PR #4865, opened 2022-01-16
Ensure uses of LOG instance are not abbreviating keys

### LINKED ISSUE #4843: Don't abbreviate hashes/public keys in logs
### Description
Currently most hashes and public keys logged by teku are truncated to reduce space. However block explorers and the REST API require the full hash/key and there's no good way to convert the truncated hash to a full one which makes incident investigation and response significantly more difficult.

Teku should log full hashes and public keys by default with only a few exceptions where the shorter version is necessary and it's unlikely the full value is required.

courtneyeh:
fixes #4843

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #4869, opened 2022-01-17
Use SSL for validator api.

rolfyone:
Because the validator api uses a bearer token, wrap the requests in ssl so that the bearer token is protected.

This will enable an MVP of a password protected validator api that allows addition and removal of validator keys, that is protected via bearer token.

## Documentation

## Changelog


infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/RestApiBuilder.java
```diff
@@ -197,7 +206,17 @@ private void configureCors(final JavalinConfig config) {
...
 
   private Server createJettyServer() {
-    final Server server = new Server(InetSocketAddress.createUnresolved(listenAddress, port));
+    final Server server = new Server();
+    final ServerConnector connector;
+    if (maybeKeystorePath.isPresent()) {
+      connector = new ServerConnector(server, getSslContextFactory());
+    } else {
+      connector = new ServerConnector(server);
+    }
```
rolfyone, line 215: This builder is currently only used by the validator api, so we won't be creating a non-ssl connector - if this api ends up in use by the beacon-api, then creating a non ssl version will be more important, but currently it's arguable that we don't even need the non ssl connector created here. Would be interested in your thoughts on removing this @ajsutton

ajsutton: hmm, could remove it and it wouldn't be the end of the world as it's easy to add back.  Would only really be worth doing if `maybeKeystorePath` stopped being an `Optional` though.  That makes it slightly harder to support non-ssl again but still wouldn't be a big deal.

Given our aim is to move everything to this I'd be inclined to keep it since we have it unless it causes us any kind of trouble.

rolfyone: I'll keep it - sounds fair :)

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/SimpleHttpClient.java
```diff
@@ -25,30 +25,43 @@
...
+              "Received unsuccessful response from %s: %s %s",
+              url, response.code(), response.message())
+          .isTrue();
+      final ResponseBody body = response.body();
+      assertThat(body).isNotNull();
+      return body.string();
+    } catch (IOException e) {
+      LOG.error("GET {} failed", path, e);
+      return "";
+    }
```
ajsutton: Hiding exceptions like this seems very weird. How do we tell the difference between a get that returns nothing and a get that fails?

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TrustingSimpleHttpsClient.java
```diff
@@ -0,0 +1,98 @@
...
+import okhttp3.OkHttpClient;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import tech.pegasys.teku.infrastructure.crypto.MessageDigestFactory;
+import tech.pegasys.teku.infrastructure.crypto.SecureRandomProvider;
+
+public class TrustingSimpleHttpsClient extends SimpleHttpClient {
+  private static final Logger LOG = LogManager.getLogger();
+  private static final TrustManager[] TRUST_ALL_CERTS =
```
ajsutton: This probably isn't `TRUST_ALL_CERTS` now that it's specifically trusting our test cert.

ajsutton (commented):
Generally looking like a good approach.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/SimpleHttpClient.java
rolfyone: Empty data will still give a parse error, I can probably remove this section now that it's all handled, but when i was debugging a random exception at this point was very un-helpful...

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TrustingSimpleHttpsClient.java
rolfyone: fair point...

ajsutton (approved):
LGTM.

Issue #4868, opened 2022-01-17
[Merge] isOptimisticSyncPossible updates


ajsutton:
### Description
The optimistic sync spec has changed slightly. Currently our `isOptimisticSyncPossible` method allows optimistic sync only if the merge block has been finalised. The latest spec allows optimistic sync if the merge block has been justified or if the block being imported is older than `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` (128 slots by default but must be user configurable) based on the current clock time.

PR #4871, opened 2022-01-18
Update to latest conditions for enabling optimistic sync

ajsutton:
## PR Description
Update conditions for enabling optimistic sync to match the latest spec. Either the merge block must be justified or the block to import must be a minimum number of slots in the past.

Will need to add a CLI option to allow overriding the number of slots before optimistic sync is enabled.

## Fixed Issue(s)
#4868 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigBuilder.java
```diff
@@ -701,6 +701,9 @@ public SpecConfigBuilder mergeBuilder(final Consumer<MergeBuilder> consumer) {
     private Bytes32 terminalBlockHash;
     private UInt64 terminalBlockHashActivationEpoch;
 
+    // Optimistic Sync
+    private int safeSlotsToImportOptimistically = 128;
```
tbenr, line 705: is this the right place to handle this constant?

storage/src/test/java/tech/pegasys/teku/storage/client/RecentChainDataTest.java
```diff
@@ -854,6 +877,86 @@ public void getAncestorsOnFork_unknownRoot() {
...
+                BlockOptions.create().setTerminalBlockHash(dataStructureUtil.randomBytes32())));
+
+    storageSystem.chainUpdater().setCurrentSlot(importBlockSlot.plus(safeSyncDistance));
+    assertThat(recentChainData.isOptimisticSyncPossible(importBlockSlot)).isTrue();
+  }
+
+  @Test
+  void isOptimisticSyncPossible_shouldBeFalseWhenMergeBlockIsOldButImportingBlockIsRecent() {
+    initPostGenesis();
+    final UInt64 safeSyncDistance = UInt64.valueOf(128);
```
tbenr: is intended to use two different ways to get `safeSyncDistance`? here: hardcoded, in previous test: from spec.

tbenr (approved):
LGTM. just a couple of questions

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigBuilder.java
ajsutton: I think so. It's a slightly odd case because we don't normally have defaults in `SpecConfig` - they all have to be in the config file.  This one is essentially a constant from the spec so won't be in the config file but must be overridable on the CLI (which I'm yet to add). We don't want the default to be in the CLI Options class as then it wouldn't be set when creating Teku through the programmatic API Anton added so I think it winds up here.

We might wind up with a constant for the default value when the CLI option comes in. So I'm going to leave it here for now and see how things pan out when adding the CLI option.

storage/src/test/java/tech/pegasys/teku/storage/client/RecentChainDataTest.java
ajsutton: Yeah that should have come from the spec too, good spot.

PR #4872, opened 2022-01-18
Restore ability to create attestations to finalized data.

### LINKED ISSUE #4870: attestation_data returns 503 for finalized epochs
### Description

curl "https://***:***@eth2-beacon-mainnet.infura.io/eth/v1/validator/attestation_data?slot=2969151&committee_index=63"
{"code":503,"message":"Beacon node is currently syncing and not serving requests"}

**Expected behavior:** 

As talked to @ajsutton, that’s a bug we’ve introduced. We’re likely rejecting requests for pretty much any finalized epoch.  Largely that API is expecting to only work with non-finalized blocks (because there’s no point attesting to things that are already finalized) but other than this one new check it does seem to work with finalized data so we should fix that.

* teku/v22.1.0

ajsutton:
## PR Description
While it's not something a validator would normally do, previously we supported creating attestations and blocks based off of finalized data. When introducing the check for optimistic blocks this was unnecessarily restricted to only non-finalized blocks.

This PR restores the ability to create attestations and blocks on finalised data by only rejecting blocks known to be optimistic instead of rejecting blocks the fork choice records as valid (which would exclude finalised blocks).

## Fixed Issue(s)
fixes #4870 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,4 +21,5 @@ For information on changes in released versions of Teku, see the [releases page]
 
 ### Bug Fixes
 * Rest api endpoints accepting validator IDs will no longer reject valid bytes48 hex strings that are not on the g2 curve.
-* Upgraded discovery to fix `ConcurrentModificationException`
+* Upgraded discovery to fix `ConcurrentModificationException`.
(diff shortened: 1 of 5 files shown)


rolfyone (approved):
LGTM

PR #4881, opened 2022-01-19
Add CLI option to override safe slots to import optimistically

### LINKED ISSUE #4868: [Merge] isOptimisticSyncPossible updates

ajsutton:
## PR Description
The optimistic sync spec requires the safe slots to import optimistically setting be user configurable (mostly in case of emergency).  This adds a `--X` option to override it.

## Fixed Issue(s)
fixes #4868 

## Documentation

## Changelog


ajsutton:
This has run into an odd quirk of how `SpecConfigBuilder` works so now has made `BELLATRIX_FORK_EPOCH` a required config item which it previously wasn't.  Will have to play around to work out what the right answer is...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigBuilder.java
```diff
@@ -225,7 +225,9 @@ private void validate() {
     validateConstant("depositContractAddress", depositContractAddress);
 
     altairBuilder.ifPresent(AltairBuilder::validate);
-    bellatrixBuilder.ifPresent(BellatrixBuilder::validate);
+    bellatrixBuilder
+        .filter(BellatrixBuilder::isBellatrixEnabled)
+        .ifPresent(BellatrixBuilder::validate);
```
ajsutton: We only require the Bellatrix options to be present if Bellatrix is actually being enabled. Previously this would happen naturally because if `BELLATRIX_FORK_EPOCH` wasn't set we'd never create the builder, but now that we have a couple of options that can be overridden the bellatrixBuilder in `SpecConfigBuilder` always winds up being created.  Rather than being really careful to avoid creating the builder unnecessarily, it's safer to only apply the validation if Bellatrix might be enabled.

We also ensure we don't create a `SpecConfigBellatrix` instance when building since we haven't validated the values, thus ensuring nothing can depend on them (as they aren't available).

We could ignore Bellatrix values if the `BELLATRIX_FORK_EPOCH` is set to far future but that doesn't fit as well with our existing config tests and they way they ensure we haven't missed values.  So now we're essentially using the presence of `BELLATRIX_FORK_EPOCH` as a sign that all the Bellatrix options should be present which seems quite reasonable.

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
```diff
@@ -93,6 +95,15 @@
...
   private UInt64 terminalBlockHashEpochOverride;
 
+  @Option(
+      names = {"--Xnetwork-safe-slots-to-import-optimistically"},
+      hidden = true,
+      paramLabel = "<NUMBER>",
+      description =
+          "Override the the number of slots that must pass before it is considered safe to optimistically import a block",
+      arity = "1")
+  private Integer safeSlotsToImportOptimistically = DEFAULT_SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY;
```
rolfyone, line 105: Given this is an integer, we might want to put validation around it not being negative, and maybe also build in an upper limit of what might be sane?

rolfyone (approved):
overall LGTM, but possibly just sanity check on the cli.

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
ajsutton: I don't think there is a sane upper limit - a big number essentially just disables optimistic sync which is potentially useful.
Added check it's not negative.

Issue #4850, opened 2022-01-11
Ability to specify different 'fee recipients' for different validator accounts running on a client


beetrooter:
### Description
As a non-pooling, non-custodial staking provider, we need to be able to specify different coinbase accounts / fee recipients based on the validator account selected to propose a block.

It will be a 1:Many relationship between fee recipients and validator accounts.

Without this, it becomes very difficult for us to distribute the priority fees to the appropriate validator owner.

ajsutton:
Yep makes sense. We will likely always require a default fee recipient be specified as there are a bunch of ways to add validators and we can't always guarantee a fee recipient are specified as part of them but definitely makes sense to have a way to override that default per validator.

I'm thinking we probably need to support reading the mapping of validator public key to fee recipient address from a YAML file and we'd periodically check that file for changes so you don't need to restart to change the configuration. Mostly I imagine that's useful for when validators are added without restarting so you can also add an override for them.

tbenr:
@beetrootkid I'd like to design how things will work, so we can agree on requirements before start implementing it.

So, my first proposal could be a yaml file similar to

```yaml
# file will be periodically read by teku, no restart will be required

# validator public key -> fee recipient ethereum address
fee_recipients:
  '0x12345...12345': '0x0000000000000000000000000000000000000001'
  '0x12345...54321': '0x0000000000000000000000000000000000000002'

# used for indices not present in fee_recipients
default_fee_recipient: "0x0000000000000000000000000000000000000003"
```

public keys are long but  allows you to prepare the file even before knowing the assigned index for a new validator.

The corresponding CLI argument could be `--validators-suggested-fee-recipient-address-file`.

I don't if makes sense to you to have the current `--validators-suggested-fee-recipient-address` as a fallback default value so `default_fee_recipient` in file can become optional (or even absent). I think it depends if you may need to change default without restarting. If instead we want to reduce potential confusion, the two CLI params could become mutual-exclusive and the file MUST include `default_fee_recipient`.

joaocenoura:
Hi guys, thanks for you input. Dropping some comments inline:

> So, my first proposal could be a yaml file similar to `<file_content>`

The structure makes sense to us. But see last comment.
Would be possible to have a flag to configure the pooling frequency and/or disable pooling?

> public keys are long but allows you to prepare the file even before knowing the assigned index for a new validator.

Perfect, it aligns with our backend too.

[...]

PR #4882, opened 2022-01-19
call `engine_forkChoiceUpdated` with PayloadAttributes using default fee recipient if proposer is not prepared

### LINKED ISSUE #4691: [Merge] call engine_forkChoiceUpdated with PayloadAttributes when current slot proposer is being prepared
as discussed here: https://github.com/ConsenSys/teku/pull/4678/files#r754035099

when `onUpdatePreparableProposers` is called we may receive the proposer of the current slot in the list.
So we should call forkchoiceupdated with payload attributes for the current slot if possible.

tbenr:
## PR Description

- call `engine_forkChoiceUpdated` with PayloadAttributes using default fee recipient if proposer is not prepared
- rename `--Xvalidators-suggested-fee-recipient-address` to `--Xvalidators-proposer-default-fee-recipient` (first step toward a multi fee-recipient support #4850)
- better separation between ForkChoiceNotifier and ForkChoiceUpdateData
- if running a BN without validator keys but with enabled rest API, it logs a RED warning at startup suggesting to specify proposer default fee recipient if a VC is going to use the node
  - an actual call to `prepare_beacon_proposer` api will print another status log RED warning

## Fixed Issue(s)
fixes #4691

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/PayloadAttributesCalculator.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/PayloadAttributesCalculator.java
@@ -162,4 +162,8 @@ public class PayloadAttributesCalculator {
                     .build())
         .collect(Collectors.toList());
   }
+
+  public boolean isProposerDefaultFeeRecipientDefined() {
+    return proposerDefaultFeeRecipient.isPresent();
+  }
 }
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/PayloadAttributesCalculator.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/PayloadAttributesCalculator.java
@@ -20,9 +20,12 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.async.eventthread.EventThread;
+import tech.pegasys.teku.infrastructure.ssz.type.Bytes20;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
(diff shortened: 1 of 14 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/PayloadAttributesCalculator.java
```diff
@@ -82,25 +92,47 @@ public void updateProposers(
...
+    }
+    if (proposerDefaultFeeRecipient.isPresent()) {
+      LOG.warn(
+          "Payload Attributes are required but current proposer hasn't prepared. Using Beacon Node's default values.");
+      return proposerDefaultFeeRecipient.get();
+    }
+    // TODO: rise a RED EVENT LOG?
+    LOG.error(
+        "Payload Attributes are required but current proposer hasn't prepared and there is no default fee recipient configured. FEES ARE GOING TO BE BURNED!");
+    return Bytes20.ZERO;
```
ajsutton: I suspect at this point failing to produce the block is probably the better approach. That way the error will propagate out to the validator client whereas this error will only show up in the beacon node logs which may not be in the VC user's control.

Actually speaking of that we need to think through how to handle these errors/warnings for people like Infura - they get worried when there are a heap of error messages in their logs but things like this may be triggered by external VC activities and not necessarily indicate a problem with the beacon node.  Simplest answer is to push any of this through `ValidatorLogger` since we have a CLI option to disable those logs.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -71,6 +71,23 @@ public void warnBellatrixParameterChanged(final String parameterName, final Stri
             Color.YELLOW));
   }
 
+  public void warnMissingProposerDefaultFeeRecipientWithRestAPIEnabled() {
+    log.warn(
+        print(
+            "No default proposer fee recipient has been specified and rest API are enabled. "
+                + "If a Validator Client is going to use this node is strongly recommended to "
+                + "configure it to avoid possible loss of fees gained proposing a block",
```
ajsutton:
```suggestion
            "No default proposer fee recipient has been specified and rest API is enabled. "
                + "If a Validator Client is going to use this node it is strongly recommended to "
                + "configure it to avoid possible loss of fees gained proposing a block",
```

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -71,6 +71,23 @@ public void warnBellatrixParameterChanged(final String parameterName, final Stri
...
+            Color.RED));
+  }
+
+  public void warnMissingProposerDefaultFeeRecipientWithPreparedBeaconProposerBeingCalled() {
+    log.warn(
+        print(
+            "Validator Client detected and NO DEFAULT PROPOSER FEE RECIPIENT!!! "
+                + "it is STRONGLY recommended to configure it to avoid possible LOSS OF FEES gained proposing a block",
```
ajsutton: I'd skip the all caps through here.

ajsutton (approved):
LGTM.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/PayloadAttributesCalculator.java
tbenr: OK, yes.
The usage of the default depends by the VC behaviour. But if the default is not configured is a bad configuration of the BN, but we are already noisy on that each time `prepare_beacon_proposer` is called.

PR #4884, opened 2022-01-20
Add test for backwards compatibility with different spec versions

ajsutton:
## PR Description
Add a test to confirm that adding or removing spec config items doesn't break backwards compatibility with previous versions.

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #4880, opened 2022-01-19
[keymanager api] make mvp available


rolfyone:
- un-hide the keymanager api options
```
Xvalidator-api-enabled
Xvalidator-api-docs-enabled
Xvalidator-api-keystore-file
Xvalidator-api-keystore-password-file
```

- investigate adding swagger-ui or similar to allow browsing the api.

- Needs to be documented in the teku online docs

- Needs to have an api published in a similar way to the beacon-api.

- Local key management is already working, and remote key management will be added post mvp.

 - check if this works with: https://github.com/joaquim-verges/eth2-keymanager-frontend as a sanity test.

 - SSL is currently enabled by default. Using a self-signed certificate is a little clumsy because they're not really a valid certificate, so we need to tweak it so we can turn SSL off if we're only listening on localhost or something like that. Something like if our listen address was 127.* or localhost, we would accept a flag to explicitly turn off ssl... this seems like a @benjaminion  question...

rolfyone:
Discussed swagger-ui availability with Adrian and decided not to offer on this API currently.

Checked against the frontend, found a small issue and raised against the frontend, basically it was sending slashing_protection: "" which was not valid in the spec, so we actively rejected that.

rolfyone:
https://github.com/ConsenSys/doc.teku/issues/336
https://github.com/ConsenSys/doc.teku/issues/334
everything for MVP is done, documentation is pending and has its own tracking. closing MVP ticket and the documentation can be tracked there.

PR #4885, opened 2022-01-20
Update storage routes for managed keys

rolfyone:
- Now that we're going to be storing remote keys, the naming model doesn't fit that well.

 - This should mean we don't have to move anything else when we make the api more publicly available, which would require upgrade scripts etc. for management.

Side issue of #4880

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #4784, opened 2021-12-12
[Merge] Only subscribe to block gossip during optimistic sync


ajsutton:
### Description
When optimistic sync is enabled, Teku should ensure it is not subscribed to any gossip except the block gossip topic. Other topics require the block to be fully validated so we are unable to forward gossip while optimistically syncing and would risk being downscored if we stayed subscribed.

On the initial startup, teku only enables gossip once the head block is close enough to the current slot to avoid this same problem because we haven't yet processed blocks.  This check needs to be changed to use the optimistic head to enable gossip, but then only subscribe to the block channel until our head block is verified.

PR #4886, opened 2022-01-20
Add ability to subscribe, unsubscribe and resubscribe to gossip topics

ajsutton:
## PR Description
Rework how gossip managers are structured so that there are explicit calls to `subscribe` rather than it being done in the constructor.  Apart from being cleaner code it also provides support for unsubscribing then resubscribing to gossip topics.

## Fixed Issue(s)
#4784 

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/AbstractGossipManager.java
```diff
@@ -57,48 +56,25 @@ protected AbstractGossipManager(
...
   @Override
-  public void shutdown() {
-    if (shutdown.compareAndSet(false, true)) {
-      // Close gossip channels
-      channel.close();
-      publisher.unsubscribe(subscriberId);
+  public void unsubscribe() {
+    if (channel.isPresent()) {
+      channel.get().close();
+      channel = Optional.empty();
```
rolfyone, line 77: could lead to concurrency fun with the subscribe, do they need to be synchronized?

ajsutton: They're synchronized back up at the `GossipForkManager` level.

ajsutton: hmm, except potentially for the ones that use a gossip publisher... Will go check those.

rolfyone (approved):
LGTM i was hoping that subscribe/unsubscribe might be enough, but its definitely safer to include the publish

PR #4887, opened 2022-01-21
Make swagger-ui usable from validator api

rolfyone:
- allow https to be turned off for localhost as a developer flag

 - add bearer authentication so swagger-ui offers the option

Partially addresses #4880

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/apis/PostKeys.java
```diff
@@ -43,6 +43,7 @@ public PostKeys(final KeyManager keyManager) {
             .operationId("ImportKeystores")
             .summary("Import Keystores")
             .description("Import keystores generated by the Eth2.0 deposit CLI tooling.")
+            .security("bearerAuth")
```
ajsutton: Maybe we should have a constant for `bearerAuth` since we wind up having to specify it in a bunch of places.  Or even just a `.securityBearerAuth` or (`.withBearerAuthSecurity()`) method.

ajsutton (approved):
LGTM. Just need to get the AT passing and would be good to avoid the `bearerAuth` duplication.

PR #4889, opened 2022-01-21
add interop keystores and passwords so that we can use interop keys to generate genesis outside of interop mode

rolfyone:

## Documentation

## Changelog


ajsutton (approved):
I don't think there's any reason to have these hidden so far down the source tree. I'd probably just create a new top level `interop-keys` directory so they're easy to find.

PR #4890, opened 2022-01-21
Make validator-api options visible on the command line interface.

rolfyone:
partially addresses #4880

Usage information in the validator client comes up as 
```
      --validator-api-cors-origins[=<origin>[,<origin>...]...]
                            Comma separated list of origins to allow, or * to
                              allow any origin
                              Default: []
      --validator-api-docs-enabled[=<BOOLEAN>]
                            Enable swagger-docs and swagger-ui endpoints
                              Default: false
      --validator-api-enabled[=<BOOLEAN>]
                            Enables Validator Rest API
                              Default: false
      --validator-api-host-allowlist[=<hostname>[,<hostname>...]...]
                            Comma-separated list of hostnames to allow, or * to
                              allow any host
                              Default: [127.0.0.1, localhost]
      --validator-api-interface=<NETWORK>
                            Interface of Validator Rest API
                              Default: 127.0.0.1
      --validator-api-keystore-file=<keystoreFile>
                            Keystore used for ssl for the validator api.
      --validator-api-keystore-password-file=<keystorePasswordFile>
                            Password used to decrypt the keystore for the
                              validator api.
      --validator-api-port=<INTEGER>
                            Port number of Rest API
                              Default: 5052
```

## Documentation

## Changelog


acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
```diff
@@ -163,11 +163,10 @@ public Config() {
     }
 
     public TekuValidatorNode.Config withValidatorApiEnabled() {
-      configMap.put("Xvalidator-api-enabled", true);
-      configMap.put("Xvalidator-api-docs-enabled", true);
```
ajsutton: We seem to have lost the docs-enabled property in this change. Was that deliberate?

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/ValidatorRestApiConfig.java
```diff
@@ -181,6 +181,11 @@ public ValidatorRestApiConfig build() {
                     "Could not access Validator api keystore file %s",
                     restApiKeystoreFile.toAbsolutePath()));
           }
+        } else {
+          if (restApiKeystoreFile != null) {
+            throw new IllegalArgumentException(
```
ajsutton: Should this be InvalidConfigurationException?

That said, we don't normally fail in cases like this because it's common to have a config file that might have the keystore set but then want to override it temporarily to disable SSL. You have to be quite explicit about disabling SSL so I don't think there's a security risk here.

ajsutton (approved):
LGTM once you get the ATs passing.

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/ValidatorRestApiConfig.java
rolfyone: sure, but its a specific issue that i had... and its really obscure when it happens, because you end up using ssl...

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
rolfyone: i had thought we wouldnt need it, but i'll have a look, because AT is broken atm

PR #4894, opened 2022-01-23
Add support for multiple fee recipients

### LINKED ISSUE #4850: Ability to specify different 'fee recipients' for different validator accounts running on a client

tbenr:
## PR Description

The PR introduces two new params for validator client:

- `--Xvalidators-proposer-config`: allows to specify a file path or a URL from which retrieve the configuration

- `--Xvalidators-proposer-config-refresh-enabled`: enables reload of the config every time VC is about to call `prepare_beacon_proposer` API (which currently happen at the beginning of each epoch)

the config is an extensible json file which allows to specify:

- a _**proposer configuration**_ for multiple validator public key.
- a default _**proposer configuration**_ for validator public keys not included in the previous mapping.

_**proposer configuration**_ currently contains only `fee_recipient`. It can be extended to include other information (ie block builder URL)

### behaviour details

- if `--Xvalidators-proposer-config` is specified,  VC during initialisation will try to load the configuration. If fails (or the json config is invalid\incorrect) VC will fail start up
- if `--Xvalidators-proposer-config-refresh-enabled` is activated, if config reload fails, the last valid configuration will be used.

### config json structure

```json
{
  "proposer_config": {
    }
  },
  "default_config": {
    "fee_recipient": "0x6e35733c5af9B61374A128e6F85f553aF09ff89A"
  }
}
```

`proposer_config` is optional
`default_config` is mandatory
`fee_recipient` is mandatory in each config

### How fee recipient will be actually chosen

- when `--Xvalidators-proposer-config` is used in VC
  - `fee_recipient` from a specific validator in the config
  - otherwise, `fee_recipient` from `default_config` in the config
  - otherwise, `--Xvalidator-proposer-default-fee-recipient` at beacon node

- when `--Xvalidators-proposer-default-fee-recipient`  is used in VC
  - `fee_recipient` from `--Xvalidators-proposer-default-fee-recipient` in VC
  - otherwise, `--Xvalidator-proposer-default-fee-recipient` at beacon node

### additional notes

currently, IO is blocking (file access as well as URL access). We have 5 threads on `validator` runner, to be evaluated if it is enough for the moment. It can be improved by migrating to async IO.

## Fixed Issue(s)
fixes #4850

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
```diff
@@ -15,63 +15,107 @@
...
+                              buildBeaconPreparableProposerList(
+                                  proposerConfig, blsPublicKeyIntegerMap))
+                      .exceptionally(
+                          throwable -> {
+                            LOG.warn(
+                                "An error occurred while obtaining proposer config", throwable);
+                            return buildBeaconPreparableProposerList(
+                                Optional.empty(), blsPublicKeyIntegerMap);
+                          })
+                      .join())
```
ajsutton: we shouldn't need a join here.  I suspect we want the first `thenApply` to actually just be a `thenCompose` which would let this return the future rather than calling join.

`thenApply` is like the future version of `Optional.map` whereas `thenCompose` is like `Optional.flatMap` if that helps the mental model at all (it may not). :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
```diff
@@ -15,63 +15,107 @@
...
+          .getValidatorIndexesByPublicKey()
           .thenApply(
-              integers ->
-                  integers.stream()
-                      .map(
-                          index ->
-                              new BeaconPreparableProposer(
-                                  UInt64.valueOf(index), getFeeRecipient()))
-                      .collect(Collectors.toList()))
+              blsPublicKeyIntegerMap ->
```
ajsutton: publicKeyToIndex map might be a better name?

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorIndexProvider.java
```diff
@@ -106,4 +107,8 @@ private void logNewValidatorIndices(final Map<BLSPublicKey, Integer> knownValida
         __ ->
             publicKeys.stream().flatMap(key -> getValidatorIndex(key).stream()).collect(toList()));
   }
+
+  public SafeFuture<Map<BLSPublicKey, Integer>> getValidatorIndexesByPublicKey() {
+    return firstSuccessfulRequest.thenApply(__ -> unmodifiableMap(validatorIndexesByPublicKey));
+  }
```
ajsutton: I'm not entirely convinced this is right - normally the idea is that you specify the public keys you want so the this provider knows what keys it should be requesting and updates them if needed. I'm very suspicious this is breaking the model somehow...

ajsutton (commented):
I'm a bit out of it after staying up late so not sure I've properly understood this. Left a few comments but will have to come back to review it properly. Sorry.

validator/client/src/test/java/tech/pegasys/teku/validator/client/proposerconfig/ProposerConfigProviderTest.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.async.StubAsyncRunner;
+import tech.pegasys.teku.infrastructure.ssz.type.Bytes20;
+import tech.pegasys.teku.validator.client.ProposerConfig;
+import tech.pegasys.teku.validator.client.proposerconfig.loader.ProposerConfigLoader;
+
+public class ProposerConfigProviderTest {
+  private static final String SOURCE = "some/file/path";
```
zilm13, line 31: Maybe it's better to use temporary file with some Java API like this to avoid any possible clashing?
 https://docs.oracle.com/javase/6/docs/api/java/io/File.html#createTempFile%28java.lang.String,%20java.lang.String%29

validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
tbenr: ah yes, most of the time I get it right but this time I fought a bit more. I was sure there must have been a clean version of it :)

tbenr: thanks for the mental model hint. Definitely helps as a shortcut without the need to replay the entire story in your head :smile:

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorIndexProvider.java
tbenr: I understand what you're saying. I don't have to rely on his internal state and I have to provide the collection of pubkeys. Even if wrt the current implementation things do not change that much (`getValidatorIndices` do not query for missing pubkeys at runtime),  it changes the assumptions of the provider.

ajsutton: Ah, so looking more carefully the key reason to pass in the actual validators you want is because there may have been some validators that were removed from active use but still have an index looked up previously and we want to avoid acting on that stale lookup info.

validator/client/src/main/java/tech/pegasys/teku/validator/client/proposerconfig/FileProposerConfigProvider.java
```diff
@@ -0,0 +1,37 @@
...
+
+package tech.pegasys.teku.validator.client.proposerconfig;
+
+import java.io.File;
+import tech.pegasys.teku.infrastructure.async.AsyncRunner;
+import tech.pegasys.teku.validator.client.ProposerConfig;
+import tech.pegasys.teku.validator.client.proposerconfig.loader.ProposerConfigLoader;
+
+public class FileProposerConfigProvider extends AbstractProposerConfigProvider {
+  private final File source;
```
ajsutton: Having specific subclasses for the config provider to handle URLs vs Files feels like overkill.  We could just use `File.toURL` to convert any file paths to URLs and then they could just be treated the same.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorIndexProvider.java
ajsutton: It may be sensible to refactor things so that `ValidatorIndexProvider` has a reference to `OwnedValidators` itself and then it wouldn't need to have the requested validators passed in which could simplify some things, but I think I'd tackle that as a separate PR as we'll have to be a bit careful about it.

validator/client/src/main/java/tech/pegasys/teku/validator/client/proposerconfig/AbstractProposerConfigProvider.java
```diff
@@ -0,0 +1,84 @@
...
+import tech.pegasys.teku.validator.client.ProposerConfig;
+import tech.pegasys.teku.validator.client.proposerconfig.loader.ProposerConfigLoader;
+
+public abstract class AbstractProposerConfigProvider implements ProposerConfigProvider {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private final boolean refresh;
+  private final AsyncRunner asyncRunner;
+  protected final ProposerConfigLoader proposerConfigLoader;
+  private Optional<ProposerConfig> lastProposerConfig = Optional.empty();
```
ajsutton, line 31: My suggestion would be for this to be stored as a `SafeFuture` then if we already have a request in progress we can return this future to "merge" the two requests rather than having to throw an error if we don't already have a config. While we probably won't get concurrent requests much, I don't think there's anything that explicitly prevents them happening so we probably want to just smoothly handle it.

It would also need to be `volatile` since it's being accessed from multiple threads.

ajsutton (commented):
Makes much more sense now I'm awake properly.  Looks good generally, mostly just tweaks to how concurrency is handled.

validator/client/src/main/java/tech/pegasys/teku/validator/client/proposerconfig/AbstractProposerConfigProvider.java
tbenr: I ended up synchronizing. I think it simplifies the handling a bit, without a huge cost (at the end the call frequency is very low).

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorIndexProvider.java
tbenr: do you intend something like: https://github.com/ConsenSys/teku/pull/4917/commits/bf349240a4c4eec1d167348dc21a85b84987b423

ajsutton: Yep that looks right. Surprised it didn't involve more changes actually...

ajsutton (approved):
LGTM.  Given the OwnedValidators change turned out to be smaller than I expected feel free to pull that into here before merging as well.

tbenr:
> LGTM. Given the OwnedValidators change turned out to be smaller than I expected feel free to pull that into here before merging as well.

I improved it by managing potential not-yet-existing indices

PR #4896, opened 2022-01-23
List validator keys on a remote signer that are in use

### LINKED ISSUE #4878: [keymanager api] list validator keys on a remote signer that are in use
### github 
This change is still in PR form, and may change slightly, but it's time to get started.

https://github.com/ethereum/keymanager-APIs/pull/17

### Story
As a maintainer of validator clients,
I want to be able to list remote keys in use by a validator client,
So that I can find keys and plan how I want them distributed.

### API
the `ValidatorRestApi` will have a new GET `/eth/v1/remotekey` endpoint added, similar to how the local GET has been defined in `GetKeys`. In the keymanager-api spec, this is listed as `ListRemoteKeys`.

There will be no query params or path params to this api, and it will respond with a body of all active keys in use that are to a remote signer. The response object is `ListRemoteKeysResponse` in the spec, and it contains a list of `SignerDefinition` under a `data` attribute.
```
{
  data: [
    {
      pubkey: "0xbadbeef",
      signerUrl: "https://signer-url"
    }
  ]
}
```
A `pubkey` is bytes48 encoded as a string hex, and the signer URL will be represented as string.

We should be able to add a call to the `ValidatorKeysAcceptanceTest` to show the new remote keys api is able to fetch an empty list, but until we add the other parts of the API we won't be able to see the responses populated in the acceptance test.

Because this is a LIST being returned as a response, in the case where we have no keys currently in use, the expectation is that we return a 200 (OK) response, with a payload that has an empty array for data, rather than a 404. There are other response types we should document in the spec, but they're largely the same as the `GetKeys`.

Swagger is a little complicated currently in the keymanager-api. it may be easiest to paste the swagger-docs into editor.swagger.io and look at the structure there (some errors may be less important than others in the editor). It'll convert to YAML pretty well and you'll be able to see the document structure.

To enable the validator-api, it's currently hidden in development options:
```
Xvalidator-api-enabled: true
Xvalidator-api-docs-enabled: true
```

courtneyeh:
fixes #4878

## Documentation

## Changelog


ethereum/core/src/main/java/tech/pegasys/teku/core/signatures/Signer.java
```diff
@@ -52,4 +54,6 @@
       ContributionAndProof contributionAndProof, ForkInfo forkInfo);
 
   boolean isLocal();
+
+  Optional<URL> getSigningServiceUrl();
```
rolfyone, line 60: @ajsutton we'll need a signer url for external signers, just checking you're happy with us adding this to the signer...

ajsutton: Yeah that's fine.  I wonder if we should then redefine `isLocal()` to be a default method that returns true if there is a signer service URL rather than overriding it in different implementations?

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/apis/GetRemoteKeys.java
```diff
@@ -0,0 +1,46 @@
...
+public class GetRemoteKeys extends RestApiEndpoint {
+
+  public static final String ROUTE = "/eth/v1/remotekey";
+  private final KeyManager keyManager;
+
+  public GetRemoteKeys(final KeyManager keyManager) {
+    super(
+        EndpointMetadata.get(ROUTE)
+            .operationId("ListRemoteKeys")
+            .summary("List Remote Keys")
```
rolfyone, line 35: This will be able to have
```
.tags(RestApiConstants.TAG_EXPERIMENTAL)
```
 once the tags PR is merged...

validator/client/src/test/resources/tech/pegasys/teku/validator/client/restapi/paths/_eth_v1_remotekeys.json
```diff
@@ -0,0 +1,60 @@
+{
+  "get" : {
+    "tags" : [ "Experimental" ],
```
rolfyone, line 3: awesome :)

rolfyone (approved):
LGTM - thanks for your patience getting this to this point, it's great!

PR #4900, opened 2022-01-24
Add debug-tools option to generate rest api.

rolfyone:
```
teku debug-tools generate-swagger-docs -o tmp
```

will currently create a `tmp/validator-api.json`. The beacon-api isn't able to generate its api without starting currently.

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/ValidatorRestApi.java
```diff
@@ -32,40 +32,46 @@
...
+                (throwable, url) -> new HttpErrorResponse(SC_BAD_REQUEST, throwable.getMessage()))
+            .exceptionHandler(
+                JsonProcessingException.class,
+                (throwable, url) -> new HttpErrorResponse(SC_BAD_REQUEST, throwable.getMessage()))
+            .endpoint(new GetKeys(keyManager))
+            .endpoint(new DeleteKeys(keyManager))
+            .endpoint(new PostKeys(keyManager))
+            .sslCertificate(
+                config.getRestApiKeystoreFile(), config.getRestApiKeystorePasswordFile());
+    // when generating rest-api docs, the keyManager may not be set
```
rolfyone: could potentially create a keymanager for use here (in debug-tools fn), but we really don't need to touch keys when we just want docs.. thoughts @ajsutton ?

ajsutton: I'd be tempted to have a `NoOpKeyManager` - nulls are never any fun.  Ideally this subcommand wouldn't actually write the bearer token to a file but it probably doesn't matter much if it does - it's only intended to run in CI and is better than having to start the full teku...

Can we only write the bearer file when the API is actually started rather than on construction or does that make life too difficult?

rolfyone: it's just locating the files that this is doing, I think its not written till startup but i'll have a play first thing in the morning...

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugToolsCommand.java
```diff
@@ -80,4 +86,53 @@ public int autocomplete(
...
+    api.getRestApiDocs()
+        .ifPresentOrElse(
+            (docs) -> {
+              final Path validatorApiPath = outputPath.resolve("validator-api.json");
+              System.out.println("Writing validator-api to " + validatorApiPath.toAbsolutePath());
+              try (FileWriter fileWriter =
+                  new FileWriter(validatorApiPath.toFile(), StandardCharsets.UTF_8)) {
+                fileWriter.write(docs);
+              } catch (IOException e) {
+                System.err.println("Failed to write validator-api.json: " + e.getMessage());
```
ajsutton: we should ensure we return a non-zero exit code on any failure (here and the `orElse` case if the api docs are unavailable), otherwise CI won't know it failed.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/NoOpKeyManager.java
```diff
@@ -0,0 +1,39 @@
...
+package tech.pegasys.teku.cli.subcommand.debug;
+
+import java.nio.file.Path;
+import tech.pegasys.teku.service.serviceutils.layout.DataDirLayout;
+import tech.pegasys.teku.validator.client.KeyManager;
+
+public class NoOpKeyManager extends KeyManager {
+  NoOpKeyManager() {
+    super(null, null);
+  }
```
ajsutton: I think this will get messy - we should just make `KeyManager` an interface so this can be a full implementation of the interface instead of having to pass `null` down.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -130,7 +130,8 @@ public static ValidatorClientService create(
           Optional.of(
               ValidatorRestApi.create(
                   validatorApiConfig,
-                  new KeyManager(validatorLoader, services.getDataDirLayout())));
+                  new ActiveKeyManager(validatorLoader),
+                  services.getDataDirLayout()));
```
ajsutton, line 134: nice. :)

ajsutton (approved):
LGTM.

PR #4901, opened 2022-01-24
Remove more Constants

ajsutton:
## PR Description
Remove some more values from `Constants`.

PowChainService uses the genesis spec config for access to config as it needs to function prior to genesis and doesn't have a concept of slots.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4902, opened 2022-01-25
allow keys to be uploaded easily to the signer using keymanager api.

rolfyone:
This is really just for setting up test instances with the interop keys here.

If the url is different to `http://localhost:19000`, it can be passed on the CLI as first argument. Bearer auth isn't currently supported, it's mostly to load the interop keys into web3signer.

## Documentation

## Changelog

### DIFF
--- a/interop-keys/.gitignore
+++ b/interop-keys/.gitignore
@@ -1,0 +1,2 @@
+payload.json
+result.json
--- a/interop-keys/.gitignore
+++ b/interop-keys/.gitignore
@@ -1,2 +1,0 @@
-payload.json
-result.json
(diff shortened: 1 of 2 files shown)


interop-keys/.gitignore
```diff
@@ -0,0 +1,2 @@
+payload.json
+result.json
```
ajsutton: This will ignore any file anywhere with these names.  We probably should use a specific path or better yet the script could just create a temp dir to put these files in and delete on exit.  A temp dir would also solve the issue that currently the script is writing these files to whatever the current directory happens to be which isn't ideal.

ajsutton:
For the record, the pattern to create a temp directory that's automatically cleaned up:

```
TEMP=`mktemp -d`

function cleanup() {
    rm -rf "${TEMP}"
}
trap cleanup EXIT
```

interop-keys/.gitignore
rolfyone: ugh ok i had assumed it worked like other ignore files. will remove that.

ajsutton (approved):
LGTM.

PR #4908, opened 2022-01-27
Allow remote keys to be added

### LINKED ISSUE #4895: [keymanager api] allow remote keys to be added

POST `/eth/v1/remotekey`
```
[
  { "pubkey": "0" },
]
```

```
{
  data: [ 
  { status: "imported"},
  ]
}
```

courtneyeh:
fixes #4895 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSource.java
```diff
@@ -101,11 +117,29 @@ public DeleteKeyResult deleteValidator(final BLSPublicKey publicKey) {
   }
 
   @Override
-  public AddLocalValidatorResult addValidator(
-      final KeyStoreData keyStoreData, final String password, final BLSPublicKey publicKey) {
+  public AddValidatorResult addValidator(
+      KeyStoreData keyStoreData, String password, BLSPublicKey publicKey) {
```
ajsutton: nit: I'd keep the `final` on these params.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSource.java
```diff
@@ -101,11 +117,29 @@ public DeleteKeyResult deleteValidator(final BLSPublicKey publicKey) {
...
+    if (!canUpdateValidators()) {
+      return new AddValidatorResult(
+          PostKeyResult.error("Cannot add validator to a read only source."), Optional.empty());
+    }
+
+    try {
+      final ValidatorProvider provider =
+          new ExternalValidatorSource.ExternalValidatorProvider(spec, publicKey);
+      externalValidatorSourceMap.put(publicKey, signerUrl);
+      return new AddValidatorResult(PostKeyResult.success(), Optional.of(provider.createSigner()));
```
ajsutton, line 126: So one detail we'll need to ensure we get right is ensuring that the created validator winds up being wrapped with slashing protection.  We might deal with that below but noting it now so I don't forget because it's very important. :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/SlashingProtectedValidatorSource.java
```diff
@@ -52,21 +53,25 @@ public DeleteKeyResult deleteValidator(final BLSPublicKey publicKey) {
...
-    return new AddLocalValidatorResult(
+    return new AddValidatorResult(
         delegateResult.getResult(),
         Optional.of(new SlashingProtectedSigner(publicKey, slashingProtector, signer)));
   }
 
+  @Override
+  public AddValidatorResult addValidator(BLSPublicKey publicKey, URL signerUrl) {
+    throw new UnsupportedOperationException();
+  }
```
ajsutton, line 73: I think this will be where we wrap in a `SlashingProtectedSigner` is basically the same way we do for `addValidator`.  Then just need to make sure we wrap the `ExternalValidatorSource` with a. `SlashingProtectedValidatorSource` when we create it.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -132,6 +134,35 @@ public synchronized PostKeyResult loadMutableValidator(
...
+      if (errorString.isPresent()) {
+        return PostKeyResult.error(errorString.get());
+      }
+    }
+    if (ownedValidators.hasValidator(publicKey)) {
+      return PostKeyResult.duplicate();
+    }
+
+    final AddValidatorResult validatorAddResult =
+            mutableValidatorSource.get().addValidator(publicKey, signerUrl);
```
ajsutton: This will need the separate `mutableExternalValidatorSource` which would also be declared as type `Optional<ValidatorSource>` but the implementation would be a `ExternalValidatorSource`.  Otherwise we'll try to add an external validator to `LocalValidatorSource` which will throw `UnsupportedOperationException`.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorSourceFactory.java
```diff
@@ -147,7 +147,9 @@ private boolean ensurePathExists(final Path directory) {
             config,
             externalSignerHttpClientFactory,
             publicKeyLoader,
-            asyncRunner);
+            asyncRunner,
+            true // TODO Check this
+            );
```
ajsutton: I suspect this instance of `ExternalValidatorSource` will set read only to false since it's loading validators from the command line and then we'd create a second instance of `ExternalValidatorSource` with read only set to true in `addMutableValidatorSource` (so it would create both a `LocalValidatorSource` to store in `mutableLocalValidatorSource` and a `ExternalValidatorSource` which would be stored in a new `mutalbeExternalValidatorSource` field here and that would also be passed to `ValidatorLoader`).

ajsutton (commented):
Mostly looking good, just got make sure when we're adding a external validator that it winds up calling into the mutable version of `ExternalValidatorSource` and not the mutable `LocalValidatorSource`.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/SlashingProtectedValidatorSource.java
rolfyone: We might need to dig into the design, there was a `mutableValidatorSource` which is basically a non read-only local validator source... we'll probably add the writable concept through `ExternalValidatorSource`, and have an instance of `ExternalValidatorSource` that's not read-only...

ajsutton: Let's catch up and I can fill you in on discussions before we give Courtney too much conflicting advice. :) There's definitely refactoring opportunities we've not taken so that we can get to something that works and know we've understood all the key interactions.

Issue #4906, opened 2022-01-26
Add commandline option constraints


jtraglia:
### Description

When poking around, I noticed that certain options that should only allow positive values (such as `--genesis-time` and `--validator-count` in `teku genesis mock`) will process negative values.

For negative genesis times, the `checkArgument` call will catch the problem and throw an exception.

```
$ teku genesis mock --genesis-time -1
java.lang.IllegalArgumentException: value (-1) must be >= 0
  <call stack>
```

But for most other options, it will continue as normal. For example, a negative validator count will not show any warnings. In this situation, `validatorCount` is effectively zero because `IntStream.range(0, -1)` is a range of zero. So it's not really a problem here.

Below are some other options that could benefit from constraints. I'm able to specify negative values for all of them.

```
--eth1-deposit-contract-max-request-size=<INTEGER>
--metrics-port=<INTEGER>
--p2p-advertised-port=<INTEGER>
--p2p-advertised-udp-port=<INTEGER>
--p2p-peer-lower-bound=<INTEGER>
--p2p-peer-upper-bound=<INTEGER>
--p2p-port=<INTEGER>
--p2p-udp-port=<INTEGER>
--rest-api-port=<INTEGER>
--validators-external-signer-timeout=<INTEGER>
--data-storage-archive-frequency=<FREQUENCY>
--number=<arg1>
```

### Possible Solution

Add validation constraints to some picocli options. Here is an example:

* https://picocli.info/#_jsr_380_beanvalidation

This would look like:

```java
@Min(value = 0, message = "Genesis time must be a positive value")
@Option(
    names = {"-t", "--genesis-time"},
    paramLabel = "<GENESIS_TIME>",
    description = "The genesis time")
private long genesisTime = System.currentTimeMillis() / 1000;
```

Note, another project dependency would be required.

```
implementation group: 'javax.validation', name: 'validation-api', version: '2.0.1.Final'
```

I would be happy to do this and submit a pull request if it's something we want to fix.

ajsutton:
For simple validations like ensuring the value is positive I think those annotations make a lot of sense. We're already getting the validation-api as a dependency (transitively via swagger and jackson) so no downside to using it.

It would be great if you took this on.

ajsutton:
[...]

PR #4909, opened 2022-01-28
Add value checks for some configuration options

jtraglia:
This PR begins to add simple validation checks to configuration options. This is still a work-in-progress. I expect to make more commits to this PR before it is approved. I'm just want to start the process now so that I know I'm going down the right path.

To identify options that needed checks, I used the following regex. This finds function signatures that return a `Builder` object and have an `int`, `long`, `OptionalInt`, or `Duration` in the parameter list. These types allow negative values but generally negative values are not wanted.

```
ag "public \w*Builder .*\(.*(\(|\s)(int|long|OptionalInt|Duration) "
```

There are lots of other configuration options that could probably use better validation. We can deal with those later. I only modified options that were clearly not meant to be zero/negative.

I throw `IllegalArgumentException` instead of `InvalidConfigurationException` because I ran into some circular dependency issues early on. I can work on fixing those problems or we might want to change these checks to `com.google.common.base.Preconditions.checkArgument` as I see that used a lot.

This is for issue #4906.

## Documentation

## Changelog


ajsutton:
Looks good generally.  You'll need to run `./gradlew spotlessApply` to fix the formatting and get the build passing (plus sign the CLA).

It would be good to be able to throw `InvalidConfigurationException` here I think so if you're up for it you could remove the dependency from `infrastructure:exception` to `infrastructure:async` in a separate PR first and that would be really nice.  And in terms of using `Precondition.checkArgument` that would keep things nice and succinct which would be good so maybe we should add an equivalent `public static void checkConfiguration(boolean, String)` method that to `InvalidConfigurationException` and then use that. Would keep things nice and succinct but let us still use the more meaningful exception type.

jtraglia:
I'm not done with this yet. I'd like to get #4910 merged first. And then I'd like to look into adding more validation checks and checking that everything works as intended. But all of the workflow checks pass now, so that's good.

ajsutton:
I think this is looking good.  With #4910 merged I think we should be able to use `InvalidConfigurationException` everywhere.  Just let me know when you're happy and I can give it a final review but looking good at the moment.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/internal/validator/tools/KeyGenerator.java
```diff
@@ -56,6 +57,10 @@ public KeyGenerator(
       final CommandSpec commandSpec,
       final Function<String, String> envSupplier,
       final Consumer<String> commandOutput) {
+    if (validatorCount < 0) {
+      throw new InvalidConfigurationException(
+          String.format("Invalid validatorCount: %d", validatorCount));
+    }
```
jtraglia, line 63: I'm unsure if this check belongs here. It's the `--number-of-validators` option in `KeyGenerationOptions`.

jtraglia:
So maybe dumb question. These three integration tests fail on my system, but they pass on CI system.

```
> Task :validator:client:integrationTest

ExternalSignerUpcheckTLSIntegrationTest > upcheckReturnsFalseWhenStatusCodeIsNot200() FAILED
    org.mockserver.client.SocketConnectionException at HttpClientConnectionErrorHandler.java:20

ExternalSignerUpcheckTLSIntegrationTest > upcheckReturnsFalseWhenServerIsDown() FAILED
    org.mockserver.client.SocketConnectionException at HttpClientConnectionErrorHandler.java:20

ExternalSignerUpcheckTLSIntegrationTest > upcheckReturnsTrueWhenStatusCodeIs200() FAILED
    org.mockserver.client.SocketConnectionException at HttpClientConnectionErrorHandler.java:20
```

I do not think these failures have anything to do with my changes. I suspect it's because I'm working on an M1 MacBook & I don't have an x86 system to test it on. I get the same results on the master branch too. I thought restarting would fix it, but it did not. Are you aware of any problems like this?

Other than this potential problem and the code block I left a comment on, I think I'm finished with this.

ajsutton:
I think the `ExternalSignerUpcheckTLSIntegrationTest` failures are because you're running Java 16 or 17. I get the same locally (CI runs java 11).  It's a limitation of the mock server's ssl handling which is using some internal sun classes. I'm hoping a future update of the library will make it work on java 16 or 17 but we may have to look at other ways to work around it at some point.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/internal/validator/tools/KeyGenerator.java
ajsutton: That value is only used by some debug subcommands which are useful for testing but not recommended for users, so doesn't have the usual config object pattern.  So I'd say this is probably the best place to put the check.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiConfig.java
```diff
@@ -134,6 +135,10 @@ public static BeaconRestApiConfigBuilder builder() {
     private BeaconRestApiConfigBuilder() {}
 
     public BeaconRestApiConfigBuilder restApiPort(final int restApiPort) {
+      if (restApiPort < 0 || restApiPort > 65535) {
```
ajsutton: We can take advantage of `PortAvailability.isPortValid` here to avoid some duplication. I'll push the change since I made it locally to check we didn't wind up with bad module dependencies again.

infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsConfig.java
```diff
@@ -118,6 +119,10 @@ public MetricsConfigBuilder metricsEnabled(boolean metricsEnabled) {
     }
 
     public MetricsConfigBuilder metricsPort(int metricsPort) {
+      if (metricsPort < 0 || metricsPort > 65535) {
```
ajsutton: Will use `PortAvailability` here too.

jtraglia:
Yes, I am running Java 17. Good to know! Glad it's not just me. In that case, I'm finished.

ajsutton (approved):
LGTM.

ajsutton:
Thanks for taking this on. It looks great - I've just replaced all the port checks with `PortAvailability.isPortValid` and assuming I haven't stuff something up in that process it should be good to merge.

jtraglia:
Great. That looks way better. Thank you.

PR #4910, opened 2022-01-28
Move exception handlers to `:infrastructure:async`

jtraglia:
`:infrastructure:execeptions` had a dependency on `:infrastructure:async` that would cause a circular dependency in some situations. For example, if `:infrastructure:metrics` included `:infrastructure:exceptions`:

```
Circular dependency between the following tasks:
:infrastructure:async:compileJava
\--- :infrastructure:metrics:compileJava
     \--- :infrastructure:exceptions:compileJava
          \--- :infrastructure:async:compileJava (*)
```

To resolve this problem, this commit moves the two exception handlers that used `async.SafeFuture` from `:infrastructure:execeptions:ExceptionUtil` to `:infrastructure:async:ExceptionHandlers`.

Also, the `:infrastructure:async` dependency has been removed from `:infrastructure:execeptions` as it is no longer required.

Related to issue #4906. This PR must be merged before #4909 can be completed.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/ExceptionHandlers.java
+++ b/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/ExceptionHandlers.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2020 ConsenSys AG.
+ *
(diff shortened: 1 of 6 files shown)


ajsutton (approved):
LGTM. Thanks for this.

PR #4912, opened 2022-01-30
Update dependencies

ajsutton:
## PR Description
Update dependencies.  Skipping the Besu update for now because it will trigger a vertx upgrade which should be handled separately.

Remove unused dependency on tuweni-toml.

## Documentation

## Changelog

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -80,7 +80,7 @@ dependencyManagement {
 
     dependency 'javax.annotation:javax.annotation-api:1.3.2'
 
-    dependencySet(group: 'org.apache.tuweni', version: '1.0.0') {
+    dependencySet(group: 'org.apache.tuweni', version: '2.1.0') {
       entry 'tuweni-bytes'
       entry 'tuweni-crypto'
       entry 'tuweni-junit'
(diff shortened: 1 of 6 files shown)


gradle/versions.gradle
```diff
@@ -52,7 +52,7 @@ dependencyManagement {
 
     dependency 'org.mock-server:mockserver-junit-jupiter:5.11.2'
 
-    dependencySet(group: 'io.swagger.core.v3', version: '2.1.11') {
+    dependencySet(group: 'io.swagger.core.v3', version: '2.1.12') {
```
rolfyone, line 55: Just be careful that swagger still renders, often its a requirement from javalin that there's a specific swagger, so updating this may break it...
I haven't checked this one specifically but in the past it'l stop rendering if we go and change it...

rolfyone: I checked mavencentral and it looks like it should be ok... I'd love a way to fail a test if it fails... will have a think

rolfyone (approved):
LGTM just cautious about swagger version

gradle/versions.gradle
ajsutton: Yep, confirmed it still renders.

PR #4914, opened 2022-01-31
Enable and disable gossip based on optimistic sync state.

ajsutton:
## PR Description
Moves away from the network periodically polling recentChainData to decide when to enable gossip to explicitly notifying it of sync state changes.
This allows gossip to be reduced to just blocks when optimistic sync is enabled and also allows unsubscribing from gossip when the node falls too far behind (such as when a laptop wakes from sleep).

## Fixed Issue(s)
#4784 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4809, opened 2021-12-20
[Merge] disable optimistic sync until  an execution payload is finalised

tbenr:
## PR Description

- handle failure in `ExecutionPayload` execution by making `ExecutePayloadResult` able to store low level failures (i.e. network exceptions).
- disable optimistic sync until an execution payload is finalised

the retry system will be part of an upcoming PR.

## Fixed Issue(s)
fixes #4781

## Documentation

- [x] I thought about documentation and added the `documentation` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #4916, opened 2022-01-31
Remove unneeded\misleading exception handling in `ForkChoicePayloadExecuror`

tbenr:
## PR Description

`ExcetutionEngineChannelImpl:executePayload` is already handling exceptions when calling the engine API (implemented in #4809) mapping them to failed result. So there is no case to intercept them in `ForkChoicePayloadExecuror::optimisticallyExecute` anymore.


ajsutton (approved):
LGTM.

PR #4917, opened 2022-01-31
rely on internal `OwnedValidator` to calculate validator indices in `ValidatorIndexProvider`

tbenr:
atop #4894 to improve `ValidatorIndexProvider`


tbenr:
embedded in #4894

PR #4919, opened 2022-01-31
Update to latest besu and vertx versions

ajsutton:
## PR Description
Update to the latest version of Besu for metrics. That updates to vertx 4 so pull in that update as well.  Fortunately we don't actually use vertx other than to create it, pass it to Besu's metrics and then close it on shutdown so no changes required beyond the version bump.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4922, opened 2022-02-01
Remove ConstantsReader

ajsutton:
## PR Description
Modifies Constants to manually load the remaining fields from a SpecConfig instance to ensure they're always in sync (and there's only three fields from the spec now so no need for all the reflection magic).

Required moving Constants to the spec.config package alongside SpecConfig to get the module dependencies to work which caused quite a lot of noise.

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #4920, opened 2022-01-31
[Merge] Update to new execution engine API


ajsutton:
### Description
Implement the new execution Engine API used in spec 1.1.9:

* [x] `executePayload` becomes `notifyNewPayload` with different response codes and meanings
* [x] `forkChoiceUpdated` can now return valid/invalid/syncing/invalid_terminal_block
* [x] Rename `receipt_root` to `receipts_root` in execution payload
* [x] ~~Block gossip rule updates (IGNORE if EL reported parent block had invalid payload)~~ Turned out we already behave correctly. There is just some room for optimization.
* [x] remove message ordering by using JSON-RPC `id`

PR #4923, opened 2022-02-01
[Bellatrix] update engine API to v1.0.0-alpha.6

tbenr:
## PR Description

renamed json rpc method `engine_executePayload` to `engine_newPayload`
renamed `ExecutePayloadResult` (internal and external) to `PayloadStatus` (`V1`)
updated `ForkChoiceUpdatedResult` to admit `PayloadStatus` data
removed JSON-RPC ordering override

updated behaviour based on the additional `PayloadStatus` values
removed `OptimisticHeadValidator` in favour of managing the optimistic head update from`forkChoiceUpdated` results.

renamed `recipientRoot` to `recipientsRoot` (and all the variants, in various places).

backward compatible with kintsugi (tested)

## Fixed Issue(s)
#4920 

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/OptimisticHeadValidator.java
```diff
@@ -108,14 +108,14 @@ private void verifyOptimisticHeads() {
             });
   }
 
-  private SafeFuture<Void> reexecutePayload(
+  private SafeFuture<Void> recallNewPayload(
```
ajsutton: I think `OptimisticHeadValidator` will wind up just getting deleted.  We don't expect the EL to execute blocks from non-canonical forks anymore, they just wind up being fully validated as part of the `forkChoiceUpdated` call if they ever become canonical.

ajsutton (commented):
I think you were just doing the API rename here in which case I think that works, but we'll also need to update our handling of the semantics (expected `ACCEPTED` from newPayload calls and dealing with valid/invalid from `forkChoiceUpdated` calls which may have some knock on effects to when we're updating the chain head and/or considering optimistic sync mode).

The other challenge we have here is that implementing this will break compatibility with Kintsugi which isn't ideal. At some point we may need to do a last release that supports Kintsugi before starting to merge this work (or just tag a specific commit/branch).

I'm also a bit in two minds about how we go about this change - if we just merge the rename we'll be in a state where we don't work with any EL because we haven't yet updated the semantics, but it could be a pretty huge PR if we try to do the semantics all at once as well.

It's probably worth doing up a draft PR with all the changes so we can see what we're dealing with. Alternatively writing up the plan in a doc would potentially work too. Then we can work out how it might be split up into sensible chunks and maybe even be able to do some of them while keeping compatibility with Kintsugi.

tbenr:
> It's probably worth doing up a draft PR with all the changes so we can see what we're dealing with. Alternatively writing up the plan in a doc would potentially work too. Then we can work out how it might be split up into sensible chunks and maybe even be able to do some of them while keeping compatibility with Kintsugi.

yes I think this is the only option, i'll be more "breaking" in coming commits

tbenr:
@ajsutton I did a first iteration on making `forkchoiceUpdated` call to feedback into `ForkChoice` itself.
Haven't validated the behaviour yet.
A bit of a trick in future management which required a @SuppressWarnings("FutureReturnValueIgnored")

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifier.java
```diff
@@ -218,6 +221,7 @@ private void internalAttestationsDue(final UInt64 slot) {
     updatePayloadAttributes(slot.plus(1));
   }
 
+  @SuppressWarnings("FutureReturnValueIgnored")
   private void sendForkChoiceUpdated() {
     forkChoiceUpdateData.send(executionEngineChannel);
   }
```
ajsutton: Ignoring the return value here is definitely wrong - we'd silently ignore any exceptions (which might include the queue being full).  Could potentially use `.reportExceptions()` here if we shouldn't ever get an error from this.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -405,6 +390,31 @@ private void notifyForkChoiceUpdatedAndOptimisticSyncingChanged() {
     }
   }
 
+  private void onExecutionPayloadResult(
+      final Bytes32 blockRoot, final PayloadStatus result, final UInt64 latestFinalizedBlockSlot) {
+    onForkChoiceThread(
+            () -> {
+              if (!result.hasValidStatus()) {
+                return;
+              }
```
ajsutton: I don't think we should be ignoring non-valid statuses.  In fact I think if we get an invalid response we need to immediately run fork choice again and update the chain head because we've potentially just re-orged over to a branch that has turned out to be invalid and want to get back off of it.

I recently fixed this in our existing code but I think it becomes more important now that the invalid response may come from fork choice updates.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionPayloadStatus.java
```diff
@@ -14,7 +14,34 @@
...
 public enum ExecutionPayloadStatus {
-  VALID,
-  INVALID,
-  SYNCING
+  VALID(Validity.VALID),
+  INVALID(Validity.INVALID),
+  SYNCING(Validity.NOT_VALIDATED),
+  ACCEPTED(Validity.NOT_VALIDATED),
+  INVALID_BLOCK_HASH(Validity.INVALID),
+  INVALID_TERMINAL_BLOCK(Validity.INVALID);
```
ajsutton, line 22: I like this approach.  Am I right in thinking that this would let our internal ForkChoice code handle either the Kintsugi style Engine API or the newer one?  If so, it might be worth considering maintaining support for both versions in the engine service.  Since we know the current version will be deleted reasonably soon we don't have to make it too neat - we could just duplicate `Web3JExecutionEngineClient` to have completely separate old and new versions.  They'd both implement the same interface which would likely use the new `newPayload` method name but the "old" implementation would still send an `engine_executePayload` request for that and it would adapt the responses as required so forkChoiceUpdated `SUCCESS` would become `VALID` etc).

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifier.java
tbenr: yeah, thought the same going bed...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionPayloadStatus.java
tbenr: it wasn't my primary goal but I see your point. I'll go down to that road and see.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
tbenr: ouch, bad rebase.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionPayloadStatus.java
tbenr: @ajsutton this my approach to kintsugi backward compatibility https://github.com/ConsenSys/teku/pull/4923/commits/464c0a3ddb85afb52fa38571291998e7f3c8a35a

ajsutton: Yeah that looks good. Would be very nice if we can manage to get this stuff merged without breaking Kintsugi.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/RandomChainBuilderForkChoiceStrategy.java
```diff
@@ -113,6 +113,11 @@ public boolean isOptimistic(final Bytes32 blockRoot) {
     return false;
   }
 
+  @Override
+  public boolean isFullyValidated(Bytes32 blockRoot) {
+    return false;
```
ajsutton: I think this should be true since it's always returning `false` for `isOptimistic`
```suggestion
    return true;
```

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifierTest.java
```diff
@@ -263,7 +269,9 @@ void onForkChoiceUpdated_shouldNotSendNotificationOfOrderedPayloadAttributes() {
 
     // current slot is 1
 
-    notifier.onForkChoiceUpdated(forkChoiceState); // calculate attributes for slot 2
+    notifier
+        .onForkChoiceUpdated(forkChoiceState)
+        .reportExceptions(); // calculate attributes for slot 2
```
ajsutton: We probably don't want to use `reportExceptions` in tests because it will log and ignore any exceptions.  Generally better to use `assertThat(...).isCompleted()` so that we know the operation did complete synchronously.

ajsutton (commented):
Looks good. Just some little details to clean up but the logic looks great.

It looks like at the moment we're always using the `KintsugiWeb3JExecutionEngineClient`? It's probably worth a follow up PR to add a `--X` CLI option to enable the kiln version so that we can begin compatibility testing ahead of kiln actually starting.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/RandomChainBuilderForkChoiceStrategy.java
tbenr: right

tbenr:
> Looks good. Just some little details to clean up but the logic looks great.
> 
> It looks like at the moment we're always using the `KintsugiWeb3JExecutionEngineClient`? It's probably worth a follow up PR to add a `--X` CLI option to enable the kiln version so that we can begin compatibility testing ahead of kiln actually starting.

It is Kintsugi, hardcoded. Yes makes total sense to add the option, I'll to that in a follow up PR, this is already too big and painful to rebase

ajsutton (approved):

PR #4924, opened 2022-02-02
Use latestValidHash value from EL

### LINKED ISSUE #4665: [Merge] Use latestValidHash value from ExecutePayloadResponse
### Description
We should take advantage of the `latestValidHash` value in `ExecutionPayloadResponse` to mark more blocks as valid/invalid when execution fails to avoid having to try executing each block.

ajsutton:
## PR Description
Allows marking more blocks valid or invalid without having to individually execute them.
Also fix an issue where blocks were left optimistic when the OptimisticHeadValidator found they were invalid.

## Fixed Issue(s)
fixes #4665 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4925, opened 2022-02-02
Use optimistic head slot when finding common ancestor

### LINKED ISSUE #4785: [Merge] Search for common ancestors starting from optimistic head
### Description
At the start of a sync or when we need to switch sync targets, we need to find a common ancestor between our chain and the target chain. Currently we search for that starting from the validated chain head, but this could be a long way behind the optimistic head, resulting in downloading a lot of blocks we have already process (but been unable to validated with the EL).

The search for common ancestor should start from the optimistic chain head instead to avoid this problem.

`MultipeerCommonAncestorFinder` is probably the place to start for this.

ajsutton:
## PR Description
Use the optimistic head slot (if there is one) rather than the fully verified head slot when finding a common ancestor to start syncing from.

Avoids re-downloading blocks we have already optimistically imported.

## Fixed Issue(s)
fixes #4785 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4926, opened 2022-02-03
Remove Constants.SLOTS_PER_EPOCH

ajsutton:
## PR Description
Remove `Constants.SLOTS_PER_EPOCH` and use the value from `SpecConfig` everywhere.

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/PendingPool.java
```diff
@@ -49,8 +48,7 @@
 
   private static final Comparator<SlotAndRoot> SLOT_AND_ROOT_COMPARATOR =
       Comparator.comparing(SlotAndRoot::getSlot).thenComparing(SlotAndRoot::getRoot);
-  private static final UInt64 DEFAULT_HISTORICAL_SLOT_TOLERANCE =
-      UInt64.valueOf(Constants.SLOTS_PER_EPOCH * 10);
+  private static final UInt64 DEFAULT_HISTORICAL_SLOT_TOLERANCE = UInt64.valueOf(320);
```
tbenr, line 51: having `DEFAULT_HISTORICAL_SLOT_TOLERANCE` as an "independent by `slotsPerEpoch`" constant is acceptable?

ajsutton: Yeah it's a fairly arbitrary number.  We'll be holding onto more stuff than we really need to in minimal config but that won't hurt anything.

The real catch with values set like this is that sometimes they wind up being set before the right value is assigned to `Constants.SLOTS_PER_EPOCH` anyway since `Constants` aren't actually constant so what this was doing is a bad idea.

tbenr (approved):
LGTM

PR #4927, opened 2022-02-03
Remove Constants.SLOTS_PER_HISTORICAL_ROOT and util module

ajsutton:
## PR Description
Remove `Constants.SLOTS_PER_HISTORICAL_ROOT` which requires moving `HistoricalBatchSchema` into `SchemaDefinitions`.  That was the last thing using `SpecDependent` so it can be removed and it was the last thing in the `util` module so finally remove it too.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #4929, opened 2022-02-03
Fixes BatchSync when EL offline

### LINKED ISSUE #4877: [Merge] Better manage sync process when EL is offline
### Description
When syncing and EL is offline we mark import batches to be invalid and drop peers.

image

```
Marking batch EventThreadOnlyBatch{delegate=SyncSourceBatch{firstSlot=203123, lastSlot=203172, count=50, complete=true, requiredParent=7409a3..e2c9, firstBlock=7b9bea..6155 (203124) (parent: 7409a3..e2c9), lastBlock=c650f3..ae62 (203172)}} as invalid because it extends from an invalid block
```

we should better handle the situation.

zilm13:
## PR Description
When `BatchSync` is trying to import batch and `BatchImport` unables to verify it because it contains `ExecutionPayload` and Execution Client is offline, `BatchSync` is paused for 5 seconds to retry the same action afterwards. Before PR `BatchSync` marked batches as invalid and penalized peers in the same case.

I called event `BatchImportResult.SERVICE_OFFLINE` because I'm not sure we should disclose EL on this level, though logs are not a part of code and could be helpful for users, so I mentioned it there. 

I don't like the way I did logging in `Web3JExecutionEngineClient` but I want good UX for users who looks at node after 2 hours of stall. I discarded `EVENT_LOG` in `BatchSync` option, where similar message repeats,  because during normal block-by-block sync `BatchSync` is not used and EL could gone offline too. Any advices how could I make this better are appreciated.

Also I'm not sure peers are not penalized during normal sync at the head of the chain when importing fresh blocks. My home chain have not synchronized yet, will check tomorrow and update.

## Fixed Issue(s)

Fixes #4877 
## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3JExecutionEngineClient.java
```diff
@@ -127,13 +127,16 @@ private static PowBlock eth1BlockToPowBlock(EthBlock.Block eth1Block) {
   }
 
   private void handleError(Throwable error) {
-    if (isErrored.compareAndSet(false, true)) {
-      EVENT_LOG.executionClientIsOffline(error);
+    final long errorTime = lastError.get();
+    if (errorTime == -1 || System.currentTimeMillis() - errorTime > ERROR_REPEAT_DELAY_MILLIS) {
```
ajsutton: Probably worth extracting the `-1` value as a `NO_ERROR_TIME` constant.

And we should also inject a `TimeProvider` instance to get the time from rather than using `System.currentTimeMillis` directly.

sync/src/main/java/tech/pegasys/teku/sync/forward/multipeer/BatchSync.java
```diff
@@ -392,6 +404,22 @@ private void onImportComplete(
         LOG.debug("Marking batch {} as invalid because it extends from an invalid block", batch);
         batch.markAsInvalid();
       }
+    } else if (result == BatchImportResult.SERVICE_OFFLINE) {
+      if (!scheduledSync) {
+        LOG.info("Sync is delayed, unable to verify blocks. Probably Execution Client is offline");
```
ajsutton:
```suggestion
        LOG.warn("Unable to import blocks because execution client is offline.");
```

ajsutton (commented):
This looks good to me.  I'd say it's worth merging this in as it fixes the problem when syncing, and then we can look at whether peers are penalised if the EL is offline when processing blocks received via gossip and add fixes for that in a separate PR if required. It's always good to keep PRs small and focussed on one thing.

zilm13:
My sync is over and I see no penalization for peers on normal sync while EL is offline, so we don't need to make additional PR

zilm13:
My fault. It's not dropping peers because of exception. I've created #4936

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3JExecutionEngineClient.java
zilm13: good point!

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoicePayloadExecutorTest.java
```diff
@@ -102,6 +102,22 @@ void optimisticallyExecute_shouldValidateMergeBlockWhenThisIsTheMergeBlock() {
...
+    final ForkChoicePayloadExecutor payloadExecutor = createPayloadExecutor();
+    final boolean execution = payloadExecutor.optimisticallyExecute(defaultPayloadHeader, payload);
+
+    // Should defer execution until it has checked the terminal difficulty so we expect getPoWBlock
+    verify(executionEngine).getPowBlock(payload.getParentHash());
+    verify(executionEngine, never()).executePayload(payload);
+    assertThat(execution).isTrue();
+    assertThat(payloadExecutor.getExecutionResult()).isCompleted();
+    final ExecutePayloadResult result = payloadExecutor.getExecutionResult().getImmediately();
+    assertThat(result.hasFailedExecution()).isTrue();
```
tbenr: this can be written using `isCompletedWithValueMatching`

sync/src/main/java/tech/pegasys/teku/sync/forward/multipeer/BatchSync.java
```diff
@@ -67,15 +72,19 @@
    */
   private UInt64 lastImportTimerStartPointSeconds;
 
+  private boolean scheduledSync = false;
```
tbenr: I'd use a more descriptive name, since the general guideline in teku is to have code self-explanatory, minimising comments.

tbenr (approved):
LGTM. only a couple of nits.

Very good first PR :)

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoicePayloadExecutorTest.java
zilm13: Nice point, have not discovered it

PR #4933, opened 2022-02-04
Add ability to use a builder when deserializing objects.

ajsutton:
## PR Description
Adds support for using a separate builder object as part of a deserializable type definition. Avoids the need to make all deserializable objects mutable just to support the framework.

## Documentation

## Changelog


rolfyone (approved):
LGTM - this is a huge step towards being able to use this in the beacon node, it's great!

PR #4934, opened 2022-02-04
Create Post Remote Keys Request

courtneyeh:
#4895 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/ValidatorTypes.java
```diff
@@ -116,6 +119,48 @@
           .withField("data", listOf(REMOTE_KEY), Function.identity())
           .build();
 
+  public static DeserializableTypeDefinition<URL> URL_TYPE =
+      DeserializableTypeDefinition.string(URL.class)
+          .name("Signer")
+          .formatter(URL::toString)
+          .parser(
+              url -> { // TODO Fix
```
ajsutton: TODO?

ajsutton (approved):
LGTM. Just need to remove the TODO comment which I think has in fact been fixed.

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/ValidatorTypes.java
```diff
@@ -116,6 +119,48 @@
...
+          .formatter(URL::toString)
+          .parser(
+              url -> {
+                try {
+                  return new URL(url);
+                } catch (MalformedURLException e) {
+                  throw new IllegalArgumentException(e);
+                }
+              })
+          .build();
```
rolfyone, line 134: lets add a negative test case around this to ValidatorTypesTest to show the exception happening... should be able to just create a test that deserializes this type and uses `/` as the url...

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/ValidatorTypes.java
```diff
@@ -116,6 +119,48 @@
...
+              PUBKEY_TYPE,
+              ExternalValidator::getPublicKey,
+              ExternalValidator::setPublicKey)
+          .withOptionalField("url", URL_TYPE, ExternalValidator::getUrl, ExternalValidator::setUrl)
+          .withField(
+              "readonly",
+              BOOLEAN_TYPE,
+              ExternalValidator::isReadOnly,
+              ExternalValidator::setReadOnly)
+          .build();
```
rolfyone, line 151: adding a test around this might be good too, just a positive test serializing and deserializing the single object.. if the deserialize requires a string for comparison like the list, it might be that the pubkey_type isn't quite right, as it should be a bytes comparison...

rolfyone: I think i'm talking about making a roundtrip test - the string comparison is just because of the way the util works i think, so we currently dont really test a roundtrip of this object

rolfyone (approved):
Overall looks really good, 2 more specific test cases and I think it's an excellent base to build on.

PR #4938, opened 2022-02-07
Use ReadOnlyForkChoiceStrategy whenever possible

ajsutton:
## PR Description
Use `ReadOnlyForkChoiceStrategy` interface whenever possible rather than the full `ForkChoiceStrategy` which allows updates. Mostly this just makes it easier to see when fork choice is being read from vs when it's actually being updated (which has stricter threading considerations).

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4940, opened 2022-02-07
Retrieve Attestation, AggregateAndProof and SignedAggregateAndProof schemas from SchemaDefinitions

ajsutton:
## PR Description
`Attestation` schema used `Constants.MAX_VALIDATORS_PER_COMMITTEE` so move it into `SchemaDefinitions` and use the value from `SpecConfig` instead.  Since `Attestation` schema is used by `AggregateAndProof` and `SignedAggregateAndProof` they also need to be retrieved from `SchemaDefinitions`.

The REST API is only affected in the way it converts to internal form (has to get the ssz schemas from `SchemaDefinitions`).  The JSON fields it declares and resulting OpenAPI schema is unchanged.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4941, opened 2022-02-07
Retrieve IndexedAttestation and AttesterSlashing schemas from SchemaDefinitions

ajsutton:
## PR Description
Retrieve `IndexedAttestation` into `SchemaDefinitions` so it can access `MAX_VALIDATORS_PER_COMMITTEE` from `SpecConfig` instead of `Constants.  `AttesterSlashing` also needs to move because it includes `IndexedAttestation`.

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #4391, opened 2021-09-19
Support publishing System metrics to a remote service


regio:
This issue extends the previous story #3773 , which initially defines the publication framework.

The goal for this story is to implement the capability of publishing `System Metrics` to a remote service.

See the [specification document](https://docs.google.com/document/d/1qPWAVRjPCENlyAjUBwGkHMvz9qLdd_6u9DPZcNxDBpc/edit#).

This is initially driven by beaconcha.in, but I think we should ideally be service agnostic: users should supply the endpoint URL that data is pushed to.

Before we consider this story done we'll also need to add tests to ensure the Prometheus metrics we're grabbing here actually do exist (ie test should fail if we ever rename any of them). That will probably need to be an acceptance test that uses a stub external metrics service to publish to and can then check it got non-zero values for everything.

ajsutton:
@regio Is this actually completely finished or was it incorrectly closed?

regio:
@ajsutton, its my mistake closing it. It still need a PR with the acceptance tests.

PR #4942, opened 2022-02-07
Add beacon node publish metrics

rolfyone:
Implement the publish mechanism for beacon node metrics.

A couple of fields are defaulted but most can be provided. Added a labelled counter for bytes sent/received on libp2p, which isn't monitoring all sent bytes, but will any time we're sending gossip which is probably a very good start.

partially addresses #4391

## Documentation

## Changelog


data/publisher/src/main/java/tech/pegasys/teku/data/publisher/PrometheusMetricsPublisherSource.java
```diff
@@ -47,14 +62,119 @@ public int getValidatorsActive() {
...
+  }
+
+  private void readExecutorCategoryItem(final Observation observation) {
     switch (observation.getMetricName()) {
       case "cpu_seconds_total":
         cpuSecondsTotal = getLongValue(observation.getValue());
         break;
       case "memory_pool_bytes_used":
         addToMemoryPoolBytesUsed((Double) observation.getValue());
         break;
```
ajsutton: Should these still be here under executor? Looks like we have them under process and jvm.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/gossip/GossipHandler.java
```diff
@@ -87,6 +105,7 @@ public GossipHandler(
       return VALIDATION_IGNORED;
     }
     LOG.trace("Received message for topic {}: {} bytes", topic, bytes.size());
+    bytesReceivedCounter.inc(bytes.size());
```
ajsutton: We should be incrementing this prior to the `!processedMessages.add` check. Even though we discard the duplicate gossip messages they still used bandwidth.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/gossip/GossipHandler.java
```diff
@@ -102,8 +121,8 @@ public void gossip(Bytes bytes) {
       // We've already gossiped this data
       return;
     }
-
     LOG.trace("Gossiping {}: {} bytes", topic, bytes.size());
+    bytesSentCounter.inc(bytes.size());
```
ajsutton: This will likely count things that the libp2p level actually sees as a duplicate and so doesn't send.  A lot of that will be caught by the `processedMessages` at this level but not all.

We also won't be counting any messages that we sent as part of rebroadcasting gossip we immediately found to be valid.

data/publisher/src/main/java/tech/pegasys/teku/data/publisher/BeaconNodeMetricData.java
```diff
@@ -13,64 +13,97 @@
...
-    this.cpu_process_seconds_total = cpuProcessSecondsTotal;
-    this.memory_process_bytes = memoryProcessBytes;
-    this.client_name = clientName;
-    this.client_version = clientVersion;
-    this.network_peers_connected = networkPeersConnected;
-    this.sync_beacon_head_slot = syncBeaconHeadSlot;
+      final long timestamp, final MetricsPublisherSource source, final File dataDirectory) {
+    super(timestamp, MetricsPublishCategories.BEACON_NODE.getDisplayName(), source);
+    if (dataDirectory != null && dataDirectory.exists() && dataDirectory.isDirectory()) {
+      this.diskBytesTotal = FileUtils.sizeOfDirectory(dataDirectory);
```
ajsutton: I worry that this is potentially quite expensive as it has to walk through the entire file tree adding up file sizes. There's just over 15,000 files in the database for our prater validator and nearly 35,000 files for the prater archive node.

I wonder if we need to cache this value and only update at a slower rate since disk usage doesn't change too fast.

ajsutton (commented):
I think the network sent/received bytes metric is probably out by a fair bit and would need to be implemented in jvm-libp2p to get accurate values.  Looking pretty good otherwise.

data/publisher/src/main/java/tech/pegasys/teku/data/publisher/PrometheusMetricsPublisherSource.java
rolfyone: probably i haven't cleaned them up, yep

data/publisher/src/main/java/tech/pegasys/teku/data/publisher/BeaconNodeMetricData.java
rolfyone: yeh the metrics call for several of these, we might need a sanity check on the interval... if its every 30 seconds it might be ok, but if its too low then this will all become very expensive...

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/gossip/GossipHandler.java
rolfyone: yep mine sits on zero as a mainnet node with no validators... is it an adequate starting point, or is it not worth having? I can just leave it at 0 if we're not happy with missing all of that?

ajsutton: I think given how misleading this is we wouldn't want to make it a metric at all.

data/publisher/src/main/java/tech/pegasys/teku/data/publisher/BeaconNodeMetricData.java
rolfyone: So `FileUtils.sizeOfDirectory` is more efficient by the looks than other things ,but with a decent number of files it really does grind to a hat. with 30k extra files, it took over 1s to compute the storage space.

We could potentially return 0 for now, or rate limit this request to be infrequent if it's taking a while - something like a service that will let the request run if it is performing well and returns a cached result if it was too slow?

For now i think 0 for this PR.

ajsutton: Agreed 0 for this PR.  I don't think I'd try to do anything fancy with caching based on performance - I'd just set it to only update every x minutes and use the cached value between those times.  I suspect x is at least 5 minutes but I'd be tempted to make it more like 15.

ajsutton (approved):
LGTM.

PR #4943, opened 2022-02-07
Remove usage of SszBitlist from schemas

ajsutton:
## PR Description
Removes usage of `SszBitlist` from REST API schemas. The REST API now uses a simple `Bytes` when declaring the API and converts to `SszBitlist` as part of `asInternalBlah` calls.

This avoids the need to specify a generic max length in the `SszBitlistSerializer` and the correct SSZ schema can be used to create the SSZ object.

The resulting OpenAPI schemas should wind up unchanged due to the use of `@Schema` on the affected fields.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4945, opened 2022-02-07
executePayload prior to validating terminal block conditions.

### LINKED ISSUE #4939: [Merge] Ensure transition block is sent to EL even if terminal conditions can't be verified
### Description
Currently in `BellatrixTransitionHelpers` when it is unable to verify that the terminal total difficulty conditions are met, it returns `SYNCING` without actually executing `executePayload` (or `newPayload` in the kiln spec).  This means the block may be optimistically imported without checking the block root which is not safe.

We should executePayload first and then check terminal total difficulty condition only if the block is VALID.

ajsutton:
## PR Description
Previously SYNCING may have been returned because the parent blocks weren't available without ever calling executePayload, potentially resulting in missing the validation of the blockHash field in the execution payload.

## Fixed Issue(s)
fixes #4939 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #4948, opened 2022-02-07
[Bellatrix] New gossip rules on block validation

### LINKED ISSUE #4920: [Merge] Update to new execution engine API

tbenr:
## PR Description

atop #4923, the actual commit is: https://github.com/ConsenSys/teku/pull/4948/commits/c51b3f3b5175b772e9c2609acbffa1b40fcbc039

Implements the ignore rule on parent block execution payload validation status, as per https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/p2p-interface.md#beacon_block

*[IGNORE] The block's parent (defined by block.parent_root) passes all validation (including execution node verification of the block.body.execution_payload).*

## Fixed Issue(s)

fixes #4920

## Documentation

## Changelog


ajsutton:
I don't think this is the right change.  The aim of the new rules is that blocks CAN be gossiped even if the parent block hasn't had it's execution payload actually executed (otherwise optimistically sync'd nodes wouldn't be able to forward any gossip and would get down scored for censoring).

So this ignore rule is there to say that if we have executed the parent node and it turned out to be invalid, we shouldn't penalise peers for that but we shouldn't forward it on.  The way this code is written we wouldn't forward the gossip when the parent is optimistically sync'd because it's not fully validated and we'd then be censoring all gossip.

I think we're already implementing this ignore rule correctly because when the parent block is discovered to be invalid we remove it from protoarray and thus it becomes an unknown block.  We therefore can't process the incoming gossip because the parent is unknown and won't penalise the peer that sent it to us - I think we might say SAVE_FOR_LATER but we'd never get the parent block so I think that's fine.  We could potentially optimise a bit by keeping track of blocks that we've already found to be invalid so they can be ignored (or rejected depending on why they're invalid) without going through the pending pool.

tbenr:
Right. Started working on optimizations on a different PR

PR #4951, opened 2022-02-08
Docker image signing cross architecture

joshuafernandes:
## PR Description
Docker image signing cross architecture - this will now build and sign arm64 and amd64 images and create a manifest at the end so the old tags still work

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -412,12 +412,8 @@ tasks.register("dockerDistUntar") {
   }
 }
 
-def dockerImageName = "consensys/teku"
-def dockerVariants = [
-  "jdk17",
-  "jdk16",
-]
-def dockerPlatforms = "linux/amd64,linux/arm64"
+def dockerImage = "consensys/teku"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -412,26 +412,22 @@ tasks.register("dockerDistUntar") {
   }
 }
 
-def dockerImageName = "consensys/teku"
-def dockerVariants = [
-  "jdk17",
-  "jdk16",
-]
-def dockerPlatforms = "linux/amd64,linux/arm64"
+def dockerImage = "consensys/dctteku"
```
ajsutton: Need to switch this back 
```suggestion
def dockerImage = "consensys/teku"
```

ajsutton (approved):
LGTM. Just need to put the image name back.

build.gradle
joshuafernandes: done, cleaned up the pr description too

PR #4952, opened 2022-02-08
Specify spec consistently in database tests.

ajsutton:
## PR Description
Specify the spec to use explicitly in database tests rather than depending on the default being consistent.

Update to using Bellatrix to prove it works (and prep the way for some further tests that will be required).

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #4671, opened 2021-11-17
[Merge] Ensure merge block passes fork choice `validate_merge_block` checks


ajsutton:
### Description
During optimistic sync we may have had to skip performing the `validate_merge_block` checks to confirm the merge block matches TTD conditions.  When the execution engine has completed syncing we need to ensure we go back and revalidate the merge block for those conditions.

ajsutton:
For finalised data, I think the key to this will be having a variable in finalised storage which records the hash of the not-yet-validated finalised transition block. If null either we've validated it or there isn't one (including it not being finalised yet).

For non-finalised we may already handle it but need to check - just need to pick up when we're switching the transition block from optimistic to valid and perform the check (if if we're doing that switch because a descendant was marked as valid).

PR #4953, opened 2022-02-08
Track the execution payload of optimisically sync'd, finalized merge transition block

ajsutton:
## PR Description
So that we can validate the terminal block conditions when the EL is back online, keep a record of the execution payload for the merge transition block if it was optimistically verified and has been finalized. On disk we only store the slot of the merge block and can reload the ExecutionPayload from that.

Not done yet is adding the ability to clear the recorded payload when it has later been verified (or to actually verify it later...)

## Fixed Issue(s)
#4671 

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -87,12 +89,12 @@
   Checkpoint justifiedCheckpoint;
   Checkpoint bestJustifiedCheckpoint;
   UInt64 latestValidFinalizedSlot = UInt64.ZERO;
+  Optional<ExecutionPayload> finalizedOptimisticTransitionPayload = Optional.empty();
```
tbenr: nit: `finalizedOptimisticTransitionPayload` is always initialized in the constructor, so initialization here is redundant.

protoarray/src/main/java/tech/pegasys/teku/protoarray/BlockMetadataStore.java
```diff
@@ -43,7 +43,7 @@ void applyUpdate(
   interface HaltableNodeProcessor {
 
     static HaltableNodeProcessor fromNodeProcessor(final NodeProcessor processor) {
-      return (child, slot, parent) -> {
+      return (child, slot, parent, executionHash) -> {
```
tbenr, line 46: nit: are we using `__` on non-used lambda params?

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -625,7 +641,7 @@ VoteTracker getVote(UInt64 validatorIndex) {
     try {
       forkChoiceStrategy.processHashesInChainWhile(
           blockRoot,
-          (root, slot, parent) -> {
+          (root, slot, parent, executionHash) -> {
```
tbenr, line 644: nit: same here

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -607,6 +623,25 @@ private void updateFinalizedData(
...
               .collect(Collectors.toSet()));
     }
+
+    return optimisticTransitionPayload;
+  }
+
+  private Optional<ExecutionPayload> updateFinalizedOptimisticTransitionBlock(
+      final boolean isFinalizedOptimisticBlockRootSet,
+      final Optional<Bytes32> finalizedOptimisticTransitionBlockRoot) {
+    if (isFinalizedOptimisticBlockRootSet) {
```
tbenr, line 633: This is the reason why we need an additional boolean aside the Optional, to actually do put\delete only once.
So on DB level we can potentially remove the optimistic transition block slot. Which I guess will be done when the actual validation will be done.

tbenr (approved):
LGTM. some nits and personal reflections :)

protoarray/src/main/java/tech/pegasys/teku/protoarray/BlockMetadataStore.java
ajsutton: I'd only use that when it's the only option, and particularly then when it's actually a `Void`.  I think it makes sense to use the appropriate name for things that do have meaning.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
ajsutton: Yep exactly - the value itself may be present or absent so we either need a separate boolean to track whether it's set or to use `Optional<Optional<Bytes32>>` and the separate boolean seemed a lot neater and easier to understand.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
ajsutton: Good catch.

PR #4954, opened 2022-02-08
[Bellatrix] add `--Xee-version` param to change engine api version

tbenr:
## PR Description
Adds a new hidden CLI param `--Xee-version` that allows to change engine API dialect.
Possible values `kintsugi` (default) and `kiln`.

`kintsugi` is based on spec `v1.0.0-alpha.5`
`kiln` is based on spec `v1.0.0-alpha.6`

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionEngineOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionEngineOptions.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.cli.options;
 import static tech.pegasys.teku.config.TekuConfiguration.Builder;
 
 import picocli.CommandLine.Option;
+import tech.pegasys.teku.spec.executionengine.ExecutionEngineChannel.Version;
 
 public class ExecutionEngineOptions {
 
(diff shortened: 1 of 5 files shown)


teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionEngineOptions.java
```diff
@@ -27,7 +27,16 @@
       hidden = true)
   private String executionEngineEndpoint = null;
 
+  @Option(
+      names = {"--Xee-version"},
+      paramLabel = "<STRING>",
+      description = "Execution Engine API version. Possible values are: kintsugi (default) or kiln",
+      arity = "1",
+      hidden = true)
+  private String executionEngineVersion = null;
```
ajsutton: Ideally this would be `private Version executionEngineVersion = Version.DEFAULT_VERSION`.  Then PicoCLI will take care of validation, do case insensitive matching and display the default value.  And we then wouldn't need to use `Optional<Version>` through the code and could just require a value always be set.

tbenr: ah yes, I was not aware of that. switching.

ajsutton (approved):
LGTM.

PR #4955, opened 2022-02-08
Create RestApiEndpoint for Post Remote Keys

courtneyeh:
#4895 

## Documentation

## Changelog


validator/client/src/test/resources/tech/pegasys/teku/validator/client/restapi/schema/PostRemoteKeysRequest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "title" : "PostRemoteKeysRequest",
+  "type" : "object",
+  "properties" : {
+    "remote_keys" : {
+      "type" : "array",
+      "items" : {
+        "$ref" : "#/components/schemas/ExternalValidator"
```
rolfyone, line 8: I think this will show that we can accept a readonly flag in the request object, but we don't want that... the readonly flag will only be for response objects...

rolfyone (approved):
LGTM

PR #4959, opened 2022-02-08
Fix PicoCLI warning because of unescaped % character

ajsutton:
## PR Description
Fix PicoCLI warning:
```
[picocli WARN] Could not format 'Number of peers that should be selected randomly (default 20% of lower-bound target)' (Underlying error: Format specifier '% o'). Using raw String: '%n' format strings have not been replaced with newlines. Please ensure to escape '%' characters with another '%'.
```

It's only on a hidden option so doesn't apply to normal `teku --help` but just `teku -X`.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4960, opened 2022-02-09
Add method to find the transition block to protoarray.

ajsutton:
## PR Description
In order to ensure we validate the merge transition block we need to be able to find it in the non-finalised data so add a method to protoarray to support that.

## Fixed Issue(s)
#4671 

## Documentation

## Changelog

### DIFF
--- a/protoarray/src/test/java/tech/pegasys/teku/protoarray/ProtoArrayTest.java
+++ b/protoarray/src/test/java/tech/pegasys/teku/protoarray/ProtoArrayTest.java
@@ -469,6 +469,62 @@ class ProtoArrayTest {
     assertThat(protoArray.contains(block2a)).isFalse();
   }
 
+  @Test
+  void findMergeTransitionBlock_emptyWhenHeadNotKnown() {
+    assertThat(protoArray.findMergeTransitionBlock(block1a)).isEmpty();
+  }
+
+  @Test
+  void findMergeTransitionBlock_emptyWhenTransitionNotYetHappened() {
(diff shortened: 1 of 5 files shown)


protoarray/src/test/java/tech/pegasys/teku/protoarray/ProtoArrayTest.java
```diff
@@ -469,6 +469,62 @@ void contains_shouldNotContainDescendantsOfInvalidBlock() {
...
+    addOptimisticBlock(3, block3a, block2a);
+    assertThat(protoArray.findMergeTransitionBlock(block3a)).isEmpty();
+  }
+
+  @Test
+  void findMergeTransitionBlock_emptyWhenAfterSpecifiedHead() {
+    addOptimisticBlock(1, block1a, GENESIS_CHECKPOINT.getRoot(), Bytes32.ZERO);
+    addOptimisticBlock(2, block2a, block1a, Bytes32.ZERO);
+    addOptimisticBlock(3, block3a, block2a, Bytes32.ZERO);
+    addOptimisticBlock(3, block4a, block3a);
```
tbenr: shouldn't be slot 4?

tbenr (approved):
LGTM. just a question on test

protoarray/src/test/java/tech/pegasys/teku/protoarray/ProtoArrayTest.java
ajsutton: It doesn't actually matter to protoarray but yes that would make more sense.

PR #4961, opened 2022-02-09
Verify transition block when importing descendant

ajsutton:
## PR Description
When importing blocks that descend from the merge transition block, ensure that the transition block is fully verified before considering the imported block fully verified.

Builds on #4960 and #4953

## Fixed Issue(s)
#4671

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #4962, opened 2022-02-09
Start logging on startup earlier

### LINKED ISSUE #4339: Configure logging earlier in startup process
### Description
In a few cases we've found it problematic to handle logging because the log4j configuration is done too late in the startup process (latest example being #4324).  We should enable logging as early as possible during startup so that the logging system is available for use to as much of the init code as possible.

Suggested approach:
1. Remove `LoggingConfiguration` from `TekuConfiguration` so that it is an independent config object.
2. `LoggingOptions` would then have two methods, the existing `configure` which sets the `wireLogs` options that will remain in `TekuConfiguration` and a new `applyLoggingConfiguration` method which creates  the `LoggingConfiguration` instance and immediately calls `LoggingConfigurator.update(loggingConfig)`.
3. Call `applyLoggingConfiguration` immediately after the call to `setLogLevels` in both `BeaconNodeCommand` and `ValidatorClientCommand`.  Then logging will be available for the rest of the config creation process.

It would also be worth investigating if we can mark `logLevel` and `loggingOptions` in `BeaconNodeCommand` as `Scope.INHERIT` for PicoCLI so that they're automatically inherited by all subcommands, then call `applyLoggingConfiguration` in `setLogLevels`.  If we ensure all subcommands call `setLogLevels` they will now all benefit from logging being configured.  We could then revisit `SubCommandLogger` and get it to use logging, especially if we used a simpler default log format for subcommands that are meant to be interactive so it didn't prefix timestamp etc, essentially outputting the same as today but with the benefit of being able to override the log configuration if it is used in scripts etc.

zilm13:
## PR Description
- Moved logging startup in the beginning of node initialization sequence
- Moved `logLevel` to `LoggingOptions` + defaults it to INFO in logging builder, not sure that both changes are appreciated
- Played with `VoluntaryExitCommand` to see how subcommand logging could be presented with new logging options enabling file logging. If it's something we want to see, it could be advanced to other subcommands. If we do it, I guess we could add some documentation for this feature.

## Fixed Issue(s)

Fixes #4339 

## Documentation

## Changelog


zilm13:
Now I think that logging should be disabled by default for subcommands and enabled with any `logLevel` set. Say, if you want to see something in logs, you run it with `-l INFO`, though it looks like some magic behavior. Anyway I think we shouldn't log anything to file for subcommands by default

teku/src/test/java/tech/pegasys/teku/cli/AbstractBeaconNodeCommandTest.java
```diff
@@ -41,8 +44,24 @@
...
+        Map<String, String> environment,
+        StartAction startAction) {
+      super(outputWriter, errorWriter, environment, startAction);
+    }
+
+    @Override
+    public void startLogging() {
+      // disable logging startup, so file appenders are not really created
+    }
+  }
```
ajsutton: I wonder if instead of having to use a subclass of `BeaconNodeCommand` here we should just make the the `startLogging` method a non-static method on `LoggingConfigurator` (ie `startLogging(LoggingConfig config)` and then we can pass the instance of `LoggingConfigurator` into the `BeaconNodeCommand` constructor. For tests we'd pass a mock that does nothing.

ajsutton (commented):
I don't think we're getting any value from applying this to the submodules and it adds a fair bit more complexity.  So I think we just leave subcommands (other than validator-client) as-is.  it's still useful to have logging configured earlier in the beacon node and validator client startup so that we can log things while creating the TekuConfiguration.

teku/src/test/java/tech/pegasys/teku/cli/AbstractBeaconNodeCommandTest.java
zilm13: It could look cleaner, yeah

zilm13:
Ok, skipping subcommands, reverted those changes, though left some minor fixes in `VoluntaryExitCommand`.
Ready for review

zilm13:
Forgot to revert one small thing. Updated.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -108,11 +107,14 @@ public void run() {
         confirmExits();
       }
       getValidatorIndices(validators).forEach(this::submitExitForValidator);
-    } catch (UncheckedIOException ex) {
+    } catch (InvalidConfigurationException ex) {
       if (Throwables.getRootCause(ex) instanceof ConnectException) {
         SUB_COMMAND_LOG.error(getFailedToConnectMessage());
         System.exit(1);
       }
```
ajsutton, line 114: Could we add an `else` case here to rethrow the original exception if it isn't a `ConnectException`? Otherwise we'll silently ignore `InvalidConfigurationException` which could be quite confusing.

ajsutton (approved):
LGTM. Just one additional fix I think is worth adding to the VoluntaryExit command while we're here.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
zilm13: Nice catch, fixed. My bad I enable auto-merge, let's open another one

PR #4963, opened 2022-02-09
Block gossip optimization by early reject and ignore in `BlockValidator`

### LINKED ISSUE #4920: [Merge] Update to new execution engine API

tbenr:
## PR Description

- changed `invalidBlockRoots` from a `Set` to a `Map` holding INVALID import result
- new `executionFailedBlockRoots` `Set` of blockroots of blocks failed with `FAILED_EXECUTION_PAYLOAD_EXECUTION`

- new failure reasons:
  - `FAILED_DESCENDANT_OF_INVALID_BLOCK`: whenever `BlockManager` tries to import a block child of a block contained in `invalidBlockRoots`, it immediately returns this result.
  - if the imported block is contained `invalidBlockRoots`, we immediately return with its previous invalid reason.

- in gossip import (`validateAndImportBlock`) if block is in `invalidBlockRoots` (or is a child of one of those blocks) we immediately reject with `Block (or its parent) previously marked as invalid`. 

## Fixed Issue(s)
fixes #4920 by closing the last nice to have task

# Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Consumer;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
@@ -33,6 +34,7 @@ import tech.pegasys.teku.spec.datastructures.blocks.ReceivedBlockListener;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.logic.common.statetransition.results.BlockImportResult;
 import tech.pegasys.teku.spec.logic.common.statetransition.results.BlockImportResult.FailureReason;
+import tech.pegasys.teku.spec.logic.common.statetransition.results.BlockImportResult.FailureReason.Validity;
 import tech.pegasys.teku.statetransition.util.FutureItems;
 import tech.pegasys.teku.statetransition.util.PendingPool;
 import tech.pegasys.teku.statetransition.validation.BlockValidator;
@@ -50,7 +52,7 @@ public class BlockManager extends Service
   private final BlockValidator validator;
 
   private final FutureItems<SignedBeaconBlock> futureBlocks;
-  private final Map<Bytes32, FailureReason> invalidBlockRoots = LimitedMap.create(500);
+  private final Map<Bytes32, Validity> invalidBlockRoots = LimitedMap.create(500);
   private final Subscribers<ReceivedBlockListener> receivedBlockSubscribers =
       Subscribers.create(true);
 
@@ -86,7 +88,8 @@ public class BlockManager extends Service
   @SuppressWarnings("FutureReturnValueIgnored")
   public SafeFuture<InternalValidationResult> validateAndImportBlock(
       final SignedBeaconBlock block) {
-    SafeFuture<InternalValidationResult> validationResult = validator.validate(block);
+    SafeFuture<InternalValidationResult> validationResult =
+        validator.validate(block, this::getStrictBlockInvalidity);
     validationResult.thenAccept(
         result -> {
           if (result.code().equals(ValidationResultCode.ACCEPT)
@@ -139,7 +142,8 @@ public class BlockManager extends Service
               if (result.isSuccessful()) {
                 LOG.trace("Imported block: {}", block);
               } else {
-                switch (result.getFailureReason()) {
+                final FailureReason failureReason = result.getFailureReason();
+                switch (failureReason) {
                   case UNKNOWN_PARENT:
                     // Add to the pending pool so it is triggered once the parent is imported
                     pendingBlocks.add(block);
@@ -166,11 +170,15 @@ public class BlockManager extends Service
                         result.getFailureCause().orElse(null));
                     break;
                   default:
-                    LOG.trace(
-                        "Unable to import block for reason {}: {}",
-                        result.getFailureReason(),
-                        block);
-                    dropInvalidBlock(block, result.getFailureReason());
+                    LOG.trace("Unable to import block for reason {}: {}", failureReason, block);
+                }
+
+                switch (failureReason.validity) {
+                  case OPTIMISTIC:
+                  case INVALID:
+                    applyFailureValidity(block, failureReason.validity);
+                    break;
+                  default:
                 }
               }
             });
@@ -180,12 +188,9 @@ public class BlockManager extends Service
     if (blockIsKnown(block)) {
       return false;
     }
-    final Optional<FailureReason> failureReason = blockIsInvalid(block);
-    if (failureReason.isPresent()) {
-      dropInvalidBlock(block, failureReason.get());
-      return false;
-    }
-    return true;
+    final Optional<Validity> failureValidity = getBlockInvalidity(block);
+    failureValidity.ifPresent(validity -> applyFailureValidity(block, validity));
+    return failureValidity.map(validity -> !validity.equals(Validity.INVALID)).orElse(true);
   }
 
   private boolean blockIsKnown(final SignedBeaconBlock block) {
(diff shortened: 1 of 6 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -163,11 +170,15 @@ private void importBlockIgnoringResult(final SignedBeaconBlock block) {
...
-                    dropInvalidBlock(block);
+                    LOG.trace("Unable to import block for reason {}: {}", failureReason, block);
+                }
+
+                switch (failureReason.validity) {
+                  case OPTIMISTIC:
+                  case INVALID:
+                    applyFailureValidity(block, failureReason.validity);
+                    break;
+                  default:
```
ajsutton: I feel like we might be overthinking this. If we know the parent is invalid (either by EL or CL rules) then we don't want to try importing the child block at all and should return either that the block failed import because it's invalid (if the parent is invalid by CL rules) or `FAILED_PAYLOAD_EXECUTION` (if the parent payload failed then the child payload fails too).  If the parent is either valid or optimistic, then we just return whatever result we get when we try to execute the child.

tbenr: I start thinking that the only thing we may add is:

```java
  @SuppressWarnings("FutureReturnValueIgnored")
  public SafeFuture<InternalValidationResult> validateAndImportBlock(
      final SignedBeaconBlock block) {
    if(blockIsInvalid(block)) {
      return SafeFuture.completedFuture(InternalValidationResult.reject("Block previously marked as invalid"));
    }
    SafeFuture<InternalValidationResult> validationResult = validator.validate(block);
```

ajsutton: I'd go one level deeper to where we currently handle invalid and known blocks.  So we currently have:

```
  private SafeFuture<BlockImportResult> doImportBlock(final SignedBeaconBlock block) {
    notifyReceivedBlockSubscribers(block);
    if (!shouldImportBlock(block)) {
      return SafeFuture.completedFuture(BlockImportResult.knownBlock(block));
    }
```

That `!shouldImportBlock` is currently checking if the block is already imported or already known to be invalid. We actually want to handle three separate cases:

 1. Block is already imported, return BlockValidationResult.knownBlock as we do now
 2. Block is already known to be invalid, return the BlockValidationResult we got from it last time
 3. Parent block is already known to be invalid, return the BlockValidationResult we got from the parent (same error reason anyway - need to be careful we don't wind up with an error message that makes it look like we were trying to import the parent again). 

 If we split that into two separate cases we could then have essentially the check you're suggesting, so roughly:

```
  private SafeFuture<BlockImportResult> doImportBlock(final SignedBeaconBlock block) {
    if (isInvalid(block.getRoot)) {
      return SafeFuture.completedFuture(<previousImportResult>);
    }
    if (isInvalid(block.getParentRoot())) {
      return SafeFuture.completedFuture(<blockImportResultFromParent>);
    }
    notifyReceivedBlockSubscribers(block);
    if (isKnownBlock(block)) {
      return SafeFuture.completedFuture(BlockImportResult.knownBlock(block));
    }
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -124,87 +133,119 @@ private void importBlockIgnoringResult(final SignedBeaconBlock block) {
...
+        handleInvalidBlock(block)
+            .or(
+                () -> {
+                  notifyReceivedBlockSubscribers(block);
+                  return handleChildOfFailedExecutionPayloadBlock(block);
+                })
+            .or(() -> handleKnownBlock(block));
+    return blockImportResult
+        .map(SafeFuture::completedFuture)
+        .orElseGet(
```
tbenr: I think I can improve here by returning `Optional<SafeFuture>` in `handle*` methods.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -47,7 +50,8 @@
   private final BlockValidator validator;
 
   private final FutureItems<SignedBeaconBlock> futureBlocks;
-  private final Set<Bytes32> invalidBlockRoots = LimitedSet.create(500);
+  private final Map<Bytes32, BlockImportResult> invalidBlockRoots = LimitedMap.create(500);
+  private final Set<Bytes32> executionFailedBlockRoots = LimitedSet.create(100);
```
ajsutton: This might be a place where it's worth adding a comment to clarify exactly what a block being in these maps mean.  In particular it's a bit ambiguous if `executionFailedBlockRoots` means the EL was unavailable or if it return INVALID.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
ajsutton: Personally I think I'd just use less chaining and a more procedural approach with `if` statements to handle the optionals.  Particularly because some of the things in the `.or(` chain here actually have side effects which isn't expected in the functional style.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -124,87 +133,119 @@ private void importBlockIgnoringResult(final SignedBeaconBlock block) {
...
-  private boolean blockIsKnown(final SignedBeaconBlock block) {
-    return pendingBlocks.contains(block)
-        || futureBlocks.contains(block)
-        || recentChainData.containsBlock(block.getRoot());
+  private Optional<BlockImportResult> handleChildOfFailedExecutionPayloadBlock(
+      final SignedBeaconBlock block) {
+    if (executionFailedBlockRoots.contains(block.getParentRoot())) {
+      executionFailedBlockRoots.add(block.getRoot());
+      pendingBlocks.add(block);
+      return Optional.of(BlockImportResult.FAILED_DESCENDANT_OF_FAILED_EXECUTION_PAYLOAD_BLOCK);
```
ajsutton: I'm not sure we need this condition.  If we can't contact the EL, we won't import the block - ie we only optimistically sync the block if we actually get a SYNCING/ACCEPTED response from the EL. So then any descendant of that block is automatically moved into the pending pool and only attempts import once the parent is actually imported which is what we want. We don't want to reject descendants because the parent hasn't imported yet - they should go to the pending pool.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/block/BlockManagerTest.java
```diff
@@ -357,9 +396,57 @@ public void onBlockImported_withPendingFutureBlocks() throws Exception {
...
+
+    // EL is now alive
+    executionEngine.setPayloadStatus(PayloadStatus.VALID);
+
+    // Gossip all remaining blocks
+    blocks.subList(1, blockCount).stream()
+        .forEach(
+            block ->
+                assertImportBlockWithResult(
+                    block, BlockImportResult.FAILED_DESCENDANT_OF_FAILED_EXECUTION_PAYLOAD_BLOCK));
```
ajsutton: If we remove the descendant of failed execution payload block case from block manager, then at this point we get `BlockImportResult.FAILED_UNKNOWN_PARENT` (which is `IGNORE` in terms of gossip) and with just that change this test then passes.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
ajsutton: So for this method I wind up with:
```
private SafeFuture<BlockImportResult> doImportBlock(final SignedBeaconBlock block) {
    final Optional<BlockImportResult> invalidBlockResult = handleInvalidBlock(block);
    if (invalidBlockResult.isPresent()) {
      return SafeFuture.completedFuture(invalidBlockResult.get());
    }

    notifyReceivedBlockSubscribers(block);

    final Optional<BlockImportResult> knownBlockResult = handleKnownBlock(block);
    //noinspection OptionalIsPresent
    if (knownBlockResult.isPresent()) {
      return SafeFuture.completedFuture(knownBlockResult.get());
    }

    return blockImporter
        .importBlock(block)
        .thenPeek(
            result -> {
              if (result.isSuccessful()) {
                LOG.trace("Imported block: {}", block);
              } else {
                switch (result.getFailureReason()) {
                  case UNKNOWN_PARENT:
                    // Add to the pending pool so it is triggered once the parent is
                    // imported
                    pendingBlocks.add(block);
                    // Check if the parent was imported while we were trying to import
                    // this block and if so, remove from the pendingPool again
                    // and process now We must add the block
                    // to the pending pool before this check happens to avoid race
                    // conditions between performing the check and the parent importing.
                    if (recentChainData.containsBlock(block.getParentRoot())) {
                      pendingBlocks.remove(block);
                      importBlockIgnoringResult(block);
                    }
                    break;
                  case BLOCK_IS_FROM_FUTURE:
                    futureBlocks.add(block);
                    break;
                  case FAILED_EXECUTION_PAYLOAD_EXECUTION_SYNCING:
                    LOG.warn("Unable to import block: Execution Client is still syncing");
                    break;
                  case FAILED_EXECUTION_PAYLOAD_EXECUTION:
                    LOG.error(
                        "Unable to import block: Execution Client communication error.",
                        result.getFailureCause().orElse(null));
                    executionFailedBlockRoots.add(block.getRoot());
                    break;
                  default:
                    LOG.trace(
                        "Unable to import block for reason {}: {}",
                        result.getFailureReason(),
                        block);
                    dropInvalidBlock(block, result);
                }
              }
            });
  }
```
which is what I mean by more procedural style and it also removes the descendant of failed execution payload check.

If you prefer the functional style I'm ok with that too but would then suggest extracting the `blockImporter.importBlock(block).thenPeek(....)` section to a separate method which will help make things clearer.

ajsutton (commented):
The invalid parent stuff looks good. I think we can remove the descendant of failed execution payload handling though.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
tbenr: yes true. It just adds a specific reason for this particular case, instead of returning UNKNOWN_PARENT.
Let's reduce unnecessary lines of code, if we actually do nothing special with that return code.

tbenr:
I added invalidity propagation in `validateAndImportBlock` too, so we early reject all invalid blocks **and all the descendants**, without trying to validate\import them.
It seems reasonable to me.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -138,14 +181,14 @@ private void importBlockIgnoringResult(final SignedBeaconBlock block) {
               } else {
                 switch (result.getFailureReason()) {
                   case UNKNOWN_PARENT:
-                    // Add to the pending pool so it is triggered once the parent is imported
+                    // Add to the pending pool so it is triggered once the parent is
+                    // imported
```
ajsutton: Might exceed line limit but worth checking if we can rejoin this comment (sometimes it gets wrapped and "sticks" like that when it would fit to join).

ajsutton (approved):
LGTM.

PR #4967, opened 2022-02-10
Make CLI options not hidden for metrics publishing

### LINKED ISSUE #4391: Support publishing System metrics to a remote service

rolfyone:
fixes #4391

## Documentation

## Changelog


rolfyone:
the new help entries are:
```
      --metrics-publish-endpoint=<URL>
                             Publish metrics for node monitoring to an external
                               service
      --metrics-publish-interval=<INTEGER>
                             Interval between metric publications to the
                               external service (measured in seconds)
                               Default: 60
```

ajsutton:
It may be worth leaving them hidden for one release and put out the word that people should test it. It's one of those features that is likely to be OS and setup dependent a bit so may be worth getting a wider sample size before we call it officially supported.

rolfyone:
I can close this PR for now, or did you just want to merge post release?

ajsutton:
Probably just merge post release.  Probably due for a release early next week anyway.

ajsutton (approved):
I suspect its worth just merging this one now and we can ask people to test it out with the develop build on testnets prior to the next release.

PR #4969, opened 2022-02-10
[Bellatrix] implements `engine_exchangeTransitionConfiguration`

### LINKED ISSUE #4956: [MERGE] Use engine_exchangeTransitionConfigurationV1 to check EL config
### Description

https://github.com/ethereum/execution-apis/pull/172

Teku should periodically send `engine_exchangeTransitionConfigurationV1` requests to the EL and log warnings if the configuration doesn't match the CL config.

tbenr:
## PR Description

call `engine_exchangeTransitionConfiguration` in `TerminalPowBlockMonitor` to verify no differences between teku configuration and current configured execution engine.

Checks are:
- `TerminalTotalDifficulty` and  `TerminalBlockHash` must correspond
- remote `TerminalBlockHash` and `TerminalBlockNumber` must be both Zero or non-Zero

When one of the check fails an error log event is printed out in console.

## Fixed Issue(s)
fixes #4956

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/KintsugiWeb3JExecutionEngineClient.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/KintsugiWeb3JExecutionEngineClient.java
@@ -70,8 +70,7 @@ public class KintsugiWeb3JExecutionEngineClient extends Web3JExecutionEngineClie
   @Override
   public SafeFuture<Response<TransitionConfigurationV1>> exchangeTransitionConfiguration(
       TransitionConfigurationV1 transitionConfiguration) {
-    return SafeFuture.failedFuture(
-        new UnsupportedOperationException("Not Implemented in Kintsugi version"));
+    return SafeFuture.completedFuture(new Response<>(transitionConfiguration));
   }
 
   private Response<ForkChoiceUpdatedResult> fromKintsugiForkChoiceUpdatedResultResponse(
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/KintsugiWeb3JExecutionEngineClient.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/KintsugiWeb3JExecutionEngineClient.java
@@ -29,6 +29,7 @@ import tech.pegasys.teku.ethereum.executionlayer.client.schema.ForkChoiceUpdated
 import tech.pegasys.teku.ethereum.executionlayer.client.schema.PayloadAttributesV1;
 import tech.pegasys.teku.ethereum.executionlayer.client.schema.PayloadStatusV1;
 import tech.pegasys.teku.ethereum.executionlayer.client.schema.Response;
+import tech.pegasys.teku.ethereum.executionlayer.client.schema.TransitionConfigurationV1;
 import tech.pegasys.teku.ethereum.executionlayer.client.serialization.Bytes8Deserializer;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.ssz.type.Bytes8;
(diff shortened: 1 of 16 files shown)


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/KintsugiWeb3JExecutionEngineClient.java
```diff
@@ -66,6 +67,13 @@ public KintsugiWeb3JExecutionEngineClient(String eeEndpoint, TimeProvider timePr
     return doRequest(web3jRequest).thenApply(this::fromKintsugiForkChoiceUpdatedResultResponse);
   }
 
+  @Override
+  public SafeFuture<Response<TransitionConfigurationV1>> exchangeTransitionConfiguration(
+      TransitionConfigurationV1 transitionConfiguration) {
+    return SafeFuture.failedFuture(
+        new UnsupportedOperationException("Not Implemented in Kintsugi version"));
```
ajsutton: I wonder if in Kintsugi this should just return the given `TransitionConfigurationV1` to avoid logging errors because the request is failing? It's a little deceptive but accurately reflects the fact that on Kintsugi we just assume the EL has the same config as we do.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionEngineChannel.java
```diff
@@ -52,6 +52,12 @@
         public SafeFuture<PayloadStatus> newPayload(final ExecutionPayload executionPayload) {
           return SafeFuture.completedFuture(PayloadStatus.SYNCING);
         }
+
+        @Override
+        public SafeFuture<TransitionConfiguration> exchangeTransitionConfiguration(
+            TransitionConfiguration transitionConfiguration) {
+          return SafeFuture.completedFuture(null);
```
ajsutton: `null` is an awkward response type here.  Not sure but might be better to just return the supplied config here too.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/TerminalPowBlockMonitor.java
```diff
@@ -267,4 +265,32 @@ private void onTerminalPowBlockFound(Bytes32 blockHash) {
   private boolean notYetFound(Bytes32 blockHash) {
     return !foundTerminalBlockHash.map(blockHash::equals).orElse(false);
   }
+
+  private void callExchangeTransitionConfiguration(SpecConfigBellatrix specConfigBellatrix) {
```
ajsutton: I think the idea is that we keep polling this to ensure the config remains in sync and the EL is still available.  Probably makes sense to have the same life cycle as searching for the TTD block - once the merge is completed successfully block import will fail if the configs are different so it won't go unnoticed and we can stop doing this check as well.

Though I think we might want to start doing this checks before the merge fork actually activates which is earlier than we'd be running the `TerminalPowBlockMonitor`.

ajsutton (commented):
Looking like its heading in the right direction.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/KintsugiWeb3JExecutionEngineClient.java
tbenr: Well I actually absorb `UnsupportedOperationException` in `finish` and just log a trace, but I haven't thought about bouncing back the config. It's nicer.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/TerminalPowBlockMonitor.java
tbenr: I was definitely thinking about moving it as early as possible, so I'm more for the latter.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
```diff
@@ -194,6 +194,26 @@ public void terminalPowBlockDetected(final Bytes32 terminalBlockHash) {
         Color.GREEN);
   }
 
+  public void transitionConfiguration_TTD_TBH_mismatch(
```
ajsutton: We shouldn't be using snake case in method names. In standard java form this would be:
```suggestion
  public void transitionConfigurationTtdTbhmismatch(
```

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
```diff
@@ -194,6 +194,26 @@ public void terminalPowBlockDetected(final Bytes32 terminalBlockHash) {
...
+    final String configurationErrorEventLog =
+        String.format(
+            "Merge       *** Transition Configuration error: local TerminalTotalDifficulty and TerminalBlockHash not matching remote Execution Client values\n"
+                + "  local  configuration: %s\n"
+                + "  remote configuration: %s",
+            localConfig, remoteConfig);
+    error(configurationErrorEventLog, Color.RED);
+  }
+
+  public void transitionConfigurationRemote_TBH_TBN_inconsistency(final String remoteConfig) {
```
ajsutton:
```suggestion
  public void transitionConfigurationRemoteTbhTbninconsistency(final String remoteConfig) {
```

ajsutton (approved):
LGTM.

PR #4975, opened 2022-02-11
Provide clearer messages when execution engine payloads are missing content

ajsutton:
## PR Description
Improve the error messages we get when the EL returns something stupid - in this case `null` response to `forkChoiceUpdated`.  Previously it was an almost entirely incomprehensible exception in `CompletableFuture` due to the way futures and method references work.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4979, opened 2022-02-11
Enforce module dependency rules

ajsutton:
## PR Description
Add a custom gradle plugin to enforce simple rules about which modules are allowed to be used from other modules. Initially this just enforces that infrastructure modules only depend on other infrastructure modules.  As we continue to sort out the module layout we can add more rules to ensure we don't mess it up again in the future.

## Documentation

## Changelog


tbenr (approved):
LGTM. Cool.

PR #4980, opened 2022-02-12
JWT support for EL engine calls

### LINKED ISSUE #4958: [Merge] Support authentication for the Engine API
### Description
Support authentication for engine API as per https://github.com/ethereum/execution-apis/pull/167

arhamj:
## PR Description

Pending

- [x] Generating secret key if not present
- [x] Adding UTs

Others
1. Went with adding caching of the Jwt token
2. Retries on failure is already handled

## Fixed Issue(s)

fixes #4958

## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtAuthInterceptor.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public Response intercept(final Chain chain) throws IOException {
+    Optional<Token> optionalToken =
+        tokenProvider.token(
+            Date.from(Instant.ofEpochMilli(timeProvider.getTimeInMillis().longValue())));
+    if (optionalToken.isEmpty()) {
+      return chain.proceed(chain.request());
+    }
+    final Token token = optionalToken.get();
+    final String authHeader = String.format("%s", token.getJwtToken());
```
ajsutton: Couldn't this just be 
```suggestion
    final String authHeader = token.getJwtToken().toString();
```

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/SafeTokenProvider.java
```diff
@@ -0,0 +1,53 @@
...
+
+package tech.pegasys.teku.ethereum.executionlayer.client.auth;
+
+import java.util.Date;
+import java.util.Optional;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class SafeTokenProvider {
+  private final ReentrantLock lock = new ReentrantLock();
+  private Token token;
```
ajsutton: Probably makes sense to make this `Optional<Token>` to make it clear that it may not be set.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/SafeTokenProvider.java
```diff
@@ -0,0 +1,53 @@
...
+      lock.unlock();
+    }
+  }
+
+  private Optional<Token> refreshTokenFromSource(final Date instant) {
+    final Optional<Token> possibleToken = tokenProvider.token(instant);
+    if (possibleToken.isEmpty()) {
+      return possibleToken;
+    }
+    token = possibleToken.get();
```
ajsutton: Not entirely sure, but I think if the provider doesn't return a token we probably should clear the cached token.  If `token` was an `Optional` then we could just skip the `isEmpty()` check here and always set `token`.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/TokenProvider.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class TokenProvider {
+  private final JwtConfig jwtConfig;
+
+  public TokenProvider(final JwtConfig jwtConfig) {
+    this.jwtConfig = jwtConfig;
+  }
+
+  public Optional<Token> token(final Date instant) {
+    final long expiresInMilliseconds = jwtConfig.getExpiresInSeconds() * 1000;
```
ajsutton:
```suggestion
    final long expiresInMilliseconds = TimeUnit.SECONDS.toMillis(jwtConfig.getExpiresInSeconds());
```

gradle/versions.gradle
```diff
@@ -162,5 +162,6 @@ dependencyManagement {
 
     dependency 'org.jupnp:org.jupnp.support:2.6.0'
     dependency 'org.jupnp:org.jupnp:2.6.0'
+    dependency 'io.jsonwebtoken:jjwt:0.9.1'
```
ajsutton: This seems to be an old version - there are tags for 0.11.2.

Also, any particular reason to use this JWT library? I don't have a particular preference but it's slightly concerning it hasn't reached it's 1.0 release and there seem to be a number of potential alternatives that may be more mature. For example auth0 have a library (https://github.com/auth0/java-jwt)

teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionEngineOptions.java
```diff
@@ -36,8 +36,20 @@
       hidden = true)
   private Version executionEngineVersion = Version.DEFAULT_VERSION;
 
+  @Option(
+      hidden = true,
+      names = {"--Xjwt-secret"},
```
ajsutton: For consistency we probably should call this:
```suggestion
      names = {"--Xee-jwt-secret"},
```

gradle/versions.gradle
arhamj: [mvn repository](https://mvnrepository.com/artifact/io.jsonwebtoken/jjwt) made me believe that it was the latest version. I used this library as it had the highest number of stars. But looking at it now it does have a lot more reported issues compared to `java-jwt`. I'll check out a few libraries and update

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtAuthInterceptor.java
arhamj: Was not sure if we were going to have a prefix to the token, so just kept it this way. Will update

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/SafeTokenProvider.java
```diff
@@ -0,0 +1,50 @@
...
+  private final ReentrantLock lock = new ReentrantLock();
+  private Optional<Token> token;
+
+  private final TokenProvider tokenProvider;
+
+  public SafeTokenProvider(final TokenProvider tokenProvider) {
+    this.tokenProvider = tokenProvider;
+  }
+
+  public Optional<Token> token(final UInt64 instantInMillis) {
```
ajsutton: Small detail but it would probably be simpler to just mark this method as `synchronized` rather than using a `ReentrantLock`. We don't need any of the monitoring features of `ReentrantLock`.

ajsutton (commented):
Looking good. The build is failing because error prone wants you to use `Instant` rather than `Date`. I was wondering if that would be possible as well because things like time comparison and adding works better with `Instant`.  I'm just not sure if using `Instant` makes it too hard to work with the jwt library if it's expecting a `Date` though. If it does we could suppress the warning.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/SafeTokenProvider.java
arhamj: makes sense. will update and add tests today. Will also be updating to use the Auth0 JWT library

ethereum/executionlayer/build.gradle
```diff
@@ -10,6 +10,9 @@ dependencies {
 
   implementation 'com.google.code.gson:gson'
   implementation 'org.apache.tuweni:tuweni-units'
+  implementation 'io.jsonwebtoken:jjwt-api'
+  implementation 'io.jsonwebtoken:jjwt-impl'
+  implementation 'io.jsonwebtoken:jjwt-jackson'
```
ajsutton: I think these two should be `runtimeOnly` so they aren't on the compile class path and we're forced to only use the stable API.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoader.java
```diff
@@ -0,0 +1,56 @@
...
+    this.jwtSecretFile = jwtSecretFile;
+  }
+
+  public Key getSecretKey() {
+    return jwtSecretFile.map(this::loadSecretFromFile).orElseGet(this::generateNewSecret);
+  }
+
+  private Key generateNewSecret() {
+    LOG.info("generating new execution engine JWT secret");
+    final Key key = MacProvider.generateKey(SignatureAlgorithm.HS256);
```
ajsutton: I think the public API for this is:
```suggestion
    final Key key = Keys.secretKeyFor(SignatureAlgorithm.HS256);
```

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/SafeTokenProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public SafeTokenProvider(final TokenProvider tokenProvider) {
+    this.tokenProvider = tokenProvider;
+  }
+
+  public synchronized Optional<Token> token(final UInt64 instantInMillis) {
+    if (token.isEmpty()) {
+      return refreshTokenFromSource(instantInMillis);
+    }
+    if (token.get().isAvailableAt(instantInMillis)) {
```
ajsutton: I think we can simplify this to:

```suggestion
    if (this.token.map(token -> token.isAvailableAt(instantInMillis)).orElse(false)) {
      return token;
    }
```

So return the token if it exists and is available at the specified time.

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoaderTest.java
```diff
@@ -0,0 +1,51 @@
...
+        new JwtSecretKeyLoader(Optional.of(jwtSecretFile.toString()));
+    final Key loadedSecret = keyLoader.getSecretKey();
+    assertThat(TestHelper.secretEquals(jwtSecret, loadedSecret)).isTrue();
+  }
+
+  @Test
+  void testGetSecretKey_KeyGeneration() {
+    final JwtSecretKeyLoader keyLoader = new JwtSecretKeyLoader(Optional.empty());
+    final Key loadedSecret = keyLoader.getSecretKey();
+    assertThat(Objects.isNull(loadedSecret)).isFalse();
```
ajsutton:
```suggestion
    assertThat(loadedSecret).isNotNull();
```

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoaderTest.java
```diff
@@ -0,0 +1,51 @@
...
+    assertThat(StringUtils.isNotEmpty(Bytes.wrap(loadedSecret.getEncoded()).toHexString()))
```
ajsutton:
```suggestion
    assertThat(Bytes.wrap(loadedSecret.getEncoded()).toHexString())).isNotEmpty();
```
I think that's the right method

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/SafeTokenProviderTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Test
+  void testGetToken_TokenUpdateOnExpiry() {
+    UInt64 timeInMillis = UInt64.valueOf(System.currentTimeMillis());
+    final Optional<Token> originalToken = safeTokenProvider.token(timeInMillis);
+    assertThat(validateTokenAtInstant(originalToken, timeInMillis)).isTrue();
+    assertThat(
+            validateJwtTokenAtInstant(
+                jwtSecretKey, originalToken.get().getJwtToken(), timeInMillis))
+        .isTrue();
```
ajsutton: I'd probably move the assertions into `validateJwtTokenAtInstance` so that it can use the richer methods `assertThat` provides rather than returning a boolean.  That way we'll get much better error messages if the test fails instead of just "expected true but was false".

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/SafeTokenProviderTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    timeInMillis = timeInMillis.plus(TimeUnit.SECONDS.toMillis(JwtConfig.EXPIRES_IN_SECONDS - 1));
+    final Optional<Token> updatedToken = safeTokenProvider.token(timeInMillis);
+    assertThat(validateTokenAtInstant(updatedToken, timeInMillis)).isTrue();
+    assertThat(
+            validateJwtTokenAtInstant(jwtSecretKey, updatedToken.get().getJwtToken(), timeInMillis))
+        .isTrue();
+
+    assertThat(originalToken.equals(updatedToken)).isTrue();
```
ajsutton:
```suggestion
    assertThat(originalToken).isEqualTo(updatedToken);
```

services/executionengine/src/main/java/tech/pegasys/teku/services/executionengine/ExecutionEngineConfiguration.java
```diff
@@ -81,5 +93,11 @@ public Builder specProvider(final Spec spec) {
       this.spec = spec;
       return this;
     }
+
+    public Builder jwtSecretFile(final String jwtSecretFile) {
+      checkNotNull(jwtSecretFile);
+      this.jwtSecretFile = Optional.of(jwtSecretFile).filter(f -> !f.isBlank());
```
ajsutton: This will need to support `jwtSecretFile` being `null` since that's the default value in the cli definition. This is what's causing most of the test failures. And we can simplify a little by using the isNotBlank method from commons-lang3. So probably:
```suggestion
      this.jwtSecretFile = Optional.ofNullable(jwtSecretFile).filter(StringUtils::isNotBlank);
```

services/executionengine/src/main/java/tech/pegasys/teku/services/executionengine/ExecutionEngineConfiguration.java
```diff
@@ -59,12 +70,13 @@ public Version getVersion() {
     private Spec spec;
     private Optional<String> endpoint = Optional.empty();
     private Version version = Version.DEFAULT_VERSION;
+    private Optional<String> jwtSecretFile;
```
ajsutton: Need to provide a default value here:
```suggestion
    private Optional<String> jwtSecretFile = Optional.empty();
```

ajsutton (commented):
Looking good, thanks for your work on this.

ajsutton (approved):
LGTM. I've pushed up a couple of small changes so that we only add authentication when the kilnv2 ee version is used, and to ensure we get a clear error message if the jwt-secret file can't be read at startup.

Thanks for getting this done, it's really great work.

At this stage I think we're spec compliant and since kilnv2 isn't the default version yet I'm ok with merging it in. The next step will be testing against EL clients that support authentication to ensure it really is compatible and then working through any issue we find.

PR #4981, opened 2022-02-13
Move a couple of errors down to debug level.

### LINKED ISSUE #4907: Provide option to disable beacon node error messages caused by bad validator clients
### Description
By default the beacon node logs a number of errors to alert users to the validator client doing something invalid (submitting an invalid attestation/block, requesting attestations in the future etc).  These are very useful when the same user owns the validator client and beacon node as often the validator client can't provide the same level of information that the beacon node can.

However, when the beacon node is operated by a separate user (e.g. Infura) the error messages in the beacon node logs are just noise which makes alerting more difficult.  We need to provide a way to disable them in that case.

One option would be to include these messages with `--log-include-validator-duties-enabled`

rolfyone:
`Failed to import proposed block` moved to validator logger, which has the nett result of int becoming a console message, but it should be possible to have this as a file level log within validator logger if we need to.

Fixes #4907

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #4983, opened 2022-02-14
Use spec for correct slot when verifying finalized transition block

ajsutton:
## PR Description
Previously the spec was selected based on the slot of the block being imported, even if the transition block being verified was actually an ancestor of that block.  Now we get the slot and execution payload of the transition block so that we can use the correct slot even if the transition block has been finalised.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #4984, opened 2022-02-14
Convert some collections to use fastutil collections

jtraglia:
Converted several `Map<Integer, ?>` to `Int2ObjectMap<?>` and `Set<Integer>` to `IntSet`. These new collections are faster and more memory efficient because they use built-in types and not `Integer` objects. There were a few collections I did not convert because it just didn't make sense too.

* In two instances, I had to move `iterator.remove()` after entry references. The tests were throwing an index out of bounds exceptions. I'm unsure why using `ObjectIterator` behaves differently.
* Needed to add some `contains()` checks prior to `remove()` because in `Map` and `Set` the remove method will just return null if the value did not exist. The fastutil collections will throw an error instead.
* On the same though, converted from `get(key)` to `getDefault(key, null)` to have the same behavior.

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/PeerSubnetSubscriptions.java
```diff
@@ -138,18 +137,28 @@ public PeerScorer createScorer() {
...
+    if (count.isPresent()) {
+      return Math.max(targetSubnetSubscriberCount - count.getAsInt(), 0);
+    } else {
+      return 0;
+    }
   }
 
-  private Optional<Integer> getMinSubscriberCount() {
-    final Optional<Integer> minAttestationSubscribers =
+  private OptionalInt getMinSubscriberCount() {
```
ajsutton, line 148: I'd probably leave this as `Optional<Integer>` - `OptionalInt` still creates a new object and is missing a lot of useful methods like `map` which makes the code harder to read.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/CommitteeSubnetSubscriptions.java
```diff
@@ -50,9 +51,8 @@ public synchronized void subscribeToSubnetId(final int subnetId) {
   }
 
   public synchronized void unsubscribeFromSubnetId(final int subnetId) {
-    final RequestedSubscription subscription = subnetIdToSubscription.remove(subnetId);
-    if (subscription != null) {
-      subscription.unsubscribe();
+    if (subnetIdToSubscription.containsKey(subnetId)) {
+      subnetIdToSubscription.remove(subnetId).unsubscribe();
```
ajsutton: In my tests removing a value that doesn't exist returns `null` the same as a normal `HashMap`. As in this test passes:

```
  @Test
  void should() {
    final Int2ObjectOpenHashMap<Object> map = new Int2ObjectOpenHashMap<>();
    final Object result = map.remove(2);
    assertThat(result).isNull();
  }
```

And the Javadoc for `remove(int)` says:
```
Returns:
the old value, or the default return value if no value was present for the given key.
```

So we shouldn't need these additional `containsKey` checks.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/SyncCommitteeSubscriptionManager.java
```diff
@@ -59,7 +60,7 @@ public synchronized void subscribe(final Integer committeeSubnet, final UInt64 u
           "Update subscription of committee subnet {}, unsubscribe due at slot {}",
           committeeSubnet,
           unsubscribeSlot.toString());
-      subcommitteeToUnsubscribeSlot.put(committeeSubnet, unsubscribeSlot);
+      subcommitteeToUnsubscribeSlot.replace(committeeSubnet, unsubscribeSlot);
```
ajsutton: I think we want to stick with `put` here. While we do know there is an existing value so `replace` will wind up doing the same thing, we definitely want to put the new value, so no point in the additional contains check replace will do.

ajsutton (commented):
Looks good generally. I'd like to understand where you're seeing exceptions from the `remove` method if the key doesn't exist.  There can be some confusion between `remove(int index)` and `remove(int key)` with some collection types but as long as we use the right method there shouldn't be exceptions when removing a key that doesn't exist from a Map or removing a value from a Set that doesn't exist.

jtraglia:
So Adrian, I did some more testing on the `remove()` call during map iteration.

This works:
```java
{
  final Map<Integer, String> map = new HashMap<>();
  map.put(1, "1");
  map.put(2, "2");
  map.put(3, "3");
  final Iterator<Map.Entry<Integer, String>> iterator = map.entrySet().iterator();
  while (iterator.hasNext()) {
    final Map.Entry<Integer, String> entry = iterator.next();
    iterator.remove();
    System.out.println("key: " + entry.getKey() + " Value: " + entry.getValue());
  }
}
```
And this does not:
```java
{
  final Int2ObjectMap<String> map = new Int2ObjectOpenHashMap<>();
  map.put(1, "1");
  map.put(2, "2");
  map.put(3, "3");
  final Iterator<Int2ObjectMap.Entry<String>> iterator = map.int2ObjectEntrySet().iterator();
  while (iterator.hasNext()) {
    final Int2ObjectMap.Entry<String> entry = iterator.next();
    iterator.remove();
    System.out.println("key: " + entry.getKey() + " Value: " + entry.getValue());
  }
}
```

I see "Index -1 out of bounds for length 33" in the output. It seems that calling `interator.remove()` messes up `entry`.

I took me a while to track down this problem. Do you know why it's happening? Is this a problem in fastutil?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/SyncCommitteeSubscriptionManager.java
jtraglia: Ah good catch. I will revert this change. Sorry I forgot to mention it the PR note.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/CommitteeSubnetSubscriptions.java
jtraglia: Yes, you are correct. I just reverted this change and it works fine. I don't know why I thought this. I must have been tracking down that other `iterator.remove()` behavior difference and got confused.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/PeerSubnetSubscriptions.java
jtraglia: Gotcha. Will revert.

ajsutton:
Yeah the `iterator.remove` change makes sense to me.  The iterator is a live view of the collection - for the built-in maps `Entry` has a reference directly to the key and value so it's still available after the call to `remove` but for these the `Entry` is obviously just holding a reference to the index and so the data becomes unavailable after `remove` is called.  Makes a lot of sense to only remove the entry after we're done using it.

Not sure which behaviour is strictly correct according to the Javadoc but either way seems reasonable to me.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/PeerSubnetSubscriptions.java
jtraglia: Hmm this is actually a little harder than I thought it would be. `streamRelevantSubnets()` returns an `IntStream` because `relevantSubnets` is an `IntSet`. As a result, `min()` will return an `OptionalInt`. Should I try to convert it to an `Optional<Integer>`?

ajsutton: ah ok, stick with `OptionalInt` in that case. Just a shame they didn't put a `.map` method on it...

jtraglia:
Do you know why the "check links in changed Markdown files / markdown-link-check (pull_request)" test is failing?

ajsutton:
> Do you know why the "check links in changed Markdown files / markdown-link-check (pull_request)" test is failing?

Looks like a temporary error building the docker image.  I'm not a fan of that checks - it depends on external services being available so tends to give false failures like this. because of that, it's not required that it passes before we can merge.

ajsutton (approved):
LGTM. Thanks for this. If you're interested in carrying on, there will be places "upstream" of these changes that can low likely be switched over too.

jtraglia:
Sure. I'll take a look at those. Should I do it in this PR or a new one?

There are also other collections I would be interested in converting to fastutil:

* `Set<Integer>` -> `IntSet`
* `List<Integer>` -> `IntList`
* `Collection<Integer>` -> `IntCollection`
* `Iterable<Integer>` -> `IntIterable`
* `Function<Integer, ?>` -> `Int2ObjectFunction<?>`

Could/should I make those changes in this PR?

ajsutton:
I'd create a follow up PR - probably a few of them in the end.  Generally we try to keep PRs fairly small so they can be reviewed more easily.

PR #4987, opened 2022-02-15
Improve accept header parsing for octet-streams

### LINKED ISSUE #4978: HTTP accept quality indications not respected for REST API SSZ requests
### Description

[`Accept` quality](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) is a useful way for a REST client to indicate that SSZ is preferred but that it'll also take JSON if forced - this allows some flexibility in requesting block data from different implementations while still choosing the most efficient format supported.

Per discussion with @ajsutton , this may be an issue in Teku (and not infura)

```
# gives SSZ as expected
curl -H "Accept: application/octet-stream" https://xxx@eth2-beacon-mainnet.infura.io/eth/v2/beacon/blocks/head
# gives JSON - should give SSZ, or error
curl -H "Accept: application/octet-stream;q=1" https://xxx@eth2-beacon-mainnet.infura.io/eth/v2/beacon/blocks/head
# gives JSON - should give SSZ, or JSON with lower priority
curl -H "Accept: application/octet-stream,application/json;q=0.9" https://@eth2-beacon-mainnet.infura.io/eth/v2/beacon/blocks/head
```

rolfyone:
Any failure will fall back to `application/json`, but it's significantly better at handling accept header strings than simply an exact match on the header.

fixes #4978

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,3 +16,4 @@ For information on changes in released versions of Teku, see the [releases page]
 ### Additions and Improvements
 
 ### Bug Fixes
+ - Improved handling of the http accept header used to determine whether to send SSZ data or json for states and blocks.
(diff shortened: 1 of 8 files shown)


ajsutton (approved):
LGTM.

PR #4988, opened 2022-02-15
Mark transition block as invalid when TTD condition is not met

ajsutton:
## PR Description
When the merge transition checks fail, mark the transition block (and everything after it) as invalid rather than just the latest block we're importing.

## Fixed Issue(s)
#4671 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #4990, opened 2022-02-15
Slashing protection logs

### LINKED ISSUE #4928: Log details about slashing protection data loaded at VC startup
### Description
As a user migrating validator client, after importing slashing protection file, I'd like to see informational logs ensuring that slashing protection data are being actually loaded when validator initialises.

zilm13:
## PR Description
Added logging of validators slashing protection on startup and reload.
It looks like this. Output is only to the file, no console.
Screenshot_20220215_200244

## Fixed Issue(s)

Fixes #4928 
## Documentation

## Changelog


zilm13:
acceptance tests are falling locally too, inspecting

zilm13:
it's stalled on `forkProvider.getForkInfo(GENESIS_SLOT)` during acceptance tests

ethereum/core/src/main/java/tech/pegasys/teku/core/signatures/LocalSlashingProtector.java
```diff
@@ -41,12 +41,10 @@ public LocalSlashingProtector(
   @Override
   public synchronized SafeFuture<Boolean> maySignBlock(
       final BLSPublicKey validator, final Bytes32 genesisValidatorsRoot, final UInt64 slot) {
-    return SafeFuture.of(
-        () -> {
-          final ValidatorSigningRecord signingRecord =
-              loadSigningRecord(validator, genesisValidatorsRoot);
-          return handleResult(validator, signingRecord.maySignBlock(genesisValidatorsRoot, slot));
-        });
```
ajsutton: We want to keep the `SafeFuture.of` around this - it ensures that if any exceptions are thrown they are returned as a failed future rather than thrown as exceptions. Having it in `loadSigningRecord` is not sufficient as `writeSigningRecord` may also fail (and there may be unexpected exceptions from anywhere along the chain).

ethereum/core/src/main/java/tech/pegasys/teku/core/signatures/Signer.java
```diff
@@ -58,4 +58,8 @@ default boolean isLocal() {
   }
 
   Optional<URL> getSigningServiceUrl();
+
+  default boolean isSlashingProtectedLocally() {
```
ajsutton: nit: I'd call this `isLocalSlashingProtectionEnabled`

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -304,6 +305,10 @@ public SszBitvector randomSszBitvector(int n) {
     return SszBitvectorSchema.create(n).ofBits(bits);
   }
 
+  public BLSKeyPair randomKeyPair() {
+    return keyPairGenerator.get();
```
ajsutton: I'd just inline this - I'm not quite sure why we have the supplier pattern for public keys but definitely don't need it for key pairs.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/DefaultSlashingProtectionLogger.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.logging.log4j.Logger;
+import tech.pegasys.teku.core.signatures.SlashingProtector;
+import tech.pegasys.teku.data.signingrecord.ValidatorSigningRecord;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.spec.datastructures.state.ForkInfo;
+import tech.pegasys.teku.validator.client.ForkProvider;
+import tech.pegasys.teku.validator.client.Validator;
+
+public class DefaultSlashingProtectionLogger implements SlashingProtectionLogger {
+  private static final Logger LOG = LogManager.getLogger("Slashing Protection");
```
ajsutton: We should should just use `LogManager.getLogger()` here rather than a custom logger name. Otherwise all the package based logging rules don't match up.

I would say these messages should go via `ValidatorLogger` though so that they go to the console - it's really only useful to log this if it's easily visible since your validator will start attesting shortly afterwards.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/DefaultSlashingProtectionLogger.java
```diff
@@ -0,0 +1,113 @@
...
+
+  public DefaultSlashingProtectionLogger(
+      final SlashingProtector slashingProtector, final ForkProvider forkProvider) {
+    this.slashingProtector = slashingProtector;
+    this.forkProvider = forkProvider;
+  }
+
+  @Override
+  public void logSlashingProtection(final List<Validator> validators) {
+    final StringJoiner sj = new StringJoiner("\n");
```
ajsutton: We don't want to produce multi-line log statements as they tend to mess up log parsing.

I think we just want to reduce the amount of information we're trying to report here. I'd think we'd just have three separate log lines - one listing the shortened public keys of validators that do have slashing protection loaded, one logging the shortened public key of validators that don't have any slashing protection, and one listing validators where the slashing protection looks to be out of date. For out of date I'd say anything where the last attestation is more than 200 epochs (about a day) ago.  So something like:

```
INFO - Slashing protection loaded for validators: abc..123, def..345
WARN - No slashing protection record for validators: ghi..678, jul..890
WARN - Slashing protection last updated more than 200 epochs old for validators: xxx..yyy
```

For any of those log lines if there are going to be more than 20 validators listed, we should report the number of validators and not each individual key (otherwise we wind up logging a massively long line that doesn't mean anything).  We already have that behaviour implemented in `ValidatorLogger.formatValidators`.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/DefaultSlashingProtectionLogger.java
```diff
@@ -0,0 +1,113 @@
...
+    sj.add("Slashing Protection activated. Validators summary:");
+    forkProvider
+        .getForkInfo(GENESIS_SLOT)
```
ajsutton: This won't return until the beacon node is fully started and able to return the fork info so we need to avoid calling it here. We likely need a way to load the signing record if it exists, but not create a new one if it doesn't.

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/SlashingProtectionLoggerTest.java
```diff
@@ -0,0 +1,154 @@
...
+                validator.getPublicKey().toAbbreviatedString(), slot));
+  }
+
+  @Test
+  public void shouldPresentProtectedValidatorAttestation() {
+    final UInt64 epochSource = UInt64.valueOf(12);
+    final UInt64 epochTarget = UInt64.valueOf(20);
+    when(slashingProtector.loadSigningRecord(pubKey, forkInfo.getGenesisValidatorsRoot()))
+        .thenReturn(
+            SafeFuture.of(
```
ajsutton: Can just use `SafeFuture.completedFuture(new ValidatorSigningRecord(...`

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/DefaultSlashingProtectionLogger.java
zilm13: Yeah it's the reason it stalls in acceptance tests

zilm13:
@ajsutton 
Ok, let's try it second time.
I've made almost everything according to your review but I'm logging "Slashing protection active" instead of "Protection loaded". The reason is:
- the record is loaded only after signature is called
- on first start, nothing is loaded
- on new validator addition + config reload nothing is loaded for it
- to my mind user will be comfortable to know that all validators which are locally protected displayed somewhere


Though there are cases when your approach is better (for example, `slashingprotection` files are gone, but they will appear in outdated this way) so the final behavior is up to you, it shouldn't take much time to change.

ajsutton:
Yeah I think we want to go the other way.  The key thing is to warn when slashing protection records aren't found.  When the validator is new that will be a false alarm as it's not expected to have records yet, but that's ok.  In the case where the records have gone missing it's really important that we do provide some warning to the user.

So currently if I run with some validators so they sign and create signing records, then stop teku, delete all the slashing protection records and start again it still tells me:
```
Validator   *** Slashing protection active for validators: ab8d3a9, b7ee0ef, 86a7388, 9977f1c, a34febc, 8a298ee, ac9f4df, b245d63, 8826e82, ae241af, 9314c6d, ae00fc3, 9698d95, b89bebc, ab0bdda, b9d1d91, 84398f5, 94f0c85, aec922b, a8d4c7c… (64 total)
```

But actually none of those validators are protected because we didn't have any existing slashing protection records.   Should just be a matter of deleting this bit of code:
```
if (validatorSigningRecord.isEmpty()
            && validator.getSigner().isLocalSlashingProtectionEnabled()) {
          validatorSigningRecord = Optional.of(new ValidatorSigningRecord(Bytes32.ZERO));
        }
```
so that we don't create the signing record if it's missing.

ajsutton:
I do see what you mean about getting an outdated slashing protection message for new validators - doesn't happen on a local testnet because it's too new.  But I think saying the slashing protection is outdated is much more confusing than saying it doesn't exist.

zilm13:
@ajsutton I got your point. Also my concerns about absence of logging of local protection for external validators when local protection is not turned on were not correct as user will see them in "not loaded" warn list. So updated to focus more on loading.

ajsutton (approved):
LGTM. Thanks for your patience while we worked out the details of the requirements on this. :)

PR #4991, opened 2022-02-15
Use more fastutil collections

jtraglia:
This is an extension of #4984. It converts the following collections:

* `Set<Integer>` -> `IntSet`
* `List<Integer>` -> `IntList`
* `Collection<Integer>` -> `IntCollection`
* `Iterable<Integer>` -> `IntIterable`

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/SyncCommitteePerformanceTracker.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/SyncCommitteePerformanceTracker.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.validator.coordinator.performance;
 
+import it.unimi.dsi.fastutil.ints.IntSet;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -52,7 +53,7 @@ import tech.pegasys.teku.storage.client.CombinedChainDataClient;
 public class SyncCommitteePerformanceTracker {
   private static final Logger LOG = LogManager.getLogger();
 
-  private final NavigableMap<UInt64, Map<UInt64, Set<Integer>>>
+  private final NavigableMap<UInt64, Map<UInt64, IntSet>>
       expectedSyncCommitteeParticipantsByPeriodEndEpoch = new ConcurrentSkipListMap<>();
 
   // Slot to block root to set of indices of validators that produced a message for that slot+root
@@ -69,7 +70,7 @@ public class SyncCommitteePerformanceTracker {
   }
 
   public SafeFuture<SyncCommitteePerformance> calculatePerformance(final UInt64 epoch) {
-    final Map<UInt64, Set<Integer>> expectedSyncCommitteeParticipants =
+    final Map<UInt64, IntSet> expectedSyncCommitteeParticipants =
         getPeriodEndEpoch(epoch)
             .map(expectedSyncCommitteeParticipantsByPeriodEndEpoch::get)
             .orElse(Collections.emptyMap());
@@ -112,7 +113,7 @@ public class SyncCommitteePerformanceTracker {
 
   private SafeFuture<SyncCommitteePerformance> calculateSyncCommitteePerformance(
       final UInt64 epoch,
-      final Map<UInt64, Set<Integer>> assignedSubcommitteeIndicesByValidatorIndex,
+      final Map<UInt64, IntSet> assignedSubcommitteeIndicesByValidatorIndex,
       final Map<UInt64, Map<Bytes32, Set<UInt64>>> producingValidatorsBySlotAndBlock) {
 
     final int numberOfExpectedMessages =
@@ -184,7 +185,7 @@ public class SyncCommitteePerformanceTracker {
   }
 
   private synchronized int countIncludedMessages(
-      final Map<UInt64, Set<Integer>> assignedSubcommitteeIndicesByValidatorIndex,
+      final Map<UInt64, IntSet> assignedSubcommitteeIndicesByValidatorIndex,
       final UInt64 slot,
       final Set<UInt64> producingValidators,
       final SyncAggregate syncAggregate) {
@@ -192,7 +193,7 @@ public class SyncCommitteePerformanceTracker {
     final SszBitvector syncCommitteeBits = syncAggregate.getSyncCommitteeBits();
     int numberOfIncludedMessages = 0;
     for (UInt64 producingValidatorIndex : producingValidators) {
-      final Set<Integer> committeeIndices =
+      final IntSet committeeIndices =
           assignedSubcommitteeIndicesByValidatorIndex.get(producingValidatorIndex);
       if (committeeIndices == null) {
         LOG.debug(
@@ -201,7 +202,7 @@ public class SyncCommitteePerformanceTracker {
             slot);
         continue;
       }
-      for (Integer committeeIndex : committeeIndices) {
+      for (int committeeIndex : committeeIndices) {
         if (syncCommitteeBits.getBit(committeeIndex)) {
           numberOfIncludedMessages++;
         }
@@ -211,12 +212,12 @@ public class SyncCommitteePerformanceTracker {
   }
 
   private int countProducedMessages(
-      final Map<UInt64, Set<Integer>> assignedSubcommitteeIndicesByValidatorIndex,
+      final Map<UInt64, IntSet> assignedSubcommitteeIndicesByValidatorIndex,
       final UInt64 slot,
       final Set<UInt64> producingValidators) {
     int numberOfProducedMessages = 0;
     for (UInt64 producingValidatorIndex : producingValidators) {
-      final Set<Integer> committeeIndices =
+      final IntSet committeeIndices =
           assignedSubcommitteeIndicesByValidatorIndex.get(producingValidatorIndex);
       if (committeeIndices == null) {
         LOG.debug(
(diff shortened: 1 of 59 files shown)


jtraglia:
Weird that the assemble check is failing on CI. It works on my machine. But I'll fix it. And I'd like to convert a few more types that I missed. So this isn't ready quite yet.

ajsutton:
Ah, for the classes that define the REST API we'll need to stick with using `Set<Integer>`/`Map<Integer>` etc. They're not so commonly used so the garbage creation isn't a big deal and switching the type will cause all kinds of fun with serialisation and the way the openAPI spec comes out.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/SyncCommitteePerformanceTracker.java
```diff
@@ -201,7 +202,7 @@ private synchronized int countIncludedMessages(
             slot);
         continue;
       }
-      for (Integer committeeIndex : committeeIndices) {
+      for (int committeeIndex : committeeIndices) {
```
ajsutton: I don't think this is taking advantage of the `IntSet` - it will create a `Iterator<Integer>` which wraps each value and then the value is unwrapped back to `int`.  We'd need to use `IntIterator` directly and call `nextInt()` rather than just `next()`. In this case it may be better to use an `IntStream` instead like:
```
numberOfIncludedMessages += committeeIndices.intStream().filter(syncCommitteeBits::getBit).count();
```

data/serializer/src/main/java/tech/pegasys/teku/api/response/v1/validator/SyncCommitteeDuty.java
```diff
@@ -55,7 +56,7 @@
   public SyncCommitteeDuty(
       @JsonProperty("pubkey") final BLSPubKey pubkey,
       @JsonProperty("validator_index") final UInt64 validatorIndex,
-      @JsonProperty("validator_sync_committee_indices") final Set<Integer> committeeIndices) {
+      @JsonProperty("validator_sync_committee_indices") final IntSet committeeIndices) {
```
ajsutton: Basically anything with `@JsonProperty` will need to stay declared using the `Integer` form to ensure our OpenAPI docs come out right.  We *might* be able to get away with actually passing it an `IntSet` as the implementation but I'm not sure about that - we'll have to test it.

ajsutton (commented):
This has wound up being a really big PR.  It's probably worth doing this in smaller steps to make reviewing simpler and make sure we pick up on all the details like using `IntIterator` instead of a `Iterator<Integer>` and other gotchas.  I know a bunch of things wind up inter-related but if we can find more contained areas to do at each step that would be good.  For example the performance tracker classes could be converted over in one PR and `ValidatorApiHandler` in another.

data/serializer/src/main/java/tech/pegasys/teku/api/response/v1/validator/SyncCommitteeDuty.java
jtraglia: Good to know. I'll revert those instances.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/SyncCommitteePerformanceTracker.java
jtraglia: Ah I didn't think of that. I think you're right. I'll change it to one of those.

jtraglia:
Yeah, it ended up being larger than I intended too. I can try to split it up into multiple PRs. After starting to make changes, it took a long time until the whole thing even compiled. So it might not be that easy to split up.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/SszBitvectorImpl.java
```diff
@@ -74,8 +74,8 @@ public SszBitvector rightShift(int n) {
   }
 
   @Override
-  public List<Integer> getAllSetBits() {
-    return value.streamAllSetBits().boxed().collect(Collectors.toList());
+  public IntList getAllSetBits() {
+    return IntArrayList.toList(value.streamAllSetBits().boxed().mapToInt(Integer::intValue));
```
ajsutton: This is boxing and then unboxing the ints unnecessarily.  Should just be:
```suggestion
    return IntArrayList.toList(value.streamAllSetBits());
```

infrastructure/ssz/src/test/java/tech/pegasys/teku/infrastructure/ssz/collections/SszBitlistTest.java
```diff
@@ -99,7 +99,7 @@ void testSszRoundtrip(SszBitlist bitlist1) {
     Bytes ssz1 = bitlist1.sszSerialize();
     SszBitlist bitlist2 = bitlist1.getSchema().sszDeserialize(ssz1);
 
-    assertThat(bitlist2.getAllSetBits()).isEqualTo(bitlist1.getAllSetBits());
+    assertThat(bitlist2.getAllSetBits().equals(bitlist1.getAllSetBits())).isTrue();
```
ajsutton: hmmm, we don't want to lose the nice error messages we get from `isEqualTo`. Simplest answer is to just use:
```
    assertThat((List<Integer>) bitlist2.getAllSetBits()).isEqualTo(bitlist1.getAllSetBits());
```
but it's not particularly pretty.

I've pushed a commit that adds a new testFixtures source tree to `:infrastructure:collections` with a `PrimitiveCollectionsAssert` to hide the ugliness a bit.  It does mean tests will wind up boxing the ints when doing assertions but that's fine and gives us all the nice assertions and errors messages that assertj provides.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorIndexProvider.java
```diff
@@ -101,13 +102,14 @@ private void logNewValidatorIndices(final Map<BLSPublicKey, Integer> knownValida
...
     // Wait for at least one successful load of validator indices before attempting to read
     return firstSuccessfulRequest.thenApply(
         __ ->
-            ownedValidators.getActiveValidators().stream()
-                .flatMap(validator -> getValidatorIndex(validator.getPublicKey()).stream())
-                .collect(toList()));
+            IntArrayList.toList(
+                ownedValidators.getActiveValidators().stream()
+                    .flatMap(validator -> getValidatorIndex(validator.getPublicKey()).stream())
+                    .mapToInt(Integer::intValue)));
```
ajsutton, line 112: As a follow up we could change the `Map<BLSPublicKey, Integer> validatorIndexesByPublicKey` to `Object2IntMap<BLSPublicKey>` so the values can be primitive ints.  Not a high priority I don't think and definitely not worth adding to this PR.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java
```diff
@@ -187,9 +189,10 @@ private SyncCommitteeDuties responseToSyncCommitteeDuties(final PostSyncDutiesRe
                     new SyncCommitteeDuty(
                         duty.pubkey.asBLSPublicKey(),
                         duty.validatorIndex.intValue(),
-                        duty.committeeIndices.stream()
-                            .map(UInt64::intValue)
-                            .collect(Collectors.toSet())))
+                        IntOpenHashSet.toSet(
+                            duty.committeeIndices.stream()
+                                .map(UInt64::intValue)
+                                .mapToInt(Integer::intValue))))
```
ajsutton: Could just use:
```suggestion
                                .mapToInt(UInt64::intValue))))
```

ajsutton: Actually I think there are probably quite a few places we could do this - worth searching for `mapToInt(Integer::intValue)` to find them I think - I only just realised it when I got to the end of reviewing. :)

ajsutton (commented):
Looking great. Getting to the final details here.

There is a pretty long stream of work we can likely do to continue this switch which is great, but don't feel you have to keep on this particular issue. It's the kind of thing we can just keep picking off bit by bit, particularly focussing on areas that are actually having a performance impact.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/SszBitvectorImpl.java
jtraglia: Nice spot. That was a bit redundant. Will fix.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorIndexProvider.java
jtraglia: Good idea. I'll keep this in mind afterwards.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java
jtraglia: Oh yeah. That looks way better. Will update all instances.

jtraglia:
Thanks for the review. I'll make all of the suggested changes. I would have tried to split this into multiple smaller PRs, but it's been a busy day for me. I'm not sure it would be worth the trouble. But yeah, there's no rush for this. Picking at it for a while sounds good to me.

ajsutton:
Yep may as well keep this as one PR now. Undoing the changes to the REST API made it easier to review as well.

ajsutton (approved):
LGTM. Thanks for the amount of work you're putting into this.

PR #4993, opened 2022-02-16
Reload duties when validators are added.

### LINKED ISSUE #4768: [Keymanager API] need to calculate duties for validators being added or removed
When adding validators, the duties for those validators could be re-calculated to reduce the time taken for the new validator to start attesting.

When removing validators, the duties for the removed validators could be removed from the pending duties list, as we know that they don't need to be performed as soon as the deletion is successful.

rolfyone:
Also improved the `import_keys.sh` script so that a bearer token can be passed to the script.

Fixes #4768

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #4997, opened 2022-02-17
Save added remote keys to file

### LINKED ISSUE #4895: [keymanager api] allow remote keys to be added

### LINKED ISSUE #4982: [keymanager api] store added remote keys to file
As a manager of validator keys on a remote service,
I want to be able to have keys that I have added to teku,
so that they can be loaded on startup without further management.

As an adjunct to this,  if i change the URL in configuration that my signer is on, i don't want to re-add all my validators, I just want to restart and have it work.

So a request object looks like an array of:
```
{
  "pubkey": "0xa8d4c7c27795a725961317ef5953a7032ed6d83739db8b0e8a72353d1b8b4439427f7efa2c89caa03cc9f28f8cbab8ac",
  "url": "http://localhost:19000"
}
```
with optional URL.
For each of these objects passed in, in the same order as local mutables does it, we should store the file while we're importing the keys. (`LocalValidatorSource.addValidator`)

If the URL was not specified, use the value of the current configuration item. this will mean that a user can simply change that URL and restart, and the validators will come from a new url (if they weren't specified during add).

Also, potentially as a separate PR, when teku loads, any of these validators that have been stored need to be added, so that we can start up with running remote mutable validators.

The path to store the files will come from `ValidatorClientService.getManagedRemoteKeyPath`

courtneyeh:
#4895
#4982

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
+++ b/validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
@@ -20,17 +20,21 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_OK;
 
+import com.google.common.base.Charsets;
+import com.google.common.io.Files;
 import java.io.IOException;
 import java.net.URL;
 import java.net.http.HttpClient;
 import java.net.http.HttpResponse;
 import java.nio.file.Path;
 import java.util.Optional;
+import java.util.function.Supplier;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 import org.mockito.ArgumentMatchers;
+import tech.pegasys.techu.service.serviceutils.layout.SimpleDataDirLayout;
 import tech.pegasys.teku.bls.BLSPublicKey;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.StubAsyncRunner;
@@ -40,15 +44,23 @@ import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
 import tech.pegasys.teku.service.serviceutils.layout.DataDirLayout;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.TestSpecFactory;
+import tech.pegasys.teku.spec.util.DataStructureUtil;
 import tech.pegasys.teku.validator.api.ValidatorConfig;
+import tech.pegasys.teku.validator.client.ValidatorClientService;
+import tech.pegasys.teku.validator.client.restapi.apis.schema.ImportStatus;
 
 public class ExternalValidatorSourceTest {
+  private final DataStructureUtil dataStructureUtil =
+      new DataStructureUtil(TestSpecFactory.createDefault());
+
   private final Spec spec = TestSpecFactory.createMinimalAltair();
   private final PublicKeyLoader publicKeyLoader = mock(PublicKeyLoader.class);
   private final HttpClient httpClient = mock(HttpClient.class);
   private final MetricsSystem metricsSystem = new StubMetricsSystem();
   private final AsyncRunner asyncRunner = new StubAsyncRunner();
   private ThrottlingTaskQueue externalSignerTaskQueue;
+
+  private final Supplier<HttpClient> httpClientFactory = () -> httpClient;
 
   @SuppressWarnings("unchecked")
   private final HttpResponse<Void> httpResponse = mock(HttpResponse.class);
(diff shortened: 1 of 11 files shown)


validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -78,7 +82,22 @@ void setup() throws IOException, InterruptedException {
...
+            Optional.of(dataDirLayout(tempDir)));
+  }
+
+  public DataDirLayout dataDirLayout(final Path tempDir) {
+    return new DataDirLayout() {
+      @Override
+      public Path getBeaconDataDirectory() {
+        return tempDir;
+      }
+
```
rolfyone: you'll be able to use the `SimpleDataDirLayout` now :)

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -116,4 +128,83 @@ void shouldThrowExceptionWhenDeleteValidator() {
   void shouldSayFalseToAddValidators() {
     assertThat(validatorSource.canUpdateValidators()).isFalse();
   }
+
+  @Test
+  void shouldAddValidator_getsConfigUrlWhenNotProvided(@TempDir Path tempDir) throws IOException {
```
rolfyone: name more like `addValidator_shouldGetConfigUrlWhenNotProvided`
function under test, then underscore, then shouldXXXX where XXXX is desired behaviour

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -116,4 +128,83 @@ void shouldThrowExceptionWhenDeleteValidator() {
...
+    String fileName = publicKey.toBytesCompressed().toUnprefixedHexString() + ".json";
+    Path path =
+        ValidatorClientService.getManagedRemoteKeyPath(dataDirLayout(tempDir)).resolve(fileName);
+    assertThat(path.toFile()).exists();
+    assertThat(Files.toString(path.toFile(), Charsets.UTF_8))
+        .isEqualTo("{\"pubkey\":\"" + publicKey + "\"}");
+  }
+
+  @Test
+  void shouldAddValidator_getsUrlWhenProvided(@TempDir Path tempDir) throws IOException {
```
rolfyone: `addValidator_shouldGetUrlIfProvided`

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -116,4 +128,83 @@ void shouldThrowExceptionWhenDeleteValidator() {
...
+    String fileName = publicKey.toBytesCompressed().toUnprefixedHexString() + ".json";
+    Path path =
+        ValidatorClientService.getManagedRemoteKeyPath(dataDirLayout(tempDir)).resolve(fileName);
+    assertThat(path.toFile()).exists();
+    assertThat(Files.toString(path.toFile(), Charsets.UTF_8))
+        .isEqualTo("{\"pubkey\":\"" + publicKey + "\",\"url\":\"http://host.com\"}");
+  }
+
+  @Test
+  void shouldDetectDuplicatesOnAddValidator(@TempDir Path tempDir) throws IOException {
```
rolfyone: `addValidator_shouldDetectDuplicate`

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -116,4 +128,83 @@ void shouldThrowExceptionWhenDeleteValidator() {
...
+    BLSPublicKey publicKey = dataStructureUtil.randomPublicKey();
+    final AddValidatorResult result1 =
+        getResultFromAddingValidator(tempDir, publicKey, Optional.empty());
+    assertThat(result1.getResult().getImportStatus()).isEqualTo(ImportStatus.IMPORTED);
+    assertThat(result1.getSigner()).isNotEmpty();
+    assertThat(result1.getSigner().get().getSigningServiceUrl()).isNotEmpty();
+    assertThat(result1.getSigner().get().getSigningServiceUrl().get())
+        .isEqualTo(config.getValidatorExternalSignerUrl());
```
rolfyone: i think this code is heavily repeated, could potentially extract it to a check function that takes a pubkey

rolfyone: or just check status once its been verified that the values are right in another test...
or it may just mean the validations are mostly inside the getter fn you're using

rolfyone (approved):
overall looks really good, can cleanup test names, and potentially more cleanup of the verification... either way it's a good base once some/all of that is done.

PR #4998, opened 2022-02-17
Calculate the proposer again when validators added or for missed events.

### LINKED ISSUE #4992: Need to handle OnPossibleMissedEvents in BeaconProposerPreparer
also should implement onValidatorsAdded in the same way.

Currently it's a no-op which is not really sufficient...

rolfyone:
It is possible that adding validators will add proposals that we need to be aware of, so we should recalculate when validators are added, to give ourselves the best chance at not missing a block proposal. Similarly for possible missed events, its possible that proposers have been changed.

Fixes #4992

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
@@ -62,11 +62,11 @@ public class BeaconProposerPreparer implements ValidatorTimingChannel {
     if (slot.mod(spec.getSlotsPerEpoch(slot)).isZero() || !firstCallDone) {
       firstCallDone = true;
 
-      calculateProposer();
+      sendPreparableProposerList();
     }
   }
 
-  private void calculateProposer() {
+  private void sendPreparableProposerList() {
     SafeFuture<Optional<ProposerConfig>> proposerConfigFuture =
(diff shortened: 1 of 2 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
```diff
@@ -62,29 +62,32 @@ public void onSlot(UInt64 slot) {
...
-                            return buildBeaconPreparableProposerList(
-                                Optional.empty(), publicKeyToIndex);
-                          }))
-          .thenAccept(validatorApiChannel::prepareBeaconProposer)
-          .finish(VALIDATOR_LOGGER::beaconProposerPreparationFailed);
+      calculateProposer();
     }
   }
 
+  private void calculateProposer() {
```
ajsutton: nit: I think I'd call this `sendPreparableProposerList`.

ajsutton (approved):
LGTM.

validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
rolfyone: wow ok, that's not a name i'd have thought of but we'll go with that.

PR #5003, opened 2022-02-18
Add Bearer prefix to jwt token

ajsutton:
## PR Description
Add the "Bearer" prefix in the jwt authorisation header.

Switch order of args to use the non-deprecated version of `signWith`.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5005, opened 2022-02-18
Increase gossip subscription limits

### LINKED ISSUE #4903: Teku rejecting gossip subscriptions when 3 forks scheduled for subsequent epochs
### Description
When a custom network is configured with phase0 at genesis, Altair at epoch 1 and bellatrix at epoch 2 Teku activates all three gossip networks at startup because they’re activating within 2 epochs. Gossip topics are fork specific so teku is subscribing to 3 times as many topics as it would normally follow all at once at startup. Since it’s told to subscribe to all subnets that’s 64 individual attestation subnets plus the blocks, aggregates etc subnets * 3 forks. In total we’re subscribing to 217 topics.

To avoid DOS attacks, Teku limits the number of topics you can ask to subscribe to in a single message and the total number of topics a peer can be subscribed to at once. We set them to "absurdly high" values that were more than twice what you’d need to be subscribed to... 100 topics per subscription request and a total of 200 topics that can be subscribed to at once. Because the limits were absurd we didn’t actually get teku to split up requests so we wind up ignoring the only subscription message that’s sent on both sides and hence two teku instances in this configuration wind up not subscribed to any gossip at all.

The subscriptions also violate the maxSubscriptions and maxGraft settings for libp2p gossip (both set to 200).  libp2p probably should be splitting the messages it sends to stay under those limits automatically but isn’t.

Likely fix:
 1. Update libp2p to automatically split messages to ensure they remain under the maxSubscriptions and maxGraft sizes
 2. Increase the limit on total number of subscriptions Teku supports to 250 (just enough for 3 forks simultaneously)
 3. Increase the 100 topics per subscription request limit to 200 to match the libp2p spec (or possibly remove the restriction entirely since it's enforced at the libp2p level)

ajsutton:
## PR Description
Ensures the limits are high enough to not be exceeded even when three forks are subscribed simultaneously as is the case with custom testnets that schedule two forks in subsequent epochs.

## Fixed Issue(s)
fixes #4903 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,3 +19,4 @@ For information on changes in released versions of Teku, see the [releases page]
 ### Bug Fixes
  - Improved handling of the http accept header used to determine whether to send SSZ data or json for states and blocks.
  - Updated discovery library with improved performance.
+ - Fixed issue on custom testnets with two forks in sequential epochs where the maximum gossip topic subscription limits were exceeded.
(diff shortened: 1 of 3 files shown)


rolfyone (approved):
LGTM

PR #5009, opened 2022-02-18
sync-aware `TerminalPowBlockMonitor`

tbenr:
Let `TerminalPowBlockMonitor` do its checks when the node is full in Sync (CL and EL in sync)

The basic idea is that, even if the Terminal Block detection depends only by EL sync status only, if CL is syncing the knowledge of Terminal Block is useless.

Lowered the gravity and rephrased the TTD validation log from `error` to `warning` in the case when the EL has progressed beyond the Terminal Block, which means that it's likely that we were syncing around that time and we missed it.

Also removed `AtomicBoolean` usage, not needed since all important methods are `synchronized`

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/TerminalPowBlockMonitor.java
```diff
@@ -234,8 +243,9 @@ private void checkTerminalBlockByTTD() {
                                 "Total Difficulty of Terminal Block parent has been validated.");
                             onTerminalPowBlockFound(powBlock.getBlockHash());
                           } else {
-                            LOG.error(
-                                "Total Difficulty of Terminal Block parent is invalid! Must be lower than TTD!");
+                            LOG.warn(
+                                "A candidate Terminal Block has been found but its parent has a Total Difficulty of higher than TTD. "
```
ajsutton:
```suggestion
                                "A candidate Terminal Block has been found but its parent has a Total Difficulty greater than terminal total difficulty. "
```

ajsutton (approved):
LGTM.

PR #5011, opened 2022-02-19
Convert all "indexes" -> "indices"

jtraglia:
This is super picky, but I noticed mixed usage between "indexes" and "indices" in the codebase. It seems that "indices" is more common, so I chose to go with that one. I think it is good to have a unified terminology.

Also: https://english.stackexchange.com/a/3126

## Documentation

## Changelog


ajsutton (approved):
LGTM. Thanks - always nice to get the little details right.

PR #5012, opened 2022-02-20
Convert `Map<?, Integer>` to `Object2IntMap<?>`

jtraglia:
Adrian mentioned [0] that we could convert these maps to the fastutil version. I'm a little unsure if `SynchronizedMap` is a good replacement for `ConcurrentHashMap` but I think it is. Also, not sure if I need to change the type of `validatorIndexesByPublicKey` in `BeaconProposerPreparerTest.java` too.

There are about 20 more instances of these maps. I plan to convert those too.

[0] https://github.com/ConsenSys/teku/pull/4991#discussion_r808554792

## Documentation

## Changelog


ajsutton:
`SynchronizedMap` won't be a straight replacement for `ConcurrentHashMap` from my understanding as it will have quite different performance characteristics. We'd need to be very careful with those conversions as there's a good chance the extra contention from `SynchronizedMap` would outweigh the benefits of avoiding boxed Integer.

jtraglia:
That's what I was afraid of. I read that it was considerably slower. Unfortunately, fastutil does not have concurrent objects. Is it worth converting the other ~20 instances, that are just regular `HashMap` types.

Issue about fastutil `ConcurrentMap`: https://github.com/vigna/fastutil/issues/81

ajsutton:
Yep I'd say it's worth converting all the plain `HashMap` types that use `Integer` but just leave the `ConcurrentHashMap` ones as-is.

jtraglia:
Sounds good. I will finish this up tomorrow.

jtraglia:
I would update the `getValidatorIndices` in classes that implement `ValidatorApiChannel` but it isn't very straight-forward in one of them. The function used to get indices, `makeBatchedValidatorRequest`, uses a generic type.

https://github.com/ConsenSys/teku/blob/2113805170e0b693edc21b4636137145cbce5de2/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java#L107-L141

I would prefer not to change any of that code.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/genesis/GenesisGenerator.java
```diff
@@ -102,11 +102,11 @@ public void updateCandidateState(
...
 
   private void processActivation(final Deposit deposit) {
-    final Integer index = keyCache.get(deposit.getData().getPubkey());
-    if (index == null) {
-      // Could be null if the deposit was invalid
+    if (!keyCache.containsKey(deposit.getData().getPubkey())) {
+      // Could be absent if the deposit was invalid
       return;
     }
+    final int index = keyCache.get(deposit.getData().getPubkey());
```
ajsutton: Given we can wind up running this in a tight loop it may be worth avoiding the additional map lookup by doing something like:

```suggestion
    final int index = keyCache.getOrDefault(deposit.getData().getPubkey(), -1);
    if (index == -1) {
      // Could be absent if the deposit was invalid
      return;
    }
```

We can't have negative deposit indices so -1 works fine as a sentinel value. Not a big performance difference but an easy win for no extra code complexity.

ajsutton (approved):
LGTM. Thanks again.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/genesis/GenesisGenerator.java
ajsutton: Going to try committing this and merging - I have a terrible track record with breaking spotless when writing code in GitHub comments though so don't panic if you get failure notifications, I'll sort it... :)

ajsutton:
> I would update the getValidatorIndices in classes that implement ValidatorApiChannel but it isn't very straight-forward in one of them. The function used to get indices, makeBatchedValidatorRequest, uses a generic type.

Yeah I think it's worth waiting to see if there's really any benefit.  That batching code is definitely nice to be able to share as it's fairly complex to get right.


Not sure if you're getting sick of swapping Integer to int yet or not - we can certainly find something else unrelated if you are, but the other thing I thought might be useful to identify places we're still boxing ints unnecessarily and prevent us regressing in the future is to write a custom error prone rule that would check we don't use List<Integer> and use intIterators instead of normal ones etc.  We already have some custom rules shared with other projects at https://github.com/ConsenSys/errorprone-checks and could probably add to those, though it may be easier to have the custom rule in the teku codebase - we used to build the custom rules that way prior to splitting them out in 9ba0b2c0da39d893a154bcc6ea1c567b6e4be5d6 so that would be a pattern for re-adding custom rules directly to Teku.  The advantage of having them in Teku is it makes it easy to pull in the latest changes to the rules and we can gradually make them stricter as we update the codebase to follow them.

jtraglia:
Learning more about errorprone sounds cool. I will see what I can do there.

Somewhat related, I have been using [Semgrep](https://semgrep.dev), a static analysis tool, to identify some patterns in Teku. It seems that errorprone has really great Java integration and might work better with Teku. I will attempt to re-write some of those patterns into errorprone. I will share anything that is useful and does not have false positives.

PR #5014, opened 2022-02-20
Remove link checker action

ajsutton:
## PR Description
We've had to ignore most of the links due to DoS prevention on websites and still get intermittent failures.

At this point the cost of training people to ignore failures is higher than the value for maybe finding a broken link in the small set of links we're still checking.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5015, opened 2022-02-21
Don't log warning when block slashing protection is old

ajsutton:
## PR Description
Blocks aren't expected to be produced very often so don't consider the slashing protection records out of date based on the last block slot. If the attestation data is up to date then the slashing protection record is up to date.

Also simplified the tests a bit.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5018, opened 2022-02-21
Load added remote keys from file

### LINKED ISSUE #4895: [keymanager api] allow remote keys to be added

### LINKED ISSUE #4982: [keymanager api] store added remote keys to file

courtneyeh:
fixes #4895
fixes #4982

## Documentation

## Changelog


rolfyone:
Looks like good progress, would like to see a test loading more than 1 (2 would be plenty) external mutable validators via validatorLoader - similar to `shouldLoadMutableValidatorIfEnabled` loading from an external mutable validator... and making sure they both get added by the time loadValidators completes...

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -151,6 +155,56 @@ void addValidator_shouldDetectDuplicate(@TempDir Path tempDir) throws IOExceptio
...
+    createRemoteKeyFile(dataDirLayout, publicKey1, Optional.of(new URL("http://host.com")));
+
+    final BLSPublicKey publicKey2 = dataStructureUtil.randomPublicKey();
+    createRemoteKeyFile(dataDirLayout, publicKey2, Optional.empty());
+
+    List<ValidatorSource.ValidatorProvider> validators =
+        externalValidatorSource.getAvailableValidators();
+
+    checkExternalValidatorProvider(validators.get(0), publicKey2, new URL("http://localhost:9000"));
+    checkExternalValidatorProvider(validators.get(1), publicKey1, new URL("http://host.com"));
```
rolfyone: could re-use that as a variable...

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -151,6 +155,56 @@ void addValidator_shouldDetectDuplicate(@TempDir Path tempDir) throws IOExceptio
...
+            externalSignerTaskQueue,
+            Optional.of(dataDirLayout));
+
+    final BLSPublicKey publicKey1 = dataStructureUtil.randomPublicKey();
+    createRemoteKeyFile(dataDirLayout, publicKey1, Optional.of(new URL("http://host.com")));
+
+    final BLSPublicKey publicKey2 = dataStructureUtil.randomPublicKey();
+    createRemoteKeyFile(dataDirLayout, publicKey2, Optional.empty());
+
+    List<ValidatorSource.ValidatorProvider> validators =
```
rolfyone: nit: final :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSource.java
```diff
@@ -145,13 +208,20 @@ public AddValidatorResult addValidator(
                   ValidatorTypes.EXTERNAL_VALIDATOR_STORE)
               .getBytes(UTF_8));
 
+      URL url = signerUrl.orElse(config.getValidatorExternalSignerUrl());
```
rolfyone: final :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSource.java
```diff
@@ -99,13 +105,70 @@ public static ExternalValidatorSource create(
 
   @Override
   public List<ValidatorProvider> getAvailableValidators() {
+    if (readOnly) {
+      return getAvailableReadOnlyValidators();
+    }
+
+    // Load from files
+    List<File> files = getValidatorFiles();
```
rolfyone: final

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSource.java
```diff
@@ -99,13 +105,70 @@ public static ExternalValidatorSource create(
...
+    final Path directory = ValidatorClientService.getManagedRemoteKeyPath(dataDirLayout);
+
+    List<File> files = new ArrayList<>();
+    try (DirectoryStream<Path> stream = Files.newDirectoryStream(directory, "*.json")) {
+      stream.forEach(entry -> files.add(entry.toFile()));
+    } catch (IOException e) {
+      LOG.debug("Failed to read file ", e);
+    }
+
+    return files;
```
rolfyone: we can do something similar to...
```
    final File[] files = directory.toFile().listFiles((dir, name) -> name.toLowerCase().endsWith("json"));
    return files == null ? List.of() : Arrays.asList(files);
```

rolfyone (approved):
LGTM

PR #5019, opened 2022-02-21
Remove use dependent roots feature toggle.

ajsutton:
## PR Description
Remove feature toggle for dependent roots. We've been using dependent roots for a long time so no need to maintain the old approach.

There's a couple of differently named things this toggle controlled (which makes sense in their specific contexts):

* `useDependentRoots` - now always `true` - duties are invalidated based on the `dependent_root` metadata
* `updateHeadForEmptySlots` - now always `false` (we don't send a re-org event if an empty slot is filled)
* `useIndependentAttestationTiming` - now always `true` - the validator now handles timing independently and doesn't expect a `head` event every slot

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5028, opened 2022-02-21
Change RecentChainData.getBestState to return a future

ajsutton:
## PR Description
Change `RecentChainData.getBestState` to return a future. This is the start of a process to allow the chain head to be updated synchronously instead of potentially having to wait for the head state to be available.

## Documentation

## Changelog


data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -172,81 +167,6 @@ public void getBeaconState_shouldFindHeadState() throws ExecutionException, Inte
         .isEqualTo(beaconStateInternal.hashTreeRoot());
   }
 
-  @Test
```
tbenr, line 175: Why removing all these tests?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/TerminalPowBlockMonitor.java
```diff
@@ -125,9 +125,9 @@ private synchronized void monitor() {
     }
 
     // beaconState must be available at this stage
```
tbenr: this comment can be removed

tbenr (approved):
LGTM with a question and a nit

data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
ajsutton: Because I also removed the `validatorParameterToIndex` method which was unused.

PR #5029, opened 2022-02-21
Add mev boost url CLI argument.

### LINKED ISSUE #5021: [MEV milestone 1] Allow a CLI to specify execution-payload builders
Initially as a development option, I want to be able to specify at least 1, potentially multiple builders of execution payload for use in MEV.

At the end of this ticket, we should have something like
--Xexecution-payload-builders=[...] that accepts 1+ URLs that point to MEV API compatible endpoints, and in the first instance is hidden.

This will be a flag available on the beacon node, and will not be required in the VC.  

Once the feature is ready, this will become a documented flag on the beacon-node.

rolfyone:
Supports setting multiple URLS for use with the MEV boost api.

fixes #5021

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5030, opened 2022-02-22
Add CLI argument for using blinded beacon blocks for the validator-client

### LINKED ISSUE #5022: [MEV milestone 1] add a CLI flag for mev-enabled on the VC
We need a development flag to enable MEV on the validator client.

This is a feature toggle, so it will be a development (hidden) option, and will enable the use of mev-boost functionality if enabled.
The flag can be something like `--Xmev-boost-enabled=true`, defaulting to false.

rolfyone:
This will be needed for MEV boost, but is also likely to be used post merge to reduce block data sent to the BN.

fixes #5022

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
@@ -13,6 +13,8 @@
 
 package tech.pegasys.teku.cli.options;
 
+import static tech.pegasys.teku.validator.api.ValidatorConfig.DEFAULT_VALIDATOR_BLINDED_BLOCKS_ENABLED;
+
 import java.nio.file.Path;
 import java.util.Optional;
 import org.apache.tuweni.bytes.Bytes32;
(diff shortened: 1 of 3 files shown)


teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -99,6 +99,16 @@
...
 
+  @Option(
+      names = {"--Xvalidators-blinded-blocks-api-enabled"},
+      paramLabel = "<BOOLEAN>",
+      showDefaultValue = Visibility.ALWAYS,
+      description = "Use blinded block api calls",
+      fallbackValue = "true",
+      hidden = true,
+      arity = "0..1")
+  private boolean blindedBlocksApiEnabled = false;
```
ajsutton: We should use a constant for the default value to ensure it stays in sync between here and the config builder.

ajsutton (approved):
LGTM, just worth removing the duplicate default value.

PR #5031, opened 2022-02-22
Find common ancestor using protoarray rather than chain head state

ajsutton:
## PR Description
To avoid needing the chain head state to be loaded prior to setting the new head, find common ancestors using the information in protoarray rather than the historical block info in the head state.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5032, opened 2022-02-22
Remove fork choice slot from ChainHead

ajsutton:
## PR Description
Remove the `forkChoiceSlot` field from `ChainHead` as it is now unnecessary.  This makes `ChainHead` pretty pointless but it's about to be modified to be useful again.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5034, opened 2022-02-22
Change attestation publication errors log level

tbenr:
## PR Description

Changes logging to DEBUG when we have no peers for the topic.
In all other cases (when no succeeded submission to at least one peer) we log a WARN.

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/AttestationGossipManager.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/AttestationGossipManager.java
@@ -13,6 +13,8 @@
 
 package tech.pegasys.teku.networking.eth2.gossip;
 
+import io.libp2p.pubsub.NoPeersForOutboundMessageException;
+import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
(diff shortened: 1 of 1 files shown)


rolfyone (approved):
LGTM

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/AttestationGossipManager.java
```diff
@@ -60,7 +62,8 @@ public void onNewAttestation(final ValidateableAttestation validateableAttestati
               attestationPublishSuccessCounter.inc();
             },
             error -> {
-              LOG.trace(
+              LOG.log(
+                  error instanceof NoPeersForOutboundMessageException ? Level.DEBUG : Level.WARN,
```
ajsutton, line 66: I think this is going to spam the logs when we run a single node network as it never has any peers. Probably not the end of the world but it does get annoying.

ajsutton: Ah right sorry, in the case of no peers at all it would be debug which is fine.  I'd suggest other exceptions should be `ERROR` level then rather than `WARN` as I wouldn't expect those (unless there's some exception type I'm missing).

tbenr: I expect networking exceptions. changing to ERROR.

PR #5035, opened 2022-02-23
Convert PeerStatusIntegrationTest to using StorageSystem

ajsutton:
## PR Description
Convert PeerStatusIntegrationTest to use StorageSystem instead of BeaconChainUtil and add a test for the status after finalisation has progressed past genesis.

Adds a method to "properly" finalize the current chain head in ChainUpdater by adding real attestations to the chain.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5038, opened 2022-02-23
Remove usage of chain head state from StatusMessageFactory

ajsutton:
## PR Description
Update `StatusMessageFactory` to retrieve the finalised checkpoint from the store instead of from the chain head state. This allows it to work even if the chain head state isn't (yet) available.  A special case has to be added for when the finalised checkpoint is the genesis block where the finalised root must be zero because the genesis state can't know the genesis block root when it's being created.

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
@@ -14,7 +14,8 @@
 package tech.pegasys.teku.networking.eth2.rpc.beaconchain.methods;
 
 import java.util.Optional;
-import tech.pegasys.teku.spec.datastructures.blocks.StateAndBlockSummary;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlockSummary;
 import tech.pegasys.teku.spec.datastructures.networking.libp2p.rpc.StatusMessage;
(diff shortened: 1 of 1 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
```diff
@@ -34,14 +35,17 @@ public StatusMessageFactory(final RecentChainData recentChainData) {
...
-    final Checkpoint finalizedCheckpoint = chainHead.getState().getFinalized_checkpoint();
 
     return Optional.of(
         new StatusMessage(
             forkInfo.getForkDigest(recentChainData.getSpec()),
-            finalizedCheckpoint.getRoot(),
+            // Genesis finalized root is always ZERO because it's taken from the state and the
+            // genesis block is calculated from the state so the state can't contain the actual
+            // block root
+            finalizedCheckpoint.getEpoch().isZero() ? Bytes32.ZERO : finalizedCheckpoint.getRoot(),
```
rolfyone, line 48: test?

ajsutton: No behaviour change - the existing `PeerStatusIntegrationTest` covers this nicely.

rolfyone: ok

rolfyone (approved):
LGTM

PR #5040, opened 2022-02-23
Use more generic types

ajsutton:
## PR Description
Use more generic types in method references when referring to the chain head. Also fixes an odd quirk in `RecentChainDataTest` where it wasn't importing all blocks.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5041, opened 2022-02-23
zero uint256 should not output 0x

### LINKED ISSUE #5036: UInt256.ZERO should seralize as 0x0 in JSON-RPC
### Description
In JSON-RPC the zero value for numbers is serialised as 0x0, not 0x (as it would be in RLP).  Currently `UInt256AsHexSerializer` serialises 0 as 0x which is rejected by Geth.

rolfyone:
Used an alternative function which is similar to `toShortHexString` but outputs 0x0 if the entity is zero.

fixes #5036

## Documentation

## Changelog


ajsutton (commented):
LGTM but would like to get a second opinion from @tbenr before we merge as I don't really understand why we have both `UInt256Serializer` and `UInt256AsHexSerializer` that are doing the same thing and I'm hoping he remembers more of the story of how we got here...

rolfyone:
agree it seems odd but wasn't confident enough to clean it up. i think ideally we'd only have 1 implementation (especially because they're the same at the moment basically)

tbenr:
I think the origin is the merge branch, and I think the basic Idea was to have a clearly separated serializer\deserializer for the ee client, easy to change without any impact on rest API (they are not blindly added in when creating`JsonProvider`).
 
Seems like we shouldn't have any dependency hell by adding `implementation project(':data:serializer')`
~~We can consolidate even if in some cases we still need to override (ie Hex Deserializer to enforce hex only values as prescribed by jsonrpc).~~
Now that things are settled we could reference the data:serializer classes when work well and leave specific classes in the ee, but we are somehow linking rest api with ee.

ajsutton (approved):
Thanks Enrico. That makes sense to me and I quite like the idea of keeping the ee serialisation separate since JSON-RPC is a different format to the REST API. Plus we're locked into using Jackson annotations here to support web3j but want to move away from that on the REST API to using the newer type definitions.

So LGTM.

tbenr:
Wait a sec. Is it the correct serializer? Shouldn't be the one on ee?

rolfyone:
the second asHex is the one in ee

ajsutton:
Goode point, we should double check how this should be serialised in the REST API and make sure we don't break that. Not sure if that's as a number or as data...

tbenr:
> the second asHex is the one in ee

Oh with mobile app I haven't seen the second:)

PR #5042, opened 2022-02-23
Reduce CircleCI storage requirements

ajsutton:
## PR Description
Only capture .tar.gz distribution as a CircleCI artefact. Reduces the amount of storage space we need to use on CircleCI.

Also ensures the gradle dependencies cache starts from scratch when dependencies change to ensure we aren't keeping old versions of dependencies in the cache unnecessarily.

## Documentation

## Changelog


joshuafernandes (approved):
LGTM

PR #5043, opened 2022-02-23
Calculate dependent roots for head updates from protoarray

ajsutton:
## PR Description
Calculate dependent roots for head updates from protoarray instead of the head state. Removes a dependency on the head state being available so that it can be loaded async in the future.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5044, opened 2022-02-23
SyncState new subscribeAndUpdate method

tbenr:
## PR Description

When we subscribe to sync state we always immediately call the subscriber to set the current value.
To cleanup initialization in `BeaconChainController` I made this behaviour structural in the subscription method.

There is only one exception in `HistoricalBlockSyncService` initialization, that's why I left the original method.
In this case I'm not sure if we want to issue an immediate `fetchBlocks()`. If yes we can remove the old method.

## Documentation

## Changelog


ajsutton:
For `HistoricalBlockSyncService` we are doing an initial `fetchBlocks()` already, it's just chain after initialise:

```

  @Override
  protected SafeFuture<?> doStart() {
    LOG.debug("Start {}", getClass().getSimpleName());
    return initialize().thenRun(this::fetchBlocks);
  }
```

so could remove the `thenRun` from that if we know that the sync state subscription will always call us back.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/AttestationGossipManager.java
```diff
@@ -63,7 +63,7 @@ public void onNewAttestation(final ValidateableAttestation validateableAttestati
             },
             error -> {
               LOG.log(
-                  error instanceof NoPeersForOutboundMessageException ? Level.DEBUG : Level.WARN,
+                  error instanceof NoPeersForOutboundMessageException ? Level.DEBUG : Level.ERROR,
```
ajsutton, line 66: This change is from the other PR and probably shouldn't be in here. :)

ajsutton (approved):
LGTM.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/AttestationGossipManager.java
tbenr: 🤦‍♂️

tbenr: > For `HistoricalBlockSyncService` we are doing an initial `fetchBlocks()` already, it's just chain after initialise:
> 
> ```
> 
>   @Override
>   protected SafeFuture<?> doStart() {
>     LOG.debug("Start {}", getClass().getSimpleName());
>     return initialize().thenRun(this::fetchBlocks);
>   }
> ```
> 
> so could remove the `thenRun` from that if we know that the sync state subscription will always call us back.

It's a bit more complicated than that. If I do the change there is a side effect when we start the service and we are already in sync that cause us to not unsubscribe since we call stop() before actually set the subscriptionId in `syncStateSubscription` (which remains -1).

tbenr: > This change is from the other PR and probably shouldn't be in here. :)

I leave it here since ERROR is what it should be.

ajsutton: > It's a bit more complicated than that. If I do the change there is a side effect when we start the service and we are already in sync that cause us to not unsubscribe since we call stop() before actually set the subscriptionId in syncStateSubscription (which remains -1).

Ah subtle - good spot.

PR #5047, opened 2022-02-23
Check UDP port availability when discovery enabled only

Nashatyrev:
## PR Description

When discovery is disable no need to check the UDP port availability (it could then be used for other purpose) 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5048, opened 2022-02-23
[Java API] Use generic privateKeySource instead of privateKeyFile in NetworkConfig

Nashatyrev:
## PR Description

When instantiating Teku node programmatically it makes sense to configure private key directly with key bytes instead of creating temporary key file. 

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5049, opened 2022-02-23
warnMissingProposerDefaultFeeRecipient logged only when coming from remote

tbenr:
## PR Description

We are logging a warning on each UpdatePreparableProposers call when no default fee recipient was specified on beacon node side. This is actually relevant the call comes from a remote validator. In case we are running an in process validator the default recipient fee config is already managed by the validator cli.

So I made the api "source aware". The solution isn't that elegant. Better one are welcome :-)

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -83,7 +83,7 @@ public class StatusLogger {
   public void warnMissingProposerDefaultFeeRecipientWithPreparedBeaconProposerBeingCalled() {
     log.warn(
         print(
-            "Validator Client detected and no default proposer fee recipient configured! "
+            "Remote Validator Client detected and no default proposer fee recipient configured! "
                 + "it is strongly recommended to configure it to avoid possible block production failures",
             Color.RED));
   }
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -84,7 +84,7 @@ public class StatusLogger {
     log.warn(
         print(
             "Remote Validator Client detected and no default proposer fee recipient configured! "
-                + "it is strongly recommended to configure it to avoid possible block production failures",
+                + "It is strongly recommended to configure fee recipient to avoid possible block production failures.",
             Color.RED));
   }
 
--- a/validator/client/src/test/java/tech/pegasys/teku/validator/client/BeaconProposerPreparerTest.java
+++ b/validator/client/src/test/java/tech/pegasys/teku/validator/client/BeaconProposerPreparerTest.java
@@ -15,7 +15,6 @@ package tech.pegasys.teku.validator.client;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.Mockito.doThrow;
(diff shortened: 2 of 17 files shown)


validator/client/src/test/java/tech/pegasys/teku/validator/client/BeaconProposerPreparerTest.java
```diff
@@ -158,20 +159,23 @@ void should_notCallPrepareBeaconProposerAfterFirstSlotOfEpoch() {
   @TestTemplate
   void should_callPrepareBeaconProposerAfterOnPossibleMissedEvents() {
     beaconProposerPreparer.onPossibleMissedEvents();
-    verify(validatorApiChannel).prepareBeaconProposer(any());
+    verify(validatorApiChannel).prepareBeaconProposer(any(), anyBoolean());
   }
 
+  @TestTemplate
```
tbenr, line 164: @rolfyone this was missing, I thought it was intended to be there.

ajsutton (approved):
LGTM, but that is definitely ugly...

Two options I can think of here:

1. Can we move the warning about missing configuration to the REST API handler for get block instead of in ForkChoiceNotifier?
2. Is it possible for the beacon node to use the default fee recipient it would have used if it was an internal validator?

validator/client/src/test/java/tech/pegasys/teku/validator/client/BeaconProposerPreparerTest.java
rolfyone: probably means the test wasn't running :/ yes it looks like it should follow the pattern here.

tbenr:
> LGTM, but that is definitely ugly...
> 
> Two options I can think of here:
> 
> 1. Can we move the warning about missing configuration to the REST API handler for get block instead of in ForkChoiceNotifier?
> 2. Is it possible for the beacon node to use the default fee recipient it would have used if it was an internal validator?

well, it is already like that, and it is `--Xvalidators-proposer-default-fee-recipient` coming from `ValidatorConfig`. Let me recap how it works so I can solidify in my head too:
- from VC perspective it's easy: we always have a default that comes from the param or from the config file.
- from BN perspective, the param is not mandatory but its expected if BN is used by some remote validator. We early detect it on startup and warn if param is missing and REST api are enabled (it isn't bad situation by itself). We try to detect an actual bad situation if a VC is actively using the apis. If comes from internal VC we don't want to warn because it is already covered. if comes from remote we try to warn as soon we become aware of a VC. "we know there is a remote VC but you haven't specified a default recipient, so if for some reason a VC don't declare a recipient we may end up with nothing to set.

for option 1. We still need the "ugliness" and we are not as early as in proposerPreparation API.

another option could be to make `ValidatorDataProvider` aware of the config and print the warning from there, always in `prepareBeaconProposer`.

tbenr:
turns out there was a much better solution

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -83,7 +83,7 @@ public void warnMissingProposerDefaultFeeRecipientWithRestAPIEnabled() {
   public void warnMissingProposerDefaultFeeRecipientWithPreparedBeaconProposerBeingCalled() {
     log.warn(
         print(
-            "Validator Client detected and no default proposer fee recipient configured! "
+            "Remote Validator Client detected and no default proposer fee recipient configured! "
                 + "it is strongly recommended to configure it to avoid possible block production failures",
```
rolfyone: nit: `It` is the start of a sentence, I'd probably remove the second 'it'... something like
`Configure fee recipient to avoid possible block production failures.`

ajsutton (approved):
LGTM. Definitely seems nicer.  Just don't forget to make that `it` a capital `It`. :)

rolfyone (approved):
LGTM overall, one nit

PR #5053, opened 2022-02-24
Make head state access use a SafeFuture

ajsutton:
## PR Description
Switches `ChainHead` to expose the head state via a future. Currently the state is actually always a completed future but this gets the places that access it ready for it to be async.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5055, opened 2022-02-24
Fix BlockHeader to use the body root not block root

ajsutton:
## PR Description
Fix `BlockHeader` to assign the body root to `body_root` instead of the block root.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5056, opened 2022-02-24
Feature: Websockets and IPC connection to execution engine

### LINKED ISSUE #5013: Support websockets to talk to Execution clients
### Description
Most execution clients provide RPC APIs via JSON-RPC over both HTTP and Websocket protocols. It would be nice that users who are providing their APIs over only Websockets could use Teku as Consensus client. As it stands today, they must run HTTP style apis in order to integrate with Teku.

**Expected behavior:**
When configured with an eth1 url that starts with `ws://` a Websocket connection should be negotiated.

Teku fails to upgrade the connection from HTTP to Websocket, and fails to use the execution client.

* Software version: teku/v22.1.1+84-gc30c00c/osx-x86_64/adoptopenjdk-java-11
* Java version: adoptopenjdk-java-11
* OS Name & Version: OSX

zilm13:
## PR Description
Added support of Websockets connection to execution engine. There are several notes regarding this PR:
- `Files.readString(filePath).strip()` in `JwtSecretKeyLoader`. Similar code appears in several parts of the project, I'd add it to some `Utils` to be consistent over the codebase, but haven't found any `Utils`.
- Errors are handled too broad both with websockets and http as execution client being offline. Not semantically correct, but user see errors like this not too deep. Pretty sure `ConnectException` is not the only type for execution client connection issues, so not forcing fix.
```
13:12:49.893 ERROR - Execution Client is offline java.lang.Error: -32601: the method engine_exchangeTransitionConfigurationV1 does not exist/is not available (See log file for full stack trace)
```
- We may need to support TLS in the future for both type of connections. Besu already supports it. Shouldn't be an epic issue for `https`, but for `wss` it could. Check more below regarding implementation limitations.

There are several constraints due to limited support of websockets (comparing to http) by `web3j` library:
- we couldn't have something like log interceptor for http, though in `TRACE` logging `web3j` logs incoming and outgoing messages by itself
- `WebSocketService` which we should use with `web3j` ignores connection exception, so it may be difficult to debug for the user. For example, with wrong JWT tokem on http you see in console "403 wrong token" or something like this, but inside `WebSocketService` there is dummy thrower
```
   if (!connected) {
            throw new ConnectException("Failed to connect to WebSocket");
        }
```
- Websockets implementation in `web3j` doesn't support TLS which could be an issue in future
- I don't see simple way to plug something from 3rd party into `web3j` for Websockets

## Fixed Issue(s)

Fixes #5013
## Documentation

## Changelog


zilm13:
Http TRACE logs are currently posted by `Web3jClientBuilder`, don't like it

UPDATED: moved initialization of interceptors to `Web3jHttpClient`, this way it should look better

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3JClient.java
```diff
@@ -0,0 +1,109 @@
...
+            .sendAsync()
+            .handle(
+                (response, exception) -> {
+                  if (exception != null) {
+                    handleError(exception);
+                    return new Response<>(exception.getMessage());
+                  } else if (response.hasError()) {
+                    final String errorMessage =
+                        response.getError().getCode() + ": " + response.getError().getMessage();
+                    handleError(new Error(errorMessage));
```
ajsutton: `Error` is probably the wrong thing to use here as it generally indicates a serious JVM error.  Something like `IOException` is probably better.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jClientBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+  public Web3jClientBuilder timeProvider(TimeProvider timeProvider) {
+    this.timeProvider = timeProvider;
+    return this;
+  }
+
+  public Web3JClient build() {
+    checkNotNull(timeProvider);
+    checkNotNull(endpoint);
+    switch (endpoint.getScheme()) {
+      case "http":
```
ajsutton, line 60: We can also support `https` using the http service.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jClientBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+        final OkHttpClient okHttpClient =
+            Web3jHttpClient.createOkHttpClient(jwtConfigOpt, timeProvider);
+        Web3jService httpService = new HttpService(endpoint.toString(), okHttpClient);
+        return new Web3jHttpClient(timeProvider, httpService);
+      case "ws":
```
ajsutton, line 63: `WebSocketClient` does support `wss` urls automatically so can just include `wss` in this case. It ultimately uses https://github.com/TooTallNate/Java-WebSocket for web sockets which supports ssl.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jWebsocketClient.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private void setupJwtAuth(final Optional<JwtConfig> jwtConfig, final TimeProvider timeProvider) {
+    if (jwtConfig.isPresent()) {
+      JwtAuthWebsocketHelper jwtAuthWebsocketHelper =
+          new JwtAuthWebsocketHelper(jwtConfig.get(), timeProvider);
+      this.jwtAuth = Optional.of(jwtAuthWebsocketHelper);
+    }
+  }
+
+  private void tryToConnect() {
```
ajsutton: Currently if this connection fails, we'll still try to send the request which gives a very misleading error message.  I suspect we should handle the `ConnectException` in the actual `do*Request` methods so they can return a failed future when the connection fails with the `ConnectException` as the cause.  Currently they're reporting `WebsocketNotConnectedException` which makes it look like we didn't try to connect.

ajsutton (approved):
LGTM generally just a couple of details to sort out.

> Files.readString(filePath).strip() in JwtSecretKeyLoader. Similar code appears in several parts of the project, I'd add it to some Utils to be consistent over the codebase, but haven't found any Utils.

We deliberately avoid generic `Utils` class and modules as it just becomes a huge collection of random junk.  We do have the `:infrastructure:io` module that a helper method could be added into though.  May be worth checking if something exists in commons-io or guava first though rather than writing our own.

> Errors are handled too broad both with websockets and http as execution client being offline. Not semantically correct, but user see errors like this not too deep. Pretty sure ConnectException is not the only type for execution client connection issues, so not forcing fix.
> 13:12:49.893 ERROR - Execution Client is offline java.lang.Error: -32601: the method engine_exchangeTransitionConfigurationV1 does not exist/is not available (See log file for full stack trace)

The `engine_exchangeTransitionConfigurationV1` method is called just to check the engine is available and configured correctly so if it fails we actually do want to report that the engine is offline.  We could potentially improve this but generally if we're getting failures from the EL we do want to report a pretty clear message to the user that the EL is unavailable.  The full exception is available in the logs as well to diagnose what caused the problem.

> We may need to support TLS in the future for both type of connections. Besu already supports it. Shouldn't be an epic issue for https, but for wss it could. Check more below regarding implementation limitations.

https and wss are both fully supported by web3j - we just need to pass those URLs through.

> WebSocketService which we should use with web3j ignores connection exception, so it may be difficult to debug for the user. 

Might be worth putting up a PR for web3j to fix this. I think we'd get more information if it just registered the web socket listener before attempting to connect in the `WebSocketService.connect` as then our error handler would get a callback with the actual exception that caused the connect failure (we'd also get the generic exception after that but that's ok).

ajsutton:
btw as a follow up we should probably add support for using ws/wss for the eth1-endpoint as well (unless it adds some extra complications that make it more difficult - it's not a high priority).  And also if it's easy adding support for IPC via `IpcService` would be useful, but again not a high priority if it turns out to be difficult.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jClientBuilder.java
zilm13: Maybe I'm wrong in my understanding but when you use https in browser the connection is accepted because certificates are signed by limited list of trusted companies and browser trust this signature. In case with your own TLS secured gateway you need to provide certificates keys etc. In Besu there are flags for it: 

---

Specifies the keystore using the [--rpc-http-tls-keystore-file](https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#rpc-http-tls-keystore-file) option.
Specifies the file that contains the password to decrypt the keystore using the [--rpc-http-tls-keystore-password-file](https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#rpc-http-tls-keystore-password-file) option.
[Specifies the clients](https://besu.hyperledger.org/en/stable/HowTo/Configure/TLS/Configure-TLS/#create-the-known-clients-file) allowed to connect to Besu using the [--rpc-http-tls-known-clients-file](https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#rpc-http-tls-known-clients-file) option.
specifies the Java cipher suites using the [--rpc-http-tls-cipher-suite](https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#rpc-http-tls-cipher-suite) option.
specifies the TLS protocol version using the [--rpc-http-tls-protocol](https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#rpc-http-tls-protocol) option.

---
https://besu.hyperledger.org/en/stable/HowTo/Configure/TLS/Configure-TLS/

So to handle TLS we need to pass something like this to the client, otherwise you cannot trust this connection. So until we setup OkHttpClient with our tls keys, we cannot support it.
This way we need to implement something like this:
https://stackoverflow.com/a/67263724
or with okhttp-tls it looks less bulky:
https://github.com/square/okhttp/tree/master/okhttp-tls
Overall, we cannot trust server just because it turned https on, that's my understanding. That's why I see it as different feature

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jClientBuilder.java
zilm13: https://github.com/TooTallNate/Java-WebSocket/blob/master/src/main/example/SSLClientExample.java#L79
check this example, we need a way to pass all of this inside (plus self-signed extra), and web3j client doesn't provide open api for it

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jWebsocketClient.java
zilm13: Good point, fixing

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jClientBuilder.java
zilm13: Do we have any not self signed https api to test with? Infura or smth like this?

zilm13:
I got this bug with no execution client running and it looks very bad
```
java.util.concurrent.CompletionException: java.lang.IllegalStateException: WebSocketClient objects are not reuseable
```
Thinking what we could do with it. The pity is even we reinit it, request could be made with old Service (we couldn't swap client in service too)

zilm13:
I have a fix, I don't like it, so I submitted https://github.com/web3j/web3j/pull/1645
Hope it will be merged.
For connection exception the fixes should be applied to https://github.com/TooTallNate/Java-WebSocket first as it's not thrown from there. Looks like several places should be fixed.

zilm13:
@ajsutton I see people use `web3j`  for `wss`, I added it to supported, though I guess it will work only with not self-signed certificates. Could you explain this please, I don't understand how it works

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3jClientBuilder.java
ajsutton: There's a few ways to deal with self-signed certs - one as you point out is to build in functionality to add to the list of trusted certs.
But you can also either get a cert from somewhere that is trusted or add to the java trusted certs (https://docs.oracle.com/cd/E19906-01/820-4916/geygn/index.html).

Mostly I'd expect people not to use SSL except in big setups where they'll likely have an internal cert authority they already add as trusted for various things and can do that.

We already support https for the eth1-endpoint as it's required for use with Infura and that's the same web3j client.

zilm13: ok, right
we could definitely support it for signed certs

zilm13:
Btw found this https://docs.web3j.io/4.8.7/advanced/tls_websockets/
so it could be setup for self-signed certs in future, I guess

zilm13:
@ajsutton I've added IPC support, though I was not able to make it work with unicode filepaths. I put this on hold until web3j PR is processed. Recreate `WebSocketService` instance for every new connection is definitely a bad way. I'll add everything to `eth1-endpoint` when we have understanding with WebSockets.

zilm13:
Personally I don't like init scheme in `Web3JClient` much  but otherwise we are forced to keep all protocol-specific logic in `Web3jClientBuilder` which I dislike more. Maybe there is another good option I missed?

ajsutton (approved):
LGTM.  Could you add an entry to `CHANGELOG` for this please since supporting web sockets and ipc connections is worth noting.

PR #5058, opened 2022-02-25
Make chain head update synchronous

ajsutton:
## PR Description
Basic block data is taken synchronously from protoarray with the actual block and state being loaded async. This allows the new chain head to be set synchronously, avoiding reordering issues that previously needed to be handled.

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #4968, opened 2022-02-10
[Merge] Support execution_optimistic flag in REST API


ajsutton:
### Description
Implement changes to reflect optimistic sync status as per https://github.com/ethereum/beacon-APIs/pull/190

Need to make sure this only shows up when bellatrix is enabled.

ajsutton:
This turns out to be a fair bit of work.  We can ensure the new field doesn't appear when bellatrix isn't enabled with:

```
  @JsonInclude(Include.NON_NULL)
  @JsonProperty("execution_optimistic")
  public final Boolean execution_optimistic;
```

and then setting it to `null` when it shouldn't appear.

However there's quite a few changes required to be able to return optimistic blocks/states and track if the chain head was optimistic at the time we gathered the data.

Firstly, we currently based most requests on the fully validated chain head so while retrieving blocks by root would currently include optimistic blocks, retrieving by slot wouldn't return anything after the fully validated head.  We may want to change things so that we actually only track one chain head - the optimistic head, along with a record of whether the chain head is optimistic or not. With the REST API moving over to include optimistic blocks, I'm not sure there's a need to not update the chain head - we would need to adjust the `Syncing` log messages. We already explicitly check if the relevant block is optimistic and refuse to publish blocks and attestations.

Then for returning if the data was affected by optimistic sync, we need to ensure we do it in a thread safe way. We may change between optimistic and normal sync while we're executing the query so need to get the chain head and whether it's optimistic in an atomic operation to ensure they're consistent.

Likely we'd need to update `BlockSelector` and `StateSelector` to return both the block/state and a flag indicating if it's optimistic or not.

ajsutton:
Need to add execution_optimistic to the finalized_head event.

PR #5062, opened 2022-02-25
Only track the optimistic chain head

ajsutton:
## PR Description
Given all functions are now expected to operate off the optimistic chain head and we block creating blocks and attestations for optimistic blocks already, there's no need to track the validated ancestor head.

## Fixed Issue(s)
Part of #4968 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5065, opened 2022-02-28
add apt-get upgrade to docker images

rolfyone:
It looks like best practice is to update on the same line (and every line) as the main apt-get due to caching etc, so I've done that also.

## Documentation

## Changelog


ajsutton (commented):
Ok ran some tests to confirm this does indeed help - libsasl2-2 wasn't up to date even with the latest ubuntu base image and with this change it was updated.

My suggestion though would be that rather than adding `apt-get update` to the line that installs curl, combine installing curl with he update and upgrade line.  Essentially we want to do all apt operations in the one `RUN` line which saves us having to update twice.

ajsutton (approved):
LGTM.

PR #5066, opened 2022-02-28
Delete remote keys

### LINKED ISSUE #5059: [keymanager api] support delete remote managed key
ref. https://ethereum.github.io/keymanager-APIs/#/Remote%20Key%20Manager/DeleteRemoteKeys

In a very similar way to how local keys are removed, we will delete remote keys.  Because we're only dealing with the definition file and not a password also, the process should be a lot simpler, and we're just re-using the DeletableSigner interface for removal.

Again, because this is a batch process, we will not be failing the entire process unless the initial request fails to parse.

The request body is the same as the local keymanager request body, but the response is different, because we don't include any slashing protection data on delete (which also implies that doesn't need to be fetched when looking at the delete code)

Ordering is important, so perform actions in the same order that the local delete is performed is probably the best guidance on that...

courtneyeh:
fixes #5059 

## Documentation

## Changelog


validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -110,9 +102,53 @@ void shouldThrowExceptionWhenAddValidator() {
...
 
   @Test
-  void shouldThrowExceptionWhenDeleteValidator() {
-    assertThatThrownBy(() -> validatorSource.deleteValidator(BLSPublicKey.empty()))
-        .isInstanceOf(UnsupportedOperationException.class);
+  void shouldReturnErrorWhenDeleteReadOnlyValidator() {
+    assertThat(validatorSource.deleteValidator(BLSPublicKey.empty()))
+        .isEqualTo(
+            DeleteKeyResult.error(
+                "Cannot delete validator from read-only local validator source."));
```
rolfyone: not local :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/apis/DeleteRemoteKeys.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public DeleteRemoteKeys(final KeyManager keyManager) {
+    super(
+        EndpointMetadata.delete(ROUTE)
+            .operationId("DeleteRemoteKeys")
+            .summary("Delete Remote Keys")
+            .withBearerAuthSecurity()
+            .description(
+                "DELETE must delete all keys from `request.pubkeys` that are known to the validator client and exist in its persistent storage."
+                    + "<br>DELETE should never return a 404 response, even if all pubkeys from request.pubkeys have no existing keystores.")
```
rolfyone, line 38: we'll delete any referenced remote keys but not any keys... eg if you reference a key thats in local signers, we won't (shouldn't) be deleting that one...

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalValidatorSourceTest.java
```diff
@@ -110,9 +102,53 @@ void shouldThrowExceptionWhenAddValidator() {
...
+  @Test
+  void shouldReturnErrorWhenDeleteNonExistentValidator(@TempDir Path tempDir) {
+    final BLSPublicKey publicKey = dataStructureUtil.randomPublicKey();
+    final ExternalValidatorSource externalValidatorSource =
+        newExternalValidatorSource(tempDir, false);
+
+    final DeleteKeyResult result = externalValidatorSource.deleteValidator(publicKey);
+    assertThat(result.getStatus()).isEqualTo(DeletionStatus.ERROR);
+    assertThat(result.getMessage())
+        .isEqualTo(
```
rolfyone: should be `DeletionStatus.NOT_FOUND`

rolfyone: I think the distinciton between error and not_found in this case is 
'we had nothing in memory and nothing on disk' == not_found
'we had stuff, and we tried and failed to clean it up' == error (eg. file was read only which should be v. unlikely, but easy to reproduce)

rolfyone: equally delete of a readonly should be an error...

rolfyone: or a pubkey that's actually present in a different source but not present in the remote read-write - also should be error ...

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
```diff
@@ -159,6 +183,13 @@ DeleteKeyResult deleteValidator(
     return deleteKeyResult;
   }
 
+  DeleteKeyResult deleteExternalValidator(final Validator activeValidator) {
+    final Signer signer = activeValidator.getSigner();
+    signer.delete();
+    LOG.info("Removed validator: {}", activeValidator.getPublicKey().toString());
```
rolfyone: I wonder if we specifically want to say `removed external validator reference` or something

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -142,6 +142,14 @@ public synchronized PostKeyResult loadLocalMutableValidator(
     return PostKeyResult.success();
   }
 
+  public DeleteKeyResult deleteExternalMutableValidator(final BLSPublicKey publicKey) {
+    if (mutableExternalValidatorSource.isEmpty()) {
+      return DeleteKeyResult.error(
+          "Unable to delete validator, could not determine the storage location.");
+    }
```
rolfyone: I'd be saying `external validator` so we can understand what's going on...

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
```diff
@@ -130,6 +131,33 @@ public synchronized DeleteKeysResponse deleteValidators(
...
+    for (final BLSPublicKey publicKey : validators) {
+      Optional<Validator> maybeValidator =
+          validatorLoader.getOwnedValidators().getValidator(publicKey);
+
+      // read-only check in a non-destructive manner
+      if (maybeValidator.isPresent() && maybeValidator.get().isReadOnly()) {
+        deletionResults.add(DeleteKeyResult.error("Cannot remove read-only validator"));
+        continue;
+      }
+      // delete validator from owned validators list
```
rolfyone, line 148: after this, check the validator isn't local. should probably check on the local side that it *is* local... otherwise the call would fail anyway below

rolfyone: im not 100% sure we can get into that position, did your investigations show that kind of behaviour?

rolfyone (approved):
LGTM just a behaviour question, and we'll cover that in acceptance test change so could potentially carry that over...

PR #5067, opened 2022-02-28
Include meta data in BlockSelector result

ajsutton:
## PR Description
Update `BlockSelector` to return the requests block/blocks along with the metadata for those blocks (currently milestone and if optimistic sync affects the result).

While currently the metadata is immediately discarded, this will allow us to gradually migrate the REST APIs that return block data to use the attached metadata instead of recalculating it in a way that may be inconsistent because of concurrent changes to the optimistic sync state or chain head.

## Fixed Issue(s)
#4968 

## Documentation

## Changelog


data/provider/src/main/java/tech/pegasys/teku/api/ObjectAndMetaData.java
```diff
@@ -0,0 +1,72 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.api;
+
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
+import tech.pegasys.teku.spec.SpecMilestone;
+
+public class ObjectAndMetaData<T> {
```
ajsutton, line 20: This class is more generic than we need right now but the next step is to add beacon state with metadata and we don't want to have two separate definitions of the metadata so introducing a class that can be shared now. Little bit of a YAGNI violation but I'm feeling lucky. :)

rolfyone (approved):
LGTM

PR #5072, opened 2022-03-01
Add JSON serialization for SSZ types

ajsutton:
## PR Description
Mega-PR that defines a `TypeDefinition` allowing JSON serialisation, deserialisation and OpenAPI spec for all SSZ schemas.

This is inevitably going to wind up being big, but will try to some of the module shuffles separately to make it easier to see the actual changes.

Apart from potentially being able to pretty-print our SSZ objects, this will wind up being used as part of converting the beacon REST API over to the modern approach we're using with the key manager API.

## Documentation

## Changelog


infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/ByteTypeDefinition.java
```diff
@@ -0,0 +1,39 @@
...
+import com.fasterxml.jackson.core.JsonParser;
+import java.io.IOException;
+
+class ByteTypeDefinition implements DeserializableTypeDefinition<Byte> {
+
+  @Override
+  public void serializeOpenApiType(final JsonGenerator gen) throws IOException {
+    gen.writeStartObject();
+    gen.writeStringField("type", "number");
+    gen.writeStringField("format", "byte");
```
rolfyone: number? i thought bytes serialized to string as hex...

ajsutton: I'm unclear what the right answer for this one is - This is for a single byte, not a byte array (which does serialise as hex).  I don't think we actually use it anywhere currently so there isn't an existing example I'm aware of to check against.

rolfyone (approved):
mostly LGTM but question about type/format used for bytes

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/ByteTypeDefinition.java
rolfyone: if it serializes to hex, then its string... unless we're not quoting

PR #5073, opened 2022-03-01
Move Bytes4, Bytes8 and Bytes20 to new bytes module

ajsutton:
## PR Description
Rather than embedding these generic bytes classes in the ssz module, split them out into their own module.

## Documentation

## Changelog


rolfyone (approved):
brief scan, idea is sound. LGTM

PR #5068, opened 2022-03-01
Add execution_optimistic to GetBlockRoot

ajsutton:
## PR Description
Adds `execution_optimistic` field to get block root API if and only if bellatrix is enabled.

Builds on #5067 

## Fixed Issue(s)
#4968 

## Documentation

- [x] I thought about documentation and added the `documentation` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #5079, opened 2022-03-01
Add execution_optimistic to getBlock v2

ajsutton:
## PR Description
Add `execution_optimistic` flag to getBlock v2.  Also collapse the `getBlock` and `getBlockV2` methods into one since they were exact duplicates of each other.  The v1 getBlock method doesn't get `execution_optimistic` added but the implementation of it is updated to extract the block from the metadata wrapper for consistency.

## Fixed Issue(s)
#5068 

## Documentation

## Changelog


rolfyone (approved):
LGTM, just that this will be undocumented to my understanding...

ajsutton:
Correct. Deliberately hiding the new field from openapi docs at the moment. When the PR to add this field is merged and we're exposing the bellatrix CLI options we'll unhide them.

PR #5088, opened 2022-03-02
Split out json module

ajsutton:
## PR Description
Moves JSON type definition code out of restapi into its own module.

There is a slight oddity in that the json parsers are using exceptions which are probably REST API specific so they had to be split out too.  To save created a huge amount of conflict between this and #5072 I'm going to let that slide for now but will circle back to throw a more JSON specific exception as a follow up.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5091, opened 2022-03-03
Use lowercase names for SSZ fields

ajsutton:
## PR Description
Extracted out of #5072 - use `getFieldName()` method instead of `name()` so our SSZ fields have lower case names.

Currently those names aren't used anywhere but we need the lowercase for json serialisation.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5093, opened 2022-03-03
Only send block event after import

ajsutton:
## PR Description
Send `block` events only after the block is imported.

Previously events were sent once the block passed the gossip validation rules but with the need to include the execution_optimistic flag the event now needs to be sent after import. Blocks which pass the gossip validation checks but fail state transition will no longer generate an event.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,12 @@
 For information on changes in released versions of Teku, see the [releases page](https://github.com/ConsenSys/teku/releases).
 
 ## Unreleased Changes
+
+### Breaking Changes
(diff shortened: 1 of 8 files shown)


tbenr (approved):
LGTM.
One thing: what is the behaviour of the other clients? Are we converging or diverging with this change?

ajsutton:
Pretty sure we're converging but there's been a lot of confusion about exactly when this event should be published. Including `execution_optimistic` will force everyone to publish after import otherwise you won't know if it was optimistic or not. So long term, definitely converging. :)

PR #5104, opened 2022-03-04
Update attestation validation from reject to ignore

### LINKED ISSUE #5086: Update to v1.1.10 gossip rules
The 1.1.10 spec includes a couple of changes to gossip rules:

>_[IGNORE]_ The valid aggregate attestation defined by `hash_tree_root(aggregate)` has _not_ already been seen
>   (via aggregate gossip, within a verified block, or through the creation of an equivalent aggregate locally).

We already. have this one, spec is just catching up with the revert we agreed on previously.

But this one changed from REJECT to IGNORE:

> - _[IGNORE]_ The current `finalized_checkpoint` is an ancestor of the `block` defined by `attestation.data.beacon_block_root` -- i.e.
   `get_ancestor(store, attestation.data.beacon_block_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch))
   == store.finalized_checkpoint.root`

rolfyone:
fixes #5086

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,5 +13,6 @@ For information on changes in released versions of Teku, see the [releases page]
 
 ## Unreleased Changes
 ### Additions and Improvements
+ - Updated gossip validation rules to match consensus spec v1.1.0.
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


ajsutton (approved):
LGTM. I do wonder how we'd ever get that check to fail though given blocks that don't descend from the finalised block are removed from the Store immediately. Would be interesting to investigate that further as we could potentially optimise by removing this check entirely and rely on the attestation being ignored because the block is unknown.

PR #5106, opened 2022-03-04
Use lower-case for beacon block fields

ajsutton:
## PR Description
Use lower-case for beacon block field names in preparation for JSON serialization.

Allow enums to be used as field names directly and apply this to the state field names as well.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5107, opened 2022-03-04
Add execution_optimistic to block and reorg events

ajsutton:
## PR Description
Adds `execution_optimistic` flag to block and reorg events (when bellatrix is enabled).

## Fixed Issue(s)
#4968 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
@@ -170,7 +170,6 @@ public class BlockManager extends Service
     if (pendingBlocks.contains(block)
         || futureBlocks.contains(block)
         || recentChainData.containsBlock(block.getRoot())) {
-      // TODO: Need to make sure the result sets optimistic flag correctly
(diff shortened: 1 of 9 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -168,6 +170,7 @@ private void importBlockIgnoringResult(final SignedBeaconBlock block) {
     if (pendingBlocks.contains(block)
         || futureBlocks.contains(block)
         || recentChainData.containsBlock(block.getRoot())) {
+      // TODO: Need to make sure the result sets optimistic flag correctly
```
rolfyone: is this something we should have a ticket reference for? it'll never get done potentially if its just here...

rolfyone (approved):
LGTM just a q. about ticketing a todo

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
ajsutton: I meant to go back and fix it... probably better to do as a follow up though so yeah will log a ticket and remove the TODO.

PR #5108, opened 2022-03-04
Simplify selection of spec version for getAncestor

ajsutton:
## PR Description
Previously the highest chain head slot was used which required searching the entire protoarray which is expensive. Instead we can use the slot of the head block. If that's unavailable the response will be empty anyway as the chain is unknown but can safely fall back to using the ancestor slot requested.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5114, opened 2022-03-06
Increase timeout for ATs

ajsutton:
## PR Description
Increases timeout for ATs where no output is expected (gradle is silent during test execution).

Also add an additional parallel executor to see if it reduces execution time.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5115, opened 2022-03-07
Mark required fields in OpenAPI

ajsutton:
## PR Description
Explicitly list object required fields in the OpenAPI documentation.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/OptionalSerializableFieldDefinition.java
+++ b/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/OptionalSerializableFieldDefinition.java
@@ -45,14 +45,8 @@ class OptionalSerializableFieldDefinition<TObject, TField>
 
   @Override
   public void writeOpenApiField(final JsonGenerator gen) throws IOException {
-    gen.writeObjectFieldStart(name);
-    gen.writeArrayFieldStart("allOf");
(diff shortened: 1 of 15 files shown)


infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/OptionalSerializableFieldDefinition.java
```diff
@@ -45,8 +45,14 @@ public void writeField(final TObject source, final JsonGenerator gen) throws IOE
 
   @Override
   public void writeOpenApiField(final JsonGenerator gen) throws IOException {
-    gen.writeFieldName(name);
+    gen.writeObjectFieldStart(name);
+    gen.writeArrayFieldStart("allOf");
     type.serializeOpenApiTypeOrReference(gen);
+    gen.writeStartObject();
+    gen.writeBooleanField("required", false);
```
rolfyone: `required:true` is probably more important than required:false, that flags things as not optional, and would put a red `*` in the openapi...

ajsutton: Right well that's actually easier - we can just explicitly list the required fields in each object, so have switched over to doing it that way.

rolfyone (approved):
LGTM

PR #5116, opened 2022-03-07
Migrate GetIdentity REST API to new framework

ajsutton:
## PR Description
As a starting point to migrate the Beacon REST API over to the new framework used by the validator REST API, convert the GetIdentity function over.

We can't combine the two approaches to OpenAPI generation so while the migrated endpoints can generate OpenAPI from their type definitions, the OpenAPI continues to come from the annotations. An integration test has been setup to check the OpenAPI generated from type definitions is what we expect so that when we do finish switching over the new OpenAPI is in good shape.

The legacy OpenAPI plugin detects the annotations on any method called `handle` so the new framework `handle` method had to be renamed `handleRequest`.

Builds on #5072 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
```diff
@@ -297,6 +303,19 @@ private static OpenApiOptions getOpenApiOptions(
     return options;
   }
 
+  private static Info createApplicationInfo() {
+    return new Info()
+        .title(StringUtils.capitalize(VersionProvider.CLIENT_IDENTITY))
+        .version(VersionProvider.IMPLEMENTATION_VERSION)
+        .description(
+            "A minimal API specification for the beacon node, which enables a validator "
+                + "to connect and perform its obligations on the Ethereum 2.0 phase 0 beacon chain.")
```
rolfyone: interesting that this says phase0, maybe we should update that...

rolfyone: interesting that this says phase0, maybe we should update that...

infrastructure/http/src/main/java/tech/pegasys/teku/infrastructure/http/RestApiConstants.java
```diff
@@ -113,4 +113,19 @@
   public static final String CACHE_NONE = "max-age=0";
   // Finalized max-age equates to 1 year
   public static final String CACHE_FINALIZED = "max-age=31556952";
+
+  public enum CacheLength {
```
rolfyone: nit: not really sure this adds much value over string constants above

rolfyone: nit: not really sure this adds much value over string constants above

rolfyone (approved):
LGTM overall

infrastructure/http/src/main/java/tech/pegasys/teku/infrastructure/http/RestApiConstants.java
ajsutton: I added it just to move towards more type safety rather than having everything defined as a `String`. Probably should be in the restapi module though so will move it there.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
ajsutton: Changed to just `obligations on the Ethereum beacon chain`.

PR #5118, opened 2022-03-07
Add pretty print subcommand

### LINKED ISSUE #1714: [feature] teku `pretty` print subcommand
### Description

In (under development) new beacon-fuzz tooling, we are using client tools like `zcli`, `lcli`, `ncli`, ... and we compared the result/return value between all tools.

I would to have a `teku` subcommand to preform `pretty` print on most common type, like those ones:
- Attestation
- BeaconBlock
- BlockHeader
- Deposit
- ProposerSlashing
- SignedBeaconBlock
- SignedBlockHeader
- SignedVoluntaryExit
- BeaconState
- VoluntaryExit

That will allow us to have an easy way to test SSZ decoding and parsing between `teku` and other clients. We consider `zcli` as the reference spectest tool.

ajsutton:
## PR Description
Add a subcommand to convert SSZ object to pretty printed JSON. Automatically supports snappy decompression so you can view SSZ objects from the reference tests.

It's under debug-tools so not intended for typical user use, primarily a development tool.

## Fixed Issue(s)
fixes #1714 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5122, opened 2022-03-08
Support custom descriptions on type definitions

ajsutton:
## PR Description
Support adding custom definitions to type definitions so we can provide more information even when re-using common types (especially useful for SSZ and primitive types).

Unfortunately the swagger-ui doesn't support overriding the description when using references that the OpenAPI spec allows so we have to ensure we inline the type when a custom description is used.  Hence `name` and `title` have been separated so we can separately control whether the type is referenced or included inline and what the name of that type is.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5123, opened 2022-03-08
Make TerminalPowBlockMonitor less noisy

ajsutton:
## PR Description
Reduce log level of the syncing message from TerminalPowBlockMonitor.  We already have `Syncing` log messages making it clear the node is syncing so don't need an extra info message every slot

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5126, opened 2022-03-08
GetBlockHeader should set canonical=false if appropriate

### LINKED ISSUE #5080: getHeader and getHeaders REST API incorrectly report all blocks as canonical
### Description
The `/eth/v1/beacon/headers/:block_id` and `/eth/v1/beacon/headers` REST APIs incorrectly report all blocks as canonical.

rolfyone:
If selecting by block root, you can select a non canonical block, but the header wasn't correctly filling in the canonical boolean.

Added tests for GetHeader, also added a basic integration test for `GetBlockHeaders`, which really needs to be expanded upon so that it can return non canonical headers.

fixes #5080

## Documentation

## Changelog


data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -121,7 +121,11 @@ public GenesisData getGenesisData() {
 
   public SafeFuture<Optional<ObjectAndMetaData<BlockHeader>>> getBlockHeader(
       final String slotParameter) {
-    return fromBlock(slotParameter, block -> new BlockHeader(block, true));
+    return fromBlock(
+        slotParameter,
+        block ->
+            new BlockHeader(
+                block, combinedChainDataClient.isCanonicalBlock(block.getSlot(), block.getRoot())));
```
ajsutton: We should just get the canonical flag from the block meta data rather than recalculating it here (as we may have since re-orged and get inconsistent results).

data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
```diff
@@ -110,18 +111,20 @@ public StateSelector justifiedSelector() {
...
-                            addMetaData(
-                                state,
-                                isCheckpointOptimistic(
-                                    BeaconBlockHeader.fromState(state).getRoot()))));
+                        state -> {
+                          BeaconBlockHeader header = BeaconBlockHeader.fromState(state);
+                          return addMetaData(
+                              state,
+                              isCheckpointOptimistic(header.getRoot()),
+                              client.isCanonicalBlock(header.getSlot(), header.getRoot()));
```
ajsutton: The justified state is always canonical so can just use `true` here.

data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
```diff
@@ -145,11 +149,13 @@ public StateSelector forStateRoot(final Bytes32 stateRoot) {
...
-                            addMetaData(
-                                state,
-                                client.isOptimisticBlock(
-                                    BeaconBlockHeader.fromState(state).getRoot()))));
+                        state -> {
+                          BeaconBlockHeader header = BeaconBlockHeader.fromState(state);
+                          return addMetaData(
+                              state,
+                              client.isOptimisticBlock(header.getRoot()),
+                              client.isCanonicalBlock(header.getSlot(), header.getRoot()));
```
ajsutton: This should use `state.getSlot()` as the block header might be from an earlier slot where the block is canonical but the state isn't because the state is from an empty slot and the canonical chain has a block present.

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -333,7 +337,13 @@ private Bytes48 getBytes48FromParameter(final String validatorParameter) {
...
                   bellatrixEnabled ? executionOptimistic : null,
                   blockAndMetaDataList.stream()
-                      .map(blockData -> new BlockHeader(blockData.getData(), true))
+                      .map(
+                          blockData ->
+                              new BlockHeader(
+                                  blockData.getData(),
+                                  combinedChainDataClient.isCanonicalBlock(
+                                      blockData.getData().getSlot(),
+                                      blockData.getData().getRoot())))
```
ajsutton: Should just use the value from the meta data here too.

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
rolfyone: we've only got the block here, im not sure how you're getting to block metadata...

ajsutton: The BlockSelector returns `BlockAndMetaData`.

ajsutton: You probably can't use the `fromBlock` convenience but if you in-line that call you'll find you have `BlockAndMetaData`.

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
rolfyone: ok this is blockAndMetadata so i can understand that bit...

ajsutton (approved):
LGTM.

PR #5128, opened 2022-03-08
forkChoiceUpdated test with MockWebServer

tbenr:
## PR Description

Introduce the usage of `MockWebServer` to test `Web3JExecutionEngineClient`

Currently used only to forkChoiceUpdated

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/build.gradle
+++ b/ethereum/executionlayer/build.gradle
@@ -16,5 +16,7 @@ dependencies {
   runtimeOnly 'io.jsonwebtoken:jjwt-impl'
   runtimeOnly 'io.jsonwebtoken:jjwt-jackson'
 
+  testImplementation 'com.squareup.okhttp3:mockwebserver'
   testImplementation testFixtures(project(':ethereum:spec'))
+  testImplementation testFixtures(project(':infrastructure:time'))
 }
--- a/ethereum/executionlayer/build.gradle
+++ b/ethereum/executionlayer/build.gradle
@@ -16,7 +16,8 @@ dependencies {
   runtimeOnly 'io.jsonwebtoken:jjwt-impl'
   runtimeOnly 'io.jsonwebtoken:jjwt-jackson'
 
-  testImplementation 'com.squareup.okhttp3:mockwebserver'
   testImplementation testFixtures(project(':ethereum:spec'))
   testImplementation testFixtures(project(':infrastructure:time'))
+
+  integrationTestImplementation 'com.squareup.okhttp3:mockwebserver'
 }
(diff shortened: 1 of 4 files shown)


ethereum/executionlayer/build.gradle
```diff
@@ -16,5 +16,7 @@ dependencies {
   runtimeOnly 'io.jsonwebtoken:jjwt-impl'
   runtimeOnly 'io.jsonwebtoken:jjwt-jackson'
 
+  testImplementation 'com.squareup.okhttp3:mockwebserver'
```
ajsutton: Given mockwebserver starts an actual server and network port, it would probably be better to make these integration tests.

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3JExecutionEngineClientTest.java
```diff
@@ -496,6 +511,77 @@ void shouldTransitionConfigurationRoundtrip() {
...
+    assertThat(request.getBody().readString(StandardCharsets.UTF_8))
+        .isEqualTo(
+            "{\"jsonrpc\":\"2.0\",\"method\":\"engine_forkchoiceUpdatedV1\",\"params\":["
+                + "{\"headBlockHash\":\"0x235bc3400c2839fd856a524871200bd5e362db615fc4565e1870ed9a2a936464\","
+                + "\"safeBlockHash\":\"0x367cbd40ac7318427aadb97345a91fa2e965daf3158d7f1846f1306305f41bef\","
+                + "\"finalizedBlockHash\":\"0xfd18cf40cc907a739be483f1ca0ee23ad65cdd3df23205eabc6d660a75d1f54e\"},"
+                + "{\"timestamp\":\"0xa\","
+                + "\"prevRandao\":\"0x367cbd40ac7318427aadb97345a91fa2e965daf3158d7f1846f1306305f41bef\","
+                + "\"suggestedFeeRecipient\":\"0xfd18cf40cc907a739be483f1ca0ee23ad65cdd3d\"}],"
+                + "\"id\":0}");
```
ajsutton: I wonder if it would be better to parse the expected and actual into generic JsonNode and compare those so that the specific string formatting doesn't matter as long as it's semantically the same JSON.

ajsutton (commented):
Definitely a good test to have but probably better as an integration test if we can.

ethereum/executionlayer/build.gradle
tbenr: I got inspired by `MetricsPublisherTest`, which should be migrated to IT then

ajsutton: Yeah definitely.  I've been meaning to find which test keeps opening a socket - MacOS pops up a dialog asking permission for it every time I run a build. :)

ajsutton (approved):
LGTM.

PR #5137, opened 2022-03-09
Update getBlockHeaders to return canonical and non canonical block headers.

### LINKED ISSUE #5127: GetBlockHeaders should return non canonical block headers if appropriate
GetBlockHeaders needs to use a selector that allows non canonical blocks so that it can respond with canonical=false in the case when block headers are not on the canonical chain.

This will also mean multiple objects can be output, so it'll require a little rework.

Follows on from #5080 where non canonical blocks may come out of GetBlockHeader.

rolfyone:
Added an integration test to show non canonical block headers being returned.

partially addresses #5127

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -21,5 +21,6 @@ For information on changes in released versions of Teku, see the [releases page]
 - Reduced CPU usage when finding ancestor block roots.
 - Updated gossip validation rules to match consensus spec v1.1.0.
 - Reduced memory usage when deduplicating gossip messages.
+- GetBlockHeaders `/eth/v1/beacon/headers` can now return non canonical block headers if they are available.
```
ajsutton: nit: We can drop the "if they are available" here - non-canonical blocks are always available for non-finalised sections of the chain.

ajsutton (approved):
LGTM.

PR #5138, opened 2022-03-09
Mark blocks with default payload as VALID

### LINKED ISSUE #5082: [Merge] Mark nodes with default execution payload as VALID when first loading
### Description
When creating the protoarray state from disk, we don't have any record of if the execution payloads have been executed or not so default to marking nodes as optimistic (they are then marked valid when fork choice updated first runs if the EL is online).

For nodes prior to the merge block (where there is no execution payload to execute) we should automatically mark these as valid instead of optimistic as we won't receive any updates from the EL for them.

ajsutton:
## PR Description
When importing blocks to protoarray, mark any block with a default execution payload as fully validated even if the block processor may optimistically execute them.

## Fixed Issue(s)
fixes #5082

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5140, opened 2022-03-09
Make execution_optimistic and canonical flags consistent

ajsutton:
## PR Description
If the chain head is optimistically executed, that "taints" whether a block is canonical or not and so the `execution_optimistic` flag needs to be set.  Similarly the finalised and justified checkpoints may have been set based on later imported blocks so should be marked optimistic whenever the head is optimistic (and because they must be ancestors of the head, can't be optimistic unless the chain head is).

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5141, opened 2022-03-09
add unprefixed hex jwt secret load test and ignore (CR)LF

tbenr:
## PR Description
Since the specs are not clear if the jwt secret needs to be stored in file using "0x" prefix or not, I add a test to be sure we are able to load unprefixed key too.

it also ignores '\n' from the file, which may cause users heachache

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoader.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoader.java
@@ -35,7 +35,6 @@ public class JwtSecretKeyLoader {
   private static final Logger LOG = LogManager.getLogger();
 
   public static final String JWT_SECRET_FILE_NAME = "ee-jwt-secret.hex";
-  public static final String UNPREFIXED_JWT_SECRET_FILE_NAME = "unprefixed-ee-jwt-secret.hex";
   private final Optional<String> jwtSecretFile;
   private final Path beaconDataDirectory;
 
@@ -72,7 +71,7 @@ public class JwtSecretKeyLoader {
   private SecretKeySpec loadSecretFromFile(final String jwtSecretFile) {
     final Path filePath = Paths.get(jwtSecretFile);
     try {
-      final Bytes bytesFromHex = Bytes.fromHexString(Files.readAllLines(filePath).get(0));
+      final Bytes bytesFromHex = Bytes.fromHexString(Files.readString(filePath).trim());
       return new SecretKeySpec(bytesFromHex.toArray(), SignatureAlgorithm.HS256.getJcaName());
     } catch (final FileNotFoundException | NoSuchFileException e) {
       throw new InvalidConfigurationException(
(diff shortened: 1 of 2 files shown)


ajsutton (commented):
I don't think we should be writing out two files.  The 0x prefix version is the canonical form for hex data so I think we just write that, but should ensure we accept both. 

For the record, Geth writes the secret with the 0x prefix when it generates the file.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoader.java
```diff
@@ -71,7 +72,7 @@ private void writeGeneratedKeyToFile(final Key key) {
   private SecretKeySpec loadSecretFromFile(final String jwtSecretFile) {
     final Path filePath = Paths.get(jwtSecretFile);
     try {
-      final Bytes bytesFromHex = Bytes.fromHexString(Files.readString(filePath));
+      final Bytes bytesFromHex = Bytes.fromHexString(Files.readAllLines(filePath).get(0));
```
ajsutton: I'd probably just use `Field.readString(filePath).trim()` here so that we discard newline and any trailing spaces etc.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoader.java
```diff
@@ -35,6 +35,7 @@
   private static final Logger LOG = LogManager.getLogger();
 
   public static final String JWT_SECRET_FILE_NAME = "ee-jwt-secret.hex";
+  public static final String UNPREFIXED_JWT_SECRET_FILE_NAME = "unprefixed-ee-jwt-secret.hex";
```
ajsutton: I think my first comments misread this PR - we only seem to use this constant from the test and it does make sense to write out a test file without the prefix so should just move this to the test instead of production code.

ajsutton (commented):
Overreacted to the placement of the constant in production code. :)

LGTM with a couple of tweaks.

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoaderTest.java
```diff
@@ -32,12 +32,26 @@
   void testGetSecretKey_ReadSecretFromProvidedFilePath(@TempDir final Path tempDir)
       throws IOException {
     final Path jwtSecretFile = tempDir.resolve(JwtSecretKeyLoader.JWT_SECRET_FILE_NAME);
+    final Path unprefixedJwtSecretFile =
+        tempDir.resolve(JwtSecretKeyLoader.UNPREFIXED_JWT_SECRET_FILE_NAME);
+
```
ajsutton: Actually, probably worth splitting this into separate test methods.  So we test loading with prefix in one method, without prefix in one and with trailing whitespace in a third.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/client/auth/JwtSecretKeyLoader.java
tbenr: Yeah. I've put dirty code in this PR:)
Tomorrow I'll cleanup!

ajsutton (approved):
LGTM.

PR #5142, opened 2022-03-09
Replace deprecated stream() call for fastutil objs

jtraglia:
Found these using an in-development errorprone bugchecker.


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
```diff
@@ -315,7 +315,7 @@ private boolean isInHistoricBlockRoots(
               }
 
               IntSummaryStatistics inclusionDistanceStatistics =
-                  inclusionDistances.stream().collect(Collectors.summarizingInt(Integer::intValue));
+                  inclusionDistances.intStream().summaryStatistics();
```
ajsutton, line 318: ooo, that's much nicer too.

ajsutton (approved):
LGTM. Thanks.

PR #5143, opened 2022-03-09
Move ObjectAndMetadata

rolfyone:
This will need to be referenced in more places like ChainDataProvider, so this move will make that simpler.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5145, opened 2022-03-09
Fix error starting Teku with empty public keys

### LINKED ISSUE #5130: Error starting teku with empty public keys `
### Description
[Detailed description of the problem and the impact it has]
Starting Teku in remote signer mode with the following flags:
```
    exec /opt/teku/bin/teku \
        --network=prater \
        --data-base-path=/opt/teku/data \
        --eth1-endpoint=$HTTP_WEB3PROVIDER \
        --validators-external-signer-url=$HTTP_WEB3SIGNER \
        --validators-external-signer-public-keys=$PUBLIC_KEYS_COMMA_SEPARATED \
        --p2p-port=9000 \
        --rest-api-cors-origins="*" \
        --rest-api-interface=0.0.0.0 \
        --rest-api-port=$BEACON_API_PORT \
        --rest-api-host-allowlist="*" \
        --rest-api-enabled=true \
        --rest-api-docs-enabled=true \
        --log-destination=CONSOLE \
        --validators-graffiti=\"$GRAFFITI\" \
        $EXTRA_OPTS
``` 

The public keys loaded onto the flag `--validators-external-signer-public-keys` are obtained dynamically and parsed from the remote signer standard API https://eips.ethereum.org/EIPS/eip-3030#get-keys.

The public keys returned from the remote signer may be empty, this situation is not contemplated by teku, resulting in an error to start teku with empty public keys at `--validators-external-signer-public-keys`

I think this is not the ideal behavior, especially taking into account that the validator and beacon processes run in a monoservice process, so it does not allow to keep syncing the beacon chain if there are no validators loaded

### Steps to Reproduce (Bug)
[Please be as specific as possible]
start teku with an empty string in the flag `--validators-external-signer-public-keys`

I expected teku to realize that an empty string means no validators and keep running instead of throwing errors. 

Start teku with the flag `--validators-external-signer-public-keys` with an empty string result into an error
```
FATAL - Invalid configuration. Signer public key is invalid
```

The error is thrown on startup so the process keeps in a restarting status.

* Software version: [`teku --version` or look for log lines starting with `Teku version:`]  22.3.0
* Java version: [`java -version`]  official docker image
* OS Name & Version: [`cat /etc/*release`]   official docker image

courtneyeh:
fixes #5130 

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
@@ -21,6 +21,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
 import picocli.CommandLine;
 import tech.pegasys.teku.config.TekuConfiguration;
 import tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException;
@@ -123,7 +124,7 @@ public class ValidatorKeysOptions {
 
   private List<String> parseValidatorExternalKeys() {
     return validatorExternalSignerPublicKeys.stream()
-        .filter(val -> !Strings.isNullOrEmpty(val))
+        .filter(StringUtils::isNotBlank)
         .collect(Collectors.toList());
   }
 
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
@@ -20,6 +20,7 @@ import java.nio.file.Path;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
```diff
@@ -120,6 +121,12 @@ public void configure(TekuConfiguration.Builder builder) {
                     convertToPath(validatorExternalSignerTruststorePasswordFile)));
   }
 
+  private List<String> parseValidatorExternalKeys() {
+    return validatorExternalSignerPublicKeys.stream()
+        .filter(val -> !Strings.isNullOrEmpty(val))
```
ajsutton: I'd suggest using the commons lang3 `StringUtils.isNotBlank` here which will also filter out strings that are entirely whitespace. You also won't need to negate it so can use a method reference.

ajsutton (approved):
LGTM. Would just switch it over to also ignore strings that are entirely white space.

PR #5148, opened 2022-03-10
Set the optimistic flag for known blocks correctly

### LINKED ISSUE #5109: [Merge] Set execution_optimistic flag correctly for known blocks
### Description
When importing a block is skipped because it's already known, we should ensure the `BlockImportResult` is marked as an optimistic execution if the block is only optimistically valid.  Currently it always reports optimistic execution as false which may be incorrect.

ajsutton:
## PR Description
Ensure the execution_optimistic value is set correctly for known blocks.

## Fixed Issue(s)
fixes #5109 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5149, opened 2022-03-10
Return non-canonical block headers from finalized blocks.

### LINKED ISSUE #5127: GetBlockHeaders should return non canonical block headers if appropriate

rolfyone:
fixes #5127

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/blockselector/BlockSelectorFactory.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/blockselector/BlockSelectorFactory.java
@@ -15,10 +15,10 @@ package tech.pegasys.teku.api.blockselector;
 
 import static tech.pegasys.teku.spec.config.SpecConfig.GENESIS_SLOT;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.api.exceptions.BadRequestException;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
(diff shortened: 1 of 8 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/blockselector/BlockSelectorFactory.java
```diff
@@ -95,12 +95,8 @@ public BlockSelector nonCanonicalBlocksSelector(final UInt64 slot) {
...
       final ChainHead chainHead = maybeChainHead.get();
       return client
-          .getAllBlocksAtSlot(slot)
-          .thenApply(
-              blocks ->
-                  blocks.stream()
-                      .map(block -> lookupBlockData(block, chainHead))
-                      .collect(Collectors.toList()));
+          .getAllBlocksAtSlot(slot, chainHead.getRoot())
+          .thenApply(blocks -> new ArrayList<>(blocks));
```
ajsutton: Given `getAllBlocksAtSlot` is only used here we probably should modify it to just return a `List` directly and avoid the need to copy.

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
```diff
@@ -535,22 +536,53 @@ public boolean isOptimisticBlock(final Bytes32 blockRoot) {
...
+    canonicalBlock.ifPresent(block -> result.add(toBlockAndMetaData(block, true)));
+    return result;
+  }
+
+  private BlockAndMetaData toBlockAndMetaData(
+      final SignedBeaconBlock signedBeaconBlock, final boolean canonical) {
+    return new BlockAndMetaData(
+        signedBeaconBlock,
+        spec.atSlot(signedBeaconBlock.getSlot()).getMilestone(),
+        isOptimisticBlock(signedBeaconBlock.getRoot()),
```
ajsutton: This isn't quite right.  We should be marking blocks as optimistic if either the block is optimistic or the chain head is optimistic.

ajsutton (approved):
LGTM.

data/provider/src/main/java/tech/pegasys/teku/api/blockselector/BlockSelectorFactory.java
rolfyone: this one we'll put into another PR just to be clear on this comment

PR #5150, opened 2022-03-10
[Merge] Unhide merge related CLI options

### LINKED ISSUE #4874: [Merge] Unhide merge related CLI options
### Description
Currently all the merge related CLI options are hidden with a `--X` prefix.  Once we're confident they've stabilised we should unhide them and remove the prefix so they become fully supported options.

courtneyeh:
fixes #4874

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionEngineOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionEngineOptions.java
@@ -34,7 +34,7 @@ public class ExecutionEngineOptions {
   private String executionEngineEndpoint = null;
 
   @Option(
-      names = {"--Xee-version"},
+      names = {"--ee-version"},
       paramLabel = "<EXECUTION_ENGINE_VERSION>",
       description =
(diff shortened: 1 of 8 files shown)


ajsutton (commented):
I'd say `--Xee-version` can stay hidden since we'll likely wind up only maintaining one version pretty soon anyway (once kintsugi dies we can just use the kilnv2 version).

teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionEngineOptions.java
```diff
@@ -43,7 +43,7 @@
   private Version executionEngineVersion = Version.DEFAULT_VERSION;
 
   @Option(
-      names = {"--Xee-payload-builders"},
+      names = {"--ee-payload-builders"},
```
ajsutton: And payload builders is probably not ready to expose yet as the MEV stuff is still in progress.

`--ee-endpoint` and `--ee-jwt-secret` are good, though we should rename `--ee-jwt-secret` to `--ee-jwt-secret-file` to be consistent with our CLI naming

ajsutton (approved):
LGTM.

Issue #5131, opened 2022-03-08
java.lang.IllegalArgumentException: Cannot create attestation for future slot.


theRealYoshi:
### Description
[Detailed description of the problem and the impact it has]

### Steps to Reproduce (Bug)
[Please be as specific as possible]
Randomly get this error when the client runs:

ERROR - Validator   *** Failed to produce attestation  
java.lang.IllegalArgumentException: Cannot create attestation for future slot.
Requested 3327683 but current slot is 3327680

**Expected behavior:** [What you expect to happen]
It should be running smoothy

**Actual behavior:** [What actually happens]
Looks like it attest a slot in the future that does not exist yet

**Frequency:** [How regularly does it occur?]
Every 8-9 hours ish. 

### Versions (Add all that apply)
* Software version: [`teku --version` or look for log lines starting with `Teku version:`]
* teku-22.3.0
* Java version: [`java -version`]
* 
* OS Name & Version: [`cat /etc/*release`]
* 
* Docker Version: [`docker version`]
* Docker version 20.10.12,
* Cloud VM, type, size: [Amazon Web Services I3-large]
* AWS us-east-2 r5.8xlarge

ajsutton:
As discussed on discord: This message indicates that the beacon node is struggling to keep up - in this case from memory because the java heap size was too small.

That said we do plan to do some work to avoid this particular symptom of the node struggling in the future but likely it still wouldn't perform well. The beacon node should be load shedding to catch up much sooner rather than having its view of the current slot fall behind by so much.

benjaminion:
Closing as no modifications to Teku required.

PR #5171, opened 2022-03-14
Remove EventChannel indirection for time tick events

ajsutton:
## PR Description
Moves control of the TimerService into BeaconChainController so that it can call the onTick method directly rather than going via an event channel. This removes the queue of events and allows the quartz scheduler to manage back pressure if events aren't processed fast enough. It does this by skipping forward, making it less likely (though still possible) for the beacon chain's view of time to fall behind.

The scheduler has also been better configured to only use a single thread (matching the thread model of the event channel), consider an execution failed if it's delayed long enough that the next one is due and be explicit about some important defaults.

## Fixed Issue(s)
Should make #5131 less common, though long term the full solution is to remove the use of time tick events entirely and just schedule actions at the particular time they're required (and use plain system time for the clock).

## Documentation

## Changelog


services/timer/src/main/java/tech/pegasys/teku/services/timer/TimerService.java
```diff
@@ -39,34 +39,52 @@
...
+    final SchedulerFactory sf = createSchedulerFactory();
     try {
       sched = sf.getScheduler();
       job =
           newJob(ScheduledTimeEvent.class)
               .withIdentity("Timer-" + TIMER_ID_GENERATOR.incrementAndGet())
               .build();
-      job.getJobDataMap()
-          .put(TIME_EVENTS_CHANNEL, config.getEventChannels().getPublisher(TimeTickChannel.class));
+      job.getJobDataMap().put(TIME_EVENTS_CHANNEL, timeTickHandler);
```
tbenr: since it is no more a Channel, should we rename this?

tbenr (approved):
LGTM with question

PR #5172, opened 2022-03-14
Fail on missing braces

### LINKED ISSUE #4518: Add error prone check to enforce use of braces
### Description
Teku coding standard requires use of braces for all if/while/for etc cases, even when the body is only a single line.  We should add a custom error prone check to enforce this.

ajsutton:
## PR Description
Enable the MissingBraces errorprone check and fix all violations
Enable the Finally (no returns/throws in finally blocks) and TypeParameterUnusedInFormals checks now that they work on Java 12+.

fixes #4518 even though I thought it wasn't possible...

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
@@ -38,12 +38,12 @@ public class ValidateableAttestation {
   private final Supplier<Bytes32> hashTreeRoot;
   private final AtomicBoolean gossiped = new AtomicBoolean(false);
   private final boolean producedLocally;
+  private final OptionalInt receivedSubnetId;
 
   private volatile boolean isValidIndexedAttestation = false;
 
   private volatile Optional<IndexedAttestation> indexedAttestation = Optional.empty();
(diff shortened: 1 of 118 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
```diff
@@ -38,12 +38,12 @@
   private final Supplier<Bytes32> hashTreeRoot;
   private final AtomicBoolean gossiped = new AtomicBoolean(false);
   private final boolean producedLocally;
+  private final OptionalInt receivedSubnetId;
```
tbenr, line 41: ok to lose `volatile` here?

tbenr (approved):
LGTM with a doublecheck

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
ajsutton: Yep, a `final` field can't change so can't be `volatile`.

tbenr: fair enough

PR #5174, opened 2022-03-14
Some changes to peer scoring

### LINKED ISSUE #4264: enable peer scoring
need to check how it works etc, not as simple as enabling.

zilm13:
## PR Description
Made some changes with current peer scoring. Not sure dropping incoming banned peers works. Is it a correct place for that?
What I don't like in the current scoring:
- permanence of ban. We don't store it on a disk, but many peers could run for a month without update and ban will stay in place
- lack of transparency in reasons, when applying `unsuitable`. The list of disconnection reasons could be extended with time, we already had ban when we were locally shutting down, not an issue w/o persistence, but could cause an error in development. So I'm totally for ban reasons list instead of ban exclusions.

I've tried to fix it, ran it few times locally, currently testing on set of test nodes

## Documentation

## Changelog


zilm13:
About failed tests: looks like we need a way to make static peers protected from any connection restrictions. Though master version was not subject to this issue, I think it's anyway a good improvement to the peer scoring for future extending if static peers are protected. Will fix it.

zilm13:
removed disconnect of incoming banned peer for a while

zilm13:
Ok, I think it's ok to add it in this way, no any incoming bans, but some changes to avoid permanent ban as nodes could run without restart for a weeks.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPeer.java
```diff
@@ -89,11 +89,13 @@ public void checkPeerIdentity() {
     getAgentVersionFromIdentity()
         .thenAccept(
             maybeAgent -> {
-              LOG.debug("Connected peer has agent string: {}", maybeAgent.orElse("Unknown"));
+              LOG.debug(
+                  "Connected peer id {} ({}) has agent string: {}",
+                  peerId,
+                  connection.remoteAddress(),
+                  maybeAgent.orElse("Unknown"));
```
ajsutton: ooo, we should remove this log line or at least the agent string. It's not safe to log arbitrary content from external sources.  We've seen attacks where the peer agent was set to things like "Your Geth is out of date, click <malicious url> to upgrade".

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/reputation/ReputationManagerTest.java
```diff
@@ -31,7 +31,7 @@
 class ReputationManagerTest {
```
ajsutton, line 34: We should add tests to cover the different short and long bans we're now using.

ajsutton (commented):
Looks good generally, just need to make sure we add tests for the new behaviour.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPeer.java
zilm13: Maybe make it trace? It was already there, I've just added `peerId` and remote address

zilm13: ok, 2 votes, removing :)

zilm13: it's useful debug feature to understand what type and version of client is connecting to you

zilm13:
Acceptance looks flaky, not reproduced locally

zilm13:
rerun is ok

ajsutton (approved):
LGTM.

PR #5179, opened 2022-03-15
Make all constant variables use uppercase names

jtraglia:
## PR Description

Convert constant (`static final`) variable names from camel case to all uppercase with underscores.

These were identified using the JavaCase check in the [Consensys errorprone-checks repo](https://github.com/ConsenSys/errorprone-checks).

Needed to ignore some enums that needed to be lowercase. Their names are used as values and it would have required changing a lot to make those uppercase; I started down that path but quickly bailed.


ajsutton (approved):
LGTM. Thanks.

PR #5183, opened 2022-03-15
Add Teku Configuration logging on startup

zilm13:
## PR Description
Needed it for a second time, let's try out

## Documentation

## Changelog


ajsutton (changes requested):
It's not safe to log the entire configuration like this.  Things like Infura URLs have a secret ID as part of the URL which we avoid logging normally but would be logged here. And this reflection based approach is likely to be fairly brittle as things like the list of classes to ignore or just serialise as a string need to be updated.

An alternative approach may be to iterate the PicoCLI `CommandLine` object and find all the boolean options that end in `-enabled` and log a summary of those. That's much more focussed but still reports which features are and aren't enabled. It's going to be much easier for users to understand as well.

zilm13:
Nice point about logging of credentials, yeah, we should avoid this

zilm13:
@ajsutton let's try this way
It will be difficult to support uptodate though it's not critical and when smb notice that he needs new field status, it would be easy to add it. I'm not sure about all sensitive fields, included all urls. Also added option to turn off this logging if user wants to omit options logging.

ajsutton:
I'm not really convinced that it makes sense to log a giant blob of text like this - it's not a format that users are really going to understand and as you mention it's going to be regularly forgotten so not include all options.  It also doesn't actually give any real confidence that an option is actually being applied - you can check your config file and CLI args to see if an option is specified, but it's just as likely that Teku will fail to apply the option from the configuration as it is to ignore the CLI arg.

zilm13:
Sounds sensible. Let's drop it

PR #5187, opened 2022-03-16
Add labelled gauge showing the number of items in pending pools

ajsutton:
## PR Description
Adds a `beacon_pending_pool_size` gauge to report the number of items in the pending block and attestations pool.

Note this is somewhat complicated by the fact that the `MetricsSystem` doesn't just return the same labelled gauge if the same name is specified like it does for counters so we have to introduce `PendingPoolFactory` to hold the single instance of the labelled gauge.

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #5188, opened 2022-03-16
Report block import timings


ajsutton:
### Description
To improve diagnostics and aim in performance improvements, we need to establish a way to report key timings for the block import process.  This includes:

* How long after slot start the block is received
* How long between then and ForkChoice.onBlock first being called
* How long it takes to retrieve the parent state
* How long it takes to create the block post state
* How long it takes to actually import the block (ForkChoice.importBlockAndState)
* How long it takes for the transaction to commit

The aim is to get a view of what takes time for block import (in a real node).  We need to ensure that capturing the timings is extremely lightweight so it doesn't affect performance.

Reporting this data is challenging - metrics are only sampled periodically so while a gauge showing timings for the last imported block may be useful, it can also miss information between samples being taken. A histogram shows the average well but tends to hide outliers and doesn't show time periods were performance is unusually slow. Logging the data for every block is too noisy.

I'd suggest:

 1. Add a gauge to show timings for the last imported block (a labelled gauge with a label for each step and the value being the number of millis into the slot when that step was completed).
 2. If a block completes importing more than 4s into a slot (and the node is not in syncing mode), log a warning including the full timing details.

ajsutton:
Probably worth doing the logging of a warning when the block finishes importing more than 4s into the slot as the first step - that likely gives us really useful information with lots of detail but without being overwhelming since late blocks are always noteworthy as they cause incorrect head votes.

tbenr:
@ajsutton I was thinking to a different approach for metrics:

image we have a labeled gauge for each import stage
the label in this gauges are "0-50ms", "50-100ms", and so on with last ">1s"
and we may have an overall processing time gauge with wider time frames

so we can count blocks processing stages that fall into a particular timeframe
with the overall processing time working similarly

it is just matter of finding timeframe "bands" that make sense for each stage.

and we may activate it with a `--Xblock-processing-performance-metrics-enabled` or similar

tbenr:
replace "gauge" with "counter"

ajsutton:
Yeah I like that idea. If they're counters you can track the rate they  increase or just look at relative distribution between them. Might need to play with grafana dashboards to see how to best visualise it but I think we'd be able to make something work.

PR #5193, opened 2022-03-16
late block import warning

tbenr:
## PR Description

Adds a late block import event log which writes a yellow warning if block import finishes after 4s into the slot.
It tracks block arrival time and processing time

## Fixed Issue(s)
related to #5188 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportPerformance.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportPerformance.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2022 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.statetransition.block;
+
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.storage.client.RecentChainData;
+
+public class BlockImportPerformance {
+  private final TimeProvider timeProvider;
+  private UInt64 timeAtSlotStartMs;
+  private UInt64 blockArrivalTimeStamp;
+  private UInt64 importCompletedTimeStamp;
+  private UInt64 arrivalDelay;
+  private UInt64 timeWarningLimitMs;
+  private UInt64 processingTime;
+
+  public BlockImportPerformance(final TimeProvider timeProvider) {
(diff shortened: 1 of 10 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportPerformance.java
```diff
@@ -0,0 +1,68 @@
...
+  private UInt64 timeWarningLimitTimeStamp;
+  private UInt64 processingTime;
+
+  public BlockImportPerformance(final TimeProvider timeProvider) {
+    this.timeProvider = timeProvider;
+  }
+
+  public void arrival(final RecentChainData recentChainData, final UInt64 slot) {
+    blockArrivalTimeStamp = timeProvider.getTimeInMillis();
+    timeAtSlotStartTimeStamp = recentChainData.computeTimeAtSlot(slot).times(1000);
```
ajsutton: nit:We probably should use `TimeUtilities.secondsToMillis` to convert this just for clarity.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -81,22 +89,33 @@ public BlockManager(
...
           InternalValidationResult.reject("Block (or its parent) previously marked as invalid"));
     }
+
     SafeFuture<InternalValidationResult> validationResult = validator.validate(block);
     validationResult.thenAccept(
         result -> {
           if (result.code().equals(ValidationResultCode.ACCEPT)
               || result.code().equals(ValidationResultCode.SAVE_FOR_FUTURE)) {
-            importBlock(block).finish(err -> LOG.error("Failed to process received block.", err));
+            LOG.trace(
```
ajsutton, line 113: I don't think I'd bother with this trace - if the block is late we can just dump all the times in one message.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
```diff
@@ -197,6 +197,18 @@ public void transitionConfigurationRemoteTbhTbnInconsistency(final String remote
...
 
+  public void lateBlockImport(
+      final Bytes32 root,
+      final UInt64 slot,
+      final UInt64 arrivalDelayMs,
+      final UInt64 processingTimeMs) {
+    String reorgEventLog =
+        String.format(
+            "Late Block Import *** Arrival Delay: %s, Processing Time: %s: Block: %s",
+            processingTimeMs, arrivalDelayMs, LogFormatter.formatBlock(slot, root));
```
ajsutton: Arrival delay and processing time are around the wrong way.

ajsutton (commented):
Looks good, mostly just small tweaks.  Will push up some fixes.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
ajsutton: eh, it doesn't hurt - leaving it in.

ajsutton:
Going to merge this one as it's quite useful as is.  I suspect we ultimately want to do some more major changes so that the gossip layer records the received time prior to any validation and then the block + timing info gets passed up to BlockManager and that wrapper stays with the block the whole way through the process until it's finally imported. But we've covered the most common path now and that's the path that's most likely to cause issues so worth getting it in.

ajsutton (approved):

PR #5194, opened 2022-03-17
Optimise deposit processing

ajsutton:
## PR Description
Uses the validator index cache to find the index for existing validators.
Less major, precompute the SSZ field name to avoid repeated lower casing.

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #5196, opened 2022-03-17
Improve `AggregatingAttestationPool`


tbenr:
We should better control the items contained in the pool since it is expensive to select them to find good ones that can be included in block when requested by `getAttestationsForBlock`

What I can see so far:
- [x] we are pruning attestations `onSlot` event. We retain all attestations within last 2 epochs.
  - action taken: reduced to 32 slot #5197
- [x] we are not considering the retain rule in the `add` method, so attestation older than 2 epochs could be added and then pruned on the next `onSlot`
  - action taken:  none. During attestation validation we already ignores attestations older than 32 slots. (as per spec `ATTESTATION_PROPAGATION_SLOT_RANGE`).

- we can have an HARD cap of number of attestationData in each slot  to `MAX_ATTESTATIONS` (retrieved by the schema at slot)

```java
  private Optional<MatchingDataAttestationGroup> getOrCreateAttestationGroup(
      final AttestationData attestationData) {
    Set<Bytes> attestationDataRootSet =
        dataHashBySlot.computeIfAbsent(attestationData.getSlot(), slot -> new HashSet<>());
    if (attestationDataRootSet.size() > 128) {
      return Optional.empty();
    }
    attestationDataRootSet.add(attestationData.hashTreeRoot());
    return attestationGroupByDataHash.computeIfAbsent(
        attestationData.hashTreeRoot(),
        key -> new MatchingDataAttestationGroup(spec, attestationData));
  }
```

-  We could introduce another data structure `Map<Slot,Integer> numOfQueuedAttDataBySlot`  mapping `Slot -> sum(dataHashBySlot.get(s1).size()) where s1 <= Slot` and use this in `getOrCreateAttestationGroup` to avoid adding attestationData that VERY unlikely will be considered in `getAttestationsForBlock` due to the fact that there are already a lot of attestationData that will be considered first and will fill-up the upcoming block.

benjaminion:
> We retain all attestations within last 2 epochs.

Attestations older than 32 slots are valueless, so we don't need the whole 2 epochs' worth.

`current_epoch_attestations` and `previous_epoch_attestations` are no longer a thing in the Altair spec, so this could be re-engineered if they are still in use.

tbenr:
> Attestations older than 32 slots are valueless, so we don't need the whole 2 epochs' worth.

Yeah true, we are infact filtering valueless attestations but only during selecting them for block production. Definitely we can easily improve here.

PR #5197, opened 2022-03-17
Reduce attestation retention period

ajsutton:
## PR Description
Drop attestations from the AggregatingAttestationPool after 32 slots when they become worthless to include.

## Fixed Issue(s)
Quick win to start #5196.  Potentially we should have further controls on the size but this halves the size of the pool quickly and easily.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5200, opened 2022-03-18
Additional logging eth1 syncing status

frankisawesome:
Signed-off-by: Frank Li <b439988l@gmail.com>

## PR Description
In running my validator I found myself frequently sending curl requests to my Besu node for block number and checking Teku logs, if we log the eth1 block number when it's still syncing I think a lot of operators will only need to check Teku logs when their eth1 client is still syncing.

Two things I'm not sure:
- Is this gonna add too many logs (i don't think so based on the frequency the check is done)
- Is info appropriate? (i think so because I want to see it without going into debug logs)

## Documentation

None required

## Changelog

None required

### DIFF
--- a/pow/src/main/java/tech/pegasys/teku/pow/Web3jInSyncCheck.java
+++ b/pow/src/main/java/tech/pegasys/teku/pow/Web3jInSyncCheck.java
@@ -38,12 +38,17 @@ public class Web3jInSyncCheck {
         // If we're close to the chain head, consider the node in sync
         // Avoids marking the node as invalid while it imports the latest block
         LOG.debug(
-            "Eth1 endpiont {} syncing but close to head so considering it valid. Current block {} Highest block: {}",
+            "Eth1 endpiont {} syncing but close to head so considering it valid. Current block: {} Highest block: {}",
             id,
             currentBlock,
             highestBlock);
         return false;
       }
+      LOG.info(
+          "Eth1 endpoint {} syncing. Current block: {} Highest block: {}",
+          id,
+          currentBlock,
+          highestBlock);
       return true;
     } catch (final NumberFormatException | NullPointerException e) {
       LOG.error("Failed to parse syncing details from eth1 endpoint {}", id, e);


pow/src/main/java/tech/pegasys/teku/pow/Web3jInSyncCheck.java
```diff
@@ -38,12 +38,17 @@ public static boolean isSyncing(final String id, final EthSyncing syncingRespons
...
             id,
             currentBlock,
             highestBlock);
         return false;
       }
+      LOG.info(
+          "Eth1 endpoint {} syncing. Current block: {} Highest block: {}",
+          id,
+          currentBlock,
+          highestBlock);
```
ajsutton, line 51: Doesn't this wind up creating two separate log messages about the node syncing? This new one and the regular one we log about the node not being usable because it's syncing?

And the current block isn't a particularly good metric for progress during a fast sync since the blocks download pretty fast and then "stall" for long periods while the world state downloads.

I quite like the idea here, but it may be better to actually look at the Eth1 client logs to judge its syncing progress rather than the teku logs?

frankisawesome: I think if double logging is a concern it seems like `Web3jEth1Provider.java:264` can be replaced, I also like the idea of having highest eth1 block easily lookupable in teku logs

ajsutton: I think the big concern for me is that the current vs highest block is not the metric you want to look at to determine how much longer an EL has to go before being in sync. It's the world state that's most important so printing the block information here will be very misleading and cause people to worry that the EL isn't making progress because the current block stops increasing while the world state downloads.

It also turns out to be very useful to have the warning messages that the eth1 node is syncing have exactly the same content as it makes them much easier to filter out in tools like Kibana.

frankisawesome: hmm yeah I think that makes sense, do you know if we have any api for world state downloads/is it possible to add one easily in besu?

Issue #5151, opened 2022-03-10
Single node Bellatrix AT


ajsutton:
### Description
Add an acceptance test to confirm that a single node with Teku and an EL can start pre-merge, produce the transition block and progress the chain post-merge.

jimmygchen:
I've made some progress with the merge acceptance tests - have both EL & CL running, and able to reach TTD. However I'm having issue with Besu's terminal block check. I will do a bit more digging on Besu repo tomorrow .. but would be amazing if you could take a look to see if there's anything obvious to you, and whether I'm going in the right direction. 

https://github.com/ConsenSys/teku/compare/master...jchen86:master

I've pulled latest Teku master branch, and running besu develop image for this test.

Two issues I've noticed in logs:

1. Unable to create valid blocks after POW reach TTD (Besu)

```
2022-03-14 01:59:29.325+00:00 | docker-java-stream--1387523461 | DEBUG | BesuNode | 2022-03-14 01:59:29.352+00:00 | vert.x-worker-thread-0 | WARN  | MergeCoordinator | checking ancestor 199 (0x0c3dd4048aac142e7dadf6537b077e6b755b5e09636b124c41958a724d802f16) is valid terminal PoW for 200 (0x1000474634a40f5fb2143ece15be240522b7dfeafd44d02789dbfa029ece51be)
2022-03-14 01:59:29.325+00:00 | docker-java-stream--1387523461 | DEBUG | BesuNode | false
2022-03-14 01:59:29.325+00:00 | docker-java-stream--1387523461 | DEBUG | BesuNode | 2022-03-14 01:59:29.352+00:00 | vert.x-worker-thread-0 | WARN  | MergeCoordinator | Couldn't find terminal block, no blocks will be valid
```

2. After merge transition, Teku is unable to obtain a payloadId. I suspect this is related to the above.

```
2022-03-14 02:54:14.999+00:00 | docker-java-stream--949041815 | DEBUG | TekuNode | 2022-03-14 02:54:15.014 ERROR - Validator   *** Failed to produce block  Slot: 11 Validator: 893c934
2022-03-14 02:54:14.999+00:00 | docker-java-stream--949041815 | DEBUG | TekuNode | java.lang.IllegalStateException: Unable to obtain a payloadId
[...]

PR #5201, opened 2022-03-18
Add Bellatrix AT to test merge transition

jimmygchen:
## PR Description

Add an acceptance test to test the merge transition. Addresses #5151:
> Add an acceptance test to confirm that a single node with Teku and an EL can start pre-merge, produce the transition block and progress the chain post-merge.

The test verifies that the transition block should have a non-default execution payload and then checks that the chain continues to finalize beyond the slot that merge happens in.

This test was initially failing due to two issues, now passing with a custom Besu build and workaround below:
1. **Besu node stuck at TTD**: fixed in this Besu [PR](https://github.com/hyperledger/besu/pull/3607) and is under review. Have tested the Besu branch locally and it fixes the issue. 
2. **Teku node unable to obtain a payloadId after the merge**: (This now works after setting `Xstartup-target-peer-count` to 0 and updating the subscriber sync states in `NoopSyncService`. Thanks @ajsutton)
> ~~This only happens on a single node network. This issue is that `PayloadAttributesCalculator` only produce payload if node is in sync. `ForkChoiceNotifierImpl` assumes we are not in sync at startup; and for single node network this sync status change does not seem to get triggered. I've workaround this by adding an additional Teku node in this test for now, but be good to understand if there's any implication changing initial `inSync` value to `true` in `ForkChoiceNotifierImpl`, or better alternatives to support single node network (if needed).~~

## Documentation

## Changelog


jimmygchen:
The test now passes with a custom Besu build from here: https://github.com/hyperledger/besu/pull/3607

jimmygchen:
This test now passes on this branch against `besu:develop` image

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -291,6 +298,25 @@ public void waitForNewFinalization() {
...
+          assertThat(block.get()).isInstanceOf(SignedBeaconBlockBellatrix.class);
+
+          final SignedBeaconBlockBellatrix bellatrixBlock = (SignedBeaconBlockBellatrix) block.get();
+          final ExecutionPayload executionPayload = bellatrixBlock.getMessage().getBody().executionPayload
+              .asInternalExecutionPayload(spec, bellatrixBlock.getMessage().slot)
+              .orElseThrow();
+
+          assertThat(executionPayload.isDefault()).isFalse();
+          LOG.debug("Non default execution payload found at slot " + bellatrixBlock.getMessage().slot);
+        }, 300, MINUTES);
```
rolfyone: nit: need to set a more realistic time for wait here. The intent is basically a time at which to call it a failure. If a normal message should come within a minute, then i could see a 5 minute timeout being reasonable, but 300 minutes would mean 1 commit in a day can be verified potentially without manually going in and stopping jobs...

rolfyone (commented):
overall looks pretty good! I updated the branch so that CI would be happy, so you should get CI results now.  One minor issue with a timeout could be addressed, and then we'll wait and see how it runs on the slower CI infrastructure...

rolfyone:
if you can run `gradle spotlessJavaApply` and commit the changes, it'll cleanup the spotless check and the windows build failure (which is ultimately a spotless thing too)
Sorry importantly - I had to update on our side to get CI to run, so make sure you do a `git pull`...

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
jimmygchen: Nice catch! I got the unit wrong here.. was thinking to do 5 minutes :)  Fixed this now. Thanks @rolfyone !

jimmygchen:
> if you can run `gradle spotlessJavaApply` and commit the changes, it'll cleanup the spotless check and the windows build failure (which is ultimately a spotless thing too)
> Sorry importantly - I had to update on our side to get CI to run, so make sure you do a `git pull`...

Done, thanks @rolfyone

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
rolfyone: oh that makes sense, usually we just do 5 minutes like below :) 300 seconds would also work :)

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BellatrixMergeTransitionAcceptanceTest.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Test
+  void shouldHaveNonDefaultExecutionPayloadAndFinalizeAfterMergeTransition() {
+    tekuNode.waitForGenesis();
+    tekuNode.waitForLogMessageContaining("MERGE is completed");
+    tekuNode.waitForNewFinalization();
+
+    tekuNode.waitForNonDefaultExecutionPayload();
+    tekuNode.waitForNewFinalization();
```
ajsutton: I suspect once we see `MERGE is completed` we will have non default execution payloads and can then just wait for one new finalisation after that to confirm it's all working.  That will reduce the length of time this test runs by a fair bit since finalisation is fairly slow.

```suggestion
    tekuNode.waitForLogMessageContaining("MERGE is completed");
    tekuNode.waitForNonDefaultExecutionPayload();

    tekuNode.waitForNewFinalization();
```

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/BesuNode.java
```diff
@@ -60,7 +84,66 @@ public String getExternalJsonRpcUrl() {
...
+      configMap.put("rpc-http-api", rpcHttpApis);
+      return this;
+    }
+
+    public BesuNode.Config withDefaultEngineRpcHttpPort() {
+      configMap.put("engine-rpc-http-port", Integer.toString(ENGINE_JSON_RPC_PORT));
+      return this;
+    }
+
+    public BesuNode.Config withEngineHostAllowList(final String[] engineHostAllowList) {
```
ajsutton: nit: Using varargs here would avoid having to explicitly create an array when calling

```suggestion
    public BesuNode.Config withEngineHostAllowList(final String... engineHostAllowList) {
```

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BellatrixMergeTransitionAcceptanceTest.java
```diff
@@ -0,0 +1,82 @@
...
+        .withTotalTerminalDifficulty(UInt64.valueOf(10001).toString())
+        .withGenesisTime(genesisTime);
+  }
+
+  private BesuNode.Config configureBesuNode(BesuNode.Config config) {
+    return config
+        .withRpcHttpApi(new String[] {"ETH,NET,WEB3,ENGINE"})
+        .withDefaultEngineRpcHttpPort()
+        .withEngineHostAllowList(new String[] {"*"})
+        .withMergeSupport(true)
```
ajsutton: I wonder if we should just automatically set all of these as part of the `withMergeSupport(true)` call to keep the acceptance test classes abstracted away from the besu details a bit.

ajsutton (commented):
Looks really good. Thanks for this.  I've just suggested a couple of small tweaks to keep the AT class itself as simple as possible.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BellatrixMergeTransitionAcceptanceTest.java
jimmygchen: Good idea. I have move these to `withMergeSupport(true)`.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/BesuNode.java
jimmygchen: Nice tip, thanks :) I've removed this method as this is now done in `withMergeSupport`

ajsutton (approved):
LGTM. Thanks for this.

PR #5203, opened 2022-03-18
Migrate database command should migrate to leveldb2 by default

### LINKED ISSUE #5198: change migrate database commands default destination database
when we migrated to leveldb-tree, we have implicitly converted the migrate-database command to use that by default.

That won't work, we need to default to leveldb2 for destination.

rolfyone:
When the leveldb-tree version became default, it changed the default 'to' database type, and that won't be very successful given the storage structures.

workaround: specify `--Xto leveldb2` in the CLI, and this will override the default using the development 'to' flag.

fixes #5198

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5204, opened 2022-03-18
Migrate GetPeerCount to MigratingEndpointAdapter

courtneyeh:
## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeerCount.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeerCount.java
@@ -45,21 +45,9 @@ public class GetPeerCount extends MigratingEndpointAdapter {
 
   private static final SerializableTypeDefinition<ResponseData> PEER_COUNT_TYPE =
       SerializableTypeDefinition.object(ResponseData.class)
-          .withField(
-              "disconnected",
-              UINT64_TYPE,
-              responseData ->
-                  UInt64.valueOf(
-                      responseData.getData().stream()
-                          .filter(eth2Peer -> !eth2Peer.isConnected())
-                          .count()))
+          .withField("disconnected", UINT64_TYPE, ResponseData::getDisconnected)
           .withField("connecting", UINT64_TYPE, responseData -> UInt64.valueOf(0))
-          .withField(
-              "connected",
-              UINT64_TYPE,
-              responseData ->
-                  UInt64.valueOf(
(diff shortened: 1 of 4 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeerCount.java
```diff
@@ -63,7 +102,24 @@ public GetPeerCount(final DataProvider provider, final JsonProvider jsonProvider
...
+  static class ResponseData {
+    private final List<Eth2Peer> data;
+
+    ResponseData(List<Eth2Peer> peers) {
+      this.data = peers;
+    }
+
+    List<Eth2Peer> getData() {
+      return data;
+    }
```
rolfyone: we should maybe make use of this response object to reduce double handling...

on creation we could count the connected / not connected, store that value, and just query the value in the type...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeerCount.java
```diff
@@ -13,39 +13,78 @@
...
+  private static final SerializableTypeDefinition<ResponseData> PEER_COUNT_TYPE =
+      SerializableTypeDefinition.object(ResponseData.class)
+          .withField(
+              "disconnected",
+              UINT64_TYPE,
+              responseData ->
+                  UInt64.valueOf(
+                      responseData.getData().stream()
+                          .filter(eth2Peer -> !eth2Peer.isConnected())
+                          .count()))
```
rolfyone: if we change the responseData object this can just be a fn call :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeerCount.java
```diff
@@ -112,14 +100,31 @@ public void handleRequest(RestApiRequest request) throws JsonProcessingException
...
+      for (Eth2Peer peer : peers) {
+        if (peer.isConnected()) {
+          connected++;
+        } else {
+          disconnected++;
+        }
+      }
+
+      this.disconnected = UInt64.valueOf(disconnected);
+      this.connected = UInt64.valueOf(connected);
```
rolfyone, line 119: nit: i think these might even be able to be final, but super minor

rolfyone (approved):
LGTM

PR #5205, opened 2022-03-18
Introduce strategy to select states for attestation validation

ajsutton:
## PR Description
Previously attestations were validated using the state for the block they attested to (applying slots if required). That requires holding or generating the state for any block we receive an attestation for.

We can reduce the number of states required by being more intelligent about the state we select. Since we hold strong references to the finalised and head states, those are our preferred options. If neither of those are suitable then we use the state from the earliest slot that would have the required committees (on the same fork as the attestation's block).

In typical network conditions this means we use three states:

1. The chain head for anything attesting to a canonical block in the current epoch
2. The finalised state for anything attesting to a block 2 epochs after the finalised state
3. The justified state for pretty much everything in between. The exception being blocks that don't descend from our justified checkpoint and are in the current epoch but that's quite unusual.

In periods of non-finalisation we may need more states to cover the different forks but will be able to use at most one state per fork per epoch which is still a 32x reduction in required states.

The other advantage this has is that it causes the justified state to be used frequently and so makes it far more likely to be available in memory when the finalised checkpoint updates and the next finalised checkpoint is most likely going to be the current justified checkpoint.

## Documentation

## Changelog


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/validation/AggregateAttestationValidatorTest.java
```diff
@@ -156,20 +148,6 @@ public void shouldReturnValidForValidAggregate() {
         .isCompletedWithValue(InternalValidationResult.ACCEPT);
   }
 
-  @Test
-  public void shouldReturnValidForValidAggregate_whenManyBlocksHaveBeenSkipped() {
```
ajsutton, line 160: Removed test as it's now handled by `AttestationStateSelector`.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/validation/AggregateAttestationValidatorTest.java
```diff
@@ -219,31 +198,12 @@ public void shouldSaveForFutureWhenStateIsNotAvailable() {
...
             any(), eq(attestation), eq(OptionalInt.empty())))
-        .thenReturn(SafeFuture.completedFuture(InternalValidationResult.SAVE_FOR_FUTURE));
+        .thenReturn(SafeFuture.completedFuture(InternalValidationResultWithState.saveForFuture()));
 
     assertThat(validator.validate(attestation))
         .isCompletedWithValue(InternalValidationResult.SAVE_FOR_FUTURE);
   }
 
-  @Test
-  public void shouldRejectWhenAttestationValidatorSavesForFutureAndAggregateChecksFail() {
```
ajsutton, line 229: Removed tests as it doesn't make sense to load a state to process the aggregate when it can't be accepted now. We'll wind up doing the full validation again when we process in the future so just leave it to then.

tbenr (approved):
LGTM

PR #5206, opened 2022-03-19
Small speed up to Merkle branch validation

benjaminion:
Sorry, had to scratch this itch. It's only about 3% faster, but 100x more elegant.


tbenr (approved):
LVGTM!

PR #5208, opened 2022-03-20
`SimpleLeafNode` remove `Bytes32` objects creations and cache `hashTreeRoot`

tbenr:
SSZ optimizations in `SimpleLeafNode` by storing data in `Bytes` and cache `hashTreeRoot`.

The drawback is that, in the current form, this PR is going to use more memory

### CPU

bellatrix state serialization
OLD
SSZBenchmark.BeaconStateSerialization  thrpt   25  42454.753 ± 726.571  ops/s

NEW
SSZBenchmark.BeaconStateSerialization  thrpt   25  47264.855 ± 595.956  ops/s

-> **11% perf increase**

#### bellatrix state generation
```Java
// bellatrix dataStructureUtil
IntStream.range(0, 1000).forEach(__ -> dataStructureUtil.randomBeaconState());
```

OLD
image

NEW
generation only
image

-> **2.6% less memory pressure**

#### bellatrix state serialization
```Java
IntStream.range(0, 1000).forEach(__ -> bellatrixState.sszSerialize());
```

OLD
image

NEW
image

-> **68% less memory pressure**

## Documentation

## Changelog

### DIFF
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
@@ -22,25 +22,39 @@ import org.apache.tuweni.bytes.Bytes32;
 
 class SimpleLeafNode implements LeafNode, TreeNode {
 
-  private final byte[] data;
+  private final Bytes data;
+  private Bytes32 hashTreeRoot;
 
   public SimpleLeafNode(Bytes data) {
     checkArgument(data.size() <= MAX_BYTE_SIZE);
-    this.data = data.toArrayUnsafe();
+    if (data.size() == MAX_BYTE_SIZE) {
+      // if data is Bytes32, it will pass throw with no object creation
+      // otherwise translate types to Bytes32
+      this.hashTreeRoot = Bytes32.wrap(data.copy());
+      this.data = hashTreeRoot;
+      return;
+    }
+    // If we store data as is, some Bytes instances (ie ConcatenatedBytes) will
+    // perform worse during serialization. So we want to translate them to a
+    // single array-backed Bytes
+    //
+    // in this case, the following seems to perform better than Bytes.wrap(data.copy())
+    this.data = Bytes.wrap(data.toArrayUnsafe());
   }
 
   @Override
   public Bytes getData() {
-    return Bytes.wrap(data);
+    return data;
   }
 
   @Override
   public Bytes32 hashTreeRoot() {
-    if (data.length == MAX_BYTE_SIZE) {
-      return Bytes32.wrap(data);
-    } else {
-      return Bytes32.wrap(Arrays.copyOf(data, MAX_BYTE_SIZE));
+    if (hashTreeRoot != null) {
+      return hashTreeRoot;
     }
+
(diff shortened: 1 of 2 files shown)


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
```diff
@@ -22,25 +22,39 @@
...
 
   @Override
   public Bytes32 hashTreeRoot() {
-    if (data.length == MAX_BYTE_SIZE) {
-      return Bytes32.wrap(data);
-    } else {
-      return Bytes32.wrap(Arrays.copyOf(data, MAX_BYTE_SIZE));
+    if (hashTreeRoot != null) {
+      return hashTreeRoot;
     }
```
rolfyone, line 54: this seems like as big an improvement as the rest of it, actually storing hash and not constantly rebuilding...

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
```diff
@@ -22,25 +22,39 @@
...
+
+    hashTreeRoot = Bytes32.wrap(Arrays.copyOf(data.toArrayUnsafe(), MAX_BYTE_SIZE));
```
rolfyone, line 56: i wonder if we should just calculate this on construction anyway... then this is just a getter

rolfyone (approved):
LGTM, some minor comments

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
tbenr: Yes but I'm a bit concerned by the memory overhead but I just had another idea. Tomorrow I'll explore it

tbenr: Yes but I'm a bit concerned by the memory overhead but I just had another idea. Tomorrow I'll explore it

Nashatyrev:
> An addition could be to have a soft ref to cached hashTreeRoot to let GC release it under low mem conditions.

I'm not sure a large amount of soft-refs to small objects would make things better. However I'm not 100% sure and it probably makes sense to experiment with that

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
tbenr: yes that's the main idea. Which is a 100% win in the Bytes32 case, where hashTreeRoot and data are actually the same thing.
In case of smaller data size is where we pay the bill, lower GC pressure but higher mem consumption.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
tbenr: > i wonder if we should just calculate this on construction anyway... then this is just a getter
I'd prefer to have a lazy cache, Anton seems to agree here :)

tbenr:
merging this to see the impact in memory consumption on one prater node

PR #5210, opened 2022-03-20
Migrate GetVersion to MigratingEndpointAdapter

### LINKED ISSUE #5209: Migrate GetVersion to MigratingEndpointAdapter

courtneyeh:
fixes #5209 

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetVersion.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetVersion.java
@@ -38,13 +38,13 @@ import tech.pegasys.teku.infrastructure.version.VersionProvider;
 public class GetVersion extends MigratingEndpointAdapter {
   public static final String ROUTE = "/eth/v1/node/version";
 
-  private static final SerializableTypeDefinition<VersionData> DATA_TYPE =
-      SerializableTypeDefinition.object(VersionData.class)
-          .withField("version", STRING_TYPE, VersionData::getVersion)
+  private static final SerializableTypeDefinition<String> DATA_TYPE =
+      SerializableTypeDefinition.object(String.class)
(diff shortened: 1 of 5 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetVersion.java
```diff
@@ -13,31 +13,53 @@
...
+public class GetVersion extends MigratingEndpointAdapter {
+  public static final String ROUTE = "/eth/v1/node/version";
 
-  public GetVersion(JsonProvider jsonProvider) {
-    this.jsonProvider = jsonProvider;
-  }
+  private static final SerializableTypeDefinition<VersionData> DATA_TYPE =
+      SerializableTypeDefinition.object(VersionData.class)
+          .withField("version", STRING_TYPE, VersionData::getVersion)
+          .build();
```
rolfyone: @ajsutton this is probably a good example to ask about if we can define this serializable without creating a static object...

There's a number of similar examples of 'simple' objects that may have 1-2 fields that we're currently creating going to need to wrap with a static object in this way, and whether that's desirable...

ajsutton: Yeah so if version is a string I'd just define something like:

```
SerializableTypeDefinition<String> DATA_TYPE = SerializableTypeDefinition.object(String.class)
  .withField("version", STRING_TYPE, Function.identity())
  .build();

SerializableTypeDefinition<String> RESPONSE_TYPE = SerializableTypeDefinition.object(String.class)
  .withField("data", DATA_TYPE, Function.identity())
  .build();
```

Which takes a single string and gives you all of:
```
{
  data: { version: "my version"; }
}
```

rolfyone (approved):
LGTM

PR #5213, opened 2022-03-21
Demonstrate oneOf in json schema definitions

rolfyone:
## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
@@ -89,6 +89,16 @@ public class SchemaDefinitionsAltair extends AbstractSchemaDefinitions {
   }
 
   @Override
+  public BeaconBlockSchema getBlindedBlockSchema() {
+    return getBeaconBlockSchema();
+  }
+
+  @Override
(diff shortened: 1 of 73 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
```diff
@@ -87,6 +88,16 @@ public BeaconBlockSchema getBeaconBlockSchema() {
...
 
+  @Override
+  public BeaconBlockSchema getBlindedBlockSchema() {
+    return getBeaconBlockSchema();
+  }
+
+  @Override
+  public BeaconBlockBodySchema<?> getBlindedBlockBodySchema() {
+    return getBeaconBlockBodySchema();
+  }
```
rolfyone, line 99: @tbenr would be interested if you think this is enough?

I'm not sure it'd add anything but noise adding all the blinded interfaces, but would welcome thoughts...

tbenr: It's borderline to me. The underlying question would be: do we want to support blinded api calls in non-Bellatrix milestones? The answer seems to be yes, so for the moment I think we can leave it there

tbenr (approved):
LGTM

PR #5223, opened 2022-03-22
Metric to capture current syncing status

rolfyone:
The external metrics publishing is looking for a simple 'syncing' flag which we didn't really have a direct mapping for, so this will fill that need.

It should be noted that the console 'syncing' message seems to trigger well before we're actually in syncing mode, so there will be a timing disconnect there, and that's just because of the way console works.

It's also likely that with the granularity of publishing, you may rarely see this say 'syncing', and that's just because usually sync windows are short, and we're only reporting in 1 minute increments to the publisher.

Tested some long sync periods and the metric is updating. The graph in external service is curious when syncing falls out, but overall it's a fair representation, and the data being sent does represent our system status so that's all we can really hope for...

Added a test case to show the desired behaviour for this metric.

## Documentation

## Changelog


tbenr:
@rolfyone I'm not sure about the approach. If I had done it, I'd put the metric in `SyncStateTracker` which in Bellatrix will signal "syncing" also in the case we are in optimistic sync, which means that the EL is not in sync.
WDYT?

tbenr (approved):
LGTM

PR #5227, opened 2022-03-23
Upgrade FuzzUtil to support Bellatrix

### LINKED ISSUE #5216: Update fuzzing utilities to work with Bellatrix
Currently, `FuzzUtil` only supports Phase0 state transitions:

* `attestation`
* `attester_slashing`
* `block`
* `block_header`
* `deposit`
* `proposer_slashing`
* `voluntary_exit`

I propose we update these state transitions to work with Bellatrix & add new ones:

* `sync_aggregate`
* `execution_payload`

Rationale: I would like to update [beacon-fuzz](https://github.com/sigp/beacon-fuzz) to work with Bellatrix.

jtraglia:
## PR Description

This PR updates the fuzz utility (`FuzzUtil`) from Phase0 to Bellatrix. I thought about creating a new `FuzzUtilBellatrix` class but decided against it because I do not think we need to support old versions (Phase0 & Altair).

I am not sure how the original developers of this created all of the [pyspec](https://github.com/ethereum/consensus-specs/tree/dev/tests/core/pyspec) yaml files. I do not see any [generator](https://github.com/ethereum/consensus-specs/tree/dev/tests/generators) options for this. I will need to either figure out how to do this or delete all of the old yaml files to avoid any confusion.

Fixes #5216.

## Documentation

## Changelog


jtraglia:
So I just learned of the [zcli utility](https://github.com/protolambda/zcli) that can convert ssz files to json/yaml for inspection. With this tool, I think we should remove all of the yaml files as they are no longer produced during generation. For example:

```
$ zcli pretty bellatrix BeaconState --preset-phase0=minimal --preset-altair=minimal pre.ssz | head
{
  "genesis_time": "0",
  "genesis_validators_root": "0x5dec7ae03261fde20d5b024dfabce8bac3276c9a4908e23d50ba8c9b50b0adff",
  "slot": "0",
  "fork": {
    "previous_version": "0x01000001",
    "current_version": "0x02000001",
    "epoch": "0"
  },
  "latest_block_header": {
```

In the same spirit, since the pyspec generators now produce `.ssz_snappy` files by default, I think we should swap to that format and update the `loadSsz` method to snappy uncompress it first. This will be easier for others to reproduce. My understanding is that the generators used to produce both `.ssz` and `.yaml` files but switched at some point in the past two years, presumably to save space.

jtraglia:
For snappy compressed ssz files, you would use this command to print it:

```
$ zcli pretty bellatrix BeaconState --preset-phase0=minimal --preset-altair=minimal ssz_snappy:pre.ssz_snappy
```

ajsutton:
zcli is awesome. :). For the record we recently added the subcommand `teku debug-tools pretty-print` to support converting SSZ objects to either json or yaml. It can read `.ssz_snappy` files directly as well which I always found frustrating with zcli.

I'd agree that removing the YAML files and moving .ssz_snappy makes sense since it matches the reference tests.

fuzz/src/main/java/tech/pegasys/teku/fuzz/input/SyncAggregateFuzzInput.java
```diff
@@ -0,0 +1,61 @@
...
+import tech.pegasys.teku.spec.config.SpecConfigAltair;
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.altair.BeaconBlockBodySchemaAltair;
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.altair.SyncAggregate;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+
+public class SyncAggregateFuzzInput
+    extends Container2<SyncAggregateFuzzInput, BeaconState, SyncAggregate> {
+
+  public static ContainerSchema2<SyncAggregateFuzzInput, BeaconState, SyncAggregate> createSchema(
+      final SpecVersion spec, final SpecConfigAltair config) {
```
ajsutton: nit: Can remove the `config` parameter from here (and then probably also from the `SyncAggregateFuzzInput` constructor).

ajsutton (commented):
I can't get GitHub to comment on it but the new `sync_aggregate.ssz_snappy` file seems to have come through as a text file but needs to be marked as binary in git.  Probably should add `*.ssz_snappy binary` to `.gitattributes` and may need to remove then re-add this file to get it to apply.  Autodetection magic seems to have worked fine for the other files.

Otherwise this looks good. Thanks.

jtraglia:
Sweet. I'll make those changes. And awesome, I didn't know about `debug-tools pretty-print`!

Yup works like a charm:

```
./build/install/teku/bin/teku debug-tools pp --network minimal BELLATRIX BeaconState \
    fuzz/src/test/resources/minimal/operations/attestation/pyspec_tests/success/pre.ssz_snappy
```

fuzz/src/main/java/tech/pegasys/teku/fuzz/input/SyncAggregateFuzzInput.java
jtraglia: Good catch. I needed it originally but not anymore. It's been removed.

jtraglia:
Hmm updating `.gitattributes` and re-adding the file didn't seem to do the trick. Git (via `git check-attr`) knows that it is binary, but GitHub still shows it as text. I think it either has something to do with `eol` (end-of-line) and/or the file length; it's the smallest file of the bunch at 103 bytes. Let me try removing the `eol` attribute for binary files & re-adding the file.

```
$ git check-attr --all -- sync_aggregate.ssz_snappy
sync_aggregate.ssz_snappy: binary: set
sync_aggregate.ssz_snappy: diff: unset
sync_aggregate.ssz_snappy: merge: unset
sync_aggregate.ssz_snappy: text: unset
sync_aggregate.ssz_snappy: eol: lf

$ git check-attr --all -- pre.ssz_snappy
pre.ssz_snappy: binary: set
pre.ssz_snappy: diff: unset
pre.ssz_snappy: merge: unset
pre.ssz_snappy: text: unset
pre.ssz_snappy: eol: lf
```

jtraglia:
Nope, didn't work. I'm not sure what we can do. It might just be a problem with GitHub's file review.

ajsutton:
Cool thanks - probably is just a GitHub display issue. Must think the file is small enough to display even though it's binary.

ajsutton:
hmm, we did something bad to git attributes and broke the windowsBuild. I'll have a play with it and see if I can figure it out.

ajsutton (approved):

jtraglia:
Ah thank you for reverting that. I did not mean to make all files binary. My bad.

PR #5228, opened 2022-03-24
Start moving parameter utils to infrastructure.

rolfyone:
We don't want individual handlers to have to handle all of the complexity of dealing with query parameters, so it makes sense to use parameterUtils, but currently they're not accessible.

Also removed the `getQueryParamMap` function, as ideally we won't need to have direct access to the query or path paramter maps from the context.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5232, opened 2022-03-24
Block Import metrics via `MetricsCountersByIntervals`

### LINKED ISSUE #5188: Report block import timings

tbenr:
## PR Description

Introduces a new metric class`MetricsCountersByIntervals` which allows to specify "value intervals" per label values. 
Passing the following time "intervals" 50 100 and 250 it creates 4 counters with the following intervals as label: `[0,50)`, `[50,100)`, `[100,250)` and `[250,∞)`.
It is possible to specify different bands for different combination of label values.

Modifies `MetricsHistogram` to support multiple labels

## Fixed Issue(s)
fixes #5188

in block import produces something similar to:

```
# HELP beacon_block_import_delay_summary Histogram recording delay in milliseconds at each stage of block import
# TYPE beacon_block_import_delay_summary summary
beacon_block_import_delay_summary{stage="transaction_prepared",quantile="0.5",} 0.0
beacon_block_import_delay_summary{stage="transaction_prepared",quantile="0.95",} 0.0
beacon_block_import_delay_summary{stage="transaction_prepared",quantile="0.99",} 0.0
beacon_block_import_delay_summary{stage="transaction_prepared",quantile="1",} 0.0
beacon_block_import_delay_summary{stage="processed",quantile="0.5",} 79.0
beacon_block_import_delay_summary{stage="processed",quantile="0.95",} 143.0
beacon_block_import_delay_summary{stage="processed",quantile="0.99",} 143.0
beacon_block_import_delay_summary{stage="processed",quantile="1",} 143.0
beacon_block_import_delay_summary{stage="pre-state_retrieved",quantile="0.5",} 5.0
beacon_block_import_delay_summary{stage="pre-state_retrieved",quantile="0.95",} 11.0
beacon_block_import_delay_summary{stage="pre-state_retrieved",quantile="0.99",} 11.0
beacon_block_import_delay_summary{stage="pre-state_retrieved",quantile="1",} 4.0
beacon_block_import_delay_summary{stage="arrival",quantile="0.5",} 479.0
beacon_block_import_delay_summary{stage="arrival",quantile="0.95",} 1791.0
beacon_block_import_delay_summary{stage="arrival",quantile="0.99",} 1791.0
beacon_block_import_delay_summary{stage="arrival",quantile="1",} 1791.0
beacon_block_import_delay_summary{stage="total_processing_time",quantile="0.5",} 87.0
beacon_block_import_delay_summary{stage="total_processing_time",quantile="0.95",} 159.0
beacon_block_import_delay_summary{stage="total_processing_time",quantile="0.99",} 159.0
beacon_block_import_delay_summary{stage="total_processing_time",quantile="1",} 159.0
beacon_block_import_delay_summary{stage="completed",quantile="0.5",} 4.0
beacon_block_import_delay_summary{stage="completed",quantile="0.95",} 7.0
beacon_block_import_delay_summary{stage="completed",quantile="0.99",} 7.0
beacon_block_import_delay_summary{stage="completed",quantile="1",} 3.0
beacon_block_import_delay_summary{stage="transaction_committed",quantile="0.5",} 1.0
beacon_block_import_delay_summary{stage="transaction_committed",quantile="0.95",} 2.0
beacon_block_import_delay_summary{stage="transaction_committed",quantile="0.99",} 2.0
beacon_block_import_delay_summary{stage="transaction_committed",quantile="1",} 0.0
# HELP beacon_block_import_delay_counter Counter of blocks falling in different time frames in each import stages
# TYPE beacon_block_import_delay_counter counter
beacon_block_import_delay_counter{stage="transaction_prepared",result="success",interval="[0,50)",} 5.0
beacon_block_import_delay_counter{stage="arrival",result="success",interval="[500,1000)",} 1.0
beacon_block_import_delay_counter{stage="arrival",result="unknown_parent",interval="[0,500)",} 1.0
beacon_block_import_delay_counter{stage="completed",result="unknown_parent",interval="[0,50)",} 1.0
beacon_block_import_delay_counter{stage="total_processing_time",result="unknown_parent",interval="[0,500)",} 1.0
beacon_block_import_delay_counter{stage="pre-state_retrieved",result="success",interval="[0,50)",} 5.0
beacon_block_import_delay_counter{stage="arrival",result="success",interval="[0,500)",} 2.0
beacon_block_import_delay_counter{stage="processed",result="success",interval="[100,250)",} 1.0
beacon_block_import_delay_counter{stage="completed",result="success",interval="[0,50)",} 5.0
beacon_block_import_delay_counter{stage="arrival",result="success",interval="[1500,2000)",} 1.0
beacon_block_import_delay_counter{stage="arrival",result="success",interval="[1000,1500)",} 1.0
beacon_block_import_delay_counter{stage="processed",result="success",interval="[50,100)",} 4.0
beacon_block_import_delay_counter{stage="transaction_committed",result="success",interval="[0,50)",} 5.0
beacon_block_import_delay_counter{stage="pre-state_retrieved",result="unknown_parent",interval="[0,50)",} 1.0
beacon_block_import_delay_counter{stage="total_processing_time",result="success",interval="[0,500)",} 5.0
```

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportPerformance.java
```diff
@@ -25,54 +25,84 @@
...
+            event.left() + (lateBlockEventTimings.isEmpty() ? " " : " +") + stepDuration + "ms");
       }
-      final String combinedTimings = String.join(", ", eventTimings);
+
+      lastEvent = event;
+    }
+
+    // if(lastEvent.left().equals(IMPORT_COMPLETED_EVENT_LABEL)) {
+    //   events.get(0).right()
+    // }
```
ajsutton: nit: Remove commented out code.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportPerformance.java
```diff
@@ -25,54 +25,84 @@
...
-        final UInt64 stepDuration = event.right().minusMinZero(previousEventTimestamp);
-        eventTimings.add(
-            event.left() + (eventTimings.isEmpty() ? " " : " +") + stepDuration + "ms");
-        previousEventTimestamp = event.right();
+    final UInt64 importCompletedTimestamp = addEvent(IMPORT_COMPLETED_EVENT_LABEL);
+
+    final List<String> lateBlockEventTimings =
+        importCompletedTimestamp.isGreaterThan(timeWarningLimitTimeStamp)
+            ? new ArrayList<>()
+            : null;
```
ajsutton: I'm not really a fan of setting a `List` to `null`.  I'd probably always set this to `new ArrayList<>()` and have a separate boolean for whether we'll need to log a warning about the late block.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -60,6 +62,8 @@
   private final Subscribers<ImportedBlockListener> receivedBlockSubscribers =
       Subscribers.create(true);
 
+  private BlockImportMetrics blockImportMetrics;
```
ajsutton: Should mark this `final` and either explicitly assign `null` or make it an `Optional`. I'd be tempted to make it an `Optional` just to be completely clear even though we'll wind up using `.orElseGet()` in the one place it's used.  Or possibly we remove `blockImportPerformanceEnabled` field and use the presence or absence of this optional as the flag.

infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsHistogramWithCounters.java
```diff
@@ -0,0 +1,148 @@
...
+
+public class MetricsHistogramWithCounters extends MetricsHistogram {
+  private final List<UInt64> sortedBoundaries;
+  private final Map<UInt64, String> boundariesToIntervalLabels;
+  private final Map<String, LabelledMetric<Counter>> intervalLabelsToMetrics;
+
+  public MetricsHistogramWithCounters(
+      final Histogram histogram,
+      final List<UInt64> sortedBoundaries,
+      final Map<UInt64, String> boundariesToIntervalLabels,
```
ajsutton: This feels like it should be a `TreeMap` and then we could just use the methods in `NavigableMap` to efficiently find the key to use (e.g. `floorEntry`). Maybe I'm missing something though.

infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsHistogramWithCounters.java
```diff
@@ -0,0 +1,148 @@
...
+    final List<UInt64> sortedBoundaries =
+        boundaries.stream().sorted().collect(Collectors.toUnmodifiableList());
+    final SynchronizedHistogram histogram =
+        new SynchronizedHistogram(numberOfSignificantValueDigits);
+    final Map<UInt64, String> boundariesToIntervalLabels =
+        boundariesToIntervalLabels(sortedBoundaries);
+    final Map<String, LabelledMetric<Counter>> boundariesToMetrics =
+        createCounterMetricsForBoundaries(
+            boundariesToIntervalLabels.values(), category, counterName, counterHelp, metricsSystem);
+    return createMetric(
```
ajsutton: Somewhere in the setup of the metric it would be good to initialise all the labels to `0` explicitly. That way they immediately show up in the metrics output making it easy for users to see all the buckets that are available. Otherwise they'll only turn up when a value actually goes into that bucket.

ajsutton (commented):
I think it will be very interesting to play with this and see what is an isn't useful in practice and how to visualise it. We may need to come back and tweak these metrics a few times but this seems like a good start.

The one thing that would be nice, but looks potentially hard with the way it's set up would be to have each stage of the block process recorded as a label rather than a separate metric.  So instead of:

`block_import_arrival_delay_summary` and `block_import_prestate_retrieved_delay_summary` we'd have `block_import_delay_summary{stage=arrival}` and `block_import_delay_summary{stage=prestate_retrieved}`.  That way it's easier to build dashboards that show all the stages and we can more easily add/remove stages without constantly breaking dashboards.

It would also be useful to have a total delay metrics that just shows how long after the start of a slot the block completed importing.

tbenr:
I added `result` label to demo the new counter metrics, it may be interesting though.

tbenr:
Uff need to fix one test. Going to fix it tomorrow

ajsutton (approved):
LGTM.

PR #5234, opened 2022-03-24
Fix incorrect labeled peer address field

### LINKED ISSUE #5185: figure out the difference in standard peers interface
We did a round of checks fairly recently so this is a little surprising, but the field 'address' in data doesn't appear to be there.
So questions:
  - was it never there and we were always incorrect?
  - did it change and we didn't do the change (has standards implications, just want yes/no though)
  - are other clients all using `address` or the new field? (if they are, we fix the spec instead)

we should probably adapt to standard, and list as a `breaking change`
https://ethereum.github.io/beacon-APIs/?urls.primaryName=v1#/Node/getPeers

zilm13:
## PR Description
Rename peer field `address` to `last_seen_p2p_address` for query responses from `eth/v1/node/peers` and `eth/v1/node/peers/{peer_id}` as it should be according to the spec. (Actually, not sure whether `JsonProperty` part is anywhere used)

## Fixed Issue(s)

Fixes #5185 
## Documentation

## Changelog


zilm13:
Documentation is incorrect, the same way Teku was:
https://consensys.github.io/teku/#operation/getEthV1NodePeers
https://consensys.github.io/teku/#operation/getEthV1NodePeersWithPeer_id
I don't know how to perform update for docs

rolfyone:
- integration tests will be from ref file breakages largely (we have static json to show things changing)
 - from the spec https://ethereum.github.io/beacon-APIs/?urls.primaryName=v1#/Node/getPeers theres another 'meta' section we need to build out
 - we'll want to add this as a breaking change to the changelog, because while we're making it standard, anyone that uses current will find it doesnt work any longer...

rolfyone:
> Documentation is incorrect, the same way Teku was: https://consensys.github.io/teku/#operation/getEthV1NodePeers https://consensys.github.io/teku/#operation/getEthV1NodePeersWithPeer_id I don't know how to perform update for docs

Documentation is generated from our swagger, so it'll reflect our changes once we get to master :)

zilm13:
@rolfyone hope this time is better
The only questionable for me is whether it's ok that I made `count` in `meta` to be string. It's number by spec. General approach recommends to use strings for numbers in json but it's small and it's number in spec. I made string. Should I make it a number?

rolfyone:
Yep most places we use string for numbers, and in the standard spec you'll see they come out as type "string".

In the standard spec this time it's saying 'number' so it won't be quoted, and the easiest way to make that happen will be to use `INTEGER_TYPE` in CoreTypes, or in the reflection version just use `int`.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeers.java
```diff
@@ -88,10 +90,26 @@
...
+          .parser(UInt64::valueOf)
+          .example("1")
+          .description("Total number of items")
+          .format("uint64")
+          .build();
+
+  private static final SerializableTypeDefinition<UInt64> PEERS_META_TYPE =
+      SerializableTypeDefinition.object(UInt64.class)
+          .name("Meta")
+          .withField("count", COUNT_TYPE, Function.identity())
```
rolfyone: so `INTEGER_TYPE` will be used here instead of `COUNT_TYPE`, then `COUNT_TYPE` won't be needed (likely to need to remove a reference file when that happens)

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - The `/eth/v1/beacon/blocks/:block_id` endpoint has been deprecated in favor of the v2 Altair endpoint `/eth/v2/beacon/blocks/:block_id`
 - The `/eth/v1/validator/blocks/:slot` endpoint has been deprecated in favor of the v2 Altair endpoint `/eth/v2/validator/blocks/:slot`
 - The `/eth/v1/debug/beacon/heads` endpoint has been deprecated in favor of the v2 Bellatrix endpoint `/eth/v2/debug/beacon/heads`
+- The `/eth/v1/node/peers` endpoint had field `address` in `data` object, which is not correct according to the spec, changed to `last_seen_p2p_address`. Also `meta` object with `count` added
+- The `/eth/v1/node/peers/{peer_id}` endpoint had field `address` in `data` object, which is not correct according to the spec, changed to `last_seen_p2p_address`
```
rolfyone: This is the right content, but rather than `upcoming breaking change`, there should be a `breaking changes` section 3 sections lower.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeers.java
zilm13: I've added description to `INTEGER_TYPE`, but example field is not easy to add to our API cleanly. Could we skip it?

rolfyone: we can skip the example, and just basically use `INTEGER_TYPE.withDescription("...")` where `COUNT_TYPE` is in the withField fn call...

zilm13:
@rolfyone it's done

rolfyone (approved):
LGTM

PR #5236, opened 2022-03-25
Optimise BlockValidator

ajsutton:
## PR Description
Only retrieve the parent block slot instead of the entire block.
Only request the state at a specific slot if the parent block is from a prior epoch. This avoids unnecessarily adding to the checkpoint state cache if the state doesn't need to have any slots applied.

Combines well with #5205 to require fewer states in the caches.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5248, opened 2022-03-28
Align `SimpleLeafNode` to `SimpleBranchNode`: make cached hash `volatile`

tbenr:
## PR Description

`SimpleLeafNode` was caching the hashtree in a non-volatile variable.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
@@ -23,15 +23,15 @@ import org.apache.tuweni.bytes.Bytes32;
 class SimpleLeafNode implements LeafNode, TreeNode {
 
   private final Bytes data;
-  private Bytes32 hashTreeRoot;
+  private volatile Bytes32 cachedHash;
 
   public SimpleLeafNode(Bytes data) {
     checkArgument(data.size() <= MAX_BYTE_SIZE);
     if (data.size() == MAX_BYTE_SIZE) {
       // if data is Bytes32, it will pass throw with no object creation
       // otherwise translate types to Bytes32
-      this.hashTreeRoot = Bytes32.wrap(data.copy());
-      this.data = hashTreeRoot;
+      this.cachedHash = Bytes32.wrap(data.copy());
+      this.data = cachedHash;
       return;
     }
     // If we store data as is, some Bytes instances (ie ConcatenatedBytes) will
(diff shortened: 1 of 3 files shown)


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
```diff
@@ -49,12 +49,10 @@ public Bytes getData() {
...
   public Bytes32 hashTreeRoot() {
-    if (hashTreeRoot != null) {
-      return hashTreeRoot;
+    if (cachedHash == null) {
+      cachedHash = Bytes32.wrap(Arrays.copyOf(data.toArrayUnsafe(), MAX_BYTE_SIZE));
     }
-
-    hashTreeRoot = Bytes32.wrap(Arrays.copyOf(data.toArrayUnsafe(), MAX_BYTE_SIZE));
-    return hashTreeRoot;
+    return cachedHash;
```
ajsutton: I'm not sure if I'm over-thinking this or not but this winds up doing a volatile, read, write, read sequence which involves flushing a lot of cache lines back to memory.  I wonder if we should do:

```
Bytes32 cachedHash = this.cachedHash;
if (cachedHash == null) {
    cachedHash = Bytes32.wrap(Arrays.copyOf(data.toArrayUnsafe(), MAX_BYTE_SIZE));
    this.cachedHash = cachedHash;
}
return cachedHash;
```

That way we only do one read and if required, one write.

ajsutton (approved):
LGTM, just one suggestion which may give a very minor performance improvement.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
ajsutton: Even if this doesn't wind up as a perf improvement I think I prefer it as it ensures that each execution of the method has a consistent view of cachedHash and not assuming that what it just wrote to the volatile field hasn't changed (not that it would ever change to a bad value with this code anyway).

tbenr: yeah. Going to apply.

tbenr: @ajsutton applied to `SimpleBranchNode` too

Issue #5102, opened 2022-03-03
[MEV boost - M1] Implement Blinded block production flow via `builder` APIs


tbenr:
Details here:
https://hackmd.io/-AWMItrHSiqCjeE2l3spwQ?view#Milestone-1-spec-direct-link

Implement required changes to support `BlindedBeaconBlock` creation and submission.

this will require changes
- [x] #5260
- [x] #5261

relater to #5103

PR #5249, opened 2022-03-28
[MEV Boost] Allow Blinded block creation flow in BN - Part1

tbenr:
## PR Description

`BlockFactory` become mev_boost \ blinded block aware
`BlockOperationSelectorFactory` become mev_boost aware.

Instead of haveing a dedicated `BlindedBeaconBlockBodyBuilder`, the flow has been integrated in `BeaconBlockBodyBuilderBellatrix`. `BeaconBlockBodyBuilder` interface exposes `isBlinded` and will be used to expose to `BlockOperationSelectorFactory` to generate a builderConsumer that can reason about blinded\non-blinded blocks.

## Fixed Issue(s)
#5261

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -128,24 +135,54 @@ public BlockOperationSelectorFactory(
...
+                      SchemaDefinitionsBellatrix.required(
+                              spec.atSlot(blockSlotState.getSlot()).getSchemaDefinitions())
+                          .getExecutionPayloadSchema()
+                          .getDefault(),
+                  (payloadId) ->
+                      executionEngineChannel
+                          .getPayload(payloadId, blockSlotState.getSlot())
+                          .join()));
+
+      if (isMevBoostEnabled || blinded) {
```
ajsutton: It feels weird to me that you could ask for a non-blinded block but be given one anyway because mev boost is enabled.  Nots are it's wrong as such, just feels very unexpected.  Maybe it's just that `blinded` should be `forceBlinded`?

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -128,24 +135,54 @@ public BlockOperationSelectorFactory(
...
+                blockSlotState,
+                () ->
+                    SchemaDefinitionsBellatrix.required(
+                            spec.atSlot(blockSlotState.getSlot()).getSchemaDefinitions())
+                        .getExecutionPayloadHeaderSchema()
+                        .getDefault(),
+                (payloadId) ->
+                    executionEngineChannel
+                        .getPayloadHeader(payloadId, blockSlotState.getSlot())
+                        .join()));
```
ajsutton: Can we log a ticket to follow up and get rid of this `.join()` please? It's not a small amount of work but it would be a lot safer and more inline with our usual threading model to be able to return the `SafeFuture` here (which then means following that up the call stack quite a bit).

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryTest.java
```diff
@@ -115,7 +119,8 @@ private ExecutionPayload getExecutionPayload(final BeaconBlock block) {
     return BeaconBlockBodyBellatrix.required(block.getBody()).getExecutionPayload();
   }
 
-  private BeaconBlock assertBlockCreated(final int blockSlot, final Spec spec)
+  private BeaconBlock assertBlockCreated(
```
ajsutton, line 122: We probably should add a test for the cases where `isMevBoostEnabled` and `blinded` are true (and the various combinations).  Looks like we're always passing `false, false` at the moment.

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactoryTest.java
```diff
@@ -290,7 +299,11 @@ void shouldIncludeNonDefaultExecutionPayload() {
 
     factory
         .createSelector(
-            parentRoot, blockSlotState, dataStructureUtil.randomSignature(), Optional.empty())
+            parentRoot,
+            blockSlotState,
+            dataStructureUtil.randomSignature(),
+            Optional.empty(),
+            false)
```
ajsutton: Worth tests at this level where we need a blinded block?

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
```diff
@@ -157,12 +157,19 @@ public OkHttpValidatorRestApiClient(final HttpUrl baseEndpoint, final OkHttpClie
...
+      final BLSSignature randaoReveal,
+      final Optional<Bytes32> graffiti,
+      final boolean blinded) {
     final Map<String, String> pathParams = Map.of("slot", slot.toString());
     final Map<String, String> queryParams = new HashMap<>();
     queryParams.put("randao_reveal", encodeQueryParam(randaoReveal));
     graffiti.ifPresent(bytes32 -> queryParams.put("graffiti", encodeQueryParam(bytes32)));
 
+    if (blinded) {
+      throw new UnsupportedOperationException("blinded blocks not yet implemented");
```
ajsutton, line 170: Probably worth logging a ticket to follow up on this so it doesn't get forgotten.

ajsutton (approved):
LGTM, just small details and worth adding some tests.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
tbenr: yes, currently it's a checkbox in #5102. I'll transform them in issues.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
tbenr: Thinking again this is wrong. 
At high-level these are the conditions I want to handle. 
```java
if (isMevBoostEnabled && blinded) {
// blinded mevboost flow
// set `bodyBuilder.executionPayloadHeader` via `executionEngineChannel.getPayloadHeader`
} else if (!isMevBoostEnabled && blinded) {
// blinded non-mevboost flow (To Be implemented - issue #5103)
// set `bodyBuilder.executionPayloadHeader` via `executionEngineChannel.getPayload` (extract header from full payload)
// cache the full payload to be used to "unblind" the signed block before publication
} else {
 // non-blind standard flow
 // set `bodyBuilder.executionPayload` via `executionEngineChannel.getPayload` 
}
```

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
tbenr: tracked here: #5259

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryTest.java
tbenr: quite not there yet. I'll do a Part-2 of the PR.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
tbenr: If you don't mind I'll leave the main issue to avoid noise. I'll transform them to tickets when I start there If will make sense.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
ajsutton: ah yeah that makes sense to me.

PR #5250, opened 2022-03-29
Update GetNewBlindedBlock to use MigratingAdapter

rolfyone:
We now need the spec passed in so that we can get a definition, and we needed to make sure we had a spec for documentation to work without breaking the api when spec instances aren't defined.

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCache.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCache.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2022 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.beaconrestapi;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.SpecVersion;
+import tech.pegasys.teku.spec.schemas.SchemaDefinitions;
+
+public class SchemaDefinitionCache {
+  private final Spec spec;
+  private final Map<SpecMilestone, SchemaDefinitions> schemas = new ConcurrentHashMap<>();
+
+  public SchemaDefinitionCache(final Spec spec) {
+    this.spec = spec;
+  }
+
+  public SchemaDefinitions getSchemaDefinition(final SpecMilestone milestone) {
+    return schemas.computeIfAbsent(milestone, this::createSchemaDefinition);
+  }
+
+  public final SpecMilestone milestoneAtSlot(final UInt64 slot) {
+    return spec.atSlot(slot).getMilestone();
+  }
+
+  private SchemaDefinitions createSchemaDefinition(final SpecMilestone milestone) {
+    final SpecVersion specVersion = spec.forMilestone(milestone);
+    if (specVersion != null) {
+      return specVersion.getSchemaDefinitions();
+    }
+    return SpecVersion.create(milestone, spec.getGenesisSpecConfig())
+        .orElseThrow()
+        .getSchemaDefinitions();
+  }
+}
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
@@ -49,6 +49,7 @@ import tech.pegasys.teku.infrastructure.json.types.CoreTypes;
 import tech.pegasys.teku.infrastructure.json.types.DeserializableTypeDefinition;
 import tech.pegasys.teku.infrastructure.json.types.SerializableOneOfTypeDefinitionBuilder;
 import tech.pegasys.teku.infrastructure.json.types.SerializableTypeDefinition;
+import tech.pegasys.teku.infrastructure.json.types.StringValueTypeDefinition;
 import tech.pegasys.teku.infrastructure.restapi.endpoints.EndpointMetadata;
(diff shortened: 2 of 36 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -111,6 +111,14 @@ public SpecVersion forMilestone(final SpecMilestone milestone) {
     return specVersions.get(milestone);
   }
 
+  public SpecVersion forMilestoneOrDefault(final SpecMilestone milestone) {
```
rolfyone: @ajsutton this is the kind of logic i need for the api, not sure if this is the best spot... we'd ideally never use it unless its in api, maybe it should live in the Migrating class?

ajsutton:
Need to update some method calls to deal with Enrico's renames. :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCache.java
```diff
@@ -0,0 +1,49 @@
...
+    return spec.atSlot(slot).getMilestone();
+  }
+
+  private SchemaDefinitions createSchemaDefinition(final SpecMilestone milestone) {
+    final SpecVersion specVersion = spec.forMilestone(milestone);
+    if (specVersion != null) {
+      return specVersion.getSchemaDefinitions();
+    }
+    return SpecVersion.create(milestone, spec.getGenesisSpecConfig())
+        .orElseThrow()
```
ajsutton, line 46: hmm, this will fail if we have a config with no `BELLATRIX_FORK_EPOCH` value because `SpecConfig.toVersionBellatrx()` will return empty. Same could happen with `ALTAIR`...
In that case we actually don't have enough info to create a version for the requested milestone which makes things pretty awkward.  I think that's fine given that this is used in the beacon node so it isn't loading the config from the REST API and so will be using our presets, all of which define bellatrix options.

So we're effectively making Altair and bellatrix config options required for the beacon node with this change. Makes me a little nervous but I think that's ok.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
```diff
@@ -26,29 +27,88 @@
...
 
-  public GetNewBlindedBlock(final DataProvider dataProvider, final JsonProvider jsonProvider) {
-    super(dataProvider, jsonProvider);
+  private static final ParameterMetadata<UInt64> PARAM_SLOT =
+      new ParameterMetadata<>(
+          SLOT,
+          CoreTypes.UINT64_TYPE.withDescription(
+              "The slot for which the block should be proposed."));
+
+  private static final StringValueTypeDefinition<BLSSignature> SIGNATURE_TYPE =
```
ajsutton: nit: Is it worth splitting this out into an `EthereumTypes` class now even if it's only in the beacon rest API package for now just to avoid having it pop up in a lot of different APIs as we migrate?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
```diff
@@ -26,29 +27,88 @@
...
+          SIGNATURE_TYPE.withDescription(
+              "`BLSSignature Hex` BLS12-381 signature for the current epoch."));
+
+  private static final ParameterMetadata<Bytes32> PARAM_GRAFFITI =
+      new ParameterMetadata<>(
+          GRAFFITI, CoreTypes.BYTES32_TYPE.withDescription("`Bytes32 Hex` Graffiti."));
+  private static final DeserializableTypeDefinition<SpecMilestone> SPEC_VERSION =
+      DeserializableTypeDefinition.enumOf(SpecMilestone.class);
+
+  private final SerializableTypeDefinition<BeaconBlock> BLINDED_BLOCK_RESPONSE_TYPE;
```
ajsutton: Do we actually need this? We should be able to just call `request.respondOk` and it will get the type definition for us.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCacheTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.jupiter.api.Test;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.TestSpecFactory;
+
+public class SchemaDefinitionCacheTest {
+
+  @Test
+  void shouldGetSchemasForAllMilestones() {
```
ajsutton: Probably should write this as a parameterised test that automatically uses all the `SpecMilestone` enum values.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCacheTest.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Test
+  void shouldGetSpecMilestoneFromSpecObject() {
+    final Spec spec = mock(Spec.class);
+    when(spec.atSlot(any()))
+        .thenReturn(TestSpecFactory.createMinimalAltair().forMilestone(SpecMilestone.ALTAIR));
+    final SchemaDefinitionCache cache = new SchemaDefinitionCache(spec);
+    assertThat(cache.milestoneAtSlot(UInt64.ONE)).isEqualTo(SpecMilestone.ALTAIR);
+    verify(spec).atSlot(any());
```
ajsutton: Could use a real `Spec` here and make the assert use `isSameInstance` to confirm we used the existing one and didn't just create something equal.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
rolfyone: we could, just currently i didnt really have a reason to, even though it does seem obvious it'll happen...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
rolfyone: ah ok - hadn't done that bit so just assumed i'd need the type available, can remove

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCache.java
rolfyone: really? tests dont seem to break without it - all the normal minimals were phase 0

ajsutton (approved):
LGTM.

PR #5253, opened 2022-03-29
remove memoize hashRoot from `SszSuperNode` and `LazyBranchNode`

### LINKED ISSUE #5246: Remove usage of Suppliers.memoize from SszSuperNode
### Description
The usage of `Suppliers.memoize` results in creating a significant amount of garbage allocations.  We should move to using a volatile field similar to how `SimpleBranchNode` works.

tbenr:
## PR Description
Substitutes the memoization of `hashTreeRoot` with a simple `volatile` variable.

Profiling by applying state transition on mainnet state at 3378209, to 3378508 (about 300 blocks)

CPU benefits:

image

Memory benefits:

image

fixes #5246

## Documentation

## Changelog

### DIFF
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/LazyBranchNode.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/LazyBranchNode.java
@@ -20,7 +20,7 @@ import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.infrastructure.crypto.Hash;
 
 public class LazyBranchNode implements BranchNode {
-  private final Supplier<Bytes32> hashTreeRoot;
+  private volatile Bytes32 cachedHash;
   private final Bytes32 leftRoot;
   private final Bytes32 rightRoot;
 
@@ -33,7 +33,7 @@ public class LazyBranchNode implements BranchNode {
       final Bytes32 rightRoot,
       final Supplier<TreeNode> leftLoader,
       final Supplier<TreeNode> rightLoader) {
-    this.hashTreeRoot = () -> hashTreeRoot;
+    this.cachedHash = hashTreeRoot;
     this.leftRoot = leftRoot;
     this.rightRoot = rightRoot;
 
@@ -46,7 +46,6 @@ public class LazyBranchNode implements BranchNode {
       final Bytes32 rightRoot,
       final Supplier<TreeNode> leftLoader,
       final Supplier<TreeNode> rightLoader) {
-    this.hashTreeRoot = Suppliers.memoize(() -> Hash.sha256(leftRoot, rightRoot));
     this.leftRoot = leftRoot;
     this.rightRoot = rightRoot;
     this.leftLoader = leftLoader;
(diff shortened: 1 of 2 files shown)


Nashatyrev (approved):
LGTM

Just curious if removing `volatile` could further improve performance a bit more?

Nashatyrev:
I mean, removing `volatile` from one side increases possibility that hash would be double calculated on another thread, but from the other side it reduces synchronization overhead.

tbenr:
I think the main reason of the `volatile` is to avoid threads caching "old" hashes when they have been updated by other threads. @ajsutton may be more verbose on the topic.

ajsutton:
I don't think the hashes ever change (the node is replaced if either of its children are changed) but the volatile is required to ensure we have a consistent view of the `Bytes32` object. Compilers and CPUs are capable of and do reorder instructions and are free to do so as long as they respect the memory model rules. In this case, if `hashTreeRoot` wasn't `volatile` the instructions could be reordered such that the allocation to `hashTreeRoot` is done before the `Bytes32` constructor had actually completed (the `hashTreeRoot` variable is actually just a pointer to the allocated memory).  So we might see `hashTreeRoot` as not null but then when we use it the actual `bytes` field inside the `ArrayWrappingBytes32` could be `null` or only partially populated or the offset and length fields may not have been set yet etc.  The same effect could also result from memory visibility if those variables spanned multiple cache lines - we might get an up to date view of parts of them but pre-initialisation view of others.

Nashatyrev:
Oh, right. Good point Adrian 👍 
Though `ArrayWrappingBytes` fields are all final and this specific case still looks safe without `volatile` it's better to be safe in case if implementation changes. 

BTW do you know if the situation you've described could be ever reproduced on x86 arch? I believe I've read that this could be reproduced on RISC arches only (which of course is also targeted by us)

ajsutton:
hmm, good point on the final fields actually.  Given we're assuming we have an immutable bytes I suspect all the implementations would have that property which may make it safe to not use volatile here.  Would be interesting to know what performance difference it makes with and without.

And I think you're right that these weird cases are often risc specific - I'm not sure how they apply to ARM chipsets.  There was a really good article going into all the details from a while back that I'll have to try and dig up again.

ajsutton:
It's definitely deep into the exotic java memory model kind of territory. :)

ajsutton:
I think https://shipilev.net/blog/2014/safe-public-construction/ is the article I remember.

tbenr:
@ajsutton @Nashatyrev 

changes applied to:
SimpleLeafNode SszSuperNode SimpleBranchNode  LazyBranchNode

Without volatile: `private Bytes32 cachedHash;`
SSZBenchmark.BeaconStateSerialization  thrpt   25  46247.386 ± 952.119  ops/s
SSZBenchmark.BeaconStateSerialization  thrpt   25  47031.849 ± 643.049  ops/s
SSZBenchmark.BeaconStateSerialization  thrpt   25  47086.389 ± 635.026  ops/s

With Volatile: `private volatile Bytes32 cachedHash;`
SSZBenchmark.BeaconStateSerialization  thrpt   25  46989.661 ± 733.422  ops/s
SSZBenchmark.BeaconStateSerialization  thrpt   25  46507.963 ± 986.361  ops/s
SSZBenchmark.BeaconStateSerialization  thrpt   25  46931.979 ± 782.850  ops/s

no measurable differences in my setup

Nashatyrev:
@tbenr thanks for benchmarks!
Looks like the `volatile` makes sense here just to be safe

tbenr:
I run also `BeaconStateBenchmark.updateBalancesAndHash` which I think is the right one to test that, but no measurable differences there too.

PR #5256, opened 2022-03-30
Fix a lot of naming convention violations

ajsutton:
## PR Description
Preparing for the updated errorprone rules to come in so fixing problems.  Note we suppress the errors for beacon rest API definition classes.

Also fixes a few cases which violate the new `MathTargetType` check (https://github.com/ConsenSys/errorprone-checks/pull/7).

## Documentation

## Changelog


ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/AbstractBeaconStateTest.java
```diff
@@ -71,16 +71,18 @@ public void simpleMutableBeaconStateTest() {
...
+    final Object stateCopy = state.getSchema().createFromBackingNode(state.getBackingNode());
 
     assertThat(state == stateCopy).isFalse();
     assertThat(state).isEqualTo(stateCopy);
     assertThat(state.hashCode()).isEqualTo(stateCopy.hashCode());
   }
 
   @SuppressWarnings("unchecked")
+  @Test
   public void equals_shouldReturnFalse() {
```
jtraglia, line 86: Nice catch. I guess these weren't being tested before.

ajsutton: Yeah I was pleased to catch that. They aren't particularly important tests but still nice to fix.

rolfyone (approved):
LGTM

storage/src/testFixtures/java/tech/pegasys/teku/storage/client/ChainUpdater.java
```diff
@@ -152,7 +152,7 @@ public SignedBlockAndState finalizeCurrentChain() {
       saveBlock(newChainHead);
       updateBestBlock(newChainHead);
     }
-    final Checkpoint finalizedCheckpoint = newChainHead.getState().getFinalized_checkpoint();
+    final Checkpoint finalizedCheckpoint = newChainHead.getState().getFinalizedCheckpoint();
```
jtraglia, line 155: GitHub code scanning correctly pointed out that `newChainHead` _could_ be null. But this is just a test fixture and those values are hardcoded, so it doesn't matter.

ajsutton: Fortunately it has "Used in tests" and "Won't fix" options to make it be quiet. :)

PR #5257, opened 2022-03-30
Fix/suppress the MathTargetType findings

jtraglia:
## PR Description

These were identified using the [MathTargetType](https://github.com/ConsenSys/errorprone-checks/blob/main/src/main/java/tech/pegasys/tools/epchecks/MathTargetType.java) check in the Consensys errorprone-checks repository.

Pretty minimal changes. Thankfully there were only a few findings.

I considered changing `committeeCount` vars/funcs to `int` types but it would have been outside the scope of this PR.
## Documentation

## Changelog


jtraglia:
This is a duplicate of #5256. Didn't realize you fixed those in there. Thanks!

PR #5265, opened 2022-03-31
Define how a 'oneOf' may be selected at the metadata level

rolfyone:
For deserialization, it is necessary to allow a custom selector to define which deserializer to use based on the input data.

This is because sometimes we don't have a type selector outside of the json itself.

A practical example is postBlindedBlock, where the handler needs to determine what kind of block is being passed in based on the slot of the block provided, but other objects may require selection based on other data within the json.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5267, opened 2022-03-31
Move reading the block body into metadata and add some tests

rolfyone:

## Documentation

## Changelog

### DIFF
--- a/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
+++ b/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
@@ -24,6 +24,7 @@ import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_UNAUTHORI
 import static tech.pegasys.teku.infrastructure.json.JsonUtil.JSON_CONTENT_TYPE;
 
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import io.javalin.http.HandlerType;
 import java.io.IOException;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -103,6 +106,31 @@ public static EndpointMetaDataBuilder delete(final String path) {
...
+        (DeserializableTypeDefinition<T>) getRequestBodyType(body);
+
+    if (bodySchema == null) {
+      throw new MissingRequestBodyException();
+    }
+
+    if (!getRequestBodyType().getSelfAndReferencedTypeDefinitions().contains(bodySchema)) {
+      throw new IllegalStateException(
+          "schema determined for parsing request body is not listed in requestBodyTypes");
+    }
```
rolfyone: this is not 100% accurate, not sure if there's a better way of encapsulating oneOf AND normal types?
I can forsee this would allow us to send back UInt64 because its included somewhere deep...

ajsutton: Yeah definitely not what we want but right idea.  I think we need a new method on `OpenApiTypeDefinition` which is a bit hard to name but something like `isEquivalentDeserializableType(DeserializableTypeDefinition<?> type)`.  `OpenApiTypeDefinition` would have a default implementation which just checks that `type == this` (ie we're using the same type so it must be equivalent).

Then `SerializableOneOfTypeDefinition` would override that and for each of it's options call `option.isEquivalentDeserializableType(type)` and if any of them match return true.

I think in this case I'd actually be keen to use directly equality comparison as well - it's very difficult to say if two different instances are actual equivalent serialisations and we want to avoid creating new instances when we don't have to so I like the idea of having to use the same instance to be considered equivalent.  It will probably mean we need to suppress error prone warnings about using `==` and will need a comment explaining it but I think it's the right thing in this particular case.

ajsutton (commented):
Looks good, just need to tweak the "is it a suitable type" check.

ajsutton (approved):
LGTM.

PR #5268, opened 2022-03-31
SSZ improvements in `MessageDigest` handling

tbenr:
This pr improves `MessageDigest` handling in 2 ways
- by reusing it when possible (which avoids several objects creations) and passing around between nodes.
- leveraging "delegation" feature above, in `SimpleLeafNode`, to avoid allocating new trailing zeros. Receiving the MessageDigest from the branch node we can update it in 2 steps: adding Data and static zeros borrowed from `ZERO_LEAVES` to reach the needed 32 bytes.

New version:
Reuse `MessageDigest` almost everywhere

## BeaconStateBenchmark

Test case I used to compare profiles (from `BeaconStateBenchmark`):

```java
private static final BLSPublicKey pubkey = BLSTestUtil.randomPublicKey(0);
private static final DataStructureUtil dataStructureUtil =
    new DataStructureUtil(0).withPubKeyGenerator(() -> pubkey);
private static final BeaconState beaconState = dataStructureUtil.randomBeaconState(32 * 1024);

public static void main(String[] args) {
  IntStream.range(0, 2000)
      .forEach(
          __ -> {
            BeaconState stateW =
                beaconState.updated(
                    state -> {
                      int size = state.getBalances().size();
                      UInt64 balance = UInt64.valueOf(777);
                      for (int i = 0; i < size; i++) {
                        state.getBalances().setElement(i, balance);
                      }
                    });
            stateW.hashTreeRoot();
          });
}
```
### CPU usage reduction

initial version: 4%
image

-->> new version: 7.5%
image

###  Memory allocation reduction

initial version: 46%
image

-->> new version: 85%
image

due to:

initial version:
image

-->> new version: (not even detected)

image

initial version:
image

-->> new version: 
image

-->> new version (more pressure on Bytes32 as expected)
image

## Mainnet state transition
but in a real world scenario, profiling by applying state transition on mainnet state at 3378209, to 3378508 (about 300 blocks)

### CPU usage increase

initial version: -1.6% (worse)
image

-->> new version: 0.5%
image

###  Memory allocation reduction

initial version: 12%
image

-->> new version: 23%
image

## Documentation

## Changelog

### DIFF
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/BranchNode.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/BranchNode.java
@@ -16,10 +16,10 @@ package tech.pegasys.teku.infrastructure.ssz.tree;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.security.MessageDigest;
 import java.util.function.Function;
 import org.apache.tuweni.bytes.Bytes32;
 import org.jetbrains.annotations.NotNull;
-import tech.pegasys.teku.infrastructure.crypto.Hash;
 import tech.pegasys.teku.infrastructure.ssz.tree.GIndexUtil.NodeRelation;
 
 /**
@@ -62,7 +62,19 @@ public interface BranchNode extends TreeNode {
 
   @Override
   default Bytes32 hashTreeRoot() {
-    return Hash.sha256(left().hashTreeRoot(), right().hashTreeRoot());
+    final MessageDigest messageDigest = left().hashTreeRootDigest();
+    right().updateDigest(messageDigest);
+    return Bytes32.wrap(messageDigest.digest());
+  }
+
+  @Override
+  default MessageDigest hashTreeRootDigest() {
+    final MessageDigest messageDigest = left().hashTreeRootDigest();
+    right().updateDigest(messageDigest);
+    final Bytes32 hashTreeRoot = Bytes32.wrap(messageDigest.digest());
+    messageDigest.reset();
+    hashTreeRoot.update(messageDigest);
+    return messageDigest;
   }
 
   @NotNull
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/BranchNode.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/BranchNode.java
@@ -63,14 +63,14 @@ public interface BranchNode extends TreeNode {
   @Override
   default Bytes32 hashTreeRoot() {
     final MessageDigest messageDigest = left().hashTreeRootDigest();
-    right().updateDigest(messageDigest);
+    right().updateDigestWithHashTreeRoot(messageDigest);
     return Bytes32.wrap(messageDigest.digest());
   }
 
   @Override
   default MessageDigest hashTreeRootDigest() {
     final MessageDigest messageDigest = left().hashTreeRootDigest();
-    right().updateDigest(messageDigest);
+    right().updateDigestWithHashTreeRoot(messageDigest);
     final Bytes32 hashTreeRoot = Bytes32.wrap(messageDigest.digest());
     messageDigest.reset();
     hashTreeRoot.update(messageDigest);
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/BranchNode.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/BranchNode.java
@@ -61,20 +61,12 @@ public interface BranchNode extends TreeNode {
   BranchNode rebind(boolean left, TreeNode newNode);
 
   @Override
-  default Bytes32 hashTreeRoot() {
-    final MessageDigest messageDigest = left().hashTreeRootDigest();
-    right().updateDigestWithHashTreeRoot(messageDigest);
+  default Bytes32 hashTreeRoot(MessageDigest messageDigest) {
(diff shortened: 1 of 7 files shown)


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/TreeNode.java
```diff
@@ -35,6 +37,18 @@
    */
   Bytes32 hashTreeRoot();
 
+  /** Provide a MessageDigest and updates it with the node `hash_tree_root` */
+  default MessageDigest hashTreeRootDigest() {
+    final MessageDigest digest = MessageDigestFactory.createSha256();
+    hashTreeRoot().update(digest);
+    return digest;
+  }
```
ajsutton: This method feels pretty awkward. I think there might be a trade off here in terms of memory allocations - with this we're saving on the leaf allocations but still creating a lot of `MessageDigest` instances.  The other approach I was thinking of was to just have a single `MessageDigest` that gets reused.  So we'd have:

```java
// TreeNode interface:
default Bytes32 hashTreeRoot() {
  final MessageDigest digest = MessageDigestFactory.createSha256();
  return hashTreeRoot(digest);
}

/** Use digest to calculate and return the hash tree root. digest must be in the reset state prior to calling and
  * left in the reset state upon return.
  */
Bytes32 hashTreeRoot(final MessageDigest digest);


// SimpleBranch
public Bytes32 hashTreeRoot(final MessageDigest digest) {
  final Bytes32 leftRoot = left.hashTreeRoot(digest);
  final Bytes32 rightRoot = right.hashTreeRoot(digest);
  leftRoot.update(digest);
  rightRoot.update(digest);
  final Bytes32 root = digest.digest();
  this.cachedRoot = root;
  return root;
}

// SimpleLeaf would be something like:
public Bytes32 hashTreeRoot(final MessageDigest digest) {
     Bytes32 cachedHash = this.cachedHash;
     if (cachedHash != null) {
       return cachedHash;
       return;
     }
     
     if (data.size() < MAX_BYTE_SIZE) {
       return Bytes32.wrap(Bytes.wrap(
           data,
           ZERO_LEAVES[MAX_BYTE_SIZE - data.size()].getData());
     } else {
        return data;
     }
}
```

So leafs would have to create two wrapper classes (we could reduce that to one by adding a ConcatenatingBytes32 class) but wouldn't have to hold any additional memory.

This is definitely a "neater" approach to my mind but I'm not actually sure what the cpu and memory performance difference would be.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/BranchNode.java
```diff
@@ -62,7 +62,19 @@ static BranchNode create(TreeNode left, TreeNode right) {
...
+    right().updateDigestWithHashTreeRoot(messageDigest);
+    return Bytes32.wrap(messageDigest.digest());
+  }
+
+  @Override
+  default MessageDigest hashTreeRootDigest() {
+    final MessageDigest messageDigest = left().hashTreeRootDigest();
+    right().updateDigestWithHashTreeRoot(messageDigest);
+    final Bytes32 hashTreeRoot = Bytes32.wrap(messageDigest.digest());
+    messageDigest.reset();
```
ajsutton: The `digest()` docs say that the message digest is reset after `digest()` is called so we shouldn't need to explicitly reset.

ajsutton (approved):
Approving this as it is better than the current state but have suggested an alternative approach with its own trade offs that may be better.  Feel free to push it back on me to actually develop and do the benchmarking if you'd prefer. :)

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
```diff
@@ -57,6 +58,19 @@ public Bytes32 hashTreeRoot() {
...
+  public void updateDigestWithHashTreeRoot(MessageDigest messageDigest) {
+    Bytes32 cachedHash = this.cachedHash;
+    if (cachedHash != null) {
+      cachedHash.update(messageDigest);
+      return;
+    }
+    data.update(messageDigest);
+    if (data.size() < MAX_BYTE_SIZE) {
+      ZERO_LEAVES[MAX_BYTE_SIZE - data.size()].getData().update(messageDigest);
+    }
```
Nashatyrev: Do you think we still actually need `cachedHash` with this update? 
I believe the `SimpleLeafNode.hashTreeRoot()` is only called by the parent branch node `hashTreeRoot()`. 
Could it be the case that `cachedHash` doesn't add perf improvements anymore while still occupies memory?

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/TreeNode.java
tbenr: yeah, seems that saving additional `MessageDigest` creation is definitely better than saving `Bytes32` and the code is cleaner

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/SimpleLeafNode.java
tbenr: yes forgot to update this method accordingly, but Adrian's approach is overall better so the final version will be different again :)

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/TreeNode.java
tbenr: in the new version i use cachedHash as an alias to Data when sizes match (32). Otherwise I allocate without caching.
I suspect doesn't care that much on leafs.

ajsutton (approved):
LGTM.

PR #5269, opened 2022-03-31
Enable deprecation check & fix/suppress warnings

jtraglia:
## PR Description

Doing some spring cleaning. In this PR I enable deprecation checks (`-Xlint:deprecation`) and either fix the warnings or suppress them via a `SuppressWarnings` annotations. Enabling this deprecation check could potentially prevent some bugs.

Most of these are passing a spec to `DataStructureUtil`.

### Some Notes

In some of the asserts, I needed to add `ignoringFields` with some field names. I only added the ones that were different (ie caused the test to fail). If any of these look like they should match, maybe we will need to alter the test.

Some resources to explain the recommended changes I made:

* [hasFailed](https://www.javadoc.io/static/org.assertj/assertj-core/3.18.1/org/assertj/core/api/AbstractCompletableFutureAssert.html#hasFailed())
* [isEqualToComparingFieldByField](https://www.javadoc.io/static/org.assertj/assertj-core/3.18.1/org/assertj/core/api/AbstractObjectAssert.html#isEqualToComparingFieldByField(java.lang.Object))
* [isEqualToIgnoringGivenFields](https://www.javadoc.io/static/org.assertj/assertj-core/3.18.1/org/assertj/core/api/AbstractObjectAssert.html#isEqualToIgnoringGivenFields(java.lang.Object,java.lang.String...))

---

It's not that straight-forward to replace `BeaconChainUtil` with `ChainBuilder`.

I tried to replace `networkConfig.getConstants()` with `networkConfig.getSpec().getGenesisConfig().getRawConfig().get(field)` but I was having some problems. For example, sometimes the configuration wouldn't have a network name field; `rawConfig.get("NETWORK_NAME")` would be null.

---

I would like to replace this deprecated `com.google.common.io.Files#createTempDir` call with `java.nio.file.Files#createTempDirectory` but I would need to handle the `IOException` somehow. Not sure how to propagate that without changing a bunch of stuff; I tried but it got complicated quickly.

https://github.com/ConsenSys/teku/blob/516444f5b43ff85530d32e7a3206acc369a8b8fb/teku/src/test/java/tech/pegasys/teku/cli/subcommand/internal/validator/tools/KeyGeneratorTest.java#L69

* https://github.com/trinodb/trino/issues/6908

---

Here we are overloading a deprecated method. This seems sensitive, so I'm hesitant to attempt to fix it. The [API](https://javadoc.io/static/org.web3j/core/4.5.8/org/web3j/tx/gas/ContractGasProvider.html) wants us to use the method that takes a `contractFunc` string.

https://github.com/ConsenSys/teku/blob/335a1b7dafb0f65f07e234c94489089b5f36cf61/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/tools/deposits/DepositTransactionSender.java#L114-L120

Same with `getGasLimit` which is hardcoded.

## Documentation

## Changelog


jtraglia:
Also, I didn't know that it was this easy to find deprecated methods. This makes the `DoNotUseDeprecatedFastutilMethod` check that we have kind of pointless. I would like to delete it if we enable deprecation checks here.

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/AbstractDataBackedRestAPIIntegrationTest.java
```diff
@@ -300,7 +301,7 @@ protected Response getResponse(final String route, Map<String, String> getParams
   }
 
   protected Response post(final String route, final String postData) throws IOException {
-    final RequestBody body = RequestBody.create(JSON, postData);
+    final RequestBody body = RequestBody.create(postData, JSON);
```
jtraglia, line 304: Seems that they swapped the parameters. Got this from: https://stackoverflow.com/a/57495696

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchDataRequesterTest.java
```diff
@@ -29,13 +29,15 @@
...
+import tech.pegasys.teku.spec.TestSpecFactory;
 import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
 import tech.pegasys.teku.spec.util.DataStructureUtil;
 
 class BatchDataRequesterTest {
   private static final UInt64 BATCH_SIZE = UInt64.valueOf(50);
   private static final int MAX_PENDING_BATCHES = 5;
-  private final DataStructureUtil dataStructureUtil = new DataStructureUtil();
+  private final DataStructureUtil dataStructureUtil =
+      new DataStructureUtil(TestSpecFactory.createMinimalPhase0());
```
ajsutton, line 40: We probably should use `TestSpecFactory.createDefault()` for these since they generally don't care what the spec actually is.

ethereum/forkchoice/src/main/java/tech/pegasys/teku/ethereum/forkchoice/ProtoArrayIndices.java
```diff
@@ -31,10 +31,12 @@ public void add(final Bytes32 blockRoot, final int nodeIndex) {
...
   }
 
+  @SuppressWarnings("deprecation")
   public Optional<Integer> get(final Bytes32 root) {
     return Optional.ofNullable(rootIndices.getOrDefault(root, null));
   }
 
+  @SuppressWarnings("deprecation")
   public void remove(final Bytes32 root) {
     rootIndices.remove(root);
```
ajsutton, line 41: This one can just switch to using:
```suggestion
  public void remove(final Bytes32 root) {
    rootIndices.removeInt(root);
```

storage/src/test/java/tech/pegasys/teku/storage/client/RecentChainDataTest.java
```diff
@@ -798,7 +798,9 @@ public void commit_pruneParallelExistingBlocks() {
   @Test
   public void getAncestorsOnFork() {
     initPreGenesis();
-    final ChainBuilder chainBuilder = ChainBuilder.create(BLSKeyGenerator.generateKeyPairs(16));
+    final ChainBuilder chainBuilder =
+        ChainBuilder.create(
+            TestSpecFactory.createMinimalPhase0(), BLSKeyGenerator.generateKeyPairs(16));
```
ajsutton, line 803: Can use the existing `spec` field:
```suggestion
            spec, BLSKeyGenerator.generateKeyPairs(16));
```

storage/src/test/java/tech/pegasys/teku/storage/client/RecentChainDataTest.java
```diff
@@ -839,7 +841,9 @@ public void getAncestorsOnFork() {
   @Test
   public void getAncestorsOnFork_unknownRoot() {
     initPreGenesis();
-    final ChainBuilder chainBuilder = ChainBuilder.create(BLSKeyGenerator.generateKeyPairs(16));
+    final ChainBuilder chainBuilder =
+        ChainBuilder.create(
+            TestSpecFactory.createMinimalPhase0(), BLSKeyGenerator.generateKeyPairs(16));
```
ajsutton, line 846:
```suggestion
            spec, BLSKeyGenerator.generateKeyPairs(16));
```

ajsutton (commented):
There's a fair bit going on in this PR which is probably worth splitting into separate issues.

Specifying the `spec` for things like `DataStructureUtils` is definitely good. We've been gradually chipping away at it but it's at the point where one good push like this to fix it all is worth doing.

The changes from `isEqualComparingFieldByField` that wind up having to ignore some fields probably need some careful individual scrutiny.  For some of them I suspect it indicates either a test that isn't covering what it should or a test that's just being a bit too stupid in how it's checking results.  There will likely be some cases we actually should ignore some fields but I'm keen to minimise them.

I'm not a fan of suppressing the `deprecation` warnings and making them errors in the build. The suppression means the deprecation is no longer hilighted in the IDE so it will then tend to just be ignored forever rather than potentially fixed up when we're next working in that area (or at least serve as a warning that we shouldn't add more stuff using the current approach).

We also commonly perform migrations in smaller steps by deprecating the old method, flagging that new code shouldn't use it, and then we can gradually work through the cases rather than change everything all at once. That's allowed us to do some quite powerful refactoring that would either never get prioritised or be too dangerous if they had to happen in one giant PR.

jtraglia:
> There's a fair bit going on in this PR which is probably worth splitting into separate issues.

Good idea. I agree that this is a lot. I will split it up.

> Specifying the `spec` for things like `DataStructureUtils` is definitely good. We've been gradually chipping away at it but it's at the point where one good push like this to fix it all is worth doing.

Okay cool. I'll do that. And I'll use `TestSpecFactory.createDefault()` like you suggested.

> The changes from `isEqualComparingFieldByField` that wind up having to ignore some fields probably need some careful individual scrutiny. For some of them I suspect it indicates either a test that isn't covering what it should or a test that's just being a bit too stupid in how it's checking results. There will likely be some cases we actually should ignore some fields but I'm keen to minimise them.

Yeah this concerned me some. I'll split it up into another PR too. Starting with the ones that do not require ignoring fields. Maybe I'll make an issue for the other ones to start a discussion.

> I'm not a fan of suppressing the `deprecation` warnings and making them errors in the build. The suppression means the deprecation is no longer hilighted in the IDE so it will then tend to just be ignored forever rather than potentially fixed up when we're next working in that area (or at least serve as a warning that we shouldn't add more stuff using the current approach).

Ah I don't use an IDE, so I haven't seen those hilighted warnings. That makes sense though. I only see:

```
Note: <file>.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
```

When I redo this piecewise, I won't treat those warnings as errors.

> We also commonly perform migrations in smaller steps by deprecating the old method, flagging that new code shouldn't use it, and then we can gradually work through the cases rather than change everything all at once. That's allowed us to do some quite powerful refactoring that would either never get prioritised or be too dangerous if they had to happen in one giant PR.

I understand. That's smart. Bite-sized migrations.

PR #5273, opened 2022-04-01
[Mev Boost] Allow Blinded block creation flow in BN - Part2

tbenr:
continuation of #5249

Updates `BlockProcessorBellatrix` to allow processing of blinded blocks. 

related to #5261

## Documentation

## Changelog


eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/common/operations/DefaultOperationProcessor.java
```diff
@@ -103,10 +105,11 @@ public void processSyncCommittee(final MutableBeaconState state, final SyncAggre
   @Override
   public void processExecutionPayload(
       final MutableBeaconState state,
-      final ExecutionPayload executionPayload,
+      ExecutionPayloadHeader executionPayloadHeader,
+      Optional<ExecutionPayload> executionPayload,
```
ajsutton: nit: final
```suggestion
      final ExecutionPayloadHeader executionPayloadHeader,
      final Optional<ExecutionPayload> executionPayload,
```

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/block/BlockProcessorAltair.java
```diff
@@ -250,7 +252,8 @@ public void processSyncAggregate(
   @Override
   public void processExecutionPayload(
       final MutableBeaconState state,
-      final ExecutionPayload executionPayload,
+      ExecutionPayloadHeader payloadHeader,
+      Optional<ExecutionPayload> payload,
```
ajsutton: nit: final
```suggestion
      final ExecutionPayloadHeader payloadHeader,
      final Optional<ExecutionPayload> payload,
```

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/bellatrix/block/BlockProcessorBellatrix.java
```diff
@@ -79,68 +80,72 @@ public void processBlock(
...
+        .equals(executionPayloadHeader.getTimestamp())) {
       throw new BlockProcessingException(
           "Execution payload timestamp does not match time for state slot");
     }
 
-    final boolean optimisticallyAccept =
-        payloadExecutor.optimisticallyExecute(state.getLatestExecutionPayloadHeader(), payload);
-    if (!optimisticallyAccept) {
-      throw new BlockProcessingException("Execution payload was not optimistically accepted");
+    if (executionPayload.isPresent()) {
```
ajsutton: This makes me a bit nervous that we may wind up not having the payload here when importing the block and this would then result in it being considered valid. Not entirely sure how we resolve that - maybe a check in `ForkChoice.onBlock` to require the block not be blinded could be enough?

ajsutton (approved):
LGTM. Just worth seeing what we can do to be a bit more defensive about ensuring we don't import a blinded block.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/bellatrix/block/BlockProcessorBellatrix.java
tbenr: right.

> maybe a check in ForkChoice.onBlock to require the block not be blinded could be enough?

should work, but I'll think about an alternative.

tbenr: I think this needs to be more structural. `BlockProcessor` requires a bit more of context so I made the `PayloadExecutor` to be optional and removed the NOOP. This way we are more strict in communicating when we have to execute a payload or not. And if we have to, we require the full payload to be there.

WDYT?

ajsutton (approved):
Yeah I like that approach.  Makes it much harder to screw up. :)

PR #5274, opened 2022-04-01
Batch verify deposit signatures

### LINKED ISSUE #5254: Optimisation idea: batch verify deposit signatures in a block
Currently we check deposit signatures [one-by-one](https://github.com/ConsenSys/teku/blob/70e2d323748f668af288c89d9c5813bab28b8df1/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java#L654) when processing new deposits. Since there are up to 16 deposits per block we could likely save some time by batch verifying them together. We don't need to know which one fails (if any) as the whole block is invalid if any signature is invalid.

benjaminion:
This processes a full set of 16 deposits in around 4ms rather than 22ms - over 5x speedup.

Logic:
  - We check signatures of all the block's deposits in a batch.
      - Almost always this will succeed and we can then process all the deposits without checking any individual signatures.
  - If the batch check fails then we fall back to normal deposit processing
     -  This verifies signatures for new validators and ignores the signatures for existing validators as usual.

Resolves #5254.

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
@@ -30,6 +30,7 @@ import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.bls.BLS;
 import tech.pegasys.teku.bls.BLSPublicKey;
+import tech.pegasys.teku.bls.BLSSignature;
 import tech.pegasys.teku.bls.BLSSignatureVerifier;
 import tech.pegasys.teku.infrastructure.bytes.Bytes4;
 import tech.pegasys.teku.infrastructure.crypto.Hash;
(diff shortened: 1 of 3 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
```diff
@@ -606,13 +607,37 @@ public void processDeposits(MutableBeaconState state, SszList<? extends Deposit>
       throws BlockProcessingException {
     safelyProcess(
         () -> {
+          final boolean depositSignaturesAreAllGood = batchVerifyDepositSignatures(deposits);
```
ajsutton: Things like the fuzzer disable deposit signature verification entirely to speed up processing. So we probably want to move that check to here since the net result is that we treat all signatures as valid.

```suggestion
          final boolean depositSignaturesAreAllGood = !blsVerifyDeposit || batchVerifyDepositSignatures(deposits);
```

And then we can remote the `blsVerifyDeposit` check from `depositSignatureIsValid`.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
```diff
@@ -606,13 +607,37 @@ public void processDeposits(MutableBeaconState state, SszList<? extends Deposit>
...
+    final List<BLSSignature> signatures = new ArrayList<>();
+    for (Deposit deposit : deposits) {
+      final BLSPublicKey pubkey = deposit.getData().getPubkey();
+      final Bytes signingRoot =
+          miscHelpers.computeSigningRoot(
+              new DepositMessage(
+                  pubkey,
+                  deposit.getData().getWithdrawalCredentials(),
+                  deposit.getData().getAmount()),
+              domain);
```
ajsutton: nit: Probably worth extracting a method to calculate the signingRoot for a `Deposit` as we're doing it here and in the existing `depositSignatureIsValid` and they always have to be the same.

ajsutton (approved):
LGTM. Just a couple of minor suggestions.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
benjaminion: Good catch!

It seems difficult to remove the check from `depositSignatureIsValid` since the `BlockValidatorTest` class sets `blsVerifyDeposit` to false and ends up calling `processDepositWithoutCheckingMerkleProof` directly via `GenesisGenerator`. That is, it bypasses the batch verification.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
benjaminion: 👍

PR #5277, opened 2022-04-01
Use entrySet instead of keySet when iterating over the whole map

jtraglia:
## PR Description

This replaces `keySet` with `entrySet` for loops that iterate over the whole map.

It is more efficient to use an iterator on the `entrySet` of the map, avoiding the extra `HashMap.get(key)` lookup.

See: https://stackoverflow.com/a/3870093 & https://stackoverflow.com/a/38116169

## Documentation

## Changelog


jtraglia:
These were identified using Facebook's [Infer](https://fbinfer.com) static analysis tool. The [Inefficient Keyset Iterator](https://fbinfer.com/docs/all-issue-types/#inefficient_keyset_iterator) issue type.

ajsutton (approved):
LGTM.  For the record, we looked at Infer a little while back but found the false failure rate was far too high to be useful.

jtraglia:
I came to the same conclusion about Infer. Neat tool but lots of false negatives.

PR #5280, opened 2022-04-02
Remove deprecated ChainBuilder.create constructor

jtraglia:
## PR Description

This PR removes the deprecated constructor and updates old instances. New constructor needs a spec.

## Documentation

## Changelog


ajsutton (approved):
LGTM. Guessing the test is flaky but will double check. We've had a bit of an increase in flakiness recently that we'll need to look into and fix.

PR #5281, opened 2022-04-02
Fix "unknown enum constant" warnings

jtraglia:
## PR Description

This gets rid of the "unknown enum constant" warning messages in the build:

```
> Task :eth-benchmark-tests:compileJmhJava
warning: unknown enum constant Status.STABLE
  reason: class file for org.apiguardian.api.API$Status not found
warning: unknown enum constant Status.STABLE
warning: unknown enum constant Status.STABLE
  reason: class file for org.apiguardian.api.API$Status not found
warning: unknown enum constant Status.STABLE
warning: unknown enum constant Status.STABLE
  reason: class file for org.apiguardian.api.API$Status not found
warning: unknown enum constant Status.STABLE
warning: unknown enum constant Status.STABLE
  reason: class file for org.apiguardian.api.API$Status not found
warning: unknown enum constant Status.STABLE
```

We needed to add `apiguardian-api` as a `runtimeOnly` dependency.

See: https://stackoverflow.com/a/46950496

## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -730,7 +730,6 @@ subprojects {
 
     runtimeOnly 'org.apache.logging.log4j:log4j-core'
     runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
-    runtimeOnly 'org.apiguardian:apiguardian-api'
 
     testImplementation 'org.apache.tuweni:tuweni-junit'
     testImplementation 'org.assertj:assertj-core'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -730,6 +730,7 @@ subprojects {
 
     runtimeOnly 'org.apache.logging.log4j:log4j-core'
     runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
+    runtimeOnly 'org.apiguardian:apiguardian-api'
```
ajsutton: I suspect we can add this as a testRuntimeOnly if it's required by junit.  Adding it to `runtimeOnly` will cause it to be shipped with Teku which I don't think we want.

jtraglia:
Good call. It still prints the warnings if it is `testRuntimeOnly` but `testFixturesRuntimeOnly` does the trick. Is this fine?

ajsutton:
Yep `testFixtureRuntimeOnly` works fine. Thanks.

ajsutton (approved):
LGTM.

PR #5288, opened 2022-04-04
Migrate the publish blinded block intereface to new api.

rolfyone:
- started building out async handling for responding to a request
 - built out parser to determine the type of object based on an attribute of that object
 - publish blinded block will now actually publish phase0 or altair blocks via existing channel interface.
 - new SchemaDefinitionCacheTest to ensure that schemas aren't being recreated, but rather re-used.

 For validation, changed the block publishing api used by the VC to use this one temporarily and it is working cleanly (validation activity only).

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/schema/SignedBlindedBlockBellatrix.json
+++ b/data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/schema/SignedBlindedBlockBellatrix.json
@@ -1,0 +1,16 @@
+{
+  "title" : "SignedBlindedBlockBellatrix",
+  "type" : "object",
+  "required" : [ "message", "signature" ],
+  "properties" : {
+    "message" : {
+      "$ref" : "#/components/schemas/BlindedBlockBellatrix"
+    },
+    "signature" : {
+      "type" : "string",
+      "pattern" : "^0x[a-fA-F0-9]{2,}$",
+      "description" : "SSZ hexadecimal",
+      "format" : "bytes"
+    }
+  }
+}
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostBlindedBlock.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostBlindedBlock.java
@@ -128,7 +128,9 @@ public class PostBlindedBlock extends MigratingEndpointAdapter {
                   .getRejectionReason()
                   .get()
                   .equals(BlockImportResult.FailureReason.INTERNAL_ERROR.name())) {
-                return AsyncApiResponse.respondWithCode(SC_INTERNAL_SERVER_ERROR);
+                return AsyncApiResponse.respondWithError(
+                    SC_INTERNAL_SERVER_ERROR,
+                    "An internal error occurred, check the server logs for more details.");
               } else {
                 return AsyncApiResponse.respondWithCode(SC_ACCEPTED);
               }
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostBlindedBlock.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostBlindedBlock.java
@@ -13,9 +13,9 @@
 
 package tech.pegasys.teku.beaconrestapi.handlers.v1.beacon;
 
-import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
 import static javax.servlet.http.HttpServletResponse.SC_SERVICE_UNAVAILABLE;
 import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_ACCEPTED;
+import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_INTERNAL_SERVER_ERROR;
 import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_OK;
 import static tech.pegasys.teku.infrastructure.http.RestApiConstants.RES_ACCEPTED;
 import static tech.pegasys.teku.infrastructure.http.RestApiConstants.RES_BAD_REQUEST;
@@ -39,7 +39,6 @@ import org.jetbrains.annotations.NotNull;
 import tech.pegasys.teku.api.DataProvider;
 import tech.pegasys.teku.api.SyncDataProvider;
 import tech.pegasys.teku.api.ValidatorDataProvider;
-import tech.pegasys.teku.api.schema.ValidatorBlockResult;
 import tech.pegasys.teku.api.schema.interfaces.SignedBlindedBlock;
 import tech.pegasys.teku.beaconrestapi.MigratingEndpointAdapter;
 import tech.pegasys.teku.beaconrestapi.SchemaDefinitionCache;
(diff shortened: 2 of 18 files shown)


data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/schema/SignedBlindedBlockBellatrix.json
```diff
@@ -0,0 +1,16 @@
+{
+  "title" : "SignedBlindedBlockBellatrix",
+  "type" : "object",
+  "required" : [ "message", "signature" ],
+  "properties" : {
+    "message" : {
+      "$ref" : "#/components/schemas/BlindedBlockBellatrix"
+    },
+    "signature" : {
```
ajsutton, line 9: nit: I wonder if we should update the `SszBlsSignature` type to set a name so it gets included by reference instead of repeated since it shows up quite a lot.  Possibly not for this PR but may be worth it.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCacheTest.java
```diff
@@ -68,4 +77,30 @@ void shouldCreateSchemaIfMilestoneRequired() {
...
+            method -> {
+              try {
+                LOG.info(method.getName());
+                final SszSchema<?> first =
+                    (SszSchema<?>) method.invoke(cache.getSchemaDefinition(specMilestone));
+                final SszSchema<?> second =
+                    (SszSchema<?>) method.invoke(cache.getSchemaDefinition(specMilestone));
+                assertThat(first.getJsonTypeDefinition()).isEqualTo(second.getJsonTypeDefinition());
+              } catch (IllegalAccessException | InvocationTargetException e) {
+                LOG.error(e);
```
ajsutton: Seems like we should fail the test if there are errors here rather than just logging (so probably `Assertions.fail(e)` I think).

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/AsyncApiResponse.java
```diff
@@ -0,0 +1,45 @@
...
+  final String responseBody;
+
+  private AsyncApiResponse(final int responseCode, final String responseBody) {
+    this.responseCode = responseCode;
+    this.responseBody = responseBody;
+  }
+
+  public int getResponseCode() {
+    return 0;
+  }
```
rolfyone: need to fix this

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -213,23 +217,19 @@ public SignedBeaconBlock parseBlindedBlock(
...
+  public SafeFuture<ValidatorBlockResult> submitSignedBlindedBlock(
+      final tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock signedBeaconBlock) {
+    LOG.debug("parsed block is from slot: {}", signedBeaconBlock.getMessage().getSlot());
+    final SpecMilestone milestone = spec.atSlot(signedBeaconBlock.getSlot()).getMilestone();
+    if (milestone == SpecMilestone.PHASE0 || milestone == SpecMilestone.ALTAIR) {
+      return validatorApiChannel
+          .sendSignedBlock(signedBeaconBlock)
+          .thenApply(ValidatorDataProvider::generateSubmitSignedBlockResponse);
+    }
+    return SafeFuture.failedFuture(new IllegalStateException("Not implemented"));
```
ajsutton: This hard coding makes me nervous that we'll forget it and also that it will break our ability to produce bellatrix non-blinded blocks. I'd have thought we'd just pass through the block and let `ValidatorApiChannel` handle it.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/AsyncApiResponse.java
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.infrastructure.restapi.endpoints;
+
+public class AsyncApiResponse {
+  final int responseCode;
+  final String responseBody;
```
ajsutton: This should be an `Object` not a `String` as the `RestApiRequest` should handle serialisation with the appropriate type based on the status code (same as it does for `respondOk` now).

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/RestApiRequest.java
```diff
@@ -50,6 +50,17 @@ public void respondOk(final Object response) throws JsonProcessingException {
...
   }
 
+  public void respondAsync(final SafeFuture<AsyncApiResponse> futureResponse) {
+    context.future(
+        futureResponse.thenAccept(
+            result -> {
+              context.status(result.getResponseCode());
+              if (result.hasResponseBody()) {
+                context.json(result.getResponseBody());
+              }
```
ajsutton: Would be simpler if `result.getResponseBody()` just returned an `Optional` then you could use `.ifPresent`.

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
rolfyone: previously it'd bail well before now for all types, could throw a not implemented exception? my guess is the api channel will need to be a different interface?

ajsutton: oh sorry I missed the blinded in the method name.  I though it was just `submitSignedBlock`.  I'd still just pass it through to `ValidatorApiChannel` - it can detect if the block is blinded or not and #5273 seems to be moving towards just handling that seamlessly.

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/schema/SignedBlindedBlockBellatrix.json
rolfyone: It'll be worth it, we do get a lot of signatures in the beacon-api.. agree not this PR though :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostBlindedBlock.java
```diff
@@ -96,28 +108,77 @@ public PostBlindedBlock(final DataProvider dataProvider, final JsonProvider json
...
+    request.respondAsync(
+        result.thenApply(
+            blockResult -> {
+              if (blockResult.getRejectionReason().isEmpty()) {
+                return AsyncApiResponse.respondWithCode(SC_OK);
+              } else if (blockResult
+                  .getRejectionReason()
+                  .get()
+                  .equals(BlockImportResult.FailureReason.INTERNAL_ERROR.name())) {
+                return AsyncApiResponse.respondWithCode(SC_INTERNAL_SERVER_ERROR);
```
ajsutton: Should we be including the error message here? `PostBlock` winds up including the rejection reason when reporting an error (empty response for SC_OK and SC_ACCEPTED).

ajsutton (approved):
LGTM.

PR #5294, opened 2022-04-05
Initial schema definitions for Event types that aren't just standard objects

rolfyone:
These were missing from the schema definitions we use from swagger, this test creates a definition from the class, which is a little ugly.

Once we're using the new method this test will be a lot cleaner, but this should allow us to be confident the structure is correct, despite the lacking typing information.

## Documentation

## Changelog


data/serializer/src/test/resources/tech/pegasys/teku/api/response/v1/schema/BlockEvent.json
```diff
@@ -0,0 +1,32 @@
...
+      "properties" : {
+        "empty" : {
+          "type" : "boolean"
+        },
+        "zero" : {
+          "type" : "boolean"
+        }
+      },
+      "required" : true
+    },
```
ajsutton: This looks like it's looking into the Bytes32 type.  It doesn't matter because we don't actually output this content anywhere but it does mean these definitions will change quite a lot when we switch to type definitions.

infrastructure/restapi/src/testFixtures/java/tech/pegasys/teku/infrastructure/restapi/OpenApiTestUtil.java
```diff
@@ -82,6 +83,36 @@ public void compareToKnownDefinitions(
...
+    final String filename = clazz.getSimpleName() + ".json";
+    final String namespace = "schema";
+
+    final SchemaFactoryWrapper visitor = new SchemaFactoryWrapper();
+
+    mapper.acceptJsonFormatVisitor(mapper.constructType(clazz), visitor);
+    final JsonNode node =
+        mapper.readTree(
+            mapper.writerWithDefaultPrettyPrinter().writeValueAsString(visitor.finalSchema()));
+    Files.write(tempDir.resolve(filename), prettyJson(node).getBytes(UTF_8));
```
ajsutton: Do we need to write these to file all the time? I can see it would be useful when initially creating them to have them all written but probably not necessary every time we run the test.

ajsutton (approved):
LGTM. Just a couple of things to consider.

data/serializer/src/test/resources/tech/pegasys/teku/api/response/v1/schema/BlockEvent.json
rolfyone: its not respecting typings that we add, its the default ugliness that comes if schema tags aren't written. it'll cleanup once we change to the other way of doing things so i'll probably merge courtneys change into this branch and make it less ugly...

infrastructure/restapi/src/testFixtures/java/tech/pegasys/teku/infrastructure/restapi/OpenApiTestUtil.java
rolfyone: fair point, will see if we can get away without doing that, I don't use it when initially creating the test either...

PR #5295, opened 2022-04-05
Migrate GetGenesis to MigratingEndpointAdapter

### LINKED ISSUE #5291: Migrate GetGenesis to MigratingEndpointAdapter

courtneyeh:
fixes #5291 

## Documentation

## Changelog


data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/paths/_eth_v1_beacon_genesis.json
```diff
@@ -0,0 +1,65 @@
...
+                  "example" : "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
+                  "format" : "byte"
+                },
+                "genesis_fork_version" : {
+                  "type" : "string",
+                  "description" : "Bytes4 hexadecimal",
+                  "example" : "0xcf8e0d4e",
+                  "format" : "byte"
+                }
+              }
```
rolfyone: this will all go a level deeper once we fix impl...

rolfyone: this will all go a level deeper once we fix impl...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetGenesis.java
```diff
@@ -66,18 +81,42 @@ public GetGenesis(final DataProvider dataProvider, final JsonProvider jsonProvid
...
-    ctx.json(jsonProvider.objectToJSON(new GetGenesisResponse(maybeData.get())));
+    request.respondOk(maybeData.get());
   }
 
   private Optional<GenesisData> getGenesisData() {
     if (!chainDataProvider.isStoreAvailable()) {
       return Optional.empty();
     }
-    return Optional.of(chainDataProvider.getGenesisData());
+    return Optional.of(chainDataProvider.getStateGenesisData());
```
rolfyone: Im just not sure if we need to go down this route of using a wrapper object... it feels like we shouldn't need it any more?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetGenesis.java
```diff
@@ -66,18 +81,42 @@ public GetGenesis(final DataProvider dataProvider, final JsonProvider jsonProvid
...
+      ChainDataProvider chainDataProvider) {
+    final SerializableTypeDefinition<GenesisData> dataType =
+        SerializableTypeDefinition.object(GenesisData.class)
+            .withField("genesis_time", UINT64_TYPE, GenesisData::getGenesisTime)
+            .withField(
+                "genesis_validators_root", BYTES32_TYPE, GenesisData::getGenesisValidatorsRoot)
+            .withField(
+                "genesis_fork_version",
+                BYTES4_TYPE,
+                data -> chainDataProvider.getGenesisForkVersion())
```
rolfyone: this section here is more like what I expected all the fields to be...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetGenesis.java
courtneyeh: It isn't a wrapper object anymore, that is getting the state `GenesisData` object

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetGenesisTest.java
```diff
@@ -51,6 +51,6 @@ public void shouldReturnGenesisInformation() throws Exception {
     when(chainDataProvider.getStateGenesisData()).thenReturn(expectedGenesisData);
```
rolfyone: not sure this when will even get called in this test now

rolfyone: Hopefully we have some kind of integration test that gives us actual data, this test is borderline able to be deleted, its not doing anything any more

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -119,6 +120,17 @@ public GenesisData getGenesisData() {
         spec.atEpoch(ZERO).getConfig().getGenesisForkVersion());
   }
 
+  public tech.pegasys.teku.spec.datastructures.genesis.GenesisData getStateGenesisData() {
+    if (!isStoreAvailable()) {
+      throw new ChainDataUnavailableException();
+    }
+    return recentChainData.getGenesisData().orElseThrow(ChainDataUnavailableException::new);
+  }
```
rolfyone: This does worry me, I'm not sure it's consistent with old handling (haven't checked), we'd want to make sure that a pre-genesis chain is not doing something horrible like throwing big errors... easiest way would be to set a minimal chain with a long pre-genesis and make sure the logs are clean when we call this... I'm starting to wonder if we should go back to using an object so that we can use `request.handleOptionalResult`

ajsutton: A pre-genesis chain would return false from `isStoreAvailable()`. There's no way to have a Store without having the genesis set so this should be fine. It may in fact be worth removing the `isStoreAvailable` check and just using the `orElseThrow` to throw the `ChainDataUnavailableException` when genesis data is unavailable - currently it's essentially duplicating the check.

courtneyeh: This basically does the same as previously, except returns the actual data, rather than the wrapper class. @rolfyone

rolfyone (approved):
LGTM

PR #5296, opened 2022-04-05
Add configuration for the Gnosis beacon chain

### LINKED ISSUE #5251: Add support for Gnosis Beacon Chain
Gnosis is establishing a beacon chain ([GBC - Gnosis Beacon Chain](https://docs.gnosischain.com/)) with a plan to merge the existing Gnosis Chain (formerly xDai) with GBC shortly ahead of the actual Ethereum Merge.

GBC currently runs on Prysm and Lighthouse. Gnosis has requested that we add GBC support to Teku.

GBC [parameters](https://docs.gnosischain.com/contracts-addresses-parameters).

StefanBratanov:
## PR Description

Add configuration for the Gnosis Beacon Chain. Usage: `--network gnosis`

## Fixed Issue(s)
fixes #5251

Synced Gnosis Chain Locally
syncing-gnosis

## Documentation

## Changelog


ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/presets/gnosis/phase0.yaml
```diff
@@ -0,0 +1,94 @@
...
+# 2**24 (= 16,777,216) historical roots, ~15,243 years
+HISTORICAL_ROOTS_LIMIT: 16777216
+# 2**40 (= 1,099,511,627,776) validator spots
+VALIDATOR_REGISTRY_LIMIT: 1099511627776
+
+
+# Reward and penalty quotients
+# ---------------------------------------------------------------
+# 2**6 (= 64)
+BASE_REWARD_FACTOR: 25
```
tbenr: The comment do not match with the actual value. Maybe it was misaligned even in the source you took as reference?
The value isn't power of 2 but i think int this case we should not have problems.

tbenr (approved):
LGTM! Just a thing to double-check before merging.

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/presets/gnosis/phase0.yaml
StefanBratanov: just checked with Gnosis team, 25 is the correct value, adding a commit to fix the comment typo

Issue #5103, opened 2022-03-03
Adopt `BlindedBeaconBlocks` in standard block production flow (non-MEVBoost)


tbenr:
More details here: https://hackmd.io/-AWMItrHSiqCjeE2l3spwQ?view#Adopting-BlindedBeaconBlocks-flow-in-normal-VCBN-communication.

In short, we can adopt blinded blocks in the REST API calls between VC and BN, independently by the MEV Boost support.
It is a good optimization by itself, which will reduce traffic between VC and BN.

related to: #5017 

and to: #5030

PR #5298, opened 2022-04-05
[Mev Boost] Allow Blinded block creation flow in BN - Part3

### LINKED ISSUE #5261: [MEV Boost] Update BN block production and processing flow to support blinded block and execute the "unblinding" flow via builder API

tbenr:
continuation of #5273

Introduces `SignedBeaconBlockBlinder` and `SignedBeaconBlockUnblinder`.
The classes will be provided via `SpecLogic::getBlindBlockUtil`, so per milestone.

`SignedBeaconBlockUnblinder` is configurable at runtime, the way we provide the `executionPayload` to unblind the block is provided by `operationSelector.createUnblinderSelector()` (similar to block building)

I provide a MEV Boost UnblinderSelector for Bellatrix unblinder which retrieves the execution payload via `proposeBlindedBlock` API.

`SignedBeaconBlockBlinder` is only implemented for Bellatrix, which helped me implementing the unit tests.
We need to implement `Phase0` and `Altair` version to support blinded flow when mev is not enabled.

fixes #5261
starts work for #5103

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -201,4 +202,20 @@ public BlockOperationSelectorFactory(
...
   }
+
+  public Consumer<BeaconBlockUnblinder> createUnblinderSelector() {
+    return bodyUnblinder -> {
+      if (isMevBoostEnabled) {
+        bodyUnblinder.executionPayload(
+            () ->
+                executionEngineChannel
+                    .proposeBlindedBlock(bodyUnblinder.getSignedBlindedBeaconBlock())
+                    .join());
```
ajsutton: It would be really nice to design this to not use `.join()` here and return a `SafeFuture`. At some point in the block factory stuff we may need a `.join()` because block creation is currently synchronous but the higher up the call stack we can push it the better as it will be easier to get rid of later then.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/altair/BeaconBlockUnblinderAltair.java
```diff
@@ -0,0 +1,25 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.altair;
+
+import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.phase0.BeaconBlockUnblinderPhase0;
+import tech.pegasys.teku.spec.schemas.SchemaDefinitions;
+
+public class BeaconBlockUnblinderAltair extends BeaconBlockUnblinderPhase0 {
```
ajsutton: I'd be tempted to just use the `Phase0` version for Altair as well and not bother with an Altair specific class.  There's quite a few things that Altair inherited unchanged from phase0 already.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
tbenr: yeah, good exercise for the complete removal.

tbenr:
There's no point in comparing all block fields in unit tests.  Going to switch to comparing hashroots

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCacheTest.java
```diff
@@ -89,7 +90,7 @@ void subsequentCallsShouldGetTheSameSchemaObject(final SpecMilestone specMilesto
     // all the get methods should return the same object on subsequent calls,
     // as creation of the schema objects is relatively expensive
     methods.stream()
-        .filter(method -> method.getName().startsWith("get"))
+        .filter(method -> method.getAnnotation(NonSchema.class) == null)
```
ajsutton, line 93: Won't this still include methods like hashCode and equals?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/SignedBeaconBlockUnblinder.java
```diff
@@ -0,0 +1,28 @@
...
+import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayload;
+
+public interface SignedBeaconBlockUnblinder {
+
+  SignedBeaconBlockUnblinder executionPayload(
+      Supplier<SafeFuture<ExecutionPayload>> executionPayloadSupplier);
+
+  SignedBeaconBlock getSignedBlindedBeaconBlock();
+
+  SafeFuture<SignedBeaconBlock> unblind();
```
ajsutton: This seems like quite an unintuitive set of methods.  It may be worth adding comments here to explain how they're expected to be used (and any required ordering).

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/SignedBeaconBlockBlinderBellatrix.java
```diff
@@ -0,0 +1,73 @@
...
+
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszBytes32;
+import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlock;
+import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.common.AbstractSignedBeaconBlockBlinder;
+import tech.pegasys.teku.spec.datastructures.type.SszSignature;
+import tech.pegasys.teku.spec.schemas.SchemaDefinitions;
+
+public class SignedBeaconBlockBlinderBellatrix extends AbstractSignedBeaconBlockBlinder {
+  public SignedBeaconBlockBlinderBellatrix(SchemaDefinitions schemaDefinitions) {
```
ajsutton: We should probably require this to be `SchemaDefinitionsBellatrix` since we're casting returned types to bellatrix when using it.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/SignedBeaconBlockBlinderBellatrix.java
```diff
@@ -0,0 +1,73 @@
...
+    final BlindedBeaconBlockBodySchemaBellatrixImpl schema =
+        (BlindedBeaconBlockBodySchemaBellatrixImpl)
+            schemaDefinitions.getBlindedBeaconBlockBodySchema();
+
+    final BlindedBeaconBlockBodyBellatrixImpl blindedBody =
+        new BlindedBeaconBlockBodyBellatrixImpl(
+            schema,
+            new SszSignature(unblindedBlockBody.getRandaoReveal()),
+            unblindedBlockBody.getEth1Data(),
+            SszBytes32.of(unblindedBlockBody.getGraffiti()),
```
ajsutton: We should probably add getters to get the original `SszSignature` version of randaoReveal and graffiti instead of unwrapping and rewrapping them.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/SignedBeaconBlockUnblinderBellatrix.java
```diff
@@ -0,0 +1,98 @@
...
+                  .hashTreeRoot()
+                  .equals(blindedBody.getExecutionPayloadHeader().hashTreeRoot()),
+              "executionPayloadHeader root in blinded block do not match provided executionPayload root");
+
+          final BeaconBlockBodyBellatrix unblindedBody =
+              new BeaconBlockBodyBellatrixImpl(
+                  (BeaconBlockBodySchemaBellatrixImpl) schemaDefinitions.getBeaconBlockBodySchema(),
+                  new SszSignature(blindedBody.getRandaoReveal()),
+                  blindedBody.getEth1Data(),
+                  SszBytes32.of(blindedBody.getGraffiti()),
```
ajsutton: Would be good to avoid the unwrap/rewrap here too.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/BlockProposalUtil.java
```diff
@@ -99,6 +102,42 @@ public BeaconBlockAndState createNewUnsignedBlock(
...
+              blockUnblinder.accept(unblinder);
+              return unblinder.unblind();
+            })
+        .orElseGet(
+            () -> {
+              // this shouldn't happen: BlockFactory should skip unblinding when is not needed
+              if (signedBeaconBlock.getMessage().getBody().isBlinded()) {
+                throw new IllegalStateException(
+                    "Unblinder not available for the current spec but the given block was blinded");
+              }
```
ajsutton: nit: Could just use Preconditions.checkState for this.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitions.java
```diff
@@ -66,11 +73,24 @@ default AttestationSchema getAttestationSchema() {
 
   AttesterSlashingSchema getAttesterSlashingSchema();
 
+  @NonSchema
+  Optional<SignedBeaconBlockUnblinder> createSignedBeaconBlockUnblinder(
+      SignedBeaconBlock signedBlindedBeaconBlock);
+
+  @NonSchema
+  Optional<SignedBeaconBlockBlinder> getSignedBeaconBlockBlinder();
```
ajsutton: hmm, I'm not entirely convinced these belong in `SchemaDefinitions`. I can see the reasoning for it because they make a lot of use of schema definitions, but it feels like we've expanded the scope of `SchemaDefintiions` from providing just the definitions to providing logic now as well and that should really come from something like `BlockProposalUtil`.

ajsutton (approved):
LGTM, mostly just unsure about the changes to `SchemaDefinition` - I think the new methods would be better in something like `BlockProposalUtil` (or even a new object that `SpecLogic` holds if that makes more sense.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/SchemaDefinitionCacheTest.java
tbenr: good point but, nope. I think because we are listing methods from an `interface`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/SignedBeaconBlockBlinderBellatrix.java
tbenr: yeah.. I'll do.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/SignedBeaconBlockBlinderBellatrix.java
tbenr: very good point.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitions.java
tbenr: 🤔 right. Placing them in `BlockProposalUtil` seems more natural.

tbenr: not really:) I'll found a better place

PR #5299, opened 2022-04-05
Invalid blocks should not be considered viable for head

ajsutton:
## PR Description
This ensures ProtoArray always updates the best child/descendant links correctly when marking a node as invalid. With thanks to Paul Hauner.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5301, opened 2022-04-06
Migrate GetSpec to MigratingEndpointAdapter

### LINKED ISSUE #5290: Migrate GetSpec to MigratingEndpointAdapter

courtneyeh:
fixes #5290 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/config/GetSpec.java
```diff
@@ -53,12 +65,18 @@ public GetSpec(final DataProvider dataProvider, final JsonProvider jsonProvider)
...
-  public void handle(final Context ctx) throws Exception {
+  public void handle(@NotNull final Context ctx) throws Exception {
+    adapt(ctx);
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
     try {
-      ctx.json(jsonProvider.objectToJSON(configProvider.getConfig()));
+      request.respondOk(configProvider);
```
courtneyeh: Shouldn't need to give a value to `respondOk()` method, however the `SerializableTypeDefinition` requires a value type.

data/provider/src/main/java/tech/pegasys/teku/api/ConfigProvider.java
```diff
@@ -92,7 +99,82 @@ public GetSpecResponse getConfig() {
     return new GetSpecResponse(configAttributes);
   }
 
-  private String formatValue(final Object v) {
+  public SerializableTypeDefinition<ConfigProvider> getSpecTypeDefinition() {
```
courtneyeh: Want to see if I can avoid using the type `ConfigProvider` for the `SerializableTypeDefinition`.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/CoreTypes.java
```diff
@@ -29,6 +30,15 @@
...
 
+  public static final StringValueTypeDefinition<Bytes> BYTES_TYPE =
+      DeserializableTypeDefinition.string(Bytes.class)
+          .formatter(Bytes::toHexString)
+          .parser(Bytes::fromHexString)
+          .example("0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2")
+          .description("Bytes hexadecimal")
+          .format("byte")
+          .build();
+
```
rolfyone: I'm really not keen on an unbounded bytes type still. In the spec `BLS_WITHDRAWAL_PREFIX` is bytes1, so i think we use that to read it out, and maybe at some point we need to cleanup the code to not use bytes for it... would be interested in @ajsutton take on `BLS_WITHDRAWAL_PREFIX`

data/provider/src/main/java/tech/pegasys/teku/api/ConfigProvider.java
```diff
@@ -92,7 +99,82 @@ public GetSpecResponse getConfig() {
...
+            "DOMAIN_AGGREGATE_AND_PROOF", BYTES4_TYPE, specConfig -> Domain.AGGREGATE_AND_PROOF);
+
+    // Manually add legacy altair constants
+    config
+        .toVersionAltair()
+        .ifPresent(
+            altairConfig -> {
+              builder
+                  .withField(
+                      "DOMAIN_SYNC_COMMITTEE", BYTES4_TYPE, specConfig -> Domain.SYNC_COMMITTEE)
```
rolfyone: i think we can write either `(__)` or `()` instead of specConfig which we don't appear to use in the lambda? this is a pretty common pattern through below...

data/provider/src/main/java/tech/pegasys/teku/api/ConfigProvider.java
```diff
@@ -92,7 +99,82 @@ public GetSpecResponse getConfig() {
...
+                  .withField(
+                      "DOMAIN_CONTRIBUTION_AND_PROOF",
+                      BYTES4_TYPE,
+                      specConfig -> Domain.CONTRIBUTION_AND_PROOF)
+                  .withField(
+                      "TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE",
+                      STRING_TYPE,
+                      specConfig ->
+                          Integer.toString(
+                              ValidatorConstants.TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE, 10))
```
rolfyone: i wonder if this would be neater if we had something like
```
private String getAggregatorsPerSyncCommittee() {
    return Integer.toString(ValidatorConstants.TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE, 10);
}
```
and then could just use `(__) -> getAggregatorsPerSyncCommittee()`
or make it take whatever input and just use a fn ref... maybe have a play?

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/CoreTypes.java
ajsutton: Not sure I'd want to add a `Bytes1` class for this but yeah probably makes sense for the API to be able to specify that it's a single length byte.  No reason that can't be done with the existing `Bytes` class - just need to verify there is only one byte when parsing and serialising.

data/provider/src/main/java/tech/pegasys/teku/api/ConfigProvider.java
courtneyeh: Updated to use `Integer.toString(int i)`, since radix is default 10. But moving this to another method does not help.

rolfyone: i'd do a specific getter for each field if anything like above...

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/CoreTypes.java
rolfyone: maybe what we should do is just define this internally to spec and have a parser...
I'd really like to not have this unbounded type defined as a core type.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/config/GetSpec.java
```diff
@@ -53,12 +68,18 @@ public GetSpec(final DataProvider dataProvider, final JsonProvider jsonProvider)
...
   @Override
-  public void handle(final Context ctx) throws Exception {
+  public void handle(@NotNull final Context ctx) throws Exception {
+    adapt(ctx);
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
     try {
-      ctx.json(jsonProvider.objectToJSON(configProvider.getConfig()));
```
rolfyone, line 58: does anything still call that `getConfig` other than tests?

courtneyeh: Three test files call `getConfig`

courtneyeh: Screen Shot 2022-04-07 at 1 08 55 PM

rolfyone (approved):
LGTM. if we can clean up that old configProvider code it'd be nice to not have it all duplicated, but haven't looked at how much work that'd be.

PR #5305, opened 2022-04-06
Add stricter limits to attestation pool size

### LINKED ISSUE #5196: Improve `AggregatingAttestationPool`

ajsutton:
## PR Description
When the total number of attestations in the pool grows too large, start dropping attestations from older slots.  It's expensive to determine exactly how many attestations we could include in a block because that depends on how well they aggregate, but we can set an approximate maximum number that are ever likely to be useful and ensure the pool size stays below that.  For efficiency and simplicity we drop all attestations from the oldest slot we have when the limit is exceeded which is cheap and the oldest attestations are likely to be the least valuable.

## Fixed Issue(s)
fixes #5196 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,3 +27,4 @@ For information on changes in released versions of Teku, see the [releases page]
 
 ### Bug Fixes
 - Fixed the target database format for the `migrate-database` command.
+- Added stricter limits on attestation pool size. 
--- a/ethereum/statetransition/src/integration-test/java/tech/pegasys/teku/statetransition/attestation/AttestationManagerIntegrationTest.java
+++ b/ethereum/statetransition/src/integration-test/java/tech/pegasys/teku/statetransition/attestation/AttestationManagerIntegrationTest.java
@@ -15,6 +15,7 @@ package tech.pegasys.teku.statetransition.attestation;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
+import static tech.pegasys.teku.statetransition.attestation.AggregatingAttestationPool.DEFAULT_MAXIMUM_ATTESTATION_COUNT;
 
 import it.unimi.dsi.fastutil.ints.IntList;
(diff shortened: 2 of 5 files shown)


tbenr (approved):
LGTM

PR #5308, opened 2022-04-06
Support new blinded block creation

rolfyone:
As part of this, the async handler had to be fixed to return a body, it was previously trying to serialize its own optional.

Reworked the old block creation so that we don't have 2 sets of code.

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
@@ -108,8 +108,12 @@ public class ValidatorDataProvider {
     return combinedChainDataClient.isStoreAvailable();
   }
 
-  public SafeFuture<Optional<BeaconBlock>> getUnsignedBeaconBlockAtSlot(
-      UInt64 slot, BLSSignature randao, Optional<Bytes32> graffiti) {
+  public SafeFuture<Optional<tech.pegasys.teku.spec.datastructures.blocks.BeaconBlock>>
+      getUnsignedBeaconBlockAtSlot(
+          final UInt64 slot,
+          final tech.pegasys.teku.bls.BLSSignature randao,
+          final Optional<Bytes32> graffiti,
+          final boolean isBlinded) {
     if (slot == null) {
       throw new IllegalArgumentException(NO_SLOT_PROVIDED);
     }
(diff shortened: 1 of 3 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -125,9 +129,15 @@ public boolean isStoreAvailable() {
     if (currentSlot.isGreaterThan(slot)) {
       throw new IllegalArgumentException(CANNOT_PRODUCE_HISTORIC_BLOCK);
     }
+    return validatorApiChannel.createUnsignedBlock(slot, randao, graffiti, false);
```
tbenr: why don't you pass the `isBlinded` param?

tbenr (approved):
LGTM with a couple of question

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -125,9 +129,15 @@ public boolean isStoreAvailable() {
...
+  }
 
-    return validatorApiChannel
-        .createUnsignedBlock(
+  public SafeFuture<Optional<BeaconBlock>> getUnsignedBeaconBlockAtSlot(
+      UInt64 slot, BLSSignature randao, Optional<Bytes32> graffiti) {
+    if (randao == null) {
+      throw new IllegalArgumentException(NO_RANDAO_PROVIDED);
```
tbenr, line 138: can we remove this check? it is done in the called method in any case, right?

tbenr (approved):

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
rolfyone: i started without it but then the conversion causes issues, so putting this back was the easy option...

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
rolfyone: ah i should really, thats where i started and just a plain old bug

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
tbenr: yeah, true.

PR #5313, opened 2022-04-08
allow the VC to use blinded blocks

rolfyone:
- when using `Xvalidators-blinded-blocks-api-enabled: true` and explicitly asking for blinded blocks, the new blinded blocks api will be called, and blinded bellatrix blocks will be received if appropriate.
 - the BN would also need to be accepting the blinded_blocks interfaces and sending blinded bellatrix blocks. It will be expected that any bellatrix block coming over the blinded_blocks api will be blinded if this option is enabled.
 - This flag will likely become default at some point, but currently this feature flag is disabled by default.

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/subcommand/RemoteSpecLoader.java
```diff
@@ -89,6 +89,6 @@ static OkHttpValidatorRestApiClient createApiClient(final URI baseEndpoint) {
     // Strip any authentication info from the URL to ensure it doesn't get logged.
     apiEndpoint = apiEndpoint.newBuilder().username("").password("").build();
     final OkHttpClient okHttpClient = httpClientBuilder.build();
-    return new OkHttpValidatorRestApiClient(apiEndpoint, okHttpClient);
+    return new OkHttpValidatorRestApiClient(apiEndpoint, okHttpClient, false);
```
ajsutton: If we can access it from here we should use `ValidatorOptions.DEFAULT_VALIDATOR_BLINDED_BLOCKS_ENABLED` here.  It won't change behaviour but will be more obviously that it doesn't care about the value when we come to remove the feature toggle.

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClientTest.java
```diff
@@ -232,6 +237,32 @@ public void createUnsignedBlock_WhenSuccess_ReturnsBeaconBlock() {
...
+                asJson(
+                    new GetNewBlindedBlockResponse(
+                        SpecMilestone.BELLATRIX, new BlindedBlockBellatrix(expectedBeaconBlock)))));
+    Optional<BeaconBlock> beaconBlock =
+        apiClient.createUnsignedBlock(slot, blsSignature, graffiti, true);
+    assertThat(beaconBlock).isPresent();
+
+    // TODO make comparable
+    // Some of the underlying stores are mutable, but the values are the same.
+    // It does make it harder to compare here though
```
ajsutton: Might need to do something with this TODO?

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/ValidatorApiMethod.java
```diff
@@ -23,7 +23,9 @@
   GET_VALIDATORS("eth/v1/beacon/states/head/validators"),
   GET_DUTIES("validator/duties"),
   GET_UNSIGNED_BLOCK_V2("eth/v2/validator/blocks/:slot"),
+  GET_UNSIGNED_BLINDED_BLOCK("eth/v1/validator/blinded_blocks/:slot"),
   SEND_SIGNED_BLOCK("eth/v1/beacon/blocks"),
+  SEND_SIGNED_BLINDED_BLOCK("eth/v1/beacon/blinded_blocks"),
```
ajsutton, line 28: Did we actually use this new API anywhere?

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
```diff
@@ -166,8 +173,8 @@ public OkHttpValidatorRestApiClient(final HttpUrl baseEndpoint, final OkHttpClie
     queryParams.put("randao_reveal", encodeQueryParam(randaoReveal));
     graffiti.ifPresent(bytes32 -> queryParams.put("graffiti", encodeQueryParam(bytes32)));
 
-    if (blinded) {
-      throw new UnsupportedOperationException("blinded blocks not yet implemented");
+    if (blindedBlocksEnabled && blinded) {
```
ajsutton: Shouldn't this be `||` instead of `&&`? ie if you explicitly request a blinded block we should request a blinded block but also if we're always using the blinded block creation API (blindedBlocksEnabled) then we should

ajsutton: Actually it's possible this is the wrong level to do the feature toggle at.  It might be more intuitive to have ValidatorRestApiClient just expose separate `createUnsignedBlock` and `createUnsignedBlindedBlock` methods (matching the API) and then `RemoteValidatorApiHandler` would have the feature toggle and select between the blinded and unblinded apis.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/ValidatorApiMethod.java
rolfyone: um.... no.... doing now. :/

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClientTest.java
rolfyone: its compared above so I'm not super worried. when it not working it goes boom... it'd be nice to do...

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
rolfyone: so this is more 'if we're using blinded blocks then I care, otherwise i don't'...
that's why i've gone with `&&`, because then the CLI will disable the api option basically...
the blinded option was the way it was handling things but the CLI should disable that, so unless i rework things...

rolfyone: my theory was that's what makes it a 'feature toggle' - it toggles things off if its not set...

rolfyone:
should all be ready now, split out asInternalBlock processing so that test could work without confusing the PR.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
```diff
@@ -178,9 +186,22 @@ public Optional<BeaconBlock> createUnsignedBlock(
...
+            queryParams,
+            createHandler(GetNewBlindedBlockResponse.class))
+        .map(response -> (BeaconBlock) response.data);
+  }
+
   @Override
   public SendSignedBlockResult sendSignedBlock(final SignedBeaconBlock beaconBlock) {
-    return post(SEND_SIGNED_BLOCK, beaconBlock, createHandler())
+    return post(
+            blindedBlocksEnabled ? SEND_SIGNED_BLINDED_BLOCK : SEND_SIGNED_BLOCK,
```
ajsutton: hmm, I think if MEV is enabled we're likely to wind up using the wrong API here - it should use the blinded block API matching the "|| blinded" when creating the block but it's only looking at the feature toggle.

If possible it might work best to just check `beaconBlock.getMessage().getBody().isBlinded()` so that we always use the right API for the data we're given.  That will then automatically use the API that matches createBlock because that's where the block came from.

ajsutton (approved):
LGTM. Just need to make sure we use the send block API that matches the create block we used.

PR #5316, opened 2022-04-08
Skip state root checks when regenerating blocks

### LINKED ISSUE #5242: Don't check state hashTreeRoot when reapplying blocks
### Description
Improve performance of reapplying blocks by skipping the check that the resulting state root matches the root in the block. This would have been checked when initially importing the block so we know it is correct. Skipping this check should enable us to avoid calculating the root for the state which can be quite expensive. Since most uses of states wind up accessing the state root, the initial benefit here will be when multiple blocks need to be replayed to generate a state - we should not need to calculate the root for the intermediate states.

A further optimisation would be to have the resulting `BeaconState` take it's state root from the value in the block so that it doesn't need to calculate it even if something later needs to access it.

ajsutton:
## PR Description
Avoid checking the state hash tree root when reapplying blocks to regenerate the state.

We were already skipping the check as part of the actual state transition (same as we skip checking the block signatures) but then had a separate sanity check as part of the regeneration pipeline which was run for every applied block.

## Fixed Issue(s)
fixes #5242 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ For information on changes in released versions of Teku, see the [releases page]
 
 
 ### Additions and Improvements
-
+- Improved performance when regenerating non-finalized states that had to be dropped from memory.
 
 ### Bug Fixes
 - Added stricter limits on attestation pool size. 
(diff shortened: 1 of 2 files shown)


rolfyone (approved):
LGTM

PR #5317, opened 2022-04-08
make getSpec return a generic map

rolfyone:
For future proofing, defined a map of String -> String to pass config in and out of, which means missing or extra fields won't break the definition.

This is more in line with the old style spec api, we made it more defined but in trying to remove duplication, we've gone to being generic again.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5319, opened 2022-04-08
Fixes #5293 (Report number of unknown validators)

### LINKED ISSUE #5293: Report number of unknown validators
### Description
The `validator_local_validator_counts` reports the number of validators in each status, but validators that teku has loaded keys for but has not yet seen a deposit from are not reported anywhere.  We should add a new `unknown` status to report those validators.

Neurone:
## PR Description
Now validator status is reported as `unknow` until explicitly updated otherwise.

## Fixed Issue(s)
#5293 

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ExternalMetricPublisherAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ExternalMetricPublisherAcceptanceTest.java
@@ -19,7 +19,8 @@ import tech.pegasys.teku.test.acceptance.dsl.ExternalMetricNode;
 import tech.pegasys.teku.test.acceptance.dsl.TekuNode;
 
 public class ExternalMetricPublisherAcceptanceTest extends AcceptanceTestBase {
-  private static final int VALIDATOR_COUNT = 8;
+  private static final int ACTIVE_VALIDATOR_COUNT = 8;
+  private static final int TOTAL_VALIDATOR_COUNT = 64;
 
   @Test
   void shouldPublishDataFromPrometheus() throws Throwable {
@@ -31,11 +32,12 @@ public class ExternalMetricPublisherAcceptanceTest extends AcceptanceTestBase {
             config ->
                 config
                     .withExternalMetricsClient(externalMetricNode, 1)
-                    .withInteropValidators(0, VALIDATOR_COUNT));
+                    .withInteropNumberOfValidators(TOTAL_VALIDATOR_COUNT)
+                    .withInteropValidators(0, ACTIVE_VALIDATOR_COUNT));
     tekuNode.start();
 
     externalMetricNode.waitForBeaconNodeMetricPublication();
-    externalMetricNode.waitForValidatorMetricPublication(VALIDATOR_COUNT);
+    externalMetricNode.waitForValidatorMetricPublication(ACTIVE_VALIDATOR_COUNT);
     externalMetricNode.waitForSystemMetricPublication();
 
     tekuNode.stop();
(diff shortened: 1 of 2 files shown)


ajsutton (commented):
Thanks for this.  The failing AT appears to be related (https://app.circleci.com/pipelines/github/ConsenSys/teku/17481/workflows/43657cbf-7457-4d89-9395-69f1c1535d6a/jobs/119032).  I'm not sure of the details but I believe it's checking the number of validators reported to the external metrics (e.g. beaconcha.in's service) and it's now getting more than it previously did.  Whether that's right or not just depends on whether it's specifically meant to report active validators or total validators.

I won't get a chance to look into it in more detail today but if you're unsure about it give me a bit of time and I'll dig in to work out what the right behaviour is.

Neurone:
Yes, it seems related to total validators, that are counted using `local_validator_counts` in the AT. It seems to me AT should be updated, but I dig more.

Neurone:
I fixed the acceptance test. The problem was with the initialization of the network: by default, it starts with 64 validators in total, and they are all owned by the user.

The initial setting of the previous scenario for the acceptance test was: **8 total validators, with 64 owned by the user**. This is clearly an error but it was not noticed before because the check is done using `local_validator_counts`, and that metric did not return `unknown` values until now, so it was always equal to the validators owned by the user. With `unknown` status now visible, the bug emerged.

Now the settings for the scenario are correct (or at least it seems to me what was intended for that test, let me know what you think), so there are **64 total validators in the network, with 8 validators owned by the user**, and tests pass.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ExternalMetricPublisherAcceptanceTest.java
```diff
@@ -31,7 +31,7 @@ void shouldPublishDataFromPrometheus() throws Throwable {
             config ->
                 config
                     .withExternalMetricsClient(externalMetricNode, 1)
-                    .withInteropNumberOfValidators(VALIDATOR_COUNT));
+                    .withInteropValidators(0, VALIDATOR_COUNT));
```
ajsutton: looking at the actual metrics it's expecting it looks like it actually publishes two separate metrics: `validator_total` and `validator_active`.  Previously they've always been (incorrectly) the same value.  So I think we actually want to go back to the validator setup this had previously but probably be more explicit.  eg:

```
.withInteropNumberOfValidators(VALIDATOR_COUNT)
.withInteropValidators(0, 64)
```

And then I'd probably rename `VALIDATOR_COUNT` to `ACTIVE_VALIDATOR_COUNT` and make the 64 a constant `TOTAL_VALIDATOR_COUNT`.  We'll then need to update the line:
```
    externalMetricNode.waitForValidatorMetricPublication(VALIDATOR_COUNT);
```

to take two params, one for the expected total validator count and one for the expected active validator count.  

And I think the production code is now doing the right thing because of your changes and should report the active and total values accurately.

ajsutton (commented):
Apologies for the delay in getting back to this one - had a busy weekend. :) I think the production code is good but the AT probably needs a slightly different tweak.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ExternalMetricPublisherAcceptanceTest.java
Neurone: I can work on this in a few weeks, but I don't know if I understood correctly the first part of the request.

As I understood, `.withInteropValidators` sets the number of validators keys owned by the node under testing, while `.withInteropNumberOfValidators` sets the total number of validators in the entire network (this is why it was set to 8 in the AT before the change), but reading your comment it seems to be the contrary.

If my first interpretation is correct, the final code should look like this:

```java
.withInteropNumberOfValidators(TOTAL_VALIDATOR_COUNT)
.withInteropValidators(0, ACTIVE_VALIDATOR_COUNT)
```

Is it right?

ajsutton: Yes your interpretation looks right to me - sorry I think I got things backwards when talking about the renaming.

Neurone: Hi @ajsutton, I finally fixed the PR. Tests pass; it should be all fine now.

ajsutton (approved):
LGTM. Thanks!

PR #5320, opened 2022-04-09
Allow deserialized zero length byte arrays

jtraglia:
## PR Description

When looking at REST API object deserializers, I noticed that this method would return null when given an empty array string (`"[]"`). I do not believe this is the intended behavior. Fixing just in case this deserializer is used elsewhere in the future.

Thankfully, this is not used in any way that can be remotely triggered. `ByteArrayDeserializer` is only used to parse the `previous_epoch_participation` and `current_epoch_participation` fields in `BeaconStateAltair`. This class is part of `GetStateResponseV2` which cannot be crafted and sent to the REST API handler; it is only deserialized in [`fetchHeadState`](https://github.com/ConsenSys/teku/blob/30b644e9bf250c9cd62affd484a1db46c509059d/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java#L433-L443).

## Documentation

## Changelog


rolfyone (approved):
It looks like this was added for epoch participation in BeaconStateAltair. Either should be ok, but I think this is probably a better idea in case it's used elsewhere in future.

LGTM.

PR #5321, opened 2022-04-11
Use milliseconds for calculating slot intervals

### LINKED ISSUE #5297: Use milliseconds for calculating slot intervals
### Description

Currently the validator client uses seconds to calculate slot intervals. This will cause an issue when `SECONDS_PER_SLOT` is not exactly divisible by 3 (e.g. 5 for Gnosis Beacon Chain). The problem is that the decimal points will be stripped and attestation creation and aggregation may be skipped, because it is expected to start earlier than required. 

`TimeBasedEventAdapter` should be modified to use milliseconds instead of seconds for the following scheduled events:

```
taskScheduler.scheduleRepeatingEvent(                                                       
    nextSlotStartTime.plus(oneThirdSlotSeconds),                                            
    secondsPerSlot,                                                                         
    this::onAttestationCreationDue);                                                        
taskScheduler.scheduleRepeatingEvent(                                                       
    nextSlotStartTime.plus(twoThirdSlotSeconds), secondsPerSlot, this::onAggregationDue); 
```

Also other places where slot interval calculation should be modified are:

`Duty Metrics` -> `calculateExpectedAttestationTimeInMillis`
`Fork Choice` -> `isBeforeAttestingInterval`
`Slot Processor` -> `isSlotAttestationDue`

StefanBratanov:
## PR Description
Modified `DutyMetrics` to calculate the expected attestation time in milliseconds correctly
Modified `ForkChoice` to use milliseconds instead of seconds when checking if the current time into the slot is before the attesting interval.
Modified `SlotProcessor` to use milliseconds instead of seconds when calculating 1/3 and 2/3 of slot.
Changed visibility of the main `RepeatingTaskScheduler` constructor to private

## Fixed Issue(s)
fixes #5297 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
+++ b/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
@@ -154,6 +154,10 @@ public final class DataStructureUtil {
     return new Random(nextSeed()).nextLong();
   }
 
+  public long randomLong(final long bound) {
+    return new Random(nextSeed()).longs(0, bound).findFirst().orElse(0L);
+  }
+
   public int randomPositiveInt() {
     return randomInt(Integer.MAX_VALUE);
   }
@@ -166,6 +170,10 @@ public final class DataStructureUtil {
 
   public UInt64 randomUInt64() {
     return UInt64.fromLongBits(randomLong());
+  }
+
+  public UInt64 randomUInt64(final long bound) {
+    return UInt64.fromLongBits(randomLong(bound));
   }
 
   public UInt256 randomUInt256() {
--- a/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
+++ b/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
@@ -154,7 +154,7 @@ public final class DataStructureUtil {
     return new Random(nextSeed()).nextLong();
   }
 
-  public long randomLong(final long bound) {
+  public long randomPositiveLong(final long bound) {
     return new Random(nextSeed()).longs(0, bound).findFirst().orElse(0L);
   }
 
@@ -173,7 +173,7 @@ public final class DataStructureUtil {
   }
 
   public UInt64 randomUInt64(final long bound) {
-    return UInt64.fromLongBits(randomLong(bound));
+    return UInt64.fromLongBits(randomPositiveLong(bound));
   }
 
   public UInt256 randomUInt256() {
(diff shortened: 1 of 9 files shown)


ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -154,6 +154,10 @@ public long randomLong() {
     return new Random(nextSeed()).nextLong();
   }
 
+  public long randomLong(final long bound) {
```
zilm13: I'd call it like the next method: `randomPositiveLong`

StefanBratanov: makes sense, fixed

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -306,9 +307,8 @@ private BlockImportResult importBlockAndState(
       final int secondsPerSlot = spec.getSecondsPerSlot(block.getSlot());
       final UInt64 timeIntoSlot =
           transaction.getTime().minus(transaction.getGenesisTime()).mod(secondsPerSlot);
```
ajsutton, line 309: hmm, I don't think we're being fair here.  The idea is that if the block is received prior to the point where attestations are due it receives an extra "proposer boost" in scoring but doesn't get it if it's received after the attestations are due (thus late blocks are less likely to wind up on the canonical chain).

With this version we're rounding (or truncating) the block received time to the second but we really want to get the received time in mills which will bias them either to get or miss the boost more often.  It's not an issue for anything except Gnosis because other network's slot time divides by 3 evenly.

I suspect we're going to have to move to having `Store.time` (which is what `transaction.getTime()` is getting here) track milliseconds rather than seconds but I think that should be a separate PR. This one is fine as-is since it doesn't make things any worse (doing the comparison on seconds will bias the boost on gnosis anyway).

I'd suggest we keep `getTime()` returning seconds (maybe rename to `getTimeSeconds()`) and add a new `getTimeMillis()` method so we're very clear about when we're using seconds vs millis.  The key pieces of code for that I think will start from `BeaconChainController.onTick` which is getting the time as seconds through to `StoreTransaction.setTime`. Work backwards by finding callers of `setTime` and you'll find the path easily enough I think.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/SlotProcessor.java
```diff
@@ -101,7 +103,7 @@ public void setCurrentSlot(final UInt64 slot) {
 
   public void onTick(final UInt64 currentTime) {
     final UInt64 genesisTime = recentChainData.getGenesisTime();
-    if (currentTime.compareTo(genesisTime) < 0) {
+    if (currentTime.isLessThan(genesisTime)) {
```
ajsutton, line 106: Nice. These methods are always so much easier to read than compareTo so good to switch over to them as we go.

ajsutton (approved):
LGTM. Just some follow up I think we'll need around the proposer boost handling in ForkChoice.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
StefanBratanov: Thank you, will open another PR for the recommended changes.

PR #5326, opened 2022-04-11
Fix valid byte range in json deserializer

jtraglia:
## PR Description

The json deserializer was allowing `0 <= b <= 256` when it should only allow `0 <= b < 256`.

When deserializing `"256"` it would result in `0` without error.

<details closed>
<summary>Test Program</summary>

```java
import static tech.pegasys.teku.infrastructure.json.types.CoreTypes.BYTE_TYPE;

import com.fasterxml.jackson.core.JsonParser;
import com.fasterxml.jackson.databind.ObjectMapper;
import java.io.IOException;

public class Test
{
  static JsonParser prepareDeserializationContext(String serializedValue) throws IOException {
    String json = String.format("{\"value\":%s}", serializedValue);
    JsonParser parser = new ObjectMapper().getFactory().createParser(json);
    parser.nextToken();
    parser.nextToken();
    parser.nextToken();
    return parser;
  }

  public static void main(String[] args) throws IOException {
    JsonParser parser = prepareDeserializationContext("256");

    // Results in value of zero.
    System.out.println(BYTE_TYPE.deserialize(parser));
    // java.lang.NumberFormatException: Value 256 out of range from input 256
    System.out.println(BYTE_TYPE.deserializeFromString("256"));
  }
}
```

</details>

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5333, opened 2022-04-12
Fix some JSON deserializer problems

jtraglia:
## PR Description

* If there was a nested array, the parser would stop when it reached the internal `JsonToken.END_ARRAY` and would not parse the remaining values.

* `StringBasedPrimitiveTypeDefinition` would parse non-scalar values without throwing an exception. This is because [`jsonParser.getValueAsString()`](https://fasterxml.github.io/jackson-core/javadoc/2.9/com/fasterxml/jackson/core/JsonParser.html#getValueAsString--) will return null on problems; we expected this to throw an exception.

## Documentation

## Changelog


jtraglia:
Note, I would have liked to use `JsonParseException` instead of `MismatchedInputException` but I was having issues with its constructor. I think they are all protected.

jtraglia:
Fixed a similar issue in `DeserializableStringMapTypeDefinition`. Should we add null checks for some `JsonParser.getValueAsString()` calls? There could be some unexpected exceptions if it returns null. I'm not sure how we should handle that situation though. Thoughts?

Edit: Ah no need to handle these. The catch-all handler will deal with it:

https://github.com/ConsenSys/teku/blob/51d3b5bd097c1d46c49b3baff3cc75922ede1f0d/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java#L219-L226

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableStringMapTypeDefinition.java
```diff
@@ -49,6 +50,10 @@ public Map<String, String> deserialize(final JsonParser p) throws IOException {
     for (; t == JsonToken.FIELD_NAME; t = p.nextToken()) {
       final String fieldName = p.getCurrentName();
       p.nextToken();
+      if (!p.getCurrentToken().isScalarValue()) {
+        throw MismatchedInputException.from(
```
ajsutton, line 54: MismatchedInputException actually looks like the right type for this - it's a more explicit subtype of JsonProcessingException.

It probably doesn't matter but I'd be tempted to pass `String.class` as the expected class here.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/StringBasedPrimitiveTypeDefinition.java
```diff
@@ -59,6 +60,10 @@ public Optional<String> getTypeName() {
 
   @Override
   public T deserialize(final JsonParser parser) throws IOException {
+    if (!parser.getCurrentToken().isScalarValue()) {
+      throw MismatchedInputException.from(
+          parser, (Class<?>) null, "Expected scalar value but got " + parser.getCurrentToken());
```
ajsutton: May as well use `String.class` here as well.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableArrayTypeDefinition.java
```diff
@@ -54,6 +54,9 @@ public CollectionT deserialize(final JsonParser parser) throws IOException {
     final List<ItemT> result = new ArrayList<>();
     while (parser.nextToken() != JsonToken.END_ARRAY) {
       result.add(itemType.deserialize(parser));
+      if (parser.getCurrentToken().isStructStart()) {
+        parser.skipChildren();
+      }
```
ajsutton: hmm, this doesn't seem right - I'd have expected that if we get a nested array/object that we didn't expect we should fail, not just skip it.

All the tests seem to pass even without it so I'm not sure what case it's trying to handle.

ajsutton (commented):
Looking good for the most part, just not sure about the skipChildren call. I think having the types inside the array fail if they get an unexpected object rather than returning null and not consuming the tokens is probably the solution but you may have seen something I've missed.

> Fixed a similar issue in DeserializableStringMapTypeDefinition. Should we add null checks for some JsonParser.getValueAsString() calls? There could be some unexpected exceptions if it returns null. I'm not sure how we should handle that situation though. Thoughts?

I think the type definitions need to return null if that's what we got from the JSON and we'll have to watch to handle that in the actual API handlers - worst case as you say the catch all handler will just result in a 500 error.  

One place we could put a null check is in `RequiredDeserializableFieldDefinition` which is used when deserialising object required fields - would be quite reasonable to consider the JSON invalid if a required field is null.  If null is allowed we should declare the field as an optional one.

jtraglia:
@ajsutton, so you're right to think that this looks weird, but I believe it is necessary to support arrays of more complex types even though they aren't used right now; future proofing. Imagine if you for some reason needed an array of arrays. Without `skipChildren` it will deserialize the array and then deserialize each item in the array. And it would stop parsing at the end of the first array.

For example, given this json: `[[a,b,c], [d,e,f], [h,i,j]]`, ...

Without `skipChildren` it will parse it like this:

* `[`
  *  `[` - see array, parse array, go to next token: `a`.
      * `a` - parse, add to results.
      * `b` - parse, add to results.
      * `c` - parse, add to results.
  * `]` - reached `JsonToken.END_ARRAY`, stop parsing.

Resulting in only `[[a,b,c], a, b, c]`.

With `skipChildren` it will parse like this:

* `[`
  *  `[` - see array, parse array, jump to end via `skipChildren`.
      * `a`
      * `b`
      * `c`
  * `]` - now here, `nextToken` will beginning of next array.
  *  `[` - see array, parse array, jump to end via `skipChildren`.
      * `d`
      * `e`
      * `f`
  * `]` - now here, `nextToken` will beginning of next array.
  *  `[` - see array, parse array, jump to end via `skipChildren`.
      * `h`
      * `i`
      * `j`
  * `]` - now here, `nextToken` will be the end.
* `]` - reached `JsonToken.END_ARRAY`, stop parsing.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableStringMapTypeDefinition.java
jtraglia: Okay cool. I will give that a shot. This was just a copy paste of:

https://github.com/ConsenSys/teku/blob/a4620f9598db35252261130d9fa2fe925b52ea3f/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableArrayTypeDefinition.java#L51-L52

jtraglia:
Welp, I added a list of lists test to help explain myself, but it seems that I am wrong. I commented out the `skipChildren` call and it behaves properly without it. So I'll remove it. I'm pretty sure I understand why now, but it's not worth getting into.

Sorry, you were correct. Thank you for pointing this out.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableStringMapTypeDefinition.java
ajsutton: Ah yeah, in that case it's less clear what the class we need is (it's an array of something by TypeDefinition doesn't really know).  Whereas in this case we know we want a string.  It's a minor detail though - not sure it even turns up in the error message anywhere.

ajsutton (approved):
LGTM. Thanks for this, it was a good catch.

PR #5334, opened 2022-04-12
[Merge] always consume `ForkChoiceUpdatedResult` from `ForkChoiceUpdated`

### LINKED ISSUE #5332: [Merge] `ForkChoiceUpdatedResult` is ignored when `ForkChoiceUpdated` is called for block proposal
When running in interop mode, where all blocks are produced locally, due to this bug the head never turns VALID and remains optimistic forever, and we refuse to produce the blocks since builds on top of an optimistic block:
```
Unable to produce block at slot 306 because parent has optimistically validated payload
```

tbenr:
## PR Description

This makes sure that `ForkChoice` will always apply necessary logic when receiving a `ForkChoiceUpdatedResult` calling `engine_forkChoiceUpdated` in `ForkChoiceNotifier`.

## Fixed Issue(s)
fixes #5332 

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -450,19 +470,9 @@ private void notifyForkChoiceUpdatedAndOptimisticSyncingChanged() {
             .getForkChoiceState(
                 recentChainData.getJustifiedCheckpoint().orElseThrow(),
                 recentChainData.getFinalizedCheckpoint().orElseThrow());
-    UInt64 latestFinalizedBlockSlot = recentChainData.getStore().getLatestFinalizedBlockSlot();
```
tbenr, line 453: @ajsutton I'm a bit uncomfortable about moving this later on notification reception. Can't currently figure out if we need to ensure consistency with `ForkChoiceState`.

tbenr: to be supersafe I could make `ForkChoiceUpdatedResultNotification` to be a context created here, in advance, that brings with it `ForkChoiceState` and `latestFinalizedBlockSlot` but I suspect will require some more complex handling in `ForkChoiceNotifier`. Probably needs to be tracked with `ForkChoiceUpdateData`.

ajsutton: We never wound up using this value for anything and I don't think we even wound up writing it to disk. The optimistic sync spec moved on from needing it, so we probably should just remove it.

I'd leave it as you have it now and I'll follow up with a PR to remove it so you can get back to MEV stuff.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifier.java
```diff
@@ -38,4 +37,25 @@ SafeFuture<Optional<ForkChoiceUpdatedResult>> onForkChoiceUpdated(
...
+
+  boolean unsubscribeFromForkChoiceUpdatedResult(long subscriberId);
+
+  interface ForkChoiceUpdatedResultSubscriber {
+    void onForkChoiceUpdatedResult(
+        final ForkChoiceUpdatedResultNotification forkChoiceUpdatedResultNotification);
+
+    class ForkChoiceUpdatedResultNotification {
+      final ForkChoiceState forkChoiceState;
+      final SafeFuture<Optional<ForkChoiceUpdatedResult>> forkChoiceUpdatedResult;
```
ajsutton: To follow our usual coding convention we should make these private and use getters for them. Slightly more code but probably worth it for consistency.

I'd be tempted to split this class out into its own file as well rather than having classes nested three deep.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifierImpl.java
```diff
@@ -227,9 +240,11 @@ private void internalAttestationsDue(final UInt64 slot) {
...
 
   private void sendForkChoiceUpdated() {
-    forkChoiceUpdateData
-        .send(executionEngineChannel)
-        .finish(error -> LOG.error("forkChoiceUpdated notification failed", error));
+    subscribers.deliver(
+        ForkChoiceUpdatedResultSubscriber::onForkChoiceUpdatedResult,
+        new ForkChoiceUpdatedResultNotification(
+            forkChoiceUpdateData.getForkChoiceState(),
+            forkChoiceUpdateData.send(executionEngineChannel)));
```
ajsutton: We should move the call to `send` out of here as it's a pretty important call with side effects and is easily missed in this setup as it looks like we're just notifying subscribers.  Just extracting it to a local variable should be enough.

ethereum/statetransition/src/testFixtures/java/tech/pegasys/teku/statetransition/forkchoice/StubForkChoiceNotifier.java
```diff
@@ -18,20 +18,44 @@
...
 
+  private final Subscribers<ForkChoiceUpdatedResultSubscriber> subscribers =
+      Subscribers.create(true);
+  private SafeFuture<Optional<ForkChoiceUpdatedResult>> forkChoiceUpdatedResultNotification =
+      SafeFuture.completedFuture(Optional.empty());
+
+  public void setForkChoiceUpdatedResul(
+      final Optional<ForkChoiceUpdatedResult> forkChoiceUpdatedResul) {
+    forkChoiceUpdatedResultNotification = SafeFuture.completedFuture(forkChoiceUpdatedResul);
+  }
```
ajsutton:
```suggestion
  public void setForkChoiceUpdatedResult(
      final Optional<ForkChoiceUpdatedResult> forkChoiceUpdatedResult) {
    forkChoiceUpdatedResultNotification = SafeFuture.completedFuture(forkChoiceUpdatedResult);
  }
```

ajsutton (approved):
LGTM. Just some minor details.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
tbenr: yeah was suspecting that :)
I'll open an issue to track that and move on.

PR #5336, opened 2022-04-13
Refactor getStateRoot

rolfyone:

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/MigratingEndpointAdapter.java
```diff
@@ -35,4 +42,18 @@ protected void adapt(final Context ctx) throws Exception {
     final RestApiRequest request = new RestApiRequest(ctx, getMetadata());
     handleRequest(request);
   }
+
+  protected void handleStateByIdRequest(
```
ajsutton: Not a fan of putting this in `MigratingEndpointAdapter` since we should be able to remove it when the migration is complete.

I do really like how we are likely to be able to write a heap of endpoints by just serialising this returned state differently. Really neat.

So maybe we just need to put this method in a subclass of `MigratingEndpointAdapter` that all the endpoints that return data just from the state extend, and this method could be the `handleRequest` implementation, the subclass just provides the `EndpointMetadata` which defines the serialisation.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/MigratingEndpointAdapter.java
```diff
@@ -35,4 +42,18 @@ protected void adapt(final Context ctx) throws Exception {
...
+    final SafeFuture<Optional<StateAndMetaData>> future =
+        chainDataProvider.getBeaconStateAndMetadata(request.getPathParameter(PARAMETER_STATE_ID));
+    request.respondAsync(
+        future.thenApply(
+            maybeStateAndMetadata -> {
+              if (maybeStateAndMetadata.isEmpty()) {
+                return AsyncApiResponse.respondNotFound();
+              }
+              return AsyncApiResponse.respondOk(maybeStateAndMetadata.get());
+            }));
```
ajsutton:
```suggestion
        future.thenApply(
            maybeStateAndMetadata ->
                maybeStateAndMetadata
                    .map(AsyncApiResponse::respondOk)
                    .orElseGet(AsyncApiResponse::respondNotFound)));
```

:)

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRootTest.java
```diff
@@ -19,32 +19,58 @@
...
-import org.apache.tuweni.bytes.Bytes32;
 import org.assertj.core.api.AssertionsForClassTypes;
 import org.junit.jupiter.api.Test;
 import tech.pegasys.teku.beaconrestapi.AbstractBeaconHandlerTest;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.restapi.endpoints.RestApiRequest;
+import tech.pegasys.teku.spec.datastructures.metadata.StateAndMetaData;
 import tech.pegasys.teku.spec.util.DataStructureUtil;
 
 public class GetStateRootTest extends AbstractBeaconHandlerTest {
```
ajsutton, line 32: For another PR, but we will likely be able o share these tests across the various endpoints that just serialise the state differently as well.

ajsutton (approved):
LGTM. Just need to avoid piling functionality into MigratingEndpointAdapter I think.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/MigratingEndpointAdapter.java
rolfyone: yep that'd be the other option, was just thinking this would get most of the logic common at least... it could equally be in a utility class, doesnt have to be on a handler, just makes it pretty... will have a play

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/MigratingEndpointAdapter.java
rolfyone: v. nice

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRootTest.java
rolfyone: I agree completely... was just thinking if i have an abstract base class that maybe tests can be done on that

PR #5337, opened 2022-04-13
Apply execution engine timeouts

### LINKED ISSUE #5302: [Merge] Update execution engine Api timeouts
### Description
As per https://github.com/ethereum/execution-apis/pull/207 set execution engine call timeouts to 120s for engine endpoints with block execution semantics and 12s for calls without execution semantics.

ajsutton:
## PR Description
Apply timeouts to execution engine requests as per the latest spec.

Web3j doesn't allow setting the actual http timeout separate for different request types and web sockets and IPC don't have any concept of request timeouts at all. So we handle the timeout via the future and set the HTTP read timeout to the longest timeout we need to ensure it doesn't abort the connection before the timeout is reached.

## Fixed Issue(s)
fixes #5302 

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/build.gradle
+++ b/ethereum/executionlayer/build.gradle
@@ -17,6 +17,7 @@ dependencies {
   runtimeOnly 'io.jsonwebtoken:jjwt-jackson'
 
   testImplementation testFixtures(project(':ethereum:spec'))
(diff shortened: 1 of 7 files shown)


rolfyone (approved):
LGTM

PR #5340, opened 2022-04-13
Make GetStateFork use AbstractGetSimpleDataFromState

rolfyone:

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/AbstractGetSimpleDataFromState.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/AbstractGetSimpleDataFromState.java
@@ -14,6 +14,7 @@
 package tech.pegasys.teku.beaconrestapi.handlers.v1.beacon;
 
 import static tech.pegasys.teku.beaconrestapi.BeaconRestApiTypes.PARAMETER_STATE_ID;
+import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_BAD_REQUEST;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/AbstractGetSimpleDataFromState.java
```diff
@@ -39,6 +40,10 @@ public AbstractGetSimpleDataFromState(
   public void handleRequest(RestApiRequest request) throws JsonProcessingException {
     final SafeFuture<Optional<StateAndMetaData>> future =
         chainDataProvider.getBeaconStateAndMetadata(request.getPathParameter(PARAMETER_STATE_ID));
+    if (future == null) {
+      request.respondError(SC_BAD_REQUEST, "Bad Request - check state_id is correct");
+      return;
+    }
     request.respondAsync(
         future.thenApply(
```
rolfyone: i wonder if i put bad request on the end of this future processing as an ExceptionallyCompose, but then I guess that's potentially going to hide issues...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/AbstractGetSimpleDataFromState.java
```diff
@@ -39,6 +40,10 @@ public AbstractGetSimpleDataFromState(
   public void handleRequest(RestApiRequest request) throws JsonProcessingException {
     final SafeFuture<Optional<StateAndMetaData>> future =
         chainDataProvider.getBeaconStateAndMetadata(request.getPathParameter(PARAMETER_STATE_ID));
+    if (future == null) {
```
ajsutton: It's really awkward that future could wind up being null here. But I don't see how this would ever be null looking at the code - it looks like chainDataProvider would throw `BadRequestException` and that should then be handled but the default exception handlers to respond with bad request. Adding a `.exceptionally` block that checks if the root cause of the exception is `BadRequstException` and returns a response with the custom "check state_id is correct" message would be fine (just rethrow any other exceptions).

ajsutton (commented):
Looks good except I'm a bit confused about the future being null case.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/AbstractGetSimpleDataFromState.java
rolfyone: test was wrong - an exception gets handled outside the handler fn, so this null check has been removed.

ajsutton (approved):
LGTM.

PR #5342, opened 2022-04-13
Allow fork choice store to save and retrieve time in milliseconds

### LINKED ISSUE #5297: Use milliseconds for calculating slot intervals

StefanBratanov:
## PR Description
Modified `ReadOnlyStore` and `MutableStore` to allow to store and retrieve time in milliseconds.

## Fixed Issue(s)
will help to resolve #5297 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/ReadOnlyStore.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/ReadOnlyStore.java
@@ -13,6 +13,8 @@
 
 package tech.pegasys.teku.spec.datastructures.forkchoice;
 
+import static tech.pegasys.teku.infrastructure.time.TimeUtilities.millisToSeconds;
+
 import java.util.Collection;
 import java.util.Optional;
 import org.apache.tuweni.bytes.Bytes32;
@@ -31,7 +33,9 @@ import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
 
 public interface ReadOnlyStore {
 
-  UInt64 getTime();
+  default UInt64 getTimeSeconds() {
+    return millisToSeconds(getTimeMillis());
+  }
 
   /**
    * Returns time in milliseconds to allow for more fine-grained time calculations
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/ReadOnlyStore.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/ReadOnlyStore.java
@@ -33,6 +33,13 @@ public interface ReadOnlyStore {
 
   UInt64 getTime();
 
+  /**
+   * Returns time in milliseconds to allow for more fine-grained time calculations
+   *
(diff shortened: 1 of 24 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/ReadOnlyStore.java
```diff
@@ -33,6 +33,13 @@ public interface ReadOnlyStore {
 
   UInt64 getTime();
```
ajsutton: I'd suggest renaming this (and the setter in MutableStore) to `getTimeSeconds` so that we're always being explicit about the units to avoid any confusion.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/datastructures/forkchoice/TestStoreImpl.java
```diff
@@ -257,6 +266,11 @@ public void setTime(final UInt64 time) {
     this.time = time;
   }
 
+  @Override
+  public void setTimeMillis(final UInt64 time) {
+    this.timeMillis = time;
+  }
```
ajsutton: I worry that having the time stored in both seconds and millis will get out of sync and potentially be very confusing.  So I'd probably just store the time in millis internally.  `getTimeSeconds` and `setTimeSeconds` would then do the conversion between seconds and millis.

ajsutton (commented):
I'm really nervous about storing the time in two variables when it may be different. Happy to talk it through this evening to work out if it's worth doing this as a temporary step though.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/datastructures/forkchoice/TestStoreImpl.java
StefanBratanov: I agree, now that i am making the next changes on top, that bit got me worried as well, will resolve it in this PR first

ajsutton (approved):
LGTM.

PR #5343, opened 2022-04-13
Verbose duties logging

### LINKED ISSUE #4385: Write attestation schedule for epoch to console when received
### Description

As an operator of validating infrastructure, I would like to be able to see when my Validators are scheduled to attest / propose at the start of an epoch. This allows me to better time any restarts / updates to avoid missing attestions / proposals.

For example, `prysm` provides:

a) a full schedule of which slots active validators have been selected to attest to / propose - provided at the beginning of the epoch
b) a countdown timer until next scheduled activity - provided after the previous scheduled activity (e.g. 68 seconds until next scheduled activity... 56 seconds until next scheduled activity... 44 seconds until next scheduled activity and so on).

### Acceptance Criteria

When I can observe scheduled attestations / proposals written to the console at the beginning of the epoch.

### Additional Information

This is a feature I use heavily when running Validators on `prysm`.

zilm13:
## PR Description
Logs attestation / aggregation / block proposal duties at the beginning of the epoch with seconds countdown on every slot until duty slot. Should be helpful for single / 2 validator nodes. Too verbose for many validators. To reduce complexity I've removed separate aggregation calculation and just added extra `1 / SLOT_INTERVALS` to countdown (attestation goes at 1/3, aggregation on 2/3).
Feature is enabled by flag `--log-validator-duties-verbose-enabled`

It would be great to add sync committee assignment info too, but it's better to create separate PR for it as this one is not small. Something like "No sync committee duties for the next 10 hours" or "Sync committee duties are expected to continue for the next 2 hours".

## Fixed Issue(s)

Should fix #4385 
## Documentation

## Changelog


ajsutton (changes requested):
I don't think this approach is going to work - it's too expensive to be requesting the duties from the beacon node again and the logging version is missing a lot of details about when to invalidate the duties such as checking the dependentRoot, if events are missed etc. So we're going to wind up with our logging not always matching the actual duties we perform.

To get it right, we'd have to essentially duplicate all the logic we already have for scheduling duties.  We probably need to just shelve this for now as it won't be a quick win - when we come back to it we'll have to work through how to extract the info from the existing scheduled duties.

zilm13:
@ajsutton I don't think that extra cost is a big issue here as feature is dedicated for 1-2 validators users with 2 extra queries per epoch. But overall I'm on the same side. Even If I do what I missed with dependent roots etc, it's the same logic as we have in Scheduler <->Loaders and as its not reused it means whenever it changes we are going to have issues on duplicate side.

PR #5345, opened 2022-04-13
[Merge] provide a mocked execution engine able to emulate transition via TTD or TBH and supporting standard and mev_boost flow

### LINKED ISSUE #5331: [Merge] Provide stub execution engines
### Description
To make testing Teku easier, add a hidden CLI option that disables the real execution engine API and instead uses one that either is always syncing or always returns valid (based on what the CLI option is set to).  While this is obviously a bad idea in production, it makes it a lot easier to test teku without having to setup an EL.

We should log very noisy warnings when this option is being used to ensure the user knows they are using insecure options.

tbenr:
I promoted the textFixture `StubExecutionEngineChannel` to production code as `ExecutionEngineChannelStub`.
Can be activated by providing `--ee-endpoint stub` option

- Detects bellatrix activation by looking at the first call to `getPowBlock` or `getPowChainHead`. If block production APIs are called before, it assumes transition already happened
- Prepares terminal blocks via TTD or TBH, consider values from exchange configuration APIs. if not available will use Spec.
- supports terminal block validation (node restart before terminal block finalized)
- keeps track of payloads to support mev boost flow
- keeps track of payloads to be able to respond to `getPowChainHead`
- all payloads are considered VALID

## Fixed Issue(s)
fixes #5331

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionEngineChannelImpl.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionEngineChannelImpl.java
@@ -44,6 +44,7 @@ import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayload;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadHeader;
 import tech.pegasys.teku.spec.datastructures.execution.PowBlock;
 import tech.pegasys.teku.spec.executionengine.ExecutionEngineChannel;
+import tech.pegasys.teku.spec.executionengine.ExecutionEngineChannelMock;
 import tech.pegasys.teku.spec.executionengine.ForkChoiceState;
 import tech.pegasys.teku.spec.executionengine.PayloadAttributes;
 import tech.pegasys.teku.spec.executionengine.PayloadStatus;
@@ -56,7 +57,7 @@ public class ExecutionEngineChannelImpl implements ExecutionEngineChannel {
   private final ExecutionEngineClient executionEngineClient;
   private final Spec spec;
 
-  public static ExecutionEngineChannelImpl create(
+  public static ExecutionEngineChannel create(
       final String eeEndpoint,
       final Spec spec,
       final TimeProvider timeProvider,
@@ -65,6 +66,9 @@ public class ExecutionEngineChannelImpl implements ExecutionEngineChannel {
       final Path beaconDataDirectory) {
     checkNotNull(eeEndpoint);
     checkNotNull(version);
+    if (eeEndpoint.equalsIgnoreCase(MOCK_ENDPOINT_IDENTIFIER)) {
+      return new ExecutionEngineChannelMock(spec, timeProvider, true);
+    }
     return new ExecutionEngineChannelImpl(
         createEngineClient(eeEndpoint, timeProvider, version, jwtSecretFile, beaconDataDirectory),
         spec);
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionEngineChannelImpl.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionEngineChannelImpl.java
@@ -15,6 +15,7 @@ package tech.pegasys.teku.ethereum.executionlayer;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static tech.pegasys.teku.infrastructure.logging.EventLogger.EVENT_LOG;
 
 import java.nio.file.Path;
 import java.util.Optional;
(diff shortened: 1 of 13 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionEngineChannelMock.java
```diff
@@ -231,4 +253,38 @@ private HeadAndAttributes(Bytes32 head, PayloadAttributes attributes) {
...
+    final SpecConfigBellatrix specConfigBellatrix =
+        specVersion.getConfig().toVersionBellatrix().orElseThrow();
+
+    final UInt64 bellatrixActivationSlot =
+        spec.computeStartSlotAtEpoch(specConfigBellatrix.getBellatrixForkEpoch());
+
+    // assumes genesis happen immediately at min genesis time reached
+    final UInt64 bellatrixActivationTime =
+        spec.getSlotStartTime(
+            bellatrixActivationSlot, spec.getGenesisSpecConfig().getMinGenesisTime());
```
tbenr: @ajsutton any hint on how to infer genesis time in interop mode without having access to the State?

tbenr: Maybe its easier to implement the transition via terminal block hash

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionEngineChannelImpl.java
```diff
@@ -65,6 +66,9 @@ public static ExecutionEngineChannelImpl create(
       final Path beaconDataDirectory) {
     checkNotNull(eeEndpoint);
     checkNotNull(version);
+    if (eeEndpoint.equalsIgnoreCase(MOCK_ENDPOINT_IDENTIFIER)) {
+      return new ExecutionEngineChannelMock(spec, timeProvider, true);
```
ajsutton: Can we log a really big warning here to the console please?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionEngineChannelMock.java
```diff
@@ -13,40 +13,71 @@
...
+import tech.pegasys.teku.spec.config.SpecConfigBellatrix;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayload;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadHeader;
 import tech.pegasys.teku.spec.datastructures.execution.PowBlock;
 import tech.pegasys.teku.spec.schemas.SchemaDefinitionsBellatrix;
 
-public class StubExecutionEngineChannel implements ExecutionEngineChannel {
-
+public class ExecutionEngineChannelMock implements ExecutionEngineChannel {
```
ajsutton: I don't think I'd call this a mock as it is still technically a stub.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionEngineChannelMock.java
ajsutton: Yeah you'll need the state realistically - we may want to use the mock with things other than interop mode anyway.  Hacky work around would be to just have the transition time hard coded.  Or just have a delay after the current time before the transition block is found (the timestamp just needs to be before the transition slot and teku should handle that already because the PoW chain timestamps aren't lined up with the beacon chain at all anyway).

ajsutton:
I like the idea - we can probably gradually flesh it out further over time as well. eg it could keep track of the head block based on fcu calls etc so it works quite a lot like a real EL except considers everything valid and makes up the transition block

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionEngineChannelImpl.java
tbenr: SURE

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionEngineChannelMock.java
tbenr: ahh.. yeah.. right... I actually don't even need to know when bellatrix activates.
I can use the first `getPowChainHead` call as the bellatrix activation signal. And i can apply a delay from there.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionengine/ExecutionEngineChannelMock.java
tbenr: Just read the definition of Mock for the first time in my life.

ajsutton (approved):
LGTM.

Issue #5243, opened 2022-03-28
Optimise Epoch Transition


ajsutton:
### Description
With the increase in number of validators the epoch transition is becoming more expensive. Benchmark the transition on MainNet sized states and optimise.

ajsutton:
Did a round of optimisation here with some good improvements.  Have logged #5448 as a potential follow up.

PR #5346, opened 2022-04-14
Optimise epoch transition

ajsutton:
## PR Description
Few optimisations in this:

* Fix the EpochTransitionBenchmark and include calculating the hashTreeRoot of the resulting state since that's a fairly big part of the cost
* Use the BeaconState invariant fields to get the field index for slot (and genesis validator root and genesis time). Avoids a map lookup to find the index which is a big win when accessing the state slot in a loop
* Compute baseReward from the validator statuses instead of state. Avoids a bunch of SSZ field accesses for each validator and caches some values that are used for each calculation more efficiently.
* Avoid temporary list creation when updating primitive collections. Avoids needing to iterate through the Stream to create the list, then iterating again to actually do the processing.

Before changes:
EpochTransitionBenchmark.epochTransition              32768  thrpt   10  21.627 ± 0.875  ops/s

After changes:
EpochTransitionBenchmark.epochTransition              32768  thrpt   10  26.840 ± 0.720  ops/s

Before:
image

After:
image

## Fixed Issue(s)
#5243 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ For information on changes in released versions of Teku, see the [releases page]
 
 ### Additions and Improvements
 - Improved performance when regenerating non-finalized states that had to be dropped from memory.
-- Performance optimizations for Gnosis beacon chain
+- Performance optimizations for Gnosis beacon chain.
+- Improved performance when processing epoch transitions.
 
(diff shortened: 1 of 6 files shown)


rolfyone (approved):
LGTM

PR #5349, opened 2022-04-14
Optimise replacing all items in SszList

ajsutton:
## PR Description
Optimise setting all items in an SszList.

EpochProcessorAltair needs to set all participation flags back to zero. Instead of iterating through every individual element we can rebuild the backing tree much more quickly using pre-packed leaf nodes.

Can hopefully tidy up the code some more and need to add some tests around it but shows the concept and seems to have quite a noticeable performance improvement.

Before:
EpochTransitionBenchmark.epochTransition              32768  thrpt   10  28.711 ± 0.293  ops/s

After:
EpochTransitionBenchmark.epochTransition              32768  thrpt   10  34.520 ± 0.551  ops/s

## Fixed Issue(s)
#5346 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/statetransition/epoch/EpochProcessorAltair.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/statetransition/epoch/EpochProcessorAltair.java
@@ -94,7 +94,6 @@ public class EpochProcessorAltair extends AbstractEpochProcessor {
     state.setPreviousEpochParticipation(state.getCurrentEpochParticipation());
 
     // Reset current epoch participation flags
-    state.getCurrentEpochParticipation().clear();
     state.getCurrentEpochParticipation().setAll(SszByte.ZERO, state.getValidators().size());
   }
 
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/SszMutablePrimitiveListImpl.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/SszMutablePrimitiveListImpl.java
@@ -17,6 +17,7 @@ import tech.pegasys.teku.infrastructure.ssz.SszPrimitive;
 import tech.pegasys.teku.infrastructure.ssz.cache.IntCache;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszMutablePrimitiveList;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszPrimitiveList;
+import tech.pegasys.teku.infrastructure.ssz.impl.AbstractSszComposite;
 import tech.pegasys.teku.infrastructure.ssz.primitive.SszUInt64;
 import tech.pegasys.teku.infrastructure.ssz.schema.collections.SszPrimitiveListSchema;
 import tech.pegasys.teku.infrastructure.ssz.tree.BranchNode;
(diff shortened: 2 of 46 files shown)


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszListSchema.java
```diff
@@ -75,6 +79,43 @@ protected TreeNode createTree(TreeNode dataNode, int length) {
     return BranchNode.create(dataNode, toLengthNode(length));
   }
 
+  @Override
+  public SszListT nCopies(final ElementDataT value, final int length) {
```
Nashatyrev, line 83: Should we check the `length` bounds here?

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/SszMutablePrimitiveListImpl.java
```diff
@@ -76,6 +77,19 @@ public void set(int index, SszElementT value) {
...
+  @SuppressWarnings("unchecked")
+  public void setAll(final SszElementT value, final int startIndex, final int endIndex) {
+    if (startIndex == 0 && endIndex >= size()) {
+      // We're replacing everything so just build a new backing tree rather that iterating
+      setImmutableBackingData(
+          (AbstractSszComposite<SszElementT>) getSchema().nCopies(value, endIndex));
+      cachedSize = endIndex;
+      return;
+    }
+    SszMutablePrimitiveList.super.setAll(value, startIndex, endIndex);
```
Nashatyrev, line 90:
```suggestion
    } esle {
        SszMutablePrimitiveList.super.setAll(value, startIndex, endIndex);
    }
```
nit: just a minor style suggestion

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszListSchema.java
```diff
@@ -75,6 +79,43 @@ protected TreeNode createTree(TreeNode dataNode, int length) {
...
+                  Bytes.concatenate(Collections.nCopies(partialNodeItemCount, singleItemData)));
+        }
+      }
+    }
+
+    List<TreeNode> leafNodes = Collections.nCopies(length / elementsPerChunk, fullLeafNode);
+    if (partialNode != null) {
+      leafNodes = new ArrayList<>(leafNodes);
+      leafNodes.add(partialNode);
+    }
```
Nashatyrev, line 113: After applying #5350 does it make sense to handle the logic separately in `SszPrimitiveListSchemaImpl`and `SszListSchemaImpl`?

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszListSchema.java
```diff
@@ -75,6 +79,43 @@ protected TreeNode createTree(TreeNode dataNode, int length) {
...
   }
 
+  @Override
+  public SszListT nCopies(final ElementDataT value, final int length) {
+    final int elementsPerChunk = getElementsPerChunk();
+    final TreeNode fullLeafNode;
+    TreeNode partialNode = null;
+    if (elementsPerChunk == 1) {
+      fullLeafNode = value.getBackingNode();
+    } else {
```
Nashatyrev, line 89: Make sure that bitlist is handled specifically. 
Makes sense just to check here this is not a bitlist and override the method in `SszBitlistSchemaImpl`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/statetransition/epoch/EpochProcessorAltair.java
```diff
@@ -94,7 +94,6 @@ public void processParticipationUpdates(final MutableBeaconState genericState) {
     state.setPreviousEpochParticipation(state.getCurrentEpochParticipation());
 
     // Reset current epoch participation flags
-    state.getCurrentEpochParticipation().clear();
     state.getCurrentEpochParticipation().setAll(SszByte.ZERO, state.getValidators().size());
```
Nashatyrev, line 97: not related nit: TBH I'd remove the `SszMutableCollection.setAll(SszElementT, int)` overload at all. Looks a bit misleading

Nashatyrev (commented):
Looks cool to me 👍 
Great optimization!

You would probably need to check `SszBitlist.setAll()` for the whole list length. It would probably fail with the current implementation.

And it's probably worth splitting `nCopy` functionality across subclasses as commented above

ajsutton:
Logged a ticket to follow up on this and get it finished off.  Going to close this for now - it can be reopened when we start looking at that ticket.

PR #5351, opened 2022-04-14
[MEV Boost] fix transition to merge

tbenr:
Blinded blocks pre-merge should not use standard `Default` value for `ExecutionPayloadHeaderSchema` but instead a block header that corresponds to the `Default` value for  `ExecutionPayloadSchema`.

`MiscHelpersBellatrix::isDefaultExecution` check needs to be updated accordingly.

the unblinder needs to provide the default `ExecutionPayload` if header matches the header of the default execution payload.

## Documentation

## Changelog


ajsutton:
I need to think about this a lot more to properly understand it but I'm wondering why we'd be trying to unblind a block with default execution payload (and/or why we wouldn't just hard code the unblinder to recognise that and return the default payload without sending to the request to any external service)

tbenr:
> I need to think about this a lot more to properly understand it but I'm wondering why we'd be trying to unblind a block with default execution payload (and/or why we wouldn't just hard code the unblinder to recognise that and return the default payload without sending to the request to any external service)

I did the change quite quickly but seemed to me quite natural to have that check in the OperationSelector. We deal with the defaults in there already, during block production. So we "produce" the default header in the block production selector and we "consume" it in the unblinder selector.

ajsutton (approved):
LGTM.

PR #5354, opened 2022-04-19
Bellatrix fork + External Signer: Send block_header for block sign request

usmansaleem:
## PR Description
For Bellatrix fork, send `block_header` instead of `block` for external signer block signing request (`BLOCK_V2`). `block_header` contains `body_root` instead of complete beacon block `body` which significantly reduces the traffic between Teku and external signer for block signing request. 

Either `block_header` or `block` can be used to calculate signing root by the external signer. For backward compatibility with PHASE0 and ALTAIR forks, only `block` is sent; for Bellatrix and later forks, only `block_header` is sent. 

## Fixed Issue(s)

https://github.com/ConsenSys/web3signer/issues/437

## Documentation

## Changelog


ethereum/core/src/main/java/tech/pegasys/teku/core/signatures/SigningRootUtil.java
```diff
@@ -41,14 +42,23 @@ public Bytes signingRootForRandaoReveal(final UInt64 epoch, final ForkInfo forkI
   }
 
   public Bytes signingRootForSignBlock(final BeaconBlock block, final ForkInfo forkInfo) {
-    final SpecVersion specVersion = spec.atSlot(block.getSlot());
+    return spec.computeSigningRoot(block, getDomainForSignBlock(block.getSlot(), forkInfo));
+  }
+
+  public Bytes signingRootForSignBlock(
```
ajsutton: Maybe call this `signingRootForSignBlockHeader`?

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/ExternalSigner.java
```diff
@@ -344,11 +344,11 @@ private BLSSignature getBlsSignature(
...
 
   @VisibleForTesting
-  static Supplier<String> slashableBlockMessage(final BeaconBlock block) {
+  static Supplier<String> slashableBlockMessage(final UInt64 slot) {
     return () ->
-        "External signed refused to sign block at slot "
-            + block.getSlot()
-            + " as it may violate a slashing condition";
+        String.format(
+            "External signed refused to sign block at slot %s as it may violate a slashing condition",
```
ajsutton:
```suggestion
            "External signer refused to sign block at slot %s as it may violate a slashing condition",
```

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/ExternalSignerBlockRequestProvider.java
```diff
@@ -43,12 +47,24 @@ public Map<String, Object> getBlockMetadata(final Map<String, Object> additional
...
+        new tech.pegasys.teku.api.schema.BeaconBlockHeader(blockHeader);
+
+    switch (spec.atSlot(block.getSlot()).getMilestone()) {
+      case PHASE0:
+        metadata.put("block", beaconBlock); // backward compatible with phase0
+        break;
+      case ALTAIR:
+        metadata.put(
+            "beacon_block",
+            new BlockRequestBody(spec.atSlot(block.getSlot()).getMilestone(), beaconBlock, null));
```
ajsutton: It probably makes sense to have two separate constructors for `BlockRequestBody`, one that takes just a beacon block and one that takes just a header.  May still need the version that takes both for JSON deserialisation but would be nice to clean this up to not have to specify null.

ajsutton (approved):
LGTM.

ethereum/core/src/main/java/tech/pegasys/teku/core/signatures/SigningRootUtil.java
usmansaleem: done.

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/ExternalSigner.java
usmansaleem: updated typo.

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/ExternalSignerBlockRequestProvider.java
usmansaleem: Done, introduced overloaded constructor. Also updated unit tests to assert correct fields are populated.

PR #5355, opened 2022-04-19
Add 400k validator perf blocks

ajsutton:
## PR Description
Add an additional set of blocks to use in performance tests with 400,000 validators so we can easily run benchmarks with large validator sets.

## Documentation

## Changelog


Nashatyrev (approved):
LGTM

PR #5376, opened 2022-04-22
Use Ubuntu 20.04 explicitly

ajsutton:
## PR Description
The new jammy ubuntu release is having issues so stick with the previously LTS for now.

Note that our current releases are using 20.04 already so this is maintaining the status quo, and it's still fully supported for some time yet. Logged #5377 to remind us to upgrade again in the future when the CI caching issues are sorted.

## Documentation

## Changelog


rolfyone (approved):
Lgtm

Issue #5231, opened 2022-03-24
Merge banner art 🐼


benjaminion:
Apparently it is mandatory to have an ASCII-art banner depicting pandas to appear on the console at the point of the Merge.

We could coordinate with Besu around this. Meanwhile I'll stick a good-first-issue label on this and see if any ASCII art gurus out there want to immortalise their name.

Somewhere around [here](https://github.com/ConsenSys/teku/blob/91c03c1b323c9551286bd10332624e550868f7a3/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java#L172) might be a good starting point.

Definition of done:
  - An ASCII art banner involving pandas appears at the point when Teku confirms that the TTD has been reached on testnets and mainnet.
  - Everyone agrees that it's super cool.
  - It doesn't break anything.
  - Colour is optional but must respect the user's existing settings.

0xOsprey:
This could be a good base image for someone talented to go out and use: https://i.imgur.com/4vm3E8m.png

Would be good to set a reasonable expectation around character width as well.

rolfyone:
@issanyo I got your PR running, if you're curious about how it would look:
![Screen Shot 2022-06-10 at 9.59.05 am.png](https://images.zenhubusercontent.com/5e45dddb5e32e495879572b3/ae190821-dd8f-4a5a-bc1d-6c40143a587f)

pincente:
How about this guys...
[bannerPoS.txt](https://github.com/ConsenSys/teku/files/8874953/bannerPoS.txt)

rolfyone:
That looks pretty good too @pincente!

aertoria:
Is this ticket still open? Looks like should be closed.

If it's open I can take it

PR #5378, opened 2022-04-23
Opt-in panda emoji for merge blocks

ajsutton:
## PR Description
When the system property `teku.panda` is set to `true`, mark merge blocks with a panda in the slot event log.

Leaving this as defaulting to off in case someone is parsing the Slot Event line.  We should still add a separate log line when the merge first happens but don't want to risk breakage by defaulting this to on.

## Fixed Issue(s)
#5231 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5379, opened 2022-04-23
Avoid spamming logs when no eth1 endpoints available

ajsutton:
## PR Description
Reduce log level to debug when we fail to update eth1 chain head because all endpoints are unsuitable.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5383, opened 2022-04-24
Add support for request bodies that support multiple content types

ajsutton:
## PR Description
Adds support to the new rest framework for specifying that the request body can be provided in multiple content types. This allows supporting posting either JSON or SSZ data.

Builds on #5734

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
```diff
@@ -172,11 +172,12 @@ public String getMigratedOpenApi() {
   }
 
   private void addStandardApiHandlers(
+      final Spec spec,
       final DataProvider dataProvider,
       final EventChannels eventChannels,
       final AsyncRunner asyncRunner,
       final BeaconRestApiConfig configuration) {
-    addBeaconHandlers(dataProvider);
+    addBeaconHandlers(spec, dataProvider);
```
rolfyone: dataProvider usually first if we want to be consistent...

ajsutton: Yeah it's a bit of a toss up because spec is usually first... :)

ajsutton: Just handlers haven't previously had spec.

rolfyone: ok well its probably inconsistent in there i guess... pretty sure other handlers had spec already but could be wrong

ajsutton: I don't mind switching it around.  I'd be surprised if handlers had spec before because we usually kept them away from all the internal details with the old reflection based API.  But 🤷‍♂️

tbenr:
Am I correct that we are testing only the JSON version and not yet the octet-stream\SSZ flow?

ajsutton:
> Am I correct that we are testing only the JSON version and not yet the octet-stream\SSZ flow?

Yes that's correct currently.  I was going to add a test for the SSZ version before merging but turns out we don't have an integration test for the JSON version either.  Currently the unit test mocks `Context` so it kind of covers it (assuming we mocked things right which is a big assumption), but we want to move the unit tests to mocking `Request` since `Context` is extremely difficult to mock well and that would make the unit test agnostic of the content type.

I've added a unit test to the existing set for now but need to think about this more as part of moving to mocking Request.

rolfyone (approved):
LGTM

PR #5393, opened 2022-04-27
Stop returning StoreBuilder from StorageQueryChannel

ajsutton:
## PR Description
Stop returning StoreBuilder from StorageQueryChannel - return a pure data class instead.  Returning a builder from an event channel call is bad practice - they should be purely data classes since event channels are the boundary between components.

Move more storage API classes to API module so the API is actually in the api module.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5397, opened 2022-04-27
Remove equivocating indices from score consideration

### LINKED ISSUE #5133: Implement discarding equivocating votes from the LMD fork choice
This has now been merged into the spec: https://github.com/ethereum/consensus-specs/pull/2845

> Discard equivocating attestations from the fork choice Store, and discount all future votes equivocating validators for fork choice purposes.

Danny's comment:

> Like proposer-boost, this is an important one to roll out in the next few months and can be rolled out iteratively rather than at a timed coordination point

zilm13:
## PR Description
Activates https://github.com/ethereum/consensus-specs/pull/2845
Currently in progress, posting draft.

My current TODO:
- add something to `FFGUpdatesTest`
- add something to `VotesTest`
- add something to `ForkChoiceTest`
- testing that cli flag works

## Fixed Issue(s)

Fixes #5133 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/VoteTracker.java
```diff
@@ -26,11 +26,25 @@ public class VoteTracker {
   private final Bytes32 currentRoot;
   private final Bytes32 nextRoot;
   private final UInt64 nextEpoch;
+  private final boolean equivocated;
```
ajsutton: Votes are stored on disk so we'll need to ensure we handle migration correctly.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -553,12 +566,40 @@ public void applyIndexedAttestations(final List<ValidateableAttestation> attesta
...
+        .thenPeek(
+            justifiedCheckpointState ->
+                onForkChoiceThread(
+                        () ->
+                            slashing
+                                .getIntersectingValidatorIndices()
+                                .forEach(transaction::addEquivocatingIndex))
+                    .reportExceptions())
+        .thenPeek(__ -> transaction.commit().join())
+        .reportExceptions();
```
ajsutton: We should be starting the transaction only after we've retrieved the checkpoint state and ensure we're on the fork choice thread for the whole lifetime of the transaction.

That said, I don't see why we actually need the `justifiedCheckpointState` here. Nothing is actually using it.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -391,6 +394,16 @@ public Optional<Bytes32> getProposerBoostRoot() {
     }
   }
 
+  @Override
+  public Set<UInt64> getEquivocatingIndices() {
+    readLock.lock();
+    try {
+      return equivocatingIndices;
```
ajsutton: This isn't thread safe because the returned `equivocatingIndices` set will be accessed from outside the readLock. We'd have to return a copy of the set to make it thread safe. That's probably reasonable given the size should be quite limited.

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransaction.java
```diff
@@ -160,6 +163,18 @@ public void removeProposerBoostRoot() {
     proposerBoostRootSet = true;
   }
 
+  @Override
+  public void addEquivocatingIndex(UInt64 equivocatingIndex) {
+    if (!store.equivocatingIndices.contains(equivocatingIndex)) {
+      equivocatingIndices.add(equivocatingIndex);
+    }
+  }
```
ajsutton: How do we prune the set of equivocating indices? We only need to hold them temporarily because once the validator is marked as slashed new attestations are ignored - we're just trying to remove the weighting from previous attestations from fork choice. Since we can't process attestations more than 2 epochs after the slot they are for we only need to remember these indices for at most 2 epochs.

It may be better to not track this in `Store` but just in something like `ForkChoiceStrategy`.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -440,6 +440,7 @@ protected void initAttesterSlashingPool() {
                     slashing -> slashing.getIntersectingValidatorIndices().size())
                 .reversed());
     blockImporter.subscribeToVerifiedBlockAttesterSlashings(attesterSlashingPool::removeAll);
+    attesterSlashingPool.subscribeOperationAdded(forkChoice::onAttesterSlashing);
```
ajsutton, line 443: This will only pick up slashings received via gossip but we'll miss any attestations included in blocks that we haven't already seen.

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransaction.java
ajsutton: We also aren't storing these indices to disk so if we restart we'll forget them all.  That's probably not the end of the world but may be something we need to ask some questions about to properly understand the implications.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/VoteTracker.java
ajsutton: Actually we don't seem to have updated the storage of votes on disk at all and will probably need to - otherwise we'll reapply weightings we previously removed on a restart. `VoteTrackerSerializer` has the code for this. It should be easy enough to read data with or without the extra boolean by checking if we've reached the end of the SSZ stream before reading it.

We need to avoid writing out the new boolean when the feature is toggled off.

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransaction.java
ajsutton: The easiest approach may be to do this a completely different way:

* Remove the equivocating indices set entirely.
* Have two new fields in `VoteTracker` - `currentEquivocating` and `nextEquivocating`
* When a slashing is received, immediately update the `VoteTracker` for any slashed validators to mark it's "nextEquivocating" as true
* In `computeDeltas`
    * if `currentEquivocating` is true, we can skip processing deltas for this validator as we're already ignoring their vote
    * otherwise if `nextEquivocating` is true we treat the new balance as zero which will then calculate a delta that removes the weighting for this validator.  The `newVote` then has `currentEquivocating` set to true.

This avoids all the issues with managing the equivocating indices set and ensures equivocating validators are recorded on disk since changes to `VoteTracker` are always stored.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
zilm13: Yeah, I've realized it yesterday, while getting to sleep. Will fix it now

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/VoteTracker.java
zilm13: Got you

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransaction.java
zilm13: Sounds like a good approach with current spec. The only downside I see is minor increase of Vote data size but in bonus we have consistency on restart

zilm13:
TODO:
- add `ForkChoiceTest` with `AttesterSlashing`s in block discarding previous attestations
- fix test PR if needed https://github.com/ethereum/consensus-specs/pull/2877
- don't like serialization/deserialization, looks a bit hacky. We already not versioning it, so we could add version and have something like `BeaconBlockBody` only with some migration. Feature is not yet activated and we actually may stuck without versioning only when we need another change in votes, so could postpone it till that moment. But maybe you see other good solution

zilm13:
Test finished, ready for review

ethereum/core/src/testFixtures/java/tech/pegasys/teku/core/AttesterSlashingGenerator.java
```diff
@@ -0,0 +1,127 @@
...
+  public AttesterSlashingGenerator(final Spec spec, final List<BLSKeyPair> validatorKeys) {
+    this.spec = spec;
+    this.validatorKeys = validatorKeys;
+    this.dataStructureUtil = new DataStructureUtil(spec);
+    final Predicates predicates = new Predicates();
+    final MiscHelpers miscHelpers = new MiscHelpers(spec.getGenesisSpecConfig());
+    final BeaconStateAccessors beaconStateAccessors =
+        new BeaconStateAccessorsPhase0(spec.getGenesisSpecConfig(), predicates, miscHelpers);
+    this.attestationUtil =
+        new AttestationUtil(spec.getGenesisSchemaDefinitions(), beaconStateAccessors, miscHelpers);
```
ajsutton: We can just get `AttestationUtil` from `spec` and may as well do it for the right slot.  So in `createAttesterSlashingForAttestation` we'd just have:
```
final AttestationUtil attestationUtil = spec.atSlot(blockAndState.getSlot())
        .getAttestationUtil();
```

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/VoteTracker.java
```diff
@@ -13,63 +13,53 @@
 
 package tech.pegasys.teku.spec.datastructures.forkchoice;
 
-import com.google.common.base.MoreObjects;
-import java.util.Objects;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 
-public class VoteTracker {
+public interface VoteTracker {
```
ajsutton: I don't think we want to go as far as introducing multiple implementation of VoteTracker.  When equivocating validators is disabled we'd just always have false for the two new booleans.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/VoteTracker.java
```diff
@@ -13,63 +13,53 @@
...
-    this.currentRoot = currentRoot;
-    this.nextRoot = nextRoot;
-    this.nextEpoch = nextEpoch;
+  static VoteTracker create(Bytes32 currentRoot, Bytes32 nextRoot, UInt64 nextEpoch) {
+    return new VoteTrackerV1(currentRoot, nextRoot, nextEpoch);
   }
 
-  public Bytes32 getCurrentRoot() {
-    return currentRoot;
+  static VoteTracker markNextEquivocating(final VoteTracker voteTracker) {
```
ajsutton: This is probably better as an instance method rather than static.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/VoteTracker.java
```diff
@@ -13,63 +13,53 @@
...
+        voteTracker.getCurrentRoot(),
+        voteTracker.getNextRoot(),
+        voteTracker.getNextEpoch(),
+        true,
+        false);
   }
 
-  public Bytes32 getNextRoot() {
-    return nextRoot;
+  static VoteTracker markCurrentEquivocating(final VoteTracker voteTracker) {
```
ajsutton: I don't think this method should exist as when we apply a vote we need to ensure we update all fields, not just equivocating.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -350,8 +357,10 @@ private BlockImportResult importBlockAndState(
 
     // We only need to apply attestations from the current or previous epoch. If the block is from
     // before that, none of the attestations will be applicable so just skip the whole step.
+    // Same applies to AttesterSlashings.
```
ajsutton: We probably need to provide better explanation here.  Attester slashings *do* apply even if they aren't from the current or previous epoch (unlike attestations) but the equivocating tracking isn't required to go back that far because we're only using the slashings to track which attestations matter.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -559,6 +568,43 @@ public void applyIndexedAttestations(final List<ValidateableAttestation> attesta
...
+        .getBeaconBlock()
+        .ifPresent(
+            beaconBlock ->
+                beaconBlock
+                    .getBody()
+                    .getAttesterSlashings()
+                    .forEach(
+                        attesterSlashing ->
+                            onAttesterSlashing(
+                                attesterSlashing, InternalValidationResult.ACCEPT, true)));
```
ajsutton: We can just use signedBeaconBlock.getMessage() here and avoid the optional.

```suggestion
    signedBeaconBlock
        .getMessage()
        .getBody()
        .getAttesterSlashings()
        .forEach(
            attesterSlashing ->
                onAttesterSlashing(attesterSlashing, InternalValidationResult.ACCEPT, true));
```

ajsutton: Actually, given we already create a vote update transaction in `applyVotesFromBlock` we should combine the two so that all the updates are done in a single transaction rather than two.  We can just extract the code from `onAttesterSlashing` that actually processes the slashing into another method so it can be reused.

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ForkChoiceStrategy.java
```diff
@@ -210,9 +210,13 @@ public Optional<Bytes32> getOptimisticallySyncedTransitionBlockRoot(final Bytes3
   void processAttestation(
       VoteUpdater voteUpdater, UInt64 validatorIndex, Bytes32 blockRoot, UInt64 targetEpoch) {
     VoteTracker vote = voteUpdater.getVote(validatorIndex);
+    // Not updating anything for equivocated validators
+    if (vote.isCurrentEquivocating() || vote.isNextEquivocating()) {
```
ajsutton: Probably worth having a `isEquivocating` method on `VoteTracker` that checks both current and next.

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArrayScoreCalculator.java
```diff
@@ -97,14 +101,19 @@ private static void computeDelta(
...
       addBalance(getIndexByRoot, deltas, vote.getNextRoot(), newBalance);
-
-      VoteTracker newVote =
-          new VoteTracker(vote.getNextRoot(), vote.getNextRoot(), vote.getNextEpoch());
+      final VoteTracker newVote;
+      if (vote.isNextEquivocating()) {
+        newVote = VoteTracker.markCurrentEquivocating(vote);
+      } else {
+        newVote = VoteTracker.create(vote.getNextRoot(), vote.getNextRoot(), vote.getNextEpoch());
+      }
```
ajsutton: I'd just simplify this to:
```suggestion
      newVote = new VoteTracker(vote.getNextRoot(), vote.getNextRoot(), vote.getNextEpoch(), vote.isNextEquivocating(), vote.isNextEquivocating());
```

it probably doesn't matter if we don't actually update the current roots once the validator is equivocating but since it's simple, it makes sense to get our records straight and show that we have considered that next vote now.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/serialization/VoteTrackerSerializer.java
```diff
@@ -28,7 +31,14 @@ public VoteTracker deserialize(final byte[] data) {
...
           final UInt64 nextEpoch = UInt64.fromLongBits(reader.readUInt64());
-          return new VoteTracker(currentRoot, nextRoot, nextEpoch);
+          if (reader.isComplete()) {
+            // Old version
+            return new VoteTrackerV1(currentRoot, nextRoot, nextEpoch);
+          } else {
+            // Version with equivocation
+            return new VoteTrackerV2(
+                currentRoot, nextRoot, nextEpoch, reader.readBoolean(), reader.readBoolean());
+          }
```
ajsutton: We deliberately read values on a separate line to the constructor so the reads don't get reordered if the constructor parameter order is refactored.

And this can just build a single version like:
```suggestion
          final boolean currentEquivocating;
          final boolean nextEquivocating;
          if (reader.isComplete()) {
            currentEquivocating = false;
            nextEquivocating = false;
          } else {
            currentEquivocating = reader.readBoolean();
            nextEquivocating = reader.readBoolean();
          }
          return new VoteTracker(currentRoot, nextRoot, nextEpoch, currentEquivocating, nextEquivocating);
```

Actually we probably want to know if equivocating validators has been disabled and read the booleans but then set them to false anyway.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/serialization/VoteTrackerSerializer.java
```diff
@@ -40,6 +50,11 @@ public byte[] serialize(final VoteTracker value) {
               writer.writeFixedBytes(value.getCurrentRoot());
               writer.writeFixedBytes(value.getNextRoot());
               writer.writeUInt64(value.getNextEpoch().longValue());
+              // Version with equivocation
+              if (value.getVersion() == Version.V2) {
+                writer.writeBoolean(value.isNextEquivocating());
+                writer.writeBoolean(value.isCurrentEquivocating());
+              }
```
ajsutton: If equivocating support is enabled we should always write these fields (which then migrates the data over to the new format) and if it's disabled we never write the extra two booleans. That way we're always writing data that matches what we'd expect based on the feature toggle.

ajsutton (commented):
In terms of storing a version with vote tracker information, I wouldn't bother. It's quite easy to handle adding additional fields which is all we need here and we're unlikely to need to remove any of the existing fields.  If we do in the future we can either add versioning or just discard existing tracked votes - as a one of thing for a migration there won't be any issue with that, we'll just be less able to determine the current head after restart until we start receiving attestations again, so it resolves itself within an epoch and may not make any difference if we need to sync for more than a couple of blocks.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/VoteTracker.java
zilm13: We shouldn't update `latest_messages` vote for equivocating validator as by the spec, and dedicated method ensures that vote fields are not changed
https://github.com/ethereum/consensus-specs/blame/dev/specs/phase0/fork-choice.md#L365
Or perhaps, I don't understand what do you mean

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArrayScoreCalculator.java
zilm13: Returning to this one. I've applied your suggestion but according to the spec, we shouldn't update votes for equivocated validators. This shouldn't hurt anything from the first look, but it could be our bug in the future

zilm13: I've reverted this one too, but if you think it's safe to update it, I will put it back

zilm13:
@ajsutton 
I have two questions before continue to update this one:
1. You've suggested to update votes for equivocated validators, but I think it's against the spec https://github.com/ethereum/consensus-specs/blame/dev/specs/phase0/fork-choice.md#L365 and it could be a source for the future bugs, are you pretty sure we are on safe path here?
2. Different serialization depending on Teku flag would require to pass this flag with `SpecConfig` or something like that. DB is currently pretty independent of everything but the `spec` which is pretty rational as spec could touch the schema. Do we really want to pass `TekuConfiguration` there and make DB dependent on that? My current "versions" approach keeps this independence. Maybe it could be tweaked in different manner but making DB module dependent on `TekuConfiguration` for this case looks a bit unpleasant from my side

ajsutton:
I'll have to look at the spec in more detail but our implementation of fork choice is very different from the spec version. It's probably not a big deal which way it goes to be honest. 

I don't think the database needs all of TekuConfiguration passed through but it already has a number of config items that affect how it behaves. The current version means everything that deals with this is handling multiple versions but we want to isolate the feature toggling to just the key entry points (eg the database)

zilm13:
Generally I agree that from some angle equivocating indices feature toggle looks like a part of spec (it's a part of spec btw :)

ajsutton:
So I think you're misunderstanding what the spec is saying vs what our code is doing because while our code gives the same results as the spec the implementation is very very different.  The high level overview of it is that the spec applies attestations in a single step - by adding them to `store.latest_messages` where as our protoarray implementation applies things in two steps.  First we update `VoteTracker` to set the `next` values to the vote we received. Then later we update weightings from a whole batch of changed attestations at once - when we do that we set the `current` values to the `next` ones so that we can tell what's changed and not changed on the next run.

So for us, not applying an attestation from an equivocating validator means not updating the `next` values when we receive a new attestation from them.  We have that check in `ForkChoiceStrategy.processAttestation`. And for removing the existing weighting we have the check in `ProtoArrayScoreCalculator` to set the new balance to 0 if nextEquivocating is true.

The idea of next and current doesn't exist in the spec so it's just for our record keeping and it's more accurate to say that we have applied the latest changes for that validator (so current and next have the same values).

zilm13:
Thank you for explanation. I finally got it. The same way we set intention to change vote's target epoch, and incarnate it later in `ProtoArrayCalculator` we do with equivocation. And it's safe as we haven't changed equivocating vote root in `ForkChoice`. Ready for review.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfig.java
```diff
@@ -135,6 +135,8 @@ static SpecConfigBuilder builder() {
 
   int getProposerScoreBoost();
 
+  boolean isEquivocatingIndicesEnabled();
```
ajsutton: hmm, I don't think this belongs in `SpecConfig` as it's not coming from the config the spec provides. For the storage side it should just wind up in `StorageConfiguration`.

ajsutton (commented):
Looks good, just need to move the config out of SpecConfig to StorageConfiguration (it will likely mean passing it through as a boolean in a few layers but that will only be temporary until we remove the feature toggle again).

zilm13:
@ajsutton did you mean something like this?

ajsutton (approved):
Yep spot on.  Thanks.

PR #5401, opened 2022-04-28
Set filename and Eth-Consensus-Version header for SSZ responses

ajsutton:
## PR Description
When responding with SSZ content, set the `Content-Disposition` header and `Eth-Consensus-Version` headers appropriately.

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/EthereumTypes.java
```diff
@@ -31,4 +39,17 @@ public class EthereumTypes {
...
+      final Function<T, SpecMilestone> milestoneSelector) {
+    return new OctetStreamResponseContentTypeDefinition<>(
+        SszData::sszSerialize,
+        value ->
+            Map.of(
+                RestApiConstants.HEADER_CONSENSUS_VERSION,
+                Version.fromMilestone(milestoneSelector.apply(value)).name(),
+                RestApiConstants.HEADER_CONTENT_DISPOSITION,
+                String.format(
+                    "filename=\"%s.ssz\"", value.hashTreeRoot().toUnprefixedHexString())));
```
rolfyone: i know we do this currently, wondering if it is worth putting `className_hash.ssz` if it's going to be used for everything now rather than just state?

ajsutton: I was originally thinking `getSchema().getName()` so that's probably worth doing yeah.

ajsutton: Turns out we don't always have a name for SszData but we now include it when we do.  And for the things we're likely to be returning (states and blocks) we should always have a name.

rolfyone (approved):
LGTM. promise i wasn't stalling until i merged :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/EthereumTypes.java
```diff
@@ -31,4 +39,19 @@ public class EthereumTypes {
...
+    return new OctetStreamResponseContentTypeDefinition<>(
+        SszData::sszSerialize,
+        value ->
+            Map.of(
+                RestApiConstants.HEADER_CONSENSUS_VERSION,
+                Version.fromMilestone(milestoneSelector.apply(value)).name(),
+                RestApiConstants.HEADER_CONTENT_DISPOSITION,
+                String.format(
+                    "filename=\"%s%s.ssz\"",
+                    value.getSchema().getName().map(name -> name = "-").orElse(""),
```
tbenr: shouldn't be `name + "-"`?

ajsutton: The map adds a - to the end of the name if there is one.  Avoids having a leading - if there was no name.

tbenr: your code mismatches your description 😉

ajsutton: Yeah I caught on to what you were saying eventually. :)

Issue #5400, opened 2022-04-27
SSZ support for block creation


rolfyone:
add SSZ support to get new block (and blinded)
add SSZ support to post signed block (and blinded)

 - all 4 would be migrated.

rolfyone:
raised #5588 to document and make non dev flag, which is the outstanding portion of this ticket, so will close this.

PR #5402, opened 2022-04-28
add SSZ response type to GetNewBlock.

rolfyone:
GetNewBlindedBlock was already functional. Added an integrationTest to exercise json and ssz responses for both.

Both GetNewBlock and GetNewBlindedBlock need to be setting filename as part of the response, once that functionality is built.

partially addresses #5400

## Documentation

## Changelog


tbenr (approved):
LGTM.
We should at some point run rest API at BELLATRIX milestone and check actual blinded blocks.

PR #5388, opened 2022-04-27
Improve error message when SSL keystore not provided for validator API.

ajsutton:
## PR Description
Improve error message when the validator API can't start because no SSL key is provided.

## Fixed Issue(s)
fixes #5386 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #5403, opened 2022-04-28
Recognise logging options before or after the validator-client subcommand

ajsutton:
## PR Description
Previously the logging options like `--log-destination` were only recognised if they were specified prior to the `validator-client` subcommand.  This change ensures the same `LoggingOptions` instance is used for both the parent and subcommand so that the option works either before or after the subcommand.

We can't just use scope = INHERIT here because not all subcommands support the logging options.

## Fixed Issue(s)
Found as part of #5388 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5404, opened 2022-04-29
update openapi publisher dependencies

tbenr:
update node dependencies to update async module to 2.6.4 to solve security warning on teku dep


ajsutton (approved):
LGTM

ajsutton:
Worth noting for the record this is only used in CI to publish the open api spec Sonia just dealing with teku and should be safe but still worth keeping up to date

Issue #5396, opened 2022-04-27
[MEV boost\Builder] Builder spec `0.2.0` support


tbenr:
### Description
we want to support version `0.2.0` and . More info here:
https://github.com/ethereum/builder-specs/releases/tag/v0.2.0

https://ethereum.github.io/builder-specs/

### general flow requirements

- [x]  track builder availability status (`builder_status`)
- [x]  implement parallel calls to `builder_getHeaderV1` and `engine_getPayload` with fallback handling (error or timeout)
  - [x]  should be able to transparently handle the `builder_getPayloadV1` with the payload returned by `engine_getPayload`.
- [x]  must always use local execution engine if merge isn't completed (non-default execution payload finalized)

### Execution API Clients:
- [x] separate `builder_` api from `execution_` api clients (migrate `builder_` to `ExecutionBuilderClient`)
- [x] generic rest api client 
  - [x] implement `builder_status`
  - [x] implement `builder_registerValidatorV1`
  - [x] update `builder_getPayloadHeaderV1` to `builder_getHeaderV1`
  - [x] update `builder_proposeBlindedBlockV1` to `builder_getPayloadV1`

### ExecutionEngineChannel:
- [x] make executionengine channel more generic (rename to `ExecutionLayerChannel`).
  - Will to expose `engine_`,  `builder_` and some `eth1_` endpoints via two separate jsonrpc clients
  - analyze the the option to make the `ExecutionLayerManagerImpl` less a simple proxy for api calls and being responsible for the fallback logic. It could abstract away all the "general flow requirements".
- [x] #5610

### ExecutionEngineService:
- [x] let the service poll `builder_status` to track builder flow availability

### CLI and configuration
- [x] introduce new `eb-endpoint` (execution builder)
- [x] #5638
- [x] #5639
- [x] #5640

### Beacon Rest API:
- [x] support for validator registration (ref: https://github.com/ethereum/beacon-APIs/pull/209) in `register_validator` API. Must call `builder_registerValidator`.

### Validator Client:
- [x] #5596
- [x] #5636
- [x] #5637
- [x] #5644

### Beacon Node:
- [x] change block proposal flow to integrate `builder_` (might be easy if `ExecutionEngineChannel` will be responsible for most of the work)

### Latest Spec changes

- [x] #5595
- [x] #5613
- [x] #5873 

### Bugfix - Improvements
- [x] #5716
- [x] #5712
- [x] #5734
- [x] #5756 
- [x] #5880

[...]

PR #5406, opened 2022-05-02
[MEV-Boost\builder] implement execution builder api client

tbenr:
## PR Description

- Introduces `Web3JExecutionBuilderClient`
- Introduces all the required execution client schema files to support the new apis
- Introduced required SSZ objects (`SignedValidatorRegistrationV1` and `SignedBuilderBidV1`) to be able to handle the signature (generation\validation).

In upcoming PR I'll remove builder API from `Web3JExecutionEngineClient` and integrate `Web3JExecutionBuilderClient` in `ExecutionEngineChannel`

## Fixed Issue(s)
related to #5396 

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/integration-test/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3JExecutionBuilderClientTest.java
+++ b/ethereum/executionlayer/src/integration-test/java/tech/pegasys/teku/ethereum/executionlayer/client/Web3JExecutionBuilderClientTest.java
@@ -17,16 +17,11 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.InstanceOfAssertFactories.INTEGER;
 import static org.assertj.core.api.InstanceOfAssertFactories.STRING;
 
-import com.fasterxml.jackson.core.JsonFactory;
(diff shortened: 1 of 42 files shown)


ajsutton (commented):
Worth talking through the details of this on the call tonight - I think it generally looks right, but I am quite confused why the execution builder is returning beacon block content like slashings etc.  The execution layer ideally wouldn't know anything about those and that's where most of the API trouble is coming from. But I'm not familiar with the latest builder api specs so am probably missing something.

And I think there's a strong argument here that we should be able to have a version of these APIs that does just send SSZ instead of JSON - particularly if the whole beacon block winds up needing to be sent.

ajsutton (approved):

lightclient:
> I am quite confused why the execution builder is returning beacon block content like slashings etc. The execution layer ideally wouldn't know anything about those and that's where most of the API trouble is coming from.

@ajsutton the builder is accepting beacon block content, not returning it. They need the signed blinded beacon block to verify the proposer's signature and to be able to release it the network in case the proposer tries to withhold. If they only received the blinded beacon block header, they wouldn't have enough information to propagate the block on their own.

PR #5407, opened 2022-05-02
Migrate PostPrepareBeaconProposer api endpoint

rolfyone:

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/versions/bellatrix/BeaconPreparableProposer.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/versions/bellatrix/BeaconPreparableProposer.java
@@ -13,12 +13,36 @@
 
 package tech.pegasys.teku.spec.datastructures.operations.versions.bellatrix;
 
+import static tech.pegasys.teku.infrastructure.json.types.CoreTypes.BYTES20_TYPE;
+import static tech.pegasys.teku.infrastructure.json.types.CoreTypes.UINT64_TYPE;
+
 import com.google.common.base.MoreObjects;
 import java.util.Objects;
 import tech.pegasys.teku.infrastructure.bytes.Bytes20;
+import tech.pegasys.teku.infrastructure.json.types.DeserializableTypeDefinition;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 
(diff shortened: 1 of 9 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/versions/bellatrix/BeaconPreparableProposer.java
```diff
@@ -60,4 +84,33 @@ public String toString() {
         .add("feeRecipient", feeRecipient)
         .toString();
   }
+
+  public static DeserializableTypeDefinition<BeaconPreparableProposer> getJsonTypeDefinition() {
+    return BEACON_PREPARABLE_PROPOSER_TYPE;
+  }
```
ajsutton: Hmm, this feels like a change to the pattern of where type definitions are defined. Normally they're separate from the actual data class.  I think it might be blurring the lines between internal and external representations... Not a strong opinion though.

There is some confusion caused by this class having wound up in spec/datastructures/operations when it's not actually an operation and not even really part of the beacon chain spec. It's really just an internal class we created because we needed it and it happens to have the same data as the REST API endpoint needs.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/CoreTypes.java
```diff
@@ -38,6 +39,15 @@ public class CoreTypes {
           .format("byte")
           .build();
 
+  public static final StringValueTypeDefinition<Bytes20> BYTES20_TYPE =
```
ajsutton, line 42: Do we actually want bytes20 or eth1 address here?

ajsutton (approved):
LGTM. Just not entirely sure the type definition should actually go in `BeaconPreparableProposer` but happy to leave it to your judgement.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/CoreTypes.java
rolfyone: probably at some stage an eth1 address is my guess, but it's pretty built into the objects at the moment that it's a bytes20... probably a bit of a cleanup, bigger than the scope of this migration

PR #5408, opened 2022-05-03
Add time based tolerance before disconnecting event streams

### LINKED ISSUE #5399: Attestation event stream always closes abruptly and can sometimes trigger an Exception
### Description
I need a constant feed of all Attestation my node receives for a project I'm working on. The `http://node/eth/v1/events?topics=attestation` endpoint fits my needs perfectly, but is sadly unusable due to Teku constantly closing the stream. 

_This is both reproducible on my private node and with the endpoints infura provides, as they also use Teku._

### Steps to Reproduce (Bug)
`curl -X GET "http://node:5051/eth/v1/events?topics=attestation" -H 'Accept: text/event-stream'`

**Expected behavior:**
Well, nothing. I would expect an uninterrupted stream of attestation and no Exceptions being triggered.

**Actual behavior:** 
After a few seconds, the stream will simply be cut-off (sometimes in the middle of data), with no error in the node logs. Every now and then an Exception like [this](https://github.com/ConsenSys/teku/files/8577731/teku_exception.log) or [this](https://github.com/ConsenSys/teku/files/8577773/teku_exception.log) appears when the stream gets cut.

**Frequency:** 
The Exception occurs about 20% of the time, the cut-off itselfs always after varying durations (1-30 seconds).

### Versions
* Software version: teku/v22.4.0/linux-x86_64/-eclipseadoptium-openjdk64bitservervm-java-17
* OS Name & Version: Ubuntu 22.04 LTS
* Docker Version: 20.10.14 Community
* Cloud VM, type, size: Hetzner AX41-NVME

ajsutton:
## PR Description
Given that huge numbers of attestations are received at once, a strict queue length based limit on event stream subscribers is too strict as there's no chance to serialize and read all those events. With this change, the event queue can exceed the maximum size for up to one second, ensuring there is time for the messages to be sent if the subscriber is reading. If the subscriber is slow the timeout is still hit quickly without too many events backing up and causing memory issues.

## Fixed Issue(s)
fixes #5399 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,3 +29,4 @@ For information on changes in released versions of Teku, see the [releases page]
 - Fixed issue with loading the optimised BLST library on Windows.
 - Reduced log level for notifications that the eth1 chain head could not be retrieved because no endpoints were available.
 - Fixed issue where logging options were not recognised if specified after the `validator-client` subcommand.
+- Avoid disconnecting event stream connections subscribed to attestation events for briefly exceeding the maximum pending event queue size. A very large number of attestations are received all at once on MainNet making it almost impossible for a consumer to stay below the queue size limit at all times.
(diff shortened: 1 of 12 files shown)


rolfyone (approved):
LGTM

PR #5409, opened 2022-05-03
Migrate GetStateValidator to MigratingEndpointAdapter

### LINKED ISSUE #5398: Migrate GetStateValidator to MigratingEndpointAdapter

courtneyeh:
fixes #5398 

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/migrated/StateValidatorData.java
```diff
@@ -0,0 +1,148 @@
...
+
+  public ValidatorStatus getStatus() {
+    return status;
+  }
+
+  public Validator getValidator() {
+    return validator;
+  }
+
+  public static ValidatorStatus getValidatorStatus(
```
rolfyone: this is one we don't want to replicate, do we need to? the logic is complicated... it also has a bunch of testing around it, and we definitely will need to keep that testing :)

rolfyone (approved):
LGTM

PR #5413, opened 2022-05-03
Improve sync message when waiting for EL sync to complete

### LINKED ISSUE #5330: [Merge] Print Syncing message when attestations due
### Description
Currently Teku prints the "Syncing" message at the very start of the slot so when waiting for EL sync to complete it always reports the remaining slots as 1 because the block for the just started slot hasn't been received yet.  This suggests the beacon chain isn't completely in sync when it actually is.

The "Slot Event" message we print is delayed until attestations are due to give the block for the current slot a chance to be imported.  We should move the "Syncing" message to this same point so that it can report 0.

Key acceptance criteria:
The "Syncing" messages report "Remaining slots: 0" when optimistic sync has completed and we are just waiting on the EL to finish syncing.

Easiest way to test this is to change `ExecutionEngineService.doStart` to use `ExecutionEngineChannel.NOOP` instead of creating a real execution engine channel then run teku on kiln. To sync more quickly grab a checkpoint state from one of our kiln nodes.  The NOOP execution engine always returns syncing so Teku will stay in optimistic sync mode.

ajsutton:
## PR Description
Improve syncing message when waiting for the EL sync to complete. Previously it would always report 1 remaining slot, now it explicitly states that teku is waiting for the EL sync to complete:

```
Syncing     *** Target slot: 378413, Head slot: 378405, Waiting for execution layer sync, Connected peers: 10
```

## Fixed Issue(s)
fixes #5330 

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
@@ -87,6 +87,15 @@ public class EventLogger {
     info(syncEventLog, Color.WHITE);
   }
 
+  public void syncEventAwaitingEL(
+      final UInt64 nodeSlot, final UInt64 headSlot, final int numPeers) {
+    final String syncEventLog =
+        String.format(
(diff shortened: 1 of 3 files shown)


tbenr (approved):
LGTM

PR #5418, opened 2022-05-03
Add checksum to Eth1Address

### LINKED ISSUE #5344: `Eth1Address` should warn if checksum fails
### Description

I think that [`Eth1Address`](https://github.com/ConsenSys/teku/blob/45a394e9888b98e2baac97fee2b2ee61096f0315/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/eth1/Eth1Address.java#L20) should check the address checksum as defined in [EIP-55](https://eips.ethereum.org/EIPS/eip-55). It should warn the user if the checksum fails. Maybe only warn if we detect mixed case (at least one lowercase letter & at least one uppercase letter).

jtraglia:
## PR Description

Thought it would be useful to check addresses to help prevent any typos. This uses the mixed-case checksum as defined in EIP-55. Also, we now check that the address has the "0x" prefix. The description made it seem like it was required.

The issue I made a while back suggested we make this a warning. This implementation will throw an `IllegalArgumentException` if it encounters an address that fails checksum. If this is not the desired behavior, I'll change it.

## Fixed Issue(s)

Fixes #5344

## Documentation

Will need to update the address for `--validators-proposer-default-fee-recipient` here:
* https://docs.teku.consensys.net/en/latest/Reference/CLI/CLI-Syntax/#validators-proposer-default-fee-recipient

```diff
- 0xfe3b557e8fb62b89f4916b721be55ceb828dbd73
+ 0xFE3B557E8Fb62b89F4916B721be55cEb828dBd73
```

I do not see how to add labels to the PR. I guess someone else needs to do that.

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfig.java
```diff
@@ -138,7 +138,7 @@ static SpecConfigBuilder builder() {
 
   int getDepositNetworkId();
 
-  Bytes getDepositContractAddress();
+  String getDepositContractAddress();
```
ajsutton: I wonder if we can make this just return `Eth1Address`.  Have a vague feeling I've been around that loop once before though...  If it's easy that would be a nice win since we're changing the type anyway, but if we run into module dependencies or it just starts causing too many knock on changes it's not worth it.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/eth1/Eth1Address.java
```diff
@@ -27,17 +32,93 @@ public class Eth1Address extends Bytes20 {
...
   public static Eth1Address fromHexString(String value) {
-    return new Eth1Address(Bytes.fromHexString(value));
+    checkAddress(value);
+    Eth1Address address = new Eth1Address(value);
+    Eth1Address encodedAddress = fromHexStringWithChecksum(value);
+    checkArgument(
+        address.equals(encodedAddress),
+        "Eth1Address fails checksum:\n got: %s\n exp: %s",
+        address,
+        encodedAddress);
```
ajsutton: hmm, I'm surprised we have both `fromHexString` and `fromHexStringWithChecksum` if both are checking the checksum.

ajsutton: oh I see, `fromHexStringWithChecksum` is reformatting to add the checksum. 

So I think there's a few changes we should make here:

1. If the supplied address is either all upper or all lower case we should accept it without checking the checksum. It's very common for users to have addresses that don't have the checksum embedded and quite hard for them to calculate the checksum so requiring it is a bit harsh. If there's mixed case we enforce the checksum which I believe matches how MetaMask handles this.
2. When we print addresses we should always add the checksum. So if the user supplies the value all-lower case, when we print it it will come back as mixed case with the checksum correctly computed.
3. We probably want to avoid creating the `Bytes` instance multiple times. So I'd modify the `Eth1Address` constructor to take both `Bytes` and `String`, then we can parse to `Bytes` and check the length is right, calculate and potentially check the hash code here and then pass the pre-built `Bytes` and mixed-case `String` into the constructor.
4. We should then be able to remove `fromHexStringWithChecksum` and probably just have a private method that takes a `String` and returns it formatted with the correct checksummed case.

And I probably wouldn't require the `0x` prefix. I'd be surprised if people don't have that but no real reason to require it.

infrastructure/crypto/src/main/java/tech/pegasys/teku/infrastructure/crypto/MessageDigestFactory.java
```diff
@@ -25,17 +25,28 @@ public class MessageDigestFactory {
   private static final Logger LOG = LogManager.getLogger();
 
   public static final String SHA_256 = "SHA-256";
-  private static final Provider SECURITY_PROVIDER = selectSecurityProvider();
+  public static final String KECCAK_256 = "KECCAK-256";
+  private static final Provider SHA_256_SECURITY_PROVIDER = selectSha256SecurityProvider();
+  private static final Provider KECCAK_256_SECURITY_PROVIDER = new BouncyCastleProvider();
```
ajsutton: Probably worth having a constant `BOUNCY_CASTLE_PROVIDER` since we'll always need it now, then just use that existing instance here and in `selectSha256SecurityProvider`.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/eth1/Eth1Address.java
```diff
@@ -27,17 +32,93 @@ public class Eth1Address extends Bytes20 {
...
+      return false;
+    }
+    final Eth1Address a = (Eth1Address) o;
+    return address.equals(a.address);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(address);
+  }
```
ajsutton: We probably don't want to override equals and hashCode here so that we maintain the equality rules of `Bytes` which doesn't care about the particular implementation class.

ajsutton (commented):
Thanks for this. Will definitely be useful when users are specifying fee-recipient and getting money paid to there.

jtraglia:
Thank you for the review. I agree with all of your comments. I will make these changes soon.

jtraglia:
Alright, I think this is ready for another review. This version is a lot better IMO; the code is much cleaner.

The only thing I didn't do was:

> modify the `Eth1Address` constructor to take both `Bytes` and `String`

With the new changes, I do not think that this is a problem anymore.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/eth1/Eth1Address.java
```diff
@@ -27,17 +31,77 @@ public class Eth1Address extends Bytes20 {
...
+    }
+  }
 
-  public Eth1Address(final Bytes bytes) {
-    super(bytes);
+  private static boolean isMixedCase(String value) {
+    String valueWithoutPrefix = value.replace("0x", "");
+    return !valueWithoutPrefix.toLowerCase().equals(valueWithoutPrefix)
+        && !valueWithoutPrefix.toUpperCase().equals(valueWithoutPrefix);
   }
```
ajsutton: We can use `StringUtils.isMixedCase` (from commons lang3)

ajsutton (approved):
LGTM. I'll just push one small change to use the isMixedCase method from commons lang rather than the custom implementation then will merge. Thanks!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/eth1/Eth1Address.java
jtraglia: Ah perfect. I didn't know that existed. Thank you.

ajsutton:
Well dang, I broke it. :D Will have a look at why isMixedCase was different.

jtraglia:
Ah yeah, needed to remove the `0x` like you discovered. Thanks!

PR #5419, opened 2022-05-03
Omit stack trace when no peers are available to send attestation to

ajsutton:
## PR Description
When an attestation fails to publish because there are no peers available on the required subnet, log a warning with no stack trace.  This significantly reduces the amount of noise in the logs during local testing when there are no peers, but doesn't hide the problem which could be quite meaningful in a normal situation.

Note the previous attempt to log these at debug level failed because it wasn't checking the root cause of the exception and it would have completely hidden the message which isn't ideal.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5420, opened 2022-05-03
Improve error message when spec cannot be created

ajsutton:
## PR Description
When a custom config file is used that doesn't include the bellatrix config options, provide a more useful error message rather than just `NoSuchElement`.

Also delete an unrelated class that was unused. :)

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5421, opened 2022-05-04
Update dependencies

ajsutton:
## PR Description
Do a sweep of dependency updates. Errorprone got smarter so tweak some things to make it happy. Mostly it now detects public methods that are unused.

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #5424, opened 2022-05-04
Update javalin

ajsutton:
## PR Description
The swagger annotations updated to use {param} style instead of :param so update handlers to use that new form.
Also needed to update the swagger-ui to work with the new version.

Builds on #5421 

## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,12 +15,12 @@ buildscript {
 }
 
 plugins {
-  id 'com.diffplug.spotless' version '6.4.1'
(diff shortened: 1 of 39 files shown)


rolfyone (approved):
LGTM

ajsutton:
This is blocked by https://github.com/tipsy/javalin/issues/1532 Going to close for now - we'll have to skip this version and go to the next one so will resurrect then.

PR #5428, opened 2022-05-04
[MEV Boost\Builder] Implement builder_status and health check

StefanBratanov:
## PR Description
**Tasks:**

Retrieve builder status and implement `isBuilderAvailable`
Track builder flow availability

**Changes:**
- Introduced `ExecutionLayerManager`
- Added `isBuilderAvailable` implementation to `ExecutionLayerManager`
- Added `onSlot` logic to `ExecutionLayerManager` to track and update the builder flow availability

## Fixed Issue(s)
related to #5396 

## Documentation

## Changelog


tbenr (approved):
LGTM but I'm still rethinking about this: having the healthcheck in the `ExecutionLayerService` makes `ExecutionLayerChannelImpl` unaware of the current status (which makes sense to be aware in case we want to immediately skip to calls based on that)

StefanBratanov:
> LGTM but I'm still rethinking about this: having the healthcheck in the `ExecutionLayerService` makes `ExecutionLayerChannelImpl` unaware of the current status (which makes sense to be aware in case we want to immediately skip to calls based on that)

Let me try implementing it in the Impl and see the difference. It should be a quick one.

tbenr:
But it could be just a peek in the impl.. let's have a quick chat tomorrow if you don't find a quick nice solution

tbenr (approved):
LGTM

PR #5429, opened 2022-05-04
[MEV Boost\Builder] execution builder client part 3

tbenr:
## PR Description

- Completes the separation of builder client and engine client.
- Moves throttling from channel to clients
- Introduces `ExecutionPayloadContext` which carries the context for a new execution payload has been produced (`payloadId`, `forkChoiceState`, and `payloadAttributes`). Used to retrieve `parentHash` and (in later PRs) publicKey of the validator, required in builder apis.

related to #5396

## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerChannelImpl.java
```diff
@@ -213,39 +224,61 @@ public SafeFuture<TransitionConfiguration> engineExchangeTransitionConfiguration
   }
 
   @Override
-  public SafeFuture<ExecutionPayloadHeader> getPayloadHeader(
+  public SafeFuture<ExecutionPayloadHeader> builderGetHeader(
       final Bytes8 payloadId, final UInt64 slot) {
```
ajsutton: My guess is that we need to start passing the data this method needs in rather than just payloadId. I think `ForkChoiceNotifier.getPayloadId` is already being called on the beacon node side before it calls through to this method and it knows the proposer ID (which can be mapped to validator ID easily) and the parent execution block hash. I think it's even verifying they're the right ones for the request so the payloadID matches up so it should fit pretty well to do the work there.  Might be some complications I've missed though.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerChannelImpl.java
```diff
@@ -213,39 +224,61 @@ public SafeFuture<TransitionConfiguration> engineExchangeTransitionConfiguration
...
                     slot,
                     executionPayloadHeader));
   }
 
+  private ExecutionPayloadHeader getExecutionHeaderFromBuilderBid(
+      SignedMessage<BuilderBidV1> signedBuilderBid, UInt64 slot) {
+    ExecutionPayloadHeaderSchema executionPayloadHeaderSchema =
+        SchemaDefinitionsBellatrix.required(spec.atSlot(slot).getSchemaDefinitions())
+            .getExecutionPayloadHeaderSchema();
+    // validate signature
```
ajsutton, line 260: Where does the public key to validate the signature come from? If it's a validator key then it would have to come from state and we won't be able to verify the signature over here - we'd have to pass it all back to the beacon node unchecked which is messy...

ajsutton: Seems like we should make sure we have a ticket to come back and actually validate the signature here.

ajsutton (approved):
LGTM.

Issue #5304, opened 2022-04-06
Ignore aggregate gossip that's a subset of previous messages


ajsutton:
### Description
There's no value in processing aggregates that are a pure subset of other aggregates we've already processed so we should just IGNORE them and save some verification work. Applies to both attestation and sync committee aggregates.

See https://github.com/ethereum/consensus-specs/pull/2847 for specifics.

PR #5439, opened 2022-05-06
Ignore aggregate attestations that are a subset of aggregates we've already seen

ajsutton:
## PR Description
Change the aggregate attestation validation rules to ignore any aggregate that we've seen a non-strict superset for already. This avoids reprocessing and gossiping aggregates that add no new value.

The set of received aggregate hashes is removed because they will be caught by this new check (ie an aggregate is always a non-strict superset of itself).

## Fixed Issue(s)
#5304  (need to apply to sync committee aggregates as well before it's complete)

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5440, opened 2022-05-06
Ignore sync committee contributions that are a subset of ones we've already seen

### LINKED ISSUE #5304: Ignore aggregate gossip that's a subset of previous messages

ajsutton:
## PR Description
Sync sync contribution validation rules to ignore any contribution that we've seen a non-strict superset for already.

Also reduced the cache size for sync committee gossip. There are only 512 members of the sync committee and we automatically ignore messages that aren't for the current slot so at most we need to remember 512 seen messages.

## Fixed Issue(s)
fixes #5304 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java
```diff
@@ -35,8 +35,10 @@ public class Constants {
   // Target 2 different attestation data (aggregators normally agree) for two slots
   public static final int VALID_ATTESTATION_DATA_SET_SIZE = 2 * 64 * 2;
   public static final int VALID_VALIDATOR_SET_SIZE = 10000;
-  public static final int VALID_CONTRIBUTION_AND_PROOF_SET_SIZE = 10000;
-  public static final int VALID_SYNC_COMMITTEE_MESSAGE_SET_SIZE = 10000;
+  // Only need to maintain a cache for the current slot, so just needs to be as large as the
+  // sync committee size.
+  public static final int VALID_CONTRIBUTION_AND_PROOF_SET_SIZE = 512;
+  public static final int VALID_SYNC_COMMITTEE_MESSAGE_SET_SIZE = 512;
```
rolfyone, line 41: if it still breaks, maybe increase this to be large enough for 2 slots.. i have vague memories...

ajsutton: It should be all good now.  Having that too low would make it more likely to pass as we'd process more messages than we need to.  But for sync committee messages I'm pretty confident there's no need to keep two slots since we ignore anything not from the current slot anyway.

ajsutton: Yep, all good now.

rolfyone (approved):
LGTM

PR #5443, opened 2022-05-06
add onSlot eth1 deposit availability check and some other nits

tbenr:
## PR Description
adds a check wrt eth1 deposit availability and other little nits.

the check can be activated using a new `--Xeth1-missing-deposits-event-logging-enabled`

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
@@ -163,7 +163,6 @@ public class DepositProvider
               }
             })
         .reportExceptions();
-    ;
   }
 
   public synchronized SszList<Deposit> getDeposits(BeaconState state, Eth1Data eth1Data) {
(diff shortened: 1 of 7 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
```diff
@@ -126,24 +137,53 @@ public void onEth1Block(final Bytes32 blockHash, final UInt64 blockTimestamp) {
...
+                  depositNavigableMap.isEmpty()
+                      ? UInt64.ZERO
+                      : depositNavigableMap.lastKey().plus(ONE);
+              if (maxPossibleResultingDepositIndex.isLessThan(eth1DepositCount)) {
+                eventLogger.eth1DepositDataNotAvailable(
+                    maxPossibleResultingDepositIndex.plus(UInt64.ONE), eth1DepositCount);
+              }
+            })
+        .reportExceptions();
+    ;
```
ajsutton: Stray `;`

ajsutton (approved):
LGTM. Feature toggle probably makes sense initially but I'd be tempted to default it on anytime we have an eth1-endpoint.  Might just need to tweak the wording of the warning to allow for the possibility that the deposit fetching just hasn't completed yet. My thinking is that if there's something wrong and users aren't getting deposits we want to warn them well before they get a block proposal.

tbenr:
The reason why I added the CLI is because it's currently guaranteed a lot of noise on startup during the initial deposit retrieval. Wasn't sure it's ok for the end user.

PR #5444, opened 2022-05-07
missing deposit event log improvements

tbenr:
avoid checking deposit when still syncing
when depositNavigableMap is empty, defaults to last processed deposit in state
added happy path unit test

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5449, opened 2022-05-09
Use justified checkpoint as safe head.

### LINKED ISSUE #5427: Send justified payload block hash as safe block hash
### Description
When sending FCU updates to the EL the safe block hash should be set to the justified checkpoint. Previously it was the same as the head block hash - need to check we did actually change it.

ajsutton:
## PR Description
When sending fork choice updated to the execution engine, the safe head is now set to the justified checkpoint rather than the head block.

## Fixed Issue(s)
fixes #5427 

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #5452, opened 2022-05-09
LimitedMap and LimitedSet are no longer thread safe for iteration


ajsutton:
### Description
As part of an earlier optimisation `LimitedMap` was changed to use `SynchronizedLimitedMap` instead of guava `Cache`. While this map is thread safe for atomic operations like get/put, it is not support thread safe iteration (including via `.stream()`.

We need to audit the code to check if there are cases where we are iterating and fix. Potentially should default to the fully thread safe Cache and provide an option to use the synchronised version only when it is safe and has performance benefits.

PR #5453, opened 2022-05-09
Fix ConcurrentModificationException in SeenAggregatesCache

ajsutton:
## PR Description
Use fully thread safe set in SeenAggregatesCache since it needs to iterate the values. We're no longer limiting the total size of the internal sets but they're limited by the number of aggregators anyway so it's still safe (previously I just liked the defence in depth approach because it was essentially free).

Logged #5452 to review for other potential cases of this bug since `LimitedMap` isn't fully thread safe anymore.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5455, opened 2022-05-09
Migrate GetAttestations to MigratingEndpointAdapter

### LINKED ISSUE #5450: Migrate GetAttestations to MigratingEndpointAdapter

courtneyeh:
fixes #5450 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetAttestations.java
```diff
@@ -63,16 +93,28 @@ public GetAttestations(final DataProvider dataProvider, final JsonProvider jsonP
...
-    List<Attestation> attestations =
-        nodeDataProvider.getAttestations(maybeSlot, maybeCommitteeIndex);
-    ctx.json(jsonProvider.objectToJSON(new GetAttestationsResponse(attestations)));
+    adapt(ctx);
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
+    Optional<UInt64> slot = request.getOptionalQueryParameter(SLOT_PARAMETER);
+    Optional<UInt64> committeeIndex = request.getOptionalQueryParameter(COMMITTEE_INDEX_PARAMETER);
```
rolfyone: final

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetAttestations.java
```diff
@@ -63,16 +93,28 @@ public GetAttestations(final DataProvider dataProvider, final JsonProvider jsonP
...
+
+    List<Attestation> attestations = nodeDataProvider.getAttestations(slot, committeeIndex);
```
rolfyone: i'd probably not create this variable, just 
`request.respondOk(nodeDataProvider.getAttestations(slot, committeeIndex));`

data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
```diff
@@ -89,11 +88,10 @@ public NodeDataProvider(
     this.payloadAttributesCalculator = payloadAttributesCalculator;
   }
 
-  public List<Attestation> getAttestations(
+  public List<tech.pegasys.teku.spec.datastructures.operations.Attestation> getAttestations(
       Optional<UInt64> maybeSlot, Optional<UInt64> maybeCommitteeIndex) {
     return attestationPool
         .getAttestations(maybeSlot, maybeCommitteeIndex)
-        .map(Attestation::new)
         .collect(Collectors.toList());
```
rolfyone: we should just be able to return `attestationPool.getAttestations` now i think...

courtneyeh: That would return a `Stream<Attestation>`, which isn't compatible with the return type of `List<Attestation>`. So would either have to do some refactoring, or collect to list elsewhere.

rolfyone: Ok it's doing a bunch of filtering etc on the other end which is why its a stream, and it was probably left as a stream because we needed to map it anyway.
If we don't need to map it now, we can change it back to a list being returned (which will effect a couple of tests) and then its significantly cleaner over this side of the fence...

rolfyone (approved):
LGTM

PR #5458, opened 2022-05-09
[MEV Boost\Builder] Builder local EL fallback

### LINKED ISSUE #5103: Adopt `BlindedBeaconBlocks` in standard block production flow (non-MEVBoost)

tbenr:
- updates `engine` API timeouts as per https://github.com/ethereum/execution-apis/pull/216
- introduces other specific timeouts for `builder` APIs
- makes `BlockOperationSelectorFactory` mev-boost-enabled (builder endpoint) agnostic

fixes #5103

related to #5396 

## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
```diff
@@ -247,23 +281,80 @@ public SafeFuture<ExecutionPayloadHeader> builderGetHeader(
...
+      final SignedBeaconBlock signedBlindedBeaconBlock) {
+
+    checkArgument(
+        signedBlindedBeaconBlock.getMessage().getBody().isBlinded(),
+        "SignedBeaconBlock must be blind");
+
+    final UInt64 slot = signedBlindedBeaconBlock.getSlot();
+
+    final Optional<Optional<SafeFuture<ExecutionPayload>>> maybeFallbackPayload =
+        Optional.ofNullable(slotToLocalElFallbackData.get(slot));
```
ajsutton: The `Optional<Optional<..>>` makes this a bit confusing, particularly with the `.get().get()` later. I can see why it's needed and is probably better than an Optional var that is null.

It may be clearer if we can "unwrap" each level to a local var as we go though so we don't have `.get()` calls through the code.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
```diff
@@ -247,23 +281,80 @@ public SafeFuture<ExecutionPayloadHeader> builderGetHeader(
...
+                spec.atSlot(slot)
+                    .getSchemaDefinitions()
+                    .toVersionBellatrix()
+                    .orElseThrow()
+                    .getExecutionPayloadHeaderSchema()
+                    .createFromExecutionPayload(executionPayload))
+        .thenPeek(
+            __ -> {
+              // store the fallback payload for this slot,
+              slotToLocalElFallbackData.put(slot, Optional.of(localExecutionPayload));
```
ajsutton: It feels a little weird that we're storing a `SafeFuture` here but only after it is completed. I'm not sure if that's opening a potential race condition and we should be storing the SafeFuture before it completes or if we could remove ambiguity by storing the executionPayload rather than `SafeFuture<ExecutionPayload>` - even though we'll wind up wrapping it in `completedFuture` when we wind up getting it out of the map again it makes the logic clearer that if it's there it's complete.

ajsutton (approved):
LGTM generally.  Just check we're storing the execution payload at the right time.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
tbenr: The reason why I added a new optional is because intellij complains null-testing an Optional.
I'll unwrap it

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
tbenr: oops, right. It was my last change to `put` in the `peek` and I haven't realize it.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
ajsutton: Yeah I figured.  To be clear, by "unwrap" I don't mean going back to null but just do essentially:

```
Optional<Optional<Foo>> start = l...
if (start.isEmpty()) {
  return bah;
}
Optional<Foo> middle = start.get();
if (middle.isEmpty()) {
  return bah;
}
Foo foo = middle.get();
```

Just so we don't wind up with .get().get() and .get().isEmpty() and other hard to decipher stuff.

tbenr: yes yes, I got it. See the latest commit :)

PR #5461, opened 2022-05-10
Make in-memory store cache sizes configurable with --X CLI options

ajsutton:
## PR Description
Add --X options to configure the size of block, state and checkpoint state caches.  These aren't really intended for users to understand but makes it a lot easier for us to test different values and tune them, and doesn't provide an escape hatch if there's an unusual network situation and the defaults turn out to be very inefficient.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5464, opened 2022-05-10
Migrate GetAttesterSlashings to MigratingEndpointAdapter

### LINKED ISSUE #5460: Migrate GetAttesterSlashings to MigratingEndpointAdapter

courtneyeh:
fixes #5460 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetAttesterSlashings.java
```diff
@@ -13,36 +13,55 @@
...
+        EndpointMetadata.get(ROUTE)
+            .operationId("getAttesterSlashings")
+            .summary("Get Attester Slashings")
+            .description(
+                "Retrieves attester slashings known by the node but not necessarily incorporated into any block.")
+            .tags(TAG_BEACON)
+            .response(
+                SC_OK,
+                "Request successful",
+                getResponseType(spec.getGenesisSpecConfig())) // TODO check correct
```
courtneyeh: ^^ Just need to check this is the correct approach to getting the `SpecConfig`

rolfyone: I'd probably pass just the spec and then if we need to get different configs at a future date it'll just be an implementation detail, but this is fine too...

rolfyone (approved):
LGTM, can change specConfig or we can change it later if we ever need to, I'm not super concerned about that.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetAttesterSlashings.java
rolfyone: either way the TODO needs to disappear :)

PR #5468, opened 2022-05-11
Aggregate public keys in a single pass

ajsutton:
## PR Description
Very minor optimisation when aggregating BLS public keys - we can check each key is valid as part of the loop that aggregates them rather than iterating the list twice.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5470, opened 2022-05-11
add GetFeeRecipient to validator api.

rolfyone:
- fixed a path parameter encoding issue
 - moved some common functions from test base class so that I could call on the validator side
 - added a runtime list of proposer fee recipients that is currently empty
 - fee recipient list is now cached to allow ad-hoc queries.
 - validator index provider can't be provided to the BeaconProposerPreparer until after object creation, so it now has an initialize
 - tests to show the merge logic for merging configuration and runtime fee recipient lists.

## Documentation

## Changelog


infrastructure/restapi/src/testFixtures/java/tech/pegasys/teku/infrastructure/restapi/StubRestApiRequest.java
```diff
@@ -120,11 +120,11 @@ private void setOptionalQueryParameter(final String parameter, final String valu
   }
 
   @Override
-  @SuppressWarnings("unchecked")
   public <T> T getPathParameter(final ParameterMetadata<T> parameterMetadata) {
     assertThat(this.pathParameters.containsKey(parameterMetadata.getName())).isTrue();
-    final Object param = pathParameters.get(parameterMetadata.getName());
-    return (T) param;
+    return parameterMetadata
```
rolfyone, line 125: previously handled strings fine, which covers beacon-api, but the new parameter results in a BlsPublicKey directly so this fn had to get enhanced to use the deserializer

validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
```diff
@@ -38,39 +39,102 @@ public class BeaconProposerPreparer implements ValidatorTimingChannel {
...
+      Optional<Eth1Address> defaultFeeRecipient,
+      Spec spec) {
     this.validatorApiChannel = validatorApiChannel;
     this.validatorIndexProvider = validatorIndexProvider;
     this.proposerConfigProvider = proposerConfigProvider;
     this.defaultFeeRecipient = defaultFeeRecipient;
     this.spec = spec;
   }
 
+  public void initialize(final Optional<ValidatorIndexProvider> provider) {
```
ajsutton, line 80: Feels odd to specify this in the constructor and here.  And possibly this is more honest to just call `setValidatorIndexProvider`?

validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
```diff
@@ -38,39 +39,102 @@ public class BeaconProposerPreparer implements ValidatorTimingChannel {
...
   }
 
+  public Optional<Eth1Address> getFeeRecipient(final BLSPublicKey publicKey) {
+    if (validatorIndexProvider.isEmpty()) {
+      return Optional.empty();
+    }
+    final Optional<Integer> validatorIndex =
+        validatorIndexProvider.orElseThrow().getValidatorIndex(publicKey);
+    return validatorIndex.flatMap(this::getFeeRecipientByValidatorIndex);
+  }
```
ajsutton, line 102: hmm, I think we need to support setting and getting a fee recipient even before the validator index is known. You should be able to configure your validator client appropriately before making the deposit rather than having to wait for the validator to be active.

That probably means the config always has to be indexed by public key, not validator index. Not sure what the knock on effects of that will be though.

ajsutton (approved):
Approving as this is a big enough first step. We'll switch to indexing by public key as a follow up.

PR #5475, opened 2022-05-11
Move web3j logic in executionclient module to a separate package

StefanBratanov:
## PR Description
Move web3j logic in `executionclient` module to `web3j` package inside the module.

## Fixed Issue(s)
related to #5396

## Documentation

## Changelog


.gitignore
```diff
@@ -30,6 +30,7 @@ tmp/
 build/
 out/
 web3j/
+!**/executionclient/web3j/
```
ajsutton: We can probably remove this ignore entirely. I think it used to be there because we generated the deposit contract APIs via web3j but we don't do that anymore.

ajsutton (approved):
LGTM.

.gitignore
StefanBratanov: Great, will remove it then. Would be cleaner.

PR #5481, opened 2022-05-12
Optimise storage of attestations waiting to be applied to fork choice.

### LINKED ISSUE #5306: Improve handling of pending attestations
### Description
Attestations cannot be processed until the slot after they are published.  Currently teku achieves this by putting them in the pending attestation pool and then processing them at the start of the next slot. This winds up duplicating a number of checks and could be significantly optimised.

Rather than storing the entire attestation, we only need to track the information that would be required to apply the vote to fork choice.  Specifically, validator index, beacon block root and target epoch (see `ForkChoiceStrategy.processAttestation`).

Alternatively, it may be possible to apply the attestations to the `VoteTracker` instances but ensure that fork choice doesn't apply those scores until the start of the next slot.

Note that this applies specifically to `DEFER_FORK_CHOICE_PROCESSING` results, not `SAVED_FOR_FUTURE` which are attestations for a slot that hasn't yet started and will have deferred nearly all validation until that future slot is received.

ajsutton:
## PR Description
Attestations can only be applied to fork choice in the slot after they were created. Previously the whole attestation was stored in the generic future attestation pool. Now the minimal info required to update protoarray is stored in a dedicated structure for attestations that have already been added to the aggregating pool but not applied to fork choice.

## Fixed Issue(s)
fixes #5306 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5482, opened 2022-05-12
Be more flexible when selecting state to validate attestation with

### LINKED ISSUE #5477: Be more flexible when choosing state for attestation validation in fork choice
### Description
`ForkChoice.onAttestation` always retrieves the state to validate with using the target checkpoint from the attestation. While this will normally be shared between a lot of attestations, particularly for the first slot of the epoch this may require us to perform an additional epoch transition when we could have just used a state from the previous epoch.

We should review for safety but can likely use `AttestationStateSelector` to be more flexible in choosing the state to use, potentially allowing us to avoid regenerating states or performing additional epoch transitions.

ajsutton:
## PR Description
Update `ForkChoice.onAttestation` to use `AttestationStateSelector` to select the state to validate with. This lets it optimise for picking a state that is already available whenever possible.

## Fixed Issue(s)
fixes #5477 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTest.java
+++ b/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTest.java
@@ -838,9 +838,7 @@ class ForkChoiceTest {
     assertThat(result)
         .isCompletedWithValue(
             AttestationProcessingResult.invalid(
-                String.format(
-                    "Checkpoint state (%s) must be at or prior to checkpoint slot boundary (%s)",
-                    targetBlock.getSlot(), targetCheckpoint.getEpochStartSlot(spec))));
+                "LMD vote must be consistent with FFG vote target"));
   }
 
   private UInt64 applyAttestationFromValidator(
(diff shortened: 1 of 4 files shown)


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTest.java
```diff
@@ -838,9 +838,7 @@ void onAttestation_shouldBeInvalidWhenInvalidCheckpointThrown() {
     assertThat(result)
         .isCompletedWithValue(
             AttestationProcessingResult.invalid(
-                String.format(
-                    "Checkpoint state (%s) must be at or prior to checkpoint slot boundary (%s)",
-                    targetBlock.getSlot(), targetCheckpoint.getEpochStartSlot(spec))));
+                "LMD vote must be consistent with FFG vote target"));
```
ajsutton, line 841: We don't pick up this error as part of requesting the state anymore - it gets picked up by the actual validation process.  So same rule applied but it gets the rejection message from the spec rather than something specific to getting the checkpoint state.

rolfyone (approved):
LGTM

PR #5485, opened 2022-05-12
Migrate GetBlockRoot to MigratingEndpointAdapter

### LINKED ISSUE #5484: Migrate GetBlockRoot to MigratingEndpointAdapter

courtneyeh:
fixes #5484 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetBlockRoot.java
```diff
@@ -13,46 +13,74 @@
...
+          .withField("root", BYTES32_TYPE, Function.identity())
+          .build();
+
+  private static final SerializableTypeDefinition<ObjectAndMetaData<Bytes32>> RESPONSE_TYPE =
+      SerializableTypeDefinition.<ObjectAndMetaData<Bytes32>>object()
+          .name("GetBlockRootResponse")
+          .withField("data", ROOT_TYPE, ObjectAndMetaData::getData)
+          // TODO remove below? Eth docs have this, but causes fail of GetBlockRootIntegrationtest
+          // .withField("execution_optimistic", BOOLEAN_TYPE,
+          // ObjectAndMetaData::isExecutionOptimistic)
```
rolfyone: i think if we're supposed to provide it that we just add it in and fix the test

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetBlockRoot.java
```diff
@@ -73,16 +101,22 @@ public GetBlockRoot(final ChainDataProvider chainDataProvider, final JsonProvide
...
 
-  private Optional<String> handleResult(Context ctx, final ObjectAndMetaData<Root> response)
-      throws JsonProcessingException {
-    return Optional.of(
-        jsonProvider.objectToJSON(
-            new GetBlockRootResponse(response.getData(), response.isExecutionOptimisticForApi())));
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
+    final SafeFuture<Optional<ObjectAndMetaData<Bytes32>>> future =
+        chainDataProvider.getBlockRoot(request.getPathParameter(PARAMETER_BLOCK_ID));
```
rolfyone, line 109: we could follow the same pattern of getting the block and then using the call to get the root here if we wanted... would fit into a generic handler then... just a thought...

rolfyone: There's only 3 I guess so it might not be worth it but they're all v. similar in that they just pull data off the block

courtneyeh: Yes might not be worth it yet, but will keep it in mind for the future

rolfyone (approved):
LGTM

PR #5492, opened 2022-05-13
Serialize events to bytes instead of String

ajsutton:
## PR Description
Previously a single String instance was stored and a new byte[] created for every subscriber. Now a single byte[] is stored and reused for every subscriber.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5499, opened 2022-05-13
[Mev Boost\Builder] Makes proposer public key available for `getHeader`

tbenr:
- `PayloadAttributesCalculator` becomes `ProposersDataManager` which is now also responsible of forwarding  registration data to builder api. This data will be also used to determine if a proposer has been configured to use builder api. "prepare_beacon_proposer" and `register_validator` are not currently not supporting the option of disabling builder api. if we want to I think we may add a new attribute in "prepare_beacon_proposer" (like a `teku` object holding our custom config) which may contain a disable\enable flag.
- enrich `PayloadAttributes` (now `PayloadBuildingAttributes`) with `ValidatorRegistration` information
- various refactoring
- updates the experimental teku api to retrieve current proposer data (which now includes `prepared_proposer` and `registered_validator`)

Seems to me that in `ValidatorApiHandler` we need to remove `ForkChoiceNotifier` and add  `ProposersDataManager` 
which will handle prepare and register apis. A subscription should notify the `ForkChoiceNotifier` when prepared proposer has been called.

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ProposersDataManager.java
```diff
@@ -17,34 +17,43 @@
...
   private static final long MAX_PROPOSER_SEEN_EPOCHS = 2;
 
   private final Spec spec;
   private final EventThread eventThread;
   private final RecentChainData recentChainData;
-  private final Map<UInt64, ProposerInfo> proposerInfoByValidatorIndex = new ConcurrentHashMap<>();
+  private final Map<UInt64, PreparedProposerInfo> preparedProposerInfoByValidatorIndex =
+      new ConcurrentHashMap<>();
+  private final Map<UInt64, SignedValidatorRegistrationV1>
+      validatorRegistrationInfoByValidatorIndex = new ConcurrentHashMap<>();
```
ajsutton: If these need to stay in sync we probably should just have a single map and use a wrapper class to hold both items as the value.  Otherwise we have to go to a fair bit of effort to avoid race conditions where a value is in on map but not the other.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ProposersDataManager.java
```diff
@@ -55,25 +64,56 @@ public PayloadAttributesCalculator(
...
+      final Collection<BeaconPreparableProposer> preparedProposers, final UInt64 currentSlot) {
     // Remove expired validators
-    proposerInfoByValidatorIndex.values().removeIf(info -> info.hasExpired(currentSlot));
+    final Iterator<Entry<UInt64, PreparedProposerInfo>> iterator =
+        preparedProposerInfoByValidatorIndex.entrySet().iterator();
+    while (iterator.hasNext()) {
+      final Entry<UInt64, PreparedProposerInfo> entry = iterator.next();
+      if (entry.getValue().hasExpired(currentSlot)) {
+        validatorRegistrationInfoByValidatorIndex.remove(entry.getKey());
+        iterator.remove();
```
ajsutton: This is an example of a potential race condition. It's possible the entry isn't in `validatorRegistrationInfoByValidatorIndex` yet but is added later and we then wouldn't prune it. 

If they don't always line up then we probably need to handle expiry of validator registrations separately. It does look like we add to the maps separately so probably do need to just always treat them completely separately.

ajsutton (commented):
Looks good generally but not sure we're pruning validator registration info correctly.

ajsutton:
> if we want to I think we may add a new attribute in "prepare_beacon_proposer" (like a teku object holding our custom config) which may contain a disable\enable flag.

I think generally the model is that the beacon node controls whether a builder is used and which one. If there's a requirement to decide that per-validator we should probably get it added to the standard validator APIs (probably as an optional section).  I'm not keen to have teku specific extensions there if we can avoid it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ProposersDataManager.java
tbenr: yeah. I rechecked how the registration api against the builder infra is supposed to be used. It should follow a regular poll, so at the end i'm going to keeping them separated, mirroring the api separation.

tbenr:
> I think generally the model is that the beacon node controls whether a builder is used and which one. 

ah that makes sense. So activating\deactivating mev for a given validator turn to be moving from\to a VC that is using a BN with builder enabled\disabled.

ajsutton (approved):
LGTM.

PR #5500, opened 2022-05-13
Add propertyTest task

jtraglia:
## PR Description

This new task will run all of our Property-Based Testing (PBT) tests. With PBT, you test a certain property by defining constraints and letting the test engine produce randomized inputs; it's a lot like fuzzing. I discovered this concept while looking for an equivalent to [Golang's `go fuzz` feature](https://go.dev/doc/fuzz/), introduced in Go 1.18, that would work with Teku and its dependencies.

These first few property tests are relatively simple. I had a difficult time deciding where to start. I know that this area is already pretty well tested, most of Teku is pretty well tested IMO, but I needed to start somewhere. The plan is to add more PBT tests to existing classes in the near future.

The PBT test engine that I chose was not my first choice. Initially, I evaluated [JQF](https://github.com/rohanpadhye/JQF) which features randomized testing with coverage-guided fuzzing algorithms and the ability to run a certain test indefinitly.  Unfortunately, JQF does not work well with Teku. It requires the JUnit4 "vintage" platform and when we use the newer JUnit5 "jupiter" platform. Also, JQF is pretty tightly integrated with Maven and does not really support Gradle.  There is [a Gradle plugin](https://github.com/mborzenkov/jqf-gradle-plugin) but it's still very much in development; it doesn't support project structures like Teku.

Then, I evaluated `java-quickcheck`, which JQF is based on. As expected, it had the same problems as JQF. It doesn't feature coverage-guidance or anything cool like that but it would have still been useful.

Lastly, I tested `jqwik` which is an alternative PBT test engine that is designed to work for the JUnit5 platform. In my opinion, it has really great documentation (see the [user guide](https://jqwik.net/docs/current/user-guide.html) and [javadoc](https://jqwik.net/docs/1.6.5/javadoc/index.html)) and, from the best I can tell, it has all of the same features as `java-quickcheck`. Another good quality, there's an additional `jqwik-kotlin` module that is supposed to simplify/streamline using `jqwik` in Kotlin projects. This will be useful if we decide to introduce PBT into libp2p in the future as it is primarily written in Kotlin.

## Documentation

## Changelog


build.gradle
```diff
@@ -763,6 +776,7 @@ subprojects {
     testImplementation 'org.mockito:mockito-core'
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
     testImplementation 'org.junit.jupiter:junit-jupiter-params'
+    testImplementation 'net.jqwik:jqwik'
```
ajsutton: We probably don't want to add jqwik as a dependency for all unit tests.  We could add it automatically for `propertyTestImplementation` though as we'd expect it to be used by all property tests.
```suggestion
    propertyTestImplementation 'net.jqwik:jqwik'
```

storage/src/property-test/java/tech/pegasys/teku/storage/server/kvstore/serialization/BeaconStateSerializerPropertyTest.java
```diff
@@ -0,0 +1,44 @@
...
+      @ForAll final SpecMilestone milestone,
+      @ForAll final Eth2Network network,
+      @ForAll @IntRange(max = 1000) final int validatorCount) {
+    final Spec spec = TestSpecFactory.create(milestone, network);
+    final DataStructureUtil dataStructureUtil = new DataStructureUtil(seed, spec);
+    final KvStoreSerializer<BeaconState> serializer = createStateSerializer(spec);
+    final BeaconState value = dataStructureUtil.randomBeaconState(validatorCount);
+    final byte[] serialized = serializer.serialize(value);
+    final BeaconState deserialized = serializer.deserialize(serialized);
+    assertThat(deserialized).isEqualTo(value);
```
ajsutton: Sadly jqwick's reporting for tests in this style is pretty poor - it doesn't include any of the info about what the parameters that triggered the failure were. While we can get the seed by running gradle with `--info` that also produces a *ton* of other output so it's hard to parse.

Fortunately for cases where there's one fairly simple condition like this we can use a property based test which does give the right info by just returning a boolean. ie:

```java
return deserialised.equals(value);
```

The downside is that we don't get the actual result values printed but at least we can reproduce the problem easily and can then locally modify the test to print whatever other info is useful.  The other option I guess is to use AssertJ's `.describedAs` method to include the parameters in it's exception message but I suspect we'll regularly forget to do that or miss params etc.

ajsutton (commented):
Looks good. I think it's worth changing to use the property form of tests as getting the info required to reproduce the failure is more important than the diff I think.

I can see from the jqwick docs that custom LifeCycleHook (https://jqwik.net/docs/current/user-guide.html#lifecycle-hooks) and I wonder if it's possible to create one of those that will catch any test failures and add the specific property values (and seed) to the error message.  Then the assert based tests would automatically include the info required to reproduce them.  I didn't actually dig deep enough to know if that's possible or what it would take though - might be a dead end.

I've also added the circleCI config to actually run the property tests so will see if I got that right...

storage/src/property-test/java/tech/pegasys/teku/storage/server/kvstore/serialization/BeaconStateSerializerPropertyTest.java
jtraglia: @ajsutton there's also the [reporter object](https://jqwik.net/docs/current/user-guide.html#platform-reporting-with-reporter-object) and [footnotes](https://jqwik.net/docs/current/user-guide.html#adding-footnotes-to-failure-reports). I will investigate more in a bit.

build.gradle
jtraglia: Good call. I will fix this.

storage/src/property-test/java/tech/pegasys/teku/storage/server/kvstore/serialization/BeaconStateSerializerPropertyTest.java
jtraglia: > it doesn't include any of the info about what the parameters that triggered the failure were. While we can get the seed by running gradle with --info that also produces a ton of other output so it's hard to parse.

So I'm not entirely sure what you mean here. If I force a test to fail, I can see the arguments (`arg0`, `arg1`) that resulted in the failure. This is a screenshot of the `./gradlew propertyTest --info` output I see:

forced_failure_output

Which is similar to the output in an IDE (like IDEA) but doesn't require the `--info` flag:

forced_failure_output_idea

jtraglia: Oh yeah, I see what you mean here now. In the tests that use `dataStructureUtil` to generate a large structure, the failure output is not very good. It will print the entire structure. I don't think this is necessarily jqwik's fault though.

In other property tests, the output looks better. The following (`./gradlew propertyTest --info`) shows the user the generated arguments that were used to cause the test to fail.

forced_failure_output

jtraglia: So there were three tests that would print a lot to stdout if they failed. I updated those to not use asserts, but rather return a boolean like you mentioned. I wasn't a huge fan of the alternative fixes. I don't think we need to worry about the other tests; their outputs could be a little long, but nothing like printing a beacon state object! What do you think? Is this good enough?

ajsutton: I think the issue is more that currently we don't run gradle with the `--info` option and when we do we get a ton of extra logging like:

```
> Task :networking:nat:compileJava UP-TO-DATE
Excluding [io.netty:netty-tcnative-boringssl-static, io.netty:netty-transport-rxtx]
Excluding []
Caching disabled for task ':networking:nat:compileJava' because:
  Build cache is disabled
Skipping task ':networking:nat:compileJava' as it is up-to-date.
:networking:nat:compileJava (Thread[Execution worker for ':' Thread 8,5,main]) completed. Took 0.238 secs.
:networking:nat:classes (Thread[Execution worker for ':' Thread 8,5,main]) started.
```

If we could just enable info logging for the actual property test task that would be fine but we get it enabled for every task along the way as well.

But I'm tempted to just get started like this and we can improve over time. I suspect something that hooks into the life cycle is probably the best long term answer so we can write tests however we like and get the seed in the failure message.

ajsutton (approved):
LGTM. I've enabled --info when running propertyTests in CI for now and we'll see what happens. If we find this frustrating we can look into ways to make it easier to get the seed and reproduce failures later.

PR #5505, opened 2022-05-16
Move VC GetGenesis to be called via type definitions

rolfyone:
- Also changed HttpErrorResponse to be deserializable

 - Some of the non typedef functionality such as ResponseHandler has been re-implemented to cater for type definitions. It is likely that the apiclient version gets removed once typedefs are used throughout, as the new version doesn't require a JsonProvider.

## Documentation

## Changelog


validator/remote/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
   implementation project(':infrastructure:http')
   implementation project(':infrastructure:logging')
   implementation project(':infrastructure:metrics')
+  implementation project(':infrastructure:restapi')
```
ajsutton: hmm, slightly unfortunate to need to depend on `:infrastructure:restapi` here rather than just `:infrastructure:json`.  Ideally we'd only depend on restapi when we're actually implementing a restapi (ie it's the server component and we can use type definitions in a client without it).  Not sure how feasible that is though.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/ValidatorRestApiClient.java
```diff
@@ -0,0 +1,21 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.validator.remote.typedef;
+
+import java.util.Optional;
+import tech.pegasys.teku.spec.datastructures.genesis.GenesisData;
+
+public interface ValidatorRestApiClient {
```
ajsutton: Do we need this interface? I can see it would be helpful if we're feature toggling between the two implementations but if we're just doing a migration we probably don't need it?

ajsutton (approved):
LGTM. Mostly just worth reviewing the dependencies are what we actually need.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/ValidatorRestApiClient.java
rolfyone: true, the way its structured this could just be deleted, I guess i was mirroring the other apiclient...

validator/remote/build.gradle
rolfyone: doesnt look like we need it this PR, removed.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/ValidatorRestApiClient.java
rolfyone: deleted interface

PR #5507, opened 2022-05-16
Fix configuration for boolean cli options

ajsutton:
## PR Description
Ensure all boolean CLI options have showDefaultValues = ALWAYS and fallbackValue = "true".

That ensures the default value is shown in help and that `--option-enabled` with no argument always enables the value instead of toggling it from its default.

Also adds the new `--Xfork-choice-before-proposing-enabled` which is what I was working on before discovering the config issues.  A separate PR will make it actually do something (which is fine because it's hidden).

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5509, opened 2022-05-16
implement CreateBlockRequest with typedefs

rolfyone:
- currently only returning JSON which is consistent with existing API.

 - we're back to using JsonUtil to decode, which required pulling in the infrastructure:json package.

 Continues to support blinded block requests.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionCache.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionCache.java
@@ -32,6 +32,10 @@ public class SchemaDefinitionCache {
     return schemas.computeIfAbsent(milestone, this::createSchemaDefinition);
   }
 
+  public SchemaDefinitions atSlot(final UInt64 slot) {
+    return getSchemaDefinition(milestoneAtSlot(slot));
+  }
+
   public final SpecMilestone milestoneAtSlot(final UInt64 slot) {
     return spec.atSlot(slot).getMilestone();
   }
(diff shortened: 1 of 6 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionCache.java
```diff
@@ -32,6 +32,10 @@ public SchemaDefinitions getSchemaDefinition(final SpecMilestone milestone) {
     return schemas.computeIfAbsent(milestone, this::createSchemaDefinition);
   }
 
+  public SchemaDefinitions atSlot(final UInt64 slot) {
+    return getSchemaDefinition(milestoneAtSlot(slot));
+  }
```
ajsutton, line 37: I kind of see how this makes sense, but seems like something we should just get from `Spec` (since if the milestone is active at a slot it's guaranteed to be in Spec).

ajsutton: Actually given we only ever need to handle schemas that are actually active in Spec we can get rid of the dependency on `SchemaDefinitionsCache` entirely and just use `Spec`.  `SchemaDefinitions` only exists to create schema definitions for milestones that aren't enabled in this network so we can use them to define APIs, but we don't need that in the client - only the actually active milestones.

ajsutton (approved):
LGTM.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionCache.java
rolfyone: ah ok great - i'll just pass the specConfig like i did in POC, i was worried i'd be creating overheads of constantly getting typedefs...

PR #5510, opened 2022-05-16
Implement SendSignedBlockRequest with typedefs

rolfyone:
- currently only returning JSON which is consistent with existing API.

 Continues to support blinded blocks.

## Documentation

## Changelog


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
```diff
@@ -0,0 +1,48 @@
...
+        signedBeaconBlock.getMessage().getBody().isBlinded()
+            ? SEND_SIGNED_BLINDED_BLOCK
+            : SEND_SIGNED_BLOCK;
+    return postJson(
+            apiMethod,
+            Collections.emptyMap(),
+            signedBeaconBlock,
+            signedBeaconBlock.getSchema().getJsonTypeDefinition(),
+            new ResponseHandler<>())
+        .map(__ -> SendSignedBlockResult.success(Bytes32.ZERO))
```
tbenr: you're not fully handling the response right?

rolfyone: by this point it either has an internal block, or it failed... the post doesn't get data back unless there's an error...

ajsutton: I think we used to return `Bytes32.ZERO` as the root here anyway, but we could actually provide the real block root from `signedBeaconBlock.hashTreeRoot()` here which would probably be slightly better in case we ever do display that root somewhere.

ajsutton (approved):
LGTM.

PR #5513, opened 2022-05-16
Request octet-streams be sent from createUnsignedBlock

rolfyone:
- although the header is set to request octet-stream, will still process the content type on response, in case the BN didn't actually return octet-streams. This should let us have this mediatype set, and still allow us to use nodes that don't support octet-stream.

  - when we're sending the signed block to submit (different PR) we'll potentially need to be looking at what is returned here to make a decision on whether to attempt to send an octet-stream signed block.

  partially addresses #5400

## Documentation

## Changelog


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/CreateBlockRequest.java
```diff
@@ -85,12 +84,18 @@ public Optional<BeaconBlock> createUnsignedBlock(
       final BLSSignature randaoReveal, final Optional<Bytes32> graffiti) {
     final Map<String, String> queryParams = new HashMap<>();
     queryParams.put("randao_reveal", randaoReveal.toString());
+    final Map<String, String> headers = new HashMap<>();
     graffiti.ifPresent(bytes32 -> queryParams.put("graffiti", bytes32.toHexString()));
 
+    if (preferSszBlockEncoding) {
+      headers.put("Accept", "application/octet-stream");
```
ajsutton: We should make this explicitly state that we'd prefer octet-stream but will accept json.  Otherwise we run the risk of some clients not automatically falling back to JSON.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorClientOptions.java
```diff
@@ -29,8 +32,22 @@ public class ValidatorClientOptions {
...
+      arity = "0..1",
+      fallbackValue = "true")
+  private boolean validatorClientSszBlocksEnabled = DEFAULT_VALIDATOR_CLIENT_SSZ_BLOCKS_ENABLED;
+
   public void configure(TekuConfiguration.Builder builder) {
-    builder.validator(config -> config.beaconNodeApiEndpoint(parseApiEndpoint()));
+    builder.validator(
+        config -> {
+          config.beaconNodeApiEndpoint(parseApiEndpoint());
+          config.validatorClientUseSszBlocksEnabled(validatorClientSszBlocksEnabled);
```
ajsutton: nit: I think we can just chain these:
```suggestion
          config.beaconNodeApiEndpoint(parseApiEndpoint())
                  .validatorClientUseSszBlocksEnabled(validatorClientSszBlocksEnabled);
```

I guarantee to have broken spotless with that...

ajsutton (approved):
LGTM.

PR #5517, opened 2022-05-17
enable sending signed blocks as SSZ from VC

rolfyone:
- will fallback to JSON, and stop trying SSZ if it gets a 406 response. This will save the effort of getting continual 406's and having to send twice constantly.  This is only in the case of a 406 being sent back.

Partially addresses #5400

## Documentation

## Changelog


infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
```diff
@@ -86,6 +86,12 @@ public void respondError(final int statusCode, final String message)
     respond(statusCode, new HttpErrorResponse(statusCode, message));
   }
 
+  @Override
+  public Optional<String> getContentTypes() {
```
ajsutton: Seems odd that this has a plural name but only returns 0 or 1 results.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
```diff
@@ -13,28 +13,73 @@
...
             : SEND_SIGNED_BLOCK;
+
+    final SendSignedBlockResult result =
+        preferSszBlockEncoding
+            ? sendSignedBlockAsSszOrFallback(apiMethod, signedBeaconBlock)
+            : sendSignedBlockAsJson(apiMethod, signedBeaconBlock);
+    if (!result.isPublished() && retryAsJson) {
+      return sendSignedBlockAsJson(apiMethod, signedBeaconBlock);
+    }
+    return result;
```
ajsutton: It's probably worth changing the ?: to a full if and then only checking the retryAsJson inside the branch that sends SSZ.  It works correctly now but feels a little odd that we'd even consider retrying as JSON after having just sent JSON.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
```diff
@@ -13,28 +13,73 @@
...
+  }
+
+  public boolean isRetryAsJson() {
+    return retryAsJson;
```
ajsutton: Probably worth renaming this and the variable - `retryAsJson` makes me think that it's a flag indicating if the request is allowed to retry as json if needed rather than a flag indicating it did need to retry.  Maybe something like `sszNotAcceptable`?

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
ajsutton: Also doesn't actually seem to be called anyway so maybe we can get rid of it?

ajsutton (approved):
LGTM.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
rolfyone: oh probably can, we did need it but i may have refactored that out...

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
rolfyone: yeh the logic does work out because of the fact that the retry is only set from a 406... you'll probably want to review the change i make in this spot and see if you're happier with it.

PR #5525, opened 2022-05-18
Serialize JSON directly to response output stream

ajsutton:
## PR Description
Avoid creating the full JSON content in memory by streaming it straight to the response output stream.

Also converted a whole bunch of tests over to use the stub request rather than a mock context because it's pretty much impossible to mock the context in the way we need to stream responses.

One downside is that while JSON content is streamed, SSZ is not currently because `SszData` only provides a method that returns the SSZ as `Bytes`. SSZ tends to be much more compact though so is less of an issue.  Get state validators on the other hand returns 181Mb of JSON.

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/AbstractGetNewBlockTest.java
+++ b/data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/AbstractGetNewBlockTest.java
@@ -15,25 +15,18 @@ package tech.pegasys.teku.beaconrestapi;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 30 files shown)


rolfyone (approved):
LGTM

Nashatyrev:
> One downside is that while JSON content is streamed, SSZ is not currently because `SszData` only provides a method that returns the SSZ as `Bytes`. SSZ tends to be much more compact though so is less of an issue. Get state validators on the other hand returns 181Mb of JSON.

Just for the record there is `SszData.sszSerialize(SszWriter)` method which should actually stream SSZ. The one returning `Bytes` is just a shortcut

ajsutton:
> Just for the record there is SszData.sszSerialize(SszWriter) method which should actually stream SSZ. The one returning Bytes is just a shortcut

huh, I knew that was there under the covers I didn't realise it was so easily accessible.

PR #5526, opened 2022-05-18
RejectedExecution reporting and health effects

rolfyone:
- New metric for rejected execution:
 ```
 # HELP executor_rejected_execution_total Count of rejected executions by task queue
 # TYPE executor_rejected_execution_total counter
 executor_rejected_execution_total{name="p2p",} 12.0
 ```
  this is a counter, so will track from start the number of RejectedExecutionExceptions received.

 - A new counter tracks 10 minutes (5 second polling intervals) of RejectedExecutionExceptions (aggregated across all threads). The counter is pretty generic, it allows to specify the number of polling periods to track, and you'd be responsible for calling 'poll' when the period changes. It's been configured for 120 periods of 5 seconds (10 minutes rolling). The poll function will return the value stored for that period.

 - if the counter is non 0, the GetHealth query will return 503. This means once the first RejectedExecutionException occurs, the health check will start to respond 503. It would take 10 minutes of no RejectedExecutionExceptions for the node to be again marked healthy.

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/AbstractNode.java
```diff
@@ -124,13 +135,29 @@ public void start() {
...
     getServiceController().start().join();
+    counterMaintainer =
+        Optional.of(
+            serviceConfig
+                .createAsyncRunner("RejectedExecutionCounter", 1)
+                .runWithFixedDelay(
+                    this::pollRejectedExecutions,
+                    Duration.ofSeconds(5),
+                    (err) -> LOG.debug("rejected execution poll failed", err)));
+  }
```
rolfyone, line 145: not sure if there's a prettier way of doing this at this point, didn't seem to be anything obvious... (normally we could use `asyncRunner.runWithFixedDelay` but not here)

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/OccuranceCounter.java
```diff
@@ -0,0 +1,57 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.infrastructure.async;
+
+import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class OccuranceCounter {
```
rolfyone, line 19: not a huge fan of the name... had `RollingCounter` as an alternative, didn't really love either one...

data/provider/src/main/java/tech/pegasys/teku/api/SyncDataProvider.java
```diff
@@ -21,15 +22,21 @@
 public class SyncDataProvider {
 
   private final SyncService syncService;
+  private final IntSupplier rejectedExecutionSupplier;
 
-  public SyncDataProvider(SyncService syncService) {
+  public SyncDataProvider(SyncService syncService, final IntSupplier rejectedExecutionSupplier) {
```
ajsutton, line 27: Not entirely sure this fits into SyncProvider but not sure it's worth having a separate provider for it either. I guess it kind of makes sense since most rejected executions are from gossip which affects sync...

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/MetricTrackingExecutorFactory.java
```diff
@@ -13,28 +13,49 @@
...
+
+  private final Optional<OccuranceCounter> rejectedExecutionCounter;
 
   public MetricTrackingExecutorFactory(final MetricsSystem metricsSystem) {
+    this(metricsSystem, Optional.empty());
+  }
+
+  public MetricTrackingExecutorFactory(
+      final MetricsSystem metricsSystem,
+      final Optional<OccuranceCounter> rejectedExecutionCounter) {
```
ajsutton: Is there a reason we wouldn't always have an OccuranceCounter? Seems simpler to just require it and ignore the counts if need be there's basically no cost.

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/OccuranceCounter.java
ajsutton: I think it works - certainly can't think of anything better.

teku/src/main/java/tech/pegasys/teku/AbstractNode.java
ajsutton: It's kind of messy but yeah not really a much better option at this level. I think it's fine.

ajsutton (approved):
LGTM.

PR #5528, opened 2022-05-18
[MEV Boost\Builder] Replace the builder web3j client with a rest client

StefanBratanov:
## PR Description
- Create `RestClientProvider` similar to `Web3jClientProvider`
- Use the `RestClientProvider` in `ExecutionLayerService` instead of the Web3j one
- Delete the old Web3j builder client implementation
- Specified a builder timeout in `Constants` which is the maximum duration before timeout for all calls
- Some refactors around JwtConfig to make it reusable
- Some other small nits (ordering of variables, finals, etc)

## Fixed Issue(s)
related to #5396 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java
```diff
@@ -44,6 +44,10 @@ public class Constants {
   public static final Duration ETH1_DEPOSIT_REQUEST_RETRY_TIMEOUT = Duration.ofSeconds(2);
   public static final Duration EL_ENGINE_BLOCK_EXECUTION_TIMEOUT = Duration.ofSeconds(8);
   public static final Duration EL_ENGINE_NON_BLOCK_EXECUTION_TIMEOUT = Duration.ofSeconds(1);
+  // Maximum duration before timeout for each builder call
+  public static final Duration EL_BUILDER_CALL_TIMEOUT = Duration.ofSeconds(8);
+  // Individual durations (per method) before timeout for each builder call. They will be less than
+  // or equal to the maximum duration before timeout above.
```
tbenr:
```suggestion
  // Individual durations (per method) before timeout for each builder call. They must be less than
  // or equal to EL_BUILDER_CALL_TIMEOUT
```

tbenr (approved):
LGTM. just a nit in comments

PR #5533, opened 2022-05-19
Immediately return empty when state after the head slot is requested

ajsutton:
## PR Description
When requesting a state by slot, there's no point looking up one that is beyond the current chain head. Doing so may cause us to process empty slots but is only going to return misleading information.

Other clients also return not found in this case.

With thanks to @z3n-chada for reporting.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5535, opened 2022-05-19
[Mev Boost\Builder] implement builder registration call in `ProposersDataManager`

tbenr:
Implements the backend of the beacon rest api `register_validator` endpoint (ref: https://github.com/ethereum/beacon-APIs/pull/209)

since on the builder side the corresponding api do not accept array, this PR calls the builder api in sequence, one by one, stopping in case of networking issues or Timeout on SafeFuture. The reason is to avoid congesting the throttled rest client and delaying potential block building calls.

related to #5396 

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #5544, opened 2022-05-20
Fix a deserialization problem in SszBitvector

jtraglia:
## PR Description

I was creating property tests for all of the serialization/deseriation classes and came across this issue. When the serialized value is negative, the bit length is incorrect. I will share the property tests later after I fix some more potential bugs.

## Documentation

## Changelog


jtraglia:
Hold off on this. Something is not quite right.

build.gradle
```diff
@@ -176,6 +176,7 @@ allprojects {
       '-Xlint:finally',
       '-Xlint:static',
       '-Werror',
+      '-parameters',
```
jtraglia, line 179: Added this so that parameter names show up in property test failure output.

jtraglia:
So my original "fix" misidentified the problem and I don't know what the proper fix should be. I've added the tests I used to identify the problem. These are expected to fail. I'll continue to look at it, but if I can't get it I might need some help.

jtraglia:
I couldn't figure out how to correctly compute the bit length given the original serialized value, so I decided to redo it in a way that made sense to me. It is now serialized to a string of 0 and 1 values, like its `toString` output. The serialized value is less space efficient but potentially easier to read if required. Not sure what this will break but it's probably a lot. If someone could fix the previous implementation that would be preferred.

I think the problem is because the bitvector size is lost in translation. In other usages, the bit length is already known when deserializing. For example, the following:

https://github.com/ConsenSys/teku/blob/516444f5b43ff85530d32e7a3206acc369a8b8fb/infrastructure/ssz/src/test/java/tech/pegasys/teku/infrastructure/ssz/collections/SszBitvectorTest.java#L73-L74

It's my understanding that `.getSchema()` will keep the bitvector size.

ajsutton:
I'll need to look into the SSZ spec better to work this out.  It's definitely a bit of an issue trying to parse a Bitvector without actually having the right schema so it may not be possible to make work right in every case.  On the plus side we're migrating away from that form of JSON parsing and I think we may have already migrated the only APIs that use Bitvector. If not we could prioritise them and delete this implementation I suspect.

Will look in more detail Monday - it's going to take more focus than I'll get over the weekend. :)

jtraglia:
Sounds good! We'll deal with it then. Have a good weekend.

ajsutton:
Have put up #5555 to remove the deserialiser entirely since it isn't used in production anymore. I'm not sure there is a way to make it work reliably given it doesn't have the actual vector length from the schema.

jtraglia:
With that PR this is no longer an issue. Thanks!

jtraglia:
I will commit these new property tests in a separate PR.

PR #5545, opened 2022-05-20
[Mev Boost\Builder] validator registration beacon rest api (server side)

tbenr:
This draft PR introduces:
- new `SszList` of `SignedValidatorRegistration`
- rest API serving validator registration supporting json and ssz content types
- moved `ValidatorRegistration` and related Ssz schemas to `tech.pegasys.teku.spec.schemas.ApiSchemas`

in next PR we migrate all Ssz API objects (non-milestone specific) to this new class

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
@@ -400,7 +400,7 @@ public class BeaconRestApi {
     app.post(
         PostContributionAndProofs.ROUTE, new PostContributionAndProofs(dataProvider, jsonProvider));
     addMigratedEndpoint(new PostPrepareBeaconProposer(dataProvider));
-    addMigratedEndpoint(new PostRegisterValidator(dataProvider, spec));
+    addMigratedEndpoint(new PostRegisterValidator(dataProvider));
   }
 
   private void addBeaconHandlers(final DataProvider dataProvider, final Spec spec) {
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
@@ -96,6 +96,7 @@ import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostAggregateAndPro
 import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostAttesterDuties;
 import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostContributionAndProofs;
 import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostPrepareBeaconProposer;
+import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostRegisterValidator;
 import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostSubscribeToBeaconCommitteeSubnet;
 import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostSyncCommitteeSubscriptions;
 import tech.pegasys.teku.beaconrestapi.handlers.v1.validator.PostSyncDuties;
@@ -114,6 +115,7 @@ import tech.pegasys.teku.infrastructure.time.TimeProvider;
 import tech.pegasys.teku.infrastructure.version.VersionProvider;
 import tech.pegasys.teku.provider.JsonProvider;
 import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.datastructures.eth1.Eth1Address;
 import tech.pegasys.teku.spec.schemas.SchemaDefinitionCache;
 import tech.pegasys.teku.storage.client.ChainDataUnavailableException;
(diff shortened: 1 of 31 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostRegisterValidator.java
```diff
@@ -0,0 +1,175 @@
...
+    return schemaDefinitionCache
+        .getSchemaDefinition(SpecMilestone.BELLATRIX)
+        .toVersionBellatrix()
+        .orElseThrow()
+        .getSignedValidatorRegistrationsSchema()
+        .getJsonTypeDefinition();
+  }
+
+  @SuppressWarnings("JavaCase")
+  public static class SignedValidatorRegistrationOpenApiSchema {
```
rolfyone: you can put this in schema with the rest of the objects if you like, most of the time the annotated objects are outside of the class...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostRegisterValidator.java
```diff
@@ -0,0 +1,175 @@
...
+      @Schema(
+          type = "string",
+          pattern = PATTERN_EXECUTION_ADDRESS,
+          example = EXAMPLE_EXECUTION_ADDRESS,
+          description = DESCRIPTION_EXECUTION_ADDRESS)
+      Eth1Address fee_recipient;
+
+      @JsonProperty("gas_limit")
+      @Schema(type = "string", format = "uint64", example = EXAMPLE_UINT64)
+      UInt64 gas_limit;
```
rolfyone: most of the time schema objects have these as `public final`, the reason for that is then you don't also need to do the `@JsonProperty` annotation (or you need a getter, one of the 3 options)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
```diff
@@ -388,6 +390,9 @@ private void addValidatorHandlers(final DataProvider dataProvider, final Spec sp
     app.post(
         PostContributionAndProofs.ROUTE, new PostContributionAndProofs(dataProvider, jsonProvider));
     addMigratedEndpoint(new PostPrepareBeaconProposer(dataProvider));
+    if (spec.isMilestoneSupported(SpecMilestone.BELLATRIX)) {
+      addMigratedEndpoint(new PostRegisterValidator(dataProvider, spec, schemaCache));
+    }
```
ajsutton: I suspect we should always add the endpoint but then just return bad request if bellatrix isn't enabled. It's weird to get a 404 based on the network config.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -288,6 +290,20 @@ public ExecutionPayloadHeader deserializeJsonExecutionPayloadHeader(
         .jsonDeserialize(objectMapper.createParser(jsonFile));
   }
 
+  public SszList<SignedValidatorRegistration> deserializeSignedValidatorRegistrations(
```
ajsutton: The other deserialise methods here are detecting based on slot automatically so I'm not sure I'd add this here since it's not actually tied to a milestone - it's more tied to the API version.

I actually wonder if this is a case where the schema shouldn't be part of SchemaDefinitions. If we put it in SchemaDefinitions we can only change the API at milestone boundaries but I'm pretty sure that's not how the API is going to evolve. And if it wasn't in SchemaDefinitions we could make this API work regardless of whether Bellatrix is enabled (though it probably wouldn't have any actual effect until Bellatrix was active).

ajsutton (commented):
Looks good generally - I just think we're starting to create schema definitions that aren't actually tied to the spec version so probably need to break the rule of putting them into SchemaDefinitions. Previously all SSZ was used with the spec and so was inherently tied to the milestone but now we're just using it to define a REST API.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
tbenr: yep, that was my feeling. Going to move them in API itself

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostRegisterValidator.java
tbenr: I defined them `public` only, so I don't have to define a constructor and I don't have to use `@JsonProperty`

tbenr: seems to be the most compact version :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostRegisterValidator.java
```diff
@@ -0,0 +1,111 @@
...
+
+    try {
+      request.respondAsync(
+          validatorDataProvider
+              .registerValidators(request.getRequestBody())
+              .thenApply(AsyncApiResponse::respondOk));
+    } catch (final JsonProcessingException jsonProcessingException) {
+      request.respondError(SC_BAD_REQUEST, jsonProcessingException.getMessage());
+    }
+  }
```
tbenr: @rolfyone is it correct my bad request handling in case of wrong input?

StefanBratanov (approved):
LGTM!

PR #5547, opened 2022-05-20
Add comparison functions for two classes

jtraglia:
## PR Description

This simply adds comparison functions (`equals` & `hashCode`) for these two schema classes. The lack of these functions were causing some of my tests to fail.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5548, opened 2022-05-20
Deserialize BLSPubKey without validation

jtraglia:
## PR Description

Ran into this issue when deserializing public keys. I believe that the "empty public key" should be deserializable.

```
  com.fasterxml.jackson.databind.JsonMappingException:
    Public key 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 is invalid: Deserialization of public key bytes failed: 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 (through reference chain: tech.pegasys.teku.api.schema.altair.BeaconStateAltair["current_sync_committee"]->tech.pegasys.teku.api.schema.altair.SyncCommittee["pubkeys"]->java.util.ArrayList[51])
```

This is because `BLSPubKey.fromHexString` will automatically validate the key:

https://github.com/ConsenSys/teku/blob/335a1b7dafb0f65f07e234c94489089b5f36cf61/data/serializer/src/main/java/tech/pegasys/teku/api/schema/BLSPubKey.java#L66-L72

## Documentation

## Changelog


jtraglia:
Though maybe we should make an exception for the empty key instead of disabling it for all keys.

ajsutton (approved):
LGTM. Reviewing the code I don't think we actually use this decoder anywhere in production now anyway.  The validator ID from GetStateValidators is deliberately decoded without validation.  We do perform validation when reading the slashing protection records which is probably a good safety check to ensure the validator public key hasn't been mistyped or something like that.

PR #5556, opened 2022-05-23
Clean up duplicate query and path parameters

courtneyeh:


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
```diff
@@ -59,9 +67,15 @@ public class BeaconRestApiTypes {
...
 
-  public static final ParameterMetadata<UInt64> SLOT_PARAMETER =
+  public static final ParameterMetadata<UInt64> SLOT_QUERY_PARAMETER =
       new ParameterMetadata<>(SLOT, CoreTypes.UINT64_TYPE.withDescription(SLOT_QUERY_DESCRIPTION));
 
+  public static final ParameterMetadata<UInt64> SLOT_PATH_PARAMETER =
+      new ParameterMetadata<>(
+          SLOT,
+          CoreTypes.UINT64_TYPE.withDescription(
+              "The slot for which the block should be proposed."));
```
rolfyone: this seems like a very specific description... we likely want something more generic in this definition, and override it if it needs something somewhere else... for this comment currently, it'll only work in 1-2 places..

rolfyone: we could probably keep just SLOT_PARAMETER, and override description in the specific places the path_parameter is used

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
```diff
@@ -170,9 +155,9 @@ private static EndpointMetadata getEndpointMetaData(
                 + "will be added to as forks progress.\n\n"
                 + "Pre-Bellatrix, this endpoint will return a `BeaconBlock`.")
         .tags(TAG_VALIDATOR, TAG_VALIDATOR_REQUIRED, TAG_EXPERIMENTAL)
-        .pathParam(PARAM_SLOT)
-        .queryParamRequired(PARAM_RANDAO)
-        .queryParam(PARAM_GRAFFITI)
+        .pathParam(SLOT_PATH_PARAMETER)
```
rolfyone: so if you get rid of the path_param, this is just SLOT_PARAMETER.withDescription("...")

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
rolfyone: also at some point we seem to have flipped naming from PARAMETER_XXX to XXX_PARAMETER if we're cleaning things up... i prefer PARAMETER first so they're grouped together in hints, but at least it would be good for it to be consistent..

rolfyone (approved):
LGTM

PR #5558, opened 2022-05-23
Update 406 to 415 response.

rolfyone:
When an unsupported content type is discovered, return 415 rather than 400 or 406.

Implicitly adding 415 to possible responses for any request that adds a body.

## Documentation

## Changelog


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
```diff
@@ -55,14 +55,14 @@ public SendSignedBlockResult sendSignedBlock(SignedBeaconBlock signedBeaconBlock
...
-    if (!result.isPublished() && sszNotAcceptable) {
+    if (!result.isPublished() && unsupportedMediaType) {
       return sendSignedBlockAsJson(apiMethod, signedBeaconBlock);
     }
     return result;
   }
 
-  public boolean isSszNotAcceptable() {
-    return sszNotAcceptable;
+  public boolean inUnsupportedMediaType() {
```
ajsutton: is rather than in?

ajsutton (approved):
LGTM. Just one typo.

PR #5560, opened 2022-05-23
Add some JsonProvider property tests

jtraglia:
## PR Description

Currently, the roundTripBeaconState test fails. I see this error:

```
tech.pegasys.teku.infrastructure.ssz.sos.SszDeserializeException:
  Invalid Bitvector ssz: trailing bits are not 0
```

I think this is because it is deserializing the string bytes instead of the hex representation. In roundTripBitVector, I call Bytes.fromHexString on the string prior to deserialization. I am unsure how to do this for a serialized beacon state string.

Also, I needed to add syncAggregate & executionPayload fields to the beacon block generator. For altair, these are required. The other generator function had this already. I would see this error before:

```
java.lang.NullPointerException:
  syncAggregate must be specified
```

## Documentation

## Changelog


data/serializer/src/property-test/java/tech/pegasys/teku/provider/JsonProviderPropertyTest.java
```diff
@@ -0,0 +1,435 @@
...
+    final Class<? extends BeaconState> clazz = BEACON_STATE_CLASS_MAP.get(specMilestone);
+    final Constructor<?> constructor =
+        clazz.getConstructor(
+            tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState.class);
+    final Object original =
+        constructor.newInstance(
+            dataStructureUtil.randomBeaconState(validatorCount, numItemsInSSZLists));
+    assertInstanceOf(clazz, original);
+    final String serialized = jsonProvider.objectToJSON(original);
+    final Object deserialized = spec.deserializeBeaconState(Bytes.wrap(serialized.getBytes(UTF_8)));
```
jtraglia: Am I doing this right?

jtraglia, line 432: Am I doing this right?

jtraglia:
Maybe connected, when `childType` is `Bitvector[4]` the `childType.getSszFixedPartSize()` is 1. The deserializer only reads the first byte of the hex string, such as `"0x0d"` (`0b1101`), and fails to parse the byte`0x30` (`"0"`).

https://github.com/ConsenSys/teku/blob/d8e8f8cfc47d4d7e7f3bacb1b3bf693bf18b1bc1/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszContainerSchema.java#L288-L292

ajsutton:
Had to dig into this a bit so have pushed up some code changes that get the property tests passing.

The BeaconState is now parsed with the new type definitions since we can't parse it with the old JsonProvider since the SszBitvector parser has been removed.  Also fixed up the approach to parsing SszBitvector with type definitions. Then realised we should just use the bit vector type definition from the schema to do all the parsing so fixed that in the second commit.

This did show that the old json serialisation was handling uint256 incorrectly - it should be serialised as a decimal string (e.g. "1") in the beacon rest api but was being encoded as hex (e.g "0x01").  The JSON-RPC calls serialise this as a quantity which uses hex but the rest api uses decimal.  Would be good to get a sanity check from @tbenr that I haven't stuff something up that will cause issues with the execution engine API, but I think all those cases explicitly specify the "as hex" version of the serializer.

It's worth noting that we're currently working to get rid of `JsonProvider` entirely and will be using the JSON type definitions(`SerializableTypeDefinition` and `DeserializableTypeDefinition`) going forward. It may make more sense to focus property testing on those since it's what we're moving to.  Every SSZ object schema has a `getJsonTypeDefinition()` that provides the JSON definition for that type and there's a bunch of common types defined in `CoreTypes` and `EthereumTypes`.

jtraglia:
Thanks for fixing those things. I understand what I was doing wrong now. I will move onwards to the JSON type definitions. Just wanted to commit these tests since they seemed somewhat useful.

ajsutton (approved):
LGTM. Thanks again.

PR #5564, opened 2022-05-24
Send already seen attestation logging to debug level

### LINKED ISSUE #5534: Downgrade or suppress logging of io.libp2p.pubsub.MessageAlreadySeenException
## Background

`io.libp2p.pubsub.MessageAlreadySeenException` indicates that a node is trying to publish an attestation that it has previously seen via gossip. We currently log this at `ERROR` level.

For example,

```
2022-05-09 12:03:33.427 ERROR - Failed to publish attestation for slot 2965817
java.util.concurrent.CompletionException: io.libp2p.pubsub.MessageAlreadySeenException: Msg: PubsubMessage{data: "\367\001\034\344\000\000\0009A-\000\001\001\000\023\001\005\230\000\000\000\v8\244\240>-\270Qq\245Ej\250\327\213p\365\347\001\320#?E3%^\016\220\242\232O\254\bj\001\000\001(\210\355\354\354\233<\267\017:\262~\364\221\306\354\334A\2713\321\364`\252\023\2204y\232\a\t\301\037\022\tj\001\001\'\360\223\000\363\t\233\332%(-\3447\377v\024\367\313t\324\024\004.\236\300\226\226\275\230\344?\341\a\324\266}\204\227b\f\366 \a{\321s\030\265\346\231\001c\237\370%\033\203\355\261\315\212\213\273\000_\356z\206\323\370\356k\203`\353\022&z\273\2132\257!H\021\327\301\001.\327\n\017\203\306\374V\242\027\247M\300Ev\303\220\315\371\265\363\241\a\003N<qK3\027\207#\266\243\260m5\210\350\242K\341\f\337\000\000\000\000\000 \000\000\000\000\000\000\000\000\000\000\000\000 "
topicIDs: "/eth2/82f4a72b/beacon_attestation_19/ssz_snappy"
}
at java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:368) ~[?:?]
```

## The issue

In standard Teku configurations receiving a message before publishing it is definitely not expected and indicates something weird is going on. It doesn't seem actively harmful though, so maybe ought to be a `WARNING`, but that's not the main issue.

On non-standard deployments, however, it is expected that nodes might see attestations via gossip before they publish them. Examples are:

  - Where a validator is run in multiple locations, relying on Web3Signer for slashing protection. If one node is slower than another then it will throw this exception on trying to send an attestation although it is harmless. (Yes, this is not a recommended configuration, but people still deploy it.)
  - In distributed validator deployments multiple nodes can produce an attestation and they all publish it to the beacon chain to achieve redundancy. Again, this is harmless.

## Possible approaches

Options,
1. downgrade the severity of this to `DEBUG` so that it does not appear in normal logs, or
2. add a configuration option to suppress this message in non-standard configurations.

I think I favour the second, and maybe downgrading to `WARNING` in the normal case.

In all cases, since we've seen the message already via gossip, it likely still makes sense not to publish it again, so we should retain the current behaviour of not publishing it.

rolfyone:
Previously the already seen exception was logging to error, and it is more noisy than we'd like.

fixes #5534

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5569, opened 2022-05-24
Run fork choice prior to block proposals

### LINKED ISSUE #5465: Run fork choice prior to proposal
### Description
As per https://github.com/ethereum/consensus-specs/pull/2878

ajsutton:
## PR Description
Run fork choice prior to creating block proposals.

## Fixed Issue(s)
fixes #5465 

## Documentation

## Changelog


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/PreProposalForkChoiceTriggerTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static tech.pegasys.teku.infrastructure.async.SafeFutureAssert.assertThatSafeFuture;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+
+class PreProposalForkChoiceTriggerTest {
```
rolfyone, line 27: can we make class public so that the code scanner isn't unhappy?

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/TickProcessorTest.java
```diff
@@ -0,0 +1,151 @@
...
+import static tech.pegasys.teku.infrastructure.async.SafeFutureAssert.assertThatSafeFuture;
+
+import org.junit.jupiter.api.Test;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.storage.client.RecentChainData;
+import tech.pegasys.teku.storage.store.UpdatableStore.StoreTransaction;
+
+class TickProcessorTest {
```
rolfyone, line 32: same - maybe public class

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/validator/GetNewBlock.java
```diff
@@ -156,12 +156,10 @@ public void handleRequest(RestApiRequest request) throws JsonProcessingException
...
-            maybeBlock -> {
-              if (maybeBlock.isEmpty()) {
-                throw new ChainDataUnavailableException();
-              }
-              return AsyncApiResponse.respondOk(maybeBlock.get());
-            }));
+            maybeBlock ->
+                maybeBlock
+                    .map(AsyncApiResponse::respondOk)
+                    .orElseThrow(ChainDataUnavailableException::new)));
```
rolfyone, line 162: not the problem of this ticket specifically, but im not sure this is correct, looks like generic handler will turn this to a 'no content' (204) and most of the time ChainDataUnavailable i would have thought should be a 503
the obvious answer might be to make this throw `ServiceUnavailableException` in the throw., or potentially the ChainDataUnavailable generic handler needs adjusting

rolfyone (approved):
some minor notes / discussion points, but overall LGTM.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/PreProposalForkChoiceTriggerTest.java
ajsutton: I'd prefer not to, convention is that JUnit5 tests are package private. Might go see if there's a way to turn off the unused check on the GitHub scanner since it seems to be pretty broken.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/validator/GetNewBlock.java
ajsutton: hmm, 503 is wrong as that means the node is syncing but that's not the case here. This is the same exception we through when we're pre-genesis and so have nothing to return.  Given this goes to combinedChainDataClient to get the required state and it's done by slot, I think the only case where this can come back empty is if we're pre-genesis which would make it consistent with other APIs (not sure it's *right* but it's consistent).

PR #5572, opened 2022-05-24
Remove `.join()` calls in `BlockOperationSelectorFactory`

### LINKED ISSUE #5259: Remove `.join()` calls in `BlockOperationSelectorFactory` for execution payloads
We should remove `.join()` calls in `BlockOperationSelectorFactory` and allow returning a `Future`.
It will wind up by letting `BeaconBlockBodyBuilder::executionPayload\executionPayloadHeader` accept`Supplier<Future<ExecutionPayload\ExecutionPayloadHeader>>` and resolve the futures at build() time.

tbenr:
Get the rid of `join()` in `BlockOperationSelectorFactory`
So, block creation have to become async.

fixes #5259

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/schema/BeaconBlockBody.java
```diff
@@ -91,36 +91,38 @@ public BeaconBlockBody(
...
+                  .deposits(
+                      deposits.stream()
+                          .map(Deposit::asInternalDeposit)
+                          .collect(schema.getDepositsSchema().collector()))
+                  .voluntaryExits(
+                      voluntary_exits.stream()
+                          .map(SignedVoluntaryExit::asInternalSignedVoluntaryExit)
+                          .collect(schema.getVoluntaryExitsSchema().collector()));
+            })
+        .join();
```
tbenr, line 125: It seems to me that here it might be only place where makes sense to have the `join()` since we are providing an immediate completed future for header and payload. But I need to double-check

ajsutton: Given we want to get rid of these API classes I'd say this is probably a good point to use a join yeah. And yes we know it will already be completed anyway which makes me feel more ok with it too.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/BlockProposalUtil.java
```diff
@@ -71,40 +71,50 @@ public BeaconBlockAndState createNewUnsignedBlock(
...
+                      Optional.empty());
+
+              final Bytes32 stateRoot = newState.hashTreeRoot();
+              final BeaconBlock newCompleteBlock = block.withStateRoot(stateRoot);
+
+              return new BeaconBlockAndState(newCompleteBlock, newState);
+            })
+        .exceptionallyCompose(
+            error -> {
+              if (error instanceof BlockProcessingException) {
```
ajsutton: Probably need to check cause chain here as it tends to get wrapped in CompletionException.  `ExceptionUtil.getCause(err, BlockProcessingException.class).isPresent()` is probably the right way so that we pick it up even if there are further causes.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/BlockProposalUtil.java
```diff
@@ -71,40 +71,50 @@ public BeaconBlockAndState createNewUnsignedBlock(
...
+                return SafeFuture.failedFuture(new StateTransitionException((Exception) error));
```
ajsutton: Should just make `StateTransitionException` constructor take a `Throwable`.

ajsutton (commented):
Looks good I think. In playing around with it I've scattered a bunch of safeJoin through the test code to get it compiling so will push that up.  Feel free to just revert that commit if it gets in the way at all though.

tbenr:
I added a test to check that we are correctly intercepting `BlockProcessingException` and wrapping it with a `StateTransitionException`

ajsutton (approved):
LGTM.

PR #5573, opened 2022-05-24
[MEV Boost\Builder] Add validator remote call to register validators

StefanBratanov:
## PR Description

- Added `RegisterValidatorsRequest` following the `SendSignedBlockRequest` pattern
- Changed `OkHttpValidatorTypeDefClient` to instantiate the request objects when creating the object
- Changed up `ResponseHandler` to support 500 status code and reordered handlers by status code
- Fixed typo in `MetricRecordingValidatorApiChannel`
- Moved `OkHttpValidatorRestApiClientTest` to an integration test module

## Fixed Issue(s)
related to #5396 

## Documentation

## Changelog


tbenr (commented):
LGTM. I'd like to have final approval from @rolfyone since here we are instantiating request\responses once and now we require them to be stateless\thread-safe (which seems to me the case at the moment, but I'd like to hear from him if we want to give priority to safety in this case too)

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
```diff
@@ -47,7 +51,7 @@ public SendSignedBlockResult sendSignedBlock(SignedBeaconBlock signedBeaconBlock
             ? SEND_SIGNED_BLINDED_BLOCK
             : SEND_SIGNED_BLOCK;
 
-    return preferSszBlockEncoding
+    return preferSszBlockEncoding && !unsupportedMediaType
```
rolfyone: I think its probably simpler to handle one request. but if we do want to go this way...

preferSszEncoding was the callers way of defining that the cli is passed in. 

We don't need 2 flags, we just need to know if we're going to bother to try SSZ. so when unsupportedMediaType goes to true, we could instead set preferSszEncoding to false, and call the json endpoint.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
```diff
@@ -30,8 +30,12 @@
 
 public class SendSignedBlockRequest extends AbstractTypeDefRequest {
 
+  private final ResponseHandler<Object> responseHandler =
+      new ResponseHandler<>()
+          .withHandler(SC_UNSUPPORTED_MEDIA_TYPE, this::handleUnsupportedResponse);
+
   private final boolean preferSszBlockEncoding;
-  private boolean unsupportedMediaType = false;
+  private volatile boolean unsupportedMediaType = false;
```
rolfyone: I'm not sure how often this client is likely to be used, but potentially unsupportedMediaType disappears, and preferSszBlockEncoding becomes an atomic boolean

validator/remote/src/integration-test/java/tech/pegasys/teku/validator/remote/typedef/OkHttpValidatorTypeDefClientTest.java
```diff
@@ -0,0 +1,161 @@
...
+    RemoteServiceNotAvailableException serverException =
+        Assertions.assertThrows(
+            RemoteServiceNotAvailableException.class,
+            () -> okHttpValidatorTypeDefClient.registerValidators(validatorRegistrations));
+
+    assertThat(serverException.getMessage()).contains("Server error from Beacon Node API");
+  }
+
+  @TestTemplate
+  void registerValidators_makesSszRequestIfSszEncodingPreferred() throws InterruptedException {
```
rolfyone, line 131: probably could have a similar test that server returns a 415 and see that we end up with json returned...

so we send octet, server returns 415, we send json. (and we dont need to decode json again, just need to see the content-type is json and that the actual second call happens... then if we want to extend it a tiny bit, make a 'second' request, and that just goes straight to json...

rolfyone:
overall looks good, some minor tweaks suggested.
If we follow this pattern a bit there'll be value in reducing the code duplication but for now it'll be fine.

rolfyone (approved):
LGTM

PR #5580, opened 2022-05-25
update blocks by range to not skip blocks.

### LINKED ISSUE #5565: Deprecate BeaconBlocksByRange.step
### Description
Change our implementation of BeaconBlocksByRange to match the new deprecated implementation (only returns the first block).

Adjust any places we use a step > 1. I think this is just in CommonAncestorFinder.

See https://github.com/ethereum/consensus-specs/pull/2856

rolfyone:
In line with 1.20 spec, the 'step' of blocksByRange must now always be 1.

Need to evaluate that we can still accept a request with a step other than 1, and make it return the first block we find in that range

fixes #5565

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/CommonAncestor.java
```diff
@@ -28,9 +28,7 @@ public class CommonAncestor {
 
   static final UInt64 OPTIMISTIC_HISTORY_LENGTH = UInt64.valueOf(3000);
   // prysm allows a maximum range of 1000 blocks (endSlot - startSlot) due to database limitations
-  static final UInt64 MAX_BLOCK_RANGE = UInt64.valueOf(1000);
-  static final UInt64 SAMPLE_RATE = UInt64.valueOf(50);
-  static final UInt64 BLOCK_COUNT = MAX_BLOCK_RANGE.dividedBy(SAMPLE_RATE);
+  static final UInt64 BLOCK_COUNT = UInt64.valueOf(1000);
```
ajsutton: We probably don't want to actually request 1000 blocks. Previously we were requesting 20 blocks (just spaced out). I'd probably try something like 100 blocks (or normal batch size is 50 when syncing).  Long periods of non-finality will likely see us drop back to the finalised checkpoint quite a bit which is a shame but hard to avoid that now.  We could log a ticket to change this so that it sends a number of requests spaced out to search further back in the future.

ajsutton (approved):
LGTM. Just need to drop down the number of blocks we're requesting when finding the common ancestor.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/CommonAncestor.java
rolfyone: Ok, I'll set it to 100.

PR #5585, opened 2022-05-25
[Mev Boost\Builder] we should be using builder post-merge only

tbenr:
To reduce complexity during the transition, has been decided to start using builder infrastructure post merge only.

I'm introducing a `forceLocalFallback` in `builderGetHeader` to be used by `BlockOperationSelectorFactory` to control the fallback based on current transition state.

to summarize the flows:

blinded flow:
- pre bellatrix: `BlockFactory` transparently passthrough blocks
- post bellatrix:
  - pre transition: `BlockOperationSelectorFactory` will produce `executionHeaderOfDefaultPayload`. The flow completes in `createUnblinderSelector` which generates the corresponding `defaultPayload`
  - transition occurred, not finalized: `BlockOperationSelectorFactory` will force `forceLocalFallback` in `builderGetHeader`. The flow completes in `createUnblinderSelector` which calls `builderGetPayload` that rebuilds the payload from the fallback cache.
  - transition occurred and finalized: `BlockOperationSelectorFactory` will call `builderGetHeader` which eventually fallback. The flow completes in `createUnblinderSelector` which calls `builderGetPayload` that rebuilds the payload by calling  the builder or from the fallback cache.

unblinded flow (standard):
- pre bellatrix:  standard flow
- post bellatrix:
  - pre transition: `BlockOperationSelectorFactory` will produce `defaultPayload`
  - transition occurred, not finalized: `BlockOperationSelectorFactory` will call local `engineGetPayload`
  - transition occurred and finalized: same as above.

related to #5396 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
@@ -154,18 +154,17 @@ public class BlockOperationSelectorFactory {
                         .getHeaderOfDefaultPayload(),
                 (executionPayloadContext) -> {
                   final boolean forceLocalFallback =
-                          executionPayloadContext
-                                  .getForkChoiceState()
-                                  .getFinalizedExecutionBlockHash()
-                                  .isZero();
+                      executionPayloadContext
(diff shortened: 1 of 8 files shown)


ajsutton (approved):
LGTM.

Issue #5433, opened 2022-05-05
Add a debug endpoints to expose deposit information


ajsutton:
### Description
Add teku specific REST API endpoints that report:

 * Range of deposits held in memory (in `DepositProvider` that Teku could use to fill blocks)
 * What the Eth1 vote would be if we were to propose the next block

For a given state ID:
 * Eth1Data from that state
 * Break down of Eth1 votes in this voting period. ie for each unique Eth1Data value in the state.getEth1DataVotes(), report the number of times that Eth1Data has been voted for.
 * Total number of slots in the voting period
 * Total number of votes required to vote in the next Eth1Data (50% of total slots in period).

PR #5586, opened 2022-05-25
Debug endpoint with deposits not yet included on the BeaconChain

zilm13:
## PR Description
Starting #5433 
It's first endpoint for:
- Range of deposits held in memory (in DepositProvider that Teku could use to fill blocks)

I think it's definitely not `beacon` here in route "/eth/v1/debug/beacon/deposits", maybe "/eth/v1/debug/eth1/deposits"? We are going to have set of such endpoints, so we could add some special namespace for them.

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
```diff
@@ -0,0 +1,130 @@
...
+import tech.pegasys.teku.infrastructure.restapi.endpoints.RestApiRequest;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.datastructures.operations.Deposit;
+import tech.pegasys.teku.validator.coordinator.DepositProvider;
+
+@SuppressWarnings("unused")
+public class GetDeposits extends MigratingEndpointAdapter {
+
+  private static final Logger LOG = LogManager.getLogger();
+  public static final String ROUTE = "/eth/v1/debug/beacon/deposits";
```
rolfyone: Because this isn't a standard endpoint, we shouldn't have the route in that namespace.
It'll live in `/teku/v1` at the top, which equates to 'tekuv1' in the handlers area. can create a new debug there or put it under the beacon folder...
to me this looks like data from a state, so it should likely be 
`/teku/v1/beacon/state/{state_id}/deposits`

which means rather than hard coding 'head' we would have a path parameter that we use, and get the deposits from the resulting state...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
```diff
@@ -0,0 +1,130 @@
...
+import tech.pegasys.teku.beaconrestapi.MigratingEndpointAdapter;
+import tech.pegasys.teku.infrastructure.json.types.SerializableTypeDefinition;
+import tech.pegasys.teku.infrastructure.restapi.endpoints.AsyncApiResponse;
+import tech.pegasys.teku.infrastructure.restapi.endpoints.EndpointMetadata;
+import tech.pegasys.teku.infrastructure.restapi.endpoints.RestApiRequest;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.datastructures.operations.Deposit;
+import tech.pegasys.teku.validator.coordinator.DepositProvider;
+
+@SuppressWarnings("unused")
```
rolfyone: would prefer that we don't use suppression unless there's a need, and shouldn't be in a handler...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
```diff
@@ -0,0 +1,130 @@
...
+      final ChainDataProvider chainDataProvider,
+      final DepositProvider depositProvider,
+      final Spec spec) {
+    super(
+        EndpointMetadata.get(ROUTE)
+            .operationId("getDeposits")
+            .summary("Deposits not yet included on the chain")
+            .description(
+                "Range of deposits that could be used in new block if we are assigned to be a proposer.")
+            .tags(TAG_DEBUG)
```
rolfyone: need to tag this as teku (`TAG_TEKU`) rather than debug

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
```diff
@@ -0,0 +1,130 @@
...
+            .build());
+    this.chainDataProvider = chainDataProvider;
+    this.depositProvider = depositProvider;
+    this.spec = spec;
+  }
+
+  @OpenApi(
+      path = ROUTE,
+      method = HttpMethod.GET,
+      summary = "Deposits not yet included on the chain",
```
rolfyone: summary needs to be super short, it is used to display in a menu on our api docs here: https://consensys.github.io/teku
eg.  would be `/eth/v1/beacon/pool/attestations` which potentially is similar to this endpoint (in that it gets things from the attestations pool) - its description is 'Get Attestations', and a block selector one is 'Get Block attestations'.

Maybe 'Get deposits' or 'Get pending deposits'?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
```diff
@@ -0,0 +1,130 @@
...
+                    return AsyncApiResponse.respondOk(allDeposits);
+                  }
+                })
+            .exceptionally(
+                ex -> {
+                  LOG.error("Failed to retrieve deposits", ex);
+                  return AsyncApiResponse.respondWithError(
+                      SC_INTERNAL_SERVER_ERROR,
+                      "An internal error occurred, check the server logs for more details.");
+                }));
```
rolfyone: We should never be throwing a SC_INTERNAL_SERVER_ERROR in a handler, that literally is declaring we have a bug.
let the default handler deal with unexpected errors... is there a case this happens? because that would indicate we've got a problem...
also, can log.debug here if we really had to, but I would be inclined to simply delete this exceptionally. If we were to do something here we'd not be logging an error.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
rolfyone: with more reading maybe we want something more like
`/teku/v1/beacon/pool/deposits`

The deposits from a state does sound also like something the ticket is looking for, but it would be a different endpoint and pr...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
```diff
@@ -0,0 +1,130 @@
...
+        chainDataProvider
+            .getBeaconState(DEFAULT_STATE_ID)
+            .thenApplyChecked(
+                maybeState -> {
+                  if (maybeState.isEmpty()) {
+                    return AsyncApiResponse.respondNotFound();
+                  } else {
+                    List<Deposit> allDeposits =
+                        depositProvider.getAllPendingDeposits(maybeState.get().getData());
+                    return AsyncApiResponse.respondOk(allDeposits);
```
ajsutton: hmm, I don't think this is quite the Api we want.  There's three separate things this issue wants:

1. Report the range of deposits that are currently available. That wouldn't need a state input at all and should just return all the deposits we currently hold.
2. An API to return the Eth1Data that we'd include in a new block we propose. This would just always base it's decision on the current head state because our Eth1 block cache will have been pruned for earlier states and so won't give the same answer it would have at the time.
3. An API that takes a state ID and returns the Eth1Data from the state, plus the bits of calculated data from that state mentioned in the ticket (break down of votes, slots in the voting period etc).

We seem to have a bit of a mix of 1 and 3 here where we're getting a state and then returning a subset of deposits.

zilm13: @ajsutton I was trying to implement first one. Do you mean that it includes the whole deposits map? including deposits that are already on BeaconChain and those that are not yet?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
zilm13: @rolfyone I could catch it for example when deposits are not yet loaded, if sending query immediately after app started

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
zilm13: I like the one with a pool, it's a bit artificial but sounds in line with other pools

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
zilm13: Looking at @ajsutton comment, it should be "Get deposits"

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
ajsutton: Yes, the idea is to be able to see what we currently have in the deposit pool.

zilm13: Got it

zilm13:
Updated according to review

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/debug/GetDeposits.java
rolfyone: this will literally become an internal server error, an internal server error is what happens when an exception is uncaught, and should only ever be put in by the default exception handler.
what specific exception do we need to handlehere if deposits arent loaded?
`ChainDataUnavailableException`, `ServiceUnavailableException` and a number of others are handled generically, so I'm unclear as to why we'd be catching here unless we're looking for something, and if we're looking for something, the result is *never* going to be a sc_internal_server_error.

For specific errors, it may be worth putting a 503 (service unavailable) during startup scenario, but i don't want a generic exceptionally that catches all errors and throws it to internal error.

More likely would be that the function being called raises something appropriate - its a new function for our use. so if we find that the service isn't ready, throw ServiceUnavailableException in deposit provider and no specific handling would be required here in the handler (other than documenting a 503 may happen)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetDeposits.java
```diff
@@ -0,0 +1,87 @@
...
+          .build();
+
+  private final DepositProvider depositProvider;
+
+  public GetDeposits(final DepositProvider depositProvider) {
+    super(
+        EndpointMetadata.get(ROUTE)
+            .operationId("getDeposits")
+            .summary("Get deposits")
+            .description("Range of deposits that are currently available.")
```
ajsutton: nit: 
```suggestion
            .description("Get all deposits currently held for inclusion in future blocks.")
```

ajsutton (approved):
LGTM.

PR #5591, opened 2022-05-26
Return to using boxed integers in BLS batch verification

ajsutton:
## PR Description
Real world experience is showing that the IntStream optimisation is less efficient than this version using boxed integers so roll back to that.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5593, opened 2022-05-26
add optional query parameter to liveness

### LINKED ISSUE #5549: liveness endpoint should return non 200 if rejectedExecutions are detected
This would be a breaking change, potentially it'd need to have a query parameter to enable or similar...

Requested by infura so that they can cycle the machine if rejectedExecutionExceptions are detected..

rolfyone:
- If `failOnRejectedCount` set, and rejectedExecution count is greater than 0 in the monitoring window, liveness will return 503.
 - in the absence of the query parameter, liveness behaviour is unaltered, return 200 if it can respond.
 - Migrated liveness endpoint to the new structure and updated tests.

 Generally a teku instance should recover after rejectedExecutionExceptions, but if desired, this query parameter can be set to alter the liveness result, and potentially take corrective action in your infrastructure if required.

 fixes #5549

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5594, opened 2022-05-26
Implement keymanager-api deleteFeeRecipient

### LINKED ISSUE #5587: keymanager - delete fee recipient
keymanager - delete fee recipient

https://github.com/ethereum/keymanager-APIs/pull/30

The concept is basically that we can add a custom fee recipient to use via the api, now we need to be able to remove that mapping so that we can fall back to the default.

This api will not be able to delete from the users configuration file, that would be an error if they attempt to do that.

rolfyone:
- fee recipient configured via the config file loaded at startup are not candidates to be removed, and will throw a 403.

 - the intent is that a fee_recipient added via setFeeRecipient will be able to be deleted so that the nominated pubkey can again use the default recipient.

 - if a pubkey that we own is specified, it is not configured via config file, and was using default fee recipient, a delete will 'succeed', it'll continue to use the default fee recipient.

 - delete takes no body, and returns no body (204 response) on success.

 fixes #5587

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #5599, opened 2022-05-26
RestAPI endpoint with Eth1Data we'd put in proposed block

zilm13:
## PR Description
Continue #5433 
This is second endpoint which would return this:
- What the Eth1 vote would be if we were to propose the next block

I'm not sure whether `Eth1Vote` could be a better name for everything around this, kept `Eth1Data`. I've touched `Eth1DataCache` in understanding our logic there, could revert minor change if it is not welcome
## Documentation

## Changelog


zilm13:
forked from #5586 
will mark as ready after #5586 merged

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetEth1Data.java
```diff
@@ -0,0 +1,102 @@
...
+    request.respondAsync(
+        chainDataProvider
+            .getBeaconStateAtHead()
+            .thenApply(
+                mayBeState -> {
+                  if (mayBeState.isEmpty()) {
+                    return AsyncApiResponse.respondNotFound();
+                  } else {
+                    StateAndMetaData stateAndMetaData = mayBeState.get();
+                    Eth1Data eth1Vote = eth1DataCache.getEth1Vote(stateAndMetaData.getData());
```
rolfyone: these are nit-picking,
 - variables should be final...
 - we're getting StateAndMetadata but looks like we want BeaconState, so make the variable BeaconState
 - if you do that, then it probably becomes simple enough that you could:
 ```
final BeaconState state = maybeState.get().getData();
return AsyncApiResponse.respondOk(eth1DataCache.getEth1Votes(state));
```
- i guess from that, 'maybe' is all lowercase, not capital b B
potentially make this even simpler, if by design Eth1DataCache can have `getEth1Votes(maybeStateAndMetadata)`
and then the logic is out of the handler...
so then you would just be doing
`thenApply(eth1DataCache::getEth1Vote);`

Comment (zilm13) [reply data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetEth1Data.java]: It will be definitely easier to read with these changes, I'll apply it.

> if by design Eth1DataCache can have getEth1Votes(maybeStateAndMetadata)

I think it's a bad idea in current case as `Eth1DataCache` is not designed for that, and this method is already used by others, but for the 3rd task endpoint I'm thinking about dedicated provider with all logic there, and I will slightly refactor these 2 endpoints there to pass everything through it, and it could hide logic like this. Not final decision, but there are more logic in 3rd endpoint and handler is not a best part for it.

File: data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/paths/_teku_v1_beacon_state_eth1data.json [hunk 2]
```diff
@@ -0,0 +1,40 @@
+{
+  "get" : {
+    "tags" : [ "Teku" ],
+    "operationId" : "getEth1Data",
+    "summary" : "Get new Eth1Data",
+    "description" : "Eth1Data we'd include in a new block if we are a proposer.",
```
ajsutton:
```suggestion
    "description" : "Eth1Data that would be used in a new block created based on the current head",
```

ajsutton (approved):
LGTM. Would just tweak the API description.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetEth1Data.java
```diff
@@ -0,0 +1,101 @@
...
+import tech.pegasys.teku.infrastructure.restapi.endpoints.AsyncApiResponse;
+import tech.pegasys.teku.infrastructure.restapi.endpoints.EndpointMetadata;
+import tech.pegasys.teku.infrastructure.restapi.endpoints.RestApiRequest;
+import tech.pegasys.teku.spec.datastructures.blocks.Eth1Data;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.validator.coordinator.Eth1DataCache;
+
+public class GetEth1Data extends MigratingEndpointAdapter {
+
+  public static final String ROUTE = "/teku/v1/beacon/state/eth1data";
```
ajsutton: This probably isn't the right URL for this as `.../beacon/state` is typically followed by a state ID and `eth1data` potentially gets very confusing in that context. Also the data isn't really coming from the state but more from the pool of Eth1 data.

I'd probably make it:
```suggestion
  public static final String ROUTE = "/teku/v1/beacon/pool/eth1data";
```

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetEth1Data.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public void handleRequest(final RestApiRequest request) throws JsonProcessingException {
+    request.respondAsync(
+        chainDataProvider
+            .getBeaconStateAtHead()
+            .thenApply(
+                maybeStateAndMetadata -> {
+                  if (maybeStateAndMetadata.isEmpty()) {
+                    return AsyncApiResponse.respondNotFound();
```
ajsutton: The `EndpointMetadata` doesn't declare not found as a valid response type - need to add it in and should add a test to confirm the 404 response is sent correctly.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetEth1Data.java
zilm13: Right, it's not from the state

PR #5602, opened 2022-05-26
Removed deprecated `/teku/v1/beacon/states/{state_id}` endpoint

rolfyone:
Now that we're getting close to migrating the teku endpoints, we should probably just remove this one, as it's superceded by the standard state apis.

It's been deprecated for a long time, but still technically breaking, so added to the breaking changes list.

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ For information on changes in released versions of Teku, see the [releases page]
 ## Unreleased Changes
 
 ### Breaking Changes
+- removed `/teku/v1/beacon/states/{state_id}`, as the functionality is in the standard state apis.
```
ajsutton, line 17: Should be able to remove it from the upcoming breaking changes now (finally).

ajsutton (approved):
LGTM.

CHANGELOG.md
rolfyone: i looked at that and missed it! fair call

PR #5603, opened 2022-05-27
Revert "Implement more memory efficient deposit merkle tree (#5501)"

ajsutton:
## PR Description
While the new tree is more memory efficient it is proving to be significantly less CPU efficient while replaying deposits.  Reverting until we can investigate and either fix that or move to actually using snapshots.

This reverts commit 784799f07c9df6676dc32d7cbd198d97e19489e6.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5611, opened 2022-05-27
Optimize primitive SSZ collection iteration

Nashatyrev:
## PR Description

Essentially optimize `ValidatorStatsAltair.getValidatorStats()` method which currently takes around 250ms which means about 0.5 sec of CPU per slot which looks a bit expensive for metrics gathering. 

The major issue with byte list is that we are traversing the backing tree from the very root node every time a single byte is accessed. 

The main idea to optimize it is to introduce `CachingTreeAccessor` which is capable of caching some internal results of SszList element query to optimize the next list query. 

This PR:
- Adds `ByteListBenchmark`
- Adds basic `CachingTreeAccessor` implementation which just caches the latest accessed tree leaf so any following queries to elements packed in the same leaf would avoid tree search. Utilizes this class in `SszPrimitiveListImpl`
- Precompute `treeDepth` in `AbstractSszCollectionSchema`
- Precompute a couple of frequently used fields in `SszPrimitiveListImpl`
- Adds `SszPrimitiveSchema.createFromPackedNodeUnboxed()` to potentially avoid ssz boxing/unboxing when iterating primitive collections
- Change `SszPrimitiveListSchema` factory method to create specific `SszByteListSchema` class when the list element type is `Byte`

### Before changes:
```
Benchmark                                                Mode  Cnt  Score   Error  Units
ByteListBenchmark.iterateBoxed                          thrpt   50  5,541 ± 0,079  ops/s
ByteListBenchmark.iterateUnboxed                        thrpt   50  5,229 ± 0,102  ops/s
ByteListBenchmark.randomIndexedAccess                   thrpt   50  5,316 ± 0,073  ops/s
ByteListBenchmark.serialIndexedAccess                   thrpt   50  5,652 ± 0,156  ops/s
ByteListBenchmark.iterateBeaconStateParticipationFlags  thrpt   50  4,799 ± 0,258  ops/s
```

### After changes:
```
ByteListBenchmark.iterateBoxed                          thrpt   50  41,112 ± 0,917  ops/s
ByteListBenchmark.iterateUnboxed                        thrpt   50  75,950 ± 0,746  ops/s
ByteListBenchmark.randomIndexedAccess                   thrpt   50   6,698 ± 0,072  ops/s
ByteListBenchmark.serialIndexedAccess                   thrpt   50  84,515 ± 1,041  ops/s
ByteListBenchmark.iterateBeaconStateParticipationFlags  thrpt   50  21,813 ± 1,734  ops/s
```

### After changes with addressed #5623
```
ByteListBenchmark.iterateBeaconStateParticipationFlags  thrpt   50  36,538 ± 2,881  ops/s
```

## Further steps 
- Address the issue #5623
- Implement more advanced `CachingTreeAccessor` strategy: cache not only the leaf node but also internal branch nodes to start traversing the tree from a lower node when another leaf needs to be retrieved
- Declare `BeaconStateAltair.*_participation` fields as specific `SszByteList` instead generic `SszList<SszByte>` for utilizing faster unboxed values access methods 

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5616, opened 2022-05-29
Provide a version of LimitedMap and LimitedSet that are safe for iteration

### LINKED ISSUE #5452: LimitedMap and LimitedSet are no longer thread safe for iteration

ajsutton:
## PR Description
Provide a version of LimitedMap and LimitedSet that are safe for iteration - use it in `OperationPool` which needs to iterate.
Rename methods to make it clearer which thread safety approach is used.

## Fixed Issue(s)
fixes #5452 

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #5312, opened 2022-04-08
[Merge] Store blocks and execution payloads separately


ajsutton:
### Description
Post-merge the `ExecutionPayload` inside a block can be quite large and isn't needed for a lot of operations (only the payload header). In particular it's not required when replaying blocks to generate state.

Currently in the in-memory and on-disk stores, teku is storing the whole block with execution payload but we could likely reduce memory requirements without loss of performance by splitting the two so that normally we work with blinded blocks (containing only the payload header) and then load the execution payload separately when it's required.

ajsutton:
This would also potentially set things up well to not store the execution payload in Teku's database but retrieve it from the EL when required. Probably not the initial goal but should be considered when adjusting the APIs involved with retrieving blocks.

rolfyone:
3 benchmark tests:
 - storing hotblocks vs. storing blinded hotblocks and payloads saw a 12% slowdown
 - storing hotblocks vs. blinded blocks and no payload was faster but not significantly
- epoch transition with blinded blocks vs. unblinded blocks - blinded blocks see approx a 5% improvement in times.

At this stage it's a bit of a mixed bag, and we're not able to easily fetch payload from EE currently. We might park this change for now and re-visit once async transactions are complete, and we can fetch payloads from EE...

PR #5618, opened 2022-05-30
[Spike] Store execution payload separately

ajsutton:
## PR Description
Spike into how to store execution payload separately to the block.  Current code is focussed on automatically creating the required SSZ schemas instead of having to create yet another version for blocks and also avoid having to copy over all the data.

## Fixed Issue(s)
#5312 

## Documentation

## Changelog


ajsutton:
Went with an alternate approach in #5677.

PR #5627, opened 2022-05-30
Return JSON content when Accept: */* is specified

ajsutton:
## PR Description
Fix the beacon REST API so that it consistently returns JSON when `Accept: */*` header is specified. Supported content types are now specified as a list in priority order rather than a Map to ensure that we get a consistent ordering when multiple supported options match the requested type.

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #5369, opened 2022-04-22
[Merge] Report on quality of execution payloads produced


ajsutton:
### Description
Give an indication of how good execution payloads are in blocks we produce:

 * Time between making the preparePayload call and the getPayload call (how long did we give the EL to produce the block?)
 * Gas used vs gas limit - how "full" was the produced block
 * For MEV may be able to report the rewards we were paid

Not sure if this should be a log message or metrics. May need to experiment to see what's more useful.

rolfyone:
closing this ticket, the rest is covered in #5654

PR #5635, opened 2022-05-31
Report EL block information when creating blocks

rolfyone:
At the end of block production, now include the gas used and %, EL block number and EL block hash.

example:
```
*** Published block              Count: 1, Slot: 1754, Root: d5a288285e922d77bbcaa856d14a087bf5211aecc744f8c2e07511cb25bde587, 0 (0%) gas, EL block:  50a7cc34b5e81a6de7b9dcc142b439cd724f2c1e902d96ed4e006fd74ffd6241 (4)
```

partially addresses #5369

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
@@ -68,7 +68,7 @@ public interface BeaconBlockBody extends SszContainer {
   }
 
   default Optional<ExecutionPayloadSummary> getOptionalExecutionPayloadSummary() {
-    return getOptionalExecutionPayload().or(getOptionalExecutionPayloadHeader());
+    return Optional.empty();
   }
 
   default boolean isBlinded() {
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
@@ -68,7 +68,7 @@ public interface BeaconBlockBody extends SszContainer {
   }
 
   default Optional<ExecutionPayloadSummary> getOptionalExecutionPayloadSummary() {
-    return Optional.empty();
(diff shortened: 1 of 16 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
```diff
@@ -66,6 +67,10 @@ default Optional<ExecutionPayloadHeader> getOptionalExecutionPayloadHeader() {
     return Optional.empty();
   }
 
+  default Optional<ExecutionPayloadSummary> getOptionalExecutionPayloadSummary() {
+    return Optional.empty();
```
ajsutton, line 71: I'd suggest:
```suggestion
    return getOptionalExecutionPayload().or(getOptionalExecutionPayloadHeader());
```

That way if either of those are overridden this method just works (but still makes sense to explicitly override it as you've done as well to save checking both options).

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadSummary.java
```diff
@@ -0,0 +1,28 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.spec.datastructures.execution;
+
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+
+public interface ExecutionPayloadSummary {
```
ajsutton, line 22: While they won't be used, I'd suggest adding all the fields from payload header here since they're available from both header and full payload.

ajsutton (approved):
LGTM.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadSummary.java
rolfyone: ok cool, I can do that I just figured I'd declare the minimum I needed

PR #5645, opened 2022-05-31
Reject negative values in byte deserializer

jtraglia:
## PR Description

Due to recent changes to the byte type definition, negative integer values (`"0x80000000"` - `"0xffffffff"`) are allowed.

I do not believe this should be allowed.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/ByteTypeDefinition.java
+++ b/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/ByteTypeDefinition.java
@@ -50,7 +50,7 @@ class ByteTypeDefinition extends PrimitiveTypeDefinition<Byte> {
     final String valueToParse =
         valueAsString.startsWith("0x") ? valueAsString.substring("0x".length()) : valueAsString;
     final int value = Integer.parseUnsignedInt(valueToParse, RADIX);
-    checkArgument(0 <= value && value <= 255, "byte value %s should be between 0 and 255", value);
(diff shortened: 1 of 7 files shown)


jtraglia:
Just to mention, prior to this PR the negative value would be cast to a byte and lose to upper bits.

For example, `0x89312050` would result in a value of `0x50`.

jtraglia:
Just to mention, prior to this PR the negative value would be cast to a byte and lose the upper bits.

For example, `0x89312050` would result in a value of `0x50`.

jtraglia:
Also worth mentioning, I don't think that `Integer.parseUnsignedInt` behaves the way the author of this intended. `Integer.parseUnsignedInt("ffffffff", 16)` and `Integer.parseUnsignedInt("4294967295", 10)` will result in `-1`. I will update `UInt8TypeDefinition` to fix the same problem.

ajsutton:
Thanks. Changes look good - I've pushed up a commit that changes the tests a bit. The property tests would pass even if a valid value failed to parse which was happening for a lot of hex encoded values because they weren't string quoted. Also changed the `RestApiRequestTest` to hard code the serialised forms instead of generating them at @rolfyone 's suggestion so we don't have the problem of it encoding them incorrectly again in the future.

Can you double check my changes and if you're happy with them I'll merge please? Given I messed this up just yesterday a second pair of eyes would be good. :)

jtraglia:
Yes, your changes look good to me. Thanks for revising those tests. I also fixed a small nit in something else.

jtraglia:
Actually one second. Just noticed something.

jtraglia:
Okay I'm done now. Ready to merge.

ajsutton (approved):

PR #5653, opened 2022-06-01
Enforce responses match metadata in StubRestApiRequest

ajsutton:
## PR Description
Update `StubRestApiRequest` to require the endpoint metadata and verify that the metadata matches the actual response provided (at least that the code is declared and whether or not content is provided currently).

Fix cases where response codes were incorrectly declared.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5656, opened 2022-06-01
Remove BlockBasedEth1HeadTracker

ajsutton:
## PR Description
Block based follow distance was an older, incorrect approach. We've feature toggled over to the correct time based following a long time ago and no longer need the old implementation.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5660, opened 2022-06-01
Fix get deposits endpoint

ajsutton:
## PR Description
Include the deposit index in the response which is the key identity for the deposit
Don't include the proof to avoid failing if the proof can't be generated because the deposit is considered finalized.

On our ropsten nodes it was always failing with:
```
{
  "code": 400,
  "message": "Cannot get proof for finalized deposits"
}
```
which makes it useless for debugging issues (including the question of why it's holding a deposit it shouldn't need).

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5662, opened 2022-06-01
Skip unknown object fields correctly

ajsutton:
## PR Description
When an unknown field is reached when parsing an object, ensure the value is skipped even if it's not a simple value (e.g an object or array).

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5663, opened 2022-06-01
set objectMapper to ignore unknown fields by default

rolfyone:
JsonProvider broke on a previous release because of an extra field, which shouldnt happen.

Disabled the exception thrown when an unknown object attribute occurs via JsonProvider.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5668, opened 2022-06-01
Add roundTrip property tests for execution structs

jtraglia:
## PR Description

Adding some PBT tests that I made for execution data structures. I didn't find any problems.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/execution/SignedBuilderBidPropertyTest.java
+++ b/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/execution/SignedBuilderBidPropertyTest.java
@@ -41,10 +41,10 @@ public class SignedBuilderBidPropertyTest {
     final SignedBuilderBid bid = dataStructureUtil.randomSignedBuilderBid();
 
     final ExecutionPayloadHeader header = dataStructureUtil.randomExecutionPayloadHeader();
-    final BuilderBidSchema unsignedSchema = new BuilderBidSchema(header.getSchema());
-    final SignedBuilderBidSchema schema = new SignedBuilderBidSchema(unsignedSchema);
+    final BuilderBidSchema schema = new BuilderBidSchema(header.getSchema());
+    final SignedBuilderBidSchema signedSchema = new SignedBuilderBidSchema(schema);
     final DeserializableTypeDefinition<SignedBuilderBid> typeDefinition =
-        schema.getJsonTypeDefinition();
+        signedSchema.getJsonTypeDefinition();
     final String json = JsonUtil.serialize(bid, typeDefinition);
     final SignedBuilderBid result = JsonUtil.parse(json, typeDefinition);
     assertThat(result).isEqualTo(bid);
--- a/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/execution/SignedBuilderBidPropertyTest.java
+++ b/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/execution/SignedBuilderBidPropertyTest.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2022 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.spec.datastructures.execution;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import net.jqwik.api.Arbitraries;
+import net.jqwik.api.Arbitrary;
+import net.jqwik.api.ForAll;
+import net.jqwik.api.Property;
+import net.jqwik.api.Provide;
+import tech.pegasys.teku.infrastructure.json.JsonUtil;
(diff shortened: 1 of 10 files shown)


jtraglia:
Dumb question, why do I sometimes need to create a schema object to get an appropriate JSON type definition?

For example, with the `Transaction` class:

Screen Shot 2022-06-01 at 1 26 04 PM

If it change it to `transaction.getSchema()` it gives this error:

Screen Shot 2022-06-01 at 1 27 13 PM

What is `capture of ?` here? Is this due to this being a work in progress?

rolfyone:
That's kind of where we're up to, at the moment if we do hit this we tend to cast to the correct deserializable, so more like your second example, and use casting... not sure if @ajsutton was considering fixing this one or not...

ajsutton:
There's a limit to what we can do here. The issue is that our SSZ types have fairly complex structures (for various reasons) so the `getSchema()` method often just promises to return some subclass of `SszSchema` (most generically `SszSchema<? extends SszData>`). That means we don't know which specific type the data is though - fine when you're using it as a return type because the common `SszData` can be used but impossible to match when you need to pass a parameter of that type because the compiler doesn't know what type is required.

In many specific schemas we override this to have a more specific return type which sometimes solves the problem. For `Transaction` we should just add:
```
@Override
  public TransactionSchema getSchema() {
    return (TransactionSchema) super.getSchema();
  }
```
because the constructor forces the schema to be a `TransactionSchema` and it would all work.  For some types - particularly states - we can't be that specific because we have different implementations for different forks.

But usually it works best to not try and use `getSchema()` and just get the schema you know you need from `SchemaDefinitions` (from the spec). That usually gives you a much more concrete type and things work out better. I think there are still a couple of particularly complex cases like state where that doesn't work. You can sometimes deploy generics creatively but particularly for tests just casting is going to be a lot easier.

ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/execution/SignedBuilderBidPropertyTest.java
```diff
@@ -0,0 +1,58 @@
...
+      @ForAll("milestone") final SpecMilestone specMilestone,
+      @ForAll final Eth2Network network)
+      throws JsonProcessingException {
+    final Spec spec = TestSpecFactory.create(specMilestone, network);
+    final DataStructureUtil dataStructureUtil = new DataStructureUtil(seed, spec);
+    final SignedBuilderBid bid = dataStructureUtil.randomSignedBuilderBid();
+
+    final ExecutionPayloadHeader header = dataStructureUtil.randomExecutionPayloadHeader();
+    final BuilderBidSchema schema = new BuilderBidSchema(header.getSchema());
+    final SignedBuilderBidSchema signedSchema = new SignedBuilderBidSchema(schema);
```
ajsutton, line 45: For the record, these schemas from the builder API are a bit unique - they are the first ones we have that aren't from the beacon chain spec and so versioning won't be tied to hard forks. We don't have a central repository for them currently, so we wind up creating new instances like this.  Might be something we want to revisit though as creating SSZ schemas can be expensive so we probably want to make sure we're reusing them in production.

ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/execution/TransactionPropertyTest.java
```diff
@@ -0,0 +1,57 @@
...
+      @ForAll("milestone") final SpecMilestone specMilestone,
+      @ForAll final Eth2Network network)
+      throws JsonProcessingException {
+    final Spec spec = TestSpecFactory.create(specMilestone, network);
+    final DataStructureUtil dataStructureUtil = new DataStructureUtil(seed, spec);
+    final Transaction transaction = dataStructureUtil.randomExecutionPayloadTransaction();
+
+    final TransactionSchema schema =
+        new TransactionSchema(
+            spec.forMilestone(specMilestone).getConfig().toVersionBellatrix().orElseThrow());
```
ajsutton: In this case overriding `getSchema()` in `Transaction` would be very useful and save having to create the schema again.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -511,6 +513,16 @@ public ExecutionPayload emptyExecutionPayload() {
             List.of());
   }
 
+  public Transaction randomExecutionPayloadTransaction() {
+    final TransactionSchema schema =
+        new TransactionSchema(
+            spec.forMilestone(SpecMilestone.BELLATRIX)
+                .getConfig()
+                .toVersionBellatrix()
+                .orElseThrow());
```
ajsutton: Probably better to get this from the spec schema definitions:
```suggestion
    final TransactionSchema schema =
        SchemaDefinitionsBellatrix.required(spec.getGenesisSchemaDefinitions())
            .getExecutionPayloadSchema()
            .getTransactionSchema();
```

You'll need to make that `getTransactionSchema()` method public as well but I think that's a good change since the `ExecutionPayloadSchema` is able to guarantee the schema type for transactions and we lose some detail in the generics when getting the element schema from the transactions list.

We don't tend to use the transaction schema directly in production, hence it gets a bit buried here.

ajsutton (commented):
Looks good. Just a couple of suggested changes to how we get the `TransactionSchema` to simplify things. The others I think are all builder API related so pretty sure they need to be directly created at this stage.

jtraglia:
Thank you for the review and for explaining. I like your suggestions. I will make those changes soon.

jtraglia:
> The others I think are all builder API related so pretty sure they need to be directly created at this stage.

Would you mind clarifying regarding this? Should I change something?

ajsutton:
> > The others I think are all builder API related so pretty sure they need to be directly created at this stage.
> 
> Would you mind clarifying regarding this? Should I change something?

I meant what you have is right for them. :)

jtraglia:
Okay cool 🙂 thanks. I'm done then.

ajsutton (approved):
LGTM. Thanks.

PR #5670, opened 2022-06-01
ignore stream EOF in beacon rest api

rolfyone:
fixes #5569

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
@@ -31,6 +31,7 @@ import io.javalin.plugin.openapi.jackson.JacksonModelConverterFactory;
 import io.javalin.plugin.openapi.ui.SwaggerOptions;
 import io.swagger.v3.oas.models.info.Info;
 import io.swagger.v3.oas.models.info.License;
+import java.io.EOFException;
 import java.net.BindException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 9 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
```diff
@@ -234,15 +235,18 @@ private void addExceptionHandlers() {
...
-        (e, ctx) -> {
-          LOG.error("Failed to process request to URL {}", ctx.url(), e);
-          ctx.status(SC_INTERNAL_SERVER_ERROR);
-          setErrorBody(
-              ctx, () -> BadRequest.internalError(jsonProvider, "An unexpected error occurred"));
-        });
+    app.exception(Exception.class, this::catchAllExceptionHandler);
+  }
+
+  private void catchAllExceptionHandler(final Throwable throwable, final Context context) {
```
ajsutton: Probably need to apply the same fix to the validator API too.  Splitting this out to a separate class probably makes sense as it should be the same for both I think.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
```diff
@@ -234,15 +235,18 @@ private void addExceptionHandlers() {
...
+    if (throwable.getCause() instanceof EOFException) {
```
ajsutton: I think we need to check if anything in the cause chain is `EOFException` as it may wind up being wrapped multiple times.  So use `ThrowableUtil.getCause(throwable, EOFException.class).isPresent()` (or maybe add a `ThrowableUtil.hasCause(throwable, EOFException.class)` that just returns boolean for simplicity.

ajsutton (approved):
LGTM. We could now remove the `url` parameter from `RestApiExceptionHandler.handleException` which I think makes sense as it's quite unlikely to be used (and it will make codeql happy).

PR #5677, opened 2022-06-02
Reuse backing tree when converting between blinded and unblinded blocks

ajsutton:
## PR Description
Provide simple `blind` and `unblind` methods on `SignedBeaconBlock` that share the unchanged parts of the backing tree instead of creating an entirely new copy.

Currently this is only used in the unblinder, but prepares for storing the transactions separately.

#5312 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/SignedBeaconBlock.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/SignedBeaconBlock.java
@@ -13,10 +13,13 @@
 
 package tech.pegasys.teku.spec.datastructures.blocks;
 
+import static com.google.common.base.Preconditions.checkState;
+
 import java.util.Optional;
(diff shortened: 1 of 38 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/SignedBeaconBlock.java
```diff
@@ -45,6 +50,56 @@ public static SignedBeaconBlock create(
     return new SignedBeaconBlock(signedBeaconBlockSchema, message, signature);
   }
 
+  public SignedBeaconBlock blind(final SchemaDefinitions schemaDefinitions) {
+    if (getMessage().getBody().isBlinded()) {
+      return this;
+    }
+
+    final TreeNode unblindedTree = getBlindedTree();
```
rolfyone: nit: looks like variable should be `blindedTree` ?

rolfyone (approved):
LGTM, one potential nit.

Issue #5666, opened 2022-06-01
Sanity check on "Detected peer with inconsistent finalized block" warning


benjaminion:
Haven't noticed these before. This is on mainnet with Teku 22.5.1.

```
May 31 21:34:05 metalalbert teku[2479]: 2022-05-31 21:34:05.613 WARN  - Detected peer with inconsistent finalized block at slot 31 for peer 16Uiu2HAkzbihk8sADuWKm3tTRx6CkUhXEFVotiSCraKsEhzVdmig.  Block roots 0x03dfea04ec332d814a75c768d53196ad3f2b938a6b3927062b35c7a15810c1a1 and 0xbacd20f09da907734434f052bd4c9503aa16bab1960e89ea20610d08d064481c do not match
May 31 21:36:35 metalalbert teku[2479]: 2022-05-31 21:36:35.994 WARN  - Detected peer with inconsistent finalized block at slot 31 for peer 16Uiu2HAkzbihk8sADuWKm3tTRx6CkUhXEFVotiSCraKsEhzVdmig.  Block roots 0x03dfea04ec332d814a75c768d53196ad3f2b938a6b3927062b35c7a15810c1a1 and 0xbacd20f09da907734434f052bd4c9503aa16bab1960e89ea20610d08d064481c do not match
May 31 22:40:05 metalalbert teku[2479]: 2022-05-31 22:40:05.786 WARN  - Detected peer with inconsistent finalized block at slot 31 for peer 16Uiu2HAkzbihk8sADuWKm3tTRx6CkUhXEFVotiSCraKsEhzVdmig.  Block roots 0x03dfea04ec332d814a75c768d53196ad3f2b938a6b3927062b35c7a15810c1a1 and 0xbacd20f09da907734434f052bd4c9503aa16bab1960e89ea20610d08d064481c do not match
May 31 22:42:26 metalalbert teku[2479]: 2022-05-31 22:42:26.278 WARN  - Detected peer with inconsistent finalized block at slot 31 for peer 16Uiu2HAkzbihk8sADuWKm3tTRx6CkUhXEFVotiSCraKsEhzVdmig.  Block roots 0x03dfea04ec332d814a75c768d53196ad3f2b938a6b3927062b35c7a15810c1a1 and 0xbacd20f09da907734434f052bd4c9503aa16bab1960e89ea20610d08d064481c do not match
[...]

PR #5678, opened 2022-06-02
Clarify log message

rolfyone:
This will give us visibility of the epoch being requested, although it's possible still that the epoch may be local or remote depending on the direction that things are coming.

Previously the block hash order would vary depending on how we arrived at the log message.

partially addresses https://github.com/ConsenSys/teku/issues/5666

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5686, opened 2022-06-03
Enable fork choice before proposal on testnets

ajsutton:
## PR Description
Enable fork choice before proposals on testnet configs. While we often override the default just on our own nodes this will give us much better feedback by enabling it for all testnet nodes (but it can still be overridden to disable again).

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5687, opened 2022-06-03
Publish metrics for block duties being completed

ajsutton:
## PR Description
Add duty metrics for block timings when validators publish blocks through the beacon node.
Switch duty metrics to use counters of number of duties completed in specific time ranges rather than a histogram. Resulting data is similar but handles better granularity and is less affected by large outliers which tend to happen at startup.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5690, opened 2022-06-03
Migrate PostContributionAndProofs to MigratingEndpointAdapter

### LINKED ISSUE #5676: Migrate PostContributionAndProofs to MigratingEndpointAdapter

courtneyeh:
fixes #5676 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostContributionAndProofs.java
```diff
@@ -74,31 +93,39 @@ public PostContributionAndProofs(
...
+                  if (rootCause instanceof IllegalArgumentException) {
+                    return SafeFuture.of(
+                        () ->
+                            AsyncApiResponse.respondWithError(SC_BAD_REQUEST, error.getMessage()));
+                  }
+                  return SafeFuture.of(
+                      () ->
+                          AsyncApiResponse.respondWithError(
+                              SC_BAD_REQUEST, "Failed to submit contribution and proofs"));
+                }));
```
rolfyone: the illegal arg should be automatically handled, not sure about the second bit... would be interesting to see if its still the same result without the exceptionally...

courtneyeh: I'm thinking the whole `exceptionallyCompose` is probably not needed

rolfyone: I'll have to have a play, didnt get to it today but I'll look tomorrow morning :)

rolfyone: - validatorDataProvider will call either remoteValidatorApiHandler or ValidatorApiHandler
 - validatorApiHandler will throw IllegalArg, anything else is likely a bug
 - remoteValidatorApiHandler will decode then put onto validatorApiHandler, so will be JsonProcessingException or IllegalArg...
based on this, we're only really expecting an IllegalArgumentException or JsonProcessingException, and these are both handled by the generic exception handler.
If the old behaviour of an unexpected exception coming out were to actually happen, we'd potentially be missing useful information by swallowing it in the old way, so i think yes, remove the ExceptionallyCompose.
If we do see any 500 errors, we'll need to accept that tech debt, but whether 400 or 500, the answer is still that the POST failed.

rolfyone (approved):
LGTM after we remove that exceptionally...

PR #5696, opened 2022-06-03
Enable Error Prone's ClassName check

jtraglia:
## PR Description

Going through Error Prone's experimental checks. This check will enforce:

> The source file name should match the name of the top-level class it contains.

Seems reasonable; found a few problems.

## Documentation

## Changelog


ajsutton (approved):
LGTM. That is a handy check to have.

PR #5697, opened 2022-06-03
Enable Error Prone's DeduplicateConstants check

jtraglia:
## PR Description

This check will identify cases where:

> This expression was previously declared as a constant; consider replacing this occurrence.

## Documentation

## Changelog


ajsutton:
hmm, I'm not sure this is a good check but am quite surprised at how few problems it found so maybe it finds a good balance.  The issue with always using a constant is that just because the value is the same doesn't necessarily mean it's actually the same thing (though within a single class that's much less likely). It also reduces locality of reference - so you have to step through an level of indirection to find what the value is which just slows you down. You see that most when tests are using the same arbitrary value - they don't have to be the same value and it's easier to see what's actually being passed if the value is used directly (though sometimes it's clearer to use a constant because having it named adds value).

But I also admit that I overcorrect on this a bit because so many people blindly apply these kinds of rules so I'll look into how the rule actually works next week and see what others think as well. :)

jtraglia:
Yeah I totally understand. I was also surprised that there were so few findings. I'll leave it up to you on whether or not to keep this. I'm enabling these checks individually with that in mind, so we can pick and choose what we want.

ajsutton (approved):
Ok so it seems the reason this doesn't turn up much stuff is that it only complains if there is a constant declared for the value already (I assume in the same class). I'm not sure if that will actually add much value but I don't think it will be a nuisance or encourage bad patterns. Let's put it in and see how it goes.

PR #5701, opened 2022-06-04
Enable Error Prone's RedundantThrows check

jtraglia:
## PR Description

This will identify where:

> Thrown exception is a subtype of another.

## Documentation

## Changelog


ajsutton (approved):
LGTM. Seems to be very narrowly focussed to just declaring sub-classes which makes sense.

PR #5702, opened 2022-06-04
Enable Error Prone's RedundantOverride check

jtraglia:
## PR Description

This check will identify cases where:
> This overriding method is redundant, and can be removed.

For both of these classes, it seems that the overridden `equals` method needs to check the subtype. I was a bit confused as to why [EqualsHashCode](http://errorprone.info/bugpattern/EqualsHashCode) doesn't complain; I think it's because the `hashCode` method is overridden in the parent.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5703, opened 2022-06-04
Enable Error Prone's InterfaceWithOnlyStatics check

jtraglia:
## PR Description

The [InterfaceWithOnlyStatics](http://errorprone.info/bugpattern/InterfaceWithOnlyStatics) check will identify cases where:

> This interface only contains static fields and methods; consider making it a final class instead to prevent subclassing.

The docs provide a little more rationale:

> Interfaces should be used to define types. Using an interface as a collection of static methods and fields violates that, and can lead to confusing type hierarchies if the interface is then implemented to allow easy access to the constants.

I think this makes sense; I like this check.

## Documentation

## Changelog


ajsutton (approved):
Yep agreed. There's also been cases where I've wanted to refactor to pull out a private class but can't because it's an interface not a class.

PR #5705, opened 2022-06-05
Avoid division by zero when estimating time to TTD

ajsutton:
## PR Description
When there are no new blocks for a long period, all samples might be for the same block and thus have the same timestamp.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5715, opened 2022-06-06
[MEV Boost\Builder] remove "already in progress" warning, implements a last config validity period

### LINKED ISSUE #5648: Investigate proposer config loader warning
On kiln testnet I'm running proposer config with this params
```
  --validators-proposer-config proposerconfig.json \
  --validators-proposer-config-refresh-enabled \
```

and on every epoch I see

```
2022-05-31 17:10:36.001 WARN  - A proposer config load is already progress, waiting it instead of generating a new request
```

investigate if there is some issues (multiple onSlot() or other events that triggers reload)

tbenr:
now that we have `ValidatorRegistrator` that uses proposerConfigProvider to retrieve info, it is perfectly fine that we have multiple concurrent requests to provide the config.

So I removed the warning and improved the refresh logic which will provide last refreshed config for 5 minutes. Only after this period the config is considered expired and actually reloaded.

## Fixed Issue(s)
fixes #5648

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #5721, opened 2022-06-07
[MEV Boost\Builder] Add a readiness check for the fee recipient retrieval when registering

### LINKED ISSUE #5716: [MEV Boost\Builder] Validator Registration logs a warning when validator is not yet known by the beacon chain (pre deposit processing)
image

we see 100 of these:
```
{"@timestamp":"2022-06-06T14:05:00,012","level":"WARN","thread":"ValidatorTimingChannel-0","class":"ValidatorRegistrator","message":"There is no fee recipient configured for Validator{publicKey=0xb2eb468700f53334d9b8b0458c470d88ff13c5d6c1444412de7f4250901b2effdef689b7e08e3515fc98347b7a967
739}. Can't register.","throwable":""}
```

StefanBratanov:
## PR Description

- Changed signature of `FeeRecipientProvider.getFeeRecipient()` to add a readiness check.
- Implemented the readiness check in the `BeaconProposerPreparer` (one successful sending of proposers)
- Added the readiness check in `ValidatorRegistrator` in `onSlot` and `onValidatorsAdded`
- Changed the implementation of `getFeeRecipient` to be one optional pipeline using `or`
- Logging changes
- Changed the `warn` log to `debug` log for the missing fee recipient.

## Fixed Issue(s)
fixes #5716

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
@@ -100,10 +100,6 @@ public class BeaconProposerPreparer implements ValidatorTimingChannel, FeeRecipi
   // - default set by --validators-proposer-default-fee-recipient
   @Override
   public Optional<Eth1Address> getFeeRecipient(final BLSPublicKey publicKey) {
-    if (validatorIndexProvider.isEmpty()
-        || !validatorIndexProvider.get().containsPublicKey(publicKey)) {
-      return Optional.empty();
-    }
     Optional<Eth1Address> maybeEth1Address =
         maybeProposerConfig.flatMap(config -> getFeeRecipientFromProposerConfig(config, publicKey));
     if (maybeEth1Address.isPresent()) {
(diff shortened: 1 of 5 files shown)


tbenr:
I don't like that the two `getFeeRecipient` have two different semantics but this difference is not obvious.

I'm thinking about making them one again but adding a boolean flag in the signature like `excludeUnresolvable` or something that makes more sense.
WDYT?

StefanBratanov:
> I don't like that the two `getFeeRecipient` have two different semantics but this difference is not obvious.
> 
> I'm thinking about making them one again but adding a boolean flag in the signature like `excludeUnresolvable` or something that makes more sense. WDYT?

Agree about the one method. The problem i got was `GetFeeRecipient` endpoint using the `BeaconProposerPreparer` method without passing a `ProposerConfig`, so that's why i left both. The main goal is to make sure the `ValidatorRegistrator` is not dependent on the `ProposerConfigProvider` of the `BeaconProposerPreparer`. I was potentially thinking about changing `GetFeeRecipient` to also have a `ProposerConfigProvider` and remove the getFeeRecipient method which passes only the public key only alltogether. In that case, there would be only one method. What is your opinion?

tbenr (approved):
For the records, we continued the conversation in a meeting agreeing that there was room for improvements there.
LGTM

PR #5723, opened 2022-06-07
Ensure event thread is started in executeLater

jtraglia:
## PR Description

Noticed that the public `AsyncRunnerEventThread::executeLater` method would not check if the thread was started prior to executing which could result in a null pointer exception. The other execution methods would properly check this.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/eventthread/AsyncRunnerEventThread.java
+++ b/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/eventthread/AsyncRunnerEventThread.java
@@ -65,6 +65,9 @@ public class AsyncRunnerEventThread implements EventThread {
 
   @Override
   public void executeLater(final Runnable task) {
+    if (!started.get()) {
+      throw new IllegalStateException("EventThread not started");
+    }
     thread.runAsync(() -> recordEventThreadIdAndExecute(asSupplier(task))).reportExceptions();
   }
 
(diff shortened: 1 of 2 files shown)


infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/eventthread/AsyncRunnerEventThread.java
```diff
@@ -92,7 +95,7 @@ public <T> SafeFuture<T> executeFuture(Supplier<SafeFuture<T>> task) {
   @Override
   public void execute(final Runnable task) {
     if (!started.get()) {
-      return;
+      throw new IllegalStateException("EventThread not started");
```
ajsutton: This will cause a lot of noise at shutdown because `started` is set back to false on shutdown but we may still be sending tasks to the executor which we deliberately want to ignore.

jtraglia: Oops. Good call. I didn't think of that. I'll revert.

jtraglia: Added a comment there that's essentially a copy of your message; hope that's fine. Most of the other methods in that class will throw an exception if the thread isn't started; those are okay right?

ajsutton: Yeah those other ones return a failed future which is much more likely to be handled fairly smoothly. I do wonder if we should throw `RejectedExecutionException` rather than `IllegalStateException` but will leave it for now since it doesn't seem to be an issue.

ajsutton (approved):
LGTM. Thanks.

PR #5724, opened 2022-06-07
Enable Error Prone's PreferredInterfaceType check

jtraglia:
## PR Description

The [PreferredInterfaceType](http://errorprone.info/bugpattern/PreferredInterfaceType) check will:

> Where possible, prefer referring to some types by a subtype which conveys more information. For example: ImmutableList conveys more information (the [immutability](https://guava.dev/releases/21.0/api/docs/com/google/common/collect/ImmutableCollection.html) of the type) than List, List conveys more information than Collection, and ListMultimap conveys more information than Multimap.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsConfig.java
+++ b/infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsConfig.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.infrastructure.metrics;
 import com.google.common.collect.ImmutableSet;
 import java.net.URL;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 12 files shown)


infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsConfig.java
```diff
@@ -27,11 +28,13 @@
...
-      ImmutableSet.<MetricCategory>builder()
-          .addAll(EnumSet.allOf(StandardMetricCategory.class))
-          .addAll(EnumSet.allOf(TekuMetricCategory.class))
-          .build();
+  public static final Set<MetricCategory> DEFAULT_METRICS_CATEGORIES =
+      Collections.unmodifiableSet(
+          ImmutableSet.<MetricCategory>builder()
+              .addAll(EnumSet.allOf(StandardMetricCategory.class))
+              .addAll(EnumSet.allOf(TekuMetricCategory.class))
+              .build());
```
jtraglia, line 36: Not 100% sure this is the proper way to fix the warning. The `ImmutableSet.builder` and `EnumSet.allOf` are used together (quite cleverly!) to create a list of default metric categories. I think this is better than using `Set.of` (which would be immutable) and manually specifying each enum; which could result in problem if a new category is added and we forget to update this set. I think wrapping it in `Collections.unmodifableSet` is the better approach.

To be clear, the problem is that the check wants a more specific type (`ImmutableSet`) here:

https://github.com/ConsenSys/teku/blob/91c03c1b323c9551286bd10332624e550868f7a3/infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsConfig.java#L119

But using that type there causes lots of problems in other places, particularly testing.

ajsutton:
I'm not a fan of this check to be honest.  The usual advice is actually to go the other way and use the most generic interface
 that makes sense because then it can be used in the most places without conversion.  There's some nuance here in that using a more generic type for parameters is the most useful case and a more specific type for what you create yourself probably makes sense but there's a real tendency for more specific types to flow through to method definitions when using IDE refactoring functions.

If the more specific type is actually useful I think it quite naturally gets used. And even with this rule you can't assume that every `List` you get is mutable so you still can't really trust the information.

jtraglia:
Sounds good to me. Closing.

PR #5503, opened 2022-05-15
Unhide execution_optimistic flag

ajsutton:
## PR Description
Always include the `execution_optimistic` flag in responses and unhide it from the API schema.

## Fixed Issue(s)
fixes #5264 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #5725, opened 2022-06-07
Prevent unboxing of a nullable variable

jtraglia:
## PR Description

Recently, `executionOptimistic` was changed from `Boolean` to `boolean` but the (old style) deserializer still treated the field as optional and as a result it could be null. When `executionOptimistic` is null, it would throw a null pointer exception when unboxing the value for the `GetStateResponseV2` constructor. To be fair, it doesn't make that much of a difference; it will now throw a NPE a little earlier if the field is missing.

## Documentation

## Changelog

### DIFF
--- a/data/serializer/src/main/java/tech/pegasys/teku/provider/GetStateResponseV2Deserializer.java
+++ b/data/serializer/src/main/java/tech/pegasys/teku/provider/GetStateResponseV2Deserializer.java
@@ -39,9 +39,7 @@ public class GetStateResponseV2Deserializer extends JsonDeserializer<GetStateRes
     JsonNode node = jp.getCodec().readTree(jp);
     final Version version =
         Version.valueOf(node.findValue("version").asText().toLowerCase(Locale.ROOT));
-    final JsonNode executionOptimisticNode = node.findValue("execution_optimistic");
-    final Boolean executionOptimistic =
-        executionOptimisticNode != null ? executionOptimisticNode.asBoolean() : null;
+    final boolean executionOptimistic = node.findValue("execution_optimistic").asBoolean();
     final BeaconState state;
     switch (version) {
       case altair:


data/serializer/src/main/java/tech/pegasys/teku/provider/GetStateResponseV2Deserializer.java
```diff
@@ -39,9 +39,7 @@ public GetStateResponseV2 deserialize(final JsonParser jp, final Deserialization
     JsonNode node = jp.getCodec().readTree(jp);
     final Version version =
         Version.valueOf(node.findValue("version").asText().toLowerCase(Locale.ROOT));
-    final JsonNode executionOptimisticNode = node.findValue("execution_optimistic");
-    final Boolean executionOptimistic =
-        executionOptimisticNode != null ? executionOptimisticNode.asBoolean() : null;
+    final boolean executionOptimistic = node.findValue("execution_optimistic").asBoolean();
```
rolfyone, line 42: Probably need to be careful on this change - theres definitely cases where execution_optimistic isn't set, so null can have meaning... 
It defaults to false, so could potentially attempt to find the value and default to false if it's not present...

jtraglia: If the field isn't present, I would expect `node.findValue(...)` to throw an exception when dereferenced. It will only default to false if there's a value (any value). But are you sure there are cases where it's not set? The original code would throw a NPE in that case. PR #5503 changes it to "always include the `execution_optimistic` flag in responses."

rolfyone: that'll be why it's checking null is all, no field existed, so set value to null... I'll have a closer look once i clean up my current PR 'mess' :)

jtraglia: Thank you. Just to be clear, if you call `new GetStateResponseV2(version, executionOptimistic, state)` when `executionOptimistic` is null, it will throw an exception. This is because the parameter is a primitive:

https://github.com/ConsenSys/teku/blob/e8b4ce8ef582d158b96a3f74bbefe7f62a448185/data/serializer/src/main/java/tech/pegasys/teku/api/response/v2/debug/GetStateResponseV2.java#L46-L49

rolfyone: fair point... I'm not sure where it'd ever be used now, which was the other aspect, but this change certainly hasn't made anything worse... :)

rolfyone (approved):
LGTM

PR #5730, opened 2022-06-08
Combined Database

ajsutton:
## PR Description
Adds a `KvStoreCombinedDatabase` as a first step towards taking advantage of the fact that our hot and finalised database are the same underlying instance in most schemas.  The legacy V4 (RocksDB) schema still uses separate instances so the original `KvStoreDatabase` exists to continue to support that.

TODO: 

 * [x] This will currently break the offline migration support. Will need to work out how to fix that.
 * [x] Investigate using `KvStoreCombinedDatabase` even for V4 dbs by adding an adapter to the combined dao so `combinedUpdater()` returns an updater that encapsulates a hot and a finalized transaction.
 * [x] Investigate if there's a way to share the majority of code between `KvStoreDatabase` and `KvStoreCombinedDatabase`. May not be needed if we can just use a dao adapter instead.
 * [x] Add tests for backwards compatibility

## Fixed Issue(s)
#5312 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5732, opened 2022-06-08
[MEV Boost\Builder] Change Builder proposed Gas Limit check to be just a warning log

tbenr:
As per discussions on discord, let's demote gaslimit check to just a warning when GL doesn't move in the expected direction.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #5744, opened 2022-06-09
rename reportExceptions

rolfyone:
There's actually very few occasions where it's desirable to use `reportExceptions`, renaming to hopefully make that clearer.

Potentially most of the locations on this changelist need changing to use .finish with an error message.

## Documentation

## Changelog


ajsutton (approved):
LGTM. Unusual name but should discourage use suitably.

PR #5745, opened 2022-06-09
Merge banner art 🐼 #5231

### LINKED ISSUE #5231: Merge banner art 🐼

issanyo:
## PR Description
Merge banner art 🐼 

## Fixed Issue(s)
fixes #5231

## Documentation

- A panda emoji appears at the point when Teku confirms that the TTD has been reached on testnets and mainnet.

## Changelog

- A panda emoji appears at the point when Teku confirms that the TTD has been reached on testnets and mainnet.


tbenr:
Good point! But in teku we do Pandas with an explicit opt-in.
So you should provide a logic similar to: https://github.com/ConsenSys/teku/blob/0bce78ec4019780a83b5c156fd8b4ef8bf4844c2/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java#L178-L185

tbenr:
But consider that this will not fix the issue, which requires to provide a big ascii art banner. There are strict requirements with Pandas :-)

issanyo:
> But consider that this will not fix the issue, which requires to provide a big ascii art banner. There are strict requirements with Pandas :-)

Please check the new banner, here is it as well for better visibility:

                    ,╔▄▄▄▄▄µ,                                ,╓▄▄▄▄▄╦╓,
               ╓▄▓▓▓███████▓▓▓▓▄,                       ╓▄▓▓▓▓███████▓▓▓▄
             ,▓████▓▓╬╣╣╣╣╬╬╬▓▓█▓▓▓╦, .,,--»»»--,.  ,▄▓▓██▓▓╬╬╬╬╣╣╬╬▓▓████▌
            ▄███▓╬╬╬╬╬╬╬╬╬╬╬╬╬╣╬▀╬╙░░░░░░░░░░░░░░░░░░░╨╬▀╣╣╬╬╬╬╬╬╬╬╬╬╬╣╬████┐
           ▐███╬╬╬╬╬╬╬╬╬╬╬╬╬╩╚Γ░░░░░░!!!!!!!!!!!!!!!░░░░░░Γ╚╠╣╬╬╬╬╬╬╬╬╬╬╣▓███,
           ███╬╬╬╬╬╬╬╬╬╬╢╩Γ░░░░░∩!!!:⌐'^^^^^^^^^^':::;!!!░░░░Γ╙╠╬╬╬╬╬╬╬╬╣╣▓██▌
          ]███╬╬╬╬╬╬╬╬╠╚Γ░░░░!\!:'^^^^^'''''''''''^^^^^::!!!░░░░Γ╚╠╬╬╬╬╬╬╣╣██▓
           ███╬╬╬╬╬╬╬╩Γ░░░░\:⌐^^^^~''             ''''''^^^::!░░░░Γ╠╬╬╬╬╬╬╣██▌
           ║██▓╬╬╬╬╬▒Γ░░░!⌐^^^^'''                    '''^'^^^:!░░░Γ╚╣╬╬╬╬███⌐
            ╫██▓╬╣╩Γ░░░\⌐^^^~~'                         '.~~^^^^:!░░ΓΓ╢╬╬▓██╩
             ╫██▌▒Γ░░\⌐^^^~.'.                             .''~^^^:!░░Γ╩▓██╩
              ╚▌Γ░░!⌐^^^'.'                                ' ...'^^^:!░░╚▓┘
              φΓ░!:'^^.'.                                      .''^^'┌!!░Γ≥
            «░░!:⌐'^~''    '',╖▄▄▄▄▄,               ,▄▄▄▄▄╓''    '.''^^:┐░░φ
           φ░░!⌐^^~'     ',▄██████████▌,         ╓▓█████████▓▄'.    '''^┌:!░░»
          φ░░:^^~'     ',▓██████████████▄      .▓██████████████▄'     '''^:!░░ε
         φ░░\^^'       ]█████████████████▌     ▓████████████████▓       ''^:!░░ε
        φ░░\'^~       .████████  █████████⌐   ║█████████▀▀███████▌       ''^:!░░~
        ▒░░:^~        ╫███████ .  ████████▒   ╫████████ .  ███████Γ       '^^;░░φ
       ]▒░┐⌐^.        ▓███████    ████████▒   ╫████████    ███████▌       '.^┌░░▒
       φ▒░┐^~.        ▓████████  ████████▌    '█████████  ████████▌        .^┌¡░▒~
       ]▒░┐┌~.        ╫█████▓██  ██▓████▀┌'   .└▓███▓▓██  ████████Γ       '.,:░░▒~
       '▒░░┐...       '▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▀│┐^.....┌░╙▓▓▓▓▓▓▓▓▓▓▓▓▓▓█▀        ..┌¡░░╚
        ╚▒░░┌...        ╨▓▓▓▓▓▓▓▓▓▓▀╩░░░░,,,,,,;¡░░┤╬▓▓▓▓▓▓▓▓▓▓▓╙        ..┌;░░▒⌐
         ╚▒░░┌┌..         ╙▀▀▓▓▀▀╨░░░╦▓▓▓▓▓▓▓▓▓▓▓▓▌░░░├╠▀▓▓▓▀╝└         .,┌;░░▒Γ
          ╚▒░░\┌...           .░░'^' ╫█▓▓▓▓▓▓▓▓▓▓▓█Γ'^'!░░            ..┌:¡░φ▒╙
           ╙▒░░░;┌.....      .φ░⌐'    ╨██████████▓╙   '';░░       ....┌:¡░░▒╚'
             ╙▒▒░░;┌┌........:░░».      ╚██████▓╜      .¡φ░┌.....,,┌┌;░░φ▒╚'
               '╚▒▒φ░░░░░░░░░░░▒░░,..     ╨▀▓▀┴     ..;░φ▒░░░░░░░░░░φφ▒╚╙`
                   '╙╚╚╚╚╚╚╚╚╚╚╚╩▒▒φφ░░=ⁿ'''^''ⁿ»=≥φφφ▒▒╚╚╚╚╚╚╚╚╚╚╚╙└`
                                     '╚φ░░;,,,;░φδ╙`
                                         ^'└╙''`

rolfyone:
That panda looks amazing! 
One question that I do have to ask, is it sourced from anywhere? If you could provide a link if it is, just so we can double check any licensing is ok to be used, and whether theres rules like attributing or including in its original form or anything like that...

issanyo:
> That panda looks amazing! One question that I do have to ask, is it sourced from anywhere? If you could provide a link if it is, just so we can double check any licensing is ok to be used, and whether theres rules like attributing or including in its original form or anything like that...

Thanks, I used https://asciiart.club/ to help me generate a first version which I modified and customized to make it look better :)

rolfyone:
Ok so I think the art is great, and we should keep it. The best place is probably to create a txt file and store it against the issue if this PR ends up not being merged...
The coding side is probably more complex than initially thought...
 - we need to find the place that this needs to be called - calling it at the current location would not really notify at the right time, and it'd potentially be pretty noisy on restart.
 - we need to make sure it actually only comes out once, and never again, we don't want it every time teku starts...
 - Ideally, it could be 1 log info line (otherwise might get interrupted), but that may be fairly complex because of the earlier log4j mitigations that happened.
 - we may also need to mitigate line trimming, but I haven't spent a lot of time to actually run this code (would need to get it compiling, spin up a network etc)

I'm not sure what your level of comfort is with investigating any of the above, so if that sounds too complicated, storing the txt file against the original issue would give someone a huge leg up when they're looking at the underlying complexities of the issue...

Let me know what you think, it'd be a bit of investigation etc, and you'd likely need to spin up custom networks to actually go through merge and reach TTD, restart the instance and make sure it doesn't come out again, maybe some tests, so I understand if it's getting in a lot further than you're comfortable with, but equally we can answer questions and try to help if you'd like to keep digging!

issanyo:
> Ok so I think the art is great, and we should keep it. The best place is probably to create a txt file and store it against the issue if this PR ends up not being merged... The coding side is probably more complex than initially thought...
> 
> * we need to find the place that this needs to be called - calling it at the current location would not really notify at the right time, and it'd potentially be pretty noisy on restart.
> * we need to make sure it actually only comes out once, and never again, we don't want it every time teku starts...
> * Ideally, it could be 1 log info line (otherwise might get interrupted), but that may be fairly complex because of the earlier log4j mitigations that happened.
> * we may also need to mitigate line trimming, but I haven't spent a lot of time to actually run this code (would need to get it compiling, spin up a network etc)
> 
> I'm not sure what your level of comfort is with investigating any of the above, so if that sounds too complicated, storing the txt file against the original issue would give someone a huge leg up when they're looking at the underlying complexities of the issue...
> 
> Let me know what you think, it'd be a bit of investigation etc, and you'd likely need to spin up custom networks to actually go through merge and reach TTD, restart the instance and make sure it doesn't come out again, maybe some tests, so I understand if it's getting in a lot further than you're comfortable with, but equally we can answer questions and try to help if you'd like to keep digging!

Thanks for your reply, actually I didnt expect that complexity, I thought just a log line with panda banner in that position would be fine. 
I'll put this inside a text file and leave for you guys :)

rolfyone:
we can ref this PR when we go for art, so will close this, as it's linked to issue.

PR #5747, opened 2022-06-09
Add independent data expiration in `ProposersDataManager`

### LINKED ISSUE #5737: `ProposersDataManager` data expiration problem
`preparedProposerInfoByValidatorIndex` and `validatorRegistrationInfoByValidatorIndex` expiration should not relay on an API triggering a corresponding update. If suddenly a VC stops calling updates, data never expires.

We should subscribe `ProposersDataManager` to `SlotEventsChannel` and expire data `onSlot`

tbenr:
Makes sure that we are cleaning up prepared proposers and validator registrations when corresponding API calls are no more called. The cleanup happens on every second slot of each epoch.

Note: We also forgotten to expire validator registration, now it is covered by this PR.

## Fixed Issue(s)
fixes #5737

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #5751, opened 2022-06-09
Add optimistic sync acceptance tests

### LINKED ISSUE #5074: Bellatrix Acceptance Test
### Description
We should have acceptance tests that checks teku behaviour with Bellatrix is activated, ideally going through the transition block.
We could think about adding complexity by test several scenarios (ie. optimistic sync)

An option is to have extended AT by running a fixed besu client version capable of running as EL.

Another option is to start using kurtosis.

https://docs.kurtosistech.com/

https://github.com/kurtosis-tech/eth2-merge-kurtosis-module

Pari is already running it daily to check various client compatibility:

https://github.com/parithosh/nightly-kurtosis-test/blob/main/.github/workflows/kurtosis.yml
https://github.com/parithosh/nightly-kurtosis-test/actions

additional examples:
https://notes.ethereum.org/@ExXcnR0-SJGthjz1dwkA1A/H11OzhRAK

zilm13:
## PR Description
We have already straightforward happy case in `BellatrixMergeTransitionAcceptanceTest`, this PR adds both possible optimistic sync cases.
Notes about this PR:
- `OptimisticSyncPostMergeAcceptanceTest` verifies accidentally syncing execution layer after the merge, `OptimisticSafeSlotsSyncPostMergeAcceptanceTest` verifies that we could pass even merge slot optimistically after `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` with optimistic import if execution layer is syncing
- I've removed `.withLogConsumer(frame -> LOG.debug(frame.getUtf8String().trim()))` line from `BesuNode` initialization, it duplicates log output, maybe I'm missing something, but there is one already in `Node`
- `getInternalIpAddress()` in `BesuNode` looks ugly, but I have not found any other way to get internal container IP, hosts are not allowed in `enode`
- `OptimisticSafeSlotsSyncPostMergeAcceptanceTest` needs 5 minutes between `StatusMessage` updates in order to resume sync, so this test takes a long time. Some way to disconnect teku nodes connection and then connect  them again could be an alternative, I have tried all ways to do this, but none is working

PS: Starting with draft to do several CI runs to clarify non-flaky

## Fixed Issue(s)

Fixes #5074 
## Documentation

## Changelog


zilm13:
Regarding https://github.com/ConsenSys/teku/pull/5751/commits/8f4ef822bcbc90fce9ca25602da9ff0510d450fd
In my test case I stop EL in the middle, then start new instance, so it's down for a while. We are on a gossip sync at this moment. We get in this case following chain:
SUCCESS<-SUCCESS<-FAILED_EXECUTION_PAYLOAD_EXECUTION<-UNKNOWN_PARENT<-UNKNOWN_PARENT...
First of all, `ReexecutingExecutionPayloadBlockManager` had incorrect comparison so it pruned the block in 1 slot. I've fixed this and added test for this case in commit. At the second, why just 2 slots. Even on mainnet it's just 24 seconds and could be not enough to restart EL which suddenly gone down. I'm not sure what would be the best constant for it, say in rare case we should be able to drop bad block braking EL when it's not a correct canonical chain payload, in all other cases we'd better retry forever. 
If we pruned it in 20 slots with this fix commit, do we have a way to recover the sync without Teku restart? If, say, EL was down for 25 slots? From my understanding we are not able to pass this until we switch to BatchSync when head gap is big

ajsutton:
Sorry to be so slow in reviewing this one.  I need to be able to focus and think it through properly but haven't managed to find that time yet. Will try to get to it tomorrow.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
```diff
@@ -92,7 +92,7 @@ protected SafeFuture<?> doStop() {
   public void onSlot(UInt64 slot) {
     super.onSlot(slot);
     pendingBlocksForEPReexecution.removeIf(
-        block -> block.getSlot().isLessThan(slot.plus(EXPIRES_IN_SLOT)));
+        block -> block.getSlot().isLessThan(slot.minusMinZero(EXPIRES_IN_SLOT)));
```
ajsutton: Probably worth splitting this change out into a separate PR.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
```diff
@@ -35,7 +35,7 @@
 
 public class ReexecutingExecutionPayloadBlockManager extends BlockManager {
   private static final Logger LOG = LogManager.getLogger();
-  private static final int EXPIRES_IN_SLOT = 2;
+  protected static final int EXPIRES_IN_SLOT = 20;
```
ajsutton: We don't need to keep retrying these blocks long - we might be triggering a bug in the EL with them so repeatedly trying for a long time could cause issues.  And if we do drop them from here but they're needed as part of the ongoing chain we'll wind up fetching and importing them again as part of the normal sync and gossip handling processes (just like a block we somehow didn't get via gossip).

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/AcceptanceTestBase.java
```diff
@@ -90,7 +91,7 @@ protected TekuDepositSender createTekuDepositSender(final String networkName) {
   }
 
   protected BesuNode createBesuNode() {
-    return createBesuNode(config -> {});
+    return createBesuNode(Config::withMiningEnabled);
```
ajsutton: Probably better to just change the default value in BesuConfig and provide a `withMiningEnabled(boolean)` so it can be disabled if needed. Otherwise switching to the other version of this method to configure some extra fields suddenly disables mining which will be hard to understand.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSafeSlotsSyncPostMergeAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    .withPeers(tekuNode1)
+                    .withSafeSlotsToImportOptimistically(SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY));
+    tekuNode2.start();
+  }
+
+  @Test
+  void shouldPassMergeOptimisticallyAndBeginFinalizationAfterSafeSlotsToImport() throws Exception {
+    tekuNode2.waitForGenesis();
+    // Disconnect eth1Node2 from eth1Node1 before the merge
+    assertThat(eth1Node2.removePeer(eth1Node1Enode)).isTrue();
```
ajsutton: There's nothing to guarantee this is pre-merge.  We'd be better off starting the node with no peers and then adding the peer when we want it to start syncing.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSafeSlotsSyncPostMergeAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+    tekuNode1.waitForLogMessageContaining("MERGE is completed");
+
+    tekuNode2.waitForLogMessageContaining("optimistic sync");
```
ajsutton: There could be a lot of messages that contain optimistic sync.  We probably need to be more specific about what we're looking for.  Likely the best thing is to retrieve the head block and check that the `execution_optimistic` field is true for it which indicates it was imported optimistically.  That would also only be true once Teku had sync'd past the actual merge block so could probably replace all the "MEREGE completed" log message as well.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSafeSlotsSyncPostMergeAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    .withSafeSlotsToImportOptimistically(SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY));
+    tekuNode2.start();
+  }
+
+  @Test
+  void shouldPassMergeOptimisticallyAndBeginFinalizationAfterSafeSlotsToImport() throws Exception {
+    tekuNode2.waitForGenesis();
+    // Disconnect eth1Node2 from eth1Node1 before the merge
+    assertThat(eth1Node2.removePeer(eth1Node1Enode)).isTrue();
+    tekuNode1.waitForLogMessageContaining("MERGE is completed");
```
ajsutton: It would be better to wait until the head block has a non-default execution payload.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSafeSlotsSyncPostMergeAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+import tech.pegasys.teku.test.acceptance.dsl.TekuNode;
+import tech.pegasys.teku.test.acceptance.dsl.tools.deposits.ValidatorKeystores;
+
+public class OptimisticSafeSlotsSyncPostMergeAcceptanceTest extends AcceptanceTestBase {
+  private static final String NETWORK_NAME = "less-swift";
+  private static final Integer SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY = 4;
+
+  private final SystemTimeProvider timeProvider = new SystemTimeProvider();
+  private BesuNode eth1Node1;
+  private BesuNode eth1Node2;
```
ajsutton: In terms of terminology we probably should call these `executionNode1` or something more execution layer related.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSafeSlotsSyncPostMergeAcceptanceTest.java
```diff
@@ -0,0 +1,115 @@
...
+    tekuNode1.waitForLogMessageContaining("MERGE is completed");
+
+    tekuNode2.waitForLogMessageContaining("optimistic sync");
+    tekuNode2.waitForLogMessageContaining("MERGE is completed");
+    tekuNode2.waitForNonDefaultExecutionPayload();
+    tekuNode2.waitForNewFinalization();
+    Optional<GetBlockRootResponse> getBlockRootResponse2 = tekuNode2.fetchBeaconRootHead();
+    assertThat(getBlockRootResponse2.get().execution_optimistic).isTrue();
+    Optional<GetBlockRootResponse> getBlockRootResponse1 = tekuNode1.fetchBeaconRootHead();
+    assertThat(getBlockRootResponse1.get().execution_optimistic).isFalse();
```
ajsutton: We shouldn't be exposing `fetchBeaconRootHead` as a getter style here - it should be a wait method like `waitForHeadExecutionOptimistic` so that the DSL automatically waits.  Otherwise it's too easy to introduce intermittency into tests by not waiting for the right thing (or because later the thing to wait for changes and we don't find all the places in tests to update).

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSyncPostMergeAcceptanceTest.java
```diff
@@ -0,0 +1,125 @@
...
+    tekuNode1.waitForNonDefaultExecutionPayload();
+    tekuNode2.waitForNonDefaultExecutionPayload();
+
+    // Stopping eth1Node2
+    final String eth1Node2Alias = eth1Node2.getNodeAlias();
+    eth1Node2.stop();
+    // Set same network name to the eth1Node3 that eth1Node2 had
+    eth1Node3.setNodeAlias(eth1Node2Alias);
+    eth1Node3.start();
+    eth1Node3.waitForLogMessageContaining("Ethereum main loop is up");
```
ajsutton: We seem to be trying to just reset eth1Node2's database to force it to resync?  We really want to hide the details of this from the AT class so we can just have a method like `eth1Node2.restartWithEmptyDatabase()`. I'd be very tempted to use `container.execInContainer` to just delete the database then restart the existing container rather than recreating one with the same alias.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
zilm13: So keeping 2 slots here?
My guess is that we will try to re-request these blocks after StatusMessage exchange, which is often with, say, 30 peers connected

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
zilm13: Thought about it, detaching

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
zilm13: But it's not often in non-mainnet cases

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
zilm13: detached https://github.com/ConsenSys/teku/pull/5786

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSafeSlotsSyncPostMergeAcceptanceTest.java
zilm13: >  We'd be better off starting the node with no peers and then adding the peer when we want it to start syncing.

I don't understand this. What's your proposed test flow?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
ajsutton: I'd keep it at 2 slots yeah.  We'll be triggered to re-request and process the block as soon as we receive a new block from the network as we'll then find we don't have all the blocks leading to it and the BlockPropagationManager will pull them in.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSyncPostMergeAcceptanceTest.java
```diff
@@ -0,0 +1,103 @@
...
+  @Test
+  void shouldSwitchToOptimisticSyncAfterMergeWhenExecutionEngineIsSyncing() throws Exception {
+    tekuNode2.waitForGenesis();
+
+    // Reset execution client's DB after the merge and leave it without any chance to sync
+    tekuNode2.waitForNonDefaultExecutionPayload();
+    executionNode2.removePeer(executionNode1);
+    executionNode2.waitForRestartWithEmptyDatabase();
+
+    tekuNode2.waitForOptimisticBlock();
```
ajsutton: So I think this looks good now - maybe I was misunderstanding the flow of the previous test which is why my comment wasn't clear.

So from what I can see we have the EL (executionNode2) in sync so that tekuNode2 can transition through the merge block normally.  Then we reset executionNode2 and ensure it can't sync and teku should continue importing blocks optimistically.

It may be worth then re-adding the peer to executionNode2 and we can then wait for Teku's head block to not be optimistic anymore which would check that the EL is syncing properly.  If it makes the test take too long it may not be worth it, but if it can re-sync reasonably quickly it could be a useful test.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/BesuNode.java
```diff
@@ -73,6 +84,23 @@ public void start() throws Exception {
     container.start();
   }
 
+  public void waitForRestartWithEmptyDatabase() {
+    Waiter.waitFor(() -> assertThat(restartWithEmptyDatabase()).isCompleted());
```
ajsutton: This is probably a case where we shouldn't need a waiter since deleting the db is a one-off. Was this intermittently failing to restart without this?

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSyncPostMergeAcceptanceTest.java
zilm13: Ok, I will add this, don't think it will take long, but maybe I'm wrong. 

My question is about another, `OptimisticSafeSlotsSyncPostMergeAcceptanceTest` test. I had following flow there:
Start:
Teku1 (with vals) -> Besu1 (mining)
Teku2 (no vals, connected to Teku1) -> Besu2 (no mining, connecting to Besu1)
After BeaconChain genesis:
Besu2 disconnects from Besu1 (it's pre-merge, and I agree with you that nothing guarantee that I'm in time until the merge there, plus I should clarify that it happens on Bellatrix, I guess), now it's syncing with no luck to sync
After that in `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` Teku2 should start to import blocks and pass the merge optimistically. Your comment was on that test. Plus there is an issue with it that Teku will not start import in `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` because `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` is more than 2, so we drop payload re-execution and have gap in the chain. We request non-gossip sync only on `StatusMessage` exchange which is once in 5 minutes. So it will start to import optimistically only in 5 minutes, not after 4 slots which is `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` if re-execution is dropped on the moment when safe slots are passed.

In `OptimisticSafeSlotsSyncPostMergeAcceptanceTest` I wanted to test second optimistic import condition which could let us even pass merge optimistically. It's not very good idea to wait 5 minutes in test, so I thought you had some idea how to rework it.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/BesuNode.java
zilm13: Shouldn't be an issue, will try

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSyncPostMergeAcceptanceTest.java
ajsutton: So my understanding of the test is that we have a teku node which gets a SYNCING response from it's EL for the merge transition block (first one with a non-default payload) and it should pause imports until `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` then proceed to optimistically import.

Given there's no need to have an EL for blocks prior to the merge transition, my suggestion would be to simply start Besu2 with no peers and it will remain at genesis and return SYNCING all the time. Then we sync teku normally. If we later wanted besu to sync we could then add the peer but that's probably not necessarily.

`SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` is hard coded to 128 which is about 12 minutes on minimal, and 6.4 minutes on the `swift` config. So that will be the determining factor in how long it has to wait.  A test that takes that long isn't going to be valuable though so I'd suggest we just drop this test. Optimistically syncing through the merge is a more detailed case we'd be better off testing with integration tests.

zilm13: > Given there's no need to have an EL for blocks prior to the merge transition, my suggestion would be to simply start Besu2 with no peers and it will remain at genesis and return SYNCING all the time.

Teku will wait forever for BeaconChain genesis event in this case, it couldn't verify it

> SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY is hard coded to 128 

I've overrided it to 4 slots, we have command-line parameter for it and it works, the issue is in re-execution strategy. We stop re-execution after 2 slots, after that we have a chain:
SUCCESS <- SUCCESS <-[nothing]<- UNKNOWN_PARENT <- UNKNOWN_PARENT
We are unable to pass it on gossip, we try to sync after `StatusMessage` exchange which is once in 5 minutes with every node, but we have 1 peer in test, so exactly 5 minutes. Even with SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY == 4 we are importing blocks only after 5 minutes. I just want to clarify that we are ok with this behavior (as we have a lot of peers on mainnet and some `StatusMessage` arrives in seconds there and we need to protect EL from bad re-execution). If it's normal, I just drop safe slots test because it will take 6 minutes which is not acceptable

zilm13:
@ajsutton I've removed safe slots long running test, but, please, notice [this comment](https://github.com/ConsenSys/teku/pull/5751#discussion_r899894681)
The remaining test is broken because of Besu issue https://github.com/hyperledger/besu/issues/3988
So it happens after Besu starts syncing again, last test assertion which proves that we could switch back to normal sync. We could either wait for fix or remove this check

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/OptimisticSyncPostMergeAcceptanceTest.java
ajsutton: > Teku will wait forever for BeaconChain genesis event in this case, it couldn't verify it

Ah, it probably makes sense to switch from sending deposits for the validators to just using interop mode. We don't need to be testing genesis generation here so can simplify things and speed up the test by using interop mode.

> I've overrided it to 4 slots, we have command-line parameter for it and it works, the issue is in re-execution strategy. We stop re-execution after 2 slots, after that we have a chain:
SUCCESS <- SUCCESS <-[nothing]<- UNKNOWN_PARENT <- UNKNOWN_PARENT
We are unable to pass it on gossip, we try to sync after StatusMessage exchange which is once in 5 minutes with every node, but we have 1 peer in test, so exactly 5 minutes. Even with SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY == 4 we are importing blocks only after 5 minutes. I just want to clarify that we are ok with this behavior (as we have a lot of peers on mainnet and some StatusMessage arrives in seconds there and we need to protect EL from bad re-execution). If it's normal, I just drop safe slots test because it will take 6 minutes which is not acceptable

This is definitely not behaviour we want, but continuing to try re-executing the block is not the right solution.  We may need to adjust our gossip handling to ensure it retries execution when we get a new descendent of a failed block (in a separate PR).

zilm13:
I've restored safe slots test with interop setup, it looks much simpler now by your advice, but stalls for the same reason that is mentioned above for 5 minutes. I've cheated with `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` == 1 but it could be flaky with this settings.
I've created issue for 5 minutes problems https://github.com/ConsenSys/teku/issues/5824
We couldn't verify returning to normal sync until this issue https://github.com/hyperledger/besu/issues/3988 is fixed and reached some Besu release, Besu fails to sync when the peer is added post-merge in both tests

zilm13:
I think we shouldn't merge this at least until #5824 is fixed

influ7:
Ty

PR #5764, opened 2022-06-10
Check if discoveryConfig is null before dereference

jtraglia:
## PR Description

Noticed that `checkNotNull(discoveryConfig)` was called after dereferencing `discoveryConfig`. I believe this check should happen _before_ dereferencing the value. I also noticed some missing null checks in `build()`, notably `spec` and `currentSchemaDefinitionsSupplier` could have been null; added checks for all of the arguments to be safe.

## Documentation

## Changelog


rolfyone (approved):
LGTM - fair points, all :)

PR #5766, opened 2022-06-10
Change default archive storage mode to leveldb2

### LINKED ISSUE #5746: Switch default database format back to leveldb2
### Description
Teku currently defaults to `leveldb-tree` format which uses a lot of disk space (though queries are fast) by default. This turns out to not be what most users want even when they are doing an archive sync. We should switch the default format back to `leveldb2` which uses the snapshot based storage.

When `--data-storage-archive-frequency` is set to 1 we should automatically use `leveldb-tree` since it will then be the more space efficient storage format.

rolfyone:
Also, if using storage-frequency of 1, default back to leveldb-tree.

fixes #5746

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/BeaconNodeDataOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/BeaconNodeDataOptions.java
@@ -59,7 +59,7 @@ public class BeaconNodeDataOptions extends ValidatorClientDataOptions {
       description = "Database version to create",
       arity = "1",
       hidden = true)
-  private String createDbVersion = DatabaseVersion.DEFAULT_VERSION.getValue();
+  private String createDbVersion = null;
 
   @CommandLine.Option(
       names = {"--data-storage-non-canonical-blocks-enabled"},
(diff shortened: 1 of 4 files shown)


teku/src/main/java/tech/pegasys/teku/cli/options/BeaconNodeDataOptions.java
```diff
@@ -107,6 +107,9 @@ public void configure(final TekuConfiguration.Builder builder) {
   }
 
   private DatabaseVersion parseDatabaseVersion() {
+    if (dataStorageFrequency == 1) {
+      return DatabaseVersion.LEVELDB_TREE;
+    }
     return DatabaseVersion.fromString(createDbVersion).orElse(DatabaseVersion.DEFAULT_VERSION);
```
ajsutton: We don't want to force switching to LEVELDB_TREE if something else is explicitly specified - the user may be trying to force using rocksdb or something like that (or it may be us trying to test). Given it's a hidden option I'd probably just have the field default to null and then you can select leveldb2 or leveldb-tree based on the frequency when version is `null` and use whatever is specified when it isn't. It won't show the default value then but I don't think that's a big deal for a hidden option.

ajsutton (approved):
LGTM.

PR #5768, opened 2022-06-10
Default to burn address for fee recipient

### LINKED ISSUE #5763: Default to 0x00 fee recipient if not set
### Description
Rather than failing to produce a block, Teku should use 0x00 as the fee recipient if none is set (and continue to log very loudly).

rolfyone:
Rather than forcing inactivity penalties, default to the burn address if eth1address has not been specified.

fixes #5763

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/ValidatorLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/ValidatorLogger.java
@@ -184,6 +184,15 @@ public class ValidatorLogger {
             Color.YELLOW));
   }
 
+  public void executionPayloadPreparedUsingBurnAddressForFeeRecipient(final UInt64 slot) {
+    log.error(
+        ColorConsolePrinter.print(
+            "Producing block at slot "
+                + slot
+                + ", no proposer was prepared, and no default fee recipient defined, using 0x00 BURN address",
+            Color.RED));
+  }
+
   public void proposedBlockImportFailed(
       final String failureReason,
       final UInt64 slot,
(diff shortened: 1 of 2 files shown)


rolfyone:
looks like #5767 also fixes this so only raising draft.

ThreeMuskets12:
Probably makes more sense for this to be an error rather than a warning like I had it, closed #5767 so this can move forward from draft to open

rolfyone:
I was going to let someone else decide @ThreeMuskets12 but ok :) 
I think we were fixing it at the same time, thanks for the efforts anyway, we do always appreciate involvement from everyone :)

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/ValidatorLogger.java
```diff
@@ -184,6 +184,15 @@ public void executionPayloadPreparedUsingBeaconDefaultFeeRecipient(final UInt64
             Color.YELLOW));
   }
 
+  public void executionPayloadPreparedUsingBurnAddressForFeeRecipient(final UInt64 slot) {
+    log.error(
+        ColorConsolePrinter.print(
+            "Producing block at slot "
+                + slot
+                + ", no proposer was prepared, and no default fee recipient defined, using 0x00 BURN address",
```
ajsutton:
```suggestion
            "Producing block at slot "
                + slot
                + ", no proposer was prepared, and no default fee recipient defined, inclusion fees will be lost",
```

not sure how well people understand the idea of a burn address so probably slightly better to just be clear about what will be lost.

ajsutton (approved):
LGTM

PR #5769, opened 2022-06-11
Update copyright notice on all files

benjaminion:
Our copyright notice was outdated (ConsenSys AG), so I consulted legal who came back with this advice:

> Change for all files - "Copyright ConsenSys Software Inc., 2022"

Specifically they asked not to leave the previous dates, but to update to this year.

Sorry!


ajsutton (approved):
LGTM.

PR #5774, opened 2022-06-13
Cache empty participation flags array for faster resetting

### LINKED ISSUE #5448: Optimise resetting participation flag lists
### Description
EpochProcessorAltair needs to set all participation flags back to zero which is quite expensive. We should create an optimised way to do this reset.  #5349 is one such approach but needs some additional tidy up to get the details right.

rolfyone:
Will still need to fall back to the old `setAll` behaviour if the number of validators is below the current (older states for example that may have less validators)

```
before:
   Benchmark                                 (validatorsCount)   Mode  Cnt  Score   Error  Units
   EpochTransitionBenchmark.epochTransition             400000  thrpt   10  1.047 ± 0.070  ops/s

after:
   Benchmark                                 (validatorsCount)   Mode  Cnt  Score   Error  Units
   EpochTransitionBenchmark.epochTransition             400000  thrpt   10  1.402 ± 0.147  ops/s
```

partially addresses #5448

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/statetransition/epoch/EpochProcessorAltair.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/statetransition/epoch/EpochProcessorAltair.java
@@ -14,8 +14,8 @@
 package tech.pegasys.teku.spec.logic.versions.altair.statetransition.epoch;
 
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import java.util.Optional;
+import tech.pegasys.teku.infrastructure.ssz.SszList;
 import tech.pegasys.teku.infrastructure.ssz.SszMutableList;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszMutableUInt64List;
 import tech.pegasys.teku.infrastructure.ssz.primitive.SszByte;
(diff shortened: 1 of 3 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/statetransition/epoch/EpochProcessorAltair.java
```diff
@@ -146,4 +155,29 @@ public void processInactivityUpdates(
...
   }
+
+  private Optional<SszList<SszByte>> getZeroParticipationFlags(final BeaconStateAltair state) {
+    SszList<SszByte> flags = zeroParticipationFlags;
+    SszMutableList<SszByte> mutableFlags;
+    if (flags == null) {
+      mutableFlags = state.getCurrentEpochParticipation().getSchema().of().createWritableCopy();
+      mutableFlags.setAll(SszByte.ZERO, state.getValidators().size());
+      flags = mutableFlags.commitChanges();
+      zeroParticipationFlags = flags;
```
ajsutton, line 166: nit: Can move this line to just before the return and avoid the duplication. In fact both the flags and zeroParticipationFlags assignments could move down.

ajsutton (approved):
LGTM.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/statetransition/epoch/EpochProcessorAltair.java
rolfyone: id prefer not to, in this case we're only going mutable because we need to actually change the object, moving it outside cut the improvement in half because we're getting mutable then storing constantly...

rolfyone: so theres 3 cases, 
 - normal is we don't need to change, in which case we get the zeroParticipation and return it
 - edge: first creation of object, have to have mutable, have to commit
 - growing: need to get mutable, need to commit

PR #5776, opened 2022-06-13
Optimise creating validator statuses

ajsutton:
## PR Description
Enhanced for loop turns out to be faster than a stream (at least in this case) and avoid accessing fields from the Validator multiple times - while they are cached it incurs a cache lookup which adds up.

```
Before:
EpochTransitionBenchmark.createValidatorStatuses             400000  thrpt   10  11.028 ± 0.835  ops/s

After:
EpochTransitionBenchmark.createValidatorStatuses             400000  thrpt   10  15.271 ± 0.243  ops/s
```

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5780, opened 2022-06-14
Cache inactivity leak and finality delay in RewardsAndPenaltiesCalculator

rolfyone:
This value doesn't change so it can just be read once.

Doesn't really improve performance in our benchmark, but is cleaner.

## Documentation

## Changelog


ajsutton (approved):
LGTM

PR #5786, opened 2022-06-14
Fix payload re-execution

zilm13:
## PR Description
We have a bug in re-execution of payload, which prunes block with `FAILED_EXECUTION_PAYLOAD_EXECUTION` in the next slot instead of retrying to execute it in the next slots. Fixing it, test added.

Opened question, do we want to keep 2 slots retrying here?
My guess is that we will try to re-request these blocks via sync after StatusMessage exchange, which is often with, say, 30 peers connected. But it's not often on non-mainnet cases. With one peer connected we will wait for 5 minutes. Same time if it's bad payload which is killing EL, we don't want to continue retrying for a long.

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
```diff
@@ -92,7 +92,7 @@ protected SafeFuture<?> doStop() {
   public void onSlot(UInt64 slot) {
     super.onSlot(slot);
     pendingBlocksForEPReexecution.removeIf(
-        block -> block.getSlot().isLessThan(slot.plus(EXPIRES_IN_SLOT)));
+        block -> block.getSlot().isLessThan(slot.minusMinZero(EXPIRES_IN_SLOT)));
```
tbenr: we can save the zero check this way:
```suggestion
        block -> block.getSlot().plus(EXPIRES_IN_SLOT).isLessThan(slot);
```

tbenr (approved):
LGTM. good catch, blame on me:)
only one nit suggestion

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
zilm13: Looks much better, thank you

PR #5794, opened 2022-06-15
Use original init size when transferring cache

ajsutton:
## PR Description
Use the original cache size when resetting the values array after a transfer rather than switching back to the default size.

```
Before:
EpochTransitionBenchmark.epochTransition             400000  thrpt   10  2.447 ± 0.063  ops/s

After:
EpochTransitionBenchmark.epochTransition             400000  thrpt   10  2.684 ± 0.103  ops/s
```

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5809, opened 2022-06-16
Fix max child types check in SszUnion

jtraglia:
## PR Description

I believe this limit check is slightly off. Before, the check would limit the number of children to 126 (max index: 125). After, there can be 128 children (max index: 127) which is `MAX_SELECTOR`.

## Documentation

## Changelog


ajsutton (approved):
LGTM. Good spot.

PR #5811, opened 2022-06-16
Fix bug in GetAllBlocksAtSlot

courtneyeh:
Fixes bug where `IllegalArgumentException` would be thrown where we want to see the empty list of `BlockAndMetaData` to then return not found. This now ensures `GetAllBlocksAtSlot` handles the empty list before creating the `AllBlocksAtSlotData` object which throws the exception. This exception should never be thrown now as it is handled prior to the object creation.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5813, opened 2022-06-17
Migrate GetStateByBlockRoot to MigratingEndpointAdapter

### LINKED ISSUE #5803: Migrate GetStateByBlockRoot to MigratingEndpointAdapter

courtneyeh:
fixes #5803 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetStateByBlockRoot.java
```diff
@@ -79,26 +107,41 @@ public GetStateByBlockRoot(
...
-              ctx.header(
-                  "Content-Disposition", "filename=\"" + sszResponse.abbreviatedHash + ".ssz\"");
-              ctx.contentType("application/octet-stream");
-              return sszResponse.byteStream;
+
+              final BeaconState state = result.get();
+              request.header(
+                  "Content-Disposition",
+                  "filename=\"" + state.hashTreeRoot().toUnprefixedHexString() + ".ssz\"");
+              request.contentType("application/octet-stream");
```
rolfyone: this bit will be handled by framework i think...

rolfyone: should be v. similar to `debug/states/{state_id}`

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -241,10 +241,10 @@ private String selectContentType(
...
         ContentTypes.getResponseContentType(supportedTypes, acceptHeader)
             .orElse(defaultResponseContentType);
-    checkState(
-        supportedTypes.contains(selectedType),
-        "Response content type %s was selected but is not supported.",
-        selectedType);
+    if (!supportedTypes.contains(selectedType)) {
+      throw new IllegalArgumentException(
+          "Response content type " + selectedType + " was selected but is not supported.");
+    }
```
courtneyeh: @ajsutton Updated this with @rolfyone so that a bad request is given instead of an internal server error. Any opinions on this?

ajsutton: Hmm, I think this should be an internal server error unless I've misunderstood the context.  I think this occurs when we've selected a response type to return that we don't actually support (most commonly because the default response type isn't a supported type).  That's entirely a code error and not a bad request. If the user specifies an unsupported type in `Accept` header we should just fall back to the default.

We'd only give a bad request if we didn't support the _request_ content type as then we can't parse the request content.

rolfyone: we have an ssz request that only allows ssz, so requesting json we'd get a 500 back...

ajsutton: Then we should set the default to be application/octet-stream so we send SSZ regardless of what's requested.

ajsutton: We may not have exposed the concept of default content type much but I'm pretty sure it exists...

ajsutton: `EndpointMetadataBuilder.defaultResponseType(ContentTypes.OCTET_STREAM)` should do it.

rolfyone: potentially we need to override the default type so that in this case it can default to ssz if none is specified.. that'd be another option...

so @courtneyeh we can override default by 
```
.defaultResponseType("application/octet-stream")
```
but will still error 500 if the user requests application/json specifically

courtneyeh: Okay so its fine to get 500 if user requests application/json? I will add in that change.

ajsutton: If the user requests application/json and we don't support it we should return ssz (we shouldn't declare json as a supported type at all which would make that just work I think)

ajsutton: Same as for most endpoints if the user specifically sends `Accept: made-up/crap` we will respond with JSON.

rolfyone (approved):
LGTM

PR #5819, opened 2022-06-20
Added deserializer for ExecutionPayload

rolfyone:
This will be used for block separation for bellatrix.

Partially addresses #5312.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
@@ -53,6 +53,7 @@ import tech.pegasys.teku.spec.datastructures.blocks.Eth1Data;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlockUnblinder;
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.BeaconBlockBodyBuilder;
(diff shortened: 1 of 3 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -262,12 +263,26 @@ public BeaconState deserializeBeaconState(final Bytes serializedState) {
...
         .getSignedBeaconBlockSchema()
-        .sszDeserialize(serializedState);
+        .sszDeserialize(serializedBlock);
+  }
+
+  public ExecutionPayload deserializeExecutionPayload(final Bytes serializedPayload) {
+    final SpecVersion specVersion = forMilestone(SpecMilestone.BELLATRIX);
+    if (specVersion == null) {
+      throw new IllegalStateException("Bellatrix must be scheduled to decode execution payloads");
+    }
```
rolfyone, line 278: This is possible, just putting it out there that it's not 100% ideal if ExecutionPayload changes on a hard fork...

ajsutton: I think we can count on ExecutionPayload changing and need to design for it now, so can't just hard code the milestone to use.

Either need to lookup the milestone via the block's slot or store the milestone with the serialised payload. Ultimately we'll want to retrieve the payload from the EL rather than the DB so will have to have the block info available to do that.

rolfyone:
closing, will go in a different direction from this...

PR #5822, opened 2022-06-21
Use head state for more attestation validations

ajsutton:
## PR Description
Adjust `AttestationStateSelector` to allow using the head state for attestations that are from prior epoch. We can still correctly determine the committees needed to validate the attestation using the randao mix history even if the state is from a later epoch than the attestation.

The head state is guaranteed to be in memory, should give the best cache hit rate and any committees we have to calculate will be added to it's state and then copied when any blocks build on that state making it less likely we'll need to calculate them again.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5825, opened 2022-06-21
Enable Error Prone's PackageLocation check

jtraglia:
## PR Description

According to [the docs for this check](http://errorprone.info/bugpattern/PackageLocation):

> Java files should be located in a directory that ends with the fully qualified name of the package.

It found one issue in with the internal tools command file.

Also cleaned up the Error Prone configuration in build.gradle:
* Moved some experimental checks to the group of other experimental checks.
* Capitalized first word in one of the comments.
* Removed punctuation from some comments, none of the other had it.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5832, opened 2022-06-23
Migrate GetProtoArray to MigratingEndpointAdapter

### LINKED ISSUE #5804: Migrate GetProtoArray to MigratingEndpointAdapter

courtneyeh:
fixes #5804 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetProtoArray.java
```diff
@@ -64,10 +83,33 @@ public GetProtoArray(final ChainDataProvider chainDataProvider, final JsonProvid
...
+        chainDataProvider.getProtoArrayData().stream()
+            .map(this::mapObjectValuesToString)
+            .collect(Collectors.toList());
+    request.respondOk(data);
+  }
 
-    final GetProtoArrayResponse response =
-        new GetProtoArrayResponse(chainDataProvider.getProtoArrayData());
-    ctx.result(jsonProvider.objectToJSON(response));
+  // TODO check this correct approach to making mapping type work
```
rolfyone: I'd be super tempted to make ProtoNode::getData return a Map<String,String>... it's only used for this api...

ajsutton: I'd agree with that - makes sense to just return what we need to serialise.  Previously returning typed objects worked best but if string makes things easier that's fine.  It's just debug data so doesn't matter if the exact format of the output changes.

ajsutton (approved):
LGTM.

PR #5833, opened 2022-06-23
Iterate statuses instead of validators list in processEffectiveBalanceUpdates

ajsutton:
## PR Description
Iterating the simple list of statuses is significantly faster than the SSZList of validators so take advantage that, then only retrieve the actual `Validator` in the rare case where effective balance is actually changing.

```
Before
EpochTransitionBenchmark.epochTransition             400000  thrpt   10  2.836 ± 0.096  ops/s

After:
EpochTransitionBenchmark.epochTransition             400000  thrpt   10  3.157 ± 0.183  ops/s
```

Had to update the epoch processor test so switched it over to using the `TestSpecContext` approach to automatically run on all milestones.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5835, opened 2022-06-23
Added storage for blinded blocks and execution payloads.

rolfyone:
partially addresses #5312

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreAccessor.java
```diff
@@ -29,6 +29,8 @@ public interface KvStoreAccessor extends AutoCloseable {
 
   <K, V> Optional<V> get(KvStoreColumn<K, V> column, K key);
 
+  long size(final KvStoreColumn<?, ?> column);
```
ajsutton: nit: No need for `final` on interface method declarations.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/schema/V6SchemaCombinedTreeState.java
```diff
@@ -111,6 +121,21 @@ public KvStoreColumn<UInt64, Set<Bytes32>> getColumnNonCanonicalRootsBySlot() {
...
+  }
+
+  @Override
+  public KvStoreColumn<Bytes32, Bytes> getColumnExecutionPayloadByPayloadHash() {
+    return executionPayloadByBlockRoot;
+  }
+
+  @Override
+  public KvStoreColumn<UInt64, Bytes32> getColumnFinalizedBlockRootBySlot() {
+    return null;
```
ajsutton: This doesn't look right.

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/InMemoryV4KvStoreBlindedBlocksDatabaseTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.io.File;
+import tech.pegasys.teku.storage.server.DatabaseVersion;
+import tech.pegasys.teku.storage.server.StateStorageMode;
+import tech.pegasys.teku.storage.storageSystem.InMemoryStorageSystemBuilder;
+import tech.pegasys.teku.storage.storageSystem.StorageSystem;
+import tech.pegasys.teku.storage.store.StoreConfig;
+
+public class InMemoryV4KvStoreBlindedBlocksDatabaseTest
+    extends AbstractKvStoreDatabaseWithHotStatesTest {
```
ajsutton: Don't need to solve it in this PR, but we probably need to find a way to rework these db tests so that we can create variants better - we'll need all of the existing tests with and without blinded blocks since the change applies to older formats as well.

ajsutton (commented):
I'll need to work out a good way to review the actual database class changes which are fairly complex and the most likely place we'll break something in the unblinded format which would be bad. Have reviewed the rest though so figured I'd drop these comments now.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/schema/V6SchemaCombinedTreeState.java
rolfyone: definitely wrong... so many permutations of the same code :/

ajsutton (approved):
Once we get this merged with the new DB tests I think we're fine to merge.

PR #5838, opened 2022-06-23
Don't revalidate IndexedAttestation when attestation is coming from a reorged block

### LINKED ISSUE #5478: Some attestations being re-validated with SIMPLE BLS verifier
### Description
Normally the gossip validation performs the `isValidIndexedAttestationCheck` which is cached in `ValidatableAttestation` so we don't have to recheck it when processing the attestation.  If we do wind up rechecking it, currently we use the simple BLS signature verifier which does not do any batching.

There seem to be some attestations slipping through which aren't yet marked as valid indexed attestations and wind up actually doing signature verification with this simple approach.  I'm not yet sure what flow leads to these.

I added a check in `AttestationUtil.isValidIndexedAttestationAsync` to print a stack trace if simple verification was used and got:
```
java.lang.Exception: Using simple BLS verification
	at tech.pegasys.teku.spec.logic.common.util.AttestationUtil.isValidIndexedAttestationAsync(AttestationUtil.java:169)
	at tech.pegasys.teku.spec.logic.common.util.AttestationUtil.isValidIndexedAttestation(AttestationUtil.java:139)
	at tech.pegasys.teku.spec.logic.common.util.ForkChoiceUtil.lambda$validate$1(ForkChoiceUtil.java:232)
	at tech.pegasys.teku.spec.datastructures.util.AttestationProcessingResult.ifSuccessful(AttestationProcessingResult.java:47)
	at tech.pegasys.teku.spec.logic.common.util.ForkChoiceUtil.validate(ForkChoiceUtil.java:227)
	at tech.pegasys.teku.spec.Spec.validateAttestation(Spec.java:477)
	at tech.pegasys.teku.statetransition.forkchoice.ForkChoice.lambda$onAttestation$18(ForkChoice.java:507)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309)
	at tech.pegasys.teku.infrastructure.async.SafeFuture.thenCompose(SafeFuture.java:470)
	at tech.pegasys.teku.statetransition.forkchoice.ForkChoice.onAttestation(ForkChoice.java:503)
	at tech.pegasys.teku.statetransition.attestation.AttestationManager.onAttestation(AttestationManager.java:196)
	at tech.pegasys.teku.statetransition.attestation.AttestationManager.lambda$processInternallyValidatedAttestation$5(AttestationManager.java:142)
	at java.base/java.util.concurrent.CompletableFuture.uniAcceptNow(CompletableFuture.java:757)
	at java.base/java.util.concurrent.CompletableFuture.uniAcceptStage(CompletableFuture.java:735)
	at java.base/java.util.concurrent.CompletableFuture.thenAccept(CompletableFuture.java:2182)
	at tech.pegasys.teku.infrastructure.async.SafeFuture.thenAccept(SafeFuture.java:451)
	at tech.pegasys.teku.statetransition.attestation.AttestationManager.processInternallyValidatedAttestation(AttestationManager.java:138)
	at tech.pegasys.teku.statetransition.attestation.AttestationManager.addAttestation(AttestationManager.java:124)
	at tech.pegasys.teku.networking.eth2.gossip.topics.topichandlers.SingleAttestationTopicHandler.lambda$createHandler$0(SingleAttestationTopicHandler.java:40)
	at tech.pegasys.teku.networking.eth2.gossip.topics.topichandlers.Eth2TopicHandler.lambda$handleMessage$2(Eth2TopicHandler.java:101)
	at tech.pegasys.teku.infrastructure.async.SafeFuture.of(SafeFuture.java:73)
	at tech.pegasys.teku.infrastructure.async.ScheduledExecutorAsyncRunner.lambda$createRunnableForAction$1(ScheduledExecutorAsyncRunner.java:119)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

This is quite rare so not a significant contributor to CPU usage (and the resulting behaviour is correct and entirely safe - just potentially less efficient than it could be).  We need to investigate why this occurs and work out if we can safely use the batching signature verification for these as well.

StefanBratanov:
## PR Description

- Added `signatureValidated` in `ValidateableAttestation` object
- Marked the attestation as coming from reorged block in `OperationsReOrgManager`
- Added an ability to skip signature check by passing a flag in `AttestationUtil`
- Removed unused method from `AttestationUtil`
- Added bunch of `final`

## Fixed Issue(s)
fixes #5478 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
@@ -64,7 +64,7 @@ public class ValidateableAttestation {
 
   public static ValidateableAttestation fromReorgedBlock(
       final Spec spec, final Attestation attestation) {
-    ValidateableAttestation validateableAttestation =
+    final ValidateableAttestation validateableAttestation =
         new ValidateableAttestation(
             spec, attestation, Optional.empty(), OptionalInt.empty(), false);
     // An indexed attestation from a reorged block is valid because it already
@@ -140,7 +140,7 @@ public class ValidateableAttestation {
       return;
     }
 
-    Bytes32 committeeShufflingSeed =
+    final Bytes32 committeeShufflingSeed =
         spec.getSeed(
             state,
             spec.computeEpochAtSlot(attestation.getData().getSlot()),
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
@@ -38,6 +38,7 @@ public class ValidateableAttestation {
   private final Supplier<Bytes32> hashTreeRoot;
   private final AtomicBoolean gossiped = new AtomicBoolean(false);
   private final boolean producedLocally;
+  private final boolean producedFromNonCanonicalBlock;
   private final OptionalInt receivedSubnetId;
 
   private volatile boolean isValidIndexedAttestation = false;
@@ -47,18 +48,24 @@ public class ValidateableAttestation {
 
   public static ValidateableAttestation from(final Spec spec, Attestation attestation) {
     return new ValidateableAttestation(
-        spec, attestation, Optional.empty(), OptionalInt.empty(), false);
+        spec, attestation, Optional.empty(), OptionalInt.empty(), false, false);
   }
 
   public static ValidateableAttestation fromValidator(final Spec spec, Attestation attestation) {
     return new ValidateableAttestation(
-        spec, attestation, Optional.empty(), OptionalInt.empty(), true);
+        spec, attestation, Optional.empty(), OptionalInt.empty(), true, false);
   }
 
   public static ValidateableAttestation fromNetwork(
       final Spec spec, Attestation attestation, int receivedSubnetId) {
     return new ValidateableAttestation(
-        spec, attestation, Optional.empty(), OptionalInt.of(receivedSubnetId), false);
+        spec, attestation, Optional.empty(), OptionalInt.of(receivedSubnetId), false, false);
+  }
+
+  public static ValidateableAttestation fromNonCanonicalBlock(
+      final Spec spec, Attestation attestation) {
+    return new ValidateableAttestation(
+        spec, attestation, Optional.empty(), OptionalInt.empty(), false, true);
   }
 
   public static ValidateableAttestation aggregateFromValidator(
@@ -68,7 +75,8 @@ public class ValidateableAttestation {
         attestation.getMessage().getAggregate(),
         Optional.of(attestation),
         OptionalInt.empty(),
-        true);
+        true,
+        false);
   }
 
   public static ValidateableAttestation aggregateFromNetwork(
@@ -78,25 +86,32 @@ public class ValidateableAttestation {
         attestation.getMessage().getAggregate(),
         Optional.of(attestation),
         OptionalInt.empty(),
+        false,
         false);
   }
 
   private ValidateableAttestation(
       final Spec spec,
-      Attestation attestation,
-      Optional<SignedAggregateAndProof> aggregateAndProof,
-      OptionalInt receivedSubnetId,
-      boolean producedLocally) {
+      final Attestation attestation,
+      final Optional<SignedAggregateAndProof> aggregateAndProof,
+      final OptionalInt receivedSubnetId,
+      final boolean producedLocally,
+      final boolean producedFromNonCanonicalBlock) {
     this.spec = spec;
     this.maybeAggregate = aggregateAndProof;
     this.attestation = attestation;
     this.receivedSubnetId = receivedSubnetId;
     this.hashTreeRoot = Suppliers.memoize(attestation::hashTreeRoot);
     this.producedLocally = producedLocally;
+    this.producedFromNonCanonicalBlock = producedFromNonCanonicalBlock;
   }
 
   public boolean isProducedLocally() {
     return producedLocally;
+  }
+
+  public boolean isProducedFromNonCanonicalBlock() {
+    return producedFromNonCanonicalBlock;
   }
 
   public boolean isValidIndexedAttestation() {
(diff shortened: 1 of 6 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
```diff
@@ -38,6 +38,7 @@ public class ValidateableAttestation {
   private final Supplier<Bytes32> hashTreeRoot;
   private final AtomicBoolean gossiped = new AtomicBoolean(false);
   private final boolean producedLocally;
+  private final boolean producedFromNonCanonicalBlock;
```
rolfyone: something like `alreadyValidated` or `previouslyValidated` or similar - looking for 'what' the context is, less than 'why' (from a non canonical block)

This will allow us to re-use it, should the need arise...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/AttestationUtil.java
```diff
@@ -167,7 +167,15 @@ public SafeFuture<AttestationProcessingResult> isValidIndexedAttestationAsync(
...
             })
-        .thenCompose(att -> isValidIndexedAttestationAsync(fork, state, att, blsSignatureVerifier))
+        .thenCompose(
+            att -> {
+              // An attestation from a non-canonical block does not require signature verification
+              // because its signature already has been verified when the block was part of the
+              // canonical chain
+              boolean skipSignatureVerification = attestation.isProducedFromNonCanonicalBlock();
+              return isValidIndexedAttestationAsync(
+                  fork, state, att, blsSignatureVerifier, skipSignatureVerification);
```
rolfyone: a good variable name will mean we don't need a big explanation here :)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/OperationsReOrgManager.java
```diff
@@ -123,23 +123,24 @@ private void processNonCanonicalBlockAttestations(
...
-                      LOG.error(
-                          "Failed to process re-queued attestation from block: {}",
-                          blockRoot,
-                          err));
-        });
+        attestation ->
+            attestationManager
+                .onAttestation(
+                    ValidateableAttestation.fromNonCanonicalBlock(
+                        recentChainData.getSpec(), attestation))
```
rolfyone: this could become something like `ValidatableAttestation.fromVerifiedAttestation` or something...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidateableAttestation.java
StefanBratanov: Yeah definitely agree. Will change it across, which will cover the other points as well.

ajsutton: I'd say `signatureValidated`.  The validation is in a particular context - it's still important to check that we have the same `committeeShufflingSeed` etc or the attestation may wind up invalid when included in a different block.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/OperationsReOrgManager.java
ajsutton: Personally I'd do something like `fromReorgedBlock` (or maybe `fromValidatedBlock`). Being clearer about the source makes it easier to know which properties should be set (especially if we add new properties to `ValidatableAttestation` in the future.  So attestations from a reorged block have the `signatureValidated` property set.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/AttestationUtil.java
```diff
@@ -151,27 +154,33 @@ public AttestationProcessingResult isValidIndexedAttestation(
...
+              final boolean skipSignatureValidation = attestation.isSignatureValidated();
+              return isValidIndexedAttestationAsync(
+                  fork, state, att, blsSignatureVerifier, skipSignatureValidation);
+            })
         .thenApply(
             result -> {
               if (result.isSuccessful()) {
                 attestation.saveCommitteeShufflingSeed(state);
+                attestation.setValidSignature();
                 attestation.setValidIndexedAttestation();
```
ajsutton: I think I'd just have the `setValidIndexedAttestation` also mark the signature as valid because part of being a valid indexed attestation is that the signature is valid.

tbenr (approved):
LGTM.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/AttestationUtil.java
```diff
@@ -151,27 +154,33 @@ public AttestationProcessingResult isValidIndexedAttestation(
...
-      BeaconState state,
-      ValidateableAttestation attestation,
-      AsyncBLSSignatureVerifier blsSignatureVerifier) {
+      final Fork fork,
+      final BeaconState state,
+      final ValidateableAttestation attestation,
+      final AsyncBLSSignatureVerifier blsSignatureVerifier) {
     if (attestation.isValidIndexedAttestation()) {
       return completedFuture(AttestationProcessingResult.SUCCESSFUL);
     }
```
ajsutton, line 164: We can actually go a step further than what this PR currently does, because we know the attestation is a valid indexed attestation (ie _all_ the checks in `isValidIndexedAttestationAsync` are already known to pass) we could just have attestations from re-orged blocks be immediately marked with `setValidIndexedAttestation`. However, we'd need to change this bit of code to check that the `ValidateableAttestation` actually has an indexed attestation and set it and the committee shuffling seed if not.  Basically we'd only immediately return here if `isValidIndexedAttestation && getIndexedAttestation().isPresent()`, otherwise we'd run through the code below but entirely skip the isValidIndexedAttestationAsync` call.

It probably doesn't make much difference but probably worth a follow up PR.

ajsutton (approved):
LGTM.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/AttestationUtil.java
StefanBratanov: Good idea, let me do that

PR #5840, opened 2022-06-23
Re-execute failed execution block if descendant is added to the pendi…

### LINKED ISSUE #5824: Possible sync stall on execution layer outage
### Description
Say, we are syncing on gossip and suddenly our execution layer goes down. We have `ReexecutingExecutionPayloadBlockManager`, which will try re-execution for next 2 slots (and we don't want to increase this number, because it could be a bad payload forcing our EL to fail). After that we will drop this block and get following chain:
    SUCCESS <- SUCCESS <-[nothing]<- UNKNOWN_PARENT <- UNKNOWN_PARENT
We are unable to pass this issue on gossip, we'll return to normal sync after `StatusMessage` exchange which is once in 5 minutes with every node. We have a lot of peers on mainnet and some `StatusMessage` always arrives in a few seconds, but in private networks or tests it could be a real problem stalling sync for up to 5 minutes.

As a solution we may need to adjust our gossip handling to ensure it retries execution when we get a new descendant of block with failed execution verification (`FAILED_EXECUTION_PAYLOAD_EXECUTION`).
### Steps to Reproduce (Bug)
Shutdown execution layer after merge when on gossip sync. Start it after more than 2 slots.

zilm13:
…ng pool

## PR Description
Re-execute failed execution block even after 2 slots if descendant is added to the pending blocks.
- I'm not sure on `EXPIRES_IN_SLOTS` constant, looks like a magic number, maybe we could use other criteria to drop it
- Not focused on tailored structures, shouldn't run very often, exiting early when no action is needed
- Not sure that stack overflow protection really needed in pool, a bit paranoid on recursions

## Fixed Issue(s)

Fixes #5824 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
@@ -13,16 +13,22 @@
 
 package tech.pegasys.teku.statetransition.block;
 
+import static tech.pegasys.teku.statetransition.util.PendingPoolFactory.BLOCK_HASH_TREE_ROOT_FUNCTION;
+
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.Cancellable;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
-import tech.pegasys.teku.infrastructure.collections.LimitedSet;
+import tech.pegasys.teku.infrastructure.collections.LimitedMap;
 import tech.pegasys.teku.infrastructure.logging.EventLogger;
 import tech.pegasys.teku.infrastructure.time.TimeProvider;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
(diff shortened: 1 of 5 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
```diff
@@ -65,6 +74,8 @@ public ReexecutingExecutionPayloadBlockManager(
         eventLogger,
         blockImportMetrics);
     this.asyncRunner = asyncRunner;
+    this.pendingBlocks = pendingBlocks;
+    pendingBlocks.subscribeRequiredBlockRoot(this::requireAncestor);
```
ajsutton, line 78: If we're getting notifications that the block is required, why isn't the `RecentBlockFetcherService` fetching the block and executing it via `BlockManager`?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/ReexecutingExecutionPayloadBlockManager.java
```diff
@@ -101,16 +116,32 @@ public SafeFuture<BlockImportResult> importBlock(SignedBeaconBlock block) {
...
+      return;
+    }
+    final Bytes32 parentForFirstPending = pendingBlocks.getFirstRequiredAncestor(root);
+    final Optional<SignedBeaconBlock> reExecuteBlock =
+        pendingBlocksForEPReexecution.keySet().stream()
+            .filter(
+                block -> BLOCK_HASH_TREE_ROOT_FUNCTION.apply(block).equals(parentForFirstPending))
+            .findFirst();
+    reExecuteBlock.ifPresent(
+        block -> pendingBlocksForEPReexecution.put(block, currentSlot.plus(RETRY_SLOTS)));
```
ajsutton, line 135: This seems to assume that we already have the block and are just extending the expiry when we receive new blocks.  What happens if the block is dropped from `pendingBlocksForEPReexecution` before we receive one of its ancestors? It seems like we'd get stuck again.

ajsutton (commented):
I'm not entirely sure I'm fully understanding the approach being used here - seems to be quite a lot of complexity involved.

I suspect that when blocks are dropped from `ReexecutingExecutionPayloadBlockManager` we need to remove all their descendants from `PendingBlocks`, same as we do when a block is invalid.  Then if we get another block from that chain, we'll refetch the blocks.

And I think you may have been right along that we need to keep retrying execution for longer - particularly since we'll potentially have to refetch a bunch of descendants if we drop a block. My concern is just that if we have a non-canonical fork which is causing errors from the EL, then we don't want to keep spamming it with re-execution attempts all the time or it may hinder our ability to keep following the canonical chain because we're effectively DOS'ing the EL. That shouldn't happen but we've had issues in the past where repeatedly retrying things causes further issues for the EL so I am nervous about it.  Even with this approach we're essentially just extending the time we retry as long as we get further ancestors though so it's not much different (and this approach is more complex).

PR #5843, opened 2022-06-24
Include SyncCommitteeMessage in Contribution only on received subnet

zilm13:
## PR Description
From the [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/validator.md#construct-sync-committee-contribution), what should we do to create Contribution:
> Collect all of the (valid) `sync_committee_messages: Set[SyncCommitteeMessage]` from the `sync_committee_{subnet_id}` gossip during the selected `slot` with an equivalent `beacon_block_root` to that of the aggregator.

This means that we should count only `SyncCommitteeMessage` as contribution if it was received on particular `subnet_id`, which sounds reasonable, as if one is listening to only one `subnet_id` and message should be distributed to several subnets, he will see it only if it appears on the `subnet_id` he is listening to.
Also this fix clears noisy exception, when we were trying to add same messages to contribution which were received on different subnets, which you could see in `SyncCommitteeGossipAcceptanceTest`. Test `SyncCommitteeMessagePoolTest.shouldCreateCorrespondingContributionsForEqualMessagesFromDifferentSubnets` clarifies that Exception is not thrown anymore.

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/synccommittee/SyncCommitteeMessagePool.java
```diff
@@ -92,18 +92,21 @@ private synchronized void doAdd(final ValidateableSyncCommitteeMessage message)
...
-            subcommitteeIndex ->
-                blockRootAndCommitteeIndexToMessages
-                    .computeIfAbsent(
-                        new BlockRootAndCommitteeIndex(
-                            message.getBeaconBlockRoot(), subcommitteeIndex),
-                        __ -> new ContributionData())
-                    .add(
-                        assignments.getParticipationBitIndices(subcommitteeIndex),
-                        message.getMessage().getSignature()));
+    IntSet applicableSubnets;
```
tbenr: This is where the usage of `final` makes more sense :)

zilm13: Ashes on my head, thank you!

ajsutton (approved):
LGTM.

PR #5844, opened 2022-06-24
Engine API Client metrics and refactoring

tbenr:
- Moves the MetricRecording logic from `MetricRecordingExecutionBuilderClient` to a new abstract class `MetricRecordingAbstractClient`
- Simplify the unit testing of the logic, we don't need any more a complex test for each metric recording subclass. I decided to drop them completely because the only important thing to test would be to check that the delegate is actually called.
- In `ExecutionLayerManagerImpl` improves logging when we fallback to local EL and apply some refactoring to improve readability.

for `fcU` call, I decided to map the call to two different methods (with and without attributes) so we can have more control on the call that initiate a block production on the local EL.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5852, opened 2022-06-28
Make distributions reproducible

### LINKED ISSUE #5718: Make distribution artefacts fully reproducible
### Description
Update the build process so that distribution artefacts created from the same git commit will have exactly the same contents (ie same sha256 checksum for the .tar.gz and .zip files generated).

Most of the Teku build process is already deterministic - the only differences currently are:

1. The license plugin includes the timestamp the report was generated at. This needs to be removed, likely be overriding the template.
2. The archives need to use reproducible file order (can set an option in gradle)
3. The archives include the file timestamps which are set based on the current system time.

2 and 3 can both be solved with gradle options:
```
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: build.gradle
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ build.gradle:143 @ allprojects {
    }
  }

  tasks.withType(Jar).configureEach {
    Jar jar ->
      jar.preserveFileTimestamps = false
      jar.reproducibleFileOrder = true
  }

  tasks.withType(JavaCompile) {
    options.compilerArgs += [
      '-Xlint:unchecked',
@ build.gradle:392 @ distTar {
  }
  compression = Compression.GZIP
  archiveExtension = 'tar.gz'
  preserveFileTimestamps = false
  reproducibleFileOrder = true
}

distZip {
@ build.gradle:401 @ distZip {
  doFirst {
    delete fileTree(dir: 'build/distributions', include: '*.zip')
  }
  preserveFileTimestamps = false
  reproducibleFileOrder = true
}
```

However, disabling `preserveFileTimestamps` means the timestamps are committed and all files wind up with a timestamp of 1 Feb 1980 (at least with Mac unzip) which will potentially cause issues.  A better approach may be to have a task iterate through all files to be included in the archive and set their modification time to the timestamp of the git commit being built, and leave `preserveFileTimestamps` enabled.

jtraglia:
## PR Description

I like the idea of having reproducible builds. This PR will make it so.

## Fixed Issue(s)

Fixes: #5718

## Documentation

## Changelog


scripts/override_dates_in_dist.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+set -euo pipefail
```
ajsutton: This won't work when building on Windows.  If we can achieve the same thing in Java or groovy it would automatically be crossplatform. Downside is it's probably more verbose.

scripts/override_dates_in_dist.sh
```diff
@@ -0,0 +1,51 @@
...
+find . -exec touch -d $LAST_COMMIT_DATETIME {} \;
+
+#
+# Repack the distribution.
+#
+if [[ "$DIST_FILE_NAME" == *.zip ]]; then
+  zip -q -r $DIST_FILE_NAME $(ls .)
+else
+  tar czf $DIST_FILE_NAME $(ls .)
+fi
```
ajsutton: Wouldn't this make the file order potentially non-deterministic again?

ajsutton (commented):
ooo, very exciting. I'd definitely love to see this working.

scripts/override_dates_in_dist.sh
ajsutton: A .bat version would be an alternative as well but would mean we're maintaining two versions of the script.

And it's probably also a big win to just have reproducible builds on UNIX-like systems.

jtraglia: Hmm I didn't think of Windows. I will look into making it cross platform somehow.

scripts/override_dates_in_dist.sh
jtraglia: Ah you're right. I'll need to fix that. Good spot.

scripts/override_dates_in_dist.sh
```diff
@@ -0,0 +1,51 @@
...
+set -euo pipefail
+
+DIST_FILE_NAME=$(basename $1)
+DIST_FILE_ABSPATH=$(readlink -f $1)
+
+#
+# The timestamp of the last commit. We need to remove the last part of the
+# timestamp because the 'touch' command doesn't work with it.
+#
+LAST_COMMIT_DATETIME=$(git log -1 --format=%cI | cut -d'+' -f1)
```
jtraglia: I will need to convert this time to UTC instead of stripping the `+XX:XX` or `-XX:XX` from the end.

ajsutton: This is definitely starting to sound like writing it in Java may be easier.  We can write libraries in a `buildSrc` dir and gradle will compile it and make it available to the build process which makes life simpler.  I'd have to look up the exact details again but it's pretty straight forward, just a matter of getting the right directory structure.

jtraglia:
I will attempt to re-implement this in Java tomorrow morning. It would be nice for this to work cross-platform. I believe there are classes for dealing with zip/tar/gzip files so that will make things easier. Also, the I imagine that the team is most comfortable with Java, so that will make maintenance easier.

jtraglia:
Doing this in Java is definitely the right approach. Just pushed the fix for `distZip`, I will handle `distTar` next.

By the way, I really wanted to use `ZipFileSystem` but I ran into some problems.

```java
FileTime now = FileTime.fromMillis(System.currentTimeMillis());
try (FileSystem fs = FileSystems.newFileSystem(zippath, null)) {
  for (Path path : fs.getRootDirectories()) {
    try (Stream<Path> fileStream = Files.walk(path)) {
      fileStream.forEach(entry -> {
        try {
          Files.setLastModifiedTime(entry, now);
        } catch (IOException e) {
          throw new RuntimeException(e);
        }
      });
    }
  }
}
```

No matter what I did, it would throw a `NullPointerException` when modifying anything in the filesystem. It could read attributes just fine. I also tested with a `{"create": "true"}` environment but it didn't fix it.

On JDK 11, the error message is really vague.
```
java.lang.NullPointerException
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.removeFromTree(ZipFileSystem.java:2816)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.update(ZipFileSystem.java:1229)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.setTimes(ZipFileSystem.java:396)
        at jdk.zipfs/jdk.nio.zipfs.ZipPath.setTimes(ZipPath.java:768)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileAttributeView.setTimes(ZipFileAttributeView.java:98)
        at java.base/java.nio.file.Files.setLastModifiedTime(Files.java:2355)
        at tech.pegasys.teku.depcheck.UpdateModifyDates.lambda$setLastModifiedTime$1(UpdateModifyDates.java:62)
```

With JDK 17, the error message is a little better:
```
java.lang.NullPointerException: Cannot invoke "jdk.nio.zipfs.ZipFileSystem$IndexNode.equals(Object)" because "child" is null
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.removeFromTree(ZipFileSystem.java:1378)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.update(ZipFileSystem.java:1702)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.setTimes(ZipFileSystem.java:589)
        at jdk.zipfs/jdk.nio.zipfs.ZipPath.setTimes(ZipPath.java:808)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileAttributeView.setTimes(ZipFileAttributeView.java:83)
        at java.base/java.nio.file.Files.setLastModifiedTime(Files.java:2442)
        at tech.pegasys.teku.depcheck.UpdateModifyDates.lambda$setLastModifiedTime$1(UpdateModifyDates.java:64)
```

I'm not really sure what the problem is. I tested the other JDK just to make sure it wasn't a JDK problem. This method should not throw a null pointer exception. To make this work, I had to unzip the distribution and re-zip it back up. It would have been much cleaner the code above worked.

build.gradle
```diff
@@ -861,14 +865,14 @@ dependencies {
...
+      from("LICENSE")
       from("build/reports/licenses") {
-        into "./licenses"
+        into "licenses"
         exclude "**/dependencies-without-allowed-license.json"
         exclude "**/project-licenses-for-check-license-task.json"
       }
       from("libs") { into "native" }
-      from("build/teku.autocomplete.sh") { into "." }
+      from("build/teku.autocomplete.sh")
```
jtraglia, line 884: When trying to get this working with `ZipFileSystem` it complained about these relative paths, like:
* `teku-develop/./LICENSE`
* `teku-develop/./licenses/`

jtraglia:
Alright I've done the same thing for tar files. Now it needs some serious clean up & deduplication.

jtraglia:
Yeah that method of removing the timestamp is wrong. Lemme fix that.

jtraglia:
So those tests are failing because the binaries are losing their execute permission during repackaging. And the `COPY_ATTRIBUTES` option appears to be unsupported for zip/tar files. Fun.

```
Caused by: java.lang.UnsupportedOperationException: COPY_ATTRIBUTES not supported
        at tech.pegasys.teku.repackage.Repackage.repackageTarGz(Repackage.java:153)
        at tech.pegasys.teku.repackage.Repackage.repackage(Repackage.java:64)
```

build.gradle
```diff
@@ -881,6 +888,11 @@ def buildTime() {
   return df.format(new Date())
 }
 
+// Get the date of the last git commit.
+def lastCommitDate() {
+  return grgit.log(maxCommits: 1).get(0).date
+}
```
ajsutton: Probably need to check if grgit is available and fall back to using the current time if not.  There are always a few people who build from the downloaded source distribution rather than a git checkout.

buildSrc/src/main/java/tech/pegasys/teku/repackage/Repackage.java
```diff
@@ -0,0 +1,200 @@
...
+          .forEachRemaining(
+              entry -> {
+                try {
+                  Path path = tempDir.resolve(entry.getName());
+                  if (entry.isDirectory()) {
+                    Files.createDirectories(path);
+                  } else {
+                    Files.createDirectories(path.getParent());
+                    Files.copy(zf.getInputStream(entry), path);
+                  }
```
ajsutton: I could probably write each file directly to the new archive rather than expanding to a file. 

I think the code below should work (and seems to preserve executable status at least on my Mac):

```java
  private static void repackageZip(Path zipDist, FileTime fileTime, Path tempDir)
      throws IOException {
    Path newZipDist = tempDir.resolve(zipDist.getFileName());
    try (final ZipFile source = new ZipFile(zipDist.toFile());
        final ZipOutputStream output = new ZipOutputStream(Files.newOutputStream(newZipDist))) {
      final Enumeration<? extends ZipEntry> sourceEntries = source.entries();
      while (sourceEntries.hasMoreElements()) {
        final ZipEntry sourceEntry = sourceEntries.nextElement();
        final ZipEntry outputEntry = new ZipEntry(sourceEntry);
        outputEntry.setLastModifiedTime(fileTime);
        outputEntry.setLastAccessTime(fileTime);
        output.putNextEntry(outputEntry);
        IOUtils.copy(source.getInputStream(sourceEntry), output);
      }
    }
  }

  private static void repackageTarGz(Path tarDist, FileTime fileTime, Path tempDir)
      throws IOException {
    Path newTarDist = tempDir.resolve(tarDist.getFileName());
    try (final TarArchiveInputStream source =
            new TarArchiveInputStream(
                new GzipCompressorInputStream(Files.newInputStream(tarDist)));
        final TarArchiveOutputStream target =
            new TarArchiveOutputStream(
                new GzipCompressorOutputStream(Files.newOutputStream(newTarDist)))) {
      ArchiveEntry entry;
      while ((entry = source.getNextEntry()) != null) {
        final TarArchiveEntry tarEntry = (TarArchiveEntry) entry;
        tarEntry.setModTime(fileTime);
        target.putArchiveEntry(tarEntry);
        IOUtils.copy(source, target);
        target.closeArchiveEntry();
      }
    }
  }
```

jtraglia: Yeah wow that looks/works way better. Thank you.

jtraglia: I did verify that it works as expected too. In retrospect, this makes sense. I didn't realize it was possible to modify it in place like that. I should have taken a step back to think about it.

ajsutton: I didn't know either - I got very much nerd-sniped on this one. :)

jtraglia:
Alright, I think this is in a pretty good state now. Though, I would still like to verify this works on other operating systems, notably Windows, before merging it in. I can test this out tomorrow.

jtraglia:
There's one more problem to deal with. Inside Teku's jar files, the inner files have different timestamps. These are the only differences though. One of these distributions was built on Windows x86. These comparisons are meld.

Screen Shot 2022-06-29 at 2 24 19 PM

As you can see here, meld considers these directories to be the same, except for the attributes.

Screen Shot 2022-06-29 at 2 27 29 PM

I think we will need to repackage the Teku jar files like we do the archives. Lemme see what I can come up with.

jtraglia:
Okay, one more issue. The file sizes are differ:

```
Permissions        Size User     Date Modified Name
.rw-r--r--  147,998,536 jtraglia 29 Jun 15:23  <from-mac>/teku-develop.zip
.rw-r--r--  147,998,539 jtraglia 29 Jun 15:23  <from-win>/teku-develop.zip
```

The Windows zip distribution is 3 bytes larger than the Mac distribution.

```
Permissions        Size User     Date Modified Name
.rw-r--r--  147,661,920 jtraglia 29 Jun 15:23  <from-mac>/teku-develop.tar.gz
.rw-r--r--  147,661,918 jtraglia 29 Jun 15:23  <from-win>/teku-develop.tar.gz
```

And the Mac tar distribution is 2 bytes larger than the Windows distribution.

jtraglia:
Hmm there are lots of small differences in the zip distributions:

Screen Shot 2022-06-29 at 4 23 23 PM

Mac on the left, Windows on the right. I think that Windows may be packaging these files as Jar files. I say that because there are lots of `0x7a` (`'z'` for "zip") and `0x6a` (`'j'` for "jar") differences. Might be simple fix, but probably not worth it.

On the other hand, tar distributions are wildly different:

Screen Shot 2022-06-29 at 4 29 42 PM

Not sure how Windows handles tar-balling files, but it's not the same as Mac/UNIX.

We may have to live with the fact that Windows distributions are different than UNIX.

jtraglia:
I can confirm that Mac and Linux produce the exact same distributions. For this branch, latest commit, and JDK 11.0.5:

```
$ sha256sum build/distributions/*
7d321bc3d900bf38d67a07533004ea3402214d69df1975cdc3c105b090a16333  build/distributions/teku-develop.tar.gz
6efa3fcb03729c6b9a0aff3b669c135df9dcb2a410ea94bdc36a8e81a87d193b  build/distributions/teku-develop.zip
```

Regarding Windows, the packaged distributions are different but they do provide the exact same files when unpacked. I don't think there's anything we can do to make them the same. With that, I believe this is ready for a final review.

ajsutton:
My guess is that the file properties we get from Windows are quite different to what we get from Mac/Linux and that shows up as the changes in the .tar.gz because it tends to include file attributes more accurately.  We could potentially filter and normalise them but I think it adds risk that we'll mess up some important attribute in the process.

For the timestamps in jar files, I think rather than having to repackage all of them which will be slower, we can use the gradle option `preserveFileTimestamps = false` for the jar files.  Unlike the tar & zip archives they aren't intended to be expanded so won't cause confusion for users. I could get reproducible builds on the same box (after a clean) with the diff below:

```diff
diff --git a/build.gradle b/build.gradle
index f840410d1a..b5d50739f8 100644
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,7 @@ allprojects {
     classifier = 'sources'
     from sourceSets.main.allSource
     jar.reproducibleFileOrder = true
+    jar.preserveFileTimestamps = false
   }
 
   sourceCompatibility = '11'
@@ -780,9 +781,7 @@ subprojects {
       )
     }
     reproducibleFileOrder = true
-    doLast {
-      repackage(archiveFile.get().toString(), lastCommitDate())
-    }
+    preserveFileTimestamps = false
   }
 
   dependencies {
```

Unfortunately both with the above change and with this branch unchanged, I'm getting a different result building on Mac vs a Ubuntu docker image. The content of the distribution is identical but the actual tar and zip files differ still. That suggests my approach to repackaging isn't consistent across OS or specific JVM builds (both were Java 18 but not exactly the same JVM build).

I actually think it's valuable to have our build server create completely reproducible builds which this would give us and the build is done on a docker image so it's doable for anyone to use that same image to build from source. Potentially we could look into gradle toolchains to specify a more exact version of java to use when building as well but I'm not too familiar with how that works.

jtraglia:
Yeah I think you will need to use the same JDK version to guarantee the same result. I just tested this with JDK 17.0.3 on three different machines (mix of Mac/Linux/x86/ARM) and their files all had the same hashes. There could also be a caching issue. When in doubt, I like to run `git clean -fdx` to delete all of the untracked (build) files.

> both were Java 18 but not exactly the same JVM build

What exactly do you mean by this? Like 18.0.1 vs 18.0.2?

ajsutton:
> What exactly do you mean by this? Like 18.0.1 vs 18.0.2?

I think different builds of OpenJDK - I think Ubuntu does their own build, and my Mac I think has adoptium. The version reported by `java --version` is quite a different format so not exactly sure how closely they match.  I wasn't really expecting minor differences in JDK to change things to be honest but so be it.  I could reproduce the exact same build using the same docker image we use in CI which I think is the main thing.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ For information on changes in released versions of Teku, see the [releases page]
 
 ### Additions and Improvements
  - Log a warning instead of a verbose error if node is syncing while performing sync committee duties
+ - Distributions created from the same git commit will have exactly the same contents, *i.e.*, they are reproducible
```
ajsutton: We possibly should list this as being identical when the distribution is built using the same docker image as well as the same git commit.

jtraglia:
Ah I understand now. I didn't think to test that. I've just been testing openjdk via brew & apt.

ajsutton (approved):
LGTM, just one last nit with the changelog so we don't over promise. :)
Thanks for doing this - it's a great improvement.

CHANGELOG.md
jtraglia: Yes, I agree. Will make that change now.

jtraglia:
Alright. I'm done now. Thanks again for all of your help with this. I didn't expect it to be so complicated.

PR #5864, opened 2022-06-29
Move etheruem:core into ethereum:spec

ajsutton:
## PR Description
Makes key testing utils like ChainBuilder available to the spec module so we can test it more easily.

## Documentation

## Changelog


rolfyone (approved):
Lgtm

Issue #4653, opened 2021-11-16
move to non reflection rest api for beacon-api


rolfyone:
Build out the standard beacon-api using the framework started in the validator-api.

The existing schema classes may be still useful in the acceptance tests.

The expectation is that once this task is complete:
 - beacon-api will be using infrastructure/restapi framework
 - acceptance tests will be either decoding to Map with JsonTestUtil, or will be using existing schema objects as appropriate
 - tests that currently output openApi schema validation should should be updated

PR #5865, opened 2022-06-29
Create JsonTypeDefinitionBeaconRestApi

courtneyeh:
#4653 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/JsonTypeDefinitionBeaconRestApi.java
```diff
@@ -13,15 +13,236 @@
...
+      final DataProvider dataProvider,
+      final Eth1DataProvider eth1DataProvider,
+      final Spec spec) {
+    restApi = create(config, dataProvider, eth1DataProvider, spec);
+  }
 
   @Override
   public void start() {
-    throw new UnsupportedOperationException("Unsupported start operation");
+    restApi.start().ifExceptionGetsHereRaiseABug();
```
ajsutton: We should change `start` (and `stop`) so that it returns a `SafeFuture`.  The old implementation can just return `SafeFuture.COMPLETED` but where this is called in `BeaconChainController` it's actually creating a `SafeFuture` from the call:

```java
return initialize()
        .thenCompose(
            (__) -> SafeFuture.fromRunnable(() -> beaconRestAPI.ifPresent(BeaconRestApi::start)));
```

so that would just become:
```java
return initialize()
        .thenCompose(
            __ -> beaconRestAPI.map(BeaconRestApi::start).orElse(SafeFuture.COMPLETE)));
```

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/JsonTypeDefinitionBeaconRestApi.java
```diff
@@ -13,15 +13,236 @@
...
+        .exceptionHandler(
+            ContentTypeNotSupportedException.class,
+            (throwable) -> new HttpErrorResponse(SC_UNSUPPORTED_MEDIA_TYPE, throwable.getMessage()))
+        .exceptionHandler(
+            BadRequestException.class,
+            (throwable) -> HttpErrorResponse.badRequest(throwable.getMessage()))
+        .exceptionHandler(
+            JsonProcessingException.class,
+            (throwable) -> HttpErrorResponse.badRequest(throwable.getMessage()))
+        .exceptionHandler(
```
ajsutton, line 165: We need to add a new handler for `MissingDepositsException` in here (it was only added to the old setup while you were working on this PR).

ajsutton (commented):
Looks good.

It would be worth logging a separate ticket to convert `AbstractDataBackedRestAPIIntegrationTest` over to using the new system. There's a lot of tests that use that so if we can switch it to the new approach pretty much all the tests should switch over too. In particular `MigratedOpenApiIntegrationTest` would then confirm we didn't miss any handlers.

Ideally we'd have it automatically run both versions but I'm not sure how to achieve that without duplicating all the subclasses or having to change a heap of stuff.  Given we've got this feature toggled off I'd be happy to merge this one the start/stop futures are addressed then come back, particularly because we know some tests will fail because we don't have GetEvents or swagger UI yet.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/JsonTypeDefinitionBeaconRestApi.java
courtneyeh: Okay I have tried that, and am not sure what to do in the `orElse()` as returning a `SafeFuture<Void>` is not allowed.
Screen Shot 2022-07-01 at 2 16 53 PM

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApi.java
```diff
@@ -13,9 +13,11 @@
...
 
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+
 public interface BeaconRestApi {
 
-  public void start();
+  public SafeFuture<?> start();
 
-  public void stop();
+  public SafeFuture<?> stop();
```
ajsutton: Tiny nit: We don't need the `public` in these method declarations - all interface methods are public.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/ReflectionBasedBeaconRestApi.java
```diff
@@ -332,7 +333,7 @@ public ReflectionBasedBeaconRestApi(
   }
 
   @Override
-  public void start() {
+  public SafeFuture<?> start() {
     try {
       app.start();
     } catch (RuntimeException ex) {
```
ajsutton, line 340: Rather than the `throw`  in the line below here we should change it to be `return SafeFuture.failedFuture(new InvalidConfigurationException(...` so the exception is returned as a failed future rather than thrown.

ajsutton (approved):
LGTM. Just a couple of minor nits.

PR #5866, opened 2022-06-29
Support multiple beacon node endpoints for a validator client

### LINKED ISSUE #5237: Feature: Multiple consensus clients for validator client
### Description

Analogous to the validator client option `--beacon-node-api-endpoint`, introduce `--beacon-node-api-endpoints` that takes a comma-separated list of one or more consensus client REST API endpoints. Eventually, deprecate `--beacon-node-api-endpoint`.

### Rationale

This helps with setups that use a single validator client pointed at multiple consensus clients in the backend. For example, I am running a Lighthouse validator client that fails over between Lighthouse, Teku and Nimbus. It'd be great to have more choices for the validator client.

StefanBratanov:
## PR Description

- Introduced `RemoteValidatorApiChannelInterface` which just exposes the endpoint for the remote interface
- Introduced `FailoverRemoteValidatorApiHandler` which deals with failing over
- Deleted unused files: `BeaconChainEvent` and `BeaconChainReorgEvent`
- Made `EventSourceBeaconChainEventAdapter` failover

Things pending:
- Think about metrics
- Add a flag for failover subnet subscription (default true)

#### Testing failover for event stream.

1. Start BN 1 and BN 2
2. Stop BN1
3. Start again BN1 after some time

image

## Fixed Issue(s)
fixes #5237 

## Documentation

## Changelog


ajsutton:
That's the kind of approach I had in mind.  I think the unimplemented methods are going to be interesting ones. I'd suggest having the first PR for this just treat all requests the same way and just find the first beacon node the request succeeds for. That gives us something simple that will generally work pretty well. It just won't necessarily create aggregates properly after fail overs because the subscriptions weren't sent to the node that wound up being asked for the aggregate and it won't give any notice to the EL when creating blocks post-merge (again because the prepare proposer call may not have gone to the node we create a block for).

Then we can follow up and address those.  One option would be to track which subscriptions we need and send them to the next node in line when we fail over, though we may not be giving much notice still and doing the tracking will add a reasonable amount of complexity.

The other option is to send them to all nodes regardless. That is very likely the right answer for prepare proposer since you want the EL to have as much time as possible to create a good block and it's very low overhead if you aren't proposing. It's more expensive for subscribing to subnets though. Worth seeing what Lighthouse does and may need the policy to be configurable.

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
```diff
@@ -235,6 +236,32 @@ public static <U> SafeFuture<Void> allOfFailFast(final SafeFuture<?>... futures)
...
+   * exceptionally with the error captured from {@link #allOf(SafeFuture[])} ()}.
+   *
+   * @param futures the futures which will need to be completed
+   * @return a new {@link SafeFuture} that is completed with the first successful result from the
+   *     given futures or completed exceptionally if all futures complete exceptionally
+   */
+  @SuppressWarnings("FutureReturnValueIgnored")
+  public static <T> SafeFuture<T> firstSuccess(final List<SafeFuture<T>> futures) {
+    SafeFuture<T> result = new SafeFuture<>();
+    allOf(futures.toArray(SafeFuture[]::new))
```
ajsutton, line 251: We probably should clarify in the javadoc that this method will wait for all futures to complete, then will return the first in the list that succeeded.

Though I suspect that's not the behaviour we want to use for fail over cases as then if one of the secondary nodes is slow we'll delay everything while waiting for it even if the primary is fast.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
```diff
@@ -0,0 +1,265 @@
...
+    final SafeFuture<T> primaryResponse = request.run(primaryDelegate);
+    final List<SafeFuture<T>> failoversResponses =
+        failoverDelegates.stream().map(request::run).collect(Collectors.toList());
+    return primaryResponse
+        .exceptionallyCompose(
+            throwable -> {
+              LOG.debug(
+                  "Request which is sent to all configured Beacon nodes failed on the primary Beacon Node.",
+                  throwable);
+              return SafeFuture.firstSuccess(failoversResponses);
```
ajsutton, line 222: I suspect we actually want to return success here as soon as any fail over succeeds without waiting for them all.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/PingUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class PingUtils {
+
+  private static final int PING_TIMEOUT_MILLIS = 2500;
+
+  public static boolean hostIsReachable(final HttpUrl hostUrl) {
+    try (final Socket socket = new Socket()) {
+      final InetSocketAddress socketAddress =
+          new InetSocketAddress(InetAddress.getByName(hostUrl.host()), hostUrl.port());
+      socket.connect(socketAddress, PING_TIMEOUT_MILLIS);
```
ajsutton, line 30: This is probably not a good thing to do to servers as it opens and closes a connection without actually sending any response. It's not wrong but can lead to some weird alerts going off and doesn't necessarily check the liveness of the host (it may have just connected to a proxy which is up but the backend isn't).  Actually it may not use the right proxy config either if a specific local HTTP proxy is configured this wouldn't use it because it's creating a raw tcp socket.

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandlerTest.java
```diff
@@ -0,0 +1,404 @@
...
+import tech.pegasys.teku.validator.api.ProposerDuties;
+import tech.pegasys.teku.validator.api.RemoteValidatorApiChannel;
+import tech.pegasys.teku.validator.api.SendSignedBlockResult;
+import tech.pegasys.teku.validator.api.SubmitDataError;
+import tech.pegasys.teku.validator.api.SyncCommitteeDuties;
+import tech.pegasys.teku.validator.api.SyncCommitteeSubnetSubscription;
+import tech.pegasys.teku.validator.api.ValidatorApiChannel;
+import tech.pegasys.teku.validator.remote.FailoverValidatorApiHandler.ValidatorApiChannelRequest;
+
+class FailoverValidatorApiHandlerTest {
```
ajsutton, line 73: Would be good to add a test here where one node doesn't respond and check that we don't get stuck waiting for it.

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
```diff
@@ -300,6 +327,15 @@ public SafeFuture<T> alwaysRun(final Runnable action) {
         .thenPeek(value -> action.run());
   }
 
+  public void finish(final BiConsumer<T, Throwable> action) {
```
ajsutton, line 330: I'm not entirely convinced this is a good addition - seems like it would be better to use the separate `finish(success, error)` version.  I can't see anywhere it's actually used either.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
```diff
@@ -14,78 +14,188 @@
...
+        .client(okHttpClient);
+  }
+
+  private void schedulePingingOfPrimaryBeaconNode() {
+    final Cancellable cancellable =
+        asyncRunner.runWithFixedDelay(
+            () -> {
+              final boolean primaryBeaconNodeIsReachable =
+                  PingUtils.hostIsReachable(primaryApiEndpoint);
+              if (primaryBeaconNodeIsReachable) {
```
ajsutton, line 163: I suspect here we need to just keep trying to connect to the primary event stream and then when it reconnects drop the fail over connection. I suspect the key to making this work well is going to be a wrapper around `EventSourceHandler` that knows which event source it's getting events from and can handle the onOpen and onClosed events to trigger starting or stopping a fail-over connection. I suspect that would also let us leave reconnecting to the primary to the built-in retry capability of `EventSource` which has some nice back-off handling and so on.

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
StefanBratanov: have to delete this, thanks

StefanBratanov:
Closing since this PR will be split into smaller ones

PR #5868, opened 2022-06-29
Improve `ReexecutingExecutionPayloadBlockManager`

tbenr:
- Reduces reexecution window to only one slot.

- Adds a check where the reexecution applies only to blocks close to currenSlot (ie not already expired). This will avoid reexecution while importing blocks during sync. (not super sure about this, but in case of massive problems on EL while syncing this will avoid unnecessary over-reexecutions)

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
@@ -44,7 +44,7 @@ public class BlockManager extends Service
     implements SlotEventsChannel, BlockImportChannel, BlockImportNotifications {
   private static final Logger LOG = LogManager.getLogger();
 
-  private final RecentChainData recentChainData;
+  protected final RecentChainData recentChainData;
   private final BlockImporter blockImporter;
   private final PendingPool<SignedBeaconBlock> pendingBlocks;
   private final BlockValidator validator;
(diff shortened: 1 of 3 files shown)


ajsutton:
This looks fine to me but is pushing in the opposite direction to #5840 which makes life difficult. Do we need to step back and take an overall view of how to handle retries and when to trigger them?  Maybe there needs to be different behaviour when we're doing a long sync vs handling gossip? Seems like it's very hard to find the right balance between DOS'ing the EL and getting stuck because we didn't reexecute a block we need.

tbenr:
I definitely pro separating the logic. I don't like this class, it's difficult to have a clear understanding about relationship with gossip and sync. I'm in favour of having the retry logic closer to each area. Yesterday i was looking at the erigon-teku box and it is still unclear to me why was in stall in syncing state, without retrying anymore.

ajsutton:
> I definitely pro separating the logic. I don't like this class, it's difficult to have a clear understanding about relationship with gossip and sync. I'm in favour of having the retry logic closer to each area. Yesterday i was looking at the erigon-teku box and it is still unclear to me why was in stall in syncing state, without retrying anymore.

Yeah one option I see is to remove all descendent blocks from the pending pool when a block fails to execute. We'd solve the not retrying problem but we'd potentially keep downloading the same chain of blocks over and over every time a new block is added to the end.  So maybe we need to remove them but put them in a cache to save re-requesting them again if needed.

Would definitely be good to step back and rethink our whole approach in this area.

PR #5869, opened 2022-06-29
Resend `ForkChoiceUpdated` if not changing for 30 seconds

### LINKED ISSUE #5370: [Merge] Periodically resend forkChoiceUpdated
### Description
If the chain head hasn't changed for more than some period of time (maybe a slot?) then resend the last forkChoiceUpdated message to the EL.  Could limit this to only if the last response was SYNCING to see if the EL has now finished syncing but probably easier to always send it.

tbenr:
As per my analysis we are calling `ForkChoiceNotifier::onForkChoiceUpdated` at least on every slot via 
`tech.pegasys.teku.statetransition.forkchoice.ForkChoiceTrigger#onAttestationsDueForSlot` -> `processHead` ->  `notifyForkChoiceUpdatedAndOptimisticSyncingChanged`.

That given, to periodically resend `fcu` it's just matter of relaxing the "already sent"  condition `ForkChoiceUpdateData` and make it time dependant.

for the records, this covers also the situation in which were marking `ForkChoiceUpdateData` as `sent` even if the actual call was throwing (timeout or other networking issues). With this change we will resend it no matter what.

## Fixed Issue(s)
fixes #5370

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceUpdateData.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceUpdateData.java
@@ -40,7 +40,7 @@ public class ForkChoiceUpdateData {
   private final Optional<Bytes32> terminalBlockHash;
   private final SafeFuture<Optional<ExecutionPayloadContext>> executionPayloadContext =
       new SafeFuture<>();
-  private UInt64 sentTimestamp = UInt64.ZERO;
+  private UInt64 toBeSentAtTime = UInt64.ZERO;
 
   private long payloadBuildingAttributesSequenceProducer = 0;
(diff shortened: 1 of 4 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceUpdateData.java
```diff
@@ -208,6 +208,18 @@ public ForkChoiceState getForkChoiceState() {
     return forkChoiceState;
   }
 
+  private boolean doesNotRequireSend(final UInt64 currentTimestamp) {
```
ajsutton: nit: I'd probably call this `isSentRecently` mostly because the negative hurts my brain, but not strong on it so feel free to ignore.

ajsutton (approved):
LGTM.

PR #5871, opened 2022-06-29
add exception handler for MissingDepositsException

### LINKED ISSUE #5836: Block production can fail when eth1 deposits are not current
```
ERROR | BeaconRestApi | Failed to process request to URL ..../eth/v2/validator/blocks/....
tech.pegasys.teku.validator.coordinator.MissingDepositsException: Unable to create block because ETH1 deposits are not available. Missing deposits 410438 to 410582
```

In this case, we've failed with 500 with full stack trace.

What'd be nice is if GetNewBlock fails (v2 only is fine), and the exception is a `MissingDepositsException`, we should pass back the error message, still as a 500 error, and we won't need to fall through to the default exception handler and print the stack trace.

This will give the user a lot more context about what may be going wrong.

sragss:
## Fixed Issue(s)
Fixes #5836 

## Documentation

## Changelog


ajsutton (approved):
LGTM.  Thanks for this.

ajsutton:
@courtneyeh Note: new exception handler we'll need to add to the new type definition based rest api setup.

PR #5887, opened 2022-07-04
Fix latestValidHash was ignored for INVALID execution payload validation

zilm13:
## PR Description
Suppose we have following chain on optimistic sync:
Block 1 (optimistic) <- Block 2 (optimistic) <- Block 3 (optimistic)
When we call execution engine with `newPayload` method from the next, Block 4, it returns `INVALID` and `latestValidHash` with root from execution payload of the Block 1. It means 2 facts for us: Block 1 execution payload becomes validated, Block2 and its children are invalid. This case was ignored in our code and didn't trigger `ForkChoiceStrategy` due to absence of transition block.
Also I've changed `OptimisticExecutionPayloadExecutor.optimisticallyExecute` method interface as it is confusing.

## Documentation

## Changelog


zilm13:
I've reverted `optimisticallyExecute` interface changes because it's used in reference test. This way I don't understand why do we not mark `INVALID` payload execution to return false there. Looks like a weird hack

zilm13:
There is an ongoing discussion, whether Block 2 and Block 3 should be INVALID or stay optimistic in this case, drafting for a while until resolution
UPDATED: I've clarified 2 and 3 should be INVALID and it will not be changed as EL couldn't definitely know that 4th is INVALID without importing 2 and 3 before that

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoicePayloadExecutor.java
```diff
@@ -79,6 +79,10 @@ public boolean optimisticallyExecute(
                       if (result.hasValidStatus()) {
                         return transitionBlockValidator.verifyTransitionBlock(
                             latestExecutionPayloadHeader, block);
+                      } else if (result.hasInvalidStatus()
+                          && result.getLatestValidHash().isPresent()) {
+                        return SafeFuture.completedFuture(
+                            new PayloadValidationResult(block.getParentRoot(), result));
```
tbenr: I'm not sure this is correct. You are assuming that the parent of the current block is invalid, by treating it as an `invalidTransitionBlock`. I think we need to manage it differently. By handling it in `ForkChoice#importBlockAndState` and passing the current valid optimistic head and `lastValidBlockHash` to a new method in `ProtoArray`  that searches the payload in protoarray similarly to what is already done in `markNodeInvalid`.

Maybe it could be a good solution to store the information in a different field in result object so it can be correctly managed in 'onExecutionPayloadResult'

zilm13: You are right, I'm marking parent as invalid which could be not true, I will re-implement it in proposed way

zilm13:
Reworked

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -306,10 +306,18 @@ private BlockImportResult importBlockAndState(
...
-                      .onExecutionPayloadResult(invalidTransitionBlockRoot, payloadResult));
+                      .onExecutionPayloadResult(invalidTransitionBlockRoot, payloadResult),
+              () ->
+                  recentChainData
+                      .getChainHead()
+                      .ifPresent(
+                          chainHead ->
+                              getForkChoiceStrategy()
+                                  .onInvalidExecutionResultWithoutTransition(
+                                      chainHead.getRoot(), payloadResult)));
```
ajsutton: This will incorrectly mark our canonical chain as invalid even if the block is a non-canonical fork.  We should be passing `block.getParentRoot()` instead of the chain head root here as that's the tip of the chain we're importing to.  The block we're importing won't be known in protoarray (as we haven't finished importing it) but it's parent must be known or we'd have deferred the import.

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -286,6 +286,34 @@ public void markNodeInvalid(final Bytes32 blockRoot, final Optional<Bytes32> lat
...
+    }
+
+    final int index = maybeFirstInvalidNodeIndex.get();
+    final ProtoNode node = getNodeByIndex(index);
+    markNodeValid(node.getParentRoot());
+    node.setValidationStatus(INVALID);
+    removeBlockRoot(node.getBlockRoot());
+    markDescendantsAsInvalid(index);
+    // Applying zero deltas causes the newly marked INVALID nodes to have their weight set to 0
+    applyDeltas(new LongArrayList(Collections.nCopies(getTotalTrackedNodeCount(), 0L)));
```
ajsutton: It feels like we're duplicating a lot of code here that we also have in `markNodeInvalid` when the only difference should be that in the existing `markNodeInvalid` method we would always mark the `headRoot` as invalid and we would skip that in this version. Given the complexity involved in this duplicated code it would be good to try and refactor to avoid the duplication.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
zilm13: that was my first attempt to do this, ok, will return to that

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
zilm13: Yeah, maybe some boolean flag, will try to refactor now

zilm13:
@ajsutton came to this. Not very elegant, but the logic is clearly visible.

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -266,8 +273,21 @@ public void markNodeInvalid(final Bytes32 blockRoot, final Optional<Bytes32> lat
...
-          findFirstInvalidNodeIndex(maybeIndex.get(), latestValidHash.get());
+      final Optional<Integer> maybeFirstInvalidNodeIndex;
+      if (verifiedInvalidTransition) {
+        maybeFirstInvalidNodeIndex =
+            findFirstInvalidNodeIndex(maybeIndex.get(), latestValidHash.get());
+      } else {
+        maybeFirstInvalidNodeIndex =
+            verifyInvalidIndex(maybeIndex.get(), latestValidHash.get())
+                .flatMap(
+                    invalidIndex -> findFirstInvalidNodeIndex(invalidIndex, latestValidHash.get()));
```
ajsutton: I think it would be clearer with an if rather than using an Optional. So I think I'd do something like:

```java
  public void markNodeInvalid(
      final Bytes32 blockRoot,
      final Optional<Bytes32> latestValidHash,
      final boolean verifiedInvalidTransition) {
    if (!verifiedInvalidTransition && latestValidHash.isEmpty()) {
      // Couldn't find invalid chain segment with lack of data
      return;
    }
    final Optional<Integer> maybeIndex = indices.get(blockRoot);
    if (maybeIndex.isEmpty()) {
      LOG.debug("Couldn't update status for block {} because it was unknown", blockRoot);
      return;
    }
    final int index;
    final ProtoNode node;
    if (latestValidHash.isPresent()) {
      final Optional<Integer> maybeFirstInvalidNodeIndex =
          findFirstInvalidNodeIndex(maybeIndex.get(), latestValidHash.get());
      if (!verifiedInvalidTransition) {
        if (nodeHasExecutionHash(maybeIndex.get(), latestValidHash.get())) {
          // head blockRoot is the latestValidHash so nothing to do
          return;
        }
        if (maybeFirstInvalidNodeIndex.isEmpty()) {
          // Nothing to do: latestValidHash was not found, no proof of invalid transition
          return;
        }
      }

      index = maybeFirstInvalidNodeIndex.orElse(maybeIndex.get());
      node = getNodeByIndex(index);
      // We found the latestValidHash so mark it as valid
      if (maybeFirstInvalidNodeIndex.isPresent()) {
        markNodeValid(node.getParentRoot());
      }
    } else {
      index = maybeIndex.get();
      node = getNodeByIndex(index);
    }

    node.setValidationStatus(INVALID);
    removeBlockRoot(node.getBlockRoot());
    markDescendantsAsInvalid(index);
    // Applying zero deltas causes the newly marked INVALID nodes to have their weight set to 0
    applyDeltas(new LongArrayList(Collections.nCopies(getTotalTrackedNodeCount(), 0L)));
  }

  private boolean nodeHasExecutionHash(final int nodeIndex, final Bytes32 executionHash) {
    return getNodeByIndex(nodeIndex).getExecutionBlockHash().equals(executionHash);
  }
```

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -257,7 +257,14 @@ public void markNodeValid(final Bytes32 blockRoot) {
     }
   }
 
-  public void markNodeInvalid(final Bytes32 blockRoot, final Optional<Bytes32> latestValidHash) {
+  public void markNodeInvalid(
+      final Bytes32 blockRoot,
+      final Optional<Bytes32> latestValidHash,
+      final boolean verifiedInvalidTransition) {
```
ajsutton, line 289: Might be worth considering making this version of method with the extra boolean verifiedInvalidTransition then have two public methods that delegate to it - one `markNodeInvalid` and one `markParentChainInvalid` or something like that. Not sure about those names so would need a comment explaining what they actually do and when to use them, but it would be good to avoid having the random true/false scattered through the code where this is called as it doesn't mean much - having the two separate public methods lets us have a nice API while sharing the logic via the private method.

ajsutton (commented):
Looks good, just a couple of tweaks to try and make the markNodeInvalid method more maintainable.

zilm13:
@ajsutton I've applied requested changes

ajsutton (approved):
LGTM.

PR #5888, opened 2022-07-04
Change `prepareBeaconProposer` return type to be a future instead of void

StefanBratanov:
## PR Description

- Changed signature of `prepareBeaconProposer` in `ValidatorApiChannel` to return SafeFuture<Void> instead of void
- Changed tests to be consistent with `PostRegisterValidator` tests which covered more scenarios

## Fixed Issue(s)
will help for #5237 

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostPrepareBeaconProposer.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostPrepareBeaconProposer.java
@@ -36,6 +36,7 @@ import tech.pegasys.teku.api.ValidatorDataProvider;
 import tech.pegasys.teku.beaconrestapi.MigratingEndpointAdapter;
 import tech.pegasys.teku.infrastructure.http.HttpStatusCodes;
 import tech.pegasys.teku.infrastructure.json.types.DeserializableTypeDefinition;
+import tech.pegasys.teku.infrastructure.restapi.endpoints.AsyncApiResponse;
 import tech.pegasys.teku.infrastructure.restapi.endpoints.EndpointMetadata;
 import tech.pegasys.teku.infrastructure.restapi.endpoints.RestApiRequest;
 import tech.pegasys.teku.spec.datastructures.operations.versions.bellatrix.BeaconPreparableProposer;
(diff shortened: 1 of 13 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostPrepareBeaconProposer.java
```diff
@@ -113,8 +114,10 @@ public void handleRequest(final RestApiRequest request) throws JsonProcessingExc
     if (!isProposerDefaultFeeRecipientDefined) {
       STATUS_LOG.warnMissingProposerDefaultFeeRecipientWithPreparedBeaconProposerBeingCalled();
     }
-    validatorDataProvider.prepareBeaconProposer(request.getRequestBody());
-    request.respondWithCode(SC_OK);
+    request.respondAsync(
+        validatorDataProvider
+            .prepareBeaconProposer(request.getRequestBody())
+            .thenApply(AsyncApiResponse::respondOk));
```
rolfyone, line 120: probably worth handling the case when errors occur here so that the user won't get a 500 if it does occur... a few are handled but if there's specific exceptions this is where you'd want to be assessing them...

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java
```diff
@@ -414,18 +414,17 @@ public void subscribeToPersistentSubnets(final Set<SubnetSubscription> subnetSub
...
-                                new tech.pegasys.teku.api.schema.bellatrix.BeaconPreparableProposer(
-                                    pbp.getValidatorIndex(), pbp.getFeeRecipient()))
-                        .collect(toList())))
-        .finish(error -> LOG.error("Failed to prepare beacon proposers", error));
+    return sendRequest(
+        () ->
+            apiClient.prepareBeaconProposer(
+                beaconPreparableProposers.stream()
+                    .map(
+                        bpp ->
```
rolfyone: could probably call the variable something more like `proposer` and it'd read more easily...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostPrepareBeaconProposer.java
StefanBratanov: thanks, good to know, this specific endpoint wasn't handling specific exceptions before and don't think there are any to include as of now

tbenr (approved):
LGTM

PR #5890, opened 2022-07-05
change the way we provide additional params via environment and config file

### LINKED ISSUE #5878: Validator Options parameters not read from `--config-file` when other params are specified via CLI args
Specifying via CLI an option from `ValidatorProposerOptions` (ie `--validators-proposer-config`)  or `ValidatorOptions` (ie `--validators-graffiti`) prevents params from `ValidatorKeysOptions` to be read via yaml config file specified via `--config-file`.

### Example

CLI:
```bash
./teku --config-file ./test.yaml --validators-graffiti test
```
test.yaml
```yaml
validator-keys: ["/etc/teku/keys:/etc/teku/secrets"]
```

`validator-keys` param is ignored and teku will run as it has been never specified.

by removing `--validators-graffiti test` from CLI, `validator-keys` will be read again.

### Initial thought
it seems it's related to how picocli initializes `ArgGroup` groups and how it will apply defaults values.
Since we provide config file functionality as a default value provider, it seems that specifying an argument from a `ArgGroup` causes initialization of all its sub `ArgGroup` which then are considered initialized and no default values are provided.
The same happens if you initialize a sibling group.

A solution could be to flatten all the arguments to `ValidatorOptions` but then we have rethink how to avoid duplication in `VoluntaryExitCommand`.

tbenr:
This PR moves away from relying on picocli default provider and programmatically reconstructs the command line arguments with the additional params from environment and config file.

The basic idea is:
- get all the options available for the current commandLine (or subcommand)
- exclude the options that has been already specified as argument
- search for matching params in environment\configFile
- rebuild the CLI arguments by appending matched params to the original params

fixes #5878

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/BeaconNodeCommand.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/BeaconNodeCommand.java
@@ -30,7 +30,6 @@ import org.hyperledger.besu.plugin.services.metrics.MetricCategory;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Mixin;
-import picocli.CommandLine.Model.ArgSpec;
 import picocli.CommandLine.Model.OptionSpec;
 import picocli.CommandLine.Option;
 import picocli.CommandLine.ParameterException;
@@ -65,7 +64,7 @@ import tech.pegasys.teku.cli.subcommand.internal.InternalToolsCommand;
 import tech.pegasys.teku.cli.util.AdditionalConfigProvider;
 import tech.pegasys.teku.cli.util.CascadingDefaultProvider;
 import tech.pegasys.teku.cli.util.EnvironmentVariableDefaultProvider;
-import tech.pegasys.teku.cli.util.YamlConfigFileDefaultProvider;
+import tech.pegasys.teku.cli.util.YamlConfigFileParamsProvider;
 import tech.pegasys.teku.config.TekuConfiguration;
 import tech.pegasys.teku.infrastructure.exceptions.ExceptionUtil;
 import tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException;
@@ -235,19 +234,22 @@ public class BeaconNodeCommand implements Callable<Integer> {
       return handleParseException(e, args);
     }
 
+    // switch to last subcommand if needed
+    final ParseResult maybeSubCommandParseResult =
+        parseResult.hasSubcommand() ? parseResult.subcommand() : parseResult;
+
     // calculate potential additional params from config provider
     final List<OptionSpec> potentialAdditionalParams =
-        commandLine.getCommandSpec().args().stream()
-            .filter(ArgSpec::isOption)
-            .map(argSpec -> (OptionSpec) argSpec)
-            .filter(optionSpec -> !parseResult.matchedOptionsSet().contains(optionSpec))
+        maybeSubCommandParseResult.commandSpec().options().stream()
+            .filter(
+                optionSpec -> !maybeSubCommandParseResult.matchedOptionsSet().contains(optionSpec))
             .collect(Collectors.toUnmodifiableList());
 
     final AdditionalConfigProvider configurationValueProvider =
         additionalConfigProvider(commandLine, configFile);
 
     final Map<String, String> additionalParams =
-        configurationValueProvider.getAdditionalConfigs(potentialAdditionalParams);
+        configurationValueProvider.getAdditionalParams(potentialAdditionalParams);
 
     // build new argument list by concatenating original args and the additional params
     final String[] enrichedArgs =
@@ -289,7 +291,7 @@ public class BeaconNodeCommand implements Callable<Integer> {
 
     return new CascadingDefaultProvider(
         new EnvironmentVariableDefaultProvider(environment),
-        new YamlConfigFileDefaultProvider(commandLine, configFile.get()));
+        new YamlConfigFileParamsProvider(commandLine, configFile.get()));
   }
 
   private int handleParseException(final CommandLine.ParameterException ex, final String[] args) {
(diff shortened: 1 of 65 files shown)


teku/src/main/java/tech/pegasys/teku/cli/BeaconNodeCommand.java
```diff
@@ -216,15 +222,48 @@ public int parse(final String[] args) {
...
+
+    final ParseResult parseResult;
+    try {
+      parseResult = commandLine.parseArgs(args);
+    } catch (ParameterException e) {
+      return handleParseException(e, args);
+    }
+
+    // switch to last subcommand if needed
+    final ParseResult maybeSubCommandParseResult =
```
ajsutton: nit: The maybe in the name here keeps making me think it's an `Optional`.  Maybe we should just reuse the `parseResult` field here and reassign it? Or just find a name without the maybe.

teku/src/main/java/tech/pegasys/teku/cli/util/EnvironmentVariableParamsProvider.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Override
+  public Map<String, String> getAdditionalParams(List<OptionSpec> potentialParams) {
+    return environment.entrySet().stream()
+        .filter(envEntry -> envEntry.getKey().startsWith(ENV_VAR_PREFIX))
+        .map(envEntry -> Map.entry(translateKey(envEntry.getKey()), envEntry.getValue()))
+        .map(yamlEntry -> mapParam(potentialParams, yamlEntry))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
```
ajsutton: Could this just be?
```suggestion
        .flatMap(yamlEntry -> mapParam(potentialParams, yamlEntry).stream())
```

teku/src/main/java/tech/pegasys/teku/cli/util/YamlConfigFileParamsProvider.java
```diff
@@ -26,41 +26,83 @@
...
     }
 
-    // only options can be used in config because a name is needed for the key
-    // so we skip default for positional params
-    return argSpec.isOption() ? getConfigurationValue(((OptionSpec) argSpec)) : null;
+    return result.entrySet().stream()
+        .map(yamlEntry -> Map.entry(translateKey(yamlEntry.getKey()), yamlEntry.getValue()))
+        .map(yamlEntry -> mapParam(potentialParams, yamlEntry))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
```
ajsutton:
```suggestion
        .flatMap(yamlEntry -> mapParam(potentialParams, yamlEntry).stream())
```

teku/src/main/java/tech/pegasys/teku/cli/util/YamlConfigFileParamsProvider.java
```diff
@@ -26,41 +26,83 @@
...
+        .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+  }
+
+  private String translateKey(final String yamlKey) {
+    return "--" + yamlKey;
```
ajsutton: I don't think this correctly handles options that use a single `-` which we have a few of, mostly in subcommands.

Probably also applies to the env var handler too though apparently we never handled it properly there...

ajsutton (commented):
Generally looks good, I think we just need to check that option names with a single `-` still work in the Yaml provider (and probably fix them in the env provider).

teku/src/main/java/tech/pegasys/teku/cli/util/YamlConfigFileParamsProvider.java
tbenr: right, we were supporting it in config but not in env.
fixing it.

tbenr:
@ajsutton I managed the situation where multiple config\env referring to the same param are specified.
ie `TEKU_ETH1-ENDPOINT` and `TEKU_ETH1-ENDPOINTS`
Currently I explicitly stop and throw an exception instead of trying to resolve the conflict.
It makes me think that this could be a breaking change because in our previous version we were giving the precedence based on the ordering of `names` in the `@Option` definition.

To print a warning instead and replicate exactly the previous behaviour I should stream starting from Option names... should be doable...

yes I can definitely do the other way around.

ajsutton:
hmm, that is an interesting one.  It's probably not a particularly big deal if we can't support both being set but if we can keep it working that would definitely be better.

tbenr:
done.
Now it should be OK.
The only edge case I can reason about is:
giving an option like this:
```java
  @Option(
      names = {"-n", "--name", "--names"},
      split = ",")
  List<String> getNames();
```
and this env vars
```shell
TEKU_N=a
TEKU_NAME=b
TEKU_NAMES=c
```

previous behaviour: `name` will be `b`
new behaviour: `name` will be `a`

(because now we consider single-dash options too)

Similarly for yaml config file.

ajsutton:
> (because now we consider single-dash options too)

oh darn, I think I've given you bad advice here. I think we deliberately ignored single `-` options in the past because we had problems where teku wouldn't start because someone had an env var set called `TEKU_V` or `TEKU_H` and it triggered the version or help output.  It _could_ potentially have been avoided by only considering options that accept a value (even if it's optional) but I'm not sure which way we went now.

ajsutton:
The extractAPISpec failure is interesting too:
```
Invalid value for option '--logging': 'en_US.UTF-8' is not a valid log level. Supported values are [OFF|FATAL|WARN|INFO|DEBUG|TRACE|ALL]
```

Did we pick up an env var we didn't previously?

tbenr:
```
  @Option(
      names = {"-l", "--logging"}
```
better to filter out again single dashed options...
there must be a `TEKU_L`... let me check... oh no, I know what is happening.

was just removing 5 chars at the begining instead of filtering for `TEKU_`

image

`LC_ALL` was matching `-l`

tbenr:
I reintroduced the `startsWith("--")` requirement on the names

A quick assessment:
image

plus `-h` and `-V`

Now this version should behave exactly as the current one.

ajsutton (approved):
LGTM. Sorry for the wild goose chase.  This is great work.

PR #5892, opened 2022-07-05
Optimize `TransitionCaches.committeeShuffle` cache utilization

Nashatyrev:
## PR Description

When running Teku with high peers load (around 500-600 connected peers), noticed that `MiscHelpers.shuffleList` takes about 6%  of CPU. Turned out that `MiscHelpers.shuffleList` could be executed up to ~20 times per epoch (when expected just once)

There are 2 issues with the `TransitionCaches.committeeShuffle`:
1. Thread racing: when a number of attestations arrives simultaneously on several network threads the attestation handler  requests corresponding committee in parallel for attestations validation. Since the `LRUCache` is not synchronized (only the underlying `Map` is synchronized) the missed cache entry is calculated on several threads simultaneously. 
2. Too small cache size (2 entries): prior to epoch `N` `EpochCachePrimer` precalculates shufflings for epochs `N` and `N+1`, after that processing older attestations requires back shuffling for the epoch `N-1`. The shuffling for the epoch `N` is thus evicted and then needed to be recalculated again. 
Not 100% sure but it seems sometimes even shuffling for `N-2` is requested for ancient attestatations come with last block in the epoch `N-1`

Solution: 
1. Synchronize `LRUCache` while using non-synchronized underlying map. The synchronization performance impact should be 0
2. Increase `committeeShuffle` cache size from 2 to 4. The memory consumption is expected to increase by ~3Mb per cross-epoch fork (just 3Mb when no forks) 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/TransitionCaches.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/TransitionCaches.java
@@ -34,7 +34,7 @@ public class TransitionCaches {
   private static final int MAX_BEACON_PROPOSER_INDEX_CACHE = 1;
   private static final int MAX_BEACON_COMMITTEE_CACHE = 64 * 64;
   private static final int MAX_TOTAL_ACTIVE_BALANCE_CACHE = 2;
-  private static final int MAX_COMMITTEE_SHUFFLE_CACHE = 2;
+  private static final int MAX_COMMITTEE_SHUFFLE_CACHE = 4;
   private static final int MAX_EFFECTIVE_BALANCE_CACHE = 1;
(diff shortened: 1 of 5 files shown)


tbenr:
having the `get` synchronized makes a lot of sense.
I'm wondering how frequent is the `N-2` and based on that could make sense to set the MAX to 3 as a tradeoff between memory\CPU consumption.

tbenr (approved):
LGTM.
As per my previous comment, the only thing I'd  consider is setting the MAX cache to 3 to potentially lower down the memory overhead to 1.5MB per fork if `N-2` cache hit is very rare.

Nashatyrev:
@tbenr, you are right, just have a run during several epochs and haven't seen any such cases. Only a single shuffling is computed per epoch with cache size 3. So let's save 1.5M

Issue #5644, opened 2022-05-31
[MEV Boost\Builder] ValidatorRegistration signature - web3signer integration


tbenr:
We should add Web3Signer support as soon as they implement it on their side.

beetrooter:
Looks like web3signer is done (https://github.com/ConsenSys/web3signer/pull/596) . Do you know when this will reach a Teku release?

PR #5900, opened 2022-07-07
Remove epoch for Validator Registration

usmansaleem:
## PR Description
Remove epoch for validator registration as it is not required to determine the spec/fork to calculate the signing root.

## Fixed Issue(s)
Related to #5644

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
```diff
@@ -117,4 +117,13 @@ public Bytes signingRootForValidatorRegistration(
     final Bytes32 domain = miscHelpers.computeDomain(Domain.APPLICATION_BUILDER);
     return miscHelpers.computeSigningRoot(validatorRegistration, domain);
   }
+
+  /* overloaded method that uses the instance Spec instead of epoch to calculate the signing. Used
+  by non-Teku clients of this library. */
+  public Bytes signingRootForValidatorRegistration(
```
tbenr: I'd be more incline to use the same method but with an `Optional` epoch to avoid code duplication.
and then use `spec.getGenesisSpec().miscHelpers()` to avoid creating a new instance.

usmansaleem: 👍🏼 I'll incorporate it.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
```diff
@@ -112,8 +113,14 @@ public Bytes signingRootForSignVoluntaryExit(
   }
 
   public Bytes signingRootForValidatorRegistration(
-      final ValidatorRegistration validatorRegistration, final UInt64 epoch) {
-    final MiscHelpers miscHelpers = spec.atEpoch(epoch).miscHelpers();
+      final ValidatorRegistration validatorRegistration, final Optional<UInt64> epoch) {
```
ajsutton: Personally I'd write this to never take an epoch and always use the genesis miscHelpers.  Like deposits, the validator registration should be valid regardless of forks and that only works if we always use the genesis spec for them.

tbenr: Right! Sorry @usmansaleem for the back and forth :)

ajsutton: Sorry for the contradictions.  To be clear there should never be a need to pass `epoch` to `signingRootForValidatorRegistration` as we should always be using the genesis spec when signing to ensure it's consistent across all forks.

usmansaleem: all good, I've incorporated suggestions. Have a look at your convenience.

ajsutton (approved):
LGTM.

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@
 For information on changes in released versions of Teku, see the [releases page](https://github.com/ConsenSys/teku/releases).
 
 ## Unreleased Changes
-
+ - Validator Registration signature integration with external signer.
+ -
```
ajsutton: This should go under the `Additions and Improvements` section sub-category below (all the changes in this file are unreleased).

usmansaleem: 👍🏼 done.

PR #5901, opened 2022-07-07
Avoid replaying of deposits at startup

### LINKED ISSUE #5432: Avoid reprocessing deposits at startup

zilm13:
## PR Description
This is a very dirty first commit just to clarify that I'm on a right path to solve storing and loading `DepositTreeSnapshot`. There are no yet tests, I'm not sure the code is already working etc.
Because snapshot is not the only data we need from historical Eth1 blocks for Deposit production and validation, I store it along with `latestFinalizedDepositsBlock`  in `Eth1Cache`, because our db structure have only canonical block number key index for eth1 blocks deposit events. By this block I mean the last Eth1 block from which deposits were included on `BeaconChain` in `BeaconBlock` and finalized.

At the very beginning I clarify whether we have such cache and if the answer is yes, snapshot is used to populate tree in `DepositProvider` and `DepositStorage` replays only blocks after `latestFinalizedDepositsBlock`. I don't like it's separated into two not connected classes and think it should be reworked to be in sync, make sure we starting replaying blocks only after tree is initialized and don't miss anything.
Files to start looking at: `DepositStorage`, `DepositProvider`

Questions, questions:
1. Am I on the right path?
2. Is it ok `DepositTreeSnapshot` and `Eth1Cache` serializable entities are in `spec.datastructures`?  We already have something there which is not yet in spec and going to be included maybe one day, but `Eth1Cache` or whatever it would be renamed next is clearly our own structure while `DepositTreeSnapshot`could pave the way there.
3. Do we care about optimistic? As we are not rewinding finalized blocks, we shouldn't care I think.

## Fixed Issue(s)

Fixes #5432
## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
```diff
@@ -121,6 +132,19 @@ public void onNewFinalizedCheckpoint(
...
+            .ifPresent(
+                eth1DataEx -> {
+                  // TODO: log
+                  final StoreTransaction transaction = recentChainData.startStoreTransaction();
+                  transaction.setEth1Cache(
+                      Eth1Cache.SSZ_SCHEMA.create(
+                          depositMerkleTree.getSnapshot().asSerializable(),
+                          eth1DataEx.getBlockNumber()),
+                      fromOptimisticBlock);
+                  transaction.commit().join();
```
ajsutton, line 146: I'd be very keen to avoid the `.join()` here. I'm not entirely sure why the `eth1DataCache` is involved here at all though - it's quite separate to the deposit merkle tree and has a different life cycle for what it holds onto (based on timestamps).  There's no actual connection between the deposit we finalize which is tied to the state of the beacon chain and what we have in the eth1 data cache which is primarily tied to the time of eth1 blocks.

ethereum/pow/merkletree/src/main/java/tech/pegasys/teku/ethereum/pow/merkletree/DepositTreeSnapshot.java
```diff
@@ -31,6 +32,13 @@ public DepositTreeSnapshot(
     this.executionBlockHash = executionBlockHash;
   }
 
+  public DepositTreeSnapshot(
+      final tech.pegasys.teku.spec.datastructures.eth1.DepositTreeSnapshot depositTreeSnapshot) {
```
ajsutton, line 36: Ultimately we're going to want to be able to start from just a `tech.pegasys.teku.spec.datastructures.eth1.DepositTreeSnapshot` because we want to allow users to provide one of those and save us loading all deposits from the eth1 chain.  We can get the eth1 block number from the `executionBlockHash` included in the deposit snapshot. That will require a bunch more changes to how we start up the PowChainService and tell it where to start fetching deposits from - prior to this it's always just been given a block number, but it now needs to take a block hash and get the number from that. It may also have to deal with that block hash being unknown if the eth1 node changed and isn't in sync now too.

ethereum/pow/merkletree/src/main/java/tech/pegasys/teku/ethereum/pow/merkletree/DepositTreeSnapshot.java
```diff
@@ -43,6 +51,11 @@ public Bytes32 getExecutionBlockHash() {
     return executionBlockHash;
   }
 
+  public tech.pegasys.teku.spec.datastructures.eth1.DepositTreeSnapshot asSerializable() {
+    return tech.pegasys.teku.spec.datastructures.eth1.DepositTreeSnapshot.SSZ_SCHEMA.create(
+        finalized, UInt64.valueOf(deposits), executionBlockHash);
+  }
```
ajsutton, line 57: I don't think we want to create an SSZ version of this at this stage.  We can just write it out with a custom serialiser like we do with `DepositsFromBlockEventSerializer`.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
ajsutton: I also don't think we need to run this through the `StoreTransaction` - we can just send the merkle tree snapshot to the database directly and not wait for it to complete.  By the time we get a new finalized checkpoint here, the store updates have been persisted so we're safe to prune the merkle tree and send the new snapshot to disk.  If that new snapshot fails to write for some reason, that's not a problem either - we'll just use an older snapshot when we next start up and process more deposits than we need to.

ajsutton (commented):
> Am I on the right path?

Mostly I think. Probably gone off in the wrong direction a bit with trying to store the eth1 cache stuff and have a block number stored. We need to be able to start purely from a `DepositTreeSnapshot` so that should be the only new thing we need to store.

> Is it ok DepositTreeSnapshot and Eth1Cache serializable entities are in spec.datastructures? > We already have something there which is not yet in spec and going to be included maybe one day, but Eth1Cache or whatever it would be renamed next is clearly our own structure while DepositTreeSnapshotcould pave the way there.

I don't think we'll wind up needing those structures anyway. spec.datastructures probably isn't the ideal place for them - it's probably more part of storage:api and we'd likely have to invert how they're created (so you can pass the spec object to the serialisable object constructor rather than having a `.asSerializable()` method on the spec object) but generally if we're creating something just for serialisation it's better to just create a custom serialiser anyway.

> Do we care about optimistic? As we are not rewinding finalized blocks, we shouldn't care I think.

No need to worry about optimistic sync here. As you note if we find a finalized block is invalid that's a fatal error and teku will exit anyway because it can't undo finalisation.

zilm13:
I got a clear view how it should be done. I'd better close this one and create another PRs in the correct way

PR #5904, opened 2022-07-08
Introduce progressive balances

ajsutton:
## PR Description
Adds the ability to progressively update `TotalBalances` as part of amortising the cost of epoch transitions. Usage of the data is currently extremely limited (and toggled off by default) but it includes a checking mode which confirms the progressively calculated balances match the ones created when run in full allowing us to build up a lot more confidence in this new code path before we start depending on it.

Also fixes an issue in protoarray where only the justified and finalized epoch was stored instead of the full checkpoint - fixing that gets the last fork choice reference test passing which involves an exceptionally unlikely corner case. Nice to have it working though.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/status/TotalBalances.java
```diff
@@ -22,77 +24,168 @@ public class TotalBalances {
...
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("currentEpochActiveValidators", currentEpochActiveValidators)
+        .add("previousEpochActiveValidators", previousEpochActiveValidators)
+        .add("currentEpochSourceAttesters", currentEpochSourceAttesters)
+        .add("currentEpochTargetAttesters", currentEpochTargetAttesters)
+        .add("currentEpochHeadAttesters", currentEpochHeadAttesters)
+        .add("previousEpochSourceAttesters", previousEpochSourceAttesters)
+        .add("previousEpochTargetAttesters", previousEpochTargetAttesters)
+        .add("previousEpochHeadAttesters", previousEpochHeadAttesters)
```
rolfyone, line 188: did we intentionally leave out `effectiveBalanceIncrement` here?

ajsutton: No.  It's not a particularly important value but may as well add it for good measure - might save some head scratching one day when we somehow make it different between two instances.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/validatorInclusion/GetGlobalValidatorInclusionTest.java
```diff
@@ -67,13 +68,14 @@ void metadata_shouldHandle200() throws IOException {
...
+            UInt64.valueOf(4668326327938047084L),
             dataStructureUtil.randomUInt64(),
             dataStructureUtil.randomUInt64(),
-            dataStructureUtil.randomUInt64(),
-            dataStructureUtil.randomUInt64(),
-            dataStructureUtil.randomUInt64(),
-            dataStructureUtil.randomUInt64());
+            UInt64.valueOf(4658411424342975020L),
+            UInt64.valueOf(4663368873993027404L));
     GetGlobalValidatorInclusion.GetGlobalValidatorResponseData responseData =
```
tbenr: shouldn't we move constant values everywhere here and update the json accordingly?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/ProtoNodeData.java
```diff
@@ -102,6 +118,7 @@ public String toString() {
         .add("stateRoot", stateRoot)
         .add("executionBlockHash", executionBlockHash)
         .add("isOptimistic", isOptimistic)
+        .add("checkpoints", checkpoints)
```
tbenr, line 115: for consistency, we may want to use getters here too

tbenr (approved):
LGTM. Just minimal nits.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/validatorInclusion/GetGlobalValidatorInclusionTest.java
ajsutton: ah, those random values aren't in the JSON, but yeah makes sense to make them constants for consistency.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/ProtoNodeData.java
ajsutton: Went the other way and changed equals/hashCode to use direct variable access since we don't normally use getters for internal access (IntelliJ defaulted to generating with getters I think).

PR #5909, opened 2022-07-11
Fix config file for nested subcommands

ajsutton:
## PR Description
Subcommands that were deeply nested such as `teku debug-tools db` were not correctly applying values from the config file or environment.  Need to continue iterating down the `ParseResult` tree to find the actual subcommand that will be executed, not just take the first subcommand.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5919, opened 2022-07-12
Stop scheduling new keep alive comments when the client is closed

ajsutton:
## PR Description
When the SSE client closes, make sure we stop sending keep alive comments, otherwise we can wind up filling the execution queue for those keep alives and spend all our CPU on retrying.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5921, opened 2022-07-12
Add `ValidatorApiChannel` failover implementation

StefanBratanov:
## PR Description

In future PRs, this failover implementation will be used as part of the VC initialization when more than one BN endpoints are specified.

- Introduced` FailoverValidatorApiHandler` which implements `ValidatorApiChannel`
- Introduced `firstSuccess` in `SafeFuture`
- Introduced `RemoteValidatorApiHandler` in order to expose the endpoint as part of the interface implementations

## Fixed Issue(s)
will help for #5237 

## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
+++ b/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
@@ -238,36 +238,26 @@ public class SafeFuture<T> extends CompletableFuture<T> {
   /**
    * Returns a new {@link SafeFuture} with the result of the first successful future from the given
    * futures. If all futures complete exceptionally, the returned {@link SafeFuture} will complete
-   * exceptionally with the exception from the first future. If the provided {@link List} of futures
-   * is an empty collection, the returned {@link SafeFuture} will complete exceptionally with an
-   * {@link IllegalArgumentException} exception.
+   * exceptionally with the exception from the {@link #allOf(SafeFuture[])} call. If the provided
+   * {@link List} of futures is an empty collection, the returned {@link SafeFuture} will complete
+   * exceptionally with an {@link IllegalArgumentException} exception.
    *
-   * @param futures the futures which will need to be completed
+   * @param futures a List of futures that need to be completed
    * @return a new {@link SafeFuture} that is completed with the first successful result from the
    *     given futures or completed exceptionally if all futures complete exceptionally
    */
   public static <T> SafeFuture<T> firstSuccess(final List<SafeFuture<T>> futures) {
     if (futures.isEmpty()) {
       return SafeFuture.failedFuture(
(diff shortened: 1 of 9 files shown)


infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
```diff
@@ -235,6 +235,41 @@ public static <U> SafeFuture<Void> allOfFailFast(final SafeFuture<?>... futures)
...
+                      return !canProceed;
+                    }))
+        .thenCompose(
+            __ ->
+                futures.stream()
+                    .filter(SafeFuture::isCompletedNormally)
+                    .findFirst()
+                    // at this point, all futures have completed exceptionally, so just
+                    // return the first error future
+                    .orElse(futures.get(0)));
```
ajsutton: We need to add tests for this method. :)

I don't think `asyncDoWhile` is going to do what you expect here - I think this will wind up in a tight loop continuously executing while the futures execute instead of actually waiting for the next future to complete before looping again because the `SafeFuture.of` here is immediately returning. 

I think you can simplify it to just:
```java
    final SafeFuture<T> result = new SafeFuture<>();
    allOf(
            futures.stream()
                .map(future -> future.thenAccept(result::complete))
                .toArray(SafeFuture[]::new))
        .propagateExceptionTo(result);
    return result;
```

which takes advantage of the fact that you can only complete a future once so for each future we tell it to pass it's successful result to the returned future as soon as it finishes, then once all the futures are complete, propagate any exception. Whatever completes the `result` future first is the result that will actually be used and everything after that is just ignored.

There will likely be a warning about unchecked types to suppress because the `SafeFuture[]::new` but that's fine.

ajsutton (commented):
Looks good, just need some tests around the new SafeFuture method and make sure it really is doing what we need.

infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/SafeFutureTest.java
```diff
@@ -1077,6 +1077,44 @@ public void getImmediately_notDone() {
...
+            List.of(
+                SafeFuture.failedFuture(exception),
+                SafeFuture.failedFuture(exception),
+                SafeFuture.failedFuture(exception)));
+
+    SafeFutureAssert.assertThatSafeFuture(result)
+        .isCompletedExceptionallyWith(IllegalStateException.class)
+        .hasMessage("oopsy");
+  }
+
```
tbenr, line 1129: I would add a test passing a non completed future in the list.
It will check that we are not waiting all of them to complete.

StefanBratanov: good point, done

tbenr (approved):
LGTM

PR #5922, opened 2022-07-13
Alter AbstractDatabaseTest to use TestTemplates.

rolfyone:
- removed concrete classes, and altered name to DatabaseTest.
- introduced the blockStorage class, but only full_block is supported until it goes into the separation PR.

Partially addresses #5312

## Documentation

## Changelog


storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/DatabaseTest.java
```diff
@@ -1926,16 +1994,36 @@ private UpdatableStore recreateStore() {
   }
 
   private void createStorageSystemInternal(
```
ajsutton: I'd be very tempted to essentially move this method into `DatabaseContext`. You'd have to pass in the tmpDir and `spec` I think but that's easy enough and not an issue to create a tmp dir even if it doesn't wind up being needed. Would just make the `DatabaseContext` more of a standalone provider rather than just a thin wrapper around a couple of config values.

storage/src/testFixtures/java/tech/pegasys/teku/storage/server/TestDatabaseInvocationContextProvider.java
```diff
@@ -0,0 +1,86 @@
...
+
+    configurations.addAll(
+        databaseVersions.stream()
+            .flatMap(
+                databaseVersion ->
+                    blockStorageSet.stream()
+                        .map(
+                            blockStorageMode ->
+                                generateContext(databaseVersion, blockStorageMode, true)))
+            .collect(Collectors.toList()));
```
ajsutton: nit: can change the `.collect(toList())` to just `.forEach(configurations::add)` and remove the `addAll` call.

ajsutton (approved):
LGTM. Just a couple of suggestions.

PR #5924, opened 2022-07-13
add a flag that enables users to hardcode the timestamp in the Valida…

### LINKED ISSUE #5898: Builder API Compatability with Distributed Validators - Validator Registration
### Distributed Validators Require Signed Contents To Be The Same - ValidatorRegistrationV1

The current implementation of the validator registration isn't compatible with a distributed validator

For distributed validators it's important that the signed contents are the same across all "co-validators"  

Teku currently uses timeProvider.getTimeInSeconds() for the Timestamp in ValidatorRegistrationV1, time.now in milliseconds are likely to never match

- The simplest solution would be an --X param that effectively hardcodes the timestamp, this could be done via an actual timestamp, a slot number or an epoch number, in this situation it's up to the co-validators to co-ordinate to ensure they all have the same config set for "timestamp"
- Since Teku uses the first signature (with the hardcoded timestamp) for any subsequent re-registrations the distributed validator will simply need to update their registration config which would include timestamp to register with new information on a relay

see here for discussion in discord https://discord.com/channels/697535391594446898/697539289042649190/994236320799002705

cc @tbenr

ciaranmcveigh5:
## PR Description

Add a flag that enables users to hardcode the timestamp in the ValidatorRegistrationV1 schema, this is required for distributed validators to come to consensus, the current implementation uses the current timestamp which will be different on each co-validator, we need the timestamp to be the same on all validators

## Fixed Issue(s)

fixes #5898 

## Documentation

- Potentially add a paragraph to this document https://hackmd.io/@StefanBratanov/BkMlo1RO9


ciaranmcveigh5:
Need to add a test case for createValidatorRegistration when validatorsRegistrationDistributedValidatorTimestamp is set to verify the ValidatorRegistrationV1 object has its timestamp set to the hardcoded value in validatorsRegistrationDistributedValidatorTimestamp & a test to verify it works as normal when validatorsRegistrationDistributedValidatorTimestamp is not set

Also can change variable naming, description etc. if needed

tbenr:
Thank you very much @ciaranmcveigh5  to bring this up.
I've taken the liberty to complete your work with tests and fixing some little things here and there.

ajsutton (approved):
LGTM. Just need to think about what the right name for this option is. While it is being added to support distributed validators it's not necessarily specific to that so maybe makes sense to make the name more about what it's doing rather than why it was first added.  Will leave that call up to you though @tbenr as you're more up to date on what the validator proposer options all are.

tbenr:
@ciaranmcveigh5 I changed the param name to `--Xvalidators-registration-timestamp-override`

PR #5932, opened 2022-07-14
Return a Bad Gateway when an error occurs calling builder registerValidator while serving BN validator registration API

### LINKED ISSUE #5880: [MEV Boost\Builder] `/eth/v1/validator/register_validator` beacon node API returns 400 when downstream builder API fails with 502
When running standalone validator, when validator registration call fails with 502 on the builder endpoint, the validator client that has initiated the registration receives a 400.

trace from VC side:
```
Server error from Beacon Node API (url = http://127.0.0.1:5051/eth/v1/validator/register_validator, status = 400, message = Invalid remote response: {“code”:502,“message”:“no successful relay response”}
```

tbenr:
During validator registration process in which the BN api forwards the request to the builder interface via our RestClient.
When unwrapping the message we were using `checkArgument` which throws an `IllegalArgumentException` which triggers the bad request ExceptionHandler at the BN api level.
~~This is a quick fix that will ensure that those kind of errors will generate a 500. I was thinking about issuing a `ServiceUnavailableException` that would have been translated to a 503, but I don't feel it would have been the right direction.~~

Moved toward intercepting the builder api exception and returning a bad gateway which fits better IMO.

fixes #5880

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostRegisterValidator.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostRegisterValidator.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.beaconrestapi.handlers.v1.validator;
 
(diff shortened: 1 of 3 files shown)


StefanBratanov (approved):
LGTM

StefanBratanov:
LGTM

PR #5935, opened 2022-07-14
builder fallback improvements

tbenr:
Three new fallback reasons added

`NOT_NEEDED`: when builder is not configured and validator is not registered, which covers the configuration in which blinded flow is active but builder endpoint and validator registration are not configured. The fallback log in this case is at `DEBUG` level

`TRANSITION_NOT_FINALIZED`: substitute `FORCED`

`BUILDER_NOT_CONFIGURED`: indicates that the validator was registered but no builder endpoint was configured.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #5936, opened 2022-07-14
Add syncing status method to `RemoteValidatorApiChannel`

StefanBratanov:
## PR Description

- Add `/eth/v1/node/syncing` get method to `RemoteValidatorApiChannel`
- Create a `SyncingStatus` model object.
- Remove two unused classes

## Fixed Issue(s)
will help for https://github.com/ConsenSys/teku/issues/5237

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -159,6 +160,11 @@ public ValidatorApiHandler(
     this.proposersDataManager = proposersDataManager;
   }
 
+  @Override
+  public SafeFuture<Optional<SyncingStatus>> getSyncingStatus() {
+    return SafeFuture.failedFuture(new UnsupportedOperationException("Not yet implemented"));
```
ajsutton: Typically we want the implementation here to be the canonical one and then the REST API basically just delegates to this. That way the remote and internal VCs are getting the same behaviour because it winds up at the same code.

So in this case we'd ideally want to take the business logic that's currently used for the `GetSyncing` API and put it here then `GetSyncing` would delegate to this. I'm not sure what that will wind up looking like so maybe it will work out better in this case to have this delegate to the same thing `GetSyncing` uses but I suspect that means `ValidatorApiHandler` which is a fairly internal kind of class winds up needing to have access to more REST API specific classes which isn't great.

You'll need to play around and see what you think works best. :)

ajsutton: Actually the other option is that we don't add this method `ValidatorApiHandler` at all and just make it available only on `RemoteValidatorApiHandler` or something like that. If it's only useful for remote validators it would be nice to avoid needing to add it here. Again, depends on the trade offs involved.

validator/api/src/main/java/tech/pegasys/teku/validator/api/SyncingStatus.java
```diff
@@ -0,0 +1,148 @@
...
+              SyncingStatus::getSyncDistance,
+              Builder::syncDistance)
+          .withField(
+              "is_syncing", CoreTypes.BOOLEAN_TYPE, SyncingStatus::isSyncing, Builder::isSyncing)
+          .withOptionalField(
+              "is_optimistic",
+              CoreTypes.BOOLEAN_TYPE,
+              SyncingStatus::getIsOptimistic,
+              Builder::isOptimistic)
+          .build();
```
ajsutton: I don't think this type definition belongs here. This is an internal class but the JSON serialisation is part of the REST API so should live with that.  We have things like `ValidatorTypes` and `BeaconRestApiTypes` to put shared JsonTypeDefinitions so they don't need to pollute the internal objects.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
```diff
@@ -78,6 +79,11 @@ public FailoverValidatorApiHandler(
     this.validatorLogger = validatorLogger;
   }
 
+  @Override
+  public SafeFuture<Optional<SyncingStatus>> getSyncingStatus() {
+    return tryRequestUntilSuccess(ValidatorApiChannel::getSyncingStatus);
+  }
```
ajsutton: This seems suspicious - wouldn't it be important to know which node wound up returning the syncing status? This may be another indication that we don't actually want to add this method to the `ValidatorApiHandler` interface.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
StefanBratanov: Definitely agree, that will be cleaner and makes more sense, since it will only be used there.

validator/api/src/main/java/tech/pegasys/teku/validator/api/SyncingStatus.java
StefanBratanov: Noted

tbenr (approved):
LGTM.
What makes me think is the `required` package. I don't have clear idea where to put those classes.

PR #5938, opened 2022-07-15
allows to specify terminal block hash for ExecutionLayerChannelStub

tbenr:
Improves the stub to allow syncing an existing net through the merge without the need of a EL.

It means that it is possible to fully sync a BN to ie sepolia by simply running:

`--network sepolia --ee-endpoint stub:0xd07cce9785d39c0dd2409b7d8e69d6bff26a69a0fa5308ac781c63ffe2a37bc1`

It think it's handy for testing for  purpose.


StefanBratanov (approved):
LVGTM

tbenr:
for the records I decided to include the option in the `stub` and not introduce a new `--X` param just to maintain the stub functionality as hidden as possible.

PR #5939, opened 2022-07-15
Fix blinded block signing for `ExternalSignerBlockRequestProvider`

tbenr:
When external signer is configured and blinded block flow is enabled, we were failing preparing the signature request to the external signer.

```
16:14:24.486 ERROR - Validator   *** Failed to produce block  Slot: ... Validator: ... java.lang.IllegalArgumentException: Expected bellatrix block body but got BlindedBeaconBlockBodyBellatrixImpl (See log file for full stack trace)
```

This is a quick fix.
`SchemaObjectProvider` might be improved to seamlessly handle blinded blocks to avoid spaghettification.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #5940, opened 2022-07-15
Migrate hot blocks if using blinded storage

rolfyone:
As a starting point for blinded storage migration

 - migrate blinded blocks so that we can start up if we were previously using full hot block storage

 - migrate the first block (lowest slot) from finalized storage - this will ensure back filling happens from the right spot.

A future PR will cater for migrating finalized storage and non canonical block storage.

Blinded block storage is still off by default, so this migration is only performed if the feature toggle is on. It's worth noting that once you've enabled that toggle, you won't be able to go back to using the old storage, as your hotblock storage will have moved and teku won't start without hotblocks to build the protoarray.

Partially addresses #5312

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/Database.java
```diff
@@ -105,6 +105,8 @@ public interface Database extends AutoCloseable {
   @MustBeClosed
   Stream<SignedBeaconBlock> streamHotBlocks();
 
+  long countFinalizedBlocks();
```
rolfyone: Added this because the count using the old stream was very slow... I was running it at the start of migration, which I think is probably something to keep, because it lets you track percent complete...  Possibly it'd be worth naming it `countFinalizedUnblindedBlocks`, as it's actually checking old storage columns...

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedHotBlockMigration.java
```diff
@@ -0,0 +1,98 @@
...
+  static <T extends KvStoreCombinedDaoBlinded & KvStoreCombinedDaoUnblinded> void migrateBlocks(
+      final T dao, final Spec spec) {
+    final BlindedHotBlockMigration<T> blindedHotBlockMigration =
+        new BlindedHotBlockMigration<>(spec, dao);
+    blindedHotBlockMigration.performBatchMigration();
+  }
+
+  private void performBatchMigration() {
+    moveHotBlocksToBlindedStorage();
+    moveFirstFinalizedBlockToBlindedStorage();
```
rolfyone: this will allow backfill to be done from the correct point at a very low cost, and will be simpler than needing to check old and new storage.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
```diff
@@ -207,10 +207,7 @@ public int getBlockCounts(
       try (Stream<SignedBeaconBlock> stream = database.streamHotBlocks()) {
         printIfPresent("Hot blocks", stream.count());
       }
-      try (Stream<SignedBeaconBlock> stream =
-          database.streamFinalizedBlocks(UInt64.ZERO, UInt64.MAX_VALUE)) {
-        printIfPresent("Finalized blocks", stream.count());
-      }
+      printIfPresent("Finalized blocks", database.countUnblindedFinalizedBlocks());
```
rolfyone, line 210: this count got added because it is significantly faster than streaming actual signed blocks.. the count uses raw streams, so that has been kept low level.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedHotBlockMigration.java
```diff
@@ -0,0 +1,98 @@
...
+    final BlindedHotBlockMigration<T> blindedHotBlockMigration =
+        new BlindedHotBlockMigration<>(spec, dao);
+    blindedHotBlockMigration.performBatchMigration();
+  }
+
+  private void performBatchMigration() {
+    moveHotBlocksToBlindedStorage();
+  }
+
+  private void moveFirstFinalizedBlockToBlindedStorage() {
```
ajsutton: very small nit: I'd move this method to after `moveHotBlocksToBlindedStorage` so the order of code in the file matches the order of execution.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedHotBlockMigration.java
```diff
@@ -0,0 +1,98 @@
...
+      for (Iterator<SignedBeaconBlock> it = blocks.iterator(); it.hasNext(); ) {
+        try (KvStoreCombinedDaoBlinded.FinalizedUpdaterBlinded blindedUpdater =
+                dao.finalizedUpdaterBlinded();
+            KvStoreCombinedDaoUnblinded.HotUpdaterUnblinded unblindedUpdater =
+                dao.hotUpdaterUnblinded()) {
+          for (int i = 0; i < BATCH_SIZE && it.hasNext(); i++) {
+            final SignedBeaconBlock block = it.next();
+            blindedUpdater.addBlindedBlock(block, block.getRoot(), spec);
+            unblindedUpdater.deleteUnblindedHotBlockOnly(block.getRoot());
+            counter++;
```
ajsutton, line 69: Just in case this winds up happening in a long period of non-finality - it may be worth printing a progress message every 1000 blocks or something like that. Will almost certainly not be needed but will be a lot better than a very long period of silence if we ever do hit this case.

ajsutton (approved):
LGTM.

Issue #5914, opened 2022-07-11
Update execution payload re-execution strategy


tbenr:
We currently have two PRs (#5840 and #5868) insisting on the same `ReexecutingExecutionPayloadBlockManager` class which handles the logic to re-execute a payload when it fails due to low level\networking exceptions, with the aim of being more resilient to short term outages.

## Current behaviour
We re-execute failing blocks every 2 seconds. We expire the re-executing when block's slot become more than 2 slot older than the current slot. It means it will re-executend for current slot and for the next 2 slots.

**Summary: blocks that arrives in time will be re-executed every 2s for 36s**

Existing bug: There is an unintended behaviour in which we are queuing for re-execution an already expired block (so it may be executed if the pruning happening `onSlot` doesn't kick-in before the 2s tick). This may happen if we hit a timeout on EL, when we may end up by re-executing it several time.

## Intents of the current PRs

#5840 wants to solve a stall problem (#5824, syncing doesn't kick-in quickly enough) by continue re-executing a block every time a descendant of that block is received. The goal is to extend the time window in which we re-execute it to be able to recover without start syncing.

#5868 wants to avoid too many re-execution by reducing slot expiry from 2 to 1 (24s likely re-execution time, 12s guaranteed)
The idea is fail fast and relay on sync process to kick-in for recovery. It also solves the bug mentioned above.

## How to proceed

Solve the mentioned bug in a dedicated pr:
- [x] #5915

We should then discuss here what are the general re-execution\retry behaviour we want to have on:

BlockImport via Gossip (currently implemented in `ReexecutingExecutionPayloadBlockManager` wrapping `BlockManager`)
BlockImport via BatchSync (nothing implemented yet)

Then we could implement a re-execution system managing both cases, or split them in two classes if the logic differs too much.

Optionally:
Improve the way we engage syncing by being more reactive (if actually needed)

ajsutton:
Going to call this done for now.  Batch sync will automatically retry already and is only syncing a single chain so unlikely to swamp the EL.  Gossip is now handled simply but sensibly and shouldn't cause issues for the EL, while the new approach has fixed problems seen in hive tests and shown that Teku doesn't retry correctly and then carry on following the chain.

PR #5941, opened 2022-07-18
Improve execution payload retrying

### LINKED ISSUE #5915: Avoid unintended reexecution of expired blocks in `ReexecutingExecutionPayloadBlockManager`
As mentioned here #5914, #5868 solves a bug that worth to be fixed while clearing the re-execution logic.

the bug:
There is an unintended behaviour in which we are queuing for re-execution an already expired block (so it may be executed if the pruning happening `onSlot` doesn't kick-in before the 2s tick). This may happen if we hit a timeout on EL, when we may end up by re-executing it several time.

the fix is here: 
https://github.com/ConsenSys/teku/pull/5868/files#diff-23d1694769644ca3b78e21c2aa378b8371cec7091d1643fce3a2ba32f2168ca8R102-R119

ajsutton:
## PR Description
Replaces `ReexecutingExecutionPayloadBlockManager` with `FailedExecutionPool`. This simplifies the design a bit by using delegation instead of inheritance but also changes the algorithm for retrying:

 * Delay between retries now uses exponential back-off up to a maximum of 30 seconds
 * Only a single block is re-executed at a time instead of all pending blocks
 * The same block is retried on timeout (assuming the EL will complete execution of that payload even though the call timed out and then be able respond from a cached result)
 * A different block is retried on error or syncing to give the best chance of finding a fork we can continue to follow and not get stuck on a side-fork that doesn't have EL data available or that the EL is consistently hitting an error when importing

## Fixed Issue(s)
fixes #5915
at least a start of #5914 and possibly enough...

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  private boolean isTimeout(final BlockImportResult importResult) {
+    return importResult
+        .getFailureCause()
+        .map(
+            error ->
+                ExceptionUtil.hasCause(error, TimeoutException.class)
+                    || ExceptionUtil.hasCause(error, SocketTimeoutException.class))
```
tbenr: By checking timeouts befaviour of the underlining OkHttp I see there is another potential exception we can receive `InterruptedIOException` (https://www.baeldung.com/okhttp-timeouts#call) 
We should not hit that because I don't think the Web3J is setting the call timeout. (In our custom restClient for builder endpoint, we do).
So for completeness I'd suggest tu add it. Might be good to have a `hasCause` accepting a `varargs` of classes and have the OR there.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
```diff
@@ -0,0 +1,121 @@
...
+import tech.pegasys.teku.infrastructure.logging.LogFormatter;
+import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
+import tech.pegasys.teku.spec.logic.common.statetransition.results.BlockImportResult;
+import tech.pegasys.teku.storage.server.ShuttingDownException;
+
+public class FailedExecutionPool {
+  private static final Logger LOG = LogManager.getLogger();
+  static final Duration MAX_RETRY_DELAY = Duration.ofSeconds(30);
+  static final Duration SHORT_DELAY = Duration.ofSeconds(2);
+  private final Queue<SignedBeaconBlock> awaitingExecution = new ArrayBlockingQueue<>(10);
```
tbenr: nit: I feel like naming it `awaitingExecutionQueue` helps the reading

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
```diff
@@ -74,8 +74,7 @@ public SafeFuture<BatchImportResult> importBatch(final Batch batch) {
               lastBlockImportResult -> {
                 if (lastBlockImportResult.isSuccessful()) {
                   return BatchImportResult.IMPORTED_ALL_BLOCKS;
-                } else if (lastBlockImportResult.getFailureReason()
-                    == BlockImportResult.FailureReason.FAILED_EXECUTION_PAYLOAD_EXECUTION) {
+                } else if (lastBlockImportResult.hasFailedExecutingExecutionPayload()) {
```
tbenr, line 77: So we do the batch retry (in 5s) also in when we receive SYNCING but we aren't in a position to optimistically import. There might be some misleading logs, though.
It also means that the batch retry will overlap with the `FailedExecutionPool` retry.


I also think that `hasFailedExecutingExecutionPayload()` is more like `hasNotExecutedExecutionPayload()`

tbenr (approved):
LGTM. With a nit and a reflection.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
```diff
@@ -0,0 +1,121 @@
...
+        currentDelay = MAX_RETRY_DELAY;
+      }
+      if (awaitingExecution.isEmpty() || isTimeout(importResult)) {
+        scheduleNextRetry(block);
+      } else {
+        // Try a different block
+        final SignedBeaconBlock nextBlock = awaitingExecution.remove();
+        awaitingExecution.add(block);
+        retryingBlock = Optional.of(nextBlock);
+        scheduleNextRetry(nextBlock);
```
zilm13: Maybe modify interface  to `scheduleNextRetry()` (no param) and get `retryingBlock` on the first line of the method to reduce complexity and avoid extra mistake possibility

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
zilm13: If I correctly understand it could be also websockets and ipc there. For Websockets it looks like timeout will be `IOException`. Anyway, this method is detached and could be improved later

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private synchronized void handleExecutionResult(
+      final SignedBeaconBlock block, final BlockImportResult importResult) {
+    if (importResult.hasFailedExecutingExecutionPayload()) {
+      currentDelay = currentDelay.multipliedBy(2);
+      if (currentDelay.compareTo(MAX_RETRY_DELAY) > 0) {
+        currentDelay = MAX_RETRY_DELAY;
+      }
+      if (awaitingExecution.isEmpty() || isTimeout(importResult)) {
+        scheduleNextRetry(block);
```
zilm13: So if we have block with something like resource exhaustion attack or any other kind of failure which looks like timeout, we will continue DoS our EL without any chance to swap payload. Maybe it's better to not handle timeout especially for safety and simplicity

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
ajsutton: hmm, interesting. The exception I got from OkHttp was an `InterruptedIOException` with a `SocketTimeoutException` as the cause.  Probably reasonable to look for InterruptedIOException directly though.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
ajsutton: Interestingly the `BatchImporter` never goes through the `BlockManager` so when things fail here they never wind up in the `FailedExecutionPool` at all and it's just the batch sync process that does the retry (which appears to work very well).

In terms of naming I think `hasNotExecutedExecutionPayload` sounds more like a `SYNCING` response where we'd optimistically import.  It's a little bit of a stretch that a SYNCING responses is a failure in the period we aren't allowed to optimistically sync but it's the best name I can think of still...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
ajsutton: Good idea.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/FailedExecutionPool.java
ajsutton: I can see the logic there, but that's not really how it works out.  The EL won't stop executing a block when the CL times out and ELs can generally only import one block at a time so if we try to execute a different block it will just be queued behind the first one and also timeout. Whereas if we execute the same block it can be easily deduplicated.  Building up a queue of other blocks to try increases memory usage and makes it even harder for the EL to continue.  Plus by retrying the same block we're most likely to get a cached response once the block finally does finish executing. We should be able to get that cached response even if we wind up executing another block in the mean time but there will be limits to how many import results fit in a cache and how long they're held for.

zilm13: Looking at the Geth code it looks like it will not forget some payload check result even if we query something after (and it sounds reasonable, we rarely make cache for 1 result in development). My point was about case of self-recovery , when client has internal timeout for execution (though Geth code doesn't contain it) or external self-recovery which will kill and restart app without signs of life. In case there exists very bad payload with looong execution we could reproduce EL torture with it again and again. But it will be definitely a network issue at this point, not only for Teku.

ajsutton: But the EL can't really time out or it could be rejecting an invalid block, so it has to keep processing until the block is done. We can't be sure if it's still processing or not (the connection may have dropped without sending a proper FIN or something like that) but if the EL is alive it will have to keep processing until it completes the block.  And if it does time out it should remember it as an invalid block (otherwise in PoW it would execute it again when it next sees it on gossip).

So we can't make it any worse, but by retrying the same block we open up some simple optimisations for the EL to avoid building up a list of blocks to execute next.

And as you say, if its feasible to create a block that takes that long to execute then the entire network will be DoSed anyway.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
tbenr: Nice, I'm still waiting the day in which I stop confusing `BlockManager` and `BlockImporter`

PR #5942, opened 2022-07-18
Failover for BN Event Stream from VC

StefanBratanov:
## PR Description

- Move `RemoteValidatorApiChannel` to the remote module and change the endpoint return type to be `HttpUrl`
- Change `EventSourceBeaconChainEventAdapter` to failover with the help of `getSyncingStatus` method
- Removed authentication, since an Auth interceptor will already be configured in the `OkHttpClient`

##### 1 VC and 2 Beacon Nodes

- Stop primary BN -> VC failovers to the other BN if the other BN is synced, otherwise it keep retrying against the primary BN
- Start primary BN -> VC goes back to the primary BN after sync is complete
- Stop both BN when current event stream BN is the primary one -> VC continues to fail and retries against the primary BN, while at the same time tries to connect to the any of the failover BN
- Stop both BN when current event stream BN is the failover one -> VC continues to fail and retries against the failover BN, while at the same time tries to connect to the primary BN

##### 1 VC and 3 Beacon Nodes
 - Stop primary BN -> VC failovers to the first BN which succeeds a syncing status check
 - Stop one of the failover BN -> VC failovers to the last remaining BN if syncing status check succeeds, otherwise it keeps retrying to connect to the current failover, while also trying to connect to the primary BN

## Fixed Issue(s)
will help for #5237 

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
@@ -59,7 +59,7 @@ public class RemoteValidatorAcceptanceTest extends AcceptanceTestBase {
   void shouldCreateAttestationsWithRemoteValidatorStartingFirst() throws Exception {
     validatorClient.start();
     validatorClient.waitForLogMessageContaining(
-        "Error while connecting to Beacon Node event stream.");
+        "Error while connecting to Beacon Node event stream");
 
     beaconNode.start();
 
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
@@ -59,7 +59,7 @@ public class RemoteValidatorAcceptanceTest extends AcceptanceTestBase {
   void shouldCreateAttestationsWithRemoteValidatorStartingFirst() throws Exception {
     validatorClient.start();
     validatorClient.waitForLogMessageContaining(
-        "Error while connecting to Beacon Node event stream");
+        "Error while connecting to beacon node event stream");
 
     beaconNode.start();
 
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
@@ -59,7 +59,7 @@ public class RemoteValidatorAcceptanceTest extends AcceptanceTestBase {
   void shouldCreateAttestationsWithRemoteValidatorStartingFirst() throws Exception {
     validatorClient.start();
     validatorClient.waitForLogMessageContaining(
-        "Error while connecting to beacon node event stream");
+        "Error while connecting to Beacon Node event stream.");
 
     beaconNode.start();
 
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
@@ -125,16 +125,15 @@ public class EventSourceBeaconChainEventAdapter implements BeaconChainEventAdapt
     return Preconditions.checkNotNull(eventSourceUrl);
   }
 
-  // connect to a failover event stream only if there are failovers configured, the syncing status
-  // call throws an exception and there is a ready failover Beacon Node
+  // connect to a failover event stream only if there are failovers configured and there is a ready
+  // failover Beacon Node
   private void switchToFailoverEventStreamIfNeeded(
       final RemoteValidatorApiChannel currentBeaconNodeApi) {
     if (failoverBeaconNodeApis.isEmpty()) {
       return;
     }
     final HttpUrl currentEndpoint = currentBeaconNodeApi.getEndpoint();
-    currentBeaconNodeApi
-        .getSyncingStatus()
+    checkBeaconNodeIsReadyForEventStreaming(currentBeaconNodeApi)
         .finish(__ -> findReadyFailoverAndSwitch(currentEndpoint));
   }
 
@@ -156,7 +155,7 @@ public class EventSourceBeaconChainEventAdapter implements BeaconChainEventAdapt
   // this method for the same failover endpoint because of the ConnectionErrorHandler callback
   private synchronized void switchToFailoverEventStream(
       final RemoteValidatorApiChannel beaconNodeApi) {
-    if (alreadyFailovered(beaconNodeApi)) {
+    if (alreadyFailedOver(beaconNodeApi)) {
       return;
     }
     final HttpUrl failoverEndpoint = beaconNodeApi.getEndpoint();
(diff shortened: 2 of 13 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
```diff
@@ -96,6 +125,112 @@ private HttpUrl createHeadEventSourceUrl(final HttpUrl endpoint) {
...
+        .map(EventSource::getHttpUrl)
+        .map(endpoint -> endpoint.equals(createHeadEventSourceUrl(beaconNodeApi.getEndpoint())))
+        .orElse(false);
+  }
+
+  private void startPrimaryBeaconNodeCheckStatusTask() {
+    final Cancellable currentCheckStatusTask = primaryBeaconNodeCheckStatusTask.get();
+    if (currentCheckStatusTask != null && !currentCheckStatusTask.isCancelled()) {
+      return;
+    }
```
tbenr: I'd extract this to a dedicated method like `switchToPrimaryEventStream` pairing nicely with the existing `switchToFailoverEventStream`

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteBeaconNodeApi.java
```diff
@@ -79,6 +78,13 @@ public static BeaconNodeApi create(
     final RemoteValidatorApiChannel primaryValidatorApi =
         createRemoteValidatorApi(
             primaryEndpoint, okHttpClient, spec, preferSszBlockEncoding, asyncRunner);
+    final List<RemoteValidatorApiChannel> failoversValidatorApis =
```
tbenr: Not a fan of `failovers`. Maybe `failoverValidatorApis` would be enough? But don't feel fully entitled to judge :)

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
StefanBratanov: Good one, done!

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteBeaconNodeApi.java
StefanBratanov: I agree, actually went back and forth with this.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
```diff
@@ -96,6 +125,113 @@ private HttpUrl createHeadEventSourceUrl(final HttpUrl endpoint) {
     return Preconditions.checkNotNull(eventSourceUrl);
   }
 
+  // connect to a failover event stream only if syncing status call throws
+  // an exception and there is a ready failover Beacon Node
+  @SuppressWarnings({"FutureReturnValueIgnored"})
```
tbenr: Played a bit to avoid the suppress
image
Could be better if it's equivalent :)

StefanBratanov: That looks better and more readable, implemented.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
```diff
@@ -59,7 +59,7 @@ void shouldCreateAttestationsWithRemoteValidator() throws Exception {
   void shouldCreateAttestationsWithRemoteValidatorStartingFirst() throws Exception {
     validatorClient.start();
     validatorClient.waitForLogMessageContaining(
-        "Error while connecting to beacon node event stream");
+        "Error while connecting to Beacon Node event stream");
```
rolfyone: we just use lower case everywhere, I'd suggest leaving it lower case in this PR, because otherwise its inconsistent with all the other messages...

rolfyone: seems like a couple have snuck in with capitals, but I think it's easier to fix those than add more :)

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
```diff
@@ -96,6 +125,112 @@ private HttpUrl createHeadEventSourceUrl(final HttpUrl endpoint) {
...
+            __ ->
+                LOG.warn(
+                    "There are no Beacon Nodes from the configured ones that are ready to be used as an event stream failover"));
+  }
+
+  // switchToFailoverEventStreamIfNeeded is async, so there could be multiple quick calls to
+  // this method for the same failover endpoint because of the ConnectionErrorHandler callback
+  private synchronized void switchToFailoverEventStream(
+      final RemoteValidatorApiChannel beaconNodeApi) {
+    if (alreadyFailovered(beaconNodeApi)) {
```
rolfyone: probably `alreadyFailedOver`

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
```diff
@@ -96,6 +125,112 @@ private HttpUrl createHeadEventSourceUrl(final HttpUrl endpoint) {
...
+  // connect to a failover event stream only if there are failovers configured, the syncing status
+  // call throws an exception and there is a ready failover Beacon Node
+  private void switchToFailoverEventStreamIfNeeded(
+      final RemoteValidatorApiChannel currentBeaconNodeApi) {
+    if (failoverBeaconNodeApis.isEmpty()) {
+      return;
+    }
+    final HttpUrl currentEndpoint = currentBeaconNodeApi.getEndpoint();
+    currentBeaconNodeApi
+        .getSyncingStatus()
```
ajsutton: I feel like we should fail over if the node reports it is syncing, not just if the call fails.

ajsutton (approved):
LGTM. I'd agree with Paul about using lower case beacon node. :)

I think this is worth merging as-is, but I do wonder if the design would come out cleaner if we had a class that encapsulated a single RemoveValidatorApiChannel, the tracking for if it was in sync or not and the ability to start/stop its event stream.  This is basically the model we use for eth1 endpoints so we're constantly keeping track of which endpoints are in a usable state and then just switching to the one we need. The top level event adapter class would then just find the first endpoint in the list that is considered ready and make sure it's event stream is started.  May not work out as well as I'm thinking but possibly worth playing around with to try and simplify the code.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/EventSourceBeaconChainEventAdapter.java
StefanBratanov: yeah that makes sense especially since we will come back to the primary one once it is in sync anyways after some time

StefanBratanov:
> LGTM. I'd agree with Paul about using lower case beacon node. :)
> 
> I think this is worth merging as-is, but I do wonder if the design would come out cleaner if we had a class that encapsulated a single RemoveValidatorApiChannel, the tracking for if it was in sync or not and the ability to start/stop its event stream. This is basically the model we use for eth1 endpoints so we're constantly keeping track of which endpoints are in a usable state and then just switching to the one we need. The top level event adapter class would then just find the first endpoint in the list that is considered ready and make sure it's event stream is started. May not work out as well as I'm thinking but possibly worth playing around with to try and simplify the code.

Decided to merge as it is. Tried a different implementation, but didn't like it a lot. I got worried also that by keeping track of the beacon node state in a scheduled process, we wouldn't react to issues as fast as using the `ConnectionErrorHandler` unless we use a callback but then i think it would make the code more difficult.

PR #5945, opened 2022-07-19
Start up ReconstructHistoricalStatesService

courtneyeh:
#5749

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBlockSyncService.java
```diff
@@ -137,6 +145,9 @@ protected SafeFuture<?> doStop() {
     LOG.debug("Stop {}", getClass().getSimpleName());
     syncStateProvider.unsubscribeFromSyncStateChanges(syncStateSubscription.get());
     badPeerCache.clear();
+    reconstructHistoricalStatesService
+        .doStop()
+        .finish(error -> LOG.error(error.getMessage())); // fixme log error not working
     return SafeFuture.COMPLETE;
```
ajsutton: I think in this case we should be returning the result of doStop():
```suggestion
    return reconstructHistoricalStatesService
        .stop();
```

because the `HistoricalBlockSyncService` is entirely encapsulating the `ReconstructHistoricalStatesService` so historical block service isn't stopped until historical states service is.  I suspect though you should be calling `.stop()` not `.doStop()`.

The .finish looks right and should log fine - it will only go to the log file by default, not the console though...

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -0,0 +1,35 @@
...
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.service.serviceutils.Service;
+
+public class ReconstructHistoricalStatesService extends Service {
+
+  @Override
+  protected SafeFuture<?> doStart() {
+    return SafeFuture.failedFuture(
+        new NotImplementedException(
+            "ReconstructHistoricalStatesService start has not been implemented"));
```
ajsutton: I'd just return `SafeFuture.COMPLETE` here and in doStop().  That way it can be hooked in but safely does nothing rather than causing failures.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBlockSyncService.java
```diff
@@ -170,6 +181,10 @@ private void fetchBlocks() {
             () -> {
               if (isSyncDone()) {
                 stop().ifExceptionGetsHereRaiseABug();
+
+                reconstructHistoricalStatesService
+                    .doStart()
+                    .finish(error -> LOG.error(error.getMessage())); // fixme log error not working
```
ajsutton: This should be `.start()` and we should only start the service if reconstructing historical states is enabled (it's safe to stop it regardless - the `Service` base class makes sure stop() does nothing if the service isn't already running.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBlockSyncService.java
courtneyeh: Do we mind that the log only goes into the log file, or should it go through the console too?

ajsutton: I think in this case you'll just return the future so won't need the finish.  If the `start()` call fails we probably should ensure the log goes to the console though.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/SyncConfig.java
```diff
@@ -18,13 +18,19 @@
 public class SyncConfig {
 
   public static final boolean DEFAULT_MULTI_PEER_SYNC_ENABLED = true;
+  public static final boolean DEFAULT_RECONSTRUCT_HISTORIC_STATES_ENABLED = false;
 
   private final boolean isEnabled;
   private final boolean isMultiPeerSyncEnabled;
+  private final boolean isReconstructHistoricStatesEnabled;
```
ajsutton: nit: I'd probably rename this to just `reconstructHistoricStatesEnabled`.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -285,6 +285,10 @@ public void eth1ServiceDown(final long interval) {
     log.warn("Eth1 service down or still syncing for {}s, retrying", interval);
   }
 
+  public void reconstructHistoricalStatesServiceDown(final Throwable throwable) {
```
ajsutton: nit: Probably rename this to `reconstructHistoricalStatesServiceFailedStartup`

ajsutton (approved):
LGTM. Just a couple of suggested renames.

PR #5951, opened 2022-07-20
Don't mark blocks before latestValidHash as fully validated

ajsutton:
## PR Description
While latestValidHash guarantees that every block after it is invalid, there are cases where the EL may not have fully executed the block returned as latestValidHash (typically because it's on a non-canonical fork). As such it is not safe to mark those nodes as fully VALID and they should be left as optimsitically imported.

## Documentation

## Changelog


zilm13:
I think the spec is straight on that: https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#payloadstatusv1
`latestValidHash`: DATA|null, 32 Bytes - the hash of the most recent valid block...
If EL didn't verified it yet it could always return null there, there is no obligation to fill this field

zilm13:
Ok, I see that some ELs are not guaranteed to follow the spec and it's a good protection in this case

zilm13 (approved):
LGTM

PR #5954, opened 2022-07-20
Start Teku from deposit snapshot

zilm13:
## PR Description
Starts Teku using deposit tree snapshot file/url as a source for deposit merkle tree. Than populates Eth1 data caches with requests of deposit events from EL instead of replaying them from the DB.
Part of #5432 
Draft until
- fix early report on "Deposits loaded up to..", they are not yet loaded when log line appears
- add tests
- add changelog entry

Attached kiln snapshot few days old. It takes about 150ms on my box from line "loading snapshot from file" to the latest deposit event fetched and added from EL, so it looks like performance of this approach is sufficient.
[kiln.zip](https://github.com/ConsenSys/teku/files/9154531/kiln.zip)

## Documentation

## Changelog


zilm13:
Finished, ready for review.
- add changelog entry -> instead toggle option to be hidden as it's experimental
-  fix early report on "Deposits loaded up to.." -> they are not loaded on log line in any case and it's difficult to follow exact deposits replay at this block number, so I changed log message to be more correct. Maybe I don't understand it well, I'm up to any fixes

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/TimeBasedEth1HeadTracker.java
```diff
@@ -41,7 +41,7 @@ public class TimeBasedEth1HeadTracker implements Eth1HeadTracker, RunLoopLogic {
   private final AsyncRunner asyncRunner;
   private final Eth1Provider eth1Provider;
 
-  private final Subscribers<HeadUpdatedSubscriber> subscribers = Subscribers.create(true);
+  private final Subscribers<HeadUpdatedSubscriber> subscribers = Subscribers.createWithReplay(true);
```
ajsutton: What drove the need for this change?

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -383,6 +386,30 @@ public void posActivated(final String pandas) {
...
   }
 
+  public void loadingDepositSnapshot(final String snapshotResource) {
+    log.info("Loading deposit tree snapshot from {}", snapshotResource);
+  }
+
+  public void loadedDepositSnapshotResource(
+      final List<Bytes32> finalized, final long deposits, final Bytes32 executionBlockHash) {
+    log.info(
+        "Loaded deposits tree state from snapshot with {} deposits and {} execution block hash. Finalized: {}.",
```
ajsutton: I don't think we need to load the finalized roots - they won't mean anything to users (or developers generally).

infrastructure/subscribers/src/main/java/tech/pegasys/teku/infrastructure/subscribers/Subscribers.java
```diff
@@ -49,12 +50,21 @@
...
+    this.replayLastActionOnSubscription = replayLastActionOnSubscription;
   }
 
   public static <T> Subscribers<T> create(final boolean suppressCallbackExceptions) {
-    return new Subscribers<>(suppressCallbackExceptions);
+    return new Subscribers<>(suppressCallbackExceptions, false);
+  }
+
+  public static <T> Subscribers<T> createWithReplay(final boolean suppressCallbackExceptions) {
+    return new Subscribers<>(suppressCallbackExceptions, true);
```
ajsutton: I'm quite sceptical about adding this here.  The need to add synchronisation has a big impact across a lot of our code base.  I suspect it could introduce some potential deadlocks as well if events are ever raised while holding some other lock.

We do have the `ObservableValue` class which is specifically designed to manage a single value and notify of updates, including providing the current value to new subscribers.

services/powchain/src/main/java/tech/pegasys/teku/services/powchain/DepositSnapshotLoader.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private SafeFuture<LoadDepositSnapshotResult> loadSnapshot() {
+    final Optional<DepositTreeSnapshot> depositTreeSnapshot =
+        depositSnapshotResourceLoader.loadDepositSnapshot(depositSnapshotResource);
+    if (depositTreeSnapshot.isEmpty() || depositTreeSnapshot.get().getDeposits() == 0) {
+      return SafeFuture.completedFuture(LoadDepositSnapshotResult.EMPTY);
+    } else {
+      final Bytes32 executionBlockHash = depositTreeSnapshot.get().getExecutionBlockHash();
+      return eth1Provider
+          .getGuaranteedEth1Block(executionBlockHash.toHexString())
```
ajsutton, line 60: The good news here is that there's a proposal to add the block number to the snapshot format so we will be able to get rid of this look up in the end. Can leave that change for a separate PR though since we haven't had to jump through too many hoops to get by hash here.

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotLoaderTest.java
```diff
@@ -0,0 +1,117 @@
...
+    this.depositSnapshotLoader =
+        new DepositSnapshotLoader(Optional.of(notFoundResource), eth1Provider);
+    final SafeFuture<LoadDepositSnapshotResult> result =
+        depositSnapshotLoader.loadDepositSnapshot();
+    assertThatThrownBy(result::get).cause().isInstanceOf(InvalidConfigurationException.class);
+  }
+
+  @Test
+  public void shouldReturnEmptyResult_whenELIsDown() throws Exception {
+    when(eth1Provider.getGuaranteedEth1Block(any(String.class)))
```
ajsutton: I believe the guaranteed eth1 block method will keep retrying until it gets the block so it should never return an error (hence the `Guaranteed`).

ajsutton (commented):
Looks good generally, mostly just worried about the Subscribers change.

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/TimeBasedEth1HeadTracker.java
zilm13: Notification is pushed before we subscribed, so we miss the head and when we are post-merge, new blocks on eth1 are only those that CL posts, which starts only when we discover enough peers and sync started (almost 1 minute every time for me on Kiln). So we miss actual head without changes and get next block (which is catched) only in a minute

infrastructure/subscribers/src/main/java/tech/pegasys/teku/infrastructure/subscribers/Subscribers.java
zilm13: I agree with you, but there is a case in async environment when we could not guarantee the order, I've described it above, may be we could come to better solution for it

ajsutton: Sounds like we should, probably in a PR by itself, move from using `Subscriber` to `ObservableValue` then.

zilm13: Good, I will take a look at it

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotLoaderTest.java
zilm13: Interesting, funny method. The only issue I see, it will continue to cycle even if hash is incorrect and block doesn't exist. But I'd keep it as it's a best way to survive with EL starting later which doesn't look like a rare case.

Btw this one looks incorrect too https://github.com/zilm13/teku/blob/b2825238533cd0395668efcca0f51da92e90db45/beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java#L173-L174

zilm13:
@ajsutton everything is fixed, ready for review

ajsutton (approved):
LGTM. Just to confirm we aren't currently storing snapshots right? Given the format is going to change we may as well update to add the block number before we start storing things. And then we can remove that getGuaranteedBlock call - I'm a bit nervous it may cause startup to be delayed until the Eth1 node is actually up which isn't good (delaying loading deposits is fine as long as the rest of the node can start up).  That gets simplified though once we have the block number in the snapshot.

zilm13:
We are not yet storing, but it was №3 in our list in #5432 
We could merge this and delay №3 for a while

ajsutton:
Yeah I think it's worth merging this as is - definitely don't want to try and do everything in one PR.

I'd probably check with SigP next but I think the plan is to add block number in which case it would make sense for us to add that next and simplify things.  Then we can look at storage.

Whenever we do look at storage we should make sure its behind a feature toggle so we can test it all out well before enabling it for all users.

PR #5956, opened 2022-07-20
Fix log message when VC refuses to perform a duty in the future

ajsutton:
## PR Description
Old message logged the current epoch but called it a slot which is confusing.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5961, opened 2022-07-21
Fix incorrect error logged when a newly imported block has an invalid payload

ajsutton:
## PR Description
When a block with an invalid payload was imported while the EL was in sync, Teku failed to find the latestValidHash because it wasn't checking the direct parent in this case (the code was originally only dealing with blocks that were imported optimistically and then later found to be invalid).

Teku's behaviour was correct (and there was even a test for it) but in incorrectly logged:

```
2022-07-21 10:53:00.508 ERROR - Could not find latest valid execution payload hash (0xfa87aa1b3ee36c763fb5b083aa3fed49fd037371ffc4949b03f894fcf5c3c099) in the non-finalized chain. Optimistic sync may have finalized an invalid transition block.
```

Now it does not.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,3 +18,4 @@ For information on changes in released versions of Teku, see the [releases page]
 
 ### Bug Fixes
  - Fix `latestValidHash`with invalid Execution Payload in response from execution engine didn't trigger appropriate ForkChoice changes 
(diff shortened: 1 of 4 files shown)


rolfyone (approved):
LGTM

PR #5962, opened 2022-07-21
Rename streamFinalizedBlocks, add getRaw

rolfyone:
- getRaw is needed to efficiently copy blocks from a column, which is not part of this PR, but will be just behind it, and with the plumbing involved it's easier to do by itself.

 - streamFinalizedBlocks is only hitting an unblinded column, so renamed to make that clear.

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockKvStoreDatabase.java
```diff
@@ -289,6 +289,10 @@ protected void storeNonCanonicalBlocks(
     if (storeNonCanonicalBlocks) {
       final Map<UInt64, Set<Bytes32>> nonCanonicalRootsBySlotBuffer = new HashMap<>();
       for (Bytes32 blockRoot : blockRoots) {
+        // TODO #5955
```
ajsutton: What exactly is the TODO here? Do we need the comment to go in?

ajsutton (approved):
LGTM.

PR #5964, opened 2022-07-22
Enable fork choice before proposing by default

ajsutton:
## PR Description
Enable fork choice before proposing by default on all networks.  We've been testing this by enabling it by default on testnets for a while and it isn't introducing any proposal delays so enable on MainNet.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5966, opened 2022-07-22
Run DatabaseTest in multiple threads

ajsutton:
## PR Description
`DatabaseTest` dynamically generates a _lot_ of tests and takes a long time to run, to the point where it becomes the critical path for completing the build.  Enable parallel execution for each of its test methods to make it faster.

Note that this puts quite a lot more pressure on disk operations so had to extend a timeout in the FileKeyValueStoreTest.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5967, opened 2022-07-22
Build migration framework through storage system

rolfyone:
- need an async runner to be able to background longer running tasks if they're not blocking startup.

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -13,32 +13,46 @@
...
     this.spec = spec;
     this.dao = dao;
+    this.asyncRunner = asyncRunner;
   }
 
   void migrateBlocks() {
     performBatchMigration();
+    asyncRunner
+        .runAsync(this::migrateRemainingBlocks)
+        .finish(error -> LOG.debug("Failed to complete block migration", error));
```
ajsutton: I suspect this should be logged at error level.

storage/src/test/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactoryTest.java
```diff
@@ -38,6 +41,9 @@ public class VersionedDatabaseFactoryTest {
   public static final int MAX_KNOWN_NODE_CACHE_SIZE = 10_000;
   private final Eth1Address eth1Address =
       Eth1Address.fromHexString("0x77f7bED277449F51505a4C54550B074030d989bC");
+
+  final AsyncRunnerFactory asyncRunnerFactory =
+      AsyncRunnerFactory.createDefault(new MetricTrackingExecutorFactory(new NoOpMetricsSystem()));
```
ajsutton: I suspect we don't want an actual async runner factory here as it makes tests multithreaded.  Likely just want something that creates StubAsyncRunner instances.

Or if we don't need the actual factory, just create a StubAsyncRunner and use that.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/admin/WeakSubjectivityCommand.java
```diff
@@ -103,11 +105,15 @@ private Database createDatabase(
       final BeaconNodeDataOptions beaconNodeDataOptions,
       final Eth2NetworkOptions eth2NetworkOptions) {
     final Spec spec = eth2NetworkOptions.getNetworkConfiguration().getSpec();
+    final AsyncRunnerFactory asyncRunnerFactory =
+        AsyncRunnerFactory.createDefault(
+            new MetricTrackingExecutorFactory(new NoOpMetricsSystem()));
```
ajsutton: hmm, I'm not entirely sure we want to trigger database migration in subcommands  so probably don't want to have to create the actual async runner...  Maybe we do need to pass in the AsyncRunnerFactory so it can create the runner only if a migration is actually required?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
rolfyone: potentially, i put it to debug initially, i'll need to be checking shutdownException etc here i think...

ajsutton (approved):
LGTM.

PR #5968, opened 2022-07-22
Build ReconstructHistoricalStatesService logic

courtneyeh:
#5749 

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBlockSyncService.java
```diff
@@ -123,9 +123,11 @@ public static HistoricalBlockSyncService create(
       final CombinedChainDataClient chainData,
       final AsyncBLSSignatureVerifier signatureVerifier,
       final SyncStateProvider syncStateProvider,
-      final boolean reconstructHistoricStatesEnabled) {
+      final boolean reconstructHistoricStatesEnabled,
+      final Optional<String> genesisStateResource) {
     ReconstructHistoricalStatesService reconstructHistoricalStatesService =
-        new ReconstructHistoricalStatesService();
+        new ReconstructHistoricalStatesService(
+            storageUpdateChannel, chainData, spec, genesisStateResource);
```
ajsutton: I wonder if we should only create this service if reconstructing states is enabled. Then instead of passing the service and a boolean through, we'd just pass `Optional<ReconstructHistoricalStatesService>` - if it's present it's enabled, if it's absent it isn't.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -13,18 +13,144 @@
...
+    final SafeFuture<Optional<Checkpoint>> future = chainDataClient.getInitialAnchor();
+    future
+        .thenAccept(
+            checkpoint -> {
+              if (checkpoint.isEmpty()) {
+                return; // todo confirm skip
+              }
+
+              final UInt64 anchorSlot = checkpoint.get().getEpochStartSlot(spec);
+              applyBlocks(genesisState, 100, anchorSlot);
```
ajsutton: We should get dataStorageFrequence from the storage config instead of hard coding 100 here.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -13,18 +13,144 @@
...
+                  return;
+                }
+
+                context.currentState =
+                    spec.replayValidatedBlock(context.currentState, block.orElseThrow());
+                if (context.checkStoringHistoricalStates()) {
+                  storageUpdateChannel
+                      .onFinalizedState(context.currentState)
+                      .finish(LOG::error); // todo ignores return value
+                }
```
ajsutton: I think we'll need to break this `thenAccept` block up so that the future returned from `storageUpdateChannel.onFinalizedState` winds up being returned in a thenCompose - that way if the state fails to store our regeneration will fail and the error propagates back up instead of being logged and ignored.

So you wind up with something of the form:

```
getBlockAtSlotExact(context.slot)
  .thenComposeChecked(block -> { // thenComposeChecked lets us avoid catching thee StateTransitionException
    if (block.isEmpty()) {
        return SafeFuture.COMPLETE;
    }
    context.currentState = spec.replayValidatedBlock(context.currentState, block.get());
    if (context.checkStoringHistoricalStates()) {
      return storageUpdateChannel.onFinalizedState(context.currentState);
    } else {
      return SafeFuture.COMPLETE;
    }
  })
  .thenRun(() -> context.incrementSlot())
  .thenCompose(__ -> applyNextBlock(context));
```

I will almost certainly have some detail of that wrong but hopefully you get the idea. :)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -13,18 +13,144 @@
...
+                LOG.error(e); // todo check approach - crash
+              }
+            })
+        .thenCompose(__ -> applyNextBlock(context));
   }
 
   @Override
   protected SafeFuture<?> doStop() {
-    return SafeFuture.COMPLETE;
+    return null;
```
ajsutton: For now stick with:
```suggestion
    return SafeFuture.COMPLETE;
```

though we will need to work out how to actually stop properly I suspect.

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBlockSyncServiceTest.java
```diff
@@ -70,8 +71,13 @@ public class HistoricalBlockSyncServiceTest {
 
   private final SyncStateProvider syncStateProvider = mock(SyncStateProvider.class);
 
+  private final CombinedChainDataClient chainData = mock(CombinedChainDataClient.class);
+  private final Optional<String> genesisStateResource =
+      Optional.of(
+          "https://2driVVdXx:5058c7@eth2-beacon-mainnet.infura.io/eth/v1/debug/beacon/states/finalized");
```
ajsutton: I probably wouldn't use an infura URL here (and if that's a real access key you need to revoke access to that project now).  Just a `https://example.com/state.ssz` type of URL is generally best for testing so there's no risk of the test actually hitting a real web server (example.com is specifically reserved for examples so can't be registered).

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -308,6 +308,21 @@ public void updateWeakSubjectivityState(WeakSubjectivityUpdate weakSubjectivityU
     }
   }
 
+  @Override
+  public void storeFinalizedState(BeaconState state) {
+    if (state == null) {
+      return;
+    }
```
ajsutton: I'd omit this check - state should never be null here.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -308,6 +308,21 @@ public void updateWeakSubjectivityState(WeakSubjectivityUpdate weakSubjectivityU
...
+
+    final Bytes32 blockRoot = spec.getBlockRootAtSlot(state, state.getSlot());
+    try (final FinalizedUpdaterCommon updater = getFinalizedUpdater()) {
+      updater.addFinalizedState(blockRoot, state);
+      updater.commit();
+    }
+  }
+
+  protected abstract FinalizedUpdaterCommon getFinalizedUpdater();
```
ajsutton: I think you can use the existing `finalizedUpdater()` method rather than adding this. The `FinalizedUpdaterT` generic type it returns is guaranteed to be a subclass of `FinalizedUpdaterCommon`.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -308,6 +308,21 @@ public void updateWeakSubjectivityState(WeakSubjectivityUpdate weakSubjectivityU
...
 
+  @Override
+  public void storeFinalizedState(BeaconState state) {
+    if (state == null) {
+      return;
+    }
+
+    final Bytes32 blockRoot = spec.getBlockRootAtSlot(state, state.getSlot());
+    try (final FinalizedUpdaterCommon updater = getFinalizedUpdater()) {
+      updater.addFinalizedState(blockRoot, state);
```
ajsutton, line 322: Worth adding a note that we'll have to come back and also record the finalized state roots as well.  It's a little bit tricky so better to do in a separate PR I think.

ajsutton (commented):
Looks like its on the right track. :)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -13,18 +13,120 @@
...
+    future
+        .thenAccept(
+            checkpoint -> {
+              if (checkpoint.isEmpty()) {
+                return; // todo confirm skip, or should I throw exception
+              }
+
+              applyBlocks(genesisState, checkpoint.get().getEpochStartSlot(spec));
+            })
+        .finish(LOG::error); // todo ignores return value
```
ajsutton: In this case we can probably just return the future - startup is then considered complete only once we loaded the initial anchor which is reasonable. And we're reporting startup failures to the console back in HistoricalBlockSyncService which is probably the best we can do.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -13,18 +13,120 @@
...
+
+    return SafeFuture.COMPLETE; // todo check
+  }
+
+  public void applyBlocks(final BeaconState genesisState, final UInt64 anchorSlot) {
+    Context context = new Context(genesisState, UInt64.ONE, anchorSlot); // todo correct slot 1?
+    applyNextBlock(context).finish(LOG::error); // todo ignores return value
```
ajsutton: This probably needs to use `.finish` but would be better to have the error go to the console - we've been unable to reconstruct historic states before of a failure.  We probably should get it to retry after a period - it may have failed because the disk was full so if it tries again later then it can resume making progress if disk space has become available. Probably best to do as a follow up change though.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -13,18 +13,118 @@
...
+              if (checkpoint.isEmpty()) {
+                return;
+              }
+              applyBlocks(genesisState, checkpoint.get().getEpochStartSlot(spec));
+            });
+  }
+
+  public void applyBlocks(final BeaconState genesisState, final UInt64 anchorSlot) {
+    Context context = new Context(genesisState, SpecConfig.GENESIS_SLOT.plus(1), anchorSlot);
+    applyNextBlock(context).finish(STATUS_LOG::reconstructHistoricalStatesServiceFailedStartup);
```
ajsutton: This probably needs to be a different status log method since it has failed while reconstructing states rather than during startup so the message would be slightly different.

ajsutton (approved):
LGTM. Just need to tweak the error message when reconstruction fails so its clear it did start, but then failed part way along.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -13,18 +13,118 @@
...
+  }
+
+  private SafeFuture<Void> applyNextBlock(Context context) {
+    if (context.checkStopApplyBlock()) {
+      return SafeFuture.COMPLETE;
```
ajsutton, line 89: Actually, probably worth logging an info message to the console here to say the reconstruction is complete.

PR #5969, opened 2022-07-22
Beacon API - Fix post blinded block in SSZ

### LINKED ISSUE #5879: `/eth/v1/beacon/blinded_blocks` API  unable to deserialize SSZ
When using `--Xbeacon-node-ssz-blocks-enabled` validator option in standalone validator, when we request to publish a blinded block, the BN api fails with error `First variable element offset doesn't match the end of fixed part`

Trace from VC perspective:
```
{"timestamp":"2022-06-30T11:18:48,066","level":"DEBUG","thread":"validator-async-0","class":"ResponseHandler","message":"Server error from Beacon Node API (url = http://127.0.0.1:5051/eth/v1/beacon/blinded_blocks, status = 400, message = First variable element offset doesn't match the end of fixed part)","throwable":""}
```

tbenr:
Deserialization fixed: the endpoint metadata was referencing the non-blinded version of the deserializer.

Added API client tests
Added API server tests

fixes #5879

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #5846, opened 2022-06-26
Improve reporting of  "Retrying with next eth1 endpoint"


benjaminion:
Version 22.6.0

I had an ISP outage for a while today. Since my Eth1 node is remote, Teku experienced timeouts in contacting it. This resulted in looong stack traces (I am logging to CONSOLE).

The stack trace is not useful for timeout errors, so I suggest we suppress it and just print the warning.

This is similar to #5416. Note that the "Error publishing metrics to remote service: timeout" is handled nicely below. Likewise I probably don't need several pages of stack traces in one second as a result of my Eth1 node timing out.

Ditto the "ERROR - Failed to update eth1 chain head" could probably be quieter if we know that the Eth1 nodes are unreachable.

```
Jun 26 19:03:03 metalalbert teku[2344]: 2022-06-26 19:03:03.468 INFO  - Slot Event  *** Slot: 4120513, Block:                                                        ... empty, Justified: 128764, Finalized: 128763, Peers: 0
Jun 26 19:03:07 metalalbert teku[2344]: 2022-06-26 19:03:07.000 WARN  - Error publishing metrics to remote service: timeout
Jun 26 19:03:11 metalalbert teku[2344]: 2022-06-26 19:03:11.307 WARN  - Retrying with next eth1 endpoint
Jun 26 19:03:11 metalalbert teku[2344]: java.util.concurrent.CompletionException: java.net.SocketTimeoutException: timeout
Jun 26 19:03:11 metalalbert teku[2344]: at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332) ~[?:?]
Jun 26 19:03:11 metalalbert teku[2344]: at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347) ~[?:?]
Jun 26 19:03:11 metalalbert teku[2344]: at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:636) ~[?:?]
Jun 26 19:03:11 metalalbert teku[2344]: at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510) ~[?:?]
Jun 26 19:03:11 metalalbert teku[2344]: at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162) ~[?:?]
[...]

PR #5972, opened 2022-07-25
5846-eth1-timeout-reporting

mehdi-aouadi:
## PR Description

Removes the stack trace from the logs when eth1 node connection times out. 

## Fixed Issue(s)
Fixes issue https://github.com/ConsenSys/teku/issues/5846

## Documentation

## Changelog


beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/FallbackAwareEth1ProviderSelectorTest.java
```diff
@@ -114,6 +114,31 @@ void shouldFallbackOnGetLogs_smallerRangeException_timeout() {
...
+            fallbackAwareEth1Provider
+                .ethGetLogs(ethLogFilter)
+                .thenRun(() -> fail("should fail!"))
+                .exceptionallyCompose(
+                    err -> {
+                      final Throwable errCause = err.getCause();
+                      assertThat(errCause).isInstanceOf(Eth1RequestException.class);
+                      assertThat(errCause.getSuppressed().length).isEqualTo(1);
+                      assertThat(errCause.getSuppressed()[0])
+                          .isInstanceOf(SocketTimeoutException.class);
```
rolfyone: given that we're doing the raising, I'd probably suggest just using
```
assertThat(err.getCause()).isInstanceOf(Eth1RequestException.class);
```

rolfyone: Could potentially simplify it by getting a SafeFuture, then afterwards evaluating the failed future...
```
SafeFuture<?> future =  fallbackAwareEth1Provider.ethGetLogs(ethLogFilter);
    assertThat(future).isCompletedExceptionally();
    assertThatThrownBy(future::get)
        .hasCauseInstanceOf(Eth1RequestException.class);
```
That shows the failure like the existing test, but what it still doesn't really show is the log not containing the exception, or failing when it does contain the trace...

mehdi-aouadi: Yep, the test doesn't actually check that the logs do not contain the stack trace. Was wondering if it is worth it to intercept the logger ... I will mock/intercept the logger and add a check for that

ajsutton: We probably don't want to mock the logger in this case. Adding a custom appender for the test is probably the best option. https://github.com/Hakky54/log-captor may be an option to make that easy. I haven't used it before so it may not work out, but if it does work it would be quite useful as it seems to provide a pretty simple approach to testing log output which we could use in other places.

mehdi-aouadi: It works and intercepts almost everything! Added it as a dependency for the beacon module only for now. We could add it elsewhere if required

gradle/versions.gradle
```diff
@@ -128,6 +128,10 @@ dependencyManagement {
       entry 'mockito-junit-jupiter'
     }
 
+    dependencySet(group: 'io.github.hakky54', version: '2.7.10') {
+      entry 'logcaptor'
+    }
```
ajsutton: Given there's only a single entry here we can probably just use the single line version of this:
```suggestion
    dependency 'io.github.hakky54:logcaptor:2.7.10'
```

ajsutton (approved):
LGTM. Just suggested a simplification of how the dependency is declared.

PR #5973, opened 2022-07-25
Debug command to print out the slots of stored finalized states

rolfyone:

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/CombinedKvStoreDao.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/CombinedKvStoreDao.java
@@ -418,6 +418,19 @@ public class CombinedKvStoreDao<S extends SchemaCombined>
   }
 
   @Override
+  @MustBeClosed
+  @SuppressWarnings("unchecked")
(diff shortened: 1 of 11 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/CombinedKvStoreDao.java
```diff
@@ -417,6 +417,19 @@ public Map<String, Long> getColumnCounts() {
...
+  @SuppressWarnings("unchecked")
+  public Stream<UInt64> streamFinalizedStateSlots(final UInt64 startSlot, final UInt64 endSlot) {
+    if (schema.getColumnMap().containsKey("FINALIZED_STATES_BY_SLOT")) {
+      KvStoreColumn<UInt64, BeaconState> column =
+          (KvStoreColumn<UInt64, BeaconState>)
+              schema.getColumnMap().get("FINALIZED_STATES_BY_SLOT");
+      return db.stream(column, startSlot, endSlot).map(ColumnEntry::getKey);
+    }
+    return Stream.empty();
+  }
```
ajsutton: I don't think this is right. It should be delegating to `V4FinalizedStateStorageLogic<S> stateStorageLogic` which will know which column to use.

ajsutton (approved):
LGTM.

PR #5980, opened 2022-07-27
Add block migration for finalized storage

rolfyone:
- ability to migrate blocks using byte copies for pre-bellatrix blocks

 - ability to migrate blocks post-bellatrix

 - ability to migrate non-canonical blocks that may be stored

 partially addresses #5312

## Documentation

## Changelog


storage/src/testFixtures/java/tech/pegasys/teku/storage/storageSystem/FileBackedStorageSystemBuilder.java
```diff
@@ -49,13 +49,28 @@ public class FileBackedStorageSystemBuilder {
...
+    return StorageSystem.create(
+        database,
+        createRestartSupplier(),
+        storageMode,
+        storeConfig,
+        spec,
+        ChainBuilder.create(spec));
+  }
+
+  public Database buildDatabaseOnly() {
```
rolfyone, line 71: this is only helpful in a pretty specific case, because the database won't be in a state to standup the storage system until hot blocks are migrated... so for the database migration test its important to be able to get access to the database directly, so that migration can be run.

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigrationTest.java
```diff
@@ -51,33 +55,83 @@ public class BlindedBlockMigrationTest {
...
+      assertHotBlockCounts(database.getColumnCounts(), blockCount, 0, blockCount, 1);
+    }
+  }
+
+  @Test
+  void shouldMigrateFinalizedBlocks(@TempDir final Path tempDir) throws Exception {
+    final int totalBlocks = 39;
+    final int finalizedBlocks = 3 * spec.getGenesisSpec().getSlotsPerEpoch() + 1;
+    // hot storage has all hot blocks plus current finalized block
+    final int hotBlocks = totalBlocks - finalizedBlocks + 1;
```
rolfyone: i need to have a small look at these numbers, just seems odd to be adding here

storage/src/test/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactoryTest.java
```diff
@@ -37,6 +37,8 @@ public class VersionedDatabaseFactoryTest {
 
   private static final StateStorageMode DATA_STORAGE_MODE = PRUNE;
   public static final int MAX_KNOWN_NODE_CACHE_SIZE = 10_000;
+
+  public static final int DEFAULT_BLOCK_BATCH_SIZE = 25;
```
rolfyone: would be easy to remove the configuration of batch size from this PR, may reduce noise, and if a follow up PR is to work this logic through, it'll all be changing anyway...

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/V4FinalizedKvStoreDao.java
```diff
@@ -270,6 +273,13 @@ public void addBlindedFinalizedBlock(
       addFinalizedBlockRootBySlot(block.getSlot(), root);
     }
 
+    @Override
+    public void addBlindedFinalizedBlockRaw(
+        final Bytes blockBytes, final Bytes32 root, final UInt64 slot) {
+      transaction.putRaw(schema.getColumnBlindedBlocksByRoot(), root, blockBytes);
+      addFinalizedBlockRootBySlot(slot, root);
+    }
```
rolfyone, line 281: The risk with this function is we can throw anything in, so if it's used incorrectly it could definitely cause problems, but it does significantly improve the performance of block migration, as long as we don't also copy in blocks that need to be blinded (bellatrix)... I'd almost like for this to only be possible as part of the migration process, and outside of that it somehow is not accessible, but I'm not sure if that's just over-complicating things...

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDaoAdapter.java
```diff
@@ -546,6 +564,11 @@ public void deleteUnblindedFinalizedBlock(final UInt64 slot, final Bytes32 block
       finalizedUpdater.deleteUnblindedFinalizedBlock(slot, blockRoot);
     }
 
+    @Override
+    public void deleteUnblindedNonCanonicalBlockOnly(final Bytes32 blockRoot) {
```
rolfyone, line 568: Similar to the `addBlindedFinalizedBlockRaw`, the only real use for this function should be during migration

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -107,4 +129,167 @@ private void moveFirstFinalizedBlockToBlindedStorage() {
...
+            counter++;
+            if (counter % LOGGING_FREQUENCY == 0) {
+              LOG.info("{} non-canonical blocks moved", counter);
+            }
+          }
+          finalizedUpdaterBlinded.commit();
+          finalizedUpdaterUnblinded.commit();
+        }
+        pause();
+      }
```
rolfyone, line 284: need to add a log to the end of the non-canonical migration

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -107,4 +129,167 @@ private void moveFirstFinalizedBlockToBlindedStorage() {
...
+            if (spec.atSlot(slot).getMilestone().isGreaterThanOrEqualTo(SpecMilestone.BELLATRIX)) {
+              preBellatrix = false;
+              LOG.info("At slot {}, detected bellatrix block, stopping direct copy", slot);
+              continue;
+            }
+            final Optional<Bytes32> maybeRoot = dao.getFinalizedBlockRootAtSlot(slot);
+            if (maybeRoot.isEmpty()) {
+              LOG.warn("Could not find block root for slot {}", slot);
+              continue;
+            }
```
rolfyone: This will skip the block from a raw copy.

The next process is a copy where we have block roots etc, so we can copy via that mechanism instead, so the block isn't lost, it's just not copied as part of this process. It could potentially not even be a warning...

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -107,4 +129,167 @@ private void moveFirstFinalizedBlockToBlindedStorage() {
...
+    return counter;
+  }
+
+  private long migrateFinalizedBlocksPreBellatrix(final long migrationCounter) {
+    long counter = 0;
+    boolean preBellatrix = true;
+    if (migrationCounter > 0) {
+      LOG.info(
+          "Migrating pre-bellatrix blocks to blinded storage, {} finalized blocks to migrate",
+          migrationCounter);
```
rolfyone, line 169: the `XX finalized blocks to migrate` is all finalized blocks, so some may not be copied as part of this step, not sure if that's misleading...
The other finalized migrate will say `XX blocks to migrate`, which is based on the total number of finalized blocks in the index minus the number migrated by this process..

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
rolfyone: changing to be debug

storage/src/test/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactoryTest.java
rolfyone: reverted this cli change, will put back in in a subsequent PR.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -13,22 +13,35 @@
...
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
+import tech.pegasys.teku.storage.server.ShuttingDownException;
 import tech.pegasys.teku.storage.server.kvstore.dataaccess.KvStoreCombinedDaoBlinded;
 import tech.pegasys.teku.storage.server.kvstore.dataaccess.KvStoreCombinedDaoUnblinded;
 
 public class BlindedBlockMigration<
     T extends KvStoreCombinedDaoBlinded & KvStoreCombinedDaoUnblinded> {
   private static final Logger LOG = LogManager.getLogger();
-  private static final int BATCH_SIZE = 1_000;
+  private static final int HOT_BLOCK_BATCH_SIZE = 1_000;
```
ajsutton: I doubt we've actually tested this with 1000 hot blocks so I'd be tempted to drop this to just 100. That will still fit all the hot blocks when the network is finalizing normally but keeps it to a much more reasonable size to ensure we don't run out of RAM.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -70,17 +86,19 @@ private void moveHotBlocksToBlindedStorage() {
...
           }
-          double percentCompleted = counter;
-          percentCompleted /= countBlocks;
-          percentCompleted *= 100;
-          LOG.info("{} hot blocks moved ({} %)", counter, String.format("%.2f", percentCompleted));
-
+          if (counter % LOGGING_FREQUENCY == 0 || counter == countBlocks) {
+            double percentCompleted = counter;
+            percentCompleted /= countBlocks;
+            percentCompleted *= 100;
```
ajsutton: nit: I'd probably just do this in a single line.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/V4FinalizedKvStoreDao.java
ajsutton: I think this is fine to leave exposed. It's hard to imagine a case where we'd call this without being aware of the potential consequences.  Trying to make it available only to the migrator is going to add another level of generics and interfaces which just makes things harder to understand I think.

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigrationTest.java
ajsutton: I believe the current finalized block is stored in both hot and finalized storage which may explain it.

ajsutton (approved):
LGTM.

PR #5981, opened 2022-07-27
Add metrics for failover BN requests

StefanBratanov:
## PR Description

- Add labelled counters for requests sent to failover beacon nodes
- Simplify variable names of request method labels

## Fixed Issue(s)
will help for #5237 

## Documentation

## Changelog


ajsutton (approved):
LGTM. I think it makes sense to have the separate metrics for now but can likely simplify once this is fully supported and `FailoverValidatorApiHandler` is always used (just potentially with an empty list of fail overs).

StefanBratanov:
> LGTM. I think it makes sense to have the separate metrics for now but can likely simplify once this is fully supported and `FailoverValidatorApiHandler` is always used (just potentially with an empty list of fail overs).

I agree. It is worth looking into simplifying it when the two different cli options are combined into one.

PR #5983, opened 2022-07-28
Pass config object to VersionedDatabaseFactory

rolfyone:
Rather than passing a whole stack of values from the config object, pass the config object to VersionedDatabaseFactory.

 - also cleaned up the constructors of VersionedDatabaseFactory.

 - added 2 CLI args to BeaconNodeDataOptions, but they're not currently read by the migration process, as I ended up cleaning up the structure first.

 - cleanup of some options in DebugDbCommand, rather than passing through defaults.

 Following up from this change will be to use the new CLI args to configure the block migrater, but the new module and moved enumerations and interfaces made this already a sizable change. The tests passed, the main thing to be aware of is we need to set the config values on the builder where we used to instantiate the factory.

## Documentation

## Changelog


services/chainstorage/api/src/main/java/tech/pegasys/teku/services/chainstorage/api/StorageConfiguration.java
```diff
@@ -104,14 +119,16 @@ public static final class Builder {
 
     private Eth1Address eth1DepositContract;
     private StateStorageMode dataStorageMode = StateStorageMode.DEFAULT_MODE;
-    private long dataStorageFrequency = VersionedDatabaseFactory.DEFAULT_STORAGE_FREQUENCY;
+    private long dataStorageFrequency = 2048L;
```
ajsutton: This probably should use a constant even if the constant winds up migrating to this class.  The value should be the same here and in `BeaconNodeDataOptions` so both should point to the same constant.

storage/api/build.gradle
```diff
@@ -4,6 +4,11 @@ dependencies {
     implementation project(':ethereum:pow:api')
     implementation project(':ethereum:spec')
 
+    implementation 'org.rocksdb:rocksdbjni'
+    implementation 'org.fusesource.leveldbjni:leveldbjni-win64'
+    implementation 'org.fusesource.leveldbjni:leveldbjni-win32'
+    implementation 'tech.pegasys:leveldb-native'
```
ajsutton: hmm, doesn't seem right that the API module depends on these implementation details.

ajsutton (commented):
Looks good but I'd be very keen to avoid having the leveldb/rocksdb dependencies added to the api module. Hopefully they can just be removed again and aren't really needed - otherwise we may need to think through options a bit more.

services/chainstorage/api/src/main/java/tech/pegasys/teku/services/chainstorage/api/StorageConfiguration.java
rolfyone: hrm i'd forgotten about that bit, I was breaking a cyclic dependency, can look at making that cleaner.

storage/api/build.gradle
rolfyone: As discussed, will attempt to move logic out of DatabaseVersion to elimnate this problem

ajsutton (approved):
LGTM.

PR #5986, opened 2022-07-28
Allow overriding of the public key in validator registrations

### LINKED ISSUE #5985: pubkey override flag for ValidatorRegistrationV1 - Distributed Validators Compatibility
Distributed Validators Require Signed Contents To Be The Same - ValidatorRegistrationV1

The current implementation of the validator registration isn't compatible with a distributed validator

For distributed validators it's important that the signed contents are the same across all "co-validators"

the "pubkey: BLSPubkey" included in the message will be the "pubshare" not the "group/root pubkey" since each co-validator in a distributed validator only knows about its pubshare not about the group pubkey.

Again the simplest implementation is likely to have another --X param for  "registration pubkey override" and then the DVT validators will have to coordinate and make sure that the flag is set correctly across all co-validators

ciaranmcveigh5:
## PR Description

Add a flag that enables users to hardcode the pubkey in the ValidatorRegistrationV1 schema, this is required for distributed validators to come to consensus since in DVT each co-validators owns a pubkeyshare rather than the pubkey meaning each registration will be 

## Fixed Issue(s)

fixes #5985

## Documentation

- Potentially add a paragraph to this document https://hackmd.io/@StefanBratanov/BkMlo1RO9

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorProposerOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorProposerOptions.java
@@ -92,6 +92,16 @@ public class ValidatorProposerOptions {
   private UInt64 builderRegistrationTimestampOverride = null;
 
   @Option(
+      names = {"--validators-builder-registration-pubkey-override"},
+      paramLabel = "<uint64>",
+      showDefaultValue = Visibility.ALWAYS,
+      description =
+          "Set a constant pubkey in BLS format to be used in validator registrations against builder infrastructure.",
+      arity = "1",
+      hidden = true)
+  private BLSPublicKey builderRegistrationPubkeyOverride = null;
+
+  @Option(
       names = {"--validators-proposer-blinded-blocks-enabled"},
       paramLabel = "<BOOLEAN>",
       showDefaultValue = Visibility.ALWAYS,
@@ -111,6 +121,7 @@ public class ValidatorProposerOptions {
                 .blindedBeaconBlocksEnabled(blindedBlocksEnabled)
                 .builderRegistrationDefaultGasLimit(builderRegistrationDefaultGasLimit)
                 .builderRegistrationSendingBatchSize(builderRegistrationSendingBatchSize)
-                .builderRegistrationTimestampOverride(builderRegistrationTimestampOverride));
+                .builderRegistrationTimestampOverride(builderRegistrationTimestampOverride)
+                .builderRegistrationPubkeyOverride(builderRegistrationPubkeyOverride));
   }
 }
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorProposerOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorProposerOptions.java
@@ -92,14 +92,14 @@ public class ValidatorProposerOptions {
   private UInt64 builderRegistrationTimestampOverride = null;
 
   @Option(
-      names = {"--validators-builder-registration-pubkey-override"},
-      paramLabel = "<BLSPublicKey>",
+      names = {"--validators-builder-registration-public-key-override"},
+      paramLabel = "<PUBLIC_KEY>",
       showDefaultValue = Visibility.ALWAYS,
       description =
-          "Set a constant pubkey in BLS format to be used in validator registrations against builder infrastructure.",
+          "Set a constant public key to be used in validator registrations against builder infrastructure.",
       arity = "1",
       hidden = true)
-  private BLSPublicKey builderRegistrationPubkeyOverride = null;
+  private String builderRegistrationPublicKeyOverride = null;
 
   @Option(
       names = {"--validators-proposer-blinded-blocks-enabled"},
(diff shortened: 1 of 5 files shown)


teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorProposerOptions.java
```diff
@@ -91,6 +91,16 @@ public class ValidatorProposerOptions {
...
 
+  @Option(
+      names = {"--validators-builder-registration-pubkey-override"},
+      paramLabel = "<BLSPublicKey>",
+      showDefaultValue = Visibility.ALWAYS,
+      description =
+          "Set a constant pubkey in BLS format to be used in validator registrations against builder infrastructure.",
+      arity = "1",
+      hidden = true)
+  private BLSPublicKey builderRegistrationPubkeyOverride = null;
```
ciaranmcveigh5: does this need to be string and then converted to type BLSPublicKey

There is an open question onA how to handle multiple validator keys on a single "Validator client"

we can have --validator-builder-registration-pubkey-override be a comma-separated list or something similar

an actual validator has pubkey-shares and the override value is the pubkeys associated with that pubkey-share

when you have more than one how are we to know which pubkey-shares relate to which pubkeys to make sure the pubkey-shares sign the correct validator registrations

we're working through ideas for currently

teku/src/test/java/tech/pegasys/teku/cli/options/ValidatorOptionsTest.java
```diff
@@ -154,6 +154,15 @@ public void shouldEnableValidatorRegistrationTimestampOverride() {
         .isEqualTo(Optional.of(UInt64.valueOf(120000)));
   }
 
+  @Test
+  public void shouldEnableValidatorRegistrationPubkeyOverride() {
+    final String[] args = {"--validators-builder-registration-pubkey-override", "120000"};
```
ciaranmcveigh5: 120000 needs to be changed to random BLSPublicKey 

not sure if there's a function for random 32 bytes

validator/client/src/test/java/tech/pegasys/teku/validator/client/ValidatorRegistratorTest.java
```diff
@@ -169,6 +169,23 @@ void registersValidators_shouldRegisterWithTimestampOverride() {
     verify(signer, times(1)).signValidatorRegistration(any());
   }
 
+  @TestTemplate
+  void registersValidators_shouldRegisterWithPubkeyOverride() {
+    when(validatorConfig.getBuilderRegistrationPubkeyOverride())
+        .thenReturn(Optional.of(BLSPublicKey.valueOf(140)));
```
ciaranmcveigh5: BLSPublicKey.valueOf(140) 

again need random BLSPublicKey not sure whats the best function to use or if a hardcoded value should be used

validator/client/src/test/java/tech/pegasys/teku/validator/client/ValidatorRegistratorTest.java
```diff
@@ -381,6 +398,11 @@ private void verifyRegistrations(
             .getBuilderRegistrationTimestampOverride()
             .orElse(stubTimeProvider.getTimeInSeconds());
 
+    final UInt64 expectedPubkey =
+        validatorConfig
+            .getBuilderRegistrationPubkeyOverride()
+            .orElse();
+
```
ciaranmcveigh5: line 398 in this file handles the testing of pubkey

ValidatorRegistration::getPublicKey

appears to do it for multiple pubkeys, not sure how this test should be altered for the pubkey-override flag

StefanBratanov:
Hi, thanks a lot for the contribution. :) I raised a PR in your fork with few changes. We can merge afterwards.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorProposerOptions.java
StefanBratanov: Yeah, it is cleaner to be a string and then we do the conversion in the `ValidatorConfig` . I made this change as part of the PR i raised in your fork.

Regarding the open question, let us know when you have more clarity and we can think about the implementation.

ciaranmcveigh5:
hey @StefanBratanov big thanks for adding the required changes! This will unblock our testnet testing in the interim while we think about the best implementation to handle multiple validators on a single VC

ajsutton (approved):
LGTM. Though I'm slightly confused with how this would work if you are running multiple validators in a single VC (that may be multiple distributed validator "parts" from separate actual validators or just a non-distributed validator alongside a distributed "part"). It looks to me like this override would only ever let the validator register a single public key and all other validators are out of luck.

Given it's a --X hidden option and has fairly minimal impact that's fine to keep things moving forward but doesn't seem like the right long term solution (unless I've misunderstood the implications here).

StefanBratanov:
> LGTM. Though I'm slightly confused with how this would work if you are running multiple validators in a single VC (that may be multiple distributed validator "parts" from separate actual validators or just a non-distributed validator alongside a distributed "part"). It looks to me like this override would only ever let the validator register a single public key and all other validators are out of luck.
> 
> Given it's a --X hidden option and has fairly minimal impact that's fine to keep things moving forward but doesn't seem like the right long term solution (unless I've misunderstood the implications here).

Yeah, I have discussed with @ciaranmcveigh5 and he raised https://github.com/ConsenSys/teku/issues/6009 to cover the multiple validators in single VC scenario and a long term solution. This specific change will unblock them in the meantime for their testing.

PR #5990, opened 2022-07-28
Enable use of ee-endpoint as eth1-endpoint before bellatrix is scheduled

ajsutton:
## PR Description
If the user moves from `--eth1-endpoint` to `--ee-endpoint` before Bellatrix is scheduled we should still use the `ee-endpoint` to retrieve deposits instead of ignoring it.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #5991, opened 2022-07-28
Use block migration CLI in block-migration

rolfyone:

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
@@ -36,10 +36,10 @@ public class BlindedBlockMigration<
     T extends KvStoreCombinedDaoBlinded & KvStoreCombinedDaoUnblinded> {
   private static final Logger LOG = LogManager.getLogger();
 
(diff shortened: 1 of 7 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -85,16 +92,16 @@ private void moveHotBlocksToBlindedStorage() {
...
             counter++;
-          }
-          if (counter % 128 == 0 || counter == countBlocks) {
-            double percentCompleted = counter * 100.0 / countBlocks;
-            LOG.info(
-                "{} hot blocks moved ({} %)", counter, String.format("%.2f", percentCompleted));
+            if (counter % 128 == 0 || counter == countBlocks) {
+              double percentCompleted = counter * 100.0 / countBlocks;
+              LOG.info(
+                  "{} hot blocks moved ({} %)", counter, String.format("%.2f", percentCompleted));
```
rolfyone, line 103: the status counter reporting just moved here so that it always works, regardless of whether the inner loop terminates at a specific point, which makes the reporting period easier to maintain

ajsutton (approved):
LGTM.

PR #5995, opened 2022-07-29
debug tool to validate the finalized chain

rolfyone:
- verify parents are present

 - verify there are no finalized indices missing, and none in slots that are believed to be empty (based on parent-child relationships)

 partially addresses #5312

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/UnblindedBlockKvStoreDatabase.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/UnblindedBlockKvStoreDatabase.java
@@ -109,7 +109,7 @@ public class UnblindedBlockKvStoreDatabase
 
   @Override
   public Optional<Bytes32> getFinalizedBlockRootBySlot(final UInt64 slot) {
-    return Optional.empty();
+    return dao.getFinalizedBlockAtSlot(slot).map(SignedBeaconBlock::getRoot);
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/UnblindedBlockKvStoreDatabase.java
```diff
@@ -100,6 +100,18 @@ public Optional<SignedBeaconBlock> getEarliestAvailableBlock() {
...
+  @Override
+  public Optional<SignedBeaconBlock> getLastAvailableFinalizedBlock() {
+    return dao.getFinalizedCheckpoint()
+        .flatMap(
+            checkpoint -> getFinalizedBlock(checkpoint.toSlotAndBlockRoot(spec).getBlockRoot()));
+  }
+
+  @Override
+  public Optional<Bytes32> getFinalizedBlockRootBySlot(final UInt64 slot) {
+    return Optional.empty();
```
ajsutton: Given how "normal" and potentially useful this method looks, I think its worth implementing it here. It would have to load the block and compute the hash tree root which isn't particularly efficient but better than suggesting the block doesn't exist.

ajsutton (approved):
LGTM. I think its worth defensively implementing the dao method, but I probably wouldn't change the `checkFinalizedIndices` to try and check the slot to root index when using unblinded blocks because there isn't an index for that in that db schema.

Issue #5818, opened 2022-06-19
Log warning when disk operations take too long


ajsutton:
### Description
When there is very high contention for disk IOPs, Teku's performance may suffer, including reduced attestation effectiveness but particularly an inability to catch up to the chain head when syncing. Since Teku is functioning correctly but is bottlenecked by the slow disk access, there are no indications in the logs to indicate why sync is progressing so slowly.

We should investigate ways to identify when database access is slow and log warnings. Threshold for raising the alerts should likely be fairly high as we really just want to highly when disk speed is causing significant issues, to the point of it appearing that Teku isn't working.  Metrics can be used to investigate less significant issues.

PR #5998, opened 2022-07-29
5818 disk operations warning

mehdi-aouadi:
## PR Description
Add a database operations monitoring:
- Execution duration alert threshold can be set with cli option `--Xlog-db-op-alert-threshold` (Value in milliseconds)
- No alerts when the value is set to zero (default value)
- For now only two db operations are monitored: LevelDB write batch and KvStoreDatabase updates

## Fixed Issue(s)
Solution to https://github.com/ConsenSys/teku/issues/5818

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/DbLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/DbLogger.java
@@ -15,6 +15,7 @@ package tech.pegasys.teku.infrastructure.logging;
 
 import static tech.pegasys.teku.infrastructure.logging.ColorConsolePrinter.print;
 
+import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import tech.pegasys.teku.infrastructure.logging.ColorConsolePrinter.Color;
@@ -23,7 +24,7 @@ public class DbLogger {
 
   public static final DbLogger DB_LOGGER = new DbLogger(LoggingConfigurator.DB_LOGGER_NAME);
 
-  private final int dbOpAlertThreshold = LoggingConfigurator.dbOpAlertThreshold();
+  private final int dbOpAlertThresholdMillis = LoggingConfigurator.dbOpAlertThresholdMillis();
 
   @SuppressWarnings("PrivateStaticFinalLoggers")
   private final Logger logger;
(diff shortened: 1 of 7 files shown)


infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/DbLogger.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @SuppressWarnings("PrivateStaticFinalLoggers")
+  private final Logger logger;
+
+  public DbLogger(String name) {
+    this.logger = LogManager.getLogger(name);
+  }
+
+  public void onDbOpAlertThreshold(String opName, long startTime, long endTime) {
+    long duration = (endTime - startTime) / 1000000;
```
ajsutton: We should probably name the params to include the units (e.g. `startTimeNanos`) and then here we can use `TimeUnits.NANOS.toMills` or whatever the appropriate conversion is rather than hard coding the constant.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/DbLogger.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import tech.pegasys.teku.infrastructure.logging.ColorConsolePrinter.Color;
+
+public class DbLogger {
+
+  public static final DbLogger DB_LOGGER = new DbLogger(LoggingConfigurator.DB_LOGGER_NAME);
+
+  private final int dbOpAlertThreshold = LoggingConfigurator.dbOpAlertThreshold();
```
ajsutton: nit: Include the units in the name.

storage/src/main/java/tech/pegasys/teku/storage/server/leveldb/LevelDbTransaction.java
```diff
@@ -92,7 +93,10 @@ public void commit() {
     applyUpdate(
         () -> {
           try {
+            long startTime = System.nanoTime();
             db.write(writeBatch);
+            long endTime = System.nanoTime();
+            DB_LOGGER.onDbOpAlertThreshold("LevelDB update", startTime, endTime);
```
ajsutton:
```suggestion
            DB_LOGGER.onDbOpAlertThreshold("Transaction commit", startTime, endTime);
```

Just to make it a little more user centric.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/LoggingConfigurator.java
```diff
@@ -42,6 +42,7 @@ public class LoggingConfigurator {
   static final String STATUS_LOGGER_NAME = "teku-status-log";
   static final String VALIDATOR_LOGGER_NAME = "teku-validator-log";
   static final String P2P_LOGGER_NAME = "teku-p2p-log";
+  static final String DB_LOGGER_NAME = "teku-db-log";
```
ajsutton, line 45: The `LoggingConfigurator` would have to be updated to enable this log and send its contents to the console same as we do for `teku-validator-log` for example.  Should be able to follow the same pattern by looking at usages of `VALIDATOR_LOGGER_NAME`.  Worth testing it works correctly in the three logging modes (file, console and both).

storage/src/main/java/tech/pegasys/teku/storage/server/leveldb/LevelDbTransaction.java
mehdi-aouadi: I also renamed the `KvStoreDatabase update` to `Block Import`

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/LoggingConfigurator.java
mehdi-aouadi: Done. I tested it locally and it works (console and file and both)

ajsutton (approved):
LGTM.

ajsutton:
So I think the next thing will be to get this enabled on our testnet nodes with a fairly low threshold to see what "'normal" timings look like, then set the default threshold to something higher than that so that we only get alerts when the disk IO is causing significant issues.

PR #6003, opened 2022-08-01
Add initial state path to the provided url

### LINKED ISSUE #5978: Automatically add path to finalized state when --initial-state given domain root
### Description
To make it easier for users to use checkpoint sync, automatically detect when the URL given to --initial-state is the root of a beacon node REST API and automatically append `/eth/v2/debug/beacon/states/finalized`

The right approach here is probably to request the root URL and if it is an error response or not application/octet-stream then log an info message and try with the additional path.

I'd be tempted to restrict this to when the supplied URL has no path (or just `/`) to avoid having it kick in because of a typo in the provided URL or something like that.  So `https://example.com` and `https://example.com/` would have the path added but `https://example.com/someting/` wouldn't.

mehdi-aouadi:
## PR Description

Automatically add the path `/eth/v2/debug/beacon/states/finalized` to the finalized state's url (`--initial-state` parameter) only when:
- The provided url fails
AND
- The provided url does not have a path (https://www.example.com or https://www.example.com/)

## Fixed Issue(s)

Fixes https://github.com/ConsenSys/teku/issues/5978
## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ For information on changes in released versions of Teku, see the [releases page]
  - Updated Sepolia bootnodes
  - Enabled progressive balance tracking optimisation on MainNet
  - Replaced separate metrics for each beacon node request with one metric `beacon_node_requests_total`. Some old metrics were kept for backwards compatibility, but will be removed in future versions.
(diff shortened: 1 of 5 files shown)


ajsutton (approved):
LGTM. Nice work.

ajsutton:
Oh, it's worth adding this as a new feature in the `CHANGELOG`.

PR #6007, opened 2022-08-02
only keep block stream open for a short time during migration

rolfyone:

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
@@ -24,6 +24,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.beacon.pow.exception.RejectedRequestException;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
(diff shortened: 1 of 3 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -97,7 +107,7 @@ private void moveHotBlocksToBlindedStorage() {
             blindedUpdater.addBlindedBlock(block, block.getRoot(), spec);
             unblindedUpdater.deleteUnblindedHotBlockOnly(block.getRoot());
             counter++;
-            if (counter % 128 == 0 || counter == countBlocks) {
+            if (counter % 32 == 0 || counter == countBlocks) {
```
ajsutton: Do we need to make it log progress this often? Not against it but it seems surprisingly frequent.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
```diff
@@ -131,10 +141,31 @@ private void moveFirstFinalizedBlockToBlindedStorage() {
...
+
+    LOG.info("Migrating finalized blocks to blinded storage");
+    copyFinalizedBlockIndexToBlindedStorage();
+    long counter;
+    do {
+      counter = migrateFinalizedBlocksPreBellatrix();
+      finalizedblockTotal += counter;
+      if (finalizedblockTotal % LOGGING_FREQUENCY == 0) {
+        LOG.info("{} blocks moved", finalizedblockTotal);
+      }
```
ajsutton: nit: Could extract this as a method to avoid the duplication.

ajsutton (approved):
LGTM.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockMigration.java
rolfyone: just for hot blocks, can drop it back but at 128 almost never will see... dont have a strong opinion... was helpful debugging

PR #6012, opened 2022-08-02
Enables swagger-ui for migrated rest api

### LINKED ISSUE #5875: Fix Swagger UI issues for JsonTypeDefinitionBeaconRestApi
To allow the removal of the old api logic, we need to be able to instantiate swagger-ui and the swagger-docs endpoint using the new typedef based api.

Validator-api already has swagger-docs being published, but it's not loading a swagger-ui. This is probably the simpler of the 2 scenarios, as it's already a completely typedef based solution using the `RestApiBuilder`.

`RestApiBuilder.openApiDocsEnabled()` currently adds `swagger-docs` endpoint to the validator-api, so it's just the swagger-ui that's missing.

zilm13:
## PR Description
Not finished yet.
Bugs to fix:
- `/swagger-ui` (without trailing /) not working

Tests to do:
- `/swagger-ui/index.html` from server is 200 and correct
- `/swagger-ui/swagger-initializer.js` (and gzipped) from server is 200 and links internally to `/swagger-docs`
- `/swagger-docs` responds with correct json
- there are no duplicates in `/swagger-docs` json
- file in swagger jar `/META-INF/resources/webjars/swagger-ui/<SWAGGER-VERSION/swagger-initializer.js` is the same that in `infrastructure/restapi/src/main/resources/swagger-ui/vendor/swagger-initializer.js`, same for gzipped 

## Fixed Issue(s)

Fixes #5875 

## Documentation

## Changelog


rolfyone:
duplicate definitions in swagger should already be well covered in integration tests... every change that impacts swagger breaks an integration test and needs reference files updated, so we shouldnt need to do any further there I think...

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/SwaggerBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Optional;
+import java.util.Set;
+import javax.servlet.http.HttpServletRequest;
+import kotlin.jvm.functions.Function1;
+import tech.pegasys.teku.infrastructure.restapi.openapi.OpenApiDocBuilder;
+
+public class SwaggerBuilder {
+  private static final String SWAGGER_VERSION = "4.10.3";
+  private static final Set<String> MODIFIED_FILES =
+      Set.of("/swagger-ui/swagger-initializer.js", "/swagger-ui/swagger-initializer.js.gz");
```
rolfyone: my hope is we can use gradle to include a swagger and then just reference that version somewhere like here... This looks like we'll be downloading swagger on load?

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/OptionalDeserializableFieldDefinition.java
```diff
@@ -40,4 +41,25 @@ public void readField(final TBuilder target, final JsonParser parser) throws IOE
     final Optional<TField> value = Optional.ofNullable(deserializableType.deserialize(parser));
     setter.accept(target, value);
   }
+
+  @Override
+  public boolean equals(final Object o) {
```
ajsutton, line 46: Somewhat surprised we need these equals and hashCode methods for this change.  We shouldn't need to change the JSON output at all.

infrastructure/restapi/src/main/resources/swagger-ui/patched/swagger-initializer.js
```diff
@@ -0,0 +1,20 @@
+window.onload = function() {
```
ajsutton, line 1: Its unclear to me what we changed in this file or why it needed to be done. We'll likely need to record that somewhere so that we can handle updates in the future.  And it would be really really good to avoid needing to modify it if we can.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/SwaggerBuilder.java
zilm13: We include precompiled swagger from webjars. We need version just for the path in jar

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/OptionalDeserializableFieldDefinition.java
zilm13: At the moment migrated /swagger-docs has about of dozen of duplicates in schema. Due to absence of this. When starting swagger-ui with it, it fails to render due to schema duplicates. And btw it's incorrect json due to this, it fails on validation in json validator

infrastructure/restapi/src/main/resources/swagger-ui/patched/swagger-initializer.js
zilm13: We have changed path to json. Originally swagger-ui is a web application in Node.js and there are 4 ways to change json location. In webjars swagger is compiled and the only way to change json location is to change this file if I'm not wrong. People usually copying whole swagger-ui to their resources, changing this one and voila. Instead I've made configuration with only this one changed, as I think it will not be changed from version to version (or very rarely changed) and there will be a test that it was not changed in jar, so we could maintain upgrade safely

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/SwaggerBuilder.java
zilm13: It's the only dir in the folder, so I think I could avoid it for sure

zilm13:
Sorry @ajsutton @rolfyone I didn't explain why it's made in this way. But you have correct questions and I've answered it

zilm13:
One additional note: webjars are just static compiled versions of appropriate web applications. Static javascript assets. No logic (originally it was Node.js app with some backend logic). The good is it's updated and it's a purpose of webjars

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/SwaggerBuilder.java
zilm13: Looking at the way we could get it and assuming cache issues sometimes, when we could have two swaggers in classpath, I'd better step back to hardcoded version:
https://github.com/ConsenSys/teku/pull/6012/commits/82f389b53bd6c0db3042b85a137910df0c3b59aa

zilm13:
All done, some cleanup and polishing needed (and avoiding cross-platform test bug).
Overall `swagger-ui` distribution comes with following set of the files: https://github.com/swagger-api/swagger-ui/tree/master/dist, we change two of them and their gzipped versions (not sure that latter is really required, need to clarify): [index.html](https://github.com/swagger-api/swagger-ui/blame/master/dist/index.html) and [swagger-initializer.js](https://github.com/swagger-api/swagger-ui/blame/master/dist/swagger-initializer.js) and use all others from distributive as is. 

It's expected that we will manually patch these two if they are changed (test `SwaggerWebjarIntegrityTest` will fail in this case). Though it's not good to do this manually, you could follow the links and see that they are changed very rarely, so it shouldn't be an issue. In case with plugins this work with everything else is maintained by plugin authors (see [SwaggerHandler.kt](https://github.com/javalin/javalin-openapi/blob/main/javalin-swagger-plugin/src/main/kotlin/io/javalin/openapi/plugin/swagger/SwaggerHandler.kt#L26-L74)). Plus we avoided to copy whole asset folder, which is a plus.

UPDATE: Actually I like that plugin changes resources paths for every version, it's a plus to avoid caching issues, I think we could rework this part too.

infrastructure/restapi/src/main/resources/swagger-ui/patched/swagger-initializer.js
zilm13: You could cherry-pick tests touching `/swagger-docs` to clarify it's failing on the current master.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/SwaggerBuilder.java
rolfyone: we have a swagger based on the javalin open api plugin. once we remove that, we'll only have our own swagger, and it'll need to be maintained in versions.gradle...
While we continue to use the swagger from javalin, we need to continue to ship both, its very version dependent..

rolfyone:
Maybe i steered you in the wrong direction with webjars, i'm not 100% sure what the answer is...

My assumption was that we'd be able to include a second version of swagger.jar and load that on a route via some kind of setup...

does look like some interesting json issues that my current integration tests seem to miss, so im not sure what the answer is re. that side of things...

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/SwaggerBuilder.java
zilm13: In plugin it's hardcoded and I think it's a safer way. Not a very big deal to update one text constant along with gradle version bump. I'd vote for this option

zilm13:
@rolfyone I think it's ok, doesn't look like a lot of things to maintain, file changes we will need to maintain were changed 2 times in the last 2 years and it's kind of files that shouldn't be changed often by `swagger-ui` authors. Plus it's a minor part of the plugin which is already far behind and we will be able to drop plugin after this merge easily.

zilm13:
Anyway you will not found more fresh version of `swagger-ui` that could be easily linked up to the Java application. This jar timeliness is maintained by CI, anything involving json link changes will require manual entry (which could lead to be dated like with the plugin we are using already) or if built in some automatic manner could fail on any `swagger-ui` update

zilm13:
Here's how an update will look like
https://github.com/ConsenSys/teku/pull/6012/commits/583387fce529ef7e7f4c0c51e108fa54baa10fcf
yes, 2 places, but removing [these things](https://github.com/ConsenSys/teku/pull/6012/commits/33fdc84f7ad5b2355746d9cb3fcd5cbab707873b) looks like a pro for me

zilm13:
Ready for review, some rationale/explanations:
1. Using `org.webjars:swagger-ui` because it's an automated built jar from Node.js app so it's the most recent static version of repo release, it's not dependent on some 3rd party manual updates like plugins. It will require us to follow and update changes in 2 files (`index.html` and `swagger-initializer.js`), which doesn't look like a huge work, there were 2 changes in latest 2 years and there will be rare changes in future. If it is not done with update, [the test](https://github.com/zilm13/teku/blob/00559c629eadaef88c9e1548d0ebce1b362fb7b0/infrastructure/restapi/src/test/java/tech/pegasys/teku/infrastructure/restapi/SwaggerWebjarIntegrityTest.java) will fail.
2. `index.html` from `swagger-ui` was rebuilt with Thymeleaf template to change path from version to version to avoid browser cache issues and other weird things which could worsen client experience. And I don't like to hardcode it in a code or smth like this.
3. Update of swagger will require changes in 2 places (see https://github.com/ConsenSys/teku/commit/583387fce529ef7e7f4c0c51e108fa54baa10fcf), if change is made only in one place, some test or tests will fail.
4. As I understand with this update (if merged) we could remove some old stuff and rename some of my tests to replace old, so I'd keep current test names in this PR, they could be renamed when removing old API

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/beacon/migrated/OldMigratedOpenApiIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import tech.pegasys.teku.beaconrestapi.AbstractDataBackedRestAPIIntegrationTest;
+import tech.pegasys.teku.beaconrestapi.ReflectionBasedBeaconRestApi;
+import tech.pegasys.teku.infrastructure.restapi.OpenApiTestUtil;
+
+public class OldMigratedOpenApiIntegrationTest extends AbstractDataBackedRestAPIIntegrationTest {
```
ajsutton: The name here seems confusing.  This isn't using the migrated open api from what I can tell. So I think it's just `ReflectionBasedOptionApiIntegrationTest`?

ajsutton (approved):
LGTM.

PR #6010, opened 2022-08-02
validator builder registration request payload endpoint flag

ciaranmcveigh5:
## PR Description

Add a builder registration request payload endpoint and retrieve values to be signed from that endpoint.

This is needed in relatation to making the builder spec compatible with Dsitributed Validator technology

We have another fix pubkey-override but this only enables 1 validator per VC (validator cli) see #6009 for further details

## Fixed Issue(s)

#6009 

## Documentation

- Potentially add a paragraph to this document https://hackmd.io/@StefanBratanov/BkMlo1RO9

PR #6014, opened 2022-08-03
Add timestamp and public key registration overrides to proposer config

### LINKED ISSUE #6009: Update proposer config to include builder registration overrides
### Problem

This is directly related to #5985 

Distributed Validators Require Signed Contents To Be The Same - for builder registration ValidatorRegistrationV1 the pubkey field will be the pubkey-share for co-validators rather than the aggregated pubkey meaning the contents between the co-validators will differ

#5985 looks to address this with a pubkey override flag that enables the operators to hardcode the pubkey to be used in the registration payload.

This fix only allows 1 validator per VC as it has no way to map the pubkey-shares to the hardcoded aggregated pubkey

Ideally we need a solution that enables multiple validator on a single VC with minimal overhead to the validator operator ie not have to update config/flags everytime a new validator is added

We have a few ideas / proposals at the spec level however realise the timelines on spec changes can be lengthy and so need an interim solution

### Proposal

We can put the builder registration overrides in the proposer config - there is already a builder section that enables the operator to have builderApi enabled/disabled at the individual validator level if we can add the pubkey & timestamp fields in validator builder registration we can have individual validator control on overrides which will enable us to run multiple validators on a single VC with DVT capabilities  

proposed schema from @StefanBratanov 
```
{
  "proposer_config": {
      "builder": {
        "enabled": true,
        "gas_limit": "12345654321",
        "registration_overrides": {
          "timestamp": "xxxx",
          "public_key": "xxxx"
        }
      }
    }
  },
  "default_config": {
    "fee_recipient": "0x6e35733c5af9B61374A128e6F85f553aF09ff89A",
    "builder": {
      "enabled": true,
      "registration_overrides": {
        "timestamp": "xxxx",
        "public_key": "xxxx"
      }
    }
  }
}
```

ciaranmcveigh5:
## PR Description

Add publicKey to builder options in proposer config and removes the pubkey override flag that was a temporary solution to this problem.

This is needed in relation to making the builder spec compatible with Distributed Validator technology

This PR or #6010 is being discussed as the solution to #6009  

## Fixed Issue(s)

fixes #6009

## Documentation

Potentially add a paragraph to this document https://hackmd.io/@StefanBratanov/BkMlo1RO9

### DIFF
--- a/data/serializer/src/main/java/tech/pegasys/teku/provider/BLSPublicKeyDeserializer.java
+++ b/data/serializer/src/main/java/tech/pegasys/teku/provider/BLSPublicKeyDeserializer.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright ConsenSys Software Inc., 2022
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.provider;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import java.io.IOException;
+import tech.pegasys.teku.bls.BLSPublicKey;
+
+public class BLSPublicKeyDeserializer extends JsonDeserializer<BLSPublicKey> {
+  @Override
+  public BLSPublicKey deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+    return BLSPublicKey.fromHexString(p.getValueAsString());
+  }
+}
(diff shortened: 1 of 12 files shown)


ciaranmcveigh5:
There are 4 separate components to changing the proposer config

1. add builder registration-overrides along with public key override to the proposer config + relevant tests
2. add timestamp override to the builder registration-overrides in proposer config along with a  --Xvalidators-builder-registration-default-timestamp flag + relevant tests
3. [optional] remove the pubkey override flag + tests given this can now be achieved via proposer config
4. [optional] remove the timestamp override flag + tests given this can now be achieved via proposer config

This PR has just been updated to only addresses number 1 to reduce the size of the PR and make it easier to review.

Each subsequent item could be its own PR and released step by step rather than as a larger PR that may be harder to review.

Number 1 is the key component to enable multiple validators on a single VC for DVT with the builder api

cc @StefanBratanov

StefanBratanov:
> There are 4 separate components to changing the proposer config
> 
> 1. add builder registration-overrides along with public key override to the proposer config + relevant tests
> 2. add timestamp override to the builder registration-overrides in proposer config along with a  --Xvalidators-builder-registration-default-timestamp flag + relevant tests
> 3. [optional] remove the pubkey override flag + tests given this can now be achieved via proposer config
> 4. [optional] remove the timestamp override flag + tests given this can now be achieved via proposer config
> 
> This PR has just been updated to only addresses number 1 to reduce the size of the PR and make it easier to review.
> 
> Each subsequent item could be its own PR and released step by step rather than as a larger PR that may be harder to review.
> 
> Number 1 is the key component to enable multiple validators on a single VC for DVT with the builder api
> 
> cc @StefanBratanov

Hey, thanks a lot for the explanation and the initial PR. :) I just made a commit to your PR which covers 1 and 2. Also, I kept the two flags from before. The proposer config will take precedence over the flags if they are specified.

ajsutton (approved):
LGTM.

PR #6015, opened 2022-08-03
VC fallback to normal block request if blinded block endpoint is not available

### LINKED ISSUE #5949: Validator Client should fallback to normal block production if blinded block endpoint is not available
Wrt #5237 we need to make sure that when `--validators-proposer-blinded-blocks-enabled` is enabled, if the connected beacon node do not support blinded endpoint (404?), the validator client can still perform its duties by falling back to a non blinded flow.

StefanBratanov:
## PR Description

- Introduce `BlindedBlockEndpointNotAvailableException` which will be thrown in case blinded block endpoint is not available.
- Fallback to unblinded block in case the blinded block endpoint is not available.

## Fixed Issue(s)
fixes #5949 

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #6017, opened 2022-08-04
Add metrics to report number of blocks requested via RPC

ajsutton:
## PR Description
Adds metrics to track the number of requests and total blocks requested via blocks by range and blocks by root RPC.

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
@@ -19,7 +19,11 @@ import java.nio.channels.ClosedChannelException;
 import java.util.Optional;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.hyperledger.besu.plugin.services.MetricsSystem;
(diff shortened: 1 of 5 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
```diff
@@ -57,10 +74,12 @@ public void onIncomingMessage(
     if (storageClient.getStore() != null) {
       SafeFuture<Void> future = SafeFuture.COMPLETE;
       if (!peer.wantToMakeRequest() || !peer.wantToReceiveObjects(callback, message.size())) {
-        peer.disconnectCleanly(DisconnectReason.RATE_LIMITING).ifExceptionGetsHereRaiseABug();
```
rolfyone, line 60: im assuming this is redundant and so was removed, but gotta check..

ajsutton: oh yes. Forgot about that - it was very early in my investigations.  The `peer.wantToMakeRequest` and `peer.wantToReceiveObjects` calls automatically handle the rate limiting disconnect. This was the only place we were doing it manually.

rolfyone (approved):
LGTM

Issue #6016, opened 2022-08-04
Cannot create attestation for future slot errors


ajsutton:
### Description
Investigate why the beacon node time updates lag behind sometimes and cause attestation failures with "Cannot create attestation for future slot" error messages.

W1337lt:
I am getting about 5 of these every day now

ajsutton:
Worth noting here that the most common causes of this are 1) having a -Xmx set too low which causes Java to spend most of it's time running GC (-Xmx5g is generally recommended but feel free to ask for advice on discord for your specific setup), or 2) having contention on disk IO - particularly common with the merge coming up and a lot of people syncing an EL on the same machine. 


So one part of reducing this is moving to async storage updates (#1934) which work is in progress on. There are likely other areas that could be optimised as well. Would be good to get multiple parallel investigations going into the different aspects of this.

mread:
Also saw this post-merge. I'm running Besu, Teku BN and Teku Validator as separate processes. BN has -Xmx6g, no old-gen gcs or cpu spikes showing in prometheus monitoring for any of the processes and it's running on a NVMe drive so there shouldn't be IO issues (in theory). There's various discussion with logs and screenshots in Discord at around 15th Sept 09:55 UTC.  Is there a way of gathering useful debug logs from leveldb2, if that is the likely contender for I/O spikes?

This appears in the leveldb LOG file just after the BN errors, don't know if it's useful or not.
```
2022/09/15-09:02:10.916789 140535372158720 Compacting 4@0 + 0@1 files
```

EvilJordan:
This is happening to me constantly post-merge (but didn't happen at all pre). Happy to provide any logs necessary to help rectify!

tbenr:
> This is happening to me constantly post-merge (but didn't happen at all pre). Happy to provide any logs necessary to help rectify!

@EvilJordan A user reporting similar situation seemed to solve by deleting the db and starting from a checkpoint sync.
Can you do that? let us know if you need instructions.

EvilJordan:
Deleting Teku's db, right? Would love some expert-guidance so I don't screw it up!

tbenr:
@EvilJordan https://discord.com/channels/697535391594446898/697539289042649190
reach me out on discord: tbenr#2593

ajsutton:
[...]

PR #6018, opened 2022-08-04
Add option to track and log long tick processing times

ajsutton:
## PR Description
Adds a hidden `--Xmetrics-tick-timing-tracking-enabled` option to track how long it takes to process a tick event and logs a warning if it takes more than 500ms (at which point the next tick event would be due).

The CLI option is in `MetricsOptions` to match the block timing option though at this stage metrics haven't been added for these timings. That is likely worth doing in a follow up.

## Fixed Issue(s)
#6016 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6023, opened 2022-08-04
[Java API] Make SyncService overridable 

Nashatyrev:
## PR Description

Make `SyncService` replacement easier. 
The original use case was to disable historical block sync. 

- Extract `SyncServiceFactory` interface
- Extract `BeaconChainController.getSyncServiceFactory()` method 
- Make `HistoricalBlockSyncService` overridable

## Documentation

## Changelog


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -919,26 +920,29 @@ public void initBlockManager() {
         .subscribe(BlockImportNotifications.class, blockManager);
   }
 
+  protected SyncServiceFactory getSyncServiceFactory() {
```
ajsutton: nit: Probably rename to `createSyncServiceFactory`

ajsutton (approved):
LGTM.

PR #6024, opened 2022-08-04
Show debug message if fee recipient from builder differs from registration

jtraglia:
## PR Description

Noticed that there was no fee recipient check for execution payloads from a builder. It is possible that a dishonest builder could set this to a different address. I'm not entirely sure that showing a warning and proceeding is the correct thing to do. Maybe we should fall back to the local execution engine.

## Documentation

## Changelog


ajsutton:
Ah the catch here is that it's generally expected that the builder won't set the actual fee recipient to the supplied value. They can build a block in advance of knowing who the fee recipient is (giving them more time) using their own fee recipient then just insert one last transaction into the block to transfer the payment to the requested fee recipient. It may still be worth having an option to warn in this situation but it's not a clear sign of a dishonest builder (it probably indicates a smart builder optimising things well).

jtraglia:
Ah I understand, that makes sense. Do you want an experimental option to enable this warning? I would also tone down the severity of the message and make it shorter. I'm not sure how useful it would be now though.

ajsutton:
I don't really know what we should do here. I lean towards not bothering but I'm not the MEV expert.  @StefanBratanov and @benjaminion any thoughts?

StefanBratanov:
Hey, i remember having a discussion about this when the `BuilderBidValidator` was implemented and we decided not to verify the fee recipient since as @ajsutton mentioned it is not a sign of a dishonest builder and builders may prefer this strategy. 

I think though actually it could be useful to add this log statement as `LOG.debug` in `BuilderBidValidatorImpl`. (think it's better place to add this since you have access to the registration and the header). Also I would probably change the message to say that it is expected behavior and a small explanation.

jtraglia:
@StefanBratanov thanks! I think a debug message makes more sense and I agree that `BuilderBidValidatorImpl` is a much better place for this. I gave it a fairly brief debug message and a large comment (mostly stolen from @ajsutton) above that block of code. If a user sees this message and wants to learn more, I hope they reference that comment.

ajsutton (approved):
LGTM.

PR #6036, opened 2022-08-05
Log full block when block import fails with internal error

### LINKED ISSUE #6021: Log full block when block import files with internal error
### Description
When we fail to import a block because of an internal error, it doesn't trigger the invalid gossip logging so we currently don't get any record of the actual block content that was invalid.  We should always log the full block content (as hex encoded ssz) when block import fails with internal error so we can fully investigate.

This shouldn't depend on the setting to log invalid gossip being enabled and should only apply to internal errors, not blocks we reject for being invalid.

StefanBratanov:
## PR Description
Enhance error log when importing a block with hex encoded ssz of the block contents.

## Fixed Issue(s)
fixes #6021 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImporter.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImporter.java
@@ -218,7 +218,7 @@ public class BlockImporter {
   }
 
   private String getBlockContent(final SignedBeaconBlock block) {
-    return block.getMessage().sszSerialize().toHexString();
+    return block.sszSerialize().toHexString();
   }
 
   private ReadOnlyForkChoiceStrategy getForkChoiceStrategy() {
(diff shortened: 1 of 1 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImporter.java
```diff
@@ -213,6 +217,10 @@ private String formatBlock(final SignedBeaconBlock block) {
     return LogFormatter.formatBlock(block.getSlot(), block.getRoot());
   }
 
+  private String getBlockContent(final SignedBeaconBlock block) {
+    return block.getMessage().sszSerialize().toHexString();
```
ajsutton: We should log the whole signed block so just:
```suggestion
    return block.sszSerialize().toHexString();
```

ajsutton (approved):
LGTM just with the tweak to include the signature wrapper as well.

PR #6039, opened 2022-08-06
Fix dump-hot-blocks command so it creates the file correctly

ajsutton:
## PR Description
Previously the dump-hot-blocks debug command would only overwrite an existing file. Now it much more usefully also supports creating the output file when it doesn't exist.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6041, opened 2022-08-08
Remove unnecessary configuration exception

StefanBratanov:
## PR Description
Remove `InvalidConfigurationException` since it is unnecessary. The `ValidatorRegistrator` would prefer a value from the proposer config if specified, but otherwise it defaults to the cli option. It is good to allow this since `builder` in the proposer config is an optional element.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #6044, opened 2022-08-09
Added getExecutionPayload database interface

rolfyone:
- get Execution Payload from the database

 - use Payload Summary in the update result, and added isDefault to allow either header or full payload to be used for the update result.

 - renamed SlotAndExecutionPayload now that it's using the payload summary.

 This will ensure the update results continue to succeed once the full execution payload is not stored.

 partially addresses #5312

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayload.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayload.java
@@ -104,7 +104,7 @@ public class ExecutionPayload
   }
 
   @Override
-  public boolean isDefault() {
+  public boolean isDefaultPayload() {
     return super.isDefault();
   }
 
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayload.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayload.java
@@ -104,6 +104,11 @@ public class ExecutionPayload
   }
 
   @Override
+  public boolean isDefault() {
+    return super.isDefault();
+  }
+
+  @Override
   public ExecutionPayloadSchema getSchema() {
     return (ExecutionPayloadSchema) super.getSchema();
   }
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadHeader.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadHeader.java
@@ -103,8 +103,8 @@ public class ExecutionPayloadHeader
   }
 
   @Override
-  public boolean isDefault() {
-    return super.isDefault();
+  public boolean isDefaultPayload() {
+    return isHeaderOfDefaultPayload();
   }
 
   @Override
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadHeader.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadHeader.java
@@ -103,6 +103,11 @@ public class ExecutionPayloadHeader
   }
 
   @Override
+  public boolean isDefault() {
+    return super.isDefault();
+  }
+
+  @Override
   public ExecutionPayloadHeaderSchema getSchema() {
     return (ExecutionPayloadHeaderSchema) super.getSchema();
   }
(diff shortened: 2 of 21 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayload.java
```diff
@@ -103,6 +103,11 @@ public ExecutionPayload(
         transactions);
   }
 
+  @Override
+  public boolean isDefault() {
+    return super.isDefault();
+  }
```
ajsutton: Can probably just omit this since it's just calling the super implementation anyway.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadHeader.java
```diff
@@ -102,6 +102,11 @@
         transactionsRoot);
   }
 
+  @Override
+  public boolean isDefault() {
+    return super.isDefault();
+  }
```
ajsutton: I'm not sure this is right. The header of the default execution payload will return `false` from the SSZ `isDefault()` (it's not all zeros because it has the hash tree root of empty transaction list). I suspect we want to rename the method on the summary interface to `isDefaultPayload` to differentiate between the default SSZ object and the default payload.

And would be worth adding tests for the new isDefault method to make sure we are getting it right.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockKvStoreDatabase.java
```diff
@@ -262,6 +260,13 @@ public Optional<Bytes32> getFinalizedBlockRootBySlot(final UInt64 slot) {
     return dao.getFinalizedBlockRootAtSlot(slot);
   }
 
+  @Override
+  public Optional<ExecutionPayload> getExecutionPayload(
+      final Bytes32 blockRoot, final UInt64 slot) {
+    final Optional<Bytes> maybePayload = dao.getExecutionPayload(blockRoot);
```
ajsutton, line 266: Was there a reason the dao is treating the payload as raw bytes instead of the usual approach where the column is typed and then we use getRaw if we need the raw bytes?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/UnblindedBlockKvStoreDatabase.java
```diff
@@ -113,6 +113,12 @@ public Optional<Bytes32> getFinalizedBlockRootBySlot(final UInt64 slot) {
     return dao.getFinalizedBlockAtSlot(slot).map(SignedBeaconBlock::getRoot);
   }
 
+  @Override
+  public Optional<ExecutionPayload> getExecutionPayload(
+      final Bytes32 blockRoot, final UInt64 slot) {
+    return Optional.empty();
```
ajsutton: Should this load the block and get the payload from there? Generally wouldn't be used I guess but might be worth implementing defensively just to avoid surprises.

ajsutton (commented):
Looks good, just think there's an issue with checking if the payload header is default because of the fun quirk that the default payload is all zeros but the default payload header is not.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayload.java
rolfyone: intellij was insistent but i'll try again...

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/UnblindedBlockKvStoreDatabase.java
rolfyone: we can't switch back to unblinded, i could fetch the block and then get it off of that potentially?

ajsutton: Yeah the unblinded database would get a payload by loading the full block, then getting the payload from it. Bit wasteful but we're really only implementing it to avoid surprises.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/BlindedBlockKvStoreDatabase.java
rolfyone: it likely needs a deserializer written, has to have the spec to get the payload decoded currently...

ajsutton: We'd have the same requirement for blocks so that should work out ok.  But worth leaving to a follow up PR.

rolfyone: blocks deserializer can read from the body, we wont be able to... but yes, we can investigate

ajsutton: ah yes interesting point.  That could well be the reason we did it this way....

ajsutton (approved):
LGTM.

PR #6045, opened 2022-08-09
Implement doStop method

courtneyeh:
#5749

## Documentation

## Changelog

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
@@ -45,7 +45,7 @@ public class ReconstructHistoricalStatesService extends Service {
   private final StorageUpdateChannel storageUpdateChannel;
   private final StatusLogger statusLogger;
 
-  private final AtomicBoolean shutdown;
+  private final AtomicBoolean shutdown = new AtomicBoolean(false);
   private final SafeFuture<Void> stopped = new SafeFuture<>();
 
   public ReconstructHistoricalStatesService(
@@ -67,7 +67,6 @@ public class ReconstructHistoricalStatesService extends Service {
     this.spec = spec;
     this.genesisStateResource = genesisStateResource;
     this.statusLogger = statusLogger;
-    this.shutdown = new AtomicBoolean(false);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -93,12 +102,33 @@ protected SafeFuture<?> doStart() {
...
+              } else {
+                statusLogger.reconstructHistoricalStatesServiceFailedProcess(error);
+              }
+            });
   }
 
   private SafeFuture<Void> applyNextBlock(Context context) {
     if (context.checkStopApplyBlock()) {
       statusLogger.reconstructHistoricalStatesServiceComplete();
+      stopped.complete(null); // fixme should this be set here too?
```
ajsutton: Yep, we want to call complete here. That way if the service finishes it's work before stop() is called, the future returned by stop() will be complete - otherwise it would never complete because `applyNextBlock` wouldn't be called anymore and so we'd never check the shutDown boolean again.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -41,6 +46,9 @@ public class ReconstructHistoricalStatesService extends Service {
   private final StorageUpdateChannel storageUpdateChannel;
   private final StatusLogger statusLogger;
 
+  private final AtomicBoolean shutDown;
```
ajsutton: nit: `shutdown` with lowercase D.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -93,12 +102,33 @@ protected SafeFuture<?> doStart() {
...
     if (context.checkStopApplyBlock()) {
       statusLogger.reconstructHistoricalStatesServiceComplete();
+      stopped.complete(null); // fixme should this be set here too?
+      return SafeFuture.COMPLETE;
+    }
+
+    if (shutDown.get()) {
+      statusLogger.reconstructHistoricalStatesServiceFailedProcess( // fixme log out stop?
+          new FatalServiceFailureException(
+              "ReconstructHistoricalStatesService", "Service was stopped before complete"));
```
ajsutton: We don't need to log anything here - teku is just shutting down and we'll resume reconstructing states when we start up again.

ajsutton (commented):
Looks good. Probably worth adding some tests for stopping both while it's still reconstructing and when it's finished to make sure the future returned from `stop` completes at the right time.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/ReconstructHistoricalStatesService.java
```diff
@@ -60,6 +67,7 @@ public ReconstructHistoricalStatesService(
     this.spec = spec;
     this.genesisStateResource = genesisStateResource;
     this.statusLogger = statusLogger;
+    this.shutdown = new AtomicBoolean(false);
```
ajsutton: very tiny nit: Can initialize this in the field declaration.

ajsutton (approved):
LGTM.

PR #6046, opened 2022-08-09
Prime justified

ajsutton:
## PR Description
Active balances from the justified state are used when applying fork choice, which is often the first thing to require them after justification. Rather than delay fork choice, prime the caches as part of priming the epoch transition.

## Fixed Issue(s)
#6016 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6051, opened 2022-08-10
Avoid waiting for a specific epoch in ATs

ajsutton:
## PR Description
We don't control time in ATs so waiting for a specific epoch isn't safe as we may have already gone past that point.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6052, opened 2022-08-10
Wait for all pending events to be processed before stopping event channels

ajsutton:
## PR Description
Wait for all pending events to be processed before stopping event channels.

New events published after the channel is stopped are still dropped, but this ensures that all events published before shutdown starts are completed during a clean shutdown.

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/AbstractNode.java
+++ b/teku/src/main/java/tech/pegasys/teku/AbstractNode.java
@@ -157,7 +157,7 @@ public abstract class AbstractNode implements Node {
     eventChannels
         .stop()
         .orTimeout(30, TimeUnit.SECONDS)
-        .handleException(error -> LOG.error("Failed to stop event channels", error))
+        .handleException(error -> LOG.warn("Failed to stop event channels cleanly", error))
         .join();
 
     threadPool.shutdownNow();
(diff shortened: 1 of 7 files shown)


teku/src/main/java/tech/pegasys/teku/AbstractNode.java
```diff
@@ -154,7 +154,12 @@ private void pollRejectedExecutions() {
   @Override
   public void stop() {
     // Stop processing new events
-    eventChannels.stop();
+    eventChannels
+        .stop()
+        .orTimeout(30, TimeUnit.SECONDS)
+        .handleException(error -> LOG.error("Failed to stop event channels", error))
```
rolfyone: not sure that we want an error here in all cases... its likely mostly noise?

rolfyone (approved):
LGTM, error may be over-selling it if we're shutting down?

teku/src/main/java/tech/pegasys/teku/AbstractNode.java
ajsutton: hmm, not sure. Given this is how during normal shutdown we'll wind up ensuring that any async updates actually get written to the DB I think we want to know if that fails. This wouldn't log an error if any of the tasks fail, just if the event channels can't finish processing events within the 30 seconds.  I might bump it down to WARN since I think we want to know about it but it's probably not the end of the world.

PR #6060, opened 2022-08-11
Move unblinding of blocks into chain storage.

rolfyone:

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
@@ -194,13 +194,6 @@ public class ChainStorage
           block.unblind(spec.atSlot(block.getSlot()).getSchemaDefinitions(), maybePayload.get()));
     }
 
-    if (executionLayerChannel.isEmpty()) {
-      throw new StorageException(
-          String.format(
-              "Needed a payload, but not able to retrieve from storage for block %s (%s)",
-              block.getRoot(), block.getSlot()));
-    }
-
     throw new StorageException(
         String.format(
             "Needed a payload, but not able to retrieve from storage for block %s (%s)",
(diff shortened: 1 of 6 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
```diff
@@ -132,17 +155,67 @@ public SafeFuture<Optional<SignedBeaconBlock>> getEarliestAvailableBlock() {
...
+      throw new StorageException(
+          String.format(
+              "Needed a payload, but not able to retrieve from storage for block %s (%s)",
+              block.getRoot(), block.getSlot()));
+    }
+
+    throw new StorageException(
+        String.format(
+            "Needed a payload, but not able to retrieve from storage for block %s (%s)",
+            block.getRoot(), block.getSlot()));
```
rolfyone, line 200: this is where we'd have to do a fetch, currently should never reach here as we're not ever deleting payloads

storage/src/testFixtures/java/tech/pegasys/teku/storage/storageSystem/InMemoryStorageSystemBuilder.java
```diff
@@ -101,6 +101,7 @@ public StorageSystem build() {
         storageMode,
         storeConfig,
         spec,
+        Optional.empty(),
```
rolfyone, line 104: at the moment no need, but in-mem could always query the EL, it would just need an async runner here

storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
```diff
@@ -132,17 +155,60 @@ public SafeFuture<Optional<SignedBeaconBlock>> getEarliestAvailableBlock() {
...
+
+  private SafeFuture<Optional<SignedBeaconBlock>> unblindBlock(
+      final Optional<SignedBeaconBlock> maybeBlock) {
+    if (maybeBlock.isEmpty()) {
+      return SafeFuture.completedFuture(maybeBlock);
+    }
+    return unblindedBlock(maybeBlock.get()).thenApply(Optional::of);
+  }
+
+  private SafeFuture<SignedBeaconBlock> unblindedBlock(final SignedBeaconBlock block) {
```
ajsutton: nit: Maybe this should be called unblindBlock so it's performing the action of unblinding the block.

ajsutton (approved):
LGTM.

PR #6061, opened 2022-08-11
Detect ee-endpoint timeouts correctly

ajsutton:
## PR Description
Detect correct exception for timeouts when using the execution engine api.  The exception differs because the engine API is the only place we add an overall request timeout to the OkHttpClient configuration.  So when using `--eth1-endpoint` we get a `SocketTimeoutException` if response takes too long from the HTTP level, but the `--ee-endpoint` configures time outs and so triggers an `InterruptedIOException` when the socket is closed by the timeout.

This was causing extra noise in the logs but also the request for deposit logs wasn't being adjusted to reduce the range, which generally meant retrying it led to another timeout until eventually the disk cache was warmed up enough to get the response in time.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,3 +25,4 @@ For information on changes in released versions of Teku, see the [releases page]
 
 ### Bug Fixes
 - Fixed `NullPointerException` when checking for the terminal PoW block while the EL was syncing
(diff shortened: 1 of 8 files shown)


rolfyone (approved):
LGTM

PR #6064, opened 2022-08-12
set ignoring payload attributes message to debug level

### LINKED ISSUE #6063: reduce log level of message for payload attributes
Getting a stack of `Ignoring calculated payload building attributes since it violates ordering` in the logs.

It's got no context, and coming up as a warning. initially we should bust this to debug level, and potentially it may be something that things could be done to fix but would probably be an @tbenr question around whether it indicates an issue we need to address...

In the scope of this ticket, make a debug level.

rolfyone:
fixes #6063

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #6065, opened 2022-08-12
Include currently non-viable chain heads when updating progressive balanaces

ajsutton:
## PR Description
Since the progressive balances update may cause the node to be viable, make sure we don't skip it because it currently appears non-viable.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6066, opened 2022-08-12
Include beacon block and payload info in warning when execuction client is still syncing

ajsutton:
## PR Description
Include beacon block and payload info in warning when execuction client is still syncing.

Add toLogString methods to simplify calls to LogFormatter.formatBlock.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6071, opened 2022-08-15
Remain in optimistic mode if there is no viable head

ajsutton:
## PR Description
During optimistic sync it is possible to import blocks that include enough attestations to update the justified checkpoint, but then discover the blocks including those attestations were invalid. This potentially makes the valid chain non-viable because it doesn't (yet) include enough attestations to reach that same justified checkpoint. Note that in this case the justified checkpoint itself is valid and eventually the attestations required to justify will be included in a valid chain.

Until that point though, our node should remain in optimistic mode so that it doesn't create attestations with the source of the updated justified checkpoint which may lead to later being unable to attest without creating a surround vote.

See https://github.com/ethereum/consensus-specs/pull/2955 for more details.

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
@@ -241,7 +241,7 @@ public class ProtoArray {
 
     // Perform a sanity check that the node is indeed valid to be the head.
     if (!nodeIsViableForHead(bestNode) && !bestNode.equals(justifiedNode)) {
-      throw new RuntimeException("ProtoArray: Best node is not viable for head");
+      throw new IllegalStateException("ProtoArray: Best node is not viable for head");
     }
     return Optional.of(bestNode);
   }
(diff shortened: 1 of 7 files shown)


storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -241,7 +241,7 @@ private Optional<ProtoNode> findHead(
 
     // Perform a sanity check that the node is indeed valid to be the head.
     if (!nodeIsViableForHead(bestNode) && !bestNode.equals(justifiedNode)) {
-      throw new RuntimeException("ProtoArray: Best node is not viable for head");
+      throw new IllegalStateException("ProtoArray: Best node is not viable for head");
```
rolfyone: is it worth adding block root and slot here?

rolfyone (approved):
LGTM

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
ajsutton: Seems like a good idea yeah.

PR #6073, opened 2022-08-15
Fix intermittency in shouldMaintainValidatorsInMutableClient

ajsutton:
## PR Description
Wait a full epoch to pass so that all validators have attested and generated slashing protection data. Previously only waited until the next epoch.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6074, opened 2022-08-15
Stop using fallback logic for ee-endpoint

### LINKED ISSUE #6025: 'retrying with next eth1 endpoint' needs to take into account ee-endpoint
When an ee-endpoint is in use, the fallback logic seems fairly redundant. We had a user asking about the warning saying `Retrying with next eth1 endpoint`, because they can only specify 1 ee-endpoint, which is a fair question.

It's possible the fallback logic is now not usable, or we need to be looking at if there's a fallback before we raise this warning...

Endpoint wasn't up, and setting was `--ee-endpoint=http://localhost:8551/"`

```
2022-08-04 01:52:37.791+03:00 | pool-3-thread-5 | WARN  | FallbackAwareEth1Provider | Retrying with next eth1 endpoint
```

lucassaldanha:
## PR Description

When using `--ee-endpoint` there is no need for using `FallbackAwareEth1Provider` because there will always be a single endpoint.

## Fixed Issue(s)
fixes #6025 

## Documentation

## Changelog


lucassaldanha:
Writing unit tests for this piece was too hard. I have raised a ticket to revisit this class and make it more testable: https://github.com/ConsenSys/teku/issues/6077

We have done manual testing to ensure the desired behaviour has been achieved.

rolfyone (approved):
LGTM

Issue #5999, opened 2022-08-01
add gas-limit to validator-api


rolfyone:
https://github.com/ethereum/keymanager-APIs/pull/39

it's looking like we're going to have a gas_limit much like fee_recipient, so its probably certain enough that we can start to implement based on the content of this PR.

PR #6081, opened 2022-08-16
Extract apis constants

mehdi-aouadi:
## PR Description

Extract the common Rest API constants

## Fixed Issue(s)

Preparation for https://github.com/ConsenSys/teku/issues/5999

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/BeaconProposerPreparer.java
```diff
@@ -55,12 +57,14 @@ public class BeaconProposerPreparer implements ValidatorTimingChannel, FeeRecipi
       ValidatorIndexProvider validatorIndexProvider,
       ProposerConfigProvider proposerConfigProvider,
       Optional<Eth1Address> defaultFeeRecipient,
+      UInt64 defaultGasLimit,
```
rolfyone: variables should all really be 'final'.. in function definition and when declared, we may as well fix this file while we're here... The only time we don't use final is in an `interface` as a rule.

rolfyone (commented):
Minor nit re. variables in BeaconProposerPreparer, but overall looks good, after all the comments already on 6080 are happy, this one should be fine.

rolfyone (approved):
LGTM based on changes in other PR

PR #6084, opened 2022-08-16
Update reference tests to 1.2.0-rc.2

ajsutton:
## PR Description
Update reference tests to 1.2.0-rc.2.  Some new light client tests are ignored since we don't support light clients. Ref tests now use the real mainnet TTD so we don't need to override it back to the placeholder value.

## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -282,7 +282,7 @@ allprojects {
   }
 }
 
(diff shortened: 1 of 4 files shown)


rolfyone (approved):
LGTM

PR #6085, opened 2022-08-16
Increase memory allocation for unit tests

ajsutton:
## PR Description
Control maximum number of workers but give each more memory to see if it stops the crashing. Uses a larger executor as well to make sure there's enough memory.

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #1934, opened 2020-05-26
Make disk calls non-blocking


arash009:
Make disk transactions non-blocking by removing join() calls to the future which transaction.commit() returns. This would allow the blockImporter to proceed and complete without waiting for the database write to complete, making block import faster, but would also mean that we need to take 

https://github.com/PegaSysEng/teku/blob/a51a5e71316791aafbe5f2ddf3caf8701658057c/ethereum/statetransition/src/main/java/tech/pegasys/artemis/statetransition/blockimport/BlockImporter.java#L63

PR #6086, opened 2022-08-17
Perform database updates asynchronously

ajsutton:
## PR Description
Applies transactions to the in-memory store immediately, allowing the actual database update to be performed asynchronously. Toggled off by default - lots more testing needed and double check for any unexpected threading implications.

Will also need to handle the case where new block is stored and quickly dropped from in-memory store, then attempts to be loaded from disk again before the database write has happened. Currently updates are on a different channel to requests so a race condition exists where the read could be attempted before the write completes.  One mitigating factor here though is that the database write will hold a hard reference to the data being stored which would prevent it being GC'd and thus it wouldn't be dropped from the in-memory store.  The exception there is when a transaction updates the finalized checkpoints - the hot blocks would be immediately removed from the in-memory store but not yet available from the database.

Simplest solution is to add additional storage for blocks and states in Store using hard references. Any blocks or states sent to the database to be stored are added to those maps as part of updating the in-memory store, then remove once the database update completes. May need to tweak `CombinedChainDataClient` to be able to deal with blocks before the finalized checkpoint possibly only being available from the in-memory store. In most cases it already tries the in-memory store first anyway to avoid race conditions with the finalized checkpoint updating.

## Fixed Issue(s)
#1934 

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/RetryingStorageUpdateChannel.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private <I, O> SafeFuture<O> retry(final Function<I, SafeFuture<O>> method, final I arg) {
+    while (true) {
+      try {
+        final SafeFuture<O> result = method.apply(arg);
+        result.join();
+        return result;
+      } catch (final Throwable t) {
+        LOG.error("Storage update failed, retrying.", t);
+      }
```
rolfyone, line 80: oh this looks fun...

ajsutton: Yeah, needs some comments on that class.  If we fail to store and ignore the update, we'll corrupt the database because the next update may depend on this one.  So at some point we probably need to exit, but if it's a temporary disk access issue or full disk we may want to retry a bit.

If it really can't proceed it will wind up applying back pressure and we won't be able to send anymore updates and things will wind up blocking further block imports etc (whereas today we'd reject the block if it failed to store).  But likely we want to actually crash rather than get to that point so that it's a clear failure rather than a node that's up but not functioning correctly.

ajsutton:
Closing this one. The async support went in but with a different approach to handling queries so we don't need the buffering.

Issue #6087, opened 2022-08-17
Phase0 voluntary exit messages are invalid in Bellatrix


benjaminion:
More generally, voluntary exits are valid only when signed with the current fork version or the previous fork version. See [`get_domain()`](https://eth2book.info/altair/part3/helper/accessors#get_domain).

There are reports that Teku will include VEs signed with a Phase0 fork version in Bellatrix blocks. This would make the blocks invalid.

**Update:** reports are unclear as to whether Teku is putting stale VEs in blocks or just propagating them via gossip. In any case, we should check and fix if necessary.

**Update2:** seems like the issue might be that the beacon API is accepting and propagating stale VEs. This is non-critical, but is undesirable.

PR #6090, opened 2022-08-18
Ensure operations that are in the pool and not valid are not included

rolfyone:
- exits that are for an epoch prior to the last milestone should not be included

 - slashings that are for an epoch prior to the last milestone should not be included

 - ensure we error if a request is made to voluntary exit command to generate an exit that would be too old

partially addresses #6087

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -752,6 +752,19 @@ public boolean isMergeTransitionComplete(final BeaconState state) {
...
+    final SpecMilestone milestone = atSlot(slot).getMilestone();
+    if (milestone == SpecMilestone.PHASE0 || milestone == SpecMilestone.ALTAIR) {
+      return UInt64.ZERO;
+    }
+    // for bellatrix, we would be only allowing back to altair
+    return getForkSchedule().getForks().stream()
+        .map(Fork::getEpoch)
+        .filter(epoch -> atEpoch(epoch).getMilestone() == SpecMilestone.ALTAIR)
+        .min(UInt64::compareTo)
+        .orElse(UInt64.ZERO);
```
rolfyone: would be nice to be able to do this in a way that doesnt need to be maintained for each milestone...

ajsutton: Yeah this definitely doesn't look at all right to me. The `SpecMilestone` enum is ordered so you probably want to get the milestone at `slot` (I'd just use `fork(computeEpochAtSlot(slot))` and then use `SpecMilestone.getAllPriorMilestones` and take the last one (could actually simplify by adding a `Optional<SpecMilestone> SpecMilestone.getPreviousMilestone()` method which could do it based on the enum ordinal.

Then just look up the epoch that previous milestone is scheduled and that's the earliest epoch you can support.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -90,6 +90,8 @@ public Consumer<BeaconBlockBodyBuilder> createSelector(
     return bodyBuilder -> {
       final Eth1Data eth1Data = eth1DataCache.getEth1Vote(blockSlotState);
 
+      final UInt64 minimalViableEpoch = spec.computeMinimumViableEpoch(blockSlotState.getSlot());
```
ajsutton: nit: Probably rename this to `previousForkEpochStart`? Just want to try and explain why it's the minimum viable.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -101,22 +103,29 @@ public Consumer<BeaconBlockBodyBuilder> createSelector(
       final SszList<AttesterSlashing> attesterSlashings =
           attesterSlashingPool.getItemsForBlock(
               blockSlotState,
-              slashing -> !exitedValidators.containsAll(slashing.getIntersectingValidatorIndices()),
+              slashing ->
+                  !exitedValidators.containsAll(slashing.getIntersectingValidatorIndices())
+                      && spec.computeEpochAtSlot(slashing.getAttestation1().getData().getSlot())
+                          .isGreaterThanOrEqualTo(minimalViableEpoch),
```
ajsutton: nit: Could probably simplify this by calculating the start slot of minimalViableEpoch ahead of time and just comparing to that rather than having to convert slot to epoch here.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -288,6 +288,17 @@ private void validateOrDefaultEpoch() {
...
+        && spec.computeMinimumViableEpoch(spec.computeStartSlotAtEpoch(maybeEpoch.get()))
+            .isGreaterThan(epoch)) {
+      SUB_COMMAND_LOG.error(
+          String.format(
+              "The specified epoch %s (%s) is on a different milestone to the chains current epoch %s (%s), cannot continue.",
+              epoch,
+              spec.atEpoch(epoch).getMilestone(),
+              maybeEpoch.get(),
+              spec.atEpoch(maybeEpoch.get()).getMilestone()));
+      System.exit(1);
```
ajsutton: We shouldn't need this here because the REST API should reject the exit for us. If not we need to fix the REST API.

ajsutton (commented):
Looks good, mostly just nits and I don't think we want to add the validation to the exit subcommand as the REST API should do it.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -90,6 +90,10 @@ public Consumer<BeaconBlockBodyBuilder> createSelector(
     return bodyBuilder -> {
       final Eth1Data eth1Data = eth1DataCache.getEth1Vote(blockSlotState);
 
+      final UInt64 previousForkEpochStart =
+          spec.computePreviousForkEpochStart(blockSlotState.getSlot());
```
ajsutton, line 94: I think this is actually an epoch at which the fork starts (as opposed to the slot at the start of the epoch the previous fork is configured for).  In which case I think it should be called `previousForkStartEpoch`.  Or just `firstEpochOfPreviousFork` would work too.

ajsutton (approved):
LGTM. Just a suggested rename to avoid potential confusion.

benjaminion:
Is there a reason for not simply checking that the signature is valid with respect to our current fork rather than looking at slot numbers? That's the actual validity criterion and would probably use less code (though signature checking is a CPU overhead if we are concerned about that).

rolfyone:
Same or previous milestone, not just current fork, so if you're transitioning there'd be effectively epochs where it wouldn't work for a period

ajsutton:
> Is there a reason for not simply checking that the signature is valid with respect to our current fork rather than looking at slot numbers? That's the actual validity criterion and would probably use less code (though signature checking is a CPU overhead if we are concerned about that).

Yeah we're concerned about the CPU overhead.  And Ben's right about checking against the current `Fork` (ie the data structure) - it has the current and previous milestone details required to validate signatures.

PR #6095, opened 2022-08-18
Update DepositTreeSnapshot as of EIPs#5443 PR

### LINKED ISSUE #6062: Update deposit merkle tree snapshot format
### Description
Update deposit merkle tree snapshot as per https://github.com/ethereum/EIPs/pull/5443

zilm13:
## PR Description
Updated deposit merkle tree snapshot:
- Currently our route is `/teku/v1/beacon/deposit_snapshot`. Should I move it to the `/eth/v1/beacon/deposit_snapshot` according to the EIP-4881
- Combining `blockHeight` with `Eth1Data` lead to increase in vote and metric preparation work, should I rework it?

## Fixed Issue(s)

Fixes #6062 
## Documentation

## Changelog


beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/api/Eth1DataCachePeriodCalculator.java
```diff
@@ -36,7 +37,9 @@ public static UInt64 calculateEth1DataCacheDurationPriorToFollowDistance(
 
     // And we want to be able to create blocks for at least the past epoch
     cacheDurationSeconds += ((long) config.getSlotsPerEpoch()) * config.getSecondsPerSlot();
-    return UInt64.valueOf(cacheDurationSeconds);
+    // Some networks have very small ETH1_FOLLOW_DISTANCE but we may need to lookup Eth1Data block
+    // height from the current state
+    return UInt64.valueOf(cacheDurationSeconds).max(secondsPerEth1VotingPeriod * 2);
```
ajsutton: I don't quite understand this comment and why we need the cache duration to be a minimum of 2 * voting period, that seems very arbitrary.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
```diff
@@ -122,7 +123,15 @@ public void onNewFinalizedCheckpoint(
...
-        depositMerkleTree.finalize(finalizedState.getEth1Data());
+        final Optional<UInt64> heightOptional =
+            eth1DataCache
+                .findEth1DataWithHeight(finalizedState.getEth1Data())
+                .map(Eth1DataCache.Eth1DataAndHeight::getBlockHeight);
+        if (heightOptional.isEmpty()) {
+          LOG.warn("Eth1Data height not found in cache. Skipping DepositTree finalization");
+          return;
+        }
+        depositMerkleTree.finalize(finalizedState.getEth1Data(), heightOptional.get());
```
ajsutton: ah, this is pretty unfortunate... I don't think we want to assume that the finalized Eth1Data is still in our cache. I guess normally it will be but there's certainly nothing that would ensure that.

We can probably live with this but I suspect the right thing to do will be to try getting the data from the cache first, and if it's not present then request the block by hash from the EL again.  That will mean this method has an async element which will add some complexity unfortunately.  Maybe just log a ticket about it for now and we can see how it performs in reality and make a decision later.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/Eth1DataCache.java
```diff
@@ -163,16 +176,25 @@ protected Map<Eth1Data, Eth1Vote> countVotes(final BeaconState state) {
...
             eth1VotingPeriod.getSpecRangeUpperBound(slot, genesisTime),
             true),
-        eth1Data ->
-            eth1Data.getDepositCount().isGreaterThanOrEqualTo(dataFromState.getDepositCount()));
+        eth1DataAndHeight ->
+            eth1DataAndHeight
+                .getEth1Data()
+                .getDepositCount()
+                .isGreaterThanOrEqualTo(dataFromState.getDepositCount()));
   }
```
ajsutton: I'd suggest just transforming the values here so this still returns `NavigableMap<UInt64, Eth1Data>` and then most of the rest of the changes in this class won't be required:

```
private NavigableMap<UInt64, Eth1Data> getVotesToConsider(
      final UInt64 slot, final UInt64 genesisTime, final Eth1Data dataFromState) {
    final NavigableMap<UInt64, Eth1Data> unfiltered =
        Maps.transformValues(
            eth1ChainCache.subMap(
                eth1VotingPeriod.getSpecRangeLowerBound(slot, genesisTime),
                true,
                eth1VotingPeriod.getSpecRangeUpperBound(slot, genesisTime),
                true),
            Eth1DataAndHeight::getEth1Data);
    return Maps.filterValues(
        unfiltered,
        eth1DataAndHeight ->
            eth1DataAndHeight
                .getDepositCount()
                .isGreaterThanOrEqualTo(dataFromState.getDepositCount()));
  }
```

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/Eth1DataCache.java
```diff
@@ -163,16 +176,25 @@ protected Map<Eth1Data, Eth1Vote> countVotes(final BeaconState state) {
     return votes;
   }
 
-  private NavigableMap<UInt64, Eth1Data> getVotesToConsider(
+  protected Optional<Eth1DataAndHeight> findEth1DataWithHeight(final Eth1Data eth1Data) {
```
ajsutton: nit: I'd rename this to `getEth1DataAndHeight`

ethereum/pow/api/src/main/java/tech/pegasys/teku/ethereum/pow/api/DepositTreeSnapshot.java
```diff
@@ -19,62 +19,43 @@
...
+        schema,
+        finalized.stream().map(SszBytes32::of).collect(schema.getFinalizedSchema().collector()),
+        SszBytes32.of(depositRoot),
+        SszUInt64.of(UInt64.valueOf(depositCount)),
+        SszBytes32.of(executionBlockHash),
+        SszUInt64.of(executionBlockHeight));
   }
 
-  private DepositTreeSnapshot(final DepositTreeSnapshotSchema schema, final TreeNode node) {
+  protected DepositTreeSnapshot(final DepositTreeSnapshotSchema schema, final TreeNode node) {
```
ajsutton: This should stay private I think.

ethereum/pow/merkletree/src/main/java/tech/pegasys/teku/ethereum/pow/merkletree/DepositTree.java
```diff
@@ -19,19 +19,25 @@
...
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.ethereum.pow.api.DepositTreeSnapshot;
+import tech.pegasys.teku.ethereum.pow.api.DepositTreeSnapshotSchema;
 import tech.pegasys.teku.infrastructure.crypto.Hash;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.datastructures.blocks.Eth1Data;
 
 public class DepositTree {
+  public static final DepositTreeSnapshotSchema DEPOSIT_TREE_SNAPSHOT_SCHEMA =
+      new DepositTreeSnapshotSchema(DEPOSIT_CONTRACT_TREE_DEPTH);
```
ajsutton: I don't think it makes sense to have the SSZ snapshot schema stored here. It's certainly not where I'd go looking for it. If it's a constant it should just be in `DepositTreeSnapshot`, otherwise it would go into `SchemaDefinitions` but that's probably not right for this object as changes to it likely won't be tied to hard forks.

I think I'd just move the `DEPOSIT_CONTRACT_TREE_DEPTH` constant to something in `:ethereum:pow:api` since it's very pow service related anyway. And once you move that constant, it makes it possible to do all the other comments about where the schema instance is held and so on.

ethereum/pow/api/src/main/java/tech/pegasys/teku/ethereum/pow/api/DepositTreeSnapshot.java
```diff
@@ -19,62 +19,43 @@
...
-  public static DepositTreeSnapshot fromBytes(final Bytes bytes) {
-    return SCHEMA.sszDeserialize(bytes);
+  public static DepositTreeSnapshot fromBytes(
+      final Bytes bytes, final DepositTreeSnapshotSchema schema) {
+    return schema.sszDeserialize(bytes);
   }
 
   public DepositTreeSnapshot(
-      final List<Bytes32> finalized, final long deposits, final Bytes32 executionBlockHash) {
+      final DepositTreeSnapshotSchema schema,
```
ajsutton: We shouldn't have public constructors that take the schema as a param - this class should just know what it's schema is.

beacon/pow/build.gradle
```diff
@@ -1,5 +1,6 @@
 dependencies {
   implementation project(':ethereum:pow:api')
+  implementation project(':ethereum:pow:merkletree')
```
ajsutton: This winds up not being required if we put move the `DEPOSIT_CONTRACT_TREE_DEPTH` and leave the snapshot schema instance in `DepositTreeSnapshot` rather than in `DepositTree`.

data/beaconrestapi/build.gradle
```diff
@@ -1,6 +1,7 @@
 dependencies {
     implementation project(':beacon:validator')
     implementation project(':ethereum:pow:api')
+    implementation project(':ethereum:pow:merkletree')
```
ajsutton: This winds up not being required if we put move the `DEPOSIT_CONTRACT_TREE_DEPTH` and leave the snapshot schema instance in `DepositTreeSnapshot` rather than in `DepositTree`.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetDepositSnapshot.java
```diff
@@ -41,12 +42,16 @@
...
+ * Get Deposit Snapshot Tree, see <a
+ * href="https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4881.md">EIP-4881</a>
+ */
 public class GetDepositSnapshot extends MigratingEndpointAdapter {
 
   public static final String ROUTE = "/teku/v1/beacon/deposit_snapshot";
 
   private static final SerializableTypeDefinition<DepositTreeSnapshot> DEPOSIT_SNAPSHOT_TYPE =
-      DepositTreeSnapshot.getJsonTypeDefinition();
+      DepositTreeSnapshot.getJsonTypeDefinition(DEPOSIT_TREE_SNAPSHOT_SCHEMA);
```
ajsutton: It seems like the `DepositTreeSnapshot` and `DepositTreeSnapshotSchema` should be in the same module and so we shouldn't need to pass this in here.  The `DepositTreeSnapshot.getJsonDefinition()` is the API I think we want to stick with.

data/serializer/src/main/java/tech/pegasys/teku/api/response/v1/teku/DepositTreeSnapshotSchema.java
```diff
@@ -48,13 +56,26 @@ public class DepositTreeSnapshotSchema {
...
+
   @JsonCreator
   public DepositTreeSnapshotSchema(
       @JsonProperty("finalized") final List<Bytes32> finalized,
-      @JsonProperty("deposits") final UInt64 deposits,
-      @JsonProperty("execution_block_hash") final Bytes32 executionBlockHash) {
+      @JsonProperty("deposit_root") final Bytes32 depositRoot,
+      @JsonProperty("deposit_count") final UInt64 depositCount,
+      @JsonProperty("execution_block_hash") final Bytes32 executionBlockHash,
+      @JsonProperty("execution_block_height") final UInt64 executionBlockHeight) {
```
ajsutton: I didn't notice this earlier but I think this class should just be called `DepositTreeSnapshot`.  The schema is an SSZ concept and usually our API definition classes as just named after the object they provide in the API without a Schema prefix.

services/powchain/build.gradle
```diff
@@ -5,6 +5,7 @@ dependencies {
   implementation project(':ethereum:spec')
   implementation project(':ethereum:networks')
   implementation project(':ethereum:pow:api')
+  implementation project(':ethereum:pow:merkletree')
```
ajsutton: This also isn't required if we move the constant instead of the schema.

ethereum/pow/api/src/main/java/tech/pegasys/teku/ethereum/pow/api/DepositTreeSnapshot.java
```diff
@@ -19,62 +19,43 @@
...
-    return SCHEMA.getJsonTypeDefinition();
+  public static DeserializableTypeDefinition<DepositTreeSnapshot> getJsonTypeDefinition(
+      final DepositTreeSnapshotSchema schema) {
+    return schema.getJsonTypeDefinition();
   }
 
-  public static DepositTreeSnapshot fromBytes(final Bytes bytes) {
-    return SCHEMA.sszDeserialize(bytes);
+  public static DepositTreeSnapshot fromBytes(
+      final Bytes bytes, final DepositTreeSnapshotSchema schema) {
```
ajsutton: We won't need to pass in the schema here - it should just be held in this class.

ajsutton (commented):
Looks good generally, will just be a lot tidier if we change the way it avoids a circular dependency.

ethereum/pow/merkletree/src/main/java/tech/pegasys/teku/ethereum/pow/merkletree/DepositTree.java
zilm13: Yeah this way it looks way way better, updated

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
zilm13: Yeah I thought about calling EL for it but it's definitely an unwanted complexity. In normal flow we have this in cache. If not on some crazy testnet, it's not a big deal I think. But I will leave a ticket, ok, yeah

zilm13: Added #6102

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/api/Eth1DataCachePeriodCalculator.java
zilm13: Yeah, I was already sleeping and definitely wrong on this. I've recalculated it. Maybe I'm still wrong somewhere in the logic, but at least I added better comment. Could you, please, recheck it again? So I want to be sure we have state `Eth1Data`'s height after finalization

zilm13:
@ajsutton updated for review

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
```diff
@@ -122,7 +123,15 @@ public void onNewFinalizedCheckpoint(
       if (depositIndex.isGreaterThanOrEqualTo(finalizedState.getEth1Data().getDepositCount())
           && depositMerkleTree.getDepositCount()
               >= finalizedState.getEth1Data().getDepositCount().longValue()) {
-        depositMerkleTree.finalize(finalizedState.getEth1Data());
+        final Optional<UInt64> heightOptional =
+            eth1DataCache
+                .getEth1DataAndHeight(finalizedState.getEth1Data())
+                .map(Eth1DataCache.Eth1DataAndHeight::getBlockHeight);
+        if (heightOptional.isEmpty()) {
+          LOG.warn("Eth1Data height not found in cache. Skipping DepositTree finalization");
```
ajsutton: Just a suggestion to tweak this wording so the user doesn't freak out and think Teku is unable to process finalisation of the chain.

```suggestion
          LOG.warn("Eth1Data height not found in cache. Skipping DepositTree pruning");
```

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/api/Eth1DataCachePeriodCalculator.java
```diff
@@ -36,6 +37,13 @@ public static UInt64 calculateEth1DataCacheDurationPriorToFollowDistance(
 
     // And we want to be able to create blocks for at least the past epoch
     cacheDurationSeconds += ((long) config.getSlotsPerEpoch()) * config.getSecondsPerSlot();
+
+    // It's enough for voting, but we may need to be able to get data about candidate
+    // in the previous period for tree finalization plus 2 epochs for finalization
+    cacheDurationSeconds +=
+        secondsPerEth1VotingPeriod
+            + ((long) config.getSlotsPerEpoch()) * config.getSecondsPerSlot() * 2;
```
ajsutton: This change still makes me really nervous and means we're holding more stuff in memory which would be good to avoid unless we really need to.  What are the specific configurations where we'd need this change?

If it's only for particularly strange testnets, then its probably fine to just not wind up pruning the deposit tree - it was only recently that we implemented pruning at all so it's not an issue if the tree isn't pruned.

ajsutton (commented):
Looks good - just the cache period change makes me really nervous as its such an easy area to get wrong and surprisingly hard to spot its wrong until you get a network into odd situations and someone happens to notice the eth1 votes are wrong.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
zilm13: It's not only about pruning. If it fails we don't have object to reply with on Rest API, plus when we add persistence we will fail to store `DepositTreeSnapshot` on disk in this case.

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/api/Eth1DataCachePeriodCalculator.java
zilm13: Good point.
Actually, especially in profitable networks validators will try to use the most recent data. I've checked in the Mainnet, it uses `Eth1Data` in 8h to 14.8h from the current (from `eth1_follow_distance` to `eth1_follow_distance` + `eth1_voting_period`) so there shouldn't be any issues with old size cache on mainnet. There was an issue on Kiln with very small `eth1_follow_distance`, but Kiln will be closed soon and no other network has small `eth1_follow_distance`. Though it's proposed to make it as small as possible after the merge. If so we could solve it when it's an issue and not in advance.

Instead I've added condition to skip finalization when it's already done for this `Eth1Data`, so when running long in the network we will lookup `Eth1Data` in cache only once when it's closer to the current time.

zilm13:
@ajsutton removed cache changes, see comment

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
ajsutton: That's fine, it just means that we don't prune the deposit tree that we store on disk or provide from the API (and if we've never been able to run the pruning we won't have a snapshot at all but that's a situation that has to be handled anyway).  Finalization is an important word to users about the chain advancing so they'll freak out with the current message.

ajsutton (approved):
LGTM.

PR #6096, opened 2022-08-19
Updated api-proposer-config.json to include gas limit

rolfyone:
- both fields become Optional, and either can be set.

 - if they're both empty, the entry should be removed.

 - added a test to show removing an entry from a file

 - generally cleaned up a bit.

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

mehdi-aouadi (approved):

PR #6099, opened 2022-08-19
Remove handling genesis since remains unused

courtneyeh:
#5749

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -338,20 +337,17 @@ public void storeFinalizedState(BeaconState state, Bytes32 blockRoot) {
...
-            lastState -> {
-              final StateRootRecorder recorder =
-                  new StateRootRecorder(
-                      lastState.getSlot().increment(), updater::addFinalizedStateRoot, spec);
-              recorder.acceptNextState(state);
-            },
-            () -> updater.addFinalizedStateRoot(state.hashTreeRoot(), state.getSlot()));
-      }
+      final Optional<BeaconState> maybeLastState =
+          getLatestAvailableFinalizedState(state.getSlot().minus(ONE));
```
ajsutton: Just to be completely safe, I'd make this `minusMinZero(ONE)`. That way if we ever do get the genesis state attempting to be stored here, we'll still handle it fine, just wind up recording it's state root a second time but that's ok as the database will deduplicate it anyway.

ajsutton (approved):
LGTM.

Issue #6059, opened 2022-08-11
Handle Nethermind backwards sync


ajsutton:
### Description
We noticed sth like this on teku <> nethermind

09:18:09.096 FATAL - Exiting due to fatal error in Eth1DepositManager java.util.concurrent.CompletionException: tech.pegasys.teku.infrastructure.exceptions.FatalServiceFailureException: java.util.concurrent.CompletionException: tech.pegasys.teku.beacon.pow.exception.FailedToFindMinGenesisBlockException: Failed to retrieve min genesis block.  Check that your eth1 node is fully synced. (See log file for full stack trace)
Teku is shutting down
Thread pool did not terminate
and the problem is that we're still downloading old blocks and receipts
Nethermind is syncing to head at first and later we're downloading history of the chain. So we could response with null on old blocks

ajsutton:
This now works.

PR #6104, opened 2022-08-20
Avoid finding min genesis PoW block when genesis state already known

ajsutton:
## PR Description

Skip finding the PoW block that first satisfies the minimum genesis time condition when we already have an anchor state.

The min genesis time block is only required as part of the process to generate genesis so if we already have a BeaconState to start from, we don't need to generate the genesis state and thus don't need to find this block.

This improves compatibility with ELs (particularly Nethermind) that may not download all historic blocks or may backfill them after eth_syncing reports the node is in sync.

## Fixed Issue(s)
#6059 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6108, opened 2022-08-22
Ensure key manager api can detect validators that we owned early.

### LINKED ISSUE #6050: Key management API returns 404 on fee recipient
### Description

When calling `GET` on `eth/v1/validator/$__pubkey/feerecipient` for a `$__pubkey` that has no individual recipient set, I get a 404 with message `Fee recipient not found`

As per https://github.com/ethereum/keymanager-APIs/blob/master/apis/fee_recipient.yaml it should return the value of `--validators-proposer-default-fee-recipient` instead

Likewise, a `POST` gets me `Validator public key not found when attempting to set fee recipient.`

The keys were imported with the key manager API, they are not "read-only" file system keys

rolfyone:
partially addresses #6050

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #6115, opened 2022-08-23
Fix concurrency issue in LRUCache

zilm13:
## PR Description
This bug led to errors in resolving validator from gossip received attestation and maybe other errors. I don't want to synchronize `super` call in implementation leaving `default` in interface because it could lead to the same error in other implementations. Better to remove default at all I think. 
Also I think we could easily delete not-so-shiny test as the cause of the issue is clear.
## Documentation

## Changelog


Nashatyrev:
Good catch 👍 
The fix looks good to me

ajsutton (approved):
LGTM. That's a really good spot. Great work.

zilm13:
Removed test. It's 10 seconds and its useful only for regression, while it's a small chance we could make error in the same place next time

Issue #2716, opened 2020-09-02
Make 'bls' module Eth2 agnostic. 


Nashatyrev:
### Description

See #2706

> If the bls module is actually used and going to be used outside of Eth2 domain then may be it makes sense to slightly redesign the module API to make it totally DST neutral?

Nashatyrev:
Refactoring BLS should also include test fix to avoid #3175 
Temporary test fix is made in #3203

lucassaldanha:
Closing as not planned.

PR #6116, opened 2022-08-23
Make it possible to disable BLS verification for everything except Deposits

Nashatyrev:
## PR Description

When doing experiments it is sometimes desirable to disable BLS verification globally for performance reasons. However verifying Deposit signature is kind of special case: it is legal for a valid block to contain Deposit with invalid signature. Not being able to handle invalid deposit signature leads to consensus break. 

This PR is still a bit hacky way to handle this: replace `AbstractBlockProcessor.blsVerifyDeposit` static flag with `BLSSignatureVerifier` reference to be able to set the verifier to the working instance even if the BLS verification is globally disabled with `BLSConstants.disableBLSVerification()`. 

BTW this change reminded me that there is still pending BLS refactoring to make BLS code cleaner (also includes this issue #2716). Will try to address this in a separate PR later

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6118, opened 2022-08-23
Draft: Remote validator acceptance test fails when using the `--beacon-node-api-endpoints` parameter

jimmygchen:
## PR Description

Issue originally raised in discord. I modified the test `RemoteValidatorAcceptanceTest.shouldFailoverWhenPrimaryBeaconNodeGoesDown` and it fails when the `--beacon-node-api-endpoints` is used instead of `--beacon-node-api-endpoint`. Switching the order in the Validator client `@Option` makes both scenarios pass.

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
@@ -86,9 +86,20 @@ public class RemoteValidatorAcceptanceTest extends AcceptanceTestBase {
                     .withNetwork("swift")
                     .withInteropValidators(0, VALIDATOR_COUNT)
                     .withPrimaryBeaconNodeEventStreamReconnectAttemptPeriod(Duration.ofMillis(100))
-                    .withBeaconNodes(beaconNode, failoverBeaconNode));
+                    .withBeaconNodeApiEndpoint(beaconNode, failoverBeaconNode));
 
     validatorClient.start();
+
+    // validator using `--beacon-node-api-endpoints` instead of `--beacon-node-api-endpoint`
+    TekuValidatorNode validatorClient2 = createValidatorNode(
+            config ->
+                    config
+                            .withNetwork("swift")
+                            .withInteropValidators(0, VALIDATOR_COUNT)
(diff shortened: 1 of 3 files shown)


jimmygchen:
@StefanBratanov looks like my guess was wrong - the order of `names` in `@Options` does not matter, as the following tests pass before the change. (this is really testing the picocli rather than Teku, but just wanted to confirm)

```
package tech.pegasys.teku;

import picocli.CommandLine;

import org.junit.jupiter.api.Test;
import tech.pegasys.teku.cli.options.ValidatorClientOptions;

import static org.junit.jupiter.api.Assertions.assertEquals;

public class ValidatorClientOptionsTest {

    @Test
    public void testArgBeaconNodeApiEndpointSingular() {
        String[] args = {
                "--beacon-node-api-endpoint",
                "http://localhost1,http://localhost2"
        };

        ValidatorClientOptions app = new ValidatorClientOptions();
        new CommandLine(app).parseArgs(args);
        assertEquals(2, app.getBeaconNodeApiEndpoints().size());
    }

    @Test
    public void testArgBeaconNodeApiEndpointsPlural() {
        String[] args = {
                "--beacon-node-api-endpoints",
                "http://localhost1,http://localhost2"
        };

        ValidatorClientOptions app = new ValidatorClientOptions();
        new CommandLine(app).parseArgs(args);
        assertEquals(2, app.getBeaconNodeApiEndpoints().size());
    }
}
```

jimmygchen:
Will revert the change and keep the test to see if the test fails on CI. 

On my machine the acceptance test continues to run after 2+ minutes, but if I switch the order in `ValidatorClientOptions`, it passes after 39 seconds.

StefanBratanov:
> Will revert the change and keep the test to see if the test fails on CI.
> 
> On my machine the acceptance test continues to run after 2+ minutes, but if I switch the order in `ValidatorClientOptions`, it passes after 39 seconds.

Hey, thanks for looking into it. The reason the test fails is because in the TekuValidatorNode Config constructor we do the following which does not get overwritten if we use `beacon-node-api-endpoints` (because of priority order) :

`configMap.put("beacon-node-api-endpoint", "http://notvalid.restapi.com");`

In reality, this wouldn't be a problem because the user will define one of the two properties.

jimmygchen:
> In reality, this wouldn't be a problem because the user will define one of the two properties.

Ah I see, thank you! I will close this now :)

Issue #6020, opened 2022-08-04
Handle invalid blocks with no payload more smoothly


ajsutton:
### Description
After Bellatrix activates if Teku receives an Altair block it correctly refuses to import it but considers it an internal failure rather than an invalid block:
```
2022-08-04 22:26:01.899 | beaconchain-async-0       | ERROR | BlockImporter                  | Internal error while importing block: aa8201004942d3b246b51b1793ca5d5d49251c980d16c372f29cd65d4b3b6dd4 (3592325)
java.util.concurrent.CompletionException: java.util.NoSuchElementException: No value present
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315) ~[?:?]
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320) ~[?:?]
	at java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1159) ~[?:?]
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510) ~[?:?]
	at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2147) ~[?:?]
	at tech.pegasys.teku.dataproviders.generators.CachingTaskQueue.lambda$completePendingTask$9(CachingTaskQueue.java:199) ~[classes/:?]
	at tech.pegasys.teku.infrastructure.async.SafeFuture.fromRunnable(SafeFuture.java:154) ~[classes/:?]
	at tech.pegasys.teku.infrastructure.async.AsyncRunner.lambda$runAsync$0(AsyncRunner.java:23) ~[classes/:?]
	at tech.pegasys.teku.infrastructure.async.SafeFuture.of(SafeFuture.java:73) ~[classes/:?]
	at tech.pegasys.teku.infrastructure.async.ScheduledExecutorAsyncRunner.lambda$createRunnableForAction$1(ScheduledExecutorAsyncRunner.java:119) ~[classes/:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
Caused by: java.util.NoSuchElementException: No value present
	at java.util.Optional.orElseThrow(Optional.java:377) ~[?:?]
[...]

PR #6120, opened 2022-08-24
Bellatrix block processor no longer processes pre-bellatrix blocks

rolfyone:
Previously it was possible that a pre-bellatrix block could attempt to be loaded, and it would fail because it has no payload.

Partially addresses #6020

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #6122, opened 2022-08-24
Verify operation signatures when building blocks

ajsutton:
## PR Description
Explicitly verify operation signatures when selecting operations to include in blocks. Simpler and safer than checking the fork indirectly.

Change the fork we use when validating gossip to always use the chain head fork rather than the fork that would be in place at the slot the message is for. This aligns with the behaviour of others clients better and also means we're only accepting gossip that we could then include in the next block.  Potentially some gossip might be rejected that could have been included later after a hard fork activates but that's a pretty rare occurrence.

Reference tests for exits and slashing operations have been updated to not only check the block processing, but also confirm that the block creation filters accept or reject the operations correctly based on their validity.

To ensure we don't keep revalidating the signature of an operation that can no longer be included forever, operations are removed from the pool if they fail validation when creating a block.

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
@@ -90,10 +90,6 @@ public class BlockOperationSelectorFactory {
     return bodyBuilder -> {
       final Eth1Data eth1Data = eth1DataCache.getEth1Vote(blockSlotState);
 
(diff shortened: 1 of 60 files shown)


tbenr (approved):
LGTM

rolfyone (approved):
lgtm, maybe debug on remove as discussed

ajsutton:
Log message isn't going to work out very well because the `OperationPool` just has a generic type so doesn't have a good way to describe the item being dropped in the logs. Given we don't log when we drop things because they're included in blocks either I'm not going to fight it.

Issue #6043, opened 2022-08-08
Implement checkpoint sync api


rolfyone:
Two endpoints are required:
1. `GET /eth/v1/checkpoint/finalized_state`: Returns BeaconState object for a finalized checkpoint state from the WS period.An alias to `GET /eth/v2/debug/beacon/states/finalized`
2. `GET /eth/v1/checkpoint/finalized_blocks/{slot}/root`: Returns 404 if a block at a slot is either unavailable or not yet finalized. Otherwise, returns beacon block root. Essentially equivalent to the current `GET /eth/v1/beacon/blocks/{slot}/root` but with the additional requirement that it only returns a block root if the block is finalized (the provided slot is < to the the most recent finalized block slot).

The detailed specs are available [here](https://github.com/ethereum/beacon-APIs/pull/226)

- [x] Implement the `GET /eth/v1/checkpoint/finalized_state`
- [x] Implement the `GET /eth/v1/checkpoint/finalized_blocks/{slot}/root`

PR #6125, opened 2022-08-24
Add Get Finalized Checkpoint State API

mehdi-aouadi:
## PR Description

Adds a `GET /eth/v1/checkpoint/finalized_state` endpoint which returns a BeaconState object for finalized checkpoint state.

## Fixed Issue(s)

https://github.com/consensys/teku/issues/6043

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetFinalizedCheckpointState.java
```diff
@@ -0,0 +1,148 @@
...
+      final ChainDataProvider chainDataProvider,
+      final SchemaDefinitionCache schemaDefinitionCache) {
+    super(
+        EndpointMetadata.get(ROUTE)
+            .operationId("getFinalizedCheckpointState")
+            .summary("Get full BeaconState object for finalized checkpoint state")
+            .description(
+                "Returns full BeaconState object for a finalized checkpoint state from the WS period. \n"
+                    + "Depending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.")
+            .tags(TAG_BEACON)
```
ajsutton: We should add the experimental tag here as well.  @rolfyone What else should we do to make it clear this API may be changed or disappear without warning?

ajsutton (commented):
Looking good I think. The main thing will just be making it clear that this API isn't yet stable and may change.  In particular there's open questions about whether it should support JSON at all and what it should do if the state is only optimistically synced.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetFinalizedCheckpointState.java
mehdi-aouadi: > Looking good I think. The main thing will just be making it clear that this API isn't yet stable and may change. In particular there's open questions about whether it should support JSON at all and what it should do if the state is only optimistically synced.

Yep, for now I am keeping it as a draft PR while waiting for more clarifications... I'll adjust according to the ongoing discussion.

rolfyone: tagging as experimental without the beacon flag if we're really unsure of it's future... if we think its probably going to be pretty much right I'd leave beacon flag on it, but typically i only flag experimental until PRs are approved in beacon-apis...

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/paths/_eth_v1_checkpoint_finalized_state.json
```diff
@@ -0,0 +1,37 @@
+{
+  "get" : {
+    "tags" : [ "Beacon" ],
+    "summary" : "Get finalized checkpoint state",
+    "description" : "Returns full BeaconState object for a finalized checkpoint state from the WS period. \nDepending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.",
```
rolfyone: spell out `WS`.

data/serializer/src/main/java/tech/pegasys/teku/api/response/v2/beacon/GetFinalizedCheckpointStateResponse.java
```diff
@@ -0,0 +1,54 @@
...
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import tech.pegasys.teku.api.schema.Version;
+import tech.pegasys.teku.api.schema.altair.BeaconStateAltair;
+import tech.pegasys.teku.api.schema.bellatrix.BeaconStateBellatrix;
+import tech.pegasys.teku.api.schema.interfaces.State;
+import tech.pegasys.teku.api.schema.phase0.BeaconStatePhase0;
+
+@SuppressWarnings("JavaCase")
+public class GetFinalizedCheckpointStateResponse {
```
rolfyone: as far as i can tell this is identical to `GetStateResponseV2`, i'd be tempted to just use that unless there's a reason to publish the same object definition again?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedCheckpointState.java
```diff
@@ -0,0 +1,149 @@
...
+            "data",
+            getSchemaDefinitionForAllMilestones(
+                schemaDefinitionCache,
+                "BeaconState",
+                SchemaDefinitions::getBeaconStateSchema,
+                (beaconState, milestone) ->
+                    schemaDefinitionCache.milestoneAtSlot(beaconState.getSlot()).equals(milestone)),
+            ObjectAndMetaData::getData)
+        .build();
+  }
```
rolfyone: i'd be tempted to share the v2 get state here also...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedCheckpointState.java
```diff
@@ -0,0 +1,149 @@
...
+  public GetFinalizedCheckpointState(
+      final ChainDataProvider chainDataProvider,
+      final SchemaDefinitionCache schemaDefinitionCache) {
+    super(
+        EndpointMetadata.get(ROUTE)
+            .operationId("getFinalizedCheckpointState")
+            .summary("Get full BeaconState object for finalized checkpoint state")
+            .description(
+                "Returns full BeaconState object for a finalized checkpoint state from the WS period. \n"
+                    + "Depending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.")
```
rolfyone: we're currently returning the latest finalized state... this makes it sound way more complicated...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetFinalizedCheckpointState.java
mehdi-aouadi: The beacon-apis PR is not merged yet so I tagged it experimental.
Will update it to beacon when the PR is merged

data/serializer/src/main/java/tech/pegasys/teku/api/response/v2/beacon/GetFinalizedCheckpointStateResponse.java
mehdi-aouadi: I just removed the `execution_optimistic` flag from the `GetFinalizedCheckpointStateResponse`. We're discussing if we should keep it (It that case I'll use the existing `GetStateResponseV2`) or simply refuse to return a state if it's optimistically syncing.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedCheckpointState.java
mehdi-aouadi: Same here, I just removed the `execution_optimistic` field too which makes them slightly different

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedCheckpointState.java
mehdi-aouadi: Updated it to: Returns the latest finalized BeaconState.

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/beacon/GetFinalizedCheckpointStateIntegrationTest.java
```diff
@@ -0,0 +1,111 @@
...
+import tech.pegasys.teku.infrastructure.ssz.schema.SszSchema;
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.versions.altair.BeaconStateAltair;
+
+public class GetFinalizedCheckpointStateIntegrationTest
+    extends AbstractDataBackedRestAPIIntegrationTest {
+
+  @Test
+  public void shouldGetBellatrixStateAsJson() throws IOException {
```
ajsutton: hmm, the spec currently says this API can return JSON but I'm actually keen to see it only ever return SSZ as the JSON serialisation is expensive and supporting multiple formats makes it much harder to implement this API with a static web server (e.g have a cron job that periodically grabs a state from a real node and just saves it for a static server to serve).

So I think for our implementation we should just go with SSZ only since that's where we'd like to wind up.  If we support JSON now it may be very hard to get rid of later (and potentially causes confusion that with some versions of teku you get json and other ssz by default).

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/paths/_eth_v1_checkpoint_finalized_state.json
```diff
@@ -0,0 +1,56 @@
+{
+  "get" : {
+    "tags" : [ "Experimental" ],
+    "operationId" : "getFinalizedCheckpointState",
+    "summary" : "Get full BeaconState object for finalized checkpoint state",
+    "description" : "Returns full BeaconState object for a finalized checkpoint state from the Weak Subjectivity period. \nDepending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.",
```
ajsutton: We'd want to update this description to remove the reference to returning JSON as well.  And actually remove the application/json 200 OK response below too.

ajsutton (commented):
Looks good generally but I'd be keen to remove the JSON response type before we merge this. It doesn't match the current spec but I think it's a much better way to go and if we do lose that argument it will be less painful to add JSON later than to remove it.

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/beacon/GetFinalizedCheckpointStateIntegrationTest.java
```diff
@@ -0,0 +1,111 @@
...
+    assertThat(stateResponse.getData()).isInstanceOf(BeaconStateAltair.class);
+    assertThat(response.header(HEADER_CONSENSUS_VERSION)).isEqualTo(Version.bellatrix.name());
+  }
+
+  @Test
+  public void shouldGetBellatrixStateAsSsz() throws IOException {
+    startRestAPIAtGenesis(SpecMilestone.BELLATRIX);
+    final Response response = get();
+    assertThat(response.code()).isEqualTo(SC_OK);
+    assertThat(response.header(HEADER_CONSENSUS_VERSION)).isEqualTo(Version.bellatrix.name());
```
zilm13, line 46: It's not difficult to check body here btw. Check this code https://github.com/zilm13/teku/blob/19478cda9358bdf12f3a501bcbd9aafc793a51f7/data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/tekuv1/beacon/GetDepositSnapshotIntegrationTest.java#L82-L83

zilm13:
it would be good to add something to CHANGELOG, I think. It's a good addition to the API

ajsutton:
Probably not something to add to the changelog yet since it's still experimental.  But we should be sure to add it once it's not experimental anymore.

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/beacon/GetFinalizedCheckpointStateIntegrationTest.java
mehdi-aouadi: I see your point.
Removed it

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/beacon/GetFinalizedCheckpointStateIntegrationTest.java
mehdi-aouadi: Done

ajsutton (approved):
`tech.pegasys.teku.beaconrestapi.handlers.v1.beacon.GetFinalizedCheckpointStateTest` is failing but otherwise LGTM. Looking at the failure it looks like there's a mixup between the milestone (phase0/altair/bellatrix) in use somewhere so probably just a test issue, but worth double checking it really does still work in the real world once that's fixed.

mehdi-aouadi:
> `tech.pegasys.teku.beaconrestapi.handlers.v1.beacon.GetFinalizedCheckpointStateTest` is failing but otherwise LGTM. Looking at the failure it looks like there's a mixup between the milestone (phase0/altair/bellatrix) in use somewhere so probably just a test issue, but worth double checking it really does still work in the real world once that's fixed.

Fixed. The test was based on JSON response format plus had to return a `BeaconState` instead of `StateAndMetaData`

PR #6132, opened 2022-08-25
Add Get Finalized Block Root API

mehdi-aouadi:
## PR Description

Add the `GET /eth/v1/checkpoint/finalized_blocks/{slot}/root` endpoint which returns 404 if a block at a slot is either unavailable or not yet finalized. Otherwise, beacon block root

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6043

## Documentation

## Changelog


data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/paths/_eth_v1_checkpoint_finalized_blocks_{slot}_root.json
```diff
@@ -0,0 +1,19 @@
...
+      "in" : "path",
+      "required" : true,
+      "schema" : {
+        "type" : "string"
+      }
+    } ],
+    "responses" : {
+      "200" : {
+        "description" : "Default response"
+      }
```
rolfyone: seems not quite right

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
```diff
@@ -167,4 +169,9 @@ public class BeaconRestApiTypes {
                   + " Available values include: [`head`, `finalized_checkpoint`, `chain_reorg`, `block`, "
                   + "`attestation`, `voluntary_exit`, `contribution_and_proof`]\n\n",
               "head"));
+
+  public static final SerializableTypeDefinition<Bytes32> ROOT_TYPE =
+      SerializableTypeDefinition.object(Bytes32.class)
+          .withField("root", BYTES32_TYPE, Function.identity())
+          .build();
```
rolfyone, line 176: should re-use `BYTES32_TYPE` in CoreTypes

rolfyone: ah nevermind, i see. just re-using the one from block root.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedBlockRoot.java
```diff
@@ -0,0 +1,101 @@
...
+    this(dataProvider.getChainDataProvider());
+  }
+
+  public GetFinalizedBlockRoot(ChainDataProvider chainDataProvider) {
+    super(
+        EndpointMetadata.get(ROUTE)
+            .operationId("getFinalizedBlockRoot")
+            .summary("Get finalized block root")
+            .description(
+                "Retrieves hashTreeRoot of finalized BeaconBlock/BeaconBlockHeader.\n"
```
rolfyone: `of the finalized beacon block.`

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedBlockRoot.java
```diff
@@ -0,0 +1,101 @@
...
+                    .map(
+                        finalizedBlock -> {
+                          if (finalizedBlock.getSlot().equals(slot)) {
+                            chainDataProvider
+                                .getBlockRoot(slot.toString())
+                                .thenApply(
+                                    maybeFinalizedBlockRoot ->
+                                        maybeFinalizedBlockRoot
+                                            .map(AsyncApiResponse::respondOk)
+                                            .orElse(AsyncApiResponse.respondNotFound()));
```
rolfyone: if we have the whole block here, i don't see why we would re-fetch the root...

one possibility is we need to do more than 1 query to figure out the result, and we're trying to do some business logic at the api level?
my preference if we actually need 2 calls, is pass information to chainDataProvider, and it responds with `Optional<BlockRoot>`, and then the business logic is out of the api... 

Business logic should be ideally shared if we ever need it again, plus chainDataProvider is way easier to test than interfaces, but if we've got logic here then it's going to be complicated to prove it's what we want to do...

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -487,6 +487,11 @@ public Version getVersionAtSlot(final UInt64 slot) {
     return Version.fromMilestone(spec.atSlot(slot).getMilestone());
   }
 
+  public SafeFuture<Optional<SignedBeaconBlock>> getFinalizedBlockInEffectAtSlot(
+      final UInt64 slot) {
+    return combinedChainDataClient.getFinalizedBlockInEffectAtSlot(slot);
+  }
```
rolfyone: so this looks like there's a super simple function here, and instead we've got business logic in an api... 

make this so it returns an optional<Bytes32>, and it either responds with the block root or it responds with empty if the slot isn't finalized or whatever the logic is... but the logic will end up at this point...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedBlockRoot.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public void handle(@NotNull Context ctx) throws Exception {
+    adapt(ctx);
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
+    final UInt64 slot = request.getPathParameter(SLOT_PARAMETER);
+    final SafeFuture<Optional<SignedBeaconBlock>> eventuallyFinalizedBlock =
```
rolfyone: `futureFinalizedBlock`

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedBlockRoot.java
rolfyone: looking more into hte ticket etc, i think once we have finalizedBlockInEffect (first future), we then respond with block.getRoot(), rather than doing another query, we have the block we want, we just need to get the data from it I think...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedBlockRoot.java
ajsutton: I'd just inline it. :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetFinalizedBlockRoot.java
ajsutton: Yep so I think typically we'd just have something like:

```
maybeFinalizedBlock 
    .filter(block -> block.getSlot().equals(slot))
    .map(SignedBeaconBlock::getRoot)
    .map(AsyncApiResponse::responseOk)
    .orElseGet(AsyncApiRessponse::respondNotFound)
```

As Paul says, probably makes sense to push this logic into `ChainDataProvider` as a new `getFinalizedBlockRoot` method.

mehdi-aouadi: Done: moved the logic to the `ChainDataProvider` service and `getRoot` to avoid another query

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
mehdi-aouadi: Done: moved the logic to the ChainDataProvider service

ajsutton (approved):
LGTM.

Issue #5929, opened 2022-07-14
Add sentry node capability


benjaminion:
Background is [here](https://hackmd.io/2KXAIkIBR9Ki_yrfRbUH4g?view#Sentry-node-approach-short-term)

In short, the IP addresses of beacon nodes can be deduced from the their attestation activity. This would allow an attacker to DDoS a node that was due to make a block proposal, in order to gain MEV or otherwise disrupt the network. Sentry nodes (Cosmos terminology [afaics](https://forum.cosmos.network/t/sentry-node-architecture-overview/454)) would only publish blocks (from a different IP) and so remain anonymous.

Even shorter: the validator sends attestations via one BN, and block proposals via a different BN.

I have no idea how this interacts with MEV-Boost.

---
**Changes required:**
- [x] Read sentry node configuration file (based on format defined [here](https://hackmd.io/0jhD6wEzQLWnE3huXLkjgg#%E2%80%9CRole-centric%E2%80%9D-config))
- [x] Support reading config from local file or URL
- [x] Implement a new `SentryValidatorApiChannel` (name not final) that extends the `ValidatorApiChannel` and encapsulates the logic of choosing the right endpoint for each method. Possibly need a new implementation of `BeaconNodeApi`.
- [x] Update `ValidatorClientService` to consume sentry config (when enabled) and instantiate the correct `ValidatorApiChannel` instance.
- [x] Update cli options with new option for sentry config file
- [x] Refactor config deserializer to use a builder

**Out of scope (for now):**
- Live reload of sentry config file
- Minimise subnet subscriptions #6112

ajsutton:
There are quite a few options we may want to control when selecting which beacon node to use for specific activities. I've written up some thoughts at https://hackmd.io/0jhD6wEzQLWnE3huXLkjgg including a first stab at a config file that could support the required variations of configuring sentry nodes to avoid DoS attacks as well as potentially allowing us to support allocating specific beacon nodes to specific attestation subnets to avoid all nodes having to subscribe to all subnets.

ajsutton:
We should also consider how vouch can be configured to separate duties: https://github.com/attestantio/vouch/blob/master/docs/configuration.md

At first glance it looks roughly similar but I'm not sure of the impacts of the hierarchical nature and whether that's worth considering.

lucassaldanha:
[...]

PR #6133, opened 2022-08-25
Sentry node configuration object and loader

lucassaldanha:
## PR Description

Add sentry node configuration object and loader (from local file or URL).

Part of #5929


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/sentry/SentryNodesConfig.java
```diff
@@ -0,0 +1,51 @@
...
+  private final BeaconNodeRoleConfig dutiesProviderNodeConfig;
+  private final BeaconNodeRoleConfig blockHandlerNodeConfig;
+  private final BeaconNodeRoleConfig attestationPublisherConfig;
+
+  @JsonCreator
+  public SentryNodesConfig(
+      @JsonProperty(value = "duties_provider", required = true)
+          final BeaconNodeRoleConfig dutiesProviderNodeConfig,
+      @JsonProperty("block_handler") final BeaconNodeRoleConfig blockHandlerNodeConfig,
+      @JsonProperty("attestation_publisher")
```
rolfyone: we should probably be using `DeserializableTypeDefinition` here

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/sentry/SentryNodesConfigLoader.java
```diff
@@ -0,0 +1,107 @@
...
+import tech.pegasys.teku.provider.JsonProvider;
+
+public class SentryNodesConfigLoader {
+
+  private static final Logger LOG = LogManager.getLogger();
+  private final JsonProvider jsonProvider;
+  private final ResourceLoader resourceLoader;
+
+  public SentryNodesConfigLoader() {
+    this(new JsonProvider(), ResourceLoader.urlOrFile("application/json"));
```
rolfyone: we don't want to be creating a new JsonProvider, that will go away though if we use deserializableType, so i think we're likely going to do that...

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/sentry/SentryNodesConfigLoaderTest.java
```diff
@@ -0,0 +1,244 @@
...
+          + "        \"http://block:5051\"\n"
+          + "      ]\n"
+          + "    },\n"
+          + "    \"attestation_publisher\": {\n"
+          + "      \"endpoints\": [\n"
+          + "        \"http://attestation:5051\"\n"
+          + "      ]\n"
+          + "    }\n"
+          + "  }\n"
+          + "}";
```
rolfyone: these could probalby be resource files

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/sentry/SentryNodesConfigLoaderTest.java
```diff
@@ -0,0 +1,120 @@
...
+  public static final String SENTRY_NODE_MINIMUM_CONFIG_JSON =
+      "src/test/resources" + "/sentry_node_minimum_config.json";
+  public static final String SENTRY_NODE_MISSING_ENDPOINTS_CONFIG_JSON =
+      "src/test/resources" + "/sentry_node_missing_endpoints_config.json";
+  public static final String SENTRY_NODE_MISSING_DUTIES_PROVIDER_CONFIG_JSON =
+      "src/test" + "/resources/sentry_node_missing_duties_provider_config.json";
+  public static final String SENTRY_NODE_MISSING_BLOCK_HANDLER_CONFIG_JSON =
+      "src/test/resources" + "/sentry_node_missing_block_handler_config.json";
+  public static final String SENTRY_NODE_MISSING_ATTESTATION_PUBLISHER_CONFIG_JSON =
+      "src/test" + "/resources/sentry_node_missing_attestation_publisher_config.json";
```
rolfyone: there's an example of reading resources at `GetStateTest` in the `shouldHandle200`, using `Resources.getResource`
It might be cleaner? I'm not fussed either way particularly, but it's a fairly neat way to read resources and means its in the same route as your testing class etc...

rolfyone (approved):
LGTM, could use a builder in the deserializable if we want immutable config...

PR #6136, opened 2022-08-26
Log to console when the execution client is first confirmed as online.

ajsutton:
## PR Description
Previously we would only log if it was offline and came back online but it's reassuring for users to get an explicit log indicating that Teku connected to the execution client correctly.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6137, opened 2022-08-26
VC using failover nodes when retrieving network spec

### LINKED ISSUE #6117: Use fallback nodes when retrieving initial spec
### Description
When retrieving the initial spec, the validator client should use fallback beacon nodes to allow it to startup even when the first beacon node is offline.

StefanBratanov:
## PR Description

- Changed `RemoteSpecLoader` to retrieve spec when multiple nodes are configured
- Added tests to `ValidatorClientCommandTest`
- Added changes to the CHANGELOG

## Fixed Issue(s)
fixes #6117 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ For information on changes in released versions of Teku, see the [releases page]
 - Added Gas Limit APIs (GET/POST/DELETE)
 - Skip finding the PoW block that first satisfies the minimum genesis time condition when the genesis state is already known. Fixes an incompatibility with Nethermind's backwards sync for historic blocks.
 - Circuit breaker logic added when interacting with Builder endpoint
+- VC using failover nodes for retrieving the network spec when `--network=auto` and multiple beacon nodes are configured
 
 ### Bug Fixes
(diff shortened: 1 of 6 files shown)


tbenr (approved):
LGTM.
I see room for spec consistency check among endpoints here... like query the spec once per endpoint and do a consistency check. This might add a bit of complexity..
@ajsutton

ajsutton:
>I see room for spec consistency check among endpoints here... like query the spec once per endpoint and do a consistency check. This might add a bit of complexity..

Not a bad idea, though definitely not something I'd add in this PR.  Probably worth logging a ticket for it to be considered later.  The key problems I can see would be:

1. the config for a beacon node might change at a later point anyway
2. comparing the config is potentially tricky. There are a number of ways we could get equivalent configs that aren't identical and custom comparison is going to be too hard to maintain as new options are added
3. The config may be different because one node has been updated and has a future hard fork scheduled and the others don't yet.  Staggering the node upgrades is a good idea we wouldn't want to discourage (but at other times, different hard fork schedules would be a red flag).

PR #6139, opened 2022-08-27
Improve peer search-connect task on startup

zilm13:
## PR Description
The issue is well understandable from these logs:
1st attempt, in milk
```
2022-08-27 17:04:15.403+04:00 | nioEventLoopGroup-2-1 | INFO  | teku-status-log | Listening for connections on: /ip4/127.0.1.1/tcp/9000/p2p/16Uiu2HAkvKzJGsWAgqUEsbJf1botdLqxPzJbX8Bx5B2iBCf1MpuJ
2022-08-27 17:04:15.428+04:00 | nioEventLoopGroup-4-1 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 17:04:15.435+04:00 | nioEventLoopGroup-4-1 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@1603e783, with 0 size input
```
2nd, immediately after the first, in milk too
```
2022-08-27 17:04:15.436+04:00 | discovery-tasks-1 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 17:04:15.438+04:00 | discovery-tasks-1 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@1603e783, with 0 size input
```
3rd, +43 seconds, +36 candidates
```
2022-08-27 17:04:58.100+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 17:04:58.114+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@1603e783, with 36 size input
```

4th, +27 seconds + 14 candidates
```
2022-08-27 17:05:25.618+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 17:05:25.626+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@1603e783, with 50 size input
```

What's wrong with it:
- we don't have any peers from discovery on the 1st peer search-connect attempt in tthe `ConnectionManager`
- 2nd attempt goes in a few millis after the 1st, looks like a bug
- 3rd attempt is in 30 seconds, finally we are doing something

As a result, we almost always don't have any peers for the first 30 seconds after startup. It's 2.5 slots on Mainnet, which could be critical for real peers on any restart, as we couldn't gossip anything during this. Plus we are failing Hive tests due to these skipped blocks. I'm trying to solve it.

At first I did a code with (1, 2, 4, 8, 16, 30, 30, 30, ...) delays of search, but it looked complicated. As discovery call always took some unpredictable time, I think it's not critical to have valuable delay there, just some 1-second delays between discovery calls is ok, just not to throw out whole load immediately by error. So I did it with  (1, 1, 1, 1, 1, 30, 30, ...) delays and removed 2nd attempt, which looked definitely buggy.
Now the same startup looks like this. Please note: we are not getting more peers, but we have some peers earlier than before, it's a goal.
1st attempt
```
2022-08-27 22:44:56.495+04:00 | beaconchain-async-1 | INFO  | Javalin | Listening on http://localhost:5051/
2022-08-27 22:44:56.370+04:00 | p2p-async-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:44:56.376+04:00 | p2p-async-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 0 size input
```

2nd attempt, +14 seconds, +17 candidates
```
2022-08-27 22:45:09.190+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:45:09.208+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 17 size input
```

3rd attempt, +10 seconds, +6 candidates
```
2022-08-27 22:45:20.385+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:45:20.389+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 23 size input
```

4th attempt, +12 seconds, +10 candidates
```
2022-08-27 22:45:32.838+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:45:32.843+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 33 size input
```

5th attempt, +3 seconds, +4 candidates
```
2022-08-27 22:45:35.761+04:00 | nioEventLoopGroup-4-1 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:45:35.763+04:00 | nioEventLoopGroup-4-1 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 37 size input
```

6th attempt, +13 seconds, +8 candidates
```
2022-08-27 22:45:48.688+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:45:48.691+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 45 size input
```

7th attempt, +42 seconds, +25 candidates (we've created task after discovery search await, so it's not exactly 30 seconds, we could improve it with 2 cancellable objects, but I don't want to over-complicate)
```
2022-08-27 22:46:30.111+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:46:30.114+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 70 size input
```

8th attempt, +30 seconds, +3 candidates (it goes every 30 seconds now like it should)
```
2022-08-27 22:46:59.985+04:00 | discovery-expiration-0 | INFO  | ConnectionManager | connectToBestPeers: additionalPeersToConsider of size 0
2022-08-27 22:46:59.989+04:00 | discovery-expiration-0 | INFO  | Eth2PeerSelectionStrategy | selectPeersToConnect:tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork@519952fa, with 73 size input
```
## Documentation

## Changelog


zilm13:
- Cleaned up, fixed test, added new test
- I'm a bit paranoid on race while switching from short delay to long, maybe I'm not right and all will be ok, but if we miss long run task setup, it will make peer stall, so if you see a better way there, I'd like to hear
- `ConnectionManagerTest` becomes a bit more noisy, but previously it was not actually testing some cases, so the question about 2 disabled test: what behavior we expect from `Pools` to be on par with test

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -57,9 +60,11 @@ public class ConnectionManager extends Service {
   private final Counter failedConnectionCounter;
   private final PeerPools peerPools = new PeerPools();
   private final Collection<Predicate<DiscoveryPeer>> peerPredicates = new CopyOnWriteArrayList<>();
+  private final AtomicInteger startupCountdownStage = new AtomicInteger(5);
 
   private volatile long peerConnectedSubscriptionId;
   private volatile Cancellable periodicPeerSearch;
+  private volatile CountDownLatch activeSearch = new CountDownLatch(1);
```
ajsutton: It would probably be better to have this be returned by `searchForPeers` and passed into `connectToBestPeers` rather than stored as a field. That way we know it's correctly tied to a specific search process and can't wind up being updated by multiple searches that may wind up executing at once (shouldn't happen but better to guarantee it and I think the code will wind up being easier to follow as well).

ajsutton: The other approach which is more how we'd normally do it would be fore `searchForPeers` to return a future that only completes when the search process is complete.  That would let us avoid blocking a thread while waiting for the search process which would be better as the asyncRunner is typically shared and has a limited number of threads.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -57,9 +60,11 @@ public class ConnectionManager extends Service {
   private final Counter failedConnectionCounter;
   private final PeerPools peerPools = new PeerPools();
   private final Collection<Predicate<DiscoveryPeer>> peerPredicates = new CopyOnWriteArrayList<>();
+  private final AtomicInteger startupCountdownStage = new AtomicInteger(5);
```
ajsutton: It may make sense to search more frequently while we have less than some number of peers rather than a fixed number of times? Maybe even just switch to the slower schedule once we have some peers.

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
```diff
@@ -204,42 +203,51 @@ public void shouldConnectToKnownPeersWhenStarted() {
...
     manager.start().join();
+    advanceTimeABit();
 
     verify(network).connect(PEER1);
     verify(network).connect(PEER2);
   }
 
   @Test
+  @Disabled("Shouldn't work, disabled until clarifying")
   public void shouldNotRetryConnectionsToDiscoveredPeersOnFailure() {
```
ajsutton: I'm not entirely sure what needs clarifying here.  We shouldn't try to make a connection to a peer we've already connected to and seen fail - otherwise we just waste our time connecting to peers we already know won't work.

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
```diff
@@ -204,42 +203,51 @@ public void shouldConnectToKnownPeersWhenStarted() {
...
     connectionFuture.completeExceptionally(new RuntimeException("Failed"));
 
-    asyncRunner.executeQueuedActions();
+    advanceTimeByInitialInterval();
     verify(network, times(1)).connect(PEER1); // No further attempts to connect
   }
 
   @Test
+  @Disabled("Shouldn't work, disabled until clarifying")
   public void shouldNotRetryConnectionsToDiscoveredPeersOnDisconnect() {
```
ajsutton: Similarly this is just that we should track peers that have been disconnected and not try to establish connections to them again (because it's likely to just result in being disconnected again).

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
```diff
@@ -204,42 +203,51 @@ public void shouldConnectToKnownPeersWhenStarted() {
...
     final ConnectionManager manager = createManager();
-    when(discoveryService.streamKnownPeers()).thenReturn(Stream.of(DISCOVERY_PEER1));
+    when(discoveryService.streamKnownPeers())
+        .thenReturn(Stream.of(DISCOVERY_PEER1))
+        .thenReturn(Stream.of(DISCOVERY_PEER1));
```
ajsutton: Possibly this is the issue because the search is now happening more than twice? We'd need to return a new stream from each search so could try making this a `thenAnswer` so it always creates a fresh Stream.

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
zilm13: I mean clarifying, how and when it was working. I guess it was sometihing using pools but not getting whole idea. It's not working in the current master

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
zilm13: It's happening twice, yeah, we'd better create new stream from one statement

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
ajsutton: I'm not sure what you mean by it's not working in the current master.  The test is enabled and passing in the current master. The `ConnectionManager.isPeerValid` method does the filtering to ensure we don't retry connections we shouldn't.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
zilm13: I'd opt in to keep number of times as we could have networks with small total number of peers such as test networks or even private networks where behavior based on the predefined number of peers could be very buggy

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
zilm13: - test cuts off on the future return in the master, it doesn't reach the point where we are attempting to connect to the discovered peers for the second time
- `ConnectionManager.isPeerValid` tests only fork currently

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
zilm13: Tried it with the `SafeFuture` return, looks better

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
ajsutton: Ah right, so we need to update the test to make sure it triggers the case.  This bit of filtering actually turns out to be done in `Eth2PeerSelectionStrategy.checkCandidate` so maybe the test we need should be for `Eth2PeerSelectionStrategy` rather than here.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
ajsutton: I think if you have discovery enabled and no peers then retrying frequently makes sense. So slowing down once we have some peers would seem pretty reasonable.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ For information on changes in released versions of Teku, see the [releases page]
 ### Breaking Changes
 
 ### Additions and Improvements
+- Speeds up gossip peer discovery on startup
```
ajsutton:
```suggestion
- Faster peer discovery on startup
```

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -91,17 +94,40 @@ protected SafeFuture<?> doStart() {
...
-            DISCOVERY_INTERVAL,
-            error -> LOG.error("Error while searching for peers", error));
-    connectToBestPeers(emptyList());
-    searchForPeers();
+    startSearchPeerTask();
     peerConnectedSubscriptionId = network.subscribeConnect(this::onPeerConnected);
     return SafeFuture.COMPLETE;
   }
 
+  @SuppressWarnings("FutureReturnValueIgnored")
```
ajsutton: We shouldn't be ignoring these returned futures - it would silently ignore any exceptions they complete with.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -91,17 +94,40 @@ protected SafeFuture<?> doStart() {
...
+  private void startSearchPeerTask() {
+    this.periodicPeerSearch =
+        asyncRunner.runCancellableAfterDelay(
+            () -> searchForPeers().thenPeek(__ -> createFollowupSearchPeerTask()), Duration.ZERO);
```
ajsutton: Not sure why this is using `thenPeek`. It means the `createFollowupSearchPeerTask` won't execute of the future from `searchForPeers` completes exceptionally which would then mean discovery halts entirely.

Also, it's very weird to be saying run after delay but having a zero delay. Given we only cancel on shutdown and the cancel will only work if the task hasn't already started executing, I suspect we should just leave `periodicPeerSearch` as `null` here and call `searchForPeers()` without going through the async executor at all.

So it winds up something like:
```java
searchForPeers()
        .alwaysRun(this::createFollowupSearchPeerTask)
        .finish(error -> LOG.error("Error while searching for peers", error));
```

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -91,17 +94,40 @@ protected SafeFuture<?> doStart() {
...
+  private void createFollowupSearchPeerTask() {
+    final int currentStage = startupCountdownStage.getAndDecrement();
+    if (currentStage < 0) {
+      return;
+    }
+    if (currentStage > 0) {
+      // Startup task, run for [startupCountdownStage] times
+      this.periodicPeerSearch =
+          asyncRunner.runCancellableAfterDelay(
+              () -> searchForPeers().thenPeek(__ -> createFollowupSearchPeerTask()),
```
ajsutton: This probably should use `alwaysRun(...).finish` like above. In fact it seems to be exactly the same code so best to extract it as a new method and reuse it.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
zilm13: I think we have 3 options:
1. 5 fast retries unconditionally
2. Fast retries until certain number of peers reached. 10? I think even 10 could be not enough for exact subnet but setting more could force it to hit frequently forever
3. So, combining 2 something like: after 50 peers or 3 minutes switch to 30 seconds retries

I did first as it looks simple and works most times. I don't like 2nd. 3rd looks a bit more complex but always good in behavior. Do you think I should do something like 3rd option?

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
zilm13: I see both cases are covered in `ReputationManagerTest`, just removed them from `ConnectionManagerTest`

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
zilm13: Ok, I've tried 3rd option, check it

zilm13:
@ajsutton reworked, please check

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -91,17 +98,37 @@ protected SafeFuture<?> doStart() {
...
       staticPeers.forEach(this::createPersistentConnection);
     }
-    periodicPeerSearch =
-        asyncRunner.runWithFixedDelay(
-            this::searchForPeers,
-            DISCOVERY_INTERVAL,
-            error -> LOG.error("Error while searching for peers", error));
-    connectToBestPeers(emptyList());
-    searchForPeers();
+    asyncRunner.runAsync(this::createRecurrentSearchTask).ifExceptionGetsHereRaiseABug();
```
ajsutton: Why are we running this async? The `searchForPeers` done in `createRecurrentSearchTask` already starts an async process and returns quickly.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
ajsutton: Ultimately the problem we're trying to solve is that we take too long to start finding peers because the discovery system hasn't had time to start filling tables before the first time we ask it for peers to try to connect to, and then we don't ask for peers again for another 30 seconds (there's that slightly odd second attempt immediately after the first but it hits the same problem that discovery hasn't received any responses yet).

So we really only need to try rapidly until we find the first peer, at which point we know discovery has found some peers. You can see from the logs in the description, we're not finding many more peers between each rapid search after that first batch, so we're not gaining much by continuing to run the search so often and can back off and just run at the usual rate.  If we keep running discovery rapidly for too long, we burn a fair bit of extra CPU and could wind up getting ignored by other nodes for requesting neighbours too often.

That's why I think something really simple like retry fast until we find the first peer makes sense. Nice and simple and solves the slow start problem we're actually targeting.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
zilm13: Nice point

zilm13:
@ajsutton reworked to wait until first initiated connection but it has one flaw from my side: if, say, we have test/private network setup with 2 nodes and one is passive, just waiting for connection from other node, warmup finish will be never triggered

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -91,17 +94,36 @@ protected SafeFuture<?> doStart() {
...
+  }
+
+  private void logSearchError(final Throwable throwable) {
+    LOG.error("Error while searching for peers", throwable);
+  }
+
+  private void createNextSearchPeerTask() {
+    if (warmupDiscovery.get()) {
+      // Retry fast until first peer connection is initiated, which means Discovery is warmed up
+      this.periodicPeerSearch =
```
ajsutton: The condition we want is that we retry frequently until the first peer is found, so I'd just make the if condition `if (network.getPeerCount() > 0)`. That would then solve the issue you mention of a 2 node network as well.

It's possible we'll get an inbound peer quickly and so switch to the slower searching before discovery is fully ready but if we already have a peer that's not a big deal as we will already be able to start catching up again.

zilm13: Ok. So I added cancel of the last task as it could happen in the middle, which is not bad I think, but we shouldn't run tasks simulatenously

ajsutton (approved):
LGTM.

PR #6141, opened 2022-08-28
Utilise progressive balances in more ways

ajsutton:
## PR Description
Uses information from progressive balances to identify the future justified checkpoints in more places.  Toggled off by default for now.

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
@@ -620,15 +620,13 @@ public class ProtoArray {
     }
 
     if (progressiveBalancesMode.isFull()) {
-      final boolean correctJustified;
-      final boolean correctFinalized;
       final boolean isPreviousEpochJustified =
           justifiedCheckpoint.getEpoch().plus(1).isGreaterThanOrEqualTo(currentEpoch);
-      correctJustified =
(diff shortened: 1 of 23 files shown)


storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -605,8 +618,30 @@ public boolean nodeIsViableForHead(ProtoNode node) {
...
+        correctJustified =
+            node.getJustifiedCheckpoint()
+                .getEpoch()
+                .isGreaterThanOrEqualTo(justifiedCheckpoint.getEpoch());
+        correctFinalized = correctJustified;
+      } else {
+        correctJustified =
+            node.getJustifiedCheckpoint()
+                .getEpoch()
+                .isGreaterThanOrEqualTo(justifiedCheckpoint.getEpoch());
```
rolfyone: looks like this code is common, could be done outside the inner if?

rolfyone (approved):
LGTM, just wonder if we simplify one thing...

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
ajsutton: Yeah can do.

PR #6152, opened 2022-09-02
Return unmodifiable view for private mutable fields

jtraglia:
## PR Description

Sorry, this PR touches a lot of different files. The changes are pretty minor though, hope that makes it okay.

In the spirit of defensive programming, when a public method returns a private field that is mutable (`List`, `Set`, `Map`, etc.), I think we should return an unmodifiable view of that field. My intention is to help prevent the modification of private variables outside of the class. There were a couple instances of this happening in some tests. For example:

```java
storageSystem.chainHeadChannel().getHeadEvents().clear();
```

`getHeadEvents` returns a reference to a private field. To fix this, I created a new method, `clearHeadEvents` with the `@VisibleForTesting` decorator, that clears the clears the events internally.

I chose to use `Collections.unmodifiable*` instead of using immutable objects because immutable objects are a lot slower. The unmodifiable view just ["replaces" the modification methods with `UnsupportedOperationException`](https://stackoverflow.com/a/6604553), therefore the performance penalty will be very minor, if anything. There were two places that I decided to create unmodifiable copies; because the field was a list of lists or list of maps. It would have been possible to modify the entries.

These were identified with a pretty simple error-prone check that I just made:
 * https://github.com/ConsenSys/errorprone-checks/pull/10

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
+++ b/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
@@ -26,6 +26,7 @@ import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -277,7 +278,7 @@ public class TekuValidatorNode extends Node {
     }
 
     public Map<File, String> getConfigFileMap() {
-      return configFileMap;
+      return Collections.unmodifiableMap(configFileMap);
     }
   }
 }
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeers.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeers.java
@@ -28,6 +28,7 @@ import io.javalin.plugin.openapi.annotations.HttpMethod;
 import io.javalin.plugin.openapi.annotations.OpenApi;
 import io.javalin.plugin.openapi.annotations.OpenApiContent;
 import io.javalin.plugin.openapi.annotations.OpenApiResponse;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -155,7 +156,7 @@ public class GetPeers extends MigratingEndpointAdapter {
     }
 
     public List<Eth2Peer> getPeers() {
-      return peers;
+      return Collections.unmodifiableList(peers);
     }
 
     public Integer getCount() {
--- a/data/serializer/src/main/java/tech/pegasys/teku/api/migrated/StateSyncCommitteesData.java
+++ b/data/serializer/src/main/java/tech/pegasys/teku/api/migrated/StateSyncCommitteesData.java
@@ -13,6 +13,8 @@
 
 package tech.pegasys.teku.api.migrated;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
@@ -28,11 +30,15 @@ public class StateSyncCommitteesData {
   }
 
   public List<UInt64> getValidators() {
-    return validators;
+    return Collections.unmodifiableList(validators);
   }
 
   public List<List<UInt64>> getValidatorAggregates() {
-    return validatorAggregates;
+    final List<List<UInt64>> copy = new ArrayList<>();
+    for (List<UInt64> validatorAggregate : validatorAggregates) {
+      copy.add(Collections.unmodifiableList(validatorAggregate));
+    }
+    return Collections.unmodifiableList(copy);
   }
 
   @Override
--- a/ethereum/pow/api/src/main/java/tech/pegasys/teku/ethereum/pow/api/DepositsFromBlockEvent.java
+++ b/ethereum/pow/api/src/main/java/tech/pegasys/teku/ethereum/pow/api/DepositsFromBlockEvent.java
@@ -17,6 +17,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.stream.Collectors.toList;
 
 import com.google.common.base.MoreObjects;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
@@ -93,7 +94,7 @@ public class DepositsFromBlockEvent {
   }
 
   public List<Deposit> getDeposits() {
-    return deposits;
+    return Collections.unmodifiableList(deposits);
   }
 
   @Override
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigPhase0.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigPhase0.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.spec.config;
 
+import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 import org.apache.tuweni.bytes.Bytes;
@@ -210,7 +211,7 @@ public class SpecConfigPhase0 implements SpecConfig {
 
   @Override
   public Map<String, Object> getRawConfig() {
-    return rawConfig;
+    return Collections.unmodifiableMap(rawConfig);
   }
 
   @Override
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/status/ValidatorStatuses.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/status/ValidatorStatuses.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.spec.logic.common.statetransition.epoch.status;
 
+import java.util.Collections;
 import java.util.List;
 
 public class ValidatorStatuses {
@@ -30,7 +31,7 @@ public class ValidatorStatuses {
   }
 
   public List<ValidatorStatus> getStatuses() {
-    return statuses;
+    return Collections.unmodifiableList(statuses);
   }
 
   public int getValidatorCount() {
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregateAttestationBuilder.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregateAttestationBuilder.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.statetransition.attestation;
 import static com.google.common.base.Preconditions.checkState;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -79,6 +80,6 @@ class AggregateAttestationBuilder {
   }
 
   public Collection<ValidateableAttestation> getIncludedAttestations() {
-    return includedAttestations;
+    return Collections.unmodifiableCollection(includedAttestations);
   }
 }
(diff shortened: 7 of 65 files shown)


acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
```diff
@@ -277,7 +278,7 @@ public void publishSelfSignedCertificate(final String targetKeystoreFile)
     }
 
     public Map<File, String> getConfigFileMap() {
-      return configFileMap;
+      return Collections.unmodifiableMap(configFileMap);
```
ajsutton, line 281: I'm not sure we really gain anything by applying this rule to test code. It can be really helpful to be much more open with data structures in tests and we want to minimise the amount of boiler plate code we wind up adding there.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetPeers.java
```diff
@@ -155,7 +156,7 @@ static class PeersData {
     }
 
     public List<Eth2Peer> getPeers() {
-      return peers;
+      return Collections.unmodifiableList(peers);
```
ajsutton, line 159: This isn't required - the entire `PeersData` class this is in is a temporary object to hold the returned data.  I suspect the same is true for most of the cases through the `beaconrestapi` package.

data/serializer/src/main/java/tech/pegasys/teku/api/migrated/StateSyncCommitteesData.java
```diff
@@ -28,11 +30,15 @@ public StateSyncCommitteesData(
...
+    return Collections.unmodifiableList(validators);
   }
 
   public List<List<UInt64>> getValidatorAggregates() {
-    return validatorAggregates;
+    final List<List<UInt64>> copy = new ArrayList<>();
+    for (List<UInt64> validatorAggregate : validatorAggregates) {
+      copy.add(Collections.unmodifiableList(validatorAggregate));
+    }
+    return Collections.unmodifiableList(copy);
```
ajsutton, line 41: This class is just a wrapper so we can return multiple items and the data is calculated so there's no need to take any defensive action. Same for all the other items in `tech.pegasys.teku.api.migrated`.

ethereum/pow/api/src/main/java/tech/pegasys/teku/ethereum/pow/api/DepositsFromBlockEvent.java
```diff
@@ -93,7 +94,7 @@ public UInt64 getBlockTimestamp() {
   }
 
   public List<Deposit> getDeposits() {
-    return deposits;
+    return Collections.unmodifiableList(deposits);
```
ajsutton, line 97: We'd probably be better off doing this in the constructor rather than creating a new unmodifiable wrapper on every get call.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigPhase0.java
```diff
@@ -210,7 +211,7 @@ public SpecConfigPhase0(
 
   @Override
   public Map<String, Object> getRawConfig() {
-    return rawConfig;
+    return Collections.unmodifiableMap(rawConfig);
```
ajsutton, line 214: Probably should just build an unmodifiable Map in the `SpecConfigBuilder` for this one.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/status/ValidatorStatuses.java
```diff
@@ -30,7 +31,7 @@ public TotalBalances getTotalBalances() {
   }
 
   public List<ValidatorStatus> getStatuses() {
-    return statuses;
+    return Collections.unmodifiableList(statuses);
```
ajsutton, line 34: This probably should be made unmodifiable back in `AbstractValidatorFactory.createInitialValidatorStatuses` where it can just use `.collect(Collectors.toUnmodifiableList());`.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregateAttestationBuilder.java
```diff
@@ -79,6 +80,6 @@ public ValidateableAttestation buildAggregate() {
   }
 
   public Collection<ValidateableAttestation> getIncludedAttestations() {
-    return includedAttestations;
+    return Collections.unmodifiableCollection(includedAttestations);
   }
```
ajsutton, line 84: We can actually just delete this method - it's only used in a test that tests this method.

infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsConfig.java
```diff
@@ -92,11 +93,11 @@ public String getMetricsInterface() {
   }
 
   public Set<MetricCategory> getMetricsCategories() {
-    return metricsCategories;
+    return Collections.unmodifiableSet(metricsCategories);
   }
 
   public List<String> getMetricsHostAllowlist() {
-    return metricsHostAllowlist;
+    return Collections.unmodifiableList(metricsHostAllowlist);
```
ajsutton, line 100: Probably should make these unmodifiable in the builder.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -173,7 +173,7 @@ public Optional<String> getSecurity() {
   }
 
   public List<String> getTags() {
-    return tags;
+    return Collections.unmodifiableList(tags);
```
ajsutton, line 176: Probably worth putting in the builder.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/AbstractSszMutablePrimitiveCollection.java
```diff
@@ -135,7 +136,7 @@ public long getNodeGIndex() {
     }
 
     public List<PackedNodeUpdate<ElementT, SszElementT>> getUpdates() {
-      return updates;
+      return Collections.unmodifiableList(updates);
```
ajsutton, line 139: hmm, this is a particularly performance sensitive area and this is going to wind up creating a lot more garbage than you'd probably expect.  Given the class exposes mutation methods anyway, I don't think making this getter return an immutable version is going to be worth it.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszContainerSchema.java
```diff
@@ -246,7 +247,7 @@ public int getSszVariablePartSize(TreeNode node) {
 
   @Override
   public List<SszSchema<?>> getFieldSchemas() {
-    return childrenSchemas;
+    return Collections.unmodifiableList(childrenSchemas);
```
ajsutton, line 250: We should just make this immutable in the constructor.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszContainerSchema.java
```diff
@@ -358,7 +359,7 @@ public String getContainerName() {
 
   @Override
   public List<String> getFieldNames() {
-    return childrenNames;
+    return Collections.unmodifiableList(childrenNames);
```
ajsutton, line 362: Will take a bit of refactoring but should be able to make this immutable in the constructor as well.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/SszUnionSchemaImpl.java
```diff
@@ -132,7 +133,7 @@ public SszUnionT build() {
 
   @Override
   public List<SszSchema<?>> getChildrenSchemas() {
-    return childrenSchemas;
+    return Collections.unmodifiableList(childrenSchemas);
```
ajsutton, line 136: Can do this in the constructor.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/BeaconChainMethods.java
```diff
@@ -352,7 +353,7 @@ private static Eth2RpcMethod<PingMessage, PingMessage> createPing(
   }
 
   public Collection<RpcMethod<?, ?, ?>> all() {
-    return allMethods;
+    return Collections.unmodifiableCollection(allMethods);
```
ajsutton, line 356: This is already an immutable list.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/core/methods/VersionedEth2RpcMethod.java
```diff
@@ -94,7 +95,7 @@ void validateMethods(
 
   @Override
   public List<String> getIds() {
-    return protocolIds;
+    return Collections.unmodifiableList(protocolIds);
```
ajsutton, line 98: Can just make this immutable in the static factory method with `.toUnmodifiableList`.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/DiscoveryConfig.java
```diff
@@ -73,11 +73,11 @@ public int getAdvertisedUdpPort() {
   }
 
   public List<String> getStaticPeers() {
-    return staticPeers;
+    return Collections.unmodifiableList(staticPeers);
```
ajsutton, line 76: These can be made immutable in the builder.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/gossip/config/GossipScoringConfig.java
```diff
@@ -57,7 +58,7 @@ public GossipTopicScoringConfig getDefaultTopicScoringConfig() {
   }
 
   public Map<String, GossipTopicScoringConfig> getTopicScoringConfig() {
-    return topicScoringConfig;
+    return Collections.unmodifiableMap(topicScoringConfig);
```
ajsutton, line 61: Move to the builder.

ajsutton (commented):
Didn't get through all the changes here but figure it's worth stopping to discuss at this point. :)

There's a couple of common patterns I'm seeing:

1. Classes which are just a simple, ephemeral wrapper and are already being given a copy of the data. We could and possibly should create immutable lists back up at the source but it's pretty low value since these objects aren't shared between multiple places, they're just used to serialise a JSON response for a single request. Creating an extra unmodifiable wrapper isn't a big deal, but we definitely don't want to copy the data in this case.
2. Classes that are themselves immutable so rather than creating a new wrapper every time the getter is called, we should make the list unmodifiable in either the constructor or the builder if there is one.

The key issue with both of these is that doing the right thing and making the collections immutable back at the source won't satisfy the error prone check because it can't tell that the stored `List` is already unmodifiable.  Guava has things like `ImmutableList` class but I'm not keen to tie ourselves to a Guava specific types so heavily through all our code.

I'm not entirely sure what the right option here is - I've always been a bit sceptical of checks like this one because they do tend to encourage the easy option of just wrapping with unmodifiableList but that's often not the right option because we should be making things immutable much earlier in the process.

Potentially we could track if a collection is immutable by adding a custom `@ImmutableCollection` annotation and then have error prone rules to enforce only immutable stuff can be assigned to that but it feels like a fair bit of extra boiler plate code and relatively complex error prone rules and I'm not sure it's worth it given how rare it is that we have bugs caused by mutating collections we shouldn't.

jtraglia:
Hey, thanks for the review. All good points. I'm not sure what the best option is either. I would prefer not to introduce unnecessary complexity for a rare problem, like you said. It would have been different if this identified an actually bug. I was hoping that wrapping the return values would be sufficient but you are right that are better ways of handling these. I'll give this some more thought and maybe make something new out of it in smaller PRs.

PR #6157, opened 2022-09-05
Improve registration logging and fix NullPointer scenario

StefanBratanov:
## PR Description

- Specify that registrations were sent to the builder network rather than only to the Beacon Node.
- Avoid Null Pointer exception when builder is not defined in the proposer config and the registrations are enabled via CLI. In that case, the default gas limit would be used.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

tbenr:
> Specify that registrations were sent to the builder network rather than only to the Beacon Node.

good, this was needed.

PR #6167, opened 2022-09-07
VC connect to failover event stream if BN is not synced

### LINKED ISSUE #6142: Failover not working when primary execution layer is down or out of peers
### Description

Teku Validator Client does not failover to a secondary beacon node (CL) when the primary execution layer is down or out of peers and fell behind on blocks

### Steps to Reproduce (Bug)

Here is something I noticed when testing the failover option in Mainnet. I am running Teku CL and VC as separate processes. When I stop the Teku CL service, I see below logs on the Teku VC

`Validator   *** Switching to failover beacon node for event streaming
Validator   *** Successfully connected to beacon node event stream`

That is great. Works as expected. Then when I turn back the CL on, I see the below

`Validator   *** Primary beacon node is back and ready for event streaming. Will attempt connecting.
Validator   *** Successfully connected to beacon node event stream`

That is perfect too. 

But, my primary EL (Besu) was behaving bad with peers. It was not able to get enough peers at all. Often it goes down to 3-4 peers and sometimes zero as well. So during that time, I see the below log in the VC which is expected

`Execution Client is offline
java.util.concurrent.TimeoutException: null`

And it took 2-3 minutes (or sometimes quicker) for EL to find peers and come back online. During this time, I noticed that the VC never switches to the failover beacon node. Ideally, I would expect it to. I understand that pre-merge, EL going behind/offline for a bit is not that big of a deal. But with merge-ready Teku (22.8.1), I still think the above scenario should have triggered a failover

To test this further, I manually stopped the primary EL (Besu) to see what happens. CL clearly throws errors that it is not able to reach/talk to EL. But VC still did not failover to the secondary CL. This may be fine for pre-merge, but probably not fine post-merge. This is one of the main use case for having a secondary failover node I would think.

This is on Teku 22.8.1 running natively on Ubuntu 22.04.1

StefanBratanov:
## PR Description

- Connect to a failover event stream in case `/eth/v1/node/syncing` returns an error or shows that the node is not synced (which is the case if the EL goes down post merge)

The flow is as below:

- VC connects to the primary Beacon Node event stream
- Every 12 seconds (configurable) there is a task, which checks the syncing status of the primary Beacon Node:
- If the node is not synced or there is an error, the VC will try to connect to a failover Beacon Node event stream if failovers are configured
- In case the event stream has already failed over and the syncing status of the primary Beacon Node is back to normal, then a reconnection to the primary Beacon Node event stream will be attempted.

In case of event stream connection errors, the VC will try connecting to a failover as well. (this was the functionality before)

## Fixed Issue(s)
fixes #6142 

## Documentation

## Changelog


ajsutton (commented):
If a beacon node is syncing, we probably shouldn't be sending any requests to it. I think currently this is just covering the event stream but if we ask it to create a block or attestation it could wind up being based off of out of date data. I think if the BN is in syncing mode then those calls would return a 503 and we'd fail over, but a) it's a bit wasteful to make the request when we're already avoiding the BN for the event stream and b) not sure how this works out if the EL is completely offline.

ajsutton:
I think this is probably reasonable to merge once the AT is sorted as a first step, but probably worth thinking through the bigger picture for all API calls, not just the event stream.

tbenr (approved):
LGTM

PR #6169, opened 2022-09-07
Improve Error Messaging when parsing proposerConfigFile

### LINKED ISSUE #6162: Improve error message when proposer config file is invalid
### Description
When the proposer config file is invalid, Teku just reports:
```
Failed to load proposer config from: file:/var/lib/teku/proposerConfig.json
```

which makes it sound like the file can't be read when actually the issue is that the content isn't valid JSON.  We should at least indicate that the content was read but invalid and ideally provide some info about why it was invalid.

tbenr:
It is a bit noisy due to the `Eth1Address` refactoring but essentially:

- Moves `Eth1Address` under a new module `:ethereum:execution-types`
- Decouples `ETH1ADDRESS_TYPE` from `Eth1Address` and moves it in `EthereumTypes`
- Moves `EthereumTypes` to a new module `:ethereum:json-types`
- Moves `Eth1AddressDeserializer` under a new module `:ethereum:jackson-deserializers`

Error messages improvements:
- Improves errors while parsing `ProposerConfigFile` by providing more information (location of the error in the JSON source)
```
Failed to load proposer config from 'file:/Users/.../proposerConfigInvalid6.json' - Unexpected character ('w' (code 119)): was expecting double-quote to start field name at line: 4, column: 8

To display full help:
teku [COMMAND] --help
```
```
Failed to load proposer config from 'file:/Users/.../proposerConfigInvalid1.json' - Invalid Public Key at line: 3, column: 5

To display full help:
teku [COMMAND] --help
```

- Improves invalid default fee recipient CLI param error on startup
```
Teku failed to start: Invalid Default Fee Recipient: Invalid odd-length hex binary representation

To display full help:
teku [COMMAND] --help

```

fixes #6162

## Documentation

## Changelog


tbenr:
@ajsutton It seems like I'm hitting another dependency rule.
Was quite natural to me to add `Eth1Address` under bytes but it requires `:infrastructure:crypto`, so it won't be a `base` module anymore. I could move it into crypto itself but doesn't seem to me a good fit.

I still think that `Eth1Address` should go somewhere in `infrastructure`

rolfyone:
It could potentially be its own module in infrastructure that references bytes...
so we'd have `:infrastructure:bytes`, `:infrastructure:eth1`, `:infrastructure:crypto`...

ajsutton:
I don't think `Eth1Address` belongs in infrastructure at all - it's not a generic concept but something Ethereum specific so definitely should sit in the `ethereum` group somewhere. We've only just started cleaning up the ethereum layer so it's still pretty messy and most of the modules are too big really.

Naming-wise it would fit well in `ethereum:executionlayer` but that module actually looks pretty engine API and builder API related. Possibly it's in a new `ethereum:executiontypes` module.

Not yet sure if that helps your module dependency issues but I'm 100% sure it doesn't belong in infrastructure. :) Will take a look at the code and see if I have better suggestions then.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/CoreTypes.java
```diff
@@ -106,6 +107,15 @@ public class CoreTypes {
               HttpErrorResponse.Builder::message)
           .build();
 
+  public static final DeserializableTypeDefinition<Eth1Address> ETH1ADDRESS_TYPE =
```
ajsutton: Ah, I see why the move happened. This doesn't belong in CoreTypes (I'm not sure `HttpErrorResponse` should be in here either to be fair...). We have `EthereumTypes` which would be a good place for it, but currently that sits in the beacon rest API module and really should be moved to something shared between the beacon rest API and validator rest API - quite likely a `ethereum:json-types` module.

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
```diff
@@ -421,8 +422,13 @@ public Builder proposerDefaultFeeRecipient(final String proposerDefaultFeeRecipi
       if (proposerDefaultFeeRecipient == null) {
         this.proposerDefaultFeeRecipient = Optional.empty();
       } else {
-        this.proposerDefaultFeeRecipient =
-            Optional.of(Eth1Address.fromHexString(proposerDefaultFeeRecipient));
+        try {
+          this.proposerDefaultFeeRecipient =
+              Optional.of(Eth1Address.fromHexString(proposerDefaultFeeRecipient));
```
ajsutton, line 427: I'd be very tempted to just make this method take an `Eth1Address` instead of a `String` and make the CLI type an `Eth1Address` but we're getting well outside the scope of this PR now.

ajsutton (commented):
Looks good generally, just the module arrangement fun.

tbenr:
oky, yes, I got where you want to draw line. Agreed :)
Going to organize modules accordingly 🤞

ajsutton (approved):
LGTM.

PR #6176, opened 2022-09-09
Tighten viability check on finalized checkpoint in full progressive balances mode

ajsutton:
## PR Description
Slight tweak to tighten the restriction on nodes being viable for head to ensure the finalized checkpoint is equal to or greater than the store finalized epoch.

## Documentation

## Changelog


rolfyone (approved):
LGTM

tbenr:
Do we need a unit test covering this?

ajsutton:
> Do we need a unit test covering this?

This is picked up by the reference tests (just that it's not toggled on atm). Worth looking at how we could improve unit testing in this area though, but it's particularly tricky and ultimately it's the reference tests that give us confidence it will really work (where work is mostly defined as being the same as all other clients...)

PR #6180, opened 2022-09-10
Handle known getLogs errors

jonson:
## PR Description
As part of debugging my validator setup, I encountered an [issue along at least a few others](https://github.com/hyperledger/besu/issues/4293) where teku is continually querying besu for a 10k range of deposit logs, causing timeouts.  The fix in https://github.com/ConsenSys/teku/pull/6061 does not seem to have actually fixed this issue.  I believe the root cause is that teku is no longer creating a `FallbackAwareEth1Provider` [when PowchainConfiguration is not enabled](https://github.com/ConsenSys/teku/blob/feb46ef53a8c8519edde37fbe8bd67d803bf9d9a/services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainService.java#L84). This means no one is throwing `Eth1RequestException` anymore, which is what `DepositFetcher` is using for the auto-adjusting the range.  Meaning the range isn't being auto-adjusted.

The workaround is to change the config so the max range is 1000 (or less), in which besu responds in less than 8s, and all is good.

This is a hacky PR and there is likely a cleaner way to solve this, but I just wanted to expose the problem and a working solution.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,4 +24,5 @@ For information on changes in released versions of Teku, see the [releases page]
 ### Bug Fixes
 - Resolves an issue with public key validation.
 - Fix `/eth/v1/validator/register_validator` responding with a 400 status code and a misleading error message in case of exceptions
-- Update snakeyaml dependency to resolve cve-2022-25857 which could result in excessive memory usage when parsing YAML content.
+- Update snakeyaml dependency to resolve cve-2022-25857 which could result in excessive memory usage when parsing YAML content
+- Fixed an issue where the range requested for deposit logs was not reduced when using only `--ee-endpoint` leading to persistent timeouts with execution clients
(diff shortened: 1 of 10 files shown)


ajsutton:
oh that's a really good spot. Thank you. Your approach isn't bad but will have a play and see if there's a neater way (and work out why the build isn't running).

ajsutton:
Settled on using your basic approach of a static method in `Eth1RequestException` because we don't really need to wrap every exception with it, I just tweaked it to handle both cases so we don't need to duplicate the exception handler.

I've also moved the tests for this from being unit tests which assumed `FallbackAwareEth1Provider` was always in the chain to use an integration test that creates the same `Eth1Provider` pipeline as is used in production. That should provide better coverage in the future so we don't break this again.

rolfyone (approved):
LGTM

Issue #6053, opened 2022-08-10
Feature: ephemeral mode


benjaminion:
Calling this "ephemeral mode" to avoid a clash with "light client" which is something totally different. Open to other naming. Basically it's Teku-lite.

As a dev I'd sometimes like to quickly spin up a minimal client just to check the chain state or test some new code. It would be very short lived (hours, not days). This would look like:

1. Always checkpoint sync - possibly don't start if there is an existing db.
2. Stub out the execution client completely. (implemented but need docs)
3. Do not back fill historic blocks.
4. Do not load validators.
5. [Optional] Do not store finalised history.

I know how to do 1 and 2. Number 3 is the big one - back fill uses a lot of CPU and bandwidth. I understand that not backfilling might result in a loss of peers, but that probably doesn't matter for this application. Number 5 is probably not worth the effort to implement.

If `--ephemeral` is specified without `--initial-state` then refuse to start.

PR #6182, opened 2022-09-11
Make using execution engine stub scarier

ajsutton:
## PR Description
Change the prefix used to activate the execution engine stub to `unsafe-test-stub` and refuse to startup with a scary message if the old `stub` value is used.

Also makes the warning logged at startup a scarier sounding error and it repeats every time the transition configuration is called.

## Documentation

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -475,7 +475,7 @@ jobs:
           command: |
             mkdir -p .openapidoc/spec
 
-            build/install/teku/bin/teku --network=mainnet --ee-endpoint=stub --Xinterop-enabled=true --rest-api-enabled=true --rest-api-docs-enabled=true 2>&1 > teku_output.log &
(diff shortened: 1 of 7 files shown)


tbenr:
I'm thinking about disallowing  it for mainnet, but then we loose a testing tool. We could add a compile-time option to unable it for mainnet.

benjaminion:
I'm ok with making it very scary or even disabling it when (mainnet) validator keys are present. But this mode is useful for just running a low-value/test node. See also #6053 - it's a shame to be continually seeing with warnings in that case.

rolfyone (approved):
LGTM

ajsutton:
I'm not a fan of compile time feature toggles, they add a bunch of complexity to things generally and users will just toggle them on that way if they really want to shoot themselves in the foot. It's not like this was documented functionality anyway.

Even for a non-validating node it can be pretty dangerous to not validate payloads so while this setup is new I think we want to be really very loud about this. Once the merge goes through we'll likely remove the exchange_transitionConfiguration call entirely and that's likely a good time to revisit and potentially not repeat the warning so regularly.

Issue #6105, opened 2022-08-21
Don't make first descendant of head the new chain head


ajsutton:
### Description
Way back, we implemented an optimisation for long sync where if the imported block is the child of the current canonical chain head, it would be set as the new head without having to run fork choice.

While this still holds, during normal operation we run fork choice more often and there is more complexity around it. There is also a potential for race conditions as updating the chain head is an async operation - while they resolve themselves it leads to unnecessary reorgs.

We should investigate not automatically updating the chain head when a block is imported and leave it to be done when fork choice is next triggered to run.

We need to check if there are any async operations like retrieving a new justified/finalized state that should be preached to avoid delays.
Also need to ensure that the chain head gets updated often enough during a long sync and that performance isn't negatively affected.

zilm13:
Did I understand correctly that by optimization you mean not running explicitly [this method](https://github.com/ConsenSys/teku/blob/c3da1c9326ffe7aa6b7d7948db9b6217a7e0e473/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java#L177), so I need to:
- remove setting a head on block import, it should be set by `processHead` when it's run (separately)
- check when we run `processHead`, is there a potential for race but more likely reorgs, we should avoid it
- check what could be cached there
- check that during long sync head is not updated too rarely after changes
- ensure performance isn't negatively affected

ajsutton:
The key thing is that we want to remove the logic that automatically makes a new block the chain head if it descends from the current head block (https://github.com/ConsenSys/teku/blob/c3da1c9326ffe7aa6b7d7948db9b6217a7e0e473/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java#L475-L486).

> remove setting a head on block import, it should be set by processHead when it's run (separately)

Yes, I'm suggesting we think through what would happen if we didn't have `ForkChoice.updateForkChoiceForImportedBlock` which is updating the chain head as soon as we import the block. We would still be calling the `ForkChoice.processHead` method at the same times - that's the "normal" fork choice runs.

The first problem I can see is that we use the block import event to remove included operations from the pool if it's on the canonical chain but will now have to trigger that when we update the chain head, even if it's not strictly a reorg (ie if it just picks a new head further along the same fork).

> check when we run processHead, is there a potential for race but more likely reorgs, we should avoid it

[...]

PR #6183, opened 2022-09-12
Dont make first descendent of head the new chain head automatically

rolfyone:
Added as a feature toggle, `--Xfork-choice-first-descendent-as-head-enabled` is current behaviour, can use `--Xfork-choice-first-descendent-as-head-enabled=false` to observe change.

partially addresses #6105

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigBuilder.java
```diff
@@ -109,6 +109,7 @@ public class SpecConfigBuilder {
 
   // Bellatrix
   private Optional<BellatrixBuilder> bellatrixBuilder = Optional.empty();
+  private boolean firstDescendentAsChainHeadEnabled = true;
```
ajsutton: hmm, this probably shouldn't wind up in SpecConfig if we can avoid it. The progressive balances feature toggle wound up in there out of necessity but we shouldn't need this in the milestone specific code - just `ForkChoice` so could just pass it there I think?

rolfyone: it did seem messy, just seemed how other eth2 things got passed, will refactor

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
```diff
@@ -169,6 +169,17 @@ public class Eth2NetworkOptions {
       hidden = true)
   private Boolean forkChoiceBeforeProposingEnabled = null;
 
+  @Option(
+      names = {"--Xfork-choice-first-descendent-as-head-enabled"},
```
ajsutton: I'd be tempted to call this `--Xfork-choice-update-head-on-block-import-enabled` since we're totally skipping updating the chain head when importing a block. Sorry should have picked that up sooner to save a bunch of renaming...

ajsutton (approved):
LGTM.

PR #6184, opened 2022-09-12
Sentry beacon nodes implementation

lucassaldanha:
## PR Description

- Implemented SentryValidatorApiChannel and  SentryBeaconNodeApi
- Updated ValidatorClientService with logic to start sentry mode

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
@@ -60,9 +60,11 @@ import tech.pegasys.teku.validator.client.restapi.ValidatorRestApiConfig;
 import tech.pegasys.teku.validator.eventadapter.InProcessBeaconNodeApi;
 import tech.pegasys.teku.validator.remote.RemoteBeaconNodeApi;
 import tech.pegasys.teku.validator.remote.sentry.SentryBeaconNodeApi;
+import tech.pegasys.teku.validator.remote.sentry.SentryNodesConfig;
 import tech.pegasys.teku.validator.remote.sentry.SentryNodesConfigLoader;
 
 public class ValidatorClientService extends Service {
+
   private static final Logger LOG = LogManager.getLogger();
   private final EventChannels eventChannels;
   private final ValidatorLoader validatorLoader;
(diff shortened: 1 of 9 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -219,6 +213,57 @@ public static ValidatorClientService create(
...
+                      SentryBeaconNodeApi.create(
+                          services,
+                          asyncRunner,
+                          sentryNodesConfig,
+                          config.getSpec(),
+                          generateEarlyAttestations,
+                          preferSszBlockEncoding,
+                          failoversSendSubnetSubscriptions,
+                          beaconNodeEventStreamSyncingStatusQueryPeriod))
+              .orElseThrow(RuntimeException::new);
```
ajsutton: We should probably throw an exception with a more meaningful error message here.  Though it may be better to just write this starting with a `validatorConfig.getSentryNodeConfigurationFile().get()` and skip the map calls given we've already done the isEmpty() check above?

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteBeaconNodeEndpoints.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import okhttp3.HttpUrl;
+
+public class RemoteBeaconNodeEndpoints {
+
+  private final List<HttpUrl> endpoints = new ArrayList<>();
+
+  public RemoteBeaconNodeEndpoints(final List<URI> endpoints) {
+    this.endpoints.addAll(stripAuthentication(convertToOkHttpUrls(endpoints)));
```
ajsutton: nit: Probably simpler to use:
```suggestion
    this.endpoints = new ArrayList<>(stripAuthentication(convertToOkHttpUrls(endpoints)));
```

Then the list will be correctly sized from the start. Not that performance matters here but it's also slightly less code. :)

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/sentry/SentryBeaconNodeApi.java
```diff
@@ -0,0 +1,256 @@
...
+    final List<HttpUrl> allBeaconNodeEndpoints =
+        new ArrayList<>(
+            convertToOkHttpUrls(
+                beaconNodesSentryConfig.getDutiesProviderNodeConfig().getEndpointsAsURIs()));
+    beaconNodesSentryConfig
+        .getBlockHandlerNodeConfig()
+        .ifPresent(c -> allBeaconNodeEndpoints.addAll(convertToOkHttpUrls(c.getEndpointsAsURIs())));
+    beaconNodesSentryConfig
+        .getAttestationPublisherConfig()
+        .ifPresent(c -> allBeaconNodeEndpoints.addAll(convertToOkHttpUrls(c.getEndpointsAsURIs())));
```
ajsutton: I'd be tempted to make this a method in `BeaconNodesSentryConfig` so that it's more likely to be updated if there are new types of nodes added to the config in the future.

ajsutton (approved):
LGTM.  The thing that's making me a bit nervous though is that we're winding up with two completely different paths for creating the `ValidatorApiChannel` implementation we ultimately use. There's details like wrapping it in `MetricRecordingValidatorApiChannel` that's common and things like the `BeaconChainEventAdapter`. While it's right for now, I can easily see how we might change something important in the non-sentry case and forget to update it in the sentry node case or visa versa.

I wonder if we need to extract some common factory methods out to share the code better. Things like taking a list of endpoints and creating the `ValidatorApiChannel` implementation like we do for duties, block handlers and attestation publishers separately.  Plus probably for creating the `BeaconChainEventAdapater` too.  Probably should be a follow up PR though.

Issue #6166, opened 2022-09-07
Automatically sort blocks in transition blocks subcommand


ajsutton:
### Description
It would be really helpful to be able to apply a batch of blocks with just `teku transition blocks -I state.ssz blocks/*.ssz` but that doesn't currently work because the list of blocks have to be sorted in slot order.

The transition subcommand could do this sorting automatically.

PR #6186, opened 2022-09-12
Sort blocks by slot number

mehdi-aouadi:
## PR Description

Sort the supplied blocks by slot number when using the transition blocks subcommand

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6166

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/subcommand/TransitionCommand.java
```diff
@@ -180,6 +185,19 @@ private SignedBeaconBlock readBlock(final Spec spec, final String path) throws I
     }
   }
 
+  private void sortBlocksBySlotNumber(final List<SignedBeaconBlock> blocks) {
+    blocks.sort(
```
zilm13: I guess we could do it with something like
`blocks.sort(Comparator.comparing(SignedBeaconBlock::getSlot));`

zilm13: Also as we are modifying input, I'd leave sorting in the place to be clear visible instead of separate method

mehdi-aouadi: My bad!

zilm13 (approved):
LGTM

Issue #6164, opened 2022-09-06
Support reading blocks as hex data in transition subcommand


ajsutton:
### Description
To make it easier to replay invalid blocks, add the ability to parse blocks that are stored as hex strings in a file rather than binary ssz to the `teku transition blocks` subcommand.

Specifically we want to be able to create the block file by just copying from our invalid gossip log and pasting it into a file. Can probably autodetect this if the file fails to parse as binary SSZ, try to load it via `Bytes.fromHexString` and then parse that as SSZ.

mehdi-aouadi:
I implemented a hex deserialisation fallback when ssz fails.
The good thing is that the blocks can have mixed format but that makes the code less elegant.
I was thinking of adding a command option to choose the blocks format (which default to ssz if not present) and keep things separate.

PR #6187, opened 2022-09-12
Deserialize block as HEX when SSZ fails

mehdi-aouadi:
## PR Description

Support reading blocks as hex data in transition subcommand:
The blocks are deserialized as SSZ first. If that fails, the blocks are deserialized as HEX.
Blocks can have mixed data formats (SSZ and HEX).

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6164
## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/subcommand/TransitionCommand.java
```diff
@@ -181,11 +182,22 @@ private BeaconState readState(final Spec spec, final Bytes inData) {
...
+  private SignedBeaconBlock deserializeSignedBeaconBlockFromHex(
+      final Spec spec, final String path, final byte[] hexBlockData) {
+    try {
+      final Bytes blockData = Bytes.wrap(Hex.decode(hexBlockData));
       return spec.deserializeSignedBeaconBlock(blockData);
-    } catch (final IllegalArgumentException e) {
-      throw new SSZException("Failed to parse SSZ (" + path + "): " + e.getMessage(), e);
+    } catch (final RuntimeException e) {
+      throw new RuntimeException(
+          String.format("Failed to parse SSZ/HEX (%s): %s", path, e.getMessage()), e);
```
ajsutton: I suspect the exception message we're getting here isn't the most useful one. ie if the SSZ was invalid (from the wrong fork or something) this would just give a message about an invalid hex character which is likely to be confusing.  We probably want to track the original exception from parsing the SSZ and provide both error messages here.

ajsutton (approved):
LGTM. Would just be good to improve the error reporting.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/TransitionCommand.java
mehdi-aouadi: I propagated down the SSZ parsing exception and added its message to the final exception

PR #6188, opened 2022-09-12
Don't send VC requests to not synced BN

### LINKED ISSUE #6174: Don't send VC requests to a not synced BN
As an extension to https://github.com/ConsenSys/teku/pull/6167 and https://github.com/ConsenSys/teku/issues/6142 we can improve the `FailoverValidatorApiHandler` to not send requests to a BN if it is not synced. 

Current solution is still viable since the BN will return `503 Beacon node is currently syncing and not serving request on that endpoint` for most calls and will failover after the exception, but there will be loads of unnecessary calls, which will slow down the validator duties completion.

StefanBratanov:
## PR Description

- Introduce `BeaconNodeReadinessManager` which tracks the ready status (on aggregation due) of the multiple configured Beacon Nodes and publishes  to `RemoteBeaconNodeSyncingChannel` subscribers
- Requests to the primary Beacon Node check the readiness before making any request (using the readiness manager) and failovers immediately if not ready
- The `EventSourceBeaconChainEventAdapter` is subscribed to sync status changes and failovers if needed.

HackMd document explaining the functionality:
https://hackmd.io/gvGgZIyXSPuRPu1wgCxGBw?view

## Fixed Issue(s)
fixes #6174 

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorClientOptions.java
```diff
@@ -49,16 +49,29 @@ public class ValidatorClientOptions {
       ValidatorConfig.DEFAULT_FAILOVERS_SEND_SUBNET_SUBSCRIPTIONS_ENABLED;
 
   @CommandLine.Option(
-      names = {"--Xbeacon-node-event-stream-syncing-status-query-period"},
+      names = {"--Xbeacon-nodes-syncing-status-query-period"},
```
rolfyone: do we need millis here? I would have thought seconds should be plenty..

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorClientOptions.java
```diff
@@ -49,16 +49,29 @@ public class ValidatorClientOptions {
...
       paramLabel = "<INTEGER>",
       description =
-          "How often (in milliseconds) will the syncing status of the Beacon Node used for event streaming be queried. If the node is not synced or the query fails, the Validator Client will try to connect to an event stream of one of the failover Beacon Nodes if such are configured. "
-              + "If the primary Beacon Node has synced successfully after a failover, a reconnection to the primary Beacon Node event stream will be attempted.",
+          "How often (in milliseconds) will the syncing status of the configured Beacon Nodes be queried. "
+              + "The result of the query would be used to determine if a Validator Client should send requests to a Beacon Node or failover immediately. "
+              + "This option is only applicable if there are multiple Beacon Nodes configured.",
```
rolfyone: `beacon node` and `validator client` should both be lower case through these.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorClientOptions.java
```diff
@@ -49,16 +49,29 @@ public class ValidatorClientOptions {
...
+      names = {"--Xbeacon-node-event-stream-readiness-check-period"},
+      paramLabel = "<INTEGER>",
+      description =
+          "How often (in milliseconds) will the readiness of the Beacon Node used for event streaming be checked. If the node is not ready, the Validator Client will try to connect to an event stream of one of the failover Beacon Nodes if such are configured. "
+              + "If the primary Beacon Node becomes ready again after a failover, a reconnection to the primary Beacon Node event stream will be attempted.",
+      hidden = true,
+      showDefaultValue = CommandLine.Help.Visibility.ALWAYS,
+      arity = "1")
+  private long beaconNodeEventStreamReadinessCheckPeriod =
+      ValidatorConfig.DEFAULT_BEACON_NODE_EVENT_STREAM_READINESS_CHECK_PERIOD.toMillis();
```
rolfyone: im not sure that we need separate frequency for event stream checking, i would have thought we could just fail over to the new BN including event stream... having a separate event stream to the BN you're connected to may lead to a level of chaos..

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
```diff
@@ -199,54 +200,24 @@ private boolean currentEventStreamHasSameEndpoint(final RemoteValidatorApiChanne
...
-                        }),
-            beaconNodeEventStreamSyncingStatusQueryPeriod,
-            beaconNodeEventStreamSyncingStatusQueryPeriod,
+            () -> {
+              if (beaconNodeReadinessManager.isReady(primaryBeaconNodeApi)) {
+                if (maybeFailoverEventSource.isPresent()) {
+                  switchToPrimaryEventStream();
+                }
+              } else {
+                switchToFailoverEventStreamIfNeeded();
```
rolfyone: looks like these don't call close on the old stream, which we should do i think.. they'll probably keep trying to reconnect? also `closeCurrentEventStream` is a little confusing, it may be close to time to refactor, as it implies that we've got multiple streams that may be open that we need to close... we don't want multiple streams open...

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorClientOptions.java
StefanBratanov: > im not sure that we need separate frequency for event stream checking, i would have thought we could just fail over to the new BN including event stream... having a separate event stream to the BN you're connected to may lead to a level of chaos..

Ideally yes, but problem is a beacon node may be not synced (EL offline for example) and in that case the event stream won't throw an exception and a failover won't be triggered unless the syncing status is checked frequently

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
StefanBratanov: > looks like these don't call close on the old stream, which we should do i think.. they'll probably keep trying to reconnect? also `closeCurrentEventStream` is a little confusing, it may be close to time to refactor, as it implies that we've got multiple streams that may be open that we need to close... we don't want multiple streams open...

They both close the current stream in the `switchToPrimaryEventStream` and `switchToFailoverEventStream` methods. I will look into refactoring.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorClientOptions.java
StefanBratanov: Actually agree, I did it for the tests, but in the acceptance tests we can just use one second since they usually take bit longer anyways

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Override
+  protected SafeFuture<?> doStart() {
+    syncingQueryTask =
+        asyncRunner.runWithFixedDelay(
+            () -> {
+              final SafeFuture<Void> primaryReadinessCheck = performPrimaryReadinessCheck();
+              final Stream<SafeFuture<?>> failoverReadinessChecks =
+                  failoverBeaconNodeApis.stream().map(this::performFailoverReadinessCheck);
```
rolfyone: this could be a terrible idea... just prefacing with that...
really since we're testing all nodes all the time, we could just have a single list, where item 0 is primary and everything else is failover...
then theres no real difference between primary / non primary, although i guess theres no point flapping between failovers, you'd potentially still be looking if element 0 is available where you're checking primary... anyway just a thought, then the stream here checks all nodes...

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManagerTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.async.StubAsyncRunner;
+import tech.pegasys.teku.infrastructure.time.StubTimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.validator.api.required.SyncingStatus;
+
+class BeaconNodeReadinessManagerTest {
```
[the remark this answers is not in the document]
rolfyone: could make public class to make codeql happy...

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
```diff
@@ -205,11 +205,9 @@ public TekuValidatorNode.Config withBeaconNodes(final TekuNode... beaconNodes) {
...
     }
 
-    public TekuValidatorNode.Config withBeaconNodeEventStreamSyncingStatusQueryPeriod(
-        final Duration syncingStatusQueryPeriod) {
-      configMap.put(
-          "Xbeacon-node-event-stream-syncing-status-query-period",
-          syncingStatusQueryPeriod.toMillis());
+    public TekuValidatorNode.Config withBeaconNodesSyncingQueryPeriod(
+        final Duration syncingQueryPeriod) {
+      configMap.put("Xbeacon-nodes-syncing-query-period", syncingQueryPeriod.toSeconds());
```
rolfyone: Theres probably very sub-optimal times to be deciding to change back from failover back to primary, so we likely want to be careful...

If we're about to calculate duties or perform a duty, failing back to primary may result in inactivity. I'm not sure how we reconcile that... potentially we want to wait until duties are complete for a slot rather than doing it every N period of time...  I could forsee 12 / 6 / 4 / 3 all potentialy being very disruptive and leading to hard to track down inactivities.
block production is probably worse in bellatrix, now that we might be getting a blinded block and not having the EL payload, failover during that period would result in loss of block production which would be really not great...
I'm not sure what the answer would be other than checking it around half way through a slot - potentially would miss aggregation, but shouldnt then miss any attestation or block production duties (i think they start in the last couple of seconds of the slot so we'd want to be stable by then or queued behind the changeover)

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
StefanBratanov: Yeah, I get your point. I am mainly having the primary separate to make use of the `RemoteBeaconNodeSyncingChannel`  primary node not in sync/ back in sync callbacks.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -0,0 +1,158 @@
...
+  private final List<RemoteValidatorApiChannel> failoverBeaconNodeApis;
+  private final AsyncRunner asyncRunner;
+  private final RemoteBeaconNodeSyncingChannel remoteBeaconNodeSyncingChannel;
+  private final Duration beaconNodeSyncingQueryPeriod;
+
+  private volatile Cancellable syncingQueryTask;
+
+  public BeaconNodeReadinessManager(
+      final RemoteValidatorApiChannel primaryBeaconNodeApi,
+      final List<RemoteValidatorApiChannel> failoverBeaconNodeApis,
```
mehdi-aouadi, line 46: IMO we could simplify this by keeping a single list with the primary node at the top as suggested by @rolfyone.
I don't see why we should treat primary/failover nodes separately. 
In my understanding, we simply have a list of nodes, when the one we're using is unavailable (syncing or whatever) we iterate through the list until we get one which could do the job. 
If the primary node must be privileged, we could check if we could connect back to it at each iteration first and then check the other failover nodes...

mehdi-aouadi (approved):
LGTM as a first iteration. Good job!

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
StefanBratanov: Decided to run the syncing check on every aggregation (2/3 into the slot) so it shouldn't be disruptive

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
StefanBratanov: Mainly keeping it for the callback functionality which is used by the event stream so it is good to differentiate between the primary node and the failover nodes.

PR #6191, opened 2022-09-13
Pass through custom descriptions in SafeFutureAssert

ajsutton:
## PR Description
When test code specifies a custom description for an assertion, `SafeFutureAssert` should pass it through to any assertions it then uses so the custom description actually appears in error messages.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6193, opened 2022-09-13
Cleaning up some params on ValidatorClientService

lucassaldanha:
Removed multiple boolean parameters in favour of the config object that contains the information to minimize the number of parameters.


ajsutton (approved):
LGTM.

PR #6194, opened 2022-09-13
Support async storage updates

ajsutton:
## PR Description
Adds a feature toggle (default off) to allow updating the in-memory Store without first storing the update to disk.

Currently, newly finalized data may be temporarily unavailable as it may be removed from the in-memory store before it is persisted to disk. Bufferring will be added in a follow up PR.

## Fixed Issue(s)
#1934 

## Documentation

## Changelog


storage/src/test/java/tech/pegasys/teku/storage/server/RetryingStorageUpdateChannelTest.java
```diff
@@ -0,0 +1,159 @@
...
+        .thenAnswer(
+            invocation -> {
+              // Increase time a bit on each call
+              timeProvider.advanceTimeBySeconds(10);
+              return SafeFuture.failedFuture(new RuntimeException("Failed"));
+            });
+
+    assertThatThrownBy(() -> retryingChannel.onStorageUpdate(event))
+        .isInstanceOf(FatalServiceFailureException.class);
+    verify(delegate, times(7)).onStorageUpdate(event);
```
lucassaldanha: Where does 7 come from? Is it a predefined limit of retries?

lucassaldanha (approved):
LGTM. Just a minor q re readability.

storage/src/test/java/tech/pegasys/teku/storage/server/RetryingStorageUpdateChannelTest.java
ajsutton: Oh yeah I meant to switch actually calculating that.  There's a fixed amount of time and the test advances time a certain amount on each retry.

rolfyone (approved):
LGTM

PR #6189, opened 2022-09-12
Improve shuffling performance

tbenr:
Avoids multiple `createSha256` and `Bytes.wrap` by introducing a `Sha256` class designed to be reused.

Before:
```
# Run complete. Total time: 00:01:26

Benchmark                            (indexCount)  Mode  Cnt  Score   Error  Units
ShuffleBenchmark.shuffledIndexBench         16384    ss   15  1.088 ± 0.017   s/op
ShuffleBenchmark.shuffledIndexBench         32768    ss   15  2.283 ± 0.185   s/op
ShuffleBenchmark.shuffledListBench          16384    ss   15  0.012 ± 0.004   s/op
ShuffleBenchmark.shuffledListBench          32768    ss   15  0.026 ± 0.003   s/op
```

After:
```
# Run complete. Total time: 00:01:17

Benchmark                            (indexCount)  Mode  Cnt  Score    Error  Units
ShuffleBenchmark.shuffledIndexBench         16384    ss   15  0.992 ±  0.020   s/op
ShuffleBenchmark.shuffledIndexBench         32768    ss   15  1.958 ±  0.021   s/op
ShuffleBenchmark.shuffledListBench          16384    ss   15  0.008 ±  0.001   s/op
ShuffleBenchmark.shuffledListBench          32768    ss   15  0.014 ±  0.001   s/op
```

Memory allocations comparison on shuffle benchmark:

image

image

related to: #6016

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #6195, opened 2022-09-13
Use Sha256 where applicable

tbenr:
Applies the sha56 reusable digest pattern in `getNextSyncCommitteeIndices` and `isValidMerkleBranch`

atop: #6189

diff commit https://github.com/ConsenSys/teku/pull/6195/commits/df741bda12120cbc96361219996dc1e8869d6af6

## Documentation

## Changelog


eth-benchmark-tests/src/jmh/java/tech/pegasys/teku/benchmarks/ShuffleBenchmark.java
```diff
@@ -60,4 +61,13 @@ public void shuffledListBench(Blackhole bh) {
...
   }
+
+  public static void main(String[] args) {
+    ShuffleBenchmark benchmark = new ShuffleBenchmark();
+
+    benchmark.indexCount = 32768;
+
+    new CustomRunner(2, 2).withBench(benchmark::shuffledIndexBench).run();
+    new CustomRunner(2, 2).withBench(benchmark::shuffledListBench).run();
+  }
```
ajsutton: We probably don't need to add this.  The IntelliJ jmh plugin should be able to run the benchmarks directly.

ajsutton (approved):
LGTM.

eth-benchmark-tests/src/jmh/java/tech/pegasys/teku/benchmarks/ShuffleBenchmark.java
tbenr: I used that to run the profiler. Because it seems not possible to run benchmarks using jmh plugin and activate the intellij profiler at the same time.

PR #6198, opened 2022-09-14
Preserve ordering of updates and queries when async IO is enabled

ajsutton:
## PR Description
Moves storage updates and queries to use the same event channel when async IO is enabled. This ensures that any query sent to the channel after an update only begins executing after the update completes. To avoid updates being delayed by queries, the queries are actually executed on a separate async runner. This means that queries may be delayed by updates but not the other way around. It avoids the need to try and buffer content to deal with updates being asynchronous and finalized data being unavailable until the update completes.

Separate channels are still used when async IO is disabled so there's no impact to current behaviour by default.

## Fixed Issue(s)
#6194 

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/CombinedStorageChannelSplitter.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/CombinedStorageChannelSplitter.java
@@ -41,7 +41,7 @@ import tech.pegasys.teku.storage.api.WeakSubjectivityUpdate;
 /**
  * Splits calls to {@link CombinedStorageChannelSplitter} into the separate {@link
  * StorageUpdateChannel} and {@link StorageQueryChannel} components with updates being handled
- * syncrhonously and queries being run asyncrhonously.
+ * synchronously and queries being run asynchronously.
  *
  * <p>This guarantees that queries are only ever processed after the updates that were sent before
  * them but without allowing queries to delay updates.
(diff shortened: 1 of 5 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/CombinedStorageChannelSplitter.java
```diff
@@ -0,0 +1,174 @@
...
+import tech.pegasys.teku.storage.api.StorageUpdate;
+import tech.pegasys.teku.storage.api.StorageUpdateChannel;
+import tech.pegasys.teku.storage.api.UpdateResult;
+import tech.pegasys.teku.storage.api.WeakSubjectivityState;
+import tech.pegasys.teku.storage.api.WeakSubjectivityUpdate;
+
+/**
+ * Splits calls to {@link CombinedStorageChannelSplitter} into the separate {@link
+ * StorageUpdateChannel} and {@link StorageQueryChannel} components with updates being handled
+ * syncrhonously and queries being run asyncrhonously.
```
lucassaldanha:
```suggestion
 * synchronously and queries being run asynchronously.
```

lucassaldanha (approved):
LGTM. It might be good to have another pair of eyes going through it though.

storage/src/main/java/tech/pegasys/teku/storage/server/CombinedStorageChannelSplitter.java
ajsutton: Darn my fingers - I typed that wrong and fixed it in about 3 other places too...

tbenr (approved):
LGTM too

Issue #6098, opened 2022-08-19
Remove `UnusedValidatorClientOptions`


ajsutton:
### Description
Quite a while back we incorrectly had a number of beacon node specific options appear in the validator client subcommand as well.  That's since been fixed and we avoided breaking backwards compatibility by making them hidden and moving them to `UnusedValidatorClientOptions`.  We should remove that class now so they aren't seen as valid subcommand options anymore.

PR #6201, opened 2022-09-14
Remove UnusedValidatorClientOptions

mehdi-aouadi:
## PR Description

Remove `UnusedValidatorClientOptions` which is no longer required to ensure backwards compatibility.

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6098

## Documentation

## Changelog


ajsutton (approved):
LGTM but we need to add this to the change log. It will go in the "Breaking Changes" section and should list each of the removed options (except the ones starting with `--X`) with a note that they have been ignored for some time but are now completely removed.

PR #6202, opened 2022-09-15
StorageUpdate should not be empty when optimisticTransitionBlockRootSet is true

ajsutton:
## PR Description
As noted in #6190 `StorageUpdate.isEmpty` should return false if `optimisticTransitionBlockRootSet` is true.  In production this never happens without also having other values that are checked set, but better to be comprehensive.

Adds an ugly but effective reflection based test to ensure new parameters aren't ignored in the future.

## Documentation

## Changelog


storage/api/src/main/java/tech/pegasys/teku/storage/api/StorageUpdate.java
```diff
@@ -59,6 +61,9 @@ public StorageUpdate(
     this.stateRoots = stateRoots;
     this.optimisticTransitionBlockRootSet = optimisticTransitionBlockRootSet;
     this.optimisticTransitionBlockRoot = optimisticTransitionBlockRoot;
+    checkArgument(
```
zilm13, line 64: I saw it once in debugger but maybe I'm wrong. Can we have a case when we clear one set before with empty?

tbenr (approved):
LGTM

PR #6203, opened 2022-09-15
Prevent sentry node config being used with beacon-node-api-endpoint

lucassaldanha:
When using sentry nodes, the `duties_provider` node should be used as the "main" beacon node API endpoint. One example of this is when the VC starts up and downloads the spec from the BN.

This change creates the following rules:
1. If `--Xsentry-config-file` AND `--beacon-node-api-endpoint` are not set, we use the default beacon node api endpoint (defined on `ValidatorConfig.DEFAULT_BEACON_NODE_API_ENDPOINTS`).
2. If `--Xsentry-config-file` AND `--beacon-node-api-endpoint` are set, the node fails starting and gives a useful error message.
3. When only `--Xsentry-config-file` is set, the node will use the endpoint for the `duties_provider` (as defined in the sentry config file) as the beacon node api endpoint.


courtneyeh (approved):
LGTM

PR #6214, opened 2022-09-16
Guarantee that epoch priming happens asynchronously

### LINKED ISSUE #6213: make sure epochCacheTimer  runs asynchronously
Specifically primeCacheForEpoch should be all running asynchronously. it probably needs to run via an asyncRunner.

Seen because slow tick events highlighted that `precompute_epoch_complete` is sometimes taking a second or more.

ajsutton:
## PR Description
Use an async runner to ensure that epoch priming happens asynchronously even if the required state happens to already be in the cache. Otherwise we may block the tick processing thread.

fixes #6213 

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

rolfyone (approved):
LGTM

PR #6217, opened 2022-09-16
Reduce size of in memory block cache

ajsutton:
## PR Description
Post merge blocks are about 3 times bigger, so reduce the size of the cache. In normal finalization we would typically store up to 96 blocks so reduce that to a single epoch worth at a time to reduce GC pressure and ensure states can be held in memory.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6218, opened 2022-09-16
Enable async IO by default

### LINKED ISSUE #1934: Make disk calls non-blocking

ajsutton:
## PR Description
Enables async IO by default.  There are some tests that will need updating to accommodate this change.

## Fixed Issue(s)
fixes #1934 

## Documentation

## Changelog


StefanBratanov:
Replaced by https://github.com/ConsenSys/teku/pull/6222

PR #6220, opened 2022-09-16
Run validator registrations when two slots in the epoch instead of the beginning

StefanBratanov:
## PR Description
Since a lot of operations are done in the beginning of the epoch, the validator registrations can be moved later in the epoch.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #6222, opened 2022-09-16
Enable async IO by default

### LINKED ISSUE #1934: Make disk calls non-blocking

StefanBratanov:
## PR Description
Enable `--Xdata-storage-async-enabled` by default

## Fixed Issue(s)
fixes https://github.com/ConsenSys/teku/issues/1934

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ For information on changes in released versions of Teku, see the [releases page]
 - The `--initial-state` and `--eth1-deposit-contract-address` options has been removed from the `validator-client` subcommand. They have been ignored for some time but are now completely removed.
 
 ### Additions and Improvements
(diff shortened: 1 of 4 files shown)


tbenr (approved):
LGTM

ajsutton (approved):
LGTM. Thanks Stefan.

PR #6223, opened 2022-09-17
Update overrides for multiply-licensed dependencies

benjaminion:
Stops gradle complaining about multiply-licensed dependencies by specifying which licence we wish to use in each case.

## Documentation

## Changelog


ajsutton (approved):
LGTM. Thank you that's been annoying me but I never see it when I have time to fix it.

ajsutton:
I think you may need to log out of circleCI and back in again to make it happy though.

benjaminion:
> I think you may need to log out of circleCI and back in again to make it happy though.

I spent 45 minutes logging out and in again, revoking and re-adding permissions, and updating the PR to kick CircleCI, but I seem to have only made things worse 😞  Will come back to this next week sometime!

PR #6237, opened 2022-09-21
Change slot in the epoch for preparing, registering validators and expiry

StefanBratanov:
## PR Description

- Move preparing and registering validators to third slot of each epoch
- Move expiry to middle of the epoch and extend the expiration epoch
- Improve logging in few places

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ProposersDataManager.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ProposersDataManager.java
@@ -45,7 +45,7 @@ import tech.pegasys.teku.storage.client.RecentChainData;
 public class ProposersDataManager implements SlotEventsChannel {
   private static final Logger LOG = LogManager.getLogger();
   private static final long PROPOSER_PREPARATION_EXPIRATION_EPOCHS = 3;
-  private static final long VALIDATOR_REGISTRATION_EXPIRATION_EPOCHS = 3;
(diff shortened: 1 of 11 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ProposersDataManager.java
```diff
@@ -44,8 +44,8 @@
 
 public class ProposersDataManager implements SlotEventsChannel {
   private static final Logger LOG = LogManager.getLogger();
-  private static final long PROPOSER_PREPARATION_EXPIRATION_EPOCHS = 2;
-  private static final long VALIDATOR_REGISTRATION_EXPIRATION_EPOCHS = 2;
+  private static final long PROPOSER_PREPARATION_EXPIRATION_EPOCHS = 3;
+  private static final long VALIDATOR_REGISTRATION_EXPIRATION_EPOCHS = 3;
```
tbenr: The only thing that makes me hesitant is that, differently from `PROPOSER_PREPARATION_EXPIRATION_EPOCHS`, `VALIDATOR_REGISTRATION_EXPIRATION_EPOCHS` has an actual usecase attached to it, which is the live disabling of builder registration. Setting this to 3 means that we are adding an additional epoch of lag.
In this case it could be good to have it configurable so customers could change it based on their needs

tbenr (approved):
LGTM with an idea for a new issue

PR #6244, opened 2022-09-23
Avoid unnecessary removeFinalizedOptimisticTransitionPayload transaction

### LINKED ISSUE #6232: Async IO not always run Async
Every time we import a block and we receive a VALID from the EL in `onExecutionPayloadResult`, we always try to remove FinalizedOptimisticTransitionPayload from the Storage

so in `MergeTransitionBlockValidator#verifyFinalizedTransitionBlock` we call:

image

this will set `clearFinalizedOptimisticTransitionPayload` to `true` in transaction.

this will then cause `optimisticTransitionBlockRootSet` to be `true` when creating `StorageUpdate`: 
image

end we end up being always synchronous due to:

image

tbenr:
Avoids an unnecessary Store and Storage transaction when importing new blocks: if store has an empty `finalizedOptimisticTransitionPayload` we don't need to clear it and we can assume that is not set in Storage too.

It also early filer empty `StorageUpdate` before being sent over the `StorageUpdateChannel`

fixes: #6232

related to #6016

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #6245, opened 2022-09-23
Add dedicated lock to votes fields in `Store`

tbenr:
`Store` is updated via `StoreTransaction` and `StoreVoteUpdater`. There is no intersection between the two: they update different fields in the store and they don't relay on the others data to compute the update.

Before this PR, Store locking was done via only one lock, which means both updates systems where locking the entire store.

In particular, `StoreVoteUpdater` updates votes via `applyForkChoiceScoreChanges` which is computational intensive. It means that every time we were processing head and the entire Store were locked including timing getters (`getTimeMillis` and `getGenesisTime`) that are massively used (ie: `onTick`, processing data coming from gossip)

This PR introduces a lock dedicated to votes fields, so we will have two different locks representing the two way of updating the Store. They do not lock eachother out and more importantly the heavy `StoreVoteUpdater` will no longer lock timing getters. Timing getters are still locked by `StoreTransaction` which locks in `applyToStore` (which can be still heavy) and in `putBlock` used to cache blocks.

related to #6016

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
@@ -711,7 +711,7 @@ class Store implements UpdatableStore {
   }
 
   private void putBlock(final SignedBeaconBlock block) {
-    final Lock writeLock = votesLock.writeLock();
+    final Lock writeLock = lock.writeLock();
     writeLock.lock();
     try {
       if (containsBlock(block.getRoot())) {
--- a/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
@@ -76,6 +76,8 @@ class Store implements UpdatableStore {
   private final Lock readVotesLock = votesLock.readLock();
   private final ReadWriteLock lock = new ReentrantReadWriteLock();
   private final Lock readLock = lock.readLock();
+  private final ReadWriteLock timingLock = new ReentrantReadWriteLock();
+  private final Lock readTimingLock = timingLock.readLock();
 
   private final MetricsSystem metricsSystem;
   private Optional<SettableGauge> blockCountGauge = Optional.empty();
@@ -275,6 +277,8 @@ class Store implements UpdatableStore {
    */
   @Override
   public void startMetrics() {
+    votesLock.writeLock().lock();
+    timingLock.writeLock().lock();
     lock.writeLock().lock();
     try {
       blockCountGauge =
@@ -287,6 +291,8 @@ class Store implements UpdatableStore {
       states.startMetrics();
       checkpointStates.startMetrics();
     } finally {
+      votesLock.writeLock().unlock();
+      timingLock.writeLock().unlock();
       lock.writeLock().unlock();
     }
   }
@@ -305,7 +311,7 @@ class Store implements UpdatableStore {
   public StoreTransaction startTransaction(
       final StorageUpdateChannel storageUpdateChannel, final StoreUpdateHandler updateHandler) {
     return new tech.pegasys.teku.storage.store.StoreTransaction(
-        spec, this, lock, storageUpdateChannel, updateHandler);
+        spec, this, lock, timingLock, storageUpdateChannel, updateHandler);
   }
 
   @Override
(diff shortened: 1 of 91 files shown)


tbenr:
An example of `StoreVoteUpdater::applyForkChoiceScoreChanges` locking sync committee and attestation processing

image

tbenr:
another example
image

tbenr:
Some results:
old and new code deployed around the same time on goerli-04 and goerli-05 respectively

Geern: Old locking (https://github.com/ConsenSys/teku/commits/0d89a86)
Yellow: New locking (https://github.com/ConsenSys/teku/commits/5fa616c)

TLDR; attestation validation are taking less time to execute

`total_attestation_validations / attestation_validations_running_in_<interval_ms>` (a percentage)

New code: around 75% of the attestation validation is taking less than 25ms
Old code: slightly above 55% of the attestation validation is taking less than 25ms

image

image

On new code there are less attestation taking >100ms compared to the old one
image

Similarly for Sync Committee Validation timings:

image

image

tbenr:
Might be a coincidence but in the same time frame on the box running new code we have:
image

And on the old we have 
image

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -72,8 +72,12 @@ class Store implements UpdatableStore {
 
   private final int hotStatePersistenceFrequencyInEpochs;
 
+  private final ReadWriteLock votesLock = new ReentrantReadWriteLock();
+  private final Lock readVotesLock = votesLock.readLock();
```
ajsutton, line 76: Maybe for a follow up PR but I wonder if we should be splitting vote tracking into a separate class rather than having it in `Store`. I think (not sure) we already use a specific VoteUpdater interface for votes anyway so the concerns are pretty separate, would be interesting to see if we can make them completely separate.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -72,8 +72,12 @@ class Store implements UpdatableStore {
...
   private final ReadWriteLock lock = new ReentrantReadWriteLock();
   private final Lock readLock = lock.readLock();
+  private final ReadWriteLock timingLock = new ReentrantReadWriteLock();
+  private final Lock readTimingLock = timingLock.readLock();
```
ajsutton: I'm somewhat nervous about the timing lock being separate because the current time is a really important concept and things like onTick update both the time and current justified checkpoint but they're now under separate locks and so may become visible at different times. It may be ok but much less clear than the separate votes lock.

ajsutton (commented):
I'm not convinced that it's safe to use a separate lock for time because time and current justified checkpoint are entwined. `ForkChoiceUtil` has a number of places where it uses both the time and the justified checkpoint to make decisions and it may now get an inconsistent view - particularly `isPreviousEpochJustified`. It might be ok but I'm not convinced either way.

Would be good to have the separate votes lock by itself in a PR and we can merge that while we do some more analysis on the time lock side.

tbenr:
I see what you mean with `isPreviousEpochJustified`. Reverting back to the dedicated time lock

tbenr:
For the records, it turned out that goerli 04 and 05 where not comparable because 04 was performing worse than 05 independently by the code change. BTW I updated the internal spreadsheet in drive with additional numbers (still including votes and time lock)

ajsutton (approved):
LGTM. That's a crazy small change for the impact it has. :D

ajsutton:
BTW I have a suspicion that those fork choice functions will work out ok because they always run on the fork choice thread, but it does make me quite nervous that there could be traps elsewhere.

tbenr:
What I'm also thinking WRT time lock is this: we start the time service that runs every 500ms at startup. So our tick starts at a random distance wrt the actual start of the second. I don't have an idea of the attestation arrival distribution around the attestation due time, but if the "shape" of it significantly changes in a 500ms window it means that some time we are "lucky" to run onTick "far from" a burst of attestation receival, sometimes we run it at the very peak. And this is chosen at startup. I might be overthinking here, but just wanted to share :-)

ajsutton:
> What I'm also thinking WRT time lock is this: we start the time service that runs every 500ms at startup. So our tick starts at a random distance wrt the actual start of the second. I don't have an idea of the attestation arrival distribution around the attestation due time, but if the "shape" of it significantly changes in a 500ms window it means that some time we are "lucky" to run onTick "far from" a burst of attestation receival, sometimes we run it at the very peak. And this is chosen at startup. I might be overthinking here, but just wanted to share :-)

Yes agreed, we should make sure our ticks align with the start of the slot when we start them off. Although ultimately I'd love to get rid of the whole tick system entirely, just that opens up these race condition questions that we'd have to be careful to handle properly.

Issue #6185, opened 2022-09-12
Support voluntary-exit for validators added via REST API


ajsutton:
### Description
Currently the voluntary-exit subcommand requires passing the keys to exit via the `--validator-keys` option which works, but isn't ideal for keys that are added via the validator REST API. We should investigate suitable options for improving the usability of this subcommand with REST API added validators.

One option would be to support specifying the `--data-path`  and `--data-validator-path` options (would have to be optional) so the subcommand knows where to validators added via the rest api store their keys, then allow the user to specify just the public keys of validators to exit.

Notably we need to provide a way to exit some but not all validators added by the rest api.

PR #6250, opened 2022-09-26
Exit validators with public key

mehdi-aouadi:
## PR Description
Support voluntary exit for validators added via REST API.
The `voluntary-exit` subcommand can be used with a validator public keys filter option to support the validators added via REST API: `--public-keys-filter`.
Example: `teku voluntary-exit --beacon-node-api-endpoint=<ENDPOINT>[,<ENDPOINT>...]... --data-validator-path=<PATH> --include-keymanager-keys=<BOOLEAN> --validator-keys=<KEY_DIR>:<PASS_DIR> | <KEY_FILE>:<PASS_FILE> --public-keys-filter=<PUBKEY>[,<PUBKEY>...]...`
To include validator keys managed via keymanager APIs, the option `--include-keymanager-keys` could be set to `true` (The default value is set to `false`).

## Fixed Issue(s)

[#6185](https://github.com/consensys/teku/issues/6185)

## Documentation

## Changelog


teku/src/integration-test/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommandTest.java
```diff
@@ -0,0 +1,158 @@
...
+          "--beacon-node-api-endpoint",
+          getMockBeaconServerEndpoint(mockBeaconServer),
+          "--confirmation-enabled",
+          "false",
+          "--validator-public-keys",
+          pubKey + "," + BLSTestUtil.randomPublicKey(17).toString()
+        };
+    int parseResult = beaconNodeCommand.parse(argsNetworkOptOnParent);
+    assertThat(parseResult).isEqualTo(0);
+    assertThat(stdOut.toString(UTF_8)).contains("Exit for validator 82c2a92 submitted.");
```
tbenr: I think we could have only 2 keys instead of 10 in the `key-manager` dir.
It could be good, instead, to have a key in the standard key management specifying `--validator-keys` to check the interaction with this other additional key set.

and for completeness we can assert that the stdOut doesn't contain the Exit output for the key that wasn't specified in the argument but present in the key store. Just to be sure that we are not exiting other keys.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
```diff
@@ -122,6 +125,17 @@ public void configure(TekuConfiguration.Builder builder) {
                     convertToPath(validatorExternalSignerTruststorePasswordFile)));
   }
 
+  public void configure(
+      final TekuConfiguration.Builder builder, final ValidatorClientDataOptions dataOptions) {
+    final DataDirLayout dataDirLayout = DataDirLayout.createFrom(dataOptions.getDataConfig());
+    validatorKeys =
```
tbenr: I'm not super convinced here.
If I understand correctly you are "emulating" a validatorKey option.
I'd expected to have the ability to exit all kind of keys by specifing the new param `--validator-public-keys`.

So if you specify  `--data-validator-path`, `--validator-keys` and `--validator-public-keys` the command loads all managed keys and exit only those matching the provided pub keys.
This way you could also exit a specific pub key by providing `--validator-keys` pointing to a path and not a specific json.

teku/src/integration-test/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommandTest.java
mehdi-aouadi: I agree but the problem is that the default data path points to the machine FS.
I will reduce the keys though

mehdi-aouadi: Actually the `--validator-keys` and `--validator-public-keys` are mutually exclusive (`--validator-keys` will be ignored if `--validator-public-keys` is provided)

mehdi-aouadi: I will reduce the keys though

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
mehdi-aouadi: Check my comment above. We don't mix `--validator-keys` and `--validator-public-keys` otherwise it won't be possible to filter the validators to exit

tbenr: Why can't you filter in the `run` method?

mehdi-aouadi: Because the validators are mixed when loaded and the `--validator-keys` are supplied as paths (Strings) and not pubkeys.
Theoretically they could be but that would require a paths to pubkeys service

mehdi-aouadi:
@tbenr After our discussion your comments make sense now :) 
I implemented the discussed logic and added a `--public-keys-filter` and a `--skip-keymanager-keys` options

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -104,6 +108,26 @@ public class VoluntaryExitCommand implements Runnable {
       arity = "1")
   private boolean confirmationEnabled = true;
 
+  @CommandLine.Mixin(name = "Data")
+  private ValidatorClientDataOptions dataOptions;
+
+  @CommandLine.Option(
+      names = {"--public-keys-filter"},
+      description = "Filter the exiting validators by public keys.",
```
tbenr: I would stick to `--validator-public-keys`. A filter is a technical thing IMO.
And I'd then specify better in the description what it does, something like:
`Optionally restrict the exit command to a specified list of public keys. When the parameter is not used, all keys will be exited`

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -104,6 +108,26 @@ public class VoluntaryExitCommand implements Runnable {
...
+      paramLabel = "<STRINGS>",
+      split = ",",
+      arity = "0..1")
+  private List<String> pubKeyFilter;
+
+  @CommandLine.Option(
+      names = {"--skip-keymanager-keys"},
```
tbenr: Thinking twice lets put it in the positive form by having `--include-keymanager-keys` which is true by default
and having a description like "Include validator keys managed via keymanager APIs"

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -257,10 +292,19 @@ private void initialise() {
             new PublicKeyLoader(),
             asyncRunner,
             metricsSystem,
-            Optional.empty());
+            dataDirLayout);
 
     try {
       validatorLoader.loadValidators();
+      pubKeysToExit.ifPresent(
```
tbenr: you could name the variable `maybePubKeysToExit` so the inner will be `pubKeysToExit`

teku/src/integration-test/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommandTest.java
```diff
@@ -0,0 +1,362 @@
...
+
+  private final String validatorPubKey1 =
+      "0xa756543ed1f0bac08480d67b1d5ae30f808db65de40bd2006f41a727e4f4200d732cdbee4c81aa930e2ffdf561f0dd25";
+
+  private final String validatorPubKey2 =
+      "0xaa516161fd3d6857ac45a89b7f8f9ba604a8c87642e5ef9ccdab46d7dc4e32518a17a371673758073cb81c0d536cbff1";
+
+  private final String nonExistingKey = BLSTestUtil.randomPublicKey(17).toString();
+
+  private final String EXIT_SUBMITTED = "Exit for validator %s submitted.";
```
tbenr: you may need to add `static` here

tbenr (commented):
It looks good, just some CLI param renaming which IMO should improve UX

Don't forget to add the doc label :)

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -257,10 +292,19 @@ private void initialise() {
...
+          blsPublicKeys ->
+              validatorLoader.getOwnedValidators().getActiveValidators().stream()
```
tbenr: I'd also do something like:
```java
      final OwnedValidators ownedValidators = validatorLoader.getOwnedValidators();

      maybePubKeysToExit.ifPresent(
          pubKeysToExit ->
                  ownedValidators.getActiveValidators().stream()
                  .filter(validator -> !blsPublicKeys.contains(validator.getPublicKey()))
                  .forEach(
                      validator ->
                              ownedValidators
                              .removeValidator(validator.getPublicKey())));
      validators = ownedValidators;
```

to reduce the interaction with the loader and work with a local owned validators

tbenr: I'd ideally work with a local data structure to avoid calling `OwnedValidators::removeValidator` which is a bit scary if in the future we add some logic that persists something somewhere...

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
mehdi-aouadi: Done

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
mehdi-aouadi: I did the opposite: When there is filter on public keys, I pick the concerned validators and add them to the `OwnedValidators` `validators` variable. Plus I used a local data structure as suggested

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
mehdi-aouadi: Ok for the inversion but I think in that case it should be false by default to handle the case when the user doesn't specify a `--data-validator-path` (which has a default value) and only use the `--validator-keys` option. In that case the "key manager" validators will be exited without the user asking for it (if the default value is `true`)

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
tbenr: yeah... better, any chance to not use`OwnedValidators` at all? so we use only getter and we build our own underlining `Map`, so we will not call any setter\write method against it, just want to be detached from it as much as possible

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
tbenr: ah yes --skip-keymanager-keys has true by default so --include-keymanager-keys false by default

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
mehdi-aouadi: I completely removed the `OwnedValidators` and replaced it by a local `HashMap`. I hope that's what you had in mind

tbenr (approved):
LGTM

PR #6254, opened 2022-09-27
Remove string concatenations from format() args

jtraglia:
## PR Description

Updates a few instances of this pattern, where `String.format` is used with a concatenated string argument. This PR only fixes one real issue, which is in the import command; that output would be wrong. The others just look better IMO.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #6255, opened 2022-09-27
Access fields/methods more directly

jtraglia:
## PR Description

Some real minor changes:

* `java.sql.Date#from` would call `java.util.Date#from`, might as well call it directly.
* `FAILED_DESCENDANT_OF_INVALID_BLOCK` is defined in `BlockImportResult`.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #6261, opened 2022-09-29
Fix exclusivity check between sentry nodes and beacon-api-endpoins options

lucassaldanha:
Simplified the logic using https://picocli.info/#_mutually_exclusive_options


courtneyeh (approved):
LGTM

PR #6269, opened 2022-09-30
Add fuzzy round-trip SSZ serialization tests for datastructs

jtraglia:
## PR Description

This PR adds round-trip SSZ serialization property tests (essentially fuzzing) to the existing `roundTrip` test methods. Previously, this method only tested JSON serialization/deserialization. The extra overhead in adding this is pretty minimal as generating the random data structures is the time consuming part of the test. No bugs were found. This functionality is already very well tested, but it may help catch a regression in the future. Who knows 🤷

* Add some new methods to `DataStructureUtil` that take zero arguments.
* Add overridden `getSchema` methods for some types that were missing it.
* Add new `Supplier` classes for each data type.
  * This looks cleaner & makes it way easier to add more tests in the future.

It may be easier to review these changes in an IDE.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/blocks/BeaconBlockBodyPropertyTest.java
+++ b/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/blocks/BeaconBlockBodyPropertyTest.java
@@ -16,27 +16,18 @@ package tech.pegasys.teku.spec.datastructures.blocks;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-import net.jqwik.api.Arbitraries;
-import net.jqwik.api.Arbitrary;
-import net.jqwik.api.Combinators;
 import net.jqwik.api.ForAll;
 import net.jqwik.api.Property;
-import net.jqwik.api.Provide;
 import org.apache.tuweni.bytes.Bytes;
 import tech.pegasys.teku.infrastructure.json.JsonUtil;
 import tech.pegasys.teku.infrastructure.json.types.DeserializableTypeDefinition;
-import tech.pegasys.teku.spec.Spec;
-import tech.pegasys.teku.spec.SpecMilestone;
-import tech.pegasys.teku.spec.TestSpecFactory;
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.BeaconBlockBody;
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.BeaconBlockBodySchema;
-import tech.pegasys.teku.spec.networks.Eth2Network;
-import tech.pegasys.teku.spec.util.DataStructureUtil;
 
 public class BeaconBlockBodyPropertyTest {
   @Property
   @SuppressWarnings("unchecked")
-  void roundTrip(@ForAll("beaconBlockBody") final BeaconBlockBody body)
+  void roundTrip(@ForAll(supplier = BeaconBlockBodySupplier.class) final BeaconBlockBody body)
       throws JsonProcessingException {
     final BeaconBlockBodySchema<?> schema = body.getSchema();
     final DeserializableTypeDefinition<BeaconBlockBody> typeDefinition =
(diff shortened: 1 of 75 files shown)


ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/blocks/BeaconBlockBodyPropertyTest.java
```diff
@@ -18,34 +18,29 @@
...
+    final Bytes ssz = body.sszSerialize();
+    final BeaconBlockBody fromSsz = schema.sszDeserialize(ssz);
+    assertThat(fromSsz).isEqualTo(body);
+
+    // Round-trip JSON serialization.
     final String json = JsonUtil.serialize(body, typeDefinition);
-    final BeaconBlockBody result = JsonUtil.parse(json, typeDefinition);
-    assertThat(result).isEqualTo(body);
+    final BeaconBlockBody fromJson = JsonUtil.parse(json, typeDefinition);
+    assertThat(fromJson).isEqualTo(body);
```
ajsutton, line 44: I probably should have thought of this sooner but we could extract this into a generic helper method to reduce duplication:

```
public class PropertyTestHelper {
  public static <T extends SszData, S extends SszSchema<T>> void assertRoundTrip(final T data)
      throws JsonProcessingException {
    final S schema = (S) data.getSchema();
    // Round-trip SSZ serialization.
    final Bytes ssz = data.sszSerialize();
    final T fromSsz = schema.sszDeserialize(ssz);
    assertThat(fromSsz).isEqualTo(data);

    // Round-trip JSON serialization.
    final DeserializableTypeDefinition<T> typeDefinition = schema.getJsonTypeDefinition();
    final String json = JsonUtil.serialize(data, typeDefinition);
    final T fromJson = JsonUtil.parse(json, typeDefinition);
    assertThat(fromJson).isEqualTo(data);
  }
}
```

then we could use this for any SSZ data type.

ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/blocks/BeaconBlockBodySupplier.java
```diff
@@ -0,0 +1,37 @@
...
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+
+public class BeaconBlockBodySupplier implements ArbitrarySupplier<BeaconBlockBody> {
+  @Override
+  public Arbitrary<BeaconBlockBody> get() {
+    Arbitrary<Integer> seed = Arbitraries.integers();
+    Arbitrary<SpecMilestone> milestone = Arbitraries.of(SpecMilestone.class);
+    Arbitrary<Eth2Network> network = Arbitraries.of(Eth2Network.class);
+    Arbitrary<Spec> spec = Combinators.combine(milestone, network).as(TestSpecFactory::create);
+    Arbitrary<DataStructureUtil> dsu = Combinators.combine(seed, spec).as(DataStructureUtil::new);
```
ajsutton, line 34: Would it be possible and make sense to use a `DataStructureUtilSupplier` to create the `DataStructureUtil` instance and not need to duplicate this in each supplier?

Could use a common base class like:
```java
public abstract class DataStructureSupplier<T> implements ArbitrarySupplier<T> {
  private final Function<DataStructureUtil, T> accessor;

  protected DataStructureSupplier(final Function<DataStructureUtil, T> accessor) {
    this.accessor = accessor;
  }

  @Override
  public Arbitrary<T> get() {
    Arbitrary<Integer> seed = Arbitraries.integers();
    Arbitrary<SpecMilestone> milestone = Arbitraries.of(SpecMilestone.class);
    Arbitrary<Eth2Network> network = Arbitraries.of(Eth2Network.class);
    Arbitrary<Spec> spec = Combinators.combine(milestone, network).as(TestSpecFactory::create);
    Arbitrary<DataStructureUtil> dsu = Combinators.combine(seed, spec).as(DataStructureUtil::new);
    return dsu.map(accessor);
  }
}
```
 and then the supplier would be like:
 ```java
public class BeaconBlockHeaderSupplier extends DataStructureSupplier<BeaconBlockHeader> {
  public BeaconBlockHeaderSupplier() {
    super(DataStructureUtil::randomBeaconBlockHeader);
  }
}
```

ajsutton: Ah - the filtering on milestone changes in a few cases so might need to be able to pass the minimum required milestone up to the `DataStructureSupplier` class too.

ajsutton (approved):
LGTM. I've made some suggestions for how to reduce code duplication but I think I'll merge this as-is given it improves test coverage and provides the suppliers which are useful so it's clearly a good step forward. Then we can remove the de-duplication in follow up PRs.

jtraglia:
Thanks! I agree, I wanted to deduplicate I just wasn't quite sure the right way to do it. You're suggestions looks good. I'll make some PRs for those.

PR #6271, opened 2022-10-02
Acceptance tests for Sentry Nodes feature

lucassaldanha:
## PR Description

Running a remote validator using 3 beacon nodes, one per role. Verifies that each beacon node is being used for their designated purpose.

## Fixed Issue(s)
Related to #5929

### DIFF
--- a/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
+++ b/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
@@ -58,8 +58,9 @@ public class TekuValidatorNode extends Node {
     super(network, TEKU_DOCKER_IMAGE_NAME, version, LOG);
     this.config = config;
     if (config.configMap.containsKey("validator-api-enabled")) {
-      container.withExposedPorts(VALIDATOR_API_PORT);
+      container.addExposedPort(VALIDATOR_API_PORT);
     }
+    container.addExposedPort(METRICS_PORT);
 
     container
         .withWorkingDirectory(WORKING_DIRECTORY)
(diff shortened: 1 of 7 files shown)


acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
```diff
@@ -222,9 +268,34 @@ public TekuValidatorNode.Config withInteropValidators(
...
+
+      return this;
+    }
+
     public void writeConfigFile() throws Exception {
       final File configFile = File.createTempFile("config", ".yaml");
       configFile.deleteOnExit();
+
+      if (isUsingSentryNodeConfig) {
+        configMap.remove("beacon-node-api-endpoint");
```
rolfyone, line 296: does this reduce functionality available to the dsl? just wondering if future AT's won't understand why some api functions don't work, might be worth a log message...

lucassaldanha: Basically, one can't work with the other. If you are using sentry nodes, you cannot set a beacon node api endpoint. However, your duties provider sentry node will be used as your beacon node endpoint.

lucassaldanha: For reference: https://github.com/ConsenSys/teku/blob/master/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorClientOptions.java#L105-L125

rolfyone: as long as dsl functions like getting data via rest api still work, I'm ok with that :)

lucassaldanha: It will work exactly the same way :)

rolfyone (approved):
LGTM

Issue #3493, opened 2021-01-27
Implement Doppelganger Detection


benjaminion:
For discussion.

### Description
As a staker, 
I want my node to ensure that its validators are unique on the network when it starts,
so that I don't get slashed if another instance of the validator is running elsewhere. 

One way this could be accomplished is to wait 2 (say) epochs before starting to attest and watch the network. If any attestations are seen from validators I believe I own, this should trigger the kill-switch.

It should be optional (default, but able to override?) as staking services are unlikely to want to miss epochs.

A small number of instances of slashing on mainnet may have been avoidable with such a mechanism. Current folklore in the staking community is to wait at least two epochs after stopping a node before restarting; this mechanism would effectively automate that.

- [x] Add doppelganger detection activation flag
- [x] Add validators liveness check to Validator API Channel
- [x] Add doppelganger detection to Validator Client Service
- [x] Handle the validators added via the keymanager API
- [x] Add tests

@mehdi-aouadi [notes](https://hackmd.io/St0_OdKwROKSgYzGJhaZMw?view)

rolfyone:
ref. #4453

benjaminion:
Promoting this to "implement doppelganger detection". After discussion with users - in particular Rocket Pool and Allnodes - I'm more convinced of the importance of this. It's time to raise it on the priority list.

Doppelganger detection should be off by default, and enabled by a flag.

benjaminion:
Useful, [Doppelganger Detection: Lessons Learned](https://hackmd.io/Szjk_d1gTrWGuUsLojmoqA)

PR #6274, opened 2022-10-03
Add doppelganger detection validator client option

mehdi-aouadi:
## PR Description

Add Doppelganger Detection validator client option: `--Xdoppelganger-detection-enabled` with default value set to `false`.
The option is hidden and not used yet. It's a placeholder for the Doppelganger Detection feature.

## Fixed Issue(s)

[teku #3493](https://github.com/consensys/teku/issues/3493)

## Documentation

## Changelog


StefanBratanov (approved):
LGTM. I would just add a note to the description that the option is just a placeholder and the feature is not yet implemented.

rolfyone (approved):
I'd probably add a test case to show that it gets set correctly on the config option, but overall looks good.

PR #6276, opened 2022-10-04
Fix issue which could cause locally produced aggregates to not be gossiped

ajsutton:
## PR Description
Ensures that locally produced aggregates aren't validated twice and ignored as duplicates.

Update AT to specifically check aggregate gossip counts, not just individual attestations in the full configuration.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6277, opened 2022-10-04
Rebuild forkchoice state in `getPayloadId` during on-the-fly flow

tbenr:
When in `getPayloadId` the `isPayloadIdSuitable` check fails, it may fail because the parent block doesn't match. So this PR makes sure that we regenerate forkChoiceUpdateData (state and attributes) and recheck them before proceeding.

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -51,6 +51,7 @@ import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
 import tech.pegasys.teku.spec.datastructures.util.AttestationProcessingResult;
 import tech.pegasys.teku.spec.datastructures.util.AttestationProcessingResult.Status;
 import tech.pegasys.teku.spec.executionlayer.ExecutionLayerChannel;
+import tech.pegasys.teku.spec.executionlayer.ForkChoiceState;
 import tech.pegasys.teku.spec.executionlayer.PayloadStatus;
 import tech.pegasys.teku.spec.logic.common.statetransition.exceptions.StateTransitionException;
 import tech.pegasys.teku.spec.logic.common.statetransition.results.BlockImportResult;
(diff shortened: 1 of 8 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -517,22 +520,22 @@ private void reportInvalidBlock(final SignedBeaconBlock block, final BlockImport
...
-    if (optimisticSyncing
-        .map(oldValue -> !oldValue.equals(forkChoiceState.isHeadOptimistic()))
-        .orElse(true)) {
-      optimisticSyncing = Optional.of(forkChoiceState.isHeadOptimistic());
-      optimisticSyncSubscribers.deliver(
-          OptimisticHeadSubscriber::onOptimisticHeadChanged, forkChoiceState.isHeadOptimistic());
-    }
+    forkChoiceStateProvider
+        .getForkChoiceState()
+        .thenAccept(
```
ajsutton: It doesn't seem good that this wound up becoming async. We could wind up not being on the fork choice thread or introducing other race conditions here. With the current code it will be sync but it would be way too easy to accidentally change that in the future and cause problems.

Hard to know exactly what to do about this, probably need to either have a `ForkChoiceStateProvider.getForkChoiceStateSync()` that just checks its already on the right thread and returns sync or make `ForkChoiceStateProvider` always expect to be on the right thread and make `ForkChoiceNotifier` handle getting onto the right thread.

Third option would be to make `ForkChoiceStateProvider` always sync and add a wrapper class that makes it async that `ForkChoiceNotifier` uses. They're all a bit distasteful really but I don't think we can safely assume a return `SafeFuture` completed synchronously.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifierImpl.java
```diff
@@ -155,36 +159,46 @@ private SafeFuture<Optional<ExecutionPayloadContext>> internalGetPayloadId(
...
+          .thenCompose(
+              executionPayloadContextFuture ->
+                  executionPayloadContextFuture.thenApply(
+                      maybeExecutionPayloadContext -> {
+                        if (maybeExecutionPayloadContext.isEmpty()) {
+                          throw new IllegalStateException(
+                              "Unable to obtain an executionPayloadContext");
+                        }
+                        return maybeExecutionPayloadContext;
+                      }));
```
ajsutton: We can avoid suppressing ignored futures with:
```java
      return forkChoiceStateProvider
          .getForkChoiceState()
          .thenCombine(
              proposersDataManager.calculatePayloadBuildingAttributes(
                  blockSlot, inSync, localForkChoiceUpdateData, true),
              (forkChoiceState, payloadBuildingAttributes) -> {
                forkChoiceUpdateData =
                    localForkChoiceUpdateData
                        .withForkChoiceState(forkChoiceState)
                        .withPayloadBuildingAttributes(payloadBuildingAttributes);

                sendForkChoiceUpdated();

                if (!forkChoiceUpdateData.isPayloadIdSuitable(parentExecutionHash, timestamp)) {
                  throw new IllegalStateException(
                      "payloadId still not suitable after requesting a new one via FcU with recalculated data");
                }
                return forkChoiceUpdateData;
              })
          .thenCompose(ForkChoiceUpdateData::getExecutionPayloadContext)
          .thenApply(
              maybeExecutionPayloadContext -> {
                if (maybeExecutionPayloadContext.isEmpty()) {
                  throw new IllegalStateException("Unable to obtain an executionPayloadContext");
                }
                return maybeExecutionPayloadContext;
              });
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
tbenr: yeah I see. The executor should be returning with an immediately completed future if runs on the right thread.
But trusting the internals is not good. I'm leaning towards adding a synchronous method...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifierImpl.java
tbenr: ty :)

ajsutton (approved):
LGTM.

PR #6278, opened 2022-10-04
Prefer nodes in sync when making failover requests which are not sent to all nodes

StefanBratanov:
## PR Description
For VC requests which are not relayed to all the nodes, if the primary Beacon Node is not in sync or the request fails then:
**Before**
The same request will be tried by each failover in order of configuration regardless of their syncing status.
**Now**
The request will be tried by each failover but they will be ordered by their readiness status.

## Fixed Issue(s)
improvement related to #6174 

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #6281, opened 2022-10-04
Update dependencies

tbenr:
There is also rocksdbjni major release 7.x.x (not even tried)

I opened an issue about Javalin 4->5 migration
 
## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -199,6 +199,7 @@ allprojects {
       // Generates a ton of false alarms for no real value
       check('LongDoubleConversion', net.ltgt.gradle.errorprone.CheckSeverity.OFF)
       check('InlineMeSuggester', net.ltgt.gradle.errorprone.CheckSeverity.OFF)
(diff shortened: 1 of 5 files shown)


mehdi-aouadi (approved):

tbenr:
image
🤔

zilm13:
I finally remembered what's that
it comes from unmigrated api and it's a part of `javalin-openapi`
https://github.com/javalin/javalin-openapi/blob/main/javalin-swagger-plugin/src/main/kotlin/io/javalin/openapi/plugin/swagger/SwaggerHandler.kt#L26-L74
so until removing old api our version of swagger must match the one used in `javalin-openapi`

ajsutton:
Errorprone seem to be disabling `CanIgnoreReturnValueSuggester` by default again (https://github.com/google/error-prone/issues/3366) so definitely something we could just turn off in our build config - we already enable and disable various errorprone rules to suit our needs.

tbenr:
The new errorprone spotted a nice one

PR #6282, opened 2022-10-04
3493 doppelganger detection validator api channel

mehdi-aouadi:
## PR Description
Add doppelganger detection VC <> BN streams though the Validator API Channel.
The channels are not used yet and are just placeholders. They will be used to implement a doppelganger detection service on the VC (upcoming feature).
Some implementation details (such as `DoppelgangerDetectionResult`) are subject to change according to the upcoming business logic implementation requirements.

## Fixed Issue(s)

https://github.com/consensys/teku/issues/3493

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -657,6 +665,29 @@ public SafeFuture<Void> registerValidators(
             });
   }
 
+  @Override
+  public SafeFuture<Optional<DoppelgangerDetectionResult>> checkValidatorsDoppelganger(
+      List<UInt64> validatorIndices, final UInt64 epoch) {
```
ajsutton: nit:
```suggestion
      final List<UInt64> validatorIndices, final UInt64 epoch) {
```

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -657,6 +665,29 @@ public SafeFuture<Void> registerValidators(
...
+        .getValidatorLiveness(validatorIndices, epoch, chainDataProvider.getCurrentEpoch())
+        .thenApply(
+            validatorLivenessAtEpoches ->
+                Optional.of(
+                    getDoppelgangerDetectionResultFromLiveness(validatorLivenessAtEpoches)));
+  }
+
+  private DoppelgangerDetectionResult getDoppelgangerDetectionResultFromLiveness(
+      final Optional<List<ValidatorLivenessAtEpoch>> maybeValidatorsLiveness) {
+    Object2BooleanMap<UInt64> validatorsLiveness = new Object2BooleanArrayMap<>();
```
ajsutton: nit: 
```suggestion
    final Object2BooleanMap<UInt64> validatorsLiveness = new Object2BooleanArrayMap<>();
```

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -657,6 +665,29 @@ public SafeFuture<Void> registerValidators(
...
+    maybeValidatorsLiveness.ifPresent(
+        validatorLivenessAtEpoches ->
+            validatorLivenessAtEpoches.forEach(
+                validatorLivenessAtEpoch ->
+                    validatorsLiveness.put(
+                        validatorLivenessAtEpoch.getIndex(), validatorLivenessAtEpoch.isLive())));
+    return new DoppelgangerDetectionResult(validatorsLiveness);
```
ajsutton: Why aren't we just returning the `VlaidatorsLiveness`? We want the return value for this method to be pretty much the same as what the REST API would return (we'll need to implement this method in `RemoteValidatorApi` by making a call to the liveness REST API).

It _can_ differ when there's a good reason to but given we have `NodeDataProvider.getValidatorLiveness` already returning reasonable internal types I'd have thought we'd just use those same types here.

mehdi-aouadi: Encapsulating the result into `DoppelgangerDetectionResult` was an anticipation for the business logic implementation (Simplifying the result by removing the epoch since it will always be the same for all the validators and offering some helper methods...) and to avoid coupling the VC to the beacon API.
But we can definitely use the `List<ValidatorLivenessAtEpoch>>` returned by the `NodeDataProvider` without any encapsulation.
I'll change it to `List<ValidatorLivenessAtEpoch>>` and we'll see in the next step if such change is required.

ajsutton (approved):
LGTM.

PR #6286, opened 2022-10-04
Add DataStructureUtilSupplier to reduce duplication

jtraglia:
## PR Description

Thanks again, @ajsutton! This looks a lot better.

* Added new abstract `DataStructureUtilSupplier` class.
  * Takes a DSU method that provides some random data structure.
  * Optionally, you can provide the milestone at which the structure was introduced.
  * Really helps cut down in code duplication.
* Added new `randomSszPublicKey` and `randomSszSignature` methods to the DSU.
  * Previously these were mapped in `get`.
* Renamed `SignedBeaconBlockProvider.java` to `SignedBeaconBlockSupplier.java`.
  * Used the wrong word in the original PR.
* Fixed a couple instances of blinded block tests using the non-blinded block suppliers.
  * I did this in another PR, but I just want to be certain it gets fixed.
  * I think I missed one in the other PR.

## Documentation

## Changelog


ajsutton (approved):
LGTM. Thanks.

Issue #6289, opened 2022-10-05
Improve `validatorRegistration` signing considering external signer


tbenr:
We currently try to sign all `validatorRegistration` concurrently, which is not a good strategy when remote signer is in use.

We have to consider the worse case in which we are already performing duties and we suddenly have to sign 1k `validatorRegistration` objects and we don't want to overload the signer (and outgoing tcp connections) with "low priority" signatures that are way less important than real duty signatures. This can cause delays or even timeouts that can lead to bad attestation performances.

- Improvement 1: limit the signing concurrency.
So an improvement could be to split the validators stream in `ValidatorRegistrator#registerValidators` batches and `collectAll` each batch before moving to the next batch, so we well have a cap on the number of concurrent sign requests.

Another, more structural, approach could be to implement a throttling on the outgoing external signer requests (which will apply on all requests). But it would be difficult to decide the actual limit and will require to implement a priority scheme which should give precedence to signature on duties over the validator signatures.

- Improvement 2: better handling of signing errors (ie timeouts)
We currently fail the entire flow if we have just one failing signing future. We should let the registration process continue excluding the failed ones (will be retried automatically the next round)

ajsutton:
Note that we already throttle signatures for checking if we're the aggregator for a slot in `ExternalSigner`.  We should probably update `signValidatorRegistration` to add tasks to the `taskQueue` so it uses the same throttling.  Not sure if the upstream code needs to be updated to send registrations in batches or something rather than waiting for all signatures to complete given they could take some time now.

PR #6291, opened 2022-10-05
Don't fail the entire registration flow when a signing fails

StefanBratanov:
## PR Description
**Before:**
if there is a signing error for one validator, the entire registration flow fails.
**After the change in this PR:**
If there is a signing error for one or more validator, those validators won't be registered and signing will be retried next epoch. All successful signed registrations will be sent.

## Fixed Issue(s)
capturing Improvement 2 of #6289 

## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
+++ b/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
@@ -212,9 +212,7 @@ public class SafeFuture<T> extends CompletableFuture<T> {
     final Stream<SafeFuture<Optional<T>>> optionalFutures =
         futures.map(future -> future.thenApply(Optional::of).exceptionally(__ -> Optional.empty()));
     return collectAll(optionalFutures)
-        .thenApply(
-            results ->
-                results.stream().filter(Optional::isPresent).map(Optional::get).collect(toList()));
+        .thenApply(results -> results.stream().flatMap(Optional::stream).collect(toList()));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
```diff
@@ -192,6 +193,28 @@ public static <T> SafeFuture<List<T>> collectAll(final SafeFuture<T>... futures)
...
+   * each result in the same order as the futures were given filtering out the exceptionally
+   * completed futures.
+   *
+   * @param futures the futures to collect results from
+   * @param <T> the result type to collect
+   * @return a new future that completes when all the supplied futures complete
+   */
+  public static <T> SafeFuture<List<T>> collectAllSuccessful(final Stream<SafeFuture<T>> futures) {
+    final Stream<SafeFuture<Optional<T>>> optionalFutures =
+        futures.map(future -> future.thenApply(Optional::of).exceptionally(__ -> Optional.empty()));
```
tbenr, line 213: what about having an optional callback to be able to log errors? Or you can return a Pair of future lists, successful and failed list

tbenr: oh you're already intercepting the error...

tbenr: but you're intercepting only when the signer fails. If the exception occurs in some other places the error is lost

tbenr (approved):
LGTM. My comments where a bit off track but you got the point :-)

PR #6292, opened 2022-10-05
Add deserializeMutated test for datastructures

jtraglia:
## PR Description

The current property tests deserialize input that should be deserializable. This PR adds tests that deserialize data structures that have been slightly mutated. I believe this is smarter than deserializing random bytes. This PR is focussed on SSZ deserialization; we could do the same thing for JSON deserialization, but it would be a bit more difficult.

Overview of changes:

* Add new `Mutator` class with a static `mutate` method.
* Add new `deserializeMutated` method to each data structure property test file.
* Add new `assertDeserializeMutatedThrowsExpected` method to `PropertyTestHelper`.
  * Expected exceptions are `SszDeserializeException` and `IllegalArgumentException`.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/util/Mutator.java
+++ b/ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/util/Mutator.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright ConsenSys Software Inc., 2022
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.spec.datastructures.util;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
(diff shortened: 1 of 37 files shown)


ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/util/Mutator.java
```diff
@@ -0,0 +1,131 @@
...
+  public static Bytes mutate(final Bytes input, final int seed) {
+    final Random random = new Random(seed);
+    final List<Bytes> sections = new ArrayList<>();
+
+    for (final byte b : input.toArray()) {
+      final double randomNumber = Math.abs(random.nextGaussian());
+      if (randomNumber <= MUTATION_PROBABILITY) {
+
+        Mutation op = null;
+        for (final Map.Entry<Mutation, Range> entry : RANGE_MAP.entrySet()) {
```
tbenr: You might use a `TreeMap<Double,Mutation>.floorEntry` and you don't need a `Range` but just the upper bound of the probability

tbenr (commented):
Looks Cool.

ethereum/spec/src/property-test/java/tech/pegasys/teku/spec/datastructures/util/Mutator.java
jtraglia: Good idea. I'll look into that!

jtraglia: Really great suggestion. Thanks. I think this looks a lot better. I ended up using `TreeMap.ceilingEntry` instead of `floorEntry` because calling `floorEntry` with a number greater than the largest key would return the last entry. I think `ceilingEntry` what you meant to suggest anyway, right?

ajsutton (approved):
LGTM. Thanks.

I've pushed one little tweak to how the mutation is done so that it uses `ByteArrayOutputStream` rather than having to create so many individual `Bytes` instances. It probably helps performance but also just reads a bit more naturally.

jtraglia:
Nice. That change looks good!

PR #6298, opened 2022-10-06
teku-6236 Added event channel for execution client availability report

lucassaldanha:
Created a new interface `ExecutionClientEventsChannel` for any ExecutionClient event. Atm it only has a `onAvailabilityUpdated(bool)` method.

Updated `Web3JClient` to keep track of the EL's availability and publish an event when it changes.

Before any requests to the EL, we assume the its availability = false.

### DIFF
--- a/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
+++ b/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
@@ -18,7 +18,6 @@ import static tech.pegasys.teku.infrastructure.exceptions.ExceptionUtil.getMessa
 import java.io.IOException;
 import java.time.Duration;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import org.web3j.protocol.Web3j;
 import org.web3j.protocol.Web3jService;
@@ -46,7 +45,6 @@ public abstract class Web3JClient {
   // available but uses a very old value to make sure we log if the first request fails
   private final AtomicLong lastError = new AtomicLong(STARTUP_LAST_ERROR_TIME);
   private boolean initialized = false;
-  private final AtomicBoolean isAvailable = new AtomicBoolean(false);
 
   protected Web3JClient(
       final EventLogger eventLog,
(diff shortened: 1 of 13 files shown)


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
```diff
@@ -36,17 +38,23 @@ public abstract class Web3JClient {
...
   private final TimeProvider timeProvider;
+  private final ExecutionClientEventsChannel executionClientEventsPublisher;
   private Web3jService web3jService;
   private Web3j eth1Web3j;
 
   // Default to the provider having a previous failure at startup so we log when it is first
   // available but uses a very old value to make sure we log if the first request fails
   private final AtomicLong lastError = new AtomicLong(STARTUP_LAST_ERROR_TIME);
   private boolean initialized = false;
+  private final AtomicBoolean isAvailable = new AtomicBoolean(false);
```
ajsutton: Shouldn't need this new boolean as we can just use the `lastError` field same as we check it in `handleSuccess` and `handleError`. Should be able to just call the publisher at the same point we're calling `eventLog.executionClientRecovered` or `logExecutionClientError`

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
```diff
@@ -103,15 +109,18 @@ protected void handleError(final Throwable error, final boolean couldBeAuthError
...
       logExecutionClientError(error, couldBeAuthError);
+      executionClientEventsPublisher.onAvailabilityUpdated(false);
     }
   }
 
   protected void handleSuccess() {
     final long lastErrorTime = lastError.getAndUpdate(x -> NO_ERROR_TIME);
     if (lastErrorTime == STARTUP_LAST_ERROR_TIME) {
       eventLog.executionClientIsOnline();
+      executionClientEventsPublisher.onAvailabilityUpdated(true);
```
lucassaldanha, line 120: I'm calling the update inside each branch to avoid publishing events with subsequent successes.

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
```diff
@@ -103,15 +109,18 @@ protected void handleError(final Throwable error, final boolean couldBeAuthError
             });
     if (maybeUpdatedTime == timeNow) {
       logExecutionClientError(error, couldBeAuthError);
+      executionClientEventsPublisher.onAvailabilityUpdated(false);
```
lucassaldanha, line 112: This call means we will publish events for subsequent failures but this is probably ok as we don't expect our EL to be failing that much ;)

ajsutton (approved):
LGTM.

PR #6301, opened 2022-10-06
Throttle validator registration external signing requests

### LINKED ISSUE #6289: Improve `validatorRegistration` signing considering external signer

StefanBratanov:
## PR Description

- Use the existing throttler in ExternalSigner for validator registrations.
- Prioritize aggregation slot signing to ensure validator duties flow is not affected.

## Fixed Issue(s)
fixes #6289 

## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/ThrottlingTaskQueue.java
+++ b/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/ThrottlingTaskQueue.java
@@ -14,15 +14,16 @@
 package tech.pegasys.teku.infrastructure.async;
 
 import com.google.common.annotations.VisibleForTesting;
-import java.util.Deque;
-import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.function.Supplier;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
 
 public class ThrottlingTaskQueue {
 
-  private final Deque<Runnable> queuedTasks = new ConcurrentLinkedDeque<>();
+  private final Queue<Runnable> queuedTasks = new ConcurrentLinkedQueue<>();
+  private final Queue<Runnable> prioritizedTasks = new ConcurrentLinkedQueue<>();
 
   private final int maximumConcurrentTasks;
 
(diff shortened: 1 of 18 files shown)


infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/ThrottlingTaskQueue.java
```diff
@@ -35,18 +39,33 @@ public ThrottlingTaskQueue(
...
     final SafeFuture<T> future = new SafeFuture<>();
-    queuedTasks.add(
+    final Runnable taskToQueue =
         () -> {
           final SafeFuture<T> requestFuture = request.get();
           requestFuture.propagateTo(future);
           requestFuture.always(this::taskComplete);
-        });
+        };
+    if (prioritize) {
```
tbenr: I think this is not good. This way you're actually making the prioritized tasks to leave on a LIFO queue which is bad

infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/ThrottlingTaskQueueTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private final StubMetricsSystem stubMetricsSystem = new StubMetricsSystem();
+
+  private final StubAsyncRunner stubAsyncRunner = new StubAsyncRunner();
+
+  private final ThrottlingTaskQueue throttlingTaskQueue =
+      new ThrottlingTaskQueue(
+          MAXIMUM_CONCURRENT_TASKS, stubMetricsSystem, TekuMetricCategory.BEACON, "default");
+
+  private final ThrottlingTaskQueueWithPriority throttlingTaskQueueWithPriority =
```
tbenr: I'd have a dedicated ThrottlingTaskQueueWithPriorityTest test class, just to follow our standard

tbenr (approved):
LGTM.

PR #6302, opened 2022-10-06
force `TimerService` to run the handler in alignment with the beginning of the second

tbenr:
as discussed here: https://github.com/ConsenSys/teku/pull/6245#issuecomment-1268053355

as per my checks, time drift is always corrected back to be close to the tick of the second

## Documentation

## Changelog

### DIFF
--- a/services/timer/src/main/java/tech/pegasys/teku/services/timer/TimerService.java
+++ b/services/timer/src/main/java/tech/pegasys/teku/services/timer/TimerService.java
@@ -88,7 +88,7 @@ public class TimerService extends Service {
                       // contention), then just skip it and fire the next event when it is due
                       .withMisfireHandlingInstructionNextWithRemainingCount()
                       .repeatForever())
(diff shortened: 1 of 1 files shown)


zilm13:
If I understand everything correctly, we have 999ms* seconds delay on start here in the worst case, but we run ticks every 500ms, so we could change it to have 499ms* in the worst case. Am I right?

zilm13:
Also when system clock will be changed by some utility few ms forward or backwards, will we handle it?

tbenr:
> If I understand everything correctly, we have 999ms* seconds delay on start here in the worst case, but we run ticks every 500ms, so we could change it to have 499ms* in the worst case. Am I right?

yes that's correct

> Also when system clock will be changed by some utility few ms forward or backwards, will we handle it?

very good question..........

tbenr:
@zilm13 completely removed the startup delay :-)

tbenr:
for the second question, I think quartz calculates next fire based on current time so it should react accordingly remaining aligned, but should be proven

ajsutton (approved):
Well that was unexpectedly simple. LGTM.

PR #6309, opened 2022-10-11
Add Doppelganger Detection Service

mehdi-aouadi:
## PR Description
Add a doppelganger detection service which runs for at least 2 epochs and checks the validators indices each 12s.
In order to prevent the service from running indefinitely (If, for some reason, the service is unable to calculate the current epoch) a timeout is configured to stop the service and continue the VC tasks. In that case an error message is logged asking the user to check the previous logs and perform a new check.
The service is started after the validators keys have been loaded and before their duties are scheduled. 
If a doppelganger is detected the VC shuts down completely (This behaviour is still subject to change).
The check is performed at the VC startup. The validators loaded via the keymanager API are not handled yet. The same service can be reused for that use case.

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/3493

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
```diff
@@ -0,0 +1,181 @@
...
+    while (true) {
+      if (doppelgangerCheckFinished.get()) {
+        return SafeFuture.COMPLETE;
+      }
+    }
+  }
+
+  @Override
+  protected SafeFuture<?> doStop() {
+    LOGGER.info("No validator doppelganger detected. Stopping doppelganger detection service.");
```
rolfyone: not sure we need an info here? doStop will be called any time the service is stopped so probably need to be selective about what the message says if we do have it...

eg. start then stop straight away will cause stop to be called and it'll report...

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
```diff
@@ -0,0 +1,181 @@
...
+    return SafeFuture.fromRunnable(
+        () -> Optional.ofNullable(doppelgangerDetectionTask).ifPresent(Cancellable::cancel));
+  }
+
+  private SafeFuture<?> checkValidatorsDoppelganger() {
+    Duration duration = Duration.of(System.nanoTime() - startTime, ChronoUnit.NANOS);
+    if (duration.compareTo(timeout) > 0) {
+      LOGGER.info("Doppelganger Detection Service max allowed duration exceeded.");
```
rolfyone: i think a message here is worthwhile, but we probably want to be clear about what seeing this message means from a user perspective...

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
```diff
@@ -0,0 +1,181 @@
...
+                                      LOGGER.fatal(
+                                          "Doppelganger detected. Shutting down Validator Client.");
+                                      System.exit(1);
+                                    }
+                                    return null;
+                                  }));
+            })
+        .exceptionally(
+            throwable -> {
+              LOGGER.error("Unable to check validators doppelganger: {}", throwable.getMessage());
```
rolfyone: again, the user will read this message, they need to understand it and potentially take action...

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
```diff
@@ -0,0 +1,181 @@
...
+              .filter(
+                  validatorLivenessAtEpoch ->
+                      validatorIndices.contains(validatorLivenessAtEpoch.getIndex()))
+              .collect(Collectors.toList());
+      if (!doppelgangers.isEmpty()) {
+        LOGGER.fatal(
+            "Doppelganger detected. Validators indices: {}",
+            doppelgangers.stream()
+                .map(validatorLivenessAtEpoch -> validatorLivenessAtEpoch.getIndex().toString())
+                .collect(Collectors.joining(",")));
```
rolfyone: given this is fatal, i guess we want to shutdown at this point - potentially what the user will find more useful is the actual abbreviated pubkeys rather than the validator indices (we should abbreviated pubkeys in the screen logger)

I think this probably needs to be in the screen logger too, not in the console logger... I think a reasonable example may be `StatusLogger.validatorsInitialized` - we'd call statusLogger and pass it a list of validators that were detected...

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -65,6 +67,8 @@
 public class ValidatorClientService extends Service {
 
   private static final Logger LOG = LogManager.getLogger();
+  private static final Duration DOPPELGANGER_DETECTION_CHECK_DELAY = Duration.ofSeconds(12);
+  private static final Duration DOPPELGANGER_DETECTION_TIMEOUT = Duration.ofMinutes(20);
```
rolfyone: 20 minutes seems a long time...
I'm not 100% sure how this is going to work, but I thought we'd be checking the keys we intend to load, and if they weren't seen as active then we'd be loading the keys, otherwise we'd be aborting before actually loading a key to be able to perform duties?
Presumably the check of keys should be fast-ish... like check last epoch, check current epoch, not seen == good to go...  This is a big assumption, im not sure of the design, so i could be wrong and we need 20 minutes, but i really hope it's not up to 20 minutes before we can load keys...

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
mehdi-aouadi: Added more details about the timeout here

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
mehdi-aouadi: Done, added more details here

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
mehdi-aouadi: Used the `StatusLogger` and tried to get the doppelgangers public keys (Handled the case where they could not be retrieved but normally that's not supposed to happen. If it's the case we only log the indices)

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
mehdi-aouadi: Reduced the duration to 15 minutes (a little bit more than 2 epochs).
This timeout is for security purposes only (to make sure that the service won't be running indefinitely). That could happen if for some obscure reasons the `getGenesisTime` keeps failing for example (There is a test for this precise use case).
Otherwise the service will perform a check every 12 seconds and for 2 epochs and then stop (In the normal case scenario).
The idea behind the 2 epochs is to give some time to the beacon node to explore the network (Especially when the `ActiveValidatorCache` is empty if the `--Xbeacon-liveness-tracking-enabled` is activated at the same time as the `--Xdoppelganger-detection-enabled`)

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -192,15 +199,50 @@ public static ValidatorClientService create(
             validatorRegistrator,
             config.getSpec(),
             services.getMetricsSystem());
+
+    validatorClientService.initializeValidators(validatorApiChannel, asyncRunner);
```
tbenr, line 202: I think that if you move initializeValidators outside the asyncRunner you defeat the purpose of it, which is postponing all slow initialization and return immediately the ValidatorClientService instance.
In this case `validatorLoader.loadValidators()` will run synchronously, which is the slow part. I'd try to reorganize things so that it remains async.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -387,19 +433,30 @@ protected SafeFuture<?> doStart() {
...
+        new ValidatorTimingActions(
+            validatorStatusLogger,
+            validatorIndexProvider,
+            validatorTimingChannels,
+            spec,
+            metricsSystem));
+    validatorStatusLogger.printInitialValidatorStatuses().ifExceptionGetsHereRaiseABug();
+    return beaconNodeApi.subscribeToEvents();
+  }
+
```
tbenr:
```suggestion
              return maybeDoppelgangerDetectionService
                  .map(service -> service.start().toVoid())
                  .orElse(SafeFuture.COMPLETE);
            })
        .thenCompose(
            __ -> {
              eventChannels.subscribe(
                  ValidatorTimingChannel.class,
                  new ValidatorTimingActions(
                      validatorStatusLogger,
                      validatorIndexProvider,
                      validatorTimingChannels,
                      spec,
                      metricsSystem));
              validatorStatusLogger.printInitialValidatorStatuses().ifExceptionGetsHereRaiseABug();
              return beaconNodeApi.subscribeToEvents();
            });
```

I'd suggest a more compact\clean version

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
```diff
@@ -0,0 +1,245 @@
...
+    doppelgangerDetectionTask =
+        asyncRunner.runWithFixedDelay(
+            this::checkValidatorsDoppelganger,
+            checkDelay,
+            checkDelay,
+            throwable ->
+                LOGGER.error(
+                    "Error while checking validators doppelganger. The check could not be performed correctly.",
+                    throwable));
+    while (true) {
```
tbenr: This is not so async. You can use a `SafeFuture` instead of an `AtomicBoolean` and let it complete once everything is done.

tbenr (commented):
Added some comments suggestions

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
mehdi-aouadi: Done. The validators initialisation is now run by the async runner and propagates to the Doppelganger Service initialisation (Validators need to be loaded)

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
mehdi-aouadi: Done. Changes accepted thanks!

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetectionService.java
mehdi-aouadi: The doppelganger detection service is not supposed to be async. In the current design it's blocking until the check is finished (When activated we must avoid any slashing risk hence blocking the validators from performing their duties. And yes this could mean penalties for approximately 2 epochs missed duties but that's an expected/accepted side effect)

tbenr: Ok I understand the intent, but you can still be async and yet avoid all duties to be performed. What I'm trying to say is that having a thread looping continuously until a flag to become true is not a good design.

An option is to pass this "doppelgangerChecked" `SafeFuture` to all duty loaders, make them loading only if the `SafeFuture` is completed.
As soon as the `SafeFuture` completes **successfully**, all duty loaders should try to load the corresponding duties and schedule them.
All loaders extends `AbstractDutyLoader` which seems a good fit to have this logic shared to all duty types.

This is just an idea to make everything more "async" while maintaining the workflow sequence you want to achieve.

tbenr: Sorry I think I over commented wrongly! Duty loaders already depends on the `SafeFuture` you return here on `doStart()` so it is just matter of returning a non completed `SafeFuture` that completes where you set the `AtomicBoolean` to true.

mehdi-aouadi: Made some refactoring:

mehdi-aouadi: Done. Used a `SafeFuture` to let the caller know when the check is finished.

mehdi-aouadi:
@tbenr as discussed I made some refactoring:

mehdi-aouadi:
@tbenr as discussed, I made some refactoring to simplify the doppelganger check and give back some actions to the caller:
- The `DoppelgangerDetection` class does not implement the `Service` interface anymore (I renamed it and removed `Service` at the end of the class name)
- No more `AtomicBoolean` usage to track the end of the processing. It's now done using a `SafeFuture` 
-  Extracted some streams operations to methods
- The action to perform when a doppelganger is detected has been removed from the `DoppelgangerDetection` class. It's now up to the caller to decide what to do in such a case
- When the check is finished, the `SafeFuture` completes with a `Map<Integer, BLSPublicKey>` containing the doppelgangers Indices/Public keys (If we are unable to get the public keys, we only return the indices, I used the `BLSPublicKey.empty()` in that case)

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -158,6 +160,29 @@ public void validatorsInitialised(final List<String> validators) {
...
+    StringBuilder doppelgangersLogInfo = new StringBuilder("\n");
+    for (Map.Entry<Integer, String> doppelgangerInfo : doppelgangersInfo.entrySet()) {
+      doppelgangersLogInfo.append(
+          StringUtils.isBlank(doppelgangerInfo.getValue())
+              ? "Index: " + doppelgangerInfo.getKey() + "\n"
+              : "Index: "
+                  + doppelgangerInfo.getKey()
+                  + ", Public key: "
+                  + doppelgangerInfo.getValue()
+                  + "\n");
```
tbenr: I'd use `String.format` and avoid string concatenation

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -158,6 +160,29 @@ public void validatorsInitialised(final List<String> validators) {
     }
   }
 
+  public void validatorsDoppelgangerDetected(final Map<Integer, String> doppelgangersInfo) {
+    StringBuilder doppelgangersLogInfo = new StringBuilder("\n");
+    for (Map.Entry<Integer, String> doppelgangerInfo : doppelgangersInfo.entrySet()) {
```
tbenr: here you could use a stream and then collect to a joiner like in `SszSchemaHints.toString()` so you avoid StringBuilder

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+import tech.pegasys.teku.infrastructure.async.Cancellable;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.exceptions.ExceptionUtil;
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.validator.api.ValidatorApiChannel;
+import tech.pegasys.teku.validator.beaconnode.GenesisDataProvider;
+
+public class DoppelgangerDetection {
```
tbenr: DoppelgangerDetector?

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+            checkDelay,
+            throwable ->
+                LOGGER.error(
+                    "Error while checking validators doppelganger. The check could not be performed correctly.",
+                    throwable));
+    return doppelgangerCheckFinished.get();
+  }
+
+  private synchronized SafeFuture<?> stopDoppelgangerDetection(
+      Map<Integer, BLSPublicKey> detectedDoppelgangers) {
```
tbenr: final Map<Integer, BLSPublicKey> detectedDoppelgangers

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+import tech.pegasys.teku.infrastructure.exceptions.ExceptionUtil;
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.validator.api.ValidatorApiChannel;
+import tech.pegasys.teku.validator.beaconnode.GenesisDataProvider;
+
+public class DoppelgangerDetection {
+
+  private static final Logger LOGGER = LogManager.getLogger();
```
tbenr: we have `LOG` everywhere

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+      throw new RuntimeException("Doppelganger Detection is already stopped.");
+    }
+    doppelgangerCheckFinished.get().complete(detectedDoppelgangers);
+    doppelgangerCheckFinished = Optional.empty();
+    return SafeFuture.fromRunnable(
+        () -> Optional.ofNullable(doppelgangerDetectionTask).ifPresent(Cancellable::cancel));
+  }
+
+  private SafeFuture<?> checkValidatorsDoppelganger() {
+    Duration duration = Duration.of(System.nanoTime() - startTime, ChronoUnit.NANOS);
```
tbenr: add `final`. Do we really need nanoseconds here? :)

tbenr: You can use `SafeFuture<Void>` hear and in stopDoppelgangerDetection  instead of `SafeFuture<?>` (it isn't a service anymore). 
so you can use `toVoid()` instead of `.thenApply(o -> null);` down there.

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+            throwable -> {
+              LOGGER.error(
+                  "Unable to check validators doppelganger. Unable to get genesis time to calculate the current epoch: {}",
+                  extractErrorMessage(throwable));
+              return null;
+            });
+  }
+
+  private SafeFuture<Void> checkValidatorsLivenessAtEpoch(
+      UInt64 currentSlot, UInt64 currentEpoch, IntCollection validatorIndices) {
```
tbenr: `final` here too

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+              })
+          .ifExceptionGetsHereRaiseABug();
+    } else {
+      LOGGER.info("No validators doppelganger detected for epoch {}, slot {}", epoch, slot);
+    }
+  }
+
+  private Map<Integer, String> mapLivenessAtEpochToIndicesPubKeysStrings(
+      List<ValidatorLivenessAtEpoch> liveValidators,
+      Map<BLSPublicKey, Integer> blsPublicKeyIntegerMap) {
```
tbenr: `final`s

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+        .thenApply(doppelgangerDetected -> null);
+  }
+
+  private void checkValidatorDoppelgangers(
+      final Optional<List<ValidatorLivenessAtEpoch>> validatorLivenessAtEpoches,
+      final List<UInt64> validatorIndices,
+      final UInt64 epoch,
+      final UInt64 slot) {
+    List<ValidatorLivenessAtEpoch> liveValidators =
+        filterLiveValidators(validatorLivenessAtEpoches, validatorIndices);
```
tbenr: `final`. I'd also rename the variable to `doppelgangers` or so

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+    if (!liveValidators.isEmpty()) {
+      LOGGER.fatal("Doppelganger detected. Shutting down Validator Client.");
+      this.validatorIndexProvider
+          .getValidatorIndicesByPublicKey()
+          .thenApply(
+              blsPublicKeyIntegerMap -> {
```
tbenr: the default coming from intelliJ sometime can be improved :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
```diff
@@ -0,0 +1,288 @@
...
+                liveValidators.stream()
+                    .anyMatch(
+                        validatorLivenessAtEpoch ->
+                            validatorLivenessAtEpoch
+                                .getIndex()
+                                .equals(UInt64.valueOf(pubKeyIndexEntry.getValue()))))
+        .collect(Collectors.toMap(Map.Entry::getValue, e -> e.getKey().toString()));
+  }
+
+  private Map<Integer, BLSPublicKey> mapLivenessAtEpochToIndicesPubKeys(
```
tbenr: don't you have some code in common with `mapLivenessAtEpochToIndicesPubKeysStrings`?

tbenr (commented):
some additional comments. several missing `final` to match our code style (I haven't marked them all)

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Yep, makes more sense. Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Will get used to make all methods params final soon, promise 😅
Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/DoppelgangerDetection.java
mehdi-aouadi: Yeah the filtering step is in common. I extracted it in a separate method...

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -192,15 +198,65 @@ public static ValidatorClientService create(
...
-            () ->
-                validatorClientService.initializeValidators(
-                    config, validatorApiChannel, asyncRunner))
-        .propagateTo(validatorClientService.initializationComplete);
+            () -> validatorClientService.initializeValidators(validatorApiChannel, asyncRunner))
+        .propagateTo(validatorsInitialized);
+
+    validatorsInitialized
+        .thenCompose(
+            __ -> {
```
tbenr: at the end here you can simply do:
image

validator/client/src/test/java/tech/pegasys/teku/validator/client/DoppelgangerDetectorTest.java
```diff
@@ -0,0 +1,436 @@
...
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.TestSpecFactory;
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+import tech.pegasys.teku.validator.api.ValidatorApiChannel;
+import tech.pegasys.teku.validator.beaconnode.GenesisDataProvider;
+
+public class DoppelgangerDetectorTest {
+  private final Spec spec = TestSpecFactory.createDefault();
+  private final DataStructureUtil dataStructureUtil = new DataStructureUtil(spec);
+  private final AsyncRunner asyncRunner = DelayedExecutorAsyncRunner.create();
```
tbenr: UT are very slow because you are actually waiting for the time passing by.
You should use `StubAsyncRunner asyncRunner = new StubAsyncRunner();`

and then call
`asyncRunner.executeQueuedActions();`

you can check several usages in other UTs.

tbenr (commented):
we are close. Just a nit on production code but some things to change on UT. We already have CI slow enough...

validator/client/src/test/java/tech/pegasys/teku/validator/client/DoppelgangerDetectorTest.java
tbenr: for instance `SyncStallDetectorTest`

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
mehdi-aouadi: Good catch, since the `runAsync` already returns a `SafeFuture` that would be cleaner...
Done

validator/client/src/test/java/tech/pegasys/teku/validator/client/DoppelgangerDetectorTest.java
mehdi-aouadi: Done. As discussed, I replaced the system time usage by the `TimeProvider` and used the `StubTimeProvider` for the UTs

validator/client/src/test/java/tech/pegasys/teku/validator/client/DoppelgangerDetectorTest.java
```diff
@@ -0,0 +1,278 @@
...
+    expectLogMessage(logCaptor.getLogEvents().get(2), "INFO", performingDoppelgangerCheckLog(0, 1));
+    expectLogMessage(
+        logCaptor.getLogEvents().get(3),
+        "INFO",
+        "No validators doppelganger detected for epoch 0, slot 1");
+    expectLogMessage(
+        logCaptor.getLogEvents().get(4),
+        "INFO",
+        "No validators doppelganger detected after 2 epochs. Stopping doppelganger detection.");
+    assertThat(doppelgangerDetectorFuture).isCompleted();
```
tbenr: you can remove those `isCompleted();` if later you check with `isCompletedWithValue`

tbenr (approved):
LGTM! just a nit.

validator/client/src/test/java/tech/pegasys/teku/validator/client/DoppelgangerDetectorTest.java
mehdi-aouadi: Done

Issue #6211, opened 2022-09-16
async runner names need to be validated


rolfyone:
when we create an async runner (`serviceConfig.createAsyncRunner` for example), metrics are created on the name given, and if the name contains certain characters like '-' it breaks metrics.

We need a test so that when we create this kind of thing, we can validate the metric name and not get into a situation where we break metrics.

mehdi-aouadi:
Isn't this the same as [check metric names are valid#11](https://app.zenhub.com/workspaces/team-centaur-5ec772b542883ab933e51688/issues/consensys/teku-internal/11)?

rolfyone:
https://github.com/ConsenSys/teku/pull/6209
fixing a metric name found when trying to push a dev build... concrete example we talked about

rolfyone:
Possibly, I thought the other was resolved, would have to check…

PR #6310, opened 2022-10-12
Validate async runner names

mehdi-aouadi:
## PR Description

Validate the async runner names so they can be used for metrics.

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6211

## Documentation

## Changelog


tbenr (approved):
LGTM even if we are introducing a regex pattern here, but It is used only during initialization.
If we want to be very strict we could replace it with manually written check @ajsutton

mehdi-aouadi:
As discussed with @tbenr I will merge this PR since the regex check is done during initialization. Regex patterns are better be avoided in general though

tbenr:
and we are applying it to our own input too

PR #6313, opened 2022-10-13
Complete builder registration filtering

### LINKED ISSUE #6312: Filter validators with all post-exit statuses when registering against builder endpoint
we already filtering out `exited_slashed` and `exited_unslashed` but we have to filter `withdrawal_*` too

tbenr:
implements a more exhaustive filtering on validator statuses when we are about to send them to the builder endpoint.

We don't want to send unknown or exited validators.

## Fixed Issue(s)

fixes #6312 

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #6324, opened 2022-10-14
Add doppelganger option to the beacon node command

mehdi-aouadi:
## PR Description

Move the doppelganger detection enabling option from the `ValidatorClientOptions` to the the `ValidatorOptions` to make it available to the Beacon Node command (Otherwise will only be available when the VC is run in a separate process from the BN).

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/3493

## Documentation

## Changelog

### DIFF
--- a/teku/src/test/java/tech/pegasys/teku/cli/BeaconNodeCommandTest.java
+++ b/teku/src/test/java/tech/pegasys/teku/cli/BeaconNodeCommandTest.java
@@ -365,11 +365,11 @@ public class BeaconNodeCommandTest extends AbstractBeaconNodeCommandTest {
     beaconNodeCommand.parse(args);
     assertThat(
             beaconNodeCommand
-                    .tekuConfiguration()
-                    .validatorClient()
-                    .getValidatorConfig()
-                    .isDoppelgangerDetectionEnabled())
-            .isFalse();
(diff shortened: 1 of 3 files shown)


teku/src/test/java/tech/pegasys/teku/cli/BeaconNodeCommandTest.java
```diff
@@ -332,6 +333,32 @@ public void shouldSetNatMethod() {
         .isEqualTo(NatMethod.UPNP);
   }
 
+  @Test
+  public void doppelgangerDetectionShouldBeDisabledByDefault() {
```
zilm13, line 337: It looks like this testcase should be fixed to follow its name goal

zilm13 (changes requested):
LGTM, just one testcase should be fixed

teku/src/test/java/tech/pegasys/teku/cli/BeaconNodeCommandTest.java
mehdi-aouadi: I fixed it and added another one to test the feature enabling/disabling separately

zilm13 (approved):
LGTM

PR #6330, opened 2022-10-19
Fix /readiness endpoint always returning 503

### LINKED ISSUE #6329: /readiness API always returns 503
### Description

https://github.com/ConsenSys/teku/pull/6306 introduced a bug where the `/readiness` API is always returning 503.

### Steps to Reproduce (Bug)
N/A

**Expected behavior:**
Readiness endpoint should respond accordingly to the "readiness" rules.

**Actual behavior:**
Readiness endpoint always returns 503, even when the node is ready.

**Frequency:**
Always.

### Versions
* Teku 22.10.1

lucassaldanha:
Fixes #6329

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/ExecutionClientDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/ExecutionClientDataProvider.java
@@ -18,7 +18,7 @@ import tech.pegasys.teku.ethereum.executionclient.events.ExecutionClientEventsCh
 
 public class ExecutionClientDataProvider implements ExecutionClientEventsChannel {
 
-  private final AtomicBoolean isExecutionClientAvailable = new AtomicBoolean(false);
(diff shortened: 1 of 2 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/ExecutionClientDataProvider.java
```diff
@@ -18,7 +18,7 @@
 
 public class ExecutionClientDataProvider implements ExecutionClientEventsChannel {
 
-  private final AtomicBoolean isExecutionClientAvailable = new AtomicBoolean(false);
+  private final AtomicBoolean isExecutionClientAvailable = new AtomicBoolean(true);
```
lucassaldanha, line 21: This was a silly mistake... After I updated the code on `Web3JClient` to avoid sending events for every successful request, I didn't notice that the execution would never go through the "recover" path and publish the event that EL is available. So if your communication with the EL never fails and recovers, it will never register as available.

ajsutton (approved):
LGTM.

PR #6332, opened 2022-10-20
Include information about why publishing the attestation failed in response

ajsutton:
## PR Description
Provide information about why publishing the attestation failed in the error response instead of just a generic description that it failed.  Since we shouldn't get exceptions (should just return a rejected result), it also logs it as an error on the beacon node.

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/build.gradle
+++ b/beacon/validator/build.gradle
@@ -25,6 +25,7 @@ dependencies {
   implementation project(':data:serializer')
   implementation project(':data:provider')
   implementation project(':infrastructure:events')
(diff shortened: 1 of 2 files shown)


tbenr (approved):
LGTM with suggestion

ajsutton:
er, there was no suggestion... Maybe a comment got lost somewhere?

tbenr:
> er, there was no suggestion... Maybe a comment got lost somewhere?

Don't know where it is gone. I was suggesting to use `ExceptionUtil::getMessageOrSimpleName` to avoid potential null messages

PR #6334, opened 2022-10-21
Using json-based definitions for API methods by default

lucassaldanha:
## PR Description

Updates default value of `--Xrest-api-migrated-enabled` to `true`.

This change should not require any changes for the users and should be completely transparent. To revert to the previous behaviour run your node with `--Xrest-api-migrated-enabled=false`.

This toggle chooses between `JsonTypeDefinitionBeaconRestApi` (new default) or `ReflectionBasedBeaconRestApi` (previous default). In the future, the `--Xrest-api-migrated-enabled` option will be deprecated in favour of always using the updated `JsonTypeDefinitionBeaconRestApi`.


ajsutton (approved):
LGTM but we should add this to the changelog.

PR #6340, opened 2022-10-26
use only contexts

joshuafernandes:
## PR Description
- Use only circleCI contexts for creds so we can enforce seamless rotation of credentials

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,10 +6,9 @@ executors:
   small_executor:
     docker:
       - image: cimg/openjdk:11.0.13
-        auth: &docker-auth
-          # Don't panic, throw away account to avoid Docker rate limits when downloading.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -532,57 +536,77 @@ workflows:
...
+          context:
+            - dockerhub-quorumengineering-ro              
       - acceptanceTests:
           requires:
             - assemble
           filters:
             tags:
               <<: *filters-release-tags
+          context:
+            - dockerhub-quorumengineering-ro
```
ajsutton, line 601: Is this context actually available to PRs from forks? Previously doing it this way has broken the ability for CI to build PRs.

joshuafernandes: Quite right, I'll fix this up in w3s and just add a comment here re the above

PR #6347, opened 2022-10-28
Fix event stream failover when primary node is syncing

### LINKED ISSUE #6345: Sync committee messages not created when failing over to secondary beacon node
### Description
When a vc is configured with multiple beacon node endpoints and the primary beacon node is online but syncing, the vc fails to produce sync committee messages when it should have been able to via the fallback:

```
{"@timestamp":"2022-10-26T20:15:16,005","level":"WARN","thread":"ValidatorTimingChannel-1","class":"teku-validator-log","message":"Validator   *** Skipped producing sync_signature while node is syncing  Count: 1, Slot: 4192276","throwable":""}
```

It is producing attestations and aggregates correctly.

StefanBratanov:
## PR Description
Removing additional check `eventSource.getState().equals(ReadyState.OPEN)` since the state could be OPEN even if the node is syncing.

## Fixed Issue(s)
fixes #6345 

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #5680, opened 2022-06-02
Implement Capella upgrade


benjaminion:
[Specification](https://github.com/ethereum/consensus-specs/tree/dev/specs/capella).

Primarily just concerns withdrawals for now. But EIP-4844 is likely to be added in, see #5681.

### Spec Config

 - [x] https://github.com/ConsenSys/teku/issues/6315
 - [x] https://github.com/ConsenSys/teku/issues/6316
 - [x] https://github.com/ConsenSys/teku/issues/6317
 - [x] https://github.com/ConsenSys/teku/issues/6318

### Spec Logic

 - [x] https://github.com/ConsenSys/teku/issues/6319
 - [x] https://github.com/ConsenSys/teku/issues/6320
 - [x] https://github.com/ConsenSys/teku/issues/6321
 - [x] https://github.com/ConsenSys/teku/issues/6322

### P2P Changes
 - [x] https://github.com/ConsenSys/teku/issues/6358

### Rest API Changes
 - [x] https://github.com/ConsenSys/teku/issues/6296

### Testing
TBC

PR #6348, opened 2022-10-30
Move bellatrix configuration to be comparable to altair

rolfyone:
In order to define Capella configuration, I need to move Bellatrix config into the same kind of structure as Altair, so that I can build on it from there.

Partially addresses #5680


ajsutton (approved):
LGTM.

PR #6349, opened 2022-10-30
Added capella configuration files

rolfyone:
Used mainnet file as the basis for gnosis, and minimal as the basis for swift.

partially addresses #5680


ajsutton (approved):
LGTM.

PR #6350, opened 2022-10-31
Start building out capella logic

rolfyone:
- capella blocks have a todo against them

 - capella state builder is using bellatrix currently


ajsutton:
Pushed a fix for equals which gets the unit tests passing.  The property tests are failing because we're requiring the new capella variables in all networks (prater doesn't have them yet) - need to handle them not being present for backwards compatibility.  So basically in `SpecConfigBuilder.CapellaBuilder` we shouldn't have:

```
    public void validate() {
      validateConstant("capellaForkVersion", capellaForkVersion);
      validateConstant("capellaForkEpoch", capellaForkEpoch);
    }
```
and should just provide default values for those fields (forkEpoch is just far future epoch and fork version then won't matter I don't think).

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/GetNewBlindedBlock.java
```diff
@@ -167,19 +173,4 @@ private static EndpointMetadata getEndpointMetaData(
                 block -> spec.getForkSchedule().getSpecMilestoneAtSlot(block.getSlot())))
         .build();
   }
-
-  private static SerializableOneOfTypeDefinition<BeaconBlock> getBlindedBlockSchemaDefinition(
```
rolfyone, line 171: This cleanup basically allows us to filter the capella milestone when listing types of blinded blocks... doing our own handling here was going to mean a lot of duplication of filters.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/validator/GetNewBlock.java
```diff
@@ -159,20 +167,4 @@ public void handleRequest(RestApiRequest request) throws JsonProcessingException
                     .map(AsyncApiResponse::respondOk)
                     .orElseThrow(ChainDataUnavailableException::new)));
   }
-
-  private static SerializableOneOfTypeDefinition<BeaconBlock> getBlockSchemaDefinition(
```
rolfyone, line 163: This cleanup basically allows us to filter the capella milestone when listing types of blinded blocks... doing our own handling here was going to mean a lot of duplication of filters.

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/schema/ProduceBlockV2Response.json
```diff
@@ -16,7 +16,7 @@
     },
     "version" : {
       "type" : "string",
-      "enum" : [ "phase0", "altair", "bellatrix" ]
+      "enum" : [ "phase0", "altair", "bellatrix", "capella" ]
```
ajsutton: We probably don't want capella appearing here yet?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
```diff
@@ -32,6 +33,8 @@
 
 public class MilestoneDependentTypesUtil {
 
+  private static final Set<SpecMilestone> IGNORED_MILESTONES = Set.of(SpecMilestone.CAPELLA);
```
ajsutton, line 36: We should make sure we log an issue to unignore this when we're ready. Otherwise we'll forget to make it appear in the api spec for sure...

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/ropsten.yaml
```diff
@@ -27,8 +27,11 @@ TERMINAL_TOTAL_DIFFICULTY: 50000000000000000
 TERMINAL_BLOCK_HASH: 0x0000000000000000000000000000000000000000000000000000000000000000
 TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH: 18446744073709551615
 
+# Capella
+CAPELLA_FORK_VERSION: 0x03000071
```
rolfyone: there's not really a value that I can use here?

ajsutton: Shouldn't need to specify anything.  None of the network yaml files should change at this point - only when we actually need to run capella on them. Otherwise we'd break backwards compatibility.

ajsutton (commented):
Looks good generally, mostly just need to make sure we don't break backwards compatibility with existing network definitions.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
rolfyone: fair point #6355

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/migrated/schema/ProduceBlockV2Response.json
rolfyone: I wasn't too worried about this, but I can take a stab at excluding it.. it's just more complicated with the way we do enumerations

ajsutton (approved):
LGTM.

PR #6352, opened 2022-10-31
Logging nits

StefanBratanov:
## PR Description

- Change builder availability logging message. `is offline\is online` could be misleading
- Changed `Execution Client request succeeded after a previous failure` to `Execution Client is responding to requests again after a previous failure`

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
@@ -152,21 +152,17 @@ public class EventLogger {
         Color.YELLOW);
   }
 
-  public void executionClientRecovered() {
-    info("Execution Client request succeeded after a previous failure", Color.GREEN);
-  }
-
-  public void builderIsOffline(final String errorMessage) {
-    String builderOfflineEventLog =
+  public void builderIsNotAvailable(final String errorMessage) {
+    final String builderOfflineEventLog =
         String.format(
-            "The builder is offline: %s. Block production will fallback to the execution engine.",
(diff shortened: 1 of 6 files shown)


tbenr:
I'm OK with the first point. Not sure about the second. I see some usefulness in it, maybe we could find a message that sound better to you. @benjaminion do you have an opinion\suggestion here?

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
```diff
@@ -152,22 +152,18 @@ public void executionClientRequestTimedOut() {
...
+    warn(builderIsNotAvailableEventLog, Color.YELLOW);
   }
 
-  public void builderIsBackOnline() {
-    String builderOnlineEventLog =
-        "The builder is back online. It will be used for block production.";
-    info(builderOnlineEventLog, Color.GREEN);
+  public void builderIsAvailable() {
+    final String builderIsAvailableEventLog =
+        "The builder is available. It will be used for block production.";
```
rolfyone, line 169: should probably call out these messages have been changed - if anyone is monitoring for them they'll be missing in their old format...

ajsutton:
Yeah I think we need a message to indicate that the execution client is responding to requests again otherwise there's no way to know when you've fixed the issue.

StefanBratanov:
> Yeah I think we need a message to indicate that the execution client is responding to requests again otherwise there's no way to know when you've fixed the issue.

Yeah agree. Modified the message a bit. Think it makes more sense now.

ajsutton (approved):
LGTM.

PR #6356, opened 2022-11-01
Introduce a SpecSupplier for property tests

ajsutton:
## PR Description
Ignores invalid configurations such as networks that don't yet have config for capella combined with enabling capella. Also removes some duplication which is nice.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6367, opened 2022-11-03
Added rest-api endpoint to fetch randao given a state.

### LINKED ISSUE #6295: add randao endpoint
https://github.com/ethereum/beacon-APIs/pull/242

rolfyone:
- Fixed a spec function that was returning randao mix for an epoch given a state

fixes #6295

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRandao.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRandao.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright ConsenSys Software Inc., 2022
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.beaconrestapi.handlers.v1.beacon;
+
+import static tech.pegasys.teku.beaconrestapi.BeaconRestApiTypes.EPOCH_PARAMETER;
+import static tech.pegasys.teku.beaconrestapi.BeaconRestApiTypes.PARAMETER_STATE_ID;
+import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_BAD_REQUEST;
+import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_OK;
+import static tech.pegasys.teku.infrastructure.http.RestApiConstants.TAG_BEACON;
+import static tech.pegasys.teku.infrastructure.json.types.CoreTypes.BYTES32_TYPE;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.Optional;
+import java.util.function.Function;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.api.ChainDataProvider;
+import tech.pegasys.teku.api.DataProvider;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.json.types.SerializableTypeDefinition;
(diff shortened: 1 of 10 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRandao.java
```diff
@@ -0,0 +1,107 @@
...
+            .description("RANDAO mix for state with given 'stateId'.")
+            .withField("data", randaoType, Function.identity())
+            .build();
+    return EndpointMetadata.get(ROUTE)
+        .operationId("getStateRandao")
+        .summary("Get chain genesis details")
+        .description(
+            "Fetch the RANDAO mix for the requested epoch from the state identified by `state_id`.\n\n"
+                + "If an epoch is not specified then the RANDAO mix for the state's current epoch will be returned.\n\n"
+                + "By adjusting the `state_id` parameter you can query for any historic value of the RANDAO mix. "
```
lucassaldanha, line 71: > By adjusting the `state_id` parameter you can query for any historic value of the RANDAO mix.

Do we want to clarify that it might not be available if the node does not have historical states?

rolfyone: probably not, we'd have to say that for basically every `/beacon/states` endpoint...

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -308,6 +308,32 @@ public SafeFuture<Optional<ObjectAndMetaData<List<StateValidatorData>>>> getStat
...
+      return Optional.empty();
+    } else if (epoch.isLessThan(stateEpoch)
+        && stateEpoch.minusMinZero(epochsPerHistoricalVector).isGreaterThan(0)
+        && stateEpoch.minusMinZero(epochsPerHistoricalVector).isGreaterThanOrEqualTo(epoch)) {
+      // ignoring the first period of epoch=0 to epochsPerHistoricalVector,
+      // return empty if the epoch is not within `epochsPerHistoricalVector` of the state epoch
+      return Optional.empty();
+    }
+    return Optional.of(
+        spec.atSlot(state.getSlot()).beaconStateAccessors().getRandaoMix(state, epoch));
```
lucassaldanha, line 334: On https://github.com/ethereum/beacon-APIs/pull/242 part of the reason for having this method seems to be related to not loading the entire state for an epoch just to get the RANDAO. However, what we are doing here seems to be just hiding the fact that we are loading the state to get the RANDAO.

This is better than returning the whole state over the wire, but if most of the time spent in this operation is in the actual loading of the state, should we try to address this?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRandao.java
rolfyone: this is basically the description straight from the standard api...

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
rolfyone: we're separating business logic from interface basically, just how our rest api's are structured, we can't really do anything but load the state...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRandao.java
lucassaldanha: Maybe on the api doc we can mention that 404 might also mean that the BN does not have the historical state they are looking for?

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
rolfyone: its definitely a lot less effort over the wire to xmit 1 bytes32 payload than the entire state... i think that's the argument for the endpoint...

lucassaldanha (approved):
LGTM. Just some questions for my own learning :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetStateRandao.java
rolfyone: We could do, but generally 404 is the route was not found, so i'm not really fussed to change the 404 response to have a custom message unless someone feel strongly about it...

ajsutton (approved):
LGTM.

PR #6373, opened 2022-11-04
Update to Javalin 5

### LINKED ISSUE #6280: Migrate to Javalin 5.x.x
Javalin 5 is out, we should consider migrating to it

https://javalin.io/migration-guide-javalin-4-to-5

ajsutton:
## PR Description
Update to Javalin 5. Doesn't appear to be a way to disable the warning about not having called start within 5 seconds anymore so will have to follow up on that.

## Fixed Issue(s)
fixes #6280 

## Documentation

## Changelog


ajsutton:
Looks like the event stream isn't working right...

ajsutton:
EventStream fixed, but I'm suspiciously not getting any log messages from Javalin. Seems like slf4j config isn't working right.

ajsutton:
slf4j fixed in a separate PR and annoying log messages from Javalin suppressed.

rolfyone (approved):
LGTM

lucassaldanha (approved):
LGTM

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/ThymeleafConfigurator.java
```diff
@@ -13,24 +13,31 @@
...
 import org.thymeleaf.templateresolver.ITemplateResolver;
 
-/** ThymeLeaf templates configuration for Javalin */
-public class ThymeleafConfigurator {
+/**
+ * ThymeLeaf templates configuration for Javalin. Note that Javalin renderers are configured using a
+ * singleton so only the first call to this class will have any effect.
+ */
+class ThymeleafConfigurator {
+  private static final AtomicBoolean REGISTERED = new AtomicBoolean();
```
zilm13: Could we have `AtomicBoolean(false)` here, please. It may be obvious with boolean (not for me), but it looks like  atomics allows even `new AtomicInteger()`, which is the same as `new AtomicInteger(0)` and it sounds weird going that far

ajsutton: Yeah good point, I like making the default values explicit too.

PR #6375, opened 2022-11-04
Fix slf4j logging

ajsutton:
## PR Description
Includes the log4j2 adapater for slf4j 2.0 apis. Keeps 1.0 slf4j compatibility to be sure it works too.

## Documentation

## Changelog


ajsutton:
It turns out that logCaptor requires using its own slf4j backend and when it activates it sends log4j content through to logback which was causing failures.  So changed approach to only include the slf4j2 log4j backend for production and not tests. I'm not sure if we're getting slf4j output in tests or not but it's definitely no worse and the production logging is better.

lucassaldanha (approved):

PR #6376, opened 2022-11-04
Update Capella beacon state to include latest_withdrawal_validator_index

rolfyone:
- removed withdrawal_queue_limit and max_partial_withdrawals_per_epoch from the configuration, as they're not needed in the latest spec updates

 - BeaconState can now be created and copied through accessors.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigCapella.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigCapella.java
@@ -36,7 +36,7 @@ public interface SpecConfigCapella extends SpecConfigBellatrix {
   @Override
   Optional<SpecConfigCapella> toVersionCapella();
 
-  UInt64 getMaxWithdrawalsPerPayload();
+  int getMaxWithdrawalsPerPayload();
 
-  UInt64 getMaxBlsToExecutionChanges();
+  int getMaxBlsToExecutionChanges();
 }
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigCapella.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigCapella.java
@@ -35,4 +35,8 @@ public interface SpecConfigCapella extends SpecConfigBellatrix {
 
   @Override
   Optional<SpecConfigCapella> toVersionCapella();
+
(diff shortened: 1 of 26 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigCapella.java
```diff
@@ -35,4 +35,8 @@ static SpecConfigCapella required(SpecConfig specConfig) {
 
   @Override
   Optional<SpecConfigCapella> toVersionCapella();
+
+  UInt64 getMaxWithdrawalsPerPayload();
+
+  UInt64 getMaxBlsToExecutionChanges();
```
ajsutton: I suspect both of these are better off just being `int` given they're known small numbers. Will save a bunch of garbage and extracting ints from UInt64.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/bellatrix/BeaconStateSchemaBellatrix.java
```diff
@@ -36,7 +36,7 @@
 public class BeaconStateSchemaBellatrix
     extends AbstractBeaconStateSchema<BeaconStateBellatrix, MutableBeaconStateBellatrix> {
 
-  private static final int LATEST_EXECUTION_PAYLOAD_HEADER_FIELD_INDEX = 24;
+  public static final int LATEST_EXECUTION_PAYLOAD_HEADER_FIELD_INDEX = 24;
```
ajsutton: I think I'd keep this private and just define it again in Capella.  It happens to be the same but since they're completely separate SszSchemas I think it's worth the constant being separate.  I'm not strongly opinionated about it though so feel free to disagree. :)

ajsutton (approved):
LGTM.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/bellatrix/BeaconStateSchemaBellatrix.java
rolfyone: i was referencing it to get the next offset...

PR #6380, opened 2022-11-04
Update plurality for registration messages

jtraglia:
## PR Description

Small nit that I've been meaning to fix for a while.

To me, these three log sentences are a bit weird. I think the "(s)" should be on "registration" not "validator". And because each validator only has one registration, "validators" should not be plural.

```diff
- validator(s) registrations
+ validator registration(s)
```

Thoughts?

## Documentation

## Changelog


rolfyone (approved):
Looks fair to me, i probably wouldn't even have the brackets, but that's just me probably...

PR #6384, opened 2022-11-05
Stubbed implementation of GetLightClientBootstrap

### LINKED ISSUE #6287: Sync endpoints for light-client
We will definitely want minimal dependencies in a module for light client in case it needs sharing...

https://github.com/ethereum/beacon-APIs/pull/247

This would be part of the light client work, just raising a PR for sync endpoints so we have a task for them.

robzajac:
## PR Description
This PR is a base implementation of [GetLightClientBootstrap](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/bootstrap.yaml) with `return 501;`. It is intended to validate the interface.

## Fixed Issue(s)
fixes #6287

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetLightClientBootstrap.java
```diff
@@ -0,0 +1,76 @@
...
+            .withNotImplementedResponse()
+            .build());
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
+    request.respondError(501, "Not implemented");
+  }
+
+  private static SerializableTypeDefinition<ObjectAndMetaData<LightClientBootstrap>>
```
robzajac: Unsure whether to use `ObjectAndMetadata` here given that this API only requires `version` and not `executionOptimistic` or `canonical`

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetLightClientBootstrapTest.java
```diff
@@ -0,0 +1,88 @@
...
+  void setup() {
+    setHandler(new GetLightClientBootstrap(schemaDefinitionCache));
+    request.setPathParameter(
+        "block_root", "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2");
+  }
+
+  @Test
+  void metadata_shouldHandle200() throws IOException {
+    /*
+     LightClientBootstrap lightClientBootstrap =
```
robzajac: Need a proper mock response here. Not sure how to construct it.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
```diff
@@ -16,6 +16,7 @@
 import static tech.pegasys.teku.ethereum.json.types.EthereumTypes.SIGNATURE_TYPE;
 import static tech.pegasys.teku.infrastructure.http.RestApiConstants.ATTESTATION_DATA_ROOT;
 import static tech.pegasys.teku.infrastructure.http.RestApiConstants.BEACON_BLOCK_ROOT;
+import static tech.pegasys.teku.infrastructure.http.RestApiConstants.BLOCK_ROOT;
```
robzajac, line 19: Not sure whether to simply reuse `BEACON_BLOCK_ROOT`, they have equivalent structure.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
```diff
@@ -41,6 +42,7 @@ public class SchemaDefinitionsAltair extends AbstractSchemaDefinitions {
   private final ContributionAndProofSchema contributionAndProofSchema;
   private final SignedContributionAndProofSchema signedContributionAndProofSchema;
   private final MetadataMessageSchemaAltair metadataMessageSchema;
+  private final LightClientBootstrapSchema lightClientBootstrapSchema;
```
robzajac, line 45: Added only for Altair schema definitions. Question whether the schema should be added for all consensus versions (i.e. lift up into `SchemaDefinitions`) (though there is no change from Altair) or whether it should only be specified in `SchemaDefinitionsAltair` and used this way.

robzajac: https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/bootstrap.yaml#L31

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
rolfyone: yep if it's the same exact structure, just re-use

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetLightClientBootstrap.java
rolfyone: ObjectAndMetadata is generally what's coming out of the ChainDataClient, so using that should be ok, but it's also just an interface you're using if we end up needing to change it so it's not locked in stone...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/lightclient/LightClientBootstrap.java
```diff
@@ -0,0 +1,70 @@
...
+import tech.pegasys.teku.infrastructure.ssz.schema.collections.SszBytes32VectorSchema;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNode;
+import tech.pegasys.teku.spec.config.SpecConfigAltair;
+import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlockHeader;
+import tech.pegasys.teku.spec.datastructures.state.SyncCommittee;
+
+public class LightClientBootstrap
+    extends Container3<LightClientBootstrap, BeaconBlockHeader, SyncCommittee, SszBytes32Vector> {
+
+  // TODO: Lift into SpecConfigAltair
```
rolfyone: Yep I think moving this will be a whole thing...

It's a little involved to create an SSZ object against the schema for the correct milestone, but it's worthwhile in case it ever changes...
Initially a good object to look at would probably be `ExecutionPayload` which exists from Bellatrix... it's not broken down into versions because there's only 1, although it may start moving very soon because there's changes marked for capella...

It might actually be worth that becoming it's own smallish PR as they're a little involved.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
rolfyone: We'll want to pull it into `SchemaDefinitions` I think, rather than hard references to Altair, as that creates the altair schema definitions even though maybe only this one is used... @ajsutton may have opinions here so tagging

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetLightClientBootstrapTest.java
rolfyone: It looks like the type that'll be expected to come into the TypeDefinition is `ObjectAndMetaData<LightClientBootstrap>`
so the expectation here is you'd be creating an `ObjectAndMetadata<LightClientBootstrap>` much like you did below...
I'll dig further into this to see what happened with the test below, because i'd have thought it'd be not too far off...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
ajsutton: Given the light client types are defined in the spec they will be versioned with hard forks and so need to come from `SchemaDefinitions`.  They will still just wind up being added to `SchemaDefinitionsAltair` because `SchemaDefinitionsBellatrix` extends from that and doesn't need to make any changes so will just inherit the Altair definitions.

Since these types are new in Altair I think you've actually got it right that they get added to `SchemaDefinitionsAltair` - they're then inherited by Bellatrix and they don't need to be in Phase0.

ajsutton: Not entirely sure that comment made sense... `SchemaDefinitions` is an interface that only contains the types that are common to all hard forks.  For things new in Altair like this you'd call `SchemaDefinitions.toVersionAltair()` to get a `SchemaDefinitionsAltair` and then can use these new methods.

And Bellatrix inherits all the Altair types, then just redefines a few, so `SchemaDefinitionsBellatrix` extends `SchemaDefinitionsAltair` so these methods will also be available there (and Capella extends Bellatrix).

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
robzajac: Nevermind -- the `BEACON_BLOCK_ROOT` parameter will be serialized with name `beacon_block_root` and the path parameter we want is named `block_root`. I saw they were both `bytes32` and they represent the same data (the beacon block root), but with different names in the API.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
robzajac: Admittedly I am still a bit confused. Since `SchemaDefinitions` contains types for all consensus versions (including phase0), I think we wouldn't lift `LightClientBootstrap` there. [The spec](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/bootstrap.yaml#L31) asks only for an `Altair.LightClientBootstrap` return type on this endpoint. The other endpoint I can see that has a similar treatment is [`produceSyncCommitteeContribution`](https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/sync_committee_contribution.yaml#L42). This makes me inclined to retrieve the `SchemaDefinitionsAltair` in the endpoint with `SchemaDefinitionsAltair.required(schemaDefinitionCache.getSchemaDefinition(SpecMilestone.ALTAIR))` and from there `.getLightClientBootstrapSchema()`

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetLightClientBootstrap.java
robzajac: It will still have all fields consumers need, and the two extra that they don't. As long as clients don't have strict response validation (i.e. allowing additive fields) then it should be okay.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetLightClientBootstrapTest.java
robzajac: It was missing the mock response `getLightClientBootstrap.json` that I wasn't sure how to construct. Resolved with @rolfyone 's commit.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/lightclient/LightClientBootstrap.java
robzajac: @rolfyone Can this one be treated like `SYNC_COMMITTEE_SIZE` from `SyncCommittee`? Another Altair-specific parameter. The field would move up to `SpecConfigAltair`, and then I believe it should be read from config, or a static constants file. But I'm having trouble seeing where the config is parsed (for e.g. `SYNC_COMMITTEE_SIZE`).

I have an [example impl on a separate branch](https://github.com/robzajac/teku/commit/30c5b6f6ea69e390e13e12163bf8696eaccb47e6) but it fails the `SpecConfigBuilder` validation, making me believe the field is not being successfully read from config.

robzajac: On second pass, I got it working (was missing a method in the `Builder`) https://github.com/robzajac/teku/commit/6b1b6a49d4d6eec19205f0f2e9641a743b8e8bc9

Just awaiting a second opinion on whether we want this in config this way, and then can write it into this branch.

robzajac: Moved constant up to config in the latest push.

robzajac:
@rolfyone @ajsutton Addressed comments & outstanding todos, do you mind taking another look?
The acceptance test failure is new. Unable to replicate locally, taking another look.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/lightclient/LightClientBootstrap.java
```diff
@@ -0,0 +1,72 @@
...
+import tech.pegasys.teku.infrastructure.ssz.schema.collections.SszBytes32VectorSchema;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNode;
+import tech.pegasys.teku.spec.config.SpecConfigAltair;
+import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlockHeader;
+import tech.pegasys.teku.spec.datastructures.state.SyncCommittee;
+
+public class LightClientBootstrap
+    extends Container3<LightClientBootstrap, BeaconBlockHeader, SyncCommittee, SszBytes32Vector> {
+
+  public static class LightClientBootstrapSchema
```
ajsutton: nit: I'd pull the schema class up into its own file rather than being nested here.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
ajsutton: Sorry should have responded to this earlier, but it looks like you worked it out and done the right thing. The new method only gets added to `SchemaDefinitionsAltair`.  We can be a bit sloppy with terminology when saying things should be added to "SchemaDefinitions" - we mean to the schema definitions for the first milestone it's relevant to.

ethereum/spec/src/test/resources/tech/pegasys/teku/spec/config/bellatrixForkSpec.json
```diff
@@ -79,6 +79,7 @@
   "SLOTS_PER_EPOCH": "32",
   "SLOTS_PER_HISTORICAL_ROOT": "8192",
   "SYNC_COMMITTEE_SIZE": "512",
+  "SYNC_COMMITTEE_BRANCH_LENGTH": "5",
```
ajsutton: Sadly we can't just update these test files (including `altairForkSpec.json` and `mergeForkSpec.json`). They're a capture of the actual response the beacon node will give from the /eth/v1/node/config endpoint which the standalone validator client uses to load the spec from the beacon node. It looks like adding this new property as a required field to Altair is breaking backwards compatibility (which is what the `SpecConfigLoaderTest.shouldLoadMainnetPreservingBackwardsCompatibilityWithRestApi` test is checking for.

We probably need to provide a default value in `SpecConfigBuilder` instead of requiring the field be set. Could just use the mainnet value as the default which I think would be reasonable. It's not particularly ideal but better than breaking backwards compatibility since people don't always update the VC and BN together. The default probably never gets used anyway since the VC doesn't use it and the built-in presets will provide it for the BN.

ajsutton (commented):
Looking really good. I think the key thing is just to provide a default so we don't break backwards compatibility with the config endpoint.

ajsutton:
Oh I think that acceptance test failure is likely related to the config backwards compatibility issue as well. It's running the latest version of the validator client (with these changes) against an old release of the beacon node and it's failing to start, likely because of the incorrect config.

To run locally you'll likely need to run `./gradlew distDocker` to build the docker images from the latest code including your changes (otherwise it likely winds up running the latest version from master without your changes).  Then you can run the acceptance test in IntelliJ like a normal test.

`./gradlew acceptanceTest` should automatically build the docker image but it runs all the tests so takes a fair while which is less useful for debugging stuff.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/lightclient/LightClientBootstrap.java
robzajac: Refactored in this push

ethereum/spec/src/test/resources/tech/pegasys/teku/spec/config/bellatrixForkSpec.json
robzajac: Thanks @ajsutton I think that makes sense. I set a default in the builder, similar treatment as `updateTimeout`. To check my understanding, is there a point at which the validators are updated to receive the new configs? Or will we continue to use the default for some time?

robzajac:
@ajsutton Recreated locally & pushed up a fix 🙏

ethereum/spec/src/test/resources/tech/pegasys/teku/spec/config/bellatrixForkSpec.json
ajsutton: The validator client will load the config from the beacon node every time it starts up, so the real value should be picked up quite quickly but it's possible the VC is running against a beacon node that doesn't get updated or against a different client which doesn't yet have light client support and this config item.  So we'll probably wind up keeping the default around pretty much forever (or at least until the next hard fork if all clients include it by then since that forces everyone to update).  For the vast majority of setups though, they'll start loading the real value very quickly.

ajsutton (approved):
LGTM. Thanks for your patience on this and sorry we've been slow with reviews.

PR #6385, opened 2022-11-07
Added BlockProcessorCapella to support Capella changes in spec logic.

rolfyone:
- Created some of the helper classes

 - removed reference to TransitionHelpers which were Bellatrix specific

## Documentation

## Changelog


ajsutton (commented):
I suspect we've gone too far here. We should just keep using the `Bellatrix` (or even earlier) versions of things unless there's actually something to change. Most of these new classes won't be needed.

I suspect now is the time we want to stop just adding infrastructure and starting letting the actual implementation drive what's required.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/forktransition/CapellaStateUpgrade.java
```diff
@@ -0,0 +1,71 @@
...
+
+              state.setFork(
+                  new Fork(
+                      preState.getFork().getCurrentVersion(),
+                      specConfig.getCapellaForkVersion(),
+                      epoch));
+
+              state.setLatestExecutionPayloadHeader(
+                  schemaDefinitions.getExecutionPayloadHeaderSchema().getDefault());
+              state.setLatestWithdrawalValidatorIndex(UInt64.valueOf(state.getValidators().size()));
```
ajsutton: This is actually initialised to 0: https://github.com/ethereum/consensus-specs/pull/3068/files#diff-c06c6116e0fa7993a0d4d7925b64cd2e5f7a23015a93e3c4d8d51209fe9f4dc0R133

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/capella/BlockProcessorCapellaTest.java
```diff
@@ -0,0 +1,44 @@
...
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+
+public class BlockProcessorCapellaTest extends BlockProcessorBellatrixTest {
+  @Override
+  protected Spec createSpec() {
+    return TestSpecFactory.createMainnetCapella();
+  }
+
+  @Test
+  void shouldIgnoreBellatrixBlock() {
```
ajsutton: nit: shouldRejectBellatrixBlock? And probably rename `shouldIgnoreAltairBlock` while we're at it.

ajsutton (approved):
LGTM, but need to get the initial value for `latest_withdrawal_validator_index` right.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/forktransition/CapellaStateUpgrade.java
rolfyone: oh so it misses. the first index, thats sad, but ok

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/capella/BlockProcessorCapellaTest.java
rolfyone: fair

PR #6394, opened 2022-11-07
Fix archive node not supporting LevelDB database version

zilm13:
## PR Description
There is a bug here and in corresponding test https://github.com/ConsenSys/teku/blob/951ce03f21fa0474f2e235aeca25129c64cc74e8/teku/src/test/java/tech/pegasys/teku/cli/options/BeaconNodeDataOptionsTest.java#L107-L114
The fix should probably look like this if the test is correct

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/options/BeaconNodeDataOptions.java
```diff
@@ -160,7 +160,7 @@ public void configure(final TekuConfiguration.Builder builder) {
 
   private DatabaseVersion parseDatabaseVersion() {
     if (createDbVersion == null) {
-      return dataStorageFrequency == 1
+      return dataStorageFrequency == 1 && DatabaseVersion.isLevelDbSupported()
           ? DatabaseVersion.LEVELDB_TREE
           : DatabaseVersion.DEFAULT_VERSION;
```
rolfyone: i wonder if archive frequency is 1 and we don't support leveldb, we should just crash? full state history of frequency of 1 will not have a very long life expectancy...

ajsutton: Yeah I suspect the current implementation is probably the right starting point. If you *really* want RocksDB with frequency 1 you can explicitly set the database format rather than using the default, but I can't see that being a good option for anyone. Even frequency=2 would halve the disk space required and you'd barely notice the different response time since you always have to load the full state with RocksDB anyway.

zilm13:
Reworked to break Teku start explicitly when archive frequency is 1 and LevelDB is not supported. Test modified to match this behavior.

ajsutton (approved):
LGTM.

PR #6397, opened 2022-11-08
Remove sync storage model

### LINKED ISSUE #6221: Update tests for async IO
As part of https://github.com/ConsenSys/teku/pull/6222 we enabled async IO by default. We need to extend the tests to work with async updates.

ajsutton:
## PR Description
Async storage has been working well for a few releases so remove the feature toggle and support for the sync storage approach.

## Fixed Issue(s)
fixes #6221 

## Documentation

## Changelog


lucassaldanha (approved):
LGTM. Should we add a note on the changelog about this?

ajsutton:
Shouldn't need a changelog entry as it's only a `--X` option being removed so something we can remove at any time. And it doesn't actually change the behaviour for anyone not using the --X option (which shouldn't be anyone).

lucassaldanha:
> Shouldn't need a changelog entry as it's only a `--X` option being removed so something we can remove at any time. And it doesn't actually change the behaviour for anyone not using the --X option (which shouldn't be anyone).

I forgot we made async storage the default option before :)

PR #6399, opened 2022-11-08
Remove --Xp2p-batch-verify-attestation-signatures-enabled

ajsutton:
## PR Description
It has been enabled for a very long time. Batching can still be disabled by setting the max batch size to 1 but we'll now always use the same infrastructure to simplify things.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6404, opened 2022-11-09
Add capella operation to perform BlsToExecutionChange

rolfyone:
## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -28,6 +47,9 @@
 import tech.pegasys.teku.spec.schemas.SchemaDefinitionsBellatrix;
 
 public class BlockProcessorCapella extends BlockProcessorBellatrix {
+  private static final Bytes ETH1_WITHDRAWAL_PREFIX = Bytes.fromHexString("0x01");
+  private static final Bytes WITHDRAWAL_BUFFER = Bytes.repeat((byte) 0x00, 11);
```
rolfyone: this will likely move to capella helpers once it exists

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -55,4 +77,76 @@ public BlockProcessorCapella(
...
+              validator.withWithdrawalCredentials(
+                  withdrawalAddressFromEth1Address(addressChange.getToExecutionAddress())));
+    }
+  }
+
+  public static Bytes32 withdrawalAddressFromEth1Address(final Bytes20 toExecutionAddress) {
+    return Bytes32.wrap(
+        Bytes.concatenate(
+            ETH1_WITHDRAWAL_PREFIX, WITHDRAWAL_BUFFER, toExecutionAddress.getWrappedBytes()));
+  }
```
rolfyone: this will likely exist in helpers also rather than here, currently static for a test case.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateAccessors.java
```diff
@@ -90,6 +91,22 @@ public Optional<BLSPublicKey> getValidatorPubKey(BeaconState state, UInt64 valid
                 }));
   }
 
+  public boolean validateWithdrawalCredentials(
```
ajsutton: I'd have modelled this off of `verifyVoluntaryExits` and return a `BlockValidationResult` so that we can include a reason for it being invalid.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateAccessors.java
```diff
@@ -90,6 +91,22 @@ public Optional<BLSPublicKey> getValidatorPubKey(BeaconState state, UInt64 valid
...
+      BeaconState state, BlsToExecutionChange addressChange) {
+    final int validatorIndex = addressChange.getValidatorIndex().intValue();
+    if (state.getValidators().size() <= validatorIndex || validatorIndex < 0) {
```
ajsutton: A UInt64 can't be  `< 0` and `intValue` checks it hasn't overflowed so we can skip the `validatorIndex < 0` check here.

Also `validatorIndex` must be strictly less than `validators.size()` since the indices start from 0.

The spec has it right: `assert address_change.validator_index < len(state.validators)`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateAccessors.java
```diff
@@ -90,6 +91,22 @@ public Optional<BLSPublicKey> getValidatorPubKey(BeaconState state, UInt64 valid
...
+      return false;
+    }
+    final Bytes32 withdrawalCredentials =
+        state.getValidators().get(validatorIndex).getWithdrawalCredentials();
+    if (withdrawalCredentials.get(0) != config.getBlsWithdrawalPrefix().get(0)
+        || withdrawalCredentials.slice(1, 31)
+            != Hash.sha256(addressChange.getFromBlsPubkey().toBytesCompressed().slice(1))) {
```
ajsutton: I think the order of operations is wrong here.  It should be:
```suggestion
            != Hash.sha256(addressChange.getFromBlsPubkey().toBytesCompressed()).slice(1)) {
```

ie sha256 the full pub key, then remove the first byte of the hash. Currently we're getting the sha256 of the truncated public key.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateAccessors.java
```diff
@@ -90,6 +91,22 @@ public Optional<BLSPublicKey> getValidatorPubKey(BeaconState state, UInt64 valid
...
+  public boolean validateWithdrawalCredentials(
+      BeaconState state, BlsToExecutionChange addressChange) {
+    final int validatorIndex = addressChange.getValidatorIndex().intValue();
+    if (state.getValidators().size() <= validatorIndex || validatorIndex < 0) {
+      return false;
+    }
+    final Bytes32 withdrawalCredentials =
+        state.getValidators().get(validatorIndex).getWithdrawalCredentials();
+    if (withdrawalCredentials.get(0) != config.getBlsWithdrawalPrefix().get(0)
+        || withdrawalCredentials.slice(1, 31)
```
ajsutton: Should be able to just use `.slice(1)` here to match what we're doing to the sha256.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateAccessors.java
ajsutton: Also, why is this here in `BeaconStateAccessors`? I'd have expected this to just be part of `BlockProcessorCapella` since it's actually part of the new `process_bls_to_execution_change` method.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/operations/OperationSignatureVerifier.java
```diff
@@ -107,4 +109,29 @@ public boolean verifyVoluntaryExitSignature(
...
+      final Fork fork,
+      final BeaconState state,
+      final SignedBlsToExecutionChange signedBlsToExecutionChange,
+      final BLSSignatureVerifier signatureVerifier) {
+    final BlsToExecutionChange addressChange = signedBlsToExecutionChange.getMessage();
+    final int validatorIndex = addressChange.getValidatorIndex().intValue();
+    if (state.getValidators().size() <= validatorIndex || validatorIndex < 0) {
+      return false;
+    }
+    if (beaconStateAccessors.validateWithdrawalCredentials(state, addressChange)) {
```
ajsutton: `validateWithdrawalCredentials` is duplicating these validatorIndex checks. But in the signature verifier I don't think we should be calling validateWithdrawalCredentials at all. At which point I don't think we need to get the validator at all so no need to check the index here.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
ajsutton: I'd have this pre-concatenated with the `ETH1_WITHDRAWAL_PREFIX` to save processing when we use it.  Buffer isn't the right name for it either since it's not actually a Buffer.  Maybe just `ETH1_WITHDRAWAL_PREFIX_AND_PADDING`?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -55,4 +77,76 @@ public BlockProcessorCapella(
...
+
+  private BlockValidationResult verifyBlsToExecutionChanges(
+      final BeaconState genericState,
+      final SszList<SignedBlsToExecutionChange> signedBlsToExecutionChanges,
+      final BLSSignatureVerifier signatureVerifier) {
+    for (SignedBlsToExecutionChange signedBlsToExecutionChange : signedBlsToExecutionChanges) {
+      boolean signatureValid =
+          operationSignatureVerifier.verifyBlsToExecutionChangeSignature(
+              genericState.getFork(), genericState, signedBlsToExecutionChange, signatureVerifier);
```
ajsutton, line 167: This is where the code to check the current withdrawal credentials should live I think.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -55,4 +77,76 @@ public BlockProcessorCapella(
...
+      if (!signatureValid) {
+        return BlockValidationResult.failed(
+            "BlsToExecutionChange signature is invalid: " + signedBlsToExecutionChange);
```
ajsutton, line 170: Worth noting that this failure reason is currently incorrect - it's being used for all the validation failures that could happen, not just invalid signature.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -55,4 +77,76 @@ public BlockProcessorCapella(
         operationValidator,
         schemaDefinitions);
   }
+
+  @Override
+  @CheckReturnValue
+  protected BlockValidationResult validateBlock(
```
ajsutton: I think we actually want to override `verifyBlockSignatures` and add in the `verifyBlsToExecutionChanges` at the end of that. Otherwise we're checking the state root matches before verifying the bus execution changes and calculating the state root is a much more expensive validation to perform.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateAccessors.java
rolfyone: pretty sure i copied that... apparently from the wrong place! Will rectify

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
rolfyone: i'll just concatenate it into a const, and won't need that value separately at all

lucassaldanha: We have this value available on `MiscHelpersCapella`.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
lucassaldanha: Yeah maybe move this into `MiscHelpersCapella` or some sort of utility class.

rolfyone:
Thanks @ajsutton i've addressed your feedback.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -55,4 +83,93 @@ public BlockProcessorCapella(
...
+    for (SignedBlsToExecutionChange signedBlsToExecutionChange : signedBlsToExecutionChanges) {
+      final BlsToExecutionChange addressChange = signedBlsToExecutionChange.getMessage();
+      final int validatorIndex = addressChange.getValidatorIndex().intValue();
+      if (genericState.getValidators().size() < validatorIndex) {
+        return BlockValidationResult.failed("Validator index invalid: " + validatorIndex);
+      }
+      final Bytes32 withdrawalCredentials =
+          genericState.getValidators().get(validatorIndex).getWithdrawalCredentials();
+      if (withdrawalCredentials.get(0) != WithdrawalPrefixes.BLS_WITHDRAWAL_PREFIX.get(0)) {
+        return BlockValidationResult.failed("Validator is not a bls key: " + validatorIndex);
```
ajsutton:
```suggestion
        return BlockValidationResult.failed("Not using BLS withdrawal credentials for validator " + validatorIndex);
```

ajsutton (approved):
LGTM. Just one small nit.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
rolfyone: aagh the trap of 'commit suggestion!

PR #6405, opened 2022-11-09
Avoid creating multiple subscriptions when gossip is stopped and started

ajsutton:
## PR Description
When gossip subscriptions were started, stopped and then started again we wound up with some subscriptions being duplicated because we didn't unsubscribe when stopping gossip.

For things we can't unsubscribe from, we now ensure we only subscribe once at first startup even if gossip isn't started. They aren't things we expect events from when we're not close to in sync anyway and any events would either be handled fine or ignored (which we've been depending on when starting sync again anyway).

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/ActiveEth2P2PNetwork.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/ActiveEth2P2PNetwork.java
@@ -122,6 +122,8 @@ public class ActiveEth2P2PNetwork extends DelegatingP2PNetwork<Eth2Peer> impleme
 
   private synchronized void startup() {
     state.set(State.RUNNING);
+    processedAttestationSubscriptionProvider.subscribe(gossipForkManager::publishAttestation);
+    eventChannels.subscribe(BlockGossipChannel.class, gossipForkManager::publishBlock);
     if (isCloseToInSync()) {
       startGossip();
     }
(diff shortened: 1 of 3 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/ActiveEth2P2PNetwork.java
```diff
@@ -142,12 +144,20 @@ private synchronized void startGossip() {
...
     gossipForkManager.configureGossipForEpoch(recentChainData.getCurrentEpoch().orElseThrow());
-    processedAttestationSubscriptionProvider.subscribe(gossipForkManager::publishAttestation);
-    eventChannels.subscribe(BlockGossipChannel.class, gossipForkManager::publishBlock);
 
     setTopicScoringParams();
   }
 
+  private synchronized void stopGossip() {
+    if (gossipStarted.compareAndSet(true, false)) {
+      LOG.warn("Stopping eth2 gossip while node is syncing");
```
rolfyone: warn on the basis we can't perform operations?

rolfyone (approved):
LGTM, just a question

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/ActiveEth2P2PNetwork.java
ajsutton: Warn because we always have. :) Mostly it's just useful to know the node is no longer following gossip. There's probably an argument it could be info but it is an indication that something isn't ideal (the node is significantly out of sync when it was previously in sync).

ajsutton:
This was merged but GitHub seems to have lost the plot a bit...

PR #6413, opened 2022-11-09
Fix validator index check in verifyBlsToExecutionChanges

jtraglia:
## PR Description

When reviewing this method, I noticed that the validator index check has a bug. It should return an error when validator index is equal to `getValidators().size()`. Assuming `SszList` is zero-indexed, which I'm pretty sure it is, that entry won't exist. I would have included a test, but since it's not a static method it's not that straight-forward.

Simple external example:

```java
int index = 3;
List<Integer> list = List.of(1, 2, 3);
if (list.size() < index) {
    System.out.println("invalid index");
} else {
    System.out.println("valid index");
    System.out.println(list.get(index));
}
```

```
valid index
Exception in thread "main" 
java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
	at java.base/java.util.ImmutableCollections$ListN.get(ImmutableCollections.java:449)
	at Main.main(Main.java:19)
```

## Documentation

## Changelog


rolfyone (approved):
LGTM

rolfyone:
Sorry, I thought I got them all but apparently not. This is still actively being worked on but it's a fair PR.

jtraglia:
No worries, I understand that it's still under development. I just like to review y'alls PRs 🙂

ajsutton:
Your reviews are very. much appreciated. :) We should make sure we have reference tests for this case too. Might start a list of things to break and check the ref tests fail once we have things done (and ref tests for this form of withdrawals).

PR #6416, opened 2022-11-10
Migrate create attestation data requests to type defs

ajsutton:
## PR Description
Move createAttestationData over to using type defs for remote validator API calls.

Tidied up how OkHttpValidatorTypeDefClient testing is done with an abstract base class so individual requests can be tested in separate test cases.
 
Added missing @JsonProperty annotations to some Capella api schema objects.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #6419, opened 2022-11-10
Use single getExecutionPayloadSchema method

ajsutton:
## PR Description
Avoids the risk of getting the Bellatrix schema definition back from Capella because the getExecutionPayloadSchemaBellatrix method was used and ensures the right schema is used for blocks. Gets ssz_static reference tests passing for everything except BeaconState (being fixed separately).

Also fixes the API schema to use a full Withdrawal object instead of serialising withdrawals to SSZ.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6421, opened 2022-11-10
Eip4844 types

### LINKED ISSUE #6378: Define new types
There are several new types defined in the spc. When these types are just alias of a generic type we use the generic one. But if these types are more complex or requires specific validation we have to create a specific class.

from [polynomial commitments types](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md#custom-types) we need to define:

- [ ] `KZGCommitment`  we need an ssz version of it like SszPublicKey (will be in `BeaconBlockBody` schema to build a list of them via `SszListSchema.create(KZGCommitment, ...)`)
- [ ] `KZGProof` we need an ssz version of it like SszPublicKey (will be part of `BlobsSidecar` schema)
- [ ] `Blob` we need an ssz version defined via SszByteVectorSchema.create() (will be part of `BlobsSidecar` schema as `SszListSchema.create(Blob, ..)`)

zilm13:
## PR Description
- I put minimal logic in `KZGCommitment` to `KZGProof` as its imlementation is another task, it looks like compressed point from the spec so I left similar to BLS method
- I'm not sure what's the correct place for `Blob` datastructure, I put it in `blocks.blob`
- for `Blob` I cannot do roundtrip test until `Milestone` defined, which requires `SpecLogic`

## Fixed Issue(s)

Fixes #6378 

## Documentation

## Changelog


tbenr (commented):
Looks good!

WRT the logic on top of the kzg objects, we don't need to have them in our code. I asked in discord and all the validations mentioned in `polynomial commitments` will be responsibility of the kzg library.

WRT Blobs placement, I'd be tempted put Blobs and BlobsSidecar under `execution\versions\eip4844`. At the end we retrieve them via engine API. WDYT?

WRT round trip, can you log it on an issue so we can track it?

zilm13:
Added https://github.com/ConsenSys/teku/issues/6422

tbenr (approved):
LGTM

PR #6423, opened 2022-11-10
Added queueless withdrawals for capella

### LINKED ISSUE #6321: Implement withdrawals processing without queues
A significant change from 1.3.0-alpha.0, the plan is now to find candidates for withdrawals during block production (or caching on state transition) rather than a large buffer populated during epoch transition.

https://github.com/ethereum/consensus-specs/pull/3068

rolfyone:
- State was missing `next_withdrawal_index`
 - changed MaxWithdrawalsPerLayload to be int
 - had to breakout process_execution_payload logic so that I could add withdrawals into that, and the spec still does the TTD check in capella which is maybe to cover corner cases, not sure...

 fixes #6321

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/bellatrix/block/BlockProcessorBellatrix.java
```diff
@@ -104,17 +104,31 @@ public void processBlock(
...
+    if (isExecutionEnabled(state, block)) {
+      final BeaconBlockBody blockBody = block.getBody();
+      processExecutionPayload(
+          state, executionPayloadHeader, blockBody.getOptionalExecutionPayload(), payloadExecutor);
+    }
+  }
+
+  public boolean isExecutionEnabled(final MutableBeaconState state, final BeaconBlock block) {
+    return miscHelpersBellatrix.isExecutionEnabled(state, block);
+  }
```
ajsutton: Seems like `isExecutionEnabled` should still be in `MiscHelpersBellatrix` (which would be inherited by the capella version).

And I'd be very tempted to just extract `processExecutionPayload` and leave the if back in the original method so it's easier to override this and just add the new processWithdrawals then call super instead of having to reimplement it.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -67,7 +84,7 @@ public BlockProcessorCapella(
       final AttestationUtil attestationUtil,
       final ValidatorsUtil validatorsUtil,
       final OperationValidator operationValidator,
-      final SchemaDefinitionsBellatrix schemaDefinitions) {
+      final SchemaDefinitions schemaDefinitions) {
```
ajsutton: This probably should be defined as `SchemaDefinitionsCapella` and we'd avoid all the required calls.  SpecLogicCapella should already have these as Capella specifically.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -80,7 +97,24 @@ public BlockProcessorCapella(
         attestationUtil,
         validatorsUtil,
         operationValidator,
-        schemaDefinitions);
+        SchemaDefinitionsBellatrix.required(schemaDefinitions));
+    schemaDefinitionsCapella = SchemaDefinitionsCapella.required(schemaDefinitions);
+    this.specConfigCapella = SpecConfigCapella.required(specConfig);
```
ajsutton: Should just require `SpecConfigCapella` be passed in rather than depend on required.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -133,12 +167,98 @@ public void processBlsToExecutionChangesNoValidation(
     }
   }
 
+  // process_withdrawals
+  public void processWithdrawals(
+      final MutableBeaconState genericState, final Optional<ExecutionPayload> maybePayload)
```
ajsutton, line 166: hmm, this is an interesting case we may want to come back to. Currently we could replay blocks with just the payload header whereas this will require the full payload (which is fine as we currently do always have the payload). At some point we probably want to split this into separate processing and verification so we don't need the full payload.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -133,12 +167,98 @@ public void processBlsToExecutionChangesNoValidation(
...
+      throws BlockProcessingException {
+    final MutableBeaconStateCapella state = MutableBeaconStateCapella.required(genericState);
+    final List<Withdrawal> expectedWithdrawals = getExpectedWithdrawals(state);
```
ajsutton: nit: I'd push this down a line so that we do the various boiler plate stuff of getting capella versions and then start on the actual logic.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -133,12 +167,98 @@ public void processBlsToExecutionChangesNoValidation(
...
+    final WithdrawalSchema withdrawalSchema = schemaDefinitionsCapella.getWithdrawalSchema();
+    final UInt64 epoch = miscHelpers.computeEpochAtSlot(preState.getSlot());
+    final int validatorCount = preState.getValidators().size();
+    UInt64 withdrawalIndex = preState.getNextWithdrawalIndex();
+    int validatorIndex = preState.getNextWithdrawalValidatorIndex().intValue();
+    for (int i = 0;
+        i < validatorCount
+            && expectedWithdrawals.size() < specConfigCapella.getMaxWithdrawalsPerPayload();
+        i++) {
+      final Validator validator = preState.getValidators().get(validatorIndex);
```
ajsutton, line 219: We're going to have to come back and optimise this. Getting each validator while searching is performance intensive - we will want to do checks that just depend on the balance first and then only get the actual validator if we really need it.  But I think that's worth logging a follow up ticket as we should do benchmarks etc and this will work fine for testnets.

ajsutton (commented):
Mostly looks good.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/bellatrix/block/BlockProcessorBellatrix.java
rolfyone: i was working around the fact we didnt have a capella helper, but I think your suggestion works as well... will give that a go...

ajsutton: oh then the capella logic would just have a `MiscHelpersBellatrix` and use that.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
rolfyone: raised #6425 to capture idea!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
rolfyone: added this comment ref. to #6426

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/bellatrix/block/BlockProcessorBellatrix.java
rolfyone: updated as discussed

ajsutton (approved):
LGTM. With this change all the ssz_static tests are passing for the latest specs version 🎉

PR #6427, opened 2022-11-11
Temporary placeholder config for Capella gossip

lucassaldanha:


ajsutton (approved):
LGTM. Makes sense as a starting point.

PR #6429, opened 2022-11-11
Perform withdrawal change verification prior to block processing

ajsutton:
## PR Description
Add a pre-validation stage to block processing rather than deferring all validation until after the block is processed. Allows separation of credential change operation validation from processing.

Use equals for comparision instead of ==

All reference tests from the latest capella spec are now passing.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6431, opened 2022-11-11
Add Capella to api types

### LINKED ISSUE #6410: Add Capella classes to the Schema interfaces
Add Capella classes under `tech.pegasys.teku.api.schema.interfaces` interfaces:

- SignedBlindedBlock
- SignedBlock
- State
- UnsignedBlindedBlock
- UnsignedBlock

ajsutton:
## PR Description
Adds capella as an option for the various block and state interfaces used in reflection based parsing. These are no longer used for OpenAPI definitions so we don't need to worry about it showing up there.

fixes #6410 

## Documentation

## Changelog


rolfyone (approved):
LGTM, now that there's no json tests on the schema this shouldn't show up anywhere really other than the validators...

PR #6433, opened 2022-11-11
Throw exception on excluded enum during deserialization

jtraglia:
## PR Description

Saw that an `excludedEnumerations` parameter was added to `DeserializableTypeDefinition.enumOf` recently. It would filter out values during serialization, but not during deserialization. I would expect it to filter in both.

* Remove a lonely semicolon that I noticed.
* Make `excludedEnumerations` final.
* Rename old test to be a little more specific.
* Skip excluded values during deserialization.
* Add new test for excluded values in deserialization.

## Documentation

## Changelog


jtraglia:
Sorry, I included a few small things in this that weren't worth their own PR.

rolfyone (approved):
LGTM

PR #6435, opened 2022-11-14
Calculate sync committee duties and subscribe at most SYNC_COMMITTEE_SUBNET_COUNT ahead

### LINKED ISSUE #6212: Unable to produce sync committee messages
### Description
Teku was unable to produce sync committee messages. This is split out this issue https://github.com/hyperledger/besu/issues/4399

From the related Besu ticket
> Ok so the key thing is that when the VC was trying to create the sync signature the beacon node had already imported a block for a later slot (the vc and bn are separate components even when run in the same process), so it gets:

> Chain head has advanced beyond slot

> It's not entirely clear why that happened but the logs show the validator client wound up catching up and producing sync messages correctly again. That may be tied up with some of the performance issues we've seen with teku and are still investigating though it's not the normal symptoms we get. We'll have to dig into in more depth to know exactly what's happened but at this stage I don't think it's a besu issue and we'll probably pick this up from the teku side to investigate further.

ajsutton:
## PR Description
Calculate sync committee duties and subscribe at most SYNC_COMMITTEE_SUBNET_COUNT ahead not EPOCHS_PER_SYNC_COMMITTEE_PERIOD ahead. Otherwise if the random number of epochs ahead we wind up picking is close to or equal to `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` a reorg or late block may cause the sync committee duties to change.

Since sync committees are very stable, the REST API doesn't provide information on dependentRoots for sync committee duties and the validator client doesn't invalidate them. So in the case where we calculate duties too early and they are affected by a reorg, the validator client will fail to perform the correct duties leading to 0% inclusion rate.

## Fixed Issue(s)
fixes #6212 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,3 +30,4 @@ For information on changes in released versions of Teku, see the [releases page]
 ### Bug Fixes
  - Fix missing status filters (active, pending, exited, withdrawal) for the `/eth/v1/beacon/states/{state_id}/validators` endpoint
  - Fixed issue which could lead to duplicate processing of some events when gossip is stopped and restarted.
+ - Fixed issue which could cause sync committee duties to be calculated too early, potentially causing duties to be missed if the scheduling was changed by a reorg.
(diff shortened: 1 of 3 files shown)


rolfyone (approved):
LGTM

PR #6437, opened 2022-11-14
Enable block timing metrics by default

### LINKED ISSUE #6300: Make block timing metrics a supported option
### Description
Should keep the `--X` version as an option but also provide a non-X prefix version.

Also worth considering how we can make the log output more user-friendly without breaking people using the existing labels for metrics (or at least the key ones like arrival and EL processing time).

ajsutton:
## PR Description
Enables block timing metrics by default and unhides the option to enable/disable it. The `--X` version is still supported as an alias because there's quite a few people using it.

## Fixed Issue(s)
fixes #6300 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Upcoming Breaking Changes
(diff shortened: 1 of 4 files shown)


mehdi-aouadi (approved):
LGTM

Issue #6422, opened 2022-11-10
EIP-4844 tests


zilm13:
### Description
As implementation of EIP-4844 is performed gradually with atomic PRs, we are not able to implement some tests because they requires future PRs. In order to not forgot it, here goes a list:
- `Blob` round-trip ssz decoding/encoding, requires `Eip4844SpecLogic`
- requires `SpecLogic`: `BlobsSidecar` ssz roundtrip, `SignedBeaconBlockAndBlobsSidecar` ssz rountrip, `ExecutionPayload` ssz rountrip,  `ExecutionPayloadHeader` ssz roundtrip, `BeaconBlockBody` ssz rountrip, `BlindedBeaconBlockBody` ssz roundtrip, general `BeaconBlockBody` and `ExecutionPayload` tests on EIP-4844 `Milestone` from #6434 
- re-enable all milestone tests that have been disabled in #6439
  - ValidatorDataProviderTest
  - tech.pegasys.teku.api.schema.SignedBeaconBlockTest
  - tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlockTest
  - EpochProcessorTest
  - OkHttpValidatorTypeDefClientTest integration test
  - SchemaDefinitionCacheTest
    - shouldGetSpecMilestoneFromSpecObject
    - subsequentCallsShouldGetTheSameSchemaObject
  - MilestoneDependentTypesUtil::IGNORED_MILESTONES
  - JsonProviderPropertyTest::SIGNED_BEACON_BLOCK_CLASS_MAP
  - JsonProviderPropertyTest::BEACON_STATE_CLASS_MAP
  - EthereumTypes::MILESTONE_TYPE
  - SpecSupplier::EXCLUDE_MILESTONES_FROM

PR #6439, opened 2022-11-14
Introduce `SpecLogicEip4844` and fix attestationWorthinessCheckerCreator

tbenr:
Introduce `SpecLogicEip4844` class and `EIP4844` milestone, with `TestSpecFactory` support.

`SpecLogicEip4844` needs to be gradually migrated to use `Eip4844` classes while we gradually introduce them.

This also make Altair's `attestationWorthinessCheckerCreator` to be reused in the subsequent forks, by moving it in  `AttestationUtil` and making it fork specific.

Fixing this I also extended the unit test `shouldCreateTheRightAttestationWorthinessChecker` to check that all forks create the expected checker (which now is checked for all milestones)

Several tests have been disabled. Tracking them on #6422.

related to: #6440

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/util/AttestationUtilTest.java
+++ b/ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/util/AttestationUtilTest.java
@@ -31,7 +31,9 @@ import org.junit.jupiter.api.TestTemplate;
 import tech.pegasys.teku.bls.BLSSignature;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.ssz.Merkleizable;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.SpecVersion;
 import tech.pegasys.teku.spec.TestSpecContext;
 import tech.pegasys.teku.spec.TestSpecInvocationContextProvider.SpecContext;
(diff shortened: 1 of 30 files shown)


ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/TestSpecFactory.java
```diff
@@ -61,6 +62,8 @@ public static Spec createMainnet(final SpecMilestone specMilestone) {
         return createMainnetBellatrix();
       case CAPELLA:
         return createMainnetCapella();
+      case EIP4844:
```
zilm13, line 65: createMinimalEip4844 too?

ajsutton (commented):
hmm, so we're already reusing `AttestationWorthinessCheckerAltair` but the construction code isn't being reused properly.

It's probably more inline with the existing pattern to move the construction code we want to reuse into a separate class - sadly it's probably going to be named something like `AttestationWorthinessCheckerFactory`...  Then we'd just create the right instance of the factory in the constructor like we do all the other logic components.

The other option would be to move `createAttestationWorthinessChecker` into something like `BeaconStateAccessors` or `AttestationUtil` so it's in an existing class we're already overriding as needed and the code would just be reused by inheritance pretty naturally.  That may be the best approach even though it's not strictly a spec method.

tbenr:
yeah, thanks

tbenr:
waiting #6441, too many things failing :)

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/util/AttestationUtilTest.java
```diff
@@ -71,12 +77,56 @@ void setUp(SpecContext specContext) {
         .thenReturn(dataStructureUtil.randomBytes(10));
     when(asyncBLSSignatureVerifier.verify(anyList(), any(Bytes.class), any(BLSSignature.class)))
         .thenReturn(SafeFuture.completedFuture(true));
-    attestationUtil =
-        new AttestationUtil(specVersion.getSchemaDefinitions(), beaconStateAccessors, miscHelpers);
+    attestationUtil = spec.getGenesisSpec().getAttestationUtil();
+
+    switch (specContext.getSpecMilestone()) {
+      case PHASE0:
```
zilm13, line 83: Maybe make some light builder and use it here and in `SpecLogic`'s to determinate which util to init?

tbenr: well... in `SpecLogic` we are currently explicitly picking the specific versions, and I'm not for adding production code that  will be used only for test. We may add it together with the other classes if we want to clean up `SpecLogic` creation..

zilm13: I mean something like `AttestationUtil.create(final Milestone milestone)`

tbenr: it should be applied to others too (like `BeaconStateAccessorsAltair`). This PR is already too big, let's avoid it for the moment.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -577,6 +581,32 @@ public ExecutionPayloadHeader randomExecutionPayloadHeaderCapella() {
             randomBytes32());
   }
 
+  public ExecutionPayloadHeader randomExecutionPayloadHeaderEip4844() {
+    final SpecConfigEip4844 specConfigCapella =
```
zilm13: `specConfigEip4844` here and below?

ajsutton (approved):
LGTM with Dmitry's suggested renaming.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
tbenr: oops

PR #6441, opened 2022-11-14
Eip4844 schema definitions

### LINKED ISSUE #6388: expose schema definitions for Eip4844 fork
Create a new `SchemaDefinitionsEip4844` class exposing all schema for the new fork

zilm13:
## PR Description
- `api.schema.eip4844` structures will be added in separate PR to reduce size of this one
- please check carefully 2 fixes in Capella files https://github.com/ConsenSys/teku/pull/6441/commits/6f685e11cae95b7128065cf9280190618bbdf05b https://github.com/ConsenSys/teku/pull/6441/commits/205e7c420d1603a7114b08dc02918b356bfe573c

## Fixed Issue(s)

Fixes #6388
## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/eip4844/BeaconStateSchemaEip4844.java
```diff
@@ -0,0 +1,119 @@
...
+  private static List<SszField> getUniqueFields(final SpecConfig specConfig) {
+    final SszField latestExecutionPayloadHeaderField =
+        new SszField(
+            LATEST_EXECUTION_PAYLOAD_HEADER_FIELD_INDEX,
+            BeaconStateFields.LATEST_EXECUTION_PAYLOAD_HEADER,
+            () -> new ExecutionPayloadHeaderSchemaEip4844(SpecConfigEip4844.required(specConfig)));
+    return Stream.concat(
+            BeaconStateSchemaAltair.getUniqueFields(specConfig).stream(),
+            Stream.of(latestExecutionPayloadHeaderField))
+        .collect(Collectors.toList());
```
ajsutton: Won't this will be missing all the new fields added in Capella? It may be worth just getting all the capella fields and then replacing the latest execution payload header field. Would probably be quite useful to have a utility method around that can do that replacement.

ajsutton (commented):
Looks good generally. I think we've missed some Capella fields in the BeaconState. Probably worth someone more familiar with the 4844 schemas to cast an eye over it as well.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/eip4844/BeaconStateSchemaEip4844.java
zilm13: Looks like I copied it from Capella and it's the same bug there. Looking for a way to cover it in a test

zilm13: I need SpecLogic for tests, postpone

zilm13: Oh, it's not bug, @tbenr helped me to figure out

tbenr: I'd prefer to introduce an override pattern like this (this is for capella). To always build on top of the previous and just apply the differences.

```java
  private static List<SszField> getUniqueFields(final SpecConfig specConfig) {
    final SszField nextWithdrawalIndexField =
        new SszField(
            NEXT_WITHDRAWAL_INDEX,
            BeaconStateFields.NEXT_WITHDRAWAL_INDEX,
            () -> SszPrimitiveSchemas.UINT64_SCHEMA);
    final SszField nextWithdrawalValidatorIndexField =
        new SszField(
            NEXT_WITHDRAWAL_VALIDATOR_INDEX,
            BeaconStateFields.NEXT_WITHDRAWAL_VALIDATOR_INDEX,
            () -> SszPrimitiveSchemas.UINT64_SCHEMA);

    final Stream<SszField> overriddenBellatrixFields =
        BeaconStateSchemaBellatrix.getUniqueFields(specConfig).stream()
            .map(
                sszField -> {
                  if (sszField.getIndex() == LATEST_EXECUTION_PAYLOAD_HEADER_FIELD_INDEX) {
                    return new SszField(
                        LATEST_EXECUTION_PAYLOAD_HEADER_FIELD_INDEX,
                        BeaconStateFields.LATEST_EXECUTION_PAYLOAD_HEADER,
                        () ->
                            new ExecutionPayloadHeaderSchemaCapella(
                                SpecConfigCapella.required(specConfig)));
                  } else {
                    return sszField;
                  }
                });
    return Stream.concat(
            overriddenBellatrixFields,
            Stream.of(nextWithdrawalIndexField, nextWithdrawalValidatorIndexField))
        .collect(Collectors.toList());
  }
```

tbenr: otherwise we could end up being forced to always build on top of Altair and add keep adding everything from there for each additional fork

tbenr: And we can filter out fields when we will start deleting stuff (if we ever will)

zilm13: What about this? needs tests and refactoring, just to show an idea https://github.com/ConsenSys/teku/pull/6441/commits/9052835b49690342efd5c0d8bb285bf08100fe81
Also weird place for this logic only in BeaconStateSchema

tbenr: decided to left it for a future PR

tbenr (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/eip4844/BeaconStateSchemaEip4844.java
ajsutton: Yeah it's worth nothing that Capella didn't inherit any of the fields from Bellatrix - the execution payload is a different schema in Capella so needed to be replaced.  4844 is the first point where we're wanting to inherit fields unchanged that aren't common to all forks.

PR #6443, opened 2022-11-15
Fix issue in storing reconstructed state

courtneyeh:
#5749 

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/V4FinalizedStateSnapshotStorageLogic.java
```diff
@@ -86,6 +86,25 @@ public void addFinalizedState(
       }
     }
 
+    @Override
+    public void addReconstructedFinalizedState( // TODO repetition
```
ajsutton: I doubt you'll be able to avoid the repetition here unfortunately as it needs to access different variables.  So I'd just remove the TODO.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/V4FinalizedStateSnapshotStorageLogic.java
```diff
@@ -86,6 +86,25 @@ public void addFinalizedState(
...
+        KvStoreAccessor db, KvStoreTransaction transaction, S schema, BeaconState state) {
+      if (!loadedLastStoreState) {
+        lastStateStoredSlot =
+            db.getFloorEntry(schema.getColumnFinalizedStatesBySlot(), state.getSlot())
+                .map(ColumnEntry::getKey);
+        loadedLastStoreState = true;
```
ajsutton: These should be using new variables that are separate to those used in `addFinalizedState`.  So we'd have `lastStateStoredSlot` and `loadedLastStoreState` used in `addFinalizedState` that tracks the last state added as we progress the chain.  Then we'd add new `lastReconstructedStateStoredSlot` and `loadedLastReconstructedStoreState` which would be used in `addReconstructedFinalizedState` to track where we're up to with reconstructing states.

We should then add a test that does something like:

```
addFinalizedState(<state at slot 20>);
addReconstructedFinalizedState(<state at slot 5>);

// assert the state at slot 5 is stored
```

ajsutton (approved):
LGTM. Nice work.

PR #6446, opened 2022-11-15
Share type definitions between beacon node and validator api client

ajsutton:
## PR Description
First stab at sharing a JSON type definition and supporting data class between the beacon REST API and the validator api client.

Put it in the `:ethereum:json-types` module which seems pretty reasonable, though it could be a module level conceptually between ethereum and beacon/validator but apart from calling it "shared" or "common" I'm not really sure how to name it which suggests it may not need to exist.

The main complication here is that we don't have an internal data structure that can hold the data required for the get genesis API response so a custom wrapper class has to be defined. In this version the wrapper class is also in `:ethereum:json-types` and the type definition is a static field on that class. It works quite neatly for this case but there's a risk that the `:ethereum:json-types` module will get used as a dumping ground for types that are actually used internally and not just in the edges of the REST API and validator API client. There is some resistance to that problem though as `:ethereum:json-types` isn't a widely used dependency so when I initially tried to use the new class more widely I found a lot of compiler errors because it wasn't available. The temptation for overreach is strong however...

## Documentation

## Changelog


rolfyone:
Looks pretty good... It'd potentially be useful for anything in validatorRequired to maybe add SSZ type while we're going... just a thought... could pretty easily add them to beacon-apis too if we wanted...

ajsutton:
I'm pretty strongly against adding SSZ for non-spec stuff without a really good reason for it. It adds a heap of overhead in terms of code and winds up generating a lot more garbage if we use the current remerkalisable stuff for such short lived objects.

ajsutton:
I moved the wrapper class into a separate package with some docs to try and dissuade people from using it as a place for actual business logic classes. Also added `Api` into the name as further discouragement of general use. 

Went back to the more normal patter of the type definition being in a `*Types` class rather than in the object itself. Otherwise it encourages people to put type definitions on our internal model classes which we don't want either.

I think this is a reasonable approach at this point and we'll see how it goes.

rolfyone:
To the best of my knowledge ValidatorRequired is pretty much all ssz objects which was what I was thinking...

Some exceptions would be syncing, config/spec, and the event steram... but Ionly really was thinking about the SSZ objects we have that could support it..

ajsutton:
hmm, possibly could review SSZ though I think it increases the scope of work when we already have quite a bit of JsonProvider to try and get rid of.

rolfyone (approved):
LGTM

PR #6447, opened 2022-11-15
Started building out calling v2 of execution layer functions in tests

rolfyone:
starting to dig into the duplication to identify the pattern better...

## Documentation

## Changelog


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/schema/PayloadAttributesV2.java
```diff
@@ -31,4 +36,34 @@ public PayloadAttributesV2(
...
+                payloadAttributes.getTimestamp(),
+                payloadAttributes.getPrevRandao(),
+                payloadAttributes.getFeeRecipient(),
+                PayloadAttributesV2.getWithdrawals(payloadAttributes.getWithdrawals())));
+  }
+
+  public static List<WithdrawalV1> getWithdrawals(
+      final Optional<List<Withdrawal>> maybeWithdrawals) {
+    if (maybeWithdrawals.isEmpty()) {
+      return List.of();
```
ajsutton: Shouldn't this should be `null` when withdrawals aren't supported not an empty list?

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
```diff
@@ -223,6 +226,24 @@ public SafeFuture<PowBlock> eth1GetPowChainHead() {
         "calling engineForkChoiceUpdated(forkChoiceState={}, payloadAttributes={})",
         forkChoiceState,
         payloadBuildingAttributes);
+    if (spec.atSlot(forkChoiceState.getHeadBlockSlot())
```
ajsutton: Minor difference but I'd probably switch based on whether capella is enabled at all, even if it hasn't activated yet. We should be able to use V2 for Bellatrix fine and it would be good to test that but without changing the current Mainnet/testnet behaviour since ELs won't yet support V2.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
```diff
@@ -260,6 +281,24 @@ && isBuilderAvailable()
...
     }
+    if (spec.atSlot(slot).getMilestone().isGreaterThanOrEqualTo(SpecMilestone.CAPELLA)) {
+      return executionEngineClient
+          .getPayloadV2(executionPayloadContext.getPayloadId())
+          .thenApply(ExecutionLayerManagerImpl::unwrapExecutionClientResponseOrThrow)
+          .thenCombine(
+              SafeFuture.of(
+                  () ->
+                      SchemaDefinitionsCapella.required(spec.atSlot(slot).getSchemaDefinitions())
+                          .getExecutionPayloadSchema()),
```
ajsutton: I think this would just be `SchemaDefinitiosnBellatrix.required` - it will then work for either Bellatrix or Capella.

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/schema/PayloadAttributesV2.java
rolfyone: mm... not sure...  soemthing to think more on...

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
rolfyone: i guess i'm up to 'if capella is in play, i'll call v2 mode' currently, but I see what you're saying...

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImpl.java
```diff
@@ -283,14 +322,30 @@ && isBuilderAvailable()
   public SafeFuture<PayloadStatus> engineNewPayload(final ExecutionPayload executionPayload) {
     LOG.trace("calling engineNewPayload(executionPayload={})", executionPayload);
 
+    if (executionPayload.getOptionalWithdrawals().isPresent()) {
```
zilm13: Maybe something like `if (executionPayload.toVersionCapella().isPresent())`?

rolfyone: this was just more to see what to refactor, its modelled differently now :)

rolfyone:
should be ready for review now.

ajsutton (approved):
LGTM.

PR #6458, opened 2022-11-17
Fixed block building for capella

rolfyone:
Currently an empty list is added for BlsToExecutionChanges, that'll ultimately come from a pool.

After this change, Capella minimal network is able to progress.

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
@@ -25,15 +25,18 @@ import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.ssz.SszList;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.SpecVersion;
 import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlock;
 import tech.pegasys.teku.spec.datastructures.blocks.Eth1Data;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlockUnblinder;
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.BeaconBlockBodyBuilder;
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.capella.BeaconBlockBodySchemaCapella;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadContext;
 import tech.pegasys.teku.spec.datastructures.operations.Attestation;
 import tech.pegasys.teku.spec.datastructures.operations.AttesterSlashing;
 import tech.pegasys.teku.spec.datastructures.operations.ProposerSlashing;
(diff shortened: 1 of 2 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -135,6 +138,14 @@ public Consumer<BeaconBlockBodyBuilder> createSelector(
...
       final SpecVersion specVersion = spec.atSlot(blockSlotState.getSlot());
 
+      if (specVersion.getMilestone().isGreaterThanOrEqualTo(SpecMilestone.CAPELLA)) {
+        final SszList<SignedBlsToExecutionChange> blsToExecutionChanges =
+            BeaconBlockBodySchemaCapella.required(
+                    specVersion.getSchemaDefinitions().getBeaconBlockBodySchema())
+                .getBlsToExecutionChangesSchema()
+                .getDefault();
+        bodyBuilder.blsToExecutionChanges(() -> blsToExecutionChanges);
+      }
```
ajsutton: We shouldn't need the if statement here - the function passed to `bodyBuilder.blsToExecutionChanges` should only be called if Capella is active, so you can just do:
```suggestion
        bodyBuilder.blsToExecutionChanges(() -> {
            return BeaconBlockBodySchemaCapella.required(
                    specVersion.getSchemaDefinitions().getBeaconBlockBodySchema())
                .getBlsToExecutionChangesSchema()
                .getDefault();
        });
```

(As usual this is guaranteed to fail spotless...)

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryTest.java
```diff
@@ -404,14 +418,26 @@ private BlockFactory createBlockFactory(final Spec spec) {
...
-    executionPayload =
-        SchemaDefinitionsBellatrix.required(spec.getGenesisSpec().getSchemaDefinitions())
-            .getExecutionPayloadSchema()
-            .getDefault();
-
-    executionPayloadHeader =
-        SchemaDefinitionsBellatrix.required(spec.getGenesisSpec().getSchemaDefinitions())
-            .getExecutionPayloadHeaderSchema()
-            .getHeaderOfDefaultPayload();
+    if (spec.isMilestoneSupported(SpecMilestone.CAPELLA)) {
```
ajsutton: Shouldn't need this bit at all because `SchemaDefinitionsBellatrix.required` should work even if the genesis spec is Capella (`SchemaDefinitionsCapella` should implement `SchemaDefinitionsBellatrix`) and it would just return the Capella execution payload schemas. If that's not working, we should fix the capella schema definitions impl.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
rolfyone: yes wasn't going to attempt to accept that change :)

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryTest.java
rolfyone: ah I had partially removed some changes like that but not this one. fixed.

ajsutton (approved):
LGTM.

PR #6461, opened 2022-11-17
Fix some whitespace concat issues

jtraglia:
## PR Description

Simply fixes some minor whitespace concatenation issues that IntelliJ pointed out.

## Documentation

## Changelog

### DIFF
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/Firewall.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/Firewall.java
@@ -58,7 +58,7 @@ public class Firewall extends ChannelInboundHandlerAdapter {
       if (cause instanceof WriteTimeoutException) {
         LOG.debug("Firewall closed channel by write timeout. No writes during " + writeTimeout);
       } else {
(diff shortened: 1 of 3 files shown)


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/Firewall.java
```diff
@@ -58,7 +58,7 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
       if (cause instanceof WriteTimeoutException) {
         LOG.debug("Firewall closed channel by write timeout. No writes during " + writeTimeout);
       } else {
-        LOG.debug("Error in Firewall, disconnecting" + cause);
+        LOG.debug("Error in Firewall, disconnecting " + cause);
```
zilm13: I'd make this one in classic way: `LOG.debug("Error in Firewall, disconnecting", cause);`

jtraglia: Good call, that is better. Thanks!

zilm13 (approved):
LGTM
Thank you!

PR #6462, opened 2022-11-17
Replace reference comparison with isNotSameAs in test

jtraglia:
## PR Description

Minor change. Use the `assertThat#isNotSameAs` method instead of doing a reference comparison. To be clear, there was nothing wrong before, this is just easier to read and probably provides a better message in the event of a problem.

## Documentation

## Changelog


zilm13 (approved):
LGTM
Thank you!

PR #6463, opened 2022-11-17
Use `.equals` for SszSchema comparisons

jtraglia:
## PR Description

Associated with this potentially new check in errorprone-checks:

* https://github.com/ConsenSys/errorprone-checks/pull/11

Where this was said:

> The `SszPrimitiveSchemas` ones are more interesting because `==` is definitely safe to use because they're all defined, slightly oddly, as anonymous inner classes and only one instance is ever going to exist for them. May need to investigate which benchmarks cover that code to see if there is actually a performance benefit.

So there isn't a problem with how things are, but it is a bit strange. I'm not really sure that I like this PR, because in my opinion the `==` comparison is better; this PR fixes the false positives from that check. Under the hood, the `.equals` method should use the `==` comparison. Is there an alternative way to "fix" these findings?

Additionally, something to keep in mind, if the schema variable is null, it could result in a `NullPointerException`. This isn't a problem in the current way things are done, with reference comparisons.

Regarding performance, I made a simple benchmark:

```java
static final SszUInt64 sszUint64 = SszUInt64.ZERO;

@Benchmark()
@BenchmarkMode(Mode.Throughput)
@Warmup(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 10, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
public void schemaComparison(Blackhole bh) {
  bh.consume(sszUint64.getSchema() == SszPrimitiveSchemas.UINT64_SCHEMA);
}
```

```
Benchmark                               Mode  Cnt    Score    Error   Units
BeaconBlockBenchmark.schemaComparison  thrpt   50  596.444 ± 14.337  ops/us
```

When changed from `==` to `.equals` its performance stayed just about the same:

```java
public void schemaComparison(Blackhole bh) {
  bh.consume(sszUint64.getSchema().equals(SszPrimitiveSchemas.UINT64_SCHEMA));
}
```

```
Benchmark                               Mode  Cnt    Score    Error   Units
BeaconBlockBenchmark.schemaComparison  thrpt   50  592.575 ± 20.353  ops/us
```

## Documentation

## Changelog

### DIFF
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/collections/SszPrimitiveListSchema.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/collections/SszPrimitiveListSchema.java
@@ -38,13 +38,13 @@ public interface SszPrimitiveListSchema<
   static <PrimT, SszPrimT extends SszPrimitive<PrimT, SszPrimT>>
       SszPrimitiveListSchema<PrimT, SszPrimT, ?> create(
           SszPrimitiveSchema<PrimT, SszPrimT> elementSchema, long maxLength, SszSchemaHints hints) {
-    if (elementSchema == SszPrimitiveSchemas.BIT_SCHEMA) {
+    if (elementSchema.equals(SszPrimitiveSchemas.BIT_SCHEMA)) {
       return (SszPrimitiveListSchema<PrimT, SszPrimT, ?>) SszBitlistSchema.create(maxLength);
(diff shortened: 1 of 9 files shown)


ajsutton (approved):
LGTM. Thanks for doing the leg work on performance. Those numbers are easily within the margin for error and I would expect inlining to eliminate the difference between the two approaches quite effectively as it seems to have done.

Definitely makes sense to me to follow standard practice of using equals here so we don't have to stop and think whether its safe all the time.

Issue #6391, opened 2022-11-07
Implement MiscHelpersEip4844


tbenr:
https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#misc

implement this helper functions. The functions will require usage of the kzg library
- [ ] process_blob_kzg_commitments
- [ ] validate_blobs_sidecar
- [ ] is_data_available
  - [ ] ~~retrieve_blobs_sidecar will be represented by a `SafeFuture<BlobsSidecar>`~~ Currently the plan is not implementing an async\optimistic blob retrival. So blobs will be provided immediately if there.
- [ ] kzg_commitment_to_versioned_hash
- [ ] tx_peek_blob_versioned_hashes
- [ ] verify_kzg_commitments_against_transactions

PR #6465, opened 2022-11-17
`BEACON_BLOCK_AND_BLOBS_SIDECAR` gossip topic

### LINKED ISSUE #6460: Implement `beacon_block_and_blobs_sidecar` topic
[reference](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#global-topics)

implements a new `GossipForkSubscriptions` which handles the new topic on `SignedBeaconBlockAndBlobsSidecar`

tbenr:
This is the first time we have to unsubscribe to a topic and subscribe to a new one.

`GossipForkSubscriptions` have been refactored to leverage class inheritance but also avoid most of the code duplication when the next fork stop using previously used topics.

Basically each fork exposes methods to add `GossipManager` for all topic that has been introduced by that fork. Than it's up to the forks to call the corresponding `add` method (or delegate to `super` if the fork is a superset of the previous).
To make sure that forks can reenable previously disabled topics, we have to make sure to pass all possible `OperationProcessor`s even if the current doesn't use it. It's the case of `blockProcessor` that may be re-enabled in future forks.

Add missing eip4844 and capella network configuration handling

fixes #6460 

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsEip4844.java
```diff
@@ -0,0 +1,104 @@
...
+          signedBlsToExecutionChangeOperationProcessor) {
+    super(
+        fork,
+        spec,
+        asyncRunner,
+        metricsSystem,
+        discoveryNetwork,
+        recentChainData,
+        gossipEncoding,
+        Optional.empty(),
```
zilm13: If I got it correctly, we should continue gossiping for 2 epochs after transition.
See: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/p2p-interface.md#transitioning-the-gossip
Discussion https://github.com/ethereum/consensus-specs/pull/3090#discussion_r1020463652

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
```diff
@@ -94,6 +96,16 @@ public SafeFuture<InternalValidationResult> validate(SignedBeaconBlock block) {
       return completedFuture(reject("Parent block is after child block."));
     }
 
+    if (blobsSidecar.isPresent()) {
+      if (blobsSidecar.get().getBeaconBlockRoot().equals(block.getRoot())) {
```
zilm13: I don't understand this. Shouldn't it be opposite? And the next one

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -217,7 +233,11 @@ private Optional<SafeFuture<BlockImportResult>> handleKnownBlock(final SignedBea
 
   private SafeFuture<BlockImportResult> handleBlockImport(
       final SignedBeaconBlock block,
+      final Optional<BlobsSidecar> blobsSidecar,
       final Optional<BlockImportPerformance> blockImportPerformance) {
+    if (blobsSidecar.isPresent()) {
+      LOG.trace("blobsSidecar is present");
```
zilm13: I'd log something about it here, not very big but with some info. If we start logging...

tbenr: Yeah was just a way to use the variable and avoid the compilation issue of unused variable

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
tbenr: Oops, I forgot a `!`!

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsEip4844.java
tbenr: Ahh yes that's a good point.. I missed that issue

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
zilm13: and btw both weights for reject from my side. 
I understand this one 
> [IGNORE] the sidecar.beacon_block_slot is for the current slot (with a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. sidecar.beacon_block_slot == block.slot.

as slots for sidecar and block matches but it arrives too later or earlier. Where are your rules in spec?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
```diff
@@ -128,6 +140,9 @@ public SafeFuture<InternalValidationResult> validate(SignedBeaconBlock block) {
               if (!blockSignatureIsValidWithRespectToProposerIndex(block, postState)) {
                 return reject("Block signature is invalid");
               }
+
+              // TODO kzg validation
```
lucassaldanha: Do we have a separate ticket for doing the KZG validation? If we don't it might be worth creating one and referring to the ticket in this comment.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
```diff
@@ -94,6 +96,16 @@ public SafeFuture<InternalValidationResult> validate(SignedBeaconBlock block) {
...
     }
 
+    if (blobsSidecar.isPresent()) {
+      if (blobsSidecar.get().getBeaconBlockRoot().equals(block.getRoot())) {
+        return completedFuture(InternalValidationResult.IGNORE);
+      }
+
+      if (blobsSidecar.get().getBeaconBlockSlot().equals(block.getSlot())) {
+        return completedFuture(InternalValidationResult.IGNORE);
+      }
```
lucassaldanha: I find it useful to encapsulate each gossip rule into a method with a meaningful name. It makes it easier to refer back to the spec.

e.g. `sidecarIsForCurrentSlot(blobsSidecar, block.getSlot())`

This means that you might also have methods representing other rules (e.g. kzs rules) even if they haven't been implemented yet empty placeholder methods.

e.g. `kzgCommitmentsAreCorrectlyEncoded(..)`, `kzgCommitmentsMatchVersionedTxHashes(..)`, etc.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsBellatrix.java
```diff
@@ -41,7 +42,7 @@ public GossipForkSubscriptionsBellatrix(
       DiscoveryNetwork<?> discoveryNetwork,
       RecentChainData recentChainData,
       GossipEncoding gossipEncoding,
-      OperationProcessor<SignedBeaconBlock> blockProcessor,
+      final Optional<OperationProcessor<SignedBeaconBlock>> blockProcessor,
```
lucassaldanha: If we are making this one final it makes sense to change the other params to final as well :)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsPhase0.java
```diff
@@ -129,17 +131,22 @@ protected void addGossipManagers(final ForkInfo forkInfo) {
...
-            recentChainData,
-            spec,
-            asyncRunner,
-            discoveryNetwork,
-            gossipEncoding,
-            forkInfo,
-            blockProcessor,
-            getMessageMaxSize());
-    addGossipManager(blockGossipManager);
+    blockProcessor.ifPresentOrElse(
```
lucassaldanha: I think I am missing something. Why do we have the optional and this logic for registering the `BlockGossipManager` here? 

Wouldn't it be the case to override `addGossipManagers(..)` on `GossipForkSubscriptionEip4844` and register it there?

lucassaldanha (commented):
I have got a few comments to consider.

In terms of testing, you could look at `BlsToExecutionChangeGossipIntegrationTest.java` to see how I was testing the gossiping part (it isn't great but kinda gets the job done).

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
tbenr: yeah, agreed..

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsPhase0.java
tbenr: My intent here was to introduce a new pattern that allows gossip managers to be removed by an upcoming fork, while save the inheritance (avoid to redefine all the subscriptions from the previous forks).
The current pattern allows only additions.

I think I'm going to change strategy:
1. this thing is ugly
2. it doesn't support deferred unsubscription (ie wait n epochs before deactivating subscription from the previous fork the current fork no longer required)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsEip4844.java
tbenr: I'm going to change the way I do this. Thinking about introducing a more structural unsubscription pattern (see comment below to luca's feedback)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
tbenr: the `MAXIMUM_GOSSIP_CLOCK_DISPARITY` check is already implemented for block, while for the other checks I started a conversation here https://github.com/ethereum/consensus-specs/issues/3109

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
tbenr: I havent yet created a dedicated ticket for the gossip validation. I'd like to see how far I go here first.
But in any case it is related to #6391 and other two kzg lib tickets

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsPhase0.java
tbenr: the 2 is wrong. And the unsubscription is nicely handled by `GossipForkManager::configureGossipForEpoch`
After some back and forth thoughts, I think I'll leave the `Optional` approach and leave potential improvement for an eventual dedicated PR.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsEip4844.java
tbenr: as we saw, everything is nicely handled by `GossipForkManager::configureGossipForEpoch`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
tbenr: moving validation to a separate PR

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsPhase0.java
```diff
@@ -219,6 +239,12 @@ public void publishBlock(final SignedBeaconBlock block) {
     blockGossipManager.publishBlock(block);
   }
 
+  @Override
+  public void publishBlockAndBlobsSidecar(
+      final SignedBeaconBlockAndBlobsSidecar blockAndBlobsSidecar) {
+    // Does not apply to this fork.
```
zilm13: I'd maybe go with interface defaults for this and all others (`publishSyncCommitteeMessage`, `publishSyncCommitteeContribution`, `publishSignedBlsToExecutionChangeMessage`, `subscribeToSyncCommitteeSubnet`, `unsubscribeFromSyncCommitteeSubnet`) to not leak them to `Phase0` implementation but definitely up to you

zilm13 (approved):
LGTM

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -807,6 +808,7 @@ protected void initP2PNetwork() {
             .eventChannels(eventChannels)
             .recentChainData(recentChainData)
             .gossipedBlockProcessor(blockManager::validateAndImportBlock)
+            .gossipedBlockAndBlobsProcessor(OperationProcessor.noop())
```
lucassaldanha, line 811: Is this correct? Don't we do anything with the BlockAndBlob messages we receive?

lucassaldanha:
Where are we forwarding the `SignedBeaconBlockAndBlobsSidecar` messages to other peers? Is this something that will come in another PR?

tbenr:
> Where are we forwarding the `SignedBeaconBlockAndBlobsSidecar` messages to other peers? Is this something that will come in another PR?

yep, decided to have a better separation compared to the first draft.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
tbenr: yes I did it on purpose. I want to modify `BlockManager` and `BlockValidator`  (and maybe `BlockImporter`) in a dedicated PR

PR #6474, opened 2022-11-18
Create MiscHelpersEip4844

### LINKED ISSUE #6391: Implement MiscHelpersEip4844

zilm13:
## PR Description
- I decided to create ssz structures from EIP-4844 (not sure it was right) to avoid magic in https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#tx_peek_blob_versioned_hashes 
  UPDATE: looked what others do while looking for fixtures, Lighthouse keeps magic, Prysm uses fork `go-ethereum` type where it's blob tx is implemented as a type, got some data to check in my implementation from their test
- Do we need `VersionedHash` or `Bytes32` is ok? 
  UPDATE: added `VersionedHash`
- Looking for some test fixtures now
  UPDATE: tests added

## Fixed Issue(s)

Fixes #6391 
## Documentation

## Changelog


tbenr:
As per out-of-band discussion, we agreed to move back from SSZ to the "magic" when decoding transaction to get the versioned hash. We gain in runtime performance and reduced maintenance (we don't need to deal with potential changes in next fork. Eventual changes in new forks will be implemented with a new miscHelper).
We lose some clarity but the "magic" is well confined in one method. We can have nice comment to increase clarity.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/SpecLogicCapella.java
```diff
@@ -86,7 +86,7 @@ public static SpecLogicCapella create(
       final SpecConfigCapella config, final SchemaDefinitionsCapella schemaDefinitions) {
     // Helpers
     final Predicates predicates = new Predicates(config);
-    final MiscHelpersBellatrix miscHelpers = new MiscHelpersBellatrix(config);
+    final MiscHelpersCapella miscHelpers = new MiscHelpersCapella(config);
```
tbenr: Seems like we normally use the previous fork class if it isn't changed. See how in `SpecLogicBellatrix` we still use Altair classes.

infrastructure/crypto/build.gradle
```diff
@@ -1,7 +1,7 @@
 dependencies {
   api 'org.bouncycastle:bcprov-jdk15on'
   api 'org.apache.tuweni:tuweni-bytes'
-
+  implementation 'commons-io:commons-io'
```
tbenr: Seems like we don't need this

tbenr (commented):
looks good. Just a couple of comments

infrastructure/crypto/build.gradle
zilm13: Forgot to remove, nice catch!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/SpecLogicCapella.java
zilm13: My point is when Capella will add their helpers (there should be some, just was not done yet) we will have SpecLogicEIP4844 without Capella parent, so when processing block for example with EIP4844 speclogic, we could fail

zilm13: I could be wrong about Capella MiscHelpers should ever happen. And my example will be definitely float in our standard tests, so could change this for sure

tbenr: I see your point. But if there will be, we'll definitely notice if we forgot to extend from the new capella version since some reference tests will fail badly.

tbenr (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/SpecLogicCapella.java
ajsutton: For the record, there are no changes to MiscHelpers in Capella.  We've actually implemented all of the Capella spec changes now. :)

PR #6476, opened 2022-11-21
Refactoring OperationValidator (used for gossip and block creating) to clarify its intent

lucassaldanha:
## PR Description
- Renamed validateFully() method to capture the fact that it is actually being used whenever a new operation message from gossip is received.
- Removed method validateForStateTransition() from OperationValidator interface as the actual per-spec state transition rules are implemented on multiple implementations of `OperationStateTransitionValidator` interface and are exposed through the Spec object.


ajsutton (approved):
LGTM.

PR #6484, opened 2022-11-21
Update reference tests to 1.3.0-alpha.1

ajsutton:
## PR Description
Also adjusts sync aggregate processing to ensure safe interactions with withdrawals in the (exceptionally unlikely) case where a fully withdrawn validator appears in the sync committee in two different positions and is penalised for the first position and rewarded for the second. The balance changes must be applied in the exact order as the spec to ensure the min zero condition when reducing the balance has the same results.

And as a bonus, the new code actually benchmarks slightly faster than the old code:

```
ProcessSyncAggregateBenchmark.processSyncAggregate             400000    ss  250  0.071 ± 0.001   s/op // Without changes
ProcessSyncAggregateBenchmark.processSyncAggregate             400000    ss  250  0.069 ± 0.001   s/op // With changes.
```

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6488, opened 2022-11-22
Check if block contains more than 1 BlsToExecutionChange for the same validator

### LINKED ISSUE #6445: Prevalidation of BLS to Eth1 Credential Changes Allows Invalid Blocks
### Description
If a block includes two credential change operations for the same validator, the block should be rejected (because the second change won't have BLS credentials by the time it's included).  With the current implementation in Teku we pre-validate the BLS changes so would incorrectly consider the block valid.

lucassaldanha:
## PR Description
- Include fix for multiple operations for same validator in block
- Updates `ManualReferenceTestRunner` with logic to check the type of error on block processing (need to revisit this approach after https://github.com/ethereum/consensus-specs/issues/3122)

## Fixed Issue(s)
fixes #6445

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/sanity/SanityBlocksTestExecutor.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/sanity/SanityBlocksTestExecutor.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
+import org.assertj.core.api.AbstractThrowableAssert;
 import tech.pegasys.teku.bls.BLSSignatureVerifier;
 import tech.pegasys.teku.ethtests.finder.TestDefinition;
 import tech.pegasys.teku.reference.BlsSetting;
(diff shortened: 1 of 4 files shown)


eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/sanity/SanityBlocksTestExecutor.java
```diff
@@ -37,6 +38,8 @@
 public class SanityBlocksTestExecutor implements TestExecutor {
 
   private static final String EXPECTED_STATE_FILENAME = "post.ssz_snappy";
+  private static final String STATE_ROOT_MISMATCH_ERROR_MESSAGE =
+      "Block state root does NOT match the calculated " + "state root";
```
rolfyone: should probably put these together given they didn't wrap...

rolfyone (approved):
LGTM just a nit

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/sanity/SanityBlocksTestExecutor.java
lucassaldanha: Argh xD

Issue #6473, opened 2022-11-18
implement `beacon_block_and_blobs_sidecar` gossip validation


tbenr:
update `BlockManager` and `BlockValidation` to handle `SignedBeaconBlockAndBlobsSidecar` coming from gossip.

Sets the stage for the changes required for state transition `ForkChoice` related classes.

PR #6490, opened 2022-11-22
4844 gossip validations and chainUpdater testFixtures

tbenr:
introduces changes in `BlockManager`, `BlockImporter` and `BlockValidator` to support new gossip validation

adds support for capella and eip4844 forks in several testFixtures classes around `chainUpdater`

related to #6473 

Doesn't close the gossip validation issue because the validations will be probably updated and kzg validation is not yet plugged in.

## Documentation

## Changelog


ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/generator/BlockProposalTestUtil.java
```diff
@@ -255,6 +275,32 @@ private ExecutionPayload createExecutionPayload(
...
+                                newSlot,
+                                UInt64.valueOf(30_000_000L),
+                                UInt64.valueOf(30_000_000L),
+                                specVersion.miscHelpers().computeTimeAtSlot(state, newSlot),
+                                dataStructureUtil.randomBytes32(),
+                                UInt256.ONE,
+                                UInt64.ONE,
+                                dataStructureUtil.randomBytes32(),
+                                transactions.orElse(Collections.emptyList()),
+                                List.of())))
```
zilm13, line 303: Don't like some inconsistency here with Capella's withdrawals, but not sure, that we should do something with it on this stage

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/validation/BlockValidatorTest.java
```diff
@@ -286,7 +296,31 @@ void shouldReturnInvalidOnWrongExecutionPayloadTimestamp(SpecContext specContext
...
+    final UInt64 nextSlot = recentChainData.getHeadSlot().plus(ONE);
+    storageSystem.chainUpdater().setCurrentSlot(nextSlot);
+
+    SignedBeaconBlock block = storageSystem.chainBuilder().generateBlockAtSlot(nextSlot).getBlock();
+
+    final BlobsSidecar blobsSidecar = specContext.getDataStructureUtil().randomBlobsSidecar();
+
+    InternalValidationResult result =
+        blockValidator.validate(block, Optional.of(blobsSidecar)).join();
     assertTrue(result.isReject());
```
zilm13: I'd separate clarification of both implemented cases (only bad slot and only bad root) and also add valid test when `blobsSidecar` exists

zilm13 (commented):
Looks good, just some thoughts

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/validation/BlockValidatorTest.java
tbenr: I agree, I was a bit lazy here because I think these consistency checks will go away.. adding missing tests

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/generator/BlockProposalTestUtil.java
tbenr: I agree, I think we should have an empty list in both cases here, but I haven't tracked where the two withdrawals are actually required (if they even are).
I'd live this to upcoming tests (from capella and Eip4844)

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/validation/BlockValidatorTest.java
zilm13: Actually atomic invalids are always covered with reference tests, but I'd definitely add a valid testcase, and it will fail once kzg checking is added, which is good

tbenr: added

ajsutton (approved):
LGTM.

PR #6494, opened 2022-11-23
Introduce GenesisStateBuilder

ajsutton:
## PR Description
Provides a more flexible way to create a genesis state rather than having multiple overloaded methods to handle all the variations.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #6496, opened 2022-11-23
Feature: default network deposit snapshot

zilm13:
## PR Description
Adds default deposit snapshot for networks where it's available (sepolia, gnosis, goerli, mainnet)
Enabled by flag `--Xdeposit-snapshot-bundle-enabled`. When used along with any network from the list, loads bundled or storage deposit tree snapshot whichever is better.
Considerations on why it's done with json internally:
- only 1kb of extra data for all networks compared to ssz, so not a big deal
- could check json manually, ssz is black box (yeah there is a tool, but json is easier)
- when EIP-4881 will be a standard, people may get snapshot from other client API and it's by default JSON, not all clients could support SSZ-encoded snapshot
- `{data:<snapshot>}`-json is acceptable for better user UX, actually parse code for `ResponseData` wrapper is copied from another two places where it's already used, maybe move it to `JSONUtils`?

Bundled it with the latest snapshots at the moment of opening a PR

## Fixed Issue(s)

Not yet
## Documentation

## Changelog


beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
```diff
@@ -63,10 +83,95 @@ private Optional<DepositTreeSnapshot> loadDepositSnapshot(
         });
   }
 
-  private DepositTreeSnapshot loadFromResource(final String source) throws IOException {
-    return DepositTreeSnapshot.fromBytes(
-        ResourceLoader.urlOrFile("application/octet-stream")
-            .loadBytes(source)
-            .orElseThrow(() -> new FileNotFoundException("Not found")));
+  private DepositTreeSnapshot loadFromUrl(final String path) throws IOException {
```
ajsutton, line 77: We've wound up making this process significantly more complex than I think it really needs to be.

Firstly we can just use `ResourceLoader.urlOrFile` to load the resource  as we do in `ChainDataLoader.loadState`. The default values pointing to the class path would be specified as a URL as like in `Eth2NetworkConfiguration.Builder.genesisStateFromClasspath`.

Then there's a lot of complexity added by supporting both JSON and SSZ formats, particularly when dealing with the extra data wrapper the REST API adds. I can see the appeal of using a readable format for the snapshots but even though I can see the content, I still can't actually verify its accuracy without running a tool over it.

So I think I'd just keep it simple and only support SSZ.

ajsutton: For context, I don't really see users manually specifying these snapshots - it's just too much hassle for them when it can easily enough be loaded from the EL anyway. It will almost always just be using the default snapshot we provide.

services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
```diff
@@ -24,11 +25,13 @@
 import tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.networks.Eth2Network;
 
 public class PowchainConfiguration {
   public static final int DEFAULT_ETH1_LOGS_MAX_BLOCK_RANGE = 10_000;
   public static final boolean DEFAULT_USE_MISSING_DEPOSIT_EVENT_LOGGING = false;
   public static final boolean DEFAULT_DEPOSIT_SNAPSHOT_STORAGE_ENABLED = false;
+  public static final boolean DEFAULT_DEPOSIT_SNAPSHOT_BUNDLE_ENABLED = false;
```
ajsutton: We can probably just combine these two flags into a single `--Xdeposit-snapshot-enabled` flag just to avoid having too many different toggles that interact in strange ways.

ajsutton (commented):
The key piece we're still missing is how to validate these snapshots - I can see the sepolia deposit root matches the latest state because it hasn't had any further deposits but the others are all very opaque.

I suspect when raising a PR that updates the snapshots we need to record the current finalized epoch when we grab them for each network - then we can use beaconcha.in to compare the info in the eth1data to what's in the snapshot which should match up. We then just need to check that the execution block hash matches up with the block number.

And we should have a test to ensure the snapshots are internally consistent (that they can be parsed and that the deposit root actual matches the `finalized` roots). Creating a `DepositTree` from the snapshot checks the finalized root so we should be able to just check we can create a DepositTree from each network snapshot correctly.

zilm13:
I've done most of the things you suggested except support of JSON,s but I dropped `data` wrapped JSON and most extra code is gone.
It was the reason, I haven't removed JSONs
> I suspect when raising a PR that updates the snapshots we need to record the current finalized epoch when we grab them for each network - then we can use beaconcha.in to compare the info in the eth1data to what's in the snapshot which should match up. 

I saw instruction on updating deposit snapshots something like this:
1. Create new branch for PR
2. Set following environment variables to appropriate urls with Teku clients (like user@8.8.8.8): SERVER_GOERLI_URL, SERVER_MAINNET_URL, SERVER_SEPOLIA_URL, SERVER_GNOSIS_URL
3. Execute updating script from the repo root:
`./scripts/update-snapshots.sh services/powchain/src/main/resources/tech/pegasys/teku/services/powchain`
4. If last update was more than 9 hours ago, all snapshots should be updated. If not all snapshots are updated, investigate on issue
5. For each network open link from script output and check following:
  - slot is not significantly older than 2 epochs (12 minutes for most networks)
  - `Eth1Data` in slot points to the same `execution_block_hash` you see in JSON for this network. Sometimes you will need to move few slots to the left (earlier) in beacon chain explorer to find the same `Eth1Data`, especially on test networks. It's a correct case, `Eth1Data` are votes and could be different
  - `deposit_count` in `Eth1Data` matches one in JSON snapshot
  - `deposit_root` in `Eth1Data` matches one in JSON snapshot
  - block hash in beacon explorer links to the block which number matches `execution_block_height` in JSON for this network
6. Run test `services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotsBundleTest.java`
7. Commit changes 
8. Include block explorer links in PR description

For https://github.com/ConsenSys/teku/pull/6496/commits/2d4f7110fb7ea669f3dadb38735b8f66629fc24c the links are:
GOERLI latest finalized slot verification link: https://goerli.beaconcha.in/slot/4400352
SEPOLIA latest finalized slot verification link: https://sepolia.beaconcha.in/slot/1131552
GNOSIS latest finalized slot verification link: https://beacon.gnosischain.com/block/6063952
MAINNET latest finalized slot verification link: https://beaconcha.in/slot/5207360

The issue is that links are pointing to the last finalized blocks, but a new voting period could be already started. There are not enough votes yet to finalize new `Eth1Data`, or say case when we don't have enough votes in period on test network, latest finalized slot will again point to non-matching `Eth1Data`. We need the slot when `DepositTreeSnapshot` is finalized, which could be deep in past, up to half of voting period or several voting periods on test networks sometimes, and it's not that simple, I have no idea how to get this slot.

ajsutton:
I suspect we're going to need tool of some kind to do the validation - it's a time consuming and quite errorprone to try and validate them by hand, particularly when all the networks are being updated at once. That would make the verification much more reliable but also mean we could just support SSZ too.  Such a tool could potentially search backwards to find a point in the chain where the deposit snapshot does match up as well. 

The other possibility for verifying things would be to start from the snapshot, then use something like Infura to load deposits up to the chain head and then you should be able to verify the deposit data in any state from finalized to chain head is correct based on your merkle tree. Infura could also confirm the execution block hash and number are right. Might be easier than having to search backwards and is probably a more real world test since that's what the system will do once it actually starts up with the snapshot.

zilm13:
I don't like both suggestions:
1) Searching backwards to the place where eth1data is available. 
We cannot do it on eth1 without replaying all deposits, so we do it on eth2. This means tool must be run in the same ssh tunnels meaning it depends on our infrastructure and is affected by its bugs. If we query blocks (several states are very big), we will need to count votes in voting period which looks already over-complicated.
2. Starting with snapshot and trying to replay deposits up to the current head. 
At first it's a big part of logic that should be detached in some way. Second, Infura already doesn't support Gnosis and could drop or add support of any network independently of us. Third, if there were no deposits after snapshot (see Sepolia) we verify nothing.

Also it's unclear how to verify with both solutions that snapshot is not outdated.

So I come to another solution: download also finalized header and state and verify snapshot, header, state against each other. And I dropped JSON support. Repository is updated. After this I see snapshot updating instructions like this:
1. Create new branch for PR
2. Set following environment variables to appropriate urls with Teku clients (like user@8.8.8.8): SERVER_GOERLI_URL, SERVER_MAINNET_URL, SERVER_SEPOLIA_URL, SERVER_GNOSIS_URL
4. Execute updating script from the repo root:
`./scripts/update-snapshots.sh services/powchain/src/main/resources/tech/pegasys/teku/services/powchain`
5. For each network open link from script output and check that slot is not significantly older than 2 epochs (12 minutes for most networks, few minutes for Gnosis)
6. Remove `@Disabled` from `DepositSnapshotsBundleTest.shouldCheckConsistencyOfSnapshotAndBlockHeaderAndState` and run test `services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotsBundleTest.java`
7. Rollback `@Disabled` and remove headers and state ssz's
8. Commit changes 

What do you think?

ajsutton:
I couldn't manage to verify the snapshots in this PR using the DepositSnapshotsBundleTest. Downloading the blocks & states referenced above don't seem to match up, I think those just aren't the exact block numbers I need to be working with.

But after a bit of a play I worked out we can verify everything from public execution client JSON-RPC endpoints relatively simply.  It does require an archive node behind the endpoint but works with Infura fine (and there's a public RPC for gnosis we can use).

The deposit contract exposes `get_deposit_count` and `get_deposit_root` methods which can be used to check the deposit count and root values. You request those values specifying the execution block hash and they should be exactly the same. Then just check that the canonical block at the block height is that block hash and you've verified all of the individual values in the snapshot match the actual execution chain.

The `shouldCreateCorrectDepositTreeSnapshotFromEachBundleSnapshot` is already checking that the finalized roots included actual do create a tree with the specified deposit root so we know they're right as well.

I'll have to spend a bit of time cleaning this up but will be able to push up something actually usable tomorrow.

ajsutton:
The code as I have it currently is:
```java
@ParameterizedTest(name = "{0}")
  @MethodSource("networksAndEth1Endpoints")
  void shouldVerifyDepositSnapshots(final Eth2Network eth2Network, final String endpoint) throws Exception {
    final PowchainConfiguration.Builder powchainConfigBuilder = PowchainConfiguration.builder();
    powchainConfigBuilder
        .specProvider(SPEC)
        .depositSnapshotEnabled(true)
        .setDepositSnapshotPathForNetwork(Optional.of(eth2Network));
    final PowchainConfiguration powchainConfiguration = powchainConfigBuilder.build();
    if (powchainConfiguration.getDepositSnapshotPath().isEmpty()) {
      return;
    }

    final String depositSnapshotPath = powchainConfiguration.getDepositSnapshotPath().get();
    final DepositSnapshotFileLoader depositSnapshotLoader =
        new DepositSnapshotFileLoader(Optional.of(depositSnapshotPath));

    // Snapshot
    final DepositTreeSnapshot depositTreeSnapshot =
        depositSnapshotLoader.loadDepositSnapshot().getDepositTreeSnapshot().orElseThrow();

    final Eth2NetworkConfiguration networkConfig =
        Eth2NetworkConfiguration.builder(eth2Network).build();
    final OkHttpClient httpClient = new Builder().connectionPool(new ConnectionPool()).build();
    final Web3j web3j = Web3j.build(new HttpService(endpoint, httpClient));
    final DepositContract depositContract =
        DepositContract.load(
            networkConfig.getEth1DepositContractAddress().toHexString(),
            web3j,
            new ReadonlyTransactionManager(
                web3j, networkConfig.getEth1DepositContractAddress().toHexString()),
            null);

    depositContract.setDefaultBlockParameter(
        DefaultBlockParameter.valueOf(
            depositTreeSnapshot.getExecutionBlockHeight().bigIntegerValue()));
    final Bytes32 expectedRoot = Bytes32.wrap(depositContract.getDepositRoot().send());
    assertThat(depositTreeSnapshot.getDepositRoot()).isEqualTo(expectedRoot);

    final Bytes depositCountBytes = Bytes.wrap(depositContract.getDepositCount().send());
    final UInt64 depositCount = MathHelpers.bytesToUInt64(depositCountBytes);
    assertThat(depositTreeSnapshot.getDepositCount()).isEqualTo(depositCount.longValue());

    // Check that the block hash is the canonical block at the expected block height
    final EthBlock block =
        web3j
            .ethGetBlockByNumber(
                DefaultBlockParameter.valueOf(
                    depositTreeSnapshot.getExecutionBlockHeight().bigIntegerValue()),
                false)
            .send();
    assertThat(block.getBlock().getHash())
        .isEqualTo(depositTreeSnapshot.getExecutionBlockHash().toHexString());
  }
```

I did have to add the getDepositCount and getDepositRoot methods back to our `DepositContract` class - they're from the code generated by web3j.

zilm13:
Yeah it's an option I mentioned and I'd be happy with it, but I wonder why current approach fails on your box, never failed for me

zilm13:
For the record:
It's clear now why my approach sometimes fails: it's not guaranteed that we will finalize `DepositTreeSnapshot` because sometimes eth1 height cache is not available. In this case it could be a bit old and doesn't match finalized `Eth1Data`. 
So, deposit contract query looks like the best approach

zilm13:
@ajsutton I've finished what you've almost done with web3j. Not sure that it's very clean, but anyway. Took 8s to verify on my pc for all 3 Infura supported networks. 
The other questions what we do with Gnosis in this case? drop it or keep without contract verification.
Also, do we try to verify age of snapshot, that it's not very outdated. It could be simply done with Infura too just by setting it shouldn't be older than, say, 10,000 blocks at the time of verification.

UPDATE: I've missed that you had code for contract, I didn't know it's generated, and it seems cleaner than mine. So you done everything. I did some cleanup with download script anyway. I like this approach

data/serializer/src/main/java/tech/pegasys/teku/api/migrated/BlockHeaderData.java
```diff
@@ -29,14 +30,17 @@ public class BlockHeaderData {
...
+          .finisher(Builder::build)
+          .withField("root", BYTES32_TYPE, BlockHeaderData::getRoot, Builder::root)
+          .withField("canonical", BOOLEAN_TYPE, BlockHeaderData::isCanonical, Builder::canonical)
           .withField(
               "header",
               SignedBeaconBlockHeader.SSZ_SCHEMA.getJsonTypeDefinition(),
-              BlockHeaderData::getHeader)
+              BlockHeaderData::getHeader,
+              Builder::header)
           .build();
```
ajsutton: We can undo these changes now as we don't need to deserialise the json when verifying the snapshot.

ajsutton (approved):
LGTM. I've push a couple of commits - one to remove the changes to `BlockHeaderData` since I'd rolled that back as part of establishing we didn't need it anymore.  The other tidies up the validation test a bit:

 - Adds support for gnosis verification using a public RPC endpoint
 - Uses an assumption to skip the test if INFURA_KEY isn't set which means it won't run in CI but will run locally without having to change the code more easily.
 - Few other little cleanups to fix warnings and stuff like that.

>Also, do we try to verify age of snapshot, that it's not very outdated. It could be simply done with Infura too just by setting it shouldn't be older than, say, 10,000 blocks at the time of verification.

I'm not really worried about verifying this - the snapshot will work fine as long as it actually matches the canonical chain. It's not a big deal if it's older than we might expect and it's much harder to get an out of date snapshot than it is to get the right one so not likely to be an issue.

So if you're happy with my last changes I think we're good to merge at last. 🎉

zilm13:
Polished a little bit more, updated and verified snapshots, merging

PR #6497, opened 2022-11-23
Download script for deposit snapshots

zilm13:
## PR Description
Not sure on this approach
Should be used from repo / with all environment variables set. When not set prints out what's wrong:
`./scripts/update-snapshots.sh beacon/pow/src/main/resources/snapshots`

## Fixed Issue(s)

Not sure yet

## Documentation

## Changelog


scripts/update-snapshots.sh
```diff
@@ -0,0 +1,73 @@
...
+require_env "SERVER_GNOSIS_URL"
+require_env "MAINNET_SSH_KEY"
+
+if [[ "$all_env_set" = false ]]; then
+  echo "Not all required environment variables are set. Following environment variables should be set:"
+  echo "SERVER_GOERLI_URL: ssh url to Goerli server with Teku like 'user@8.8.8.8'"
+  echo "SERVER_MAINNET_URL: ssh url to Mainnet server with Teku like 'user@8.8.8.8'"
+  echo "SERVER_SEPOLIA_URL: ssh url to Sepolia server with Teku like 'user@8.8.8.8'"
+  echo "SERVER_GNOSIS_URL: ssh url to Gnosis server with Teku like 'user@8.8.8.8'"
+  echo "MAINNET_SSH_KEY: path to SSH key required for Mainnet server like '~/.ssh/secret-key.pem'"
```
ajsutton: I'm not really sure why mainnet is special here. We'd likely get these snapshots from our mainnet-archive box so it has the same security setup as the testnet boxes.

ajsutton (approved):
LGTM.

scripts/update-snapshots.sh
zilm13: It makes things even simpler and more secure. Checked it with archive node, all good, updated

PR #6505, opened 2022-11-24
Implements BlockProcessor for Eip4844

### LINKED ISSUE #6392: Implement BlockProcessorEip4844
The `BlockProcessor` for `Eip4844` have to implement `process_blob_kzg_commitments` function which requires Blobs to be available.
(reference: https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#block-processing)

Not implementing any optimistic behaviour is the current way to go. `BlobsSidecar` must be present, otherwise block import must fail.

tbenr:
The implementation is a bit over complicated, on purpose. It follows the potential async process to retrieve the data in the future DAS world. It helped me to understand which will be the flow. We can simplify it here if we want to (starting from returning sync result instead of a `SafeFuture`.

Introduces `KzgCommitmentsProcessor`, responsible for running the validations on kzg commitments check (ultimately calling miscHelper)

Introduces `ForkChoiceBlobsSidecarAvailabilityChecker`, responsible for retrieving (currently doing nothing) and validating the blobs when they are actually required.

I'll probably remove `isDataAvailable` from miscHelper and directly use `validateBlobSidecar`: the entier `ForkChoiceBlobsSidecarAvailabilityChecker` is responsible for that.

fixes #6392 

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobsSidecarAvailabilityChecker.java
```diff
@@ -0,0 +1,106 @@
...
+      return BlobsSidecarAvailabilityChecker.invalidResult(blobsSidecar);
+    }
+
+    return BlobsSidecarAvailabilityChecker.validResult(blobsSidecar);
+  }
+
+  private boolean isBlockInDataAvailabilityWindow() {
+    final UInt64 currentSlot = recentChainData.getCurrentSlot().orElseThrow();
+
+    if (block
```
zilm13: If I'm not wrong, it will be always `false`. Also it could be simplified. And I'd add test for it

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobsSidecarAvailabilityChecker.java
```diff
@@ -0,0 +1,106 @@
...
+  @Override
+  public SafeFuture<BlobsSidecarAndValidationResult> validateBlobsSidecar() {
+
+    // in the current 4844 specs, the blobsSidecar is immediately available with the block
+    // so if we have it we do want to validate them regardless
+    if (blobsSidecar.isPresent()) {
+      return validate(blobsSidecar.get());
+    }
+
+    if (isBlockInDataAvailabilityWindow()) {
```
zilm13, line 72: I'd add a comment here, it's not obvious

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -664,7 +668,9 @@ public BeaconState processBlock(
               blockSlotState,
               IndexedAttestationCache.NOOP,
               signatureVerifier,
-              payloadExecutor);
+              payloadExecutor,
```
zilm13, line 671: I don't like passing here these 3 which is not needed on early forks, but we create `payloadExecutor` and `blobsSidecarAvailabilityChecker` for every block, so we could not put it, say, in SpecLogic. We could put there only `kzgCommitmentsProcessor`, which is singleton. We could create very specific container, but it's definitely a bad way. Ok, let it go ;)

zilm13 (commented):
Some thoughts

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobsSidecarAvailabilityChecker.java
tbenr: give another look now. test added

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobsSidecarAvailabilityCheckerTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Test
+  void shouldReturnInvalid() {
+    prepareBlockAndBlobInAvailabilityWindow(true);
+
+    blobsSidecarAvailabilityChecker.initiateDataAvailabilityCheck();
+    assertThat(blobsSidecarAvailabilityChecker.getAvailabilityCheckResult())
+        .isCompletedWithValueMatching(
+            result ->
+                result.isFailure()
```
zilm13: I'd add valid case too, when reading this one it's especially unclear why it's invalid until checking whole logic

tbenr: added

zilm13 (approved):
LGTM

PR #6508, opened 2022-11-24
Reworked the fallback to FcUV1 logic

rolfyone:
- it's not really ideal to call v1 if payload is empty, because that happens in capella too.

## Documentation

## Changelog


ajsutton (approved):
LGTM. Getting annoying that ELs don't just support V2 for pre-capella...

PR #6510, opened 2022-11-25
Provide default values for all milestone config items

ajsutton:
## PR Description
Spec config handling has been adjusted to make sure that some value is always available for each known milestone. This is required for the REST API to access all the types used in the API definitions even if the milestone isn't actually enabled.

If a fork epoch is specified for the milestone, all values for that milestone must be explicitly specified. If no fork epoch is specified, default, roughly MainNet like values are provided for all values and the fork epoch is set to FAR_FUTURE_EPOCH. This ensures that if the milestone is enabled, the config is actually providing all required values.

When loading a remote config, the preset specified (if present, which it should always be) will now be loaded. Values from the remote config still override the preset values though.

This effectively means we're dropping support for network configs that use the old style of all constants in one file with no preset. It's been over a year and two hard forks (I think it even predates mainnet launch), I think that's ok.

## Documentation

## Changelog


ajsutton:
I'm thinking that I'll do another PR to move these build classes into their own package and then the `AltairSpecBuilder` etc can be in top level files so `SpecConfigBuilder` doesn't just keep getting bigger and bigger all the time.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigBuilder.java
```diff
@@ -19,19 +19,20 @@
...
 import org.apache.tuweni.units.bigints.UInt256;
 import tech.pegasys.teku.ethereum.execution.types.Eth1Address;
 import tech.pegasys.teku.infrastructure.bytes.Bytes4;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 
+@SuppressWarnings({"UnusedReturnValue", "unused"})
 public class SpecConfigBuilder {
+
+  // Placeholder version explicitly doesn't match MainNet (or any other known testnet)
+  private static final Bytes4 PLACEHOLDER_FORK_VERSION = Bytes4.fromHexString("0x99999999");
```
rolfyone, line 35: want to set this to `0xffffffff`?

rolfyone (approved):
LGTM, minor q.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigBuilder.java
ajsutton: eh. It's entirely arbitrary and never actually used. Just making it different to MainNet for safety really.

PR #6511, opened 2022-11-25
Move spec builders to a separate package

ajsutton:
## PR Description
Splits the fork specific builders out into their own classes to avoid SpecConfigBuilder growing bigger all the time.

Also ensures that overridden values for capella and eip4844 fork epoch are added to the raw config (used by the REST API that remote VCs load config from).

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6512, opened 2022-11-25
Changed withdrawals to use hash comparison

rolfyone:
This allows us to process withdrawals from a header and still get the correct resulting hash.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -170,48 +170,35 @@ public void processBlsToExecutionChangesNoValidation(
...
-    final SszList<Withdrawal> payloadWithdrawals = executionPayloadCapella.getWithdrawals();
-    final List<Withdrawal> expectedWithdrawals = getExpectedWithdrawals(state);
-    if (expectedWithdrawals.size() != payloadWithdrawals.size()) {
+    final SszList<Withdrawal> expectedWithdrawals =
+        schemaDefinitionsCapella
+            .getExecutionPayloadSchema()
+            .getWithdrawalsSchemaRequired()
+            .createFromElements(getExpectedWithdrawals(state));
+
+    if (payloadSummary.getOptionalWithdrawalsRoot().isPresent()
```
ajsutton: We should fail if withdrawals root isn't present as that means withdrawals were missing (and/or we've got a block from the wrong milestone somehow).

lucassaldanha (approved):

ajsutton (approved):
LGTM, just should be stricter on requiring the withdrawals root to be present.

PR #6513, opened 2022-11-25
Fixed blinded block on remote VC in capella

rolfyone:
- needed to fix how we get blsToExecutionChanges

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
@@ -35,6 +35,7 @@ import tech.pegasys.teku.spec.datastructures.operations.Attestation;
 import tech.pegasys.teku.spec.datastructures.operations.AttesterSlashing;
 import tech.pegasys.teku.spec.datastructures.operations.Deposit;
 import tech.pegasys.teku.spec.datastructures.operations.ProposerSlashing;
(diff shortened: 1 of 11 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/BeaconBlockBody.java
```diff
@@ -75,6 +76,8 @@ default Optional<ExecutionPayloadSummary> getOptionalExecutionPayloadSummary() {
     return Optional.empty();
   }
 
+  Optional<SszList<SignedBlsToExecutionChange>> getOptionalBlsToExecutionChanges();
```
tbenr, line 79: why don't you go for
```java
  default Optional<SszList<SignedBlsToExecutionChange>> getOptionalBlsToExecutionChanges() {
    return Optional.empty();
  }
 ````
 like the others `getOptional`?

rolfyone: Adrian has strong views on it being an anti-pattern, so I tried to avoid it...

ajsutton: Well strong is probably over stating it, but if the expectation is that any new implementations would want to implement the method, then it's probably better to not have a no-op default implementation - that way you can't forget to implement it.  Default methods do make a lot of sense if they're likely to be the right implementation for new classes though.

ajsutton (approved):
LGTM.

Issue #6472, opened 2022-11-18
Doppelganger Detection acceptance tests


mehdi-aouadi:
Add a doppelganger detection acceptance test

PR #6514, opened 2022-11-25
Detect doppelganger for key manager api loaded keys

mehdi-aouadi:
## PR Description
Enable doppelganger detection for keys loaded through the key manager api (local and remote)

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6472

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
```diff
@@ -214,57 +217,107 @@ DeleteKeyResult deleteExternalValidator(final Validator activeValidator) {
...
+      maybeDoppelgangerDetector
+          .map(
+              doppelgangerDetector ->
+                  doppelgangerDetector
+                      .performDoppelgangerDetection(filterImportedPubKeys(importResults))
+                      .thenAccept(
+                          doppelgangerDetected -> {
+                            if (!doppelgangerDetected.isEmpty()) {
+                              System.exit(1);
+                            }
```
rolfyone: Im not sure this is what we want... if we've got a bunch of keys already loaded and i load one that is invalid, then we're going to system.exit a running validator?

Because it's batch loading, i think potentially you could justify rejecting the whole batch once the first fails, but I'm not super keen on crashing the validator...

Probably need more opinions @ajsutton

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
```diff
@@ -214,57 +217,107 @@ DeleteKeyResult deleteExternalValidator(final Validator activeValidator) {
...
+                      .performDoppelgangerDetection(filterExternallyImportedPubKeys(importResults))
+                      .thenAccept(
+                          doppelgangerDetected -> {
+                            if (!doppelgangerDetected.isEmpty()) {
+                              System.exit(1);
+                            }
+                          }))
+          .orElse(SafeFuture.COMPLETE)
+          .thenAccept(__ -> validatorTimingChannel.onValidatorsAdded())
+          .ifExceptionGetsHereRaiseABug();
```
rolfyone: i think for external signers we need to be super careful. some people (against our recommendation) do run active/active, and this would cause them problems if they turn dopelganger on..

Ultimately, that's kind of why an external signer is in use in the first place, so downtime can be minimized moving keys etc, because the signing records are safe...

Potentially we don't run this protection on dopelganger, or we potentially do something clever, but i think ditto for the system.exit probably not being what we want...

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
rolfyone: another case that may be interesting, a user removes a key, decides they've made a mistake and re-adds to the same validator...
If slashing protection data is in place and current on import, i wonder if we could be clever...

eg. if they have slashing protection records from last epoch, there's not much point checking if they were active previous epoch, because we cant sign anyway... not a fully formed thought...

ajsutton: hmm, yeah probably not good to exit in that case. It's hard to know how to make it obvious that the key was disabled because of doppelgänger detection though - a single log line would be easy to miss. Might be worth seeing what other clients do in this case.


Not sure if we should trust slashing protection records though. You may have removed the validator then added it to a different node and then also try to readd it to this one by mistake. I think in general if you enable doppelgänger support, you're saying that you want to skip 2 epochs after a new key is added in order to be maximally safe.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
ajsutton: Again I think if you've chosen to enable doppelgänger, you're explicitly asking us to keep you safe even if that means missing some attestations when keys are first added or on restart. If you're running active/active or want to minimise downtime while moving validators you shouldn't enable doppelgänger.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
rolfyone: I'd happily reject the batch, that'd be in line with other import failures...

ajsutton: hmm, yeah possibly, but most of the time when we reject a batch we also return an API error response but in this case we're already well past that point. So it's really a question of whether it makes it more obvious there's a problem if none of the validators activate or if that just makes it cost more...

mehdi-aouadi: @rolfyone regarding the keys rejection, as mentioned by @ajsutton: The API response doesn't wait for the doppelganger check result (Which could take up to 2 epochs and hence makes it unreasonable). It returns the result of the keys import only and the doppelganger check is run separately.
One option here, instead of shutting down the whole VC, could be to ignore/exclude the detected doppelganger keys or the whole batches containing them and that would be reported in the logs only (And not in the API response).
I think we can assume that when a user activates the doppelganger detection, he should be aware of this behaviour and hence check the logs for further infos.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
mehdi-aouadi: Again, the only other option I see here instead of shutting down the VC would be to exclude/ignore the detected doppelganger keys.
I think we should make it clear that the doppelganger detection is "not perfect" and that running it clearly means missing some duties (which means penalties). That should be an accepted side effect.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
mehdi-aouadi: @rolfyone regarding the slashing protection data, I am not sure how we could use it for this purpose. The idea here is to check if the keys are used elsewhere in the network and not locally. Otherwise we could reject any keys that have been used even after a restart. Let me know if I am missing something here

ajsutton: Given Lighthouse is doing a full exit in this case and on the basis that users activated doppelgänger because they wanted to prioritise safety over availability I think I've come around to exiting.

I'd suggest not calling `System.exit` directly in the code though - it's better to inject an instance of an object that calls exit so that we can replace it with a mock in tests and check it was told to exit without actually exiting the test JVM.

rolfyone: Ok we can probably document the behaviour I guess. It seems like pretty bad UX, If you're using keymanager you and dopelganger you'll be restricted from any rewards for a further 2 epochs, and none of the validators will be active until you resolve your issue...

mehdi-aouadi:
@ajsutton @rolfyone I made the following changes:
- I encapsulated the DG detection action in a separate interface so we can easily inject/mock it for tests
- Instead of shutting down the VC in the keymanager API use case I now remove the detected DG keys and log that instead. I am not sure about the validators removal though, I need to make sure it's correct and sufficient (Followed what we do for the key deletion API and removed some unnecessary processing, mainly the one related to the slashing protection export)
- Added some UTs for the `ActiveKeyManager`

To summarize, here is the behaviour now:
Post Keys -> Import the keys -> Perform DG detection and in // return the import result -> If DG detected the imported validator doppelgangers are removed

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -161,21 +162,46 @@ public void validatorsInitialised(final List<String> validators) {
...
-                            "Index: %d, Public key: %s",
+                            "Index: %s, Public key: %s",
                             doppelgangerInfo.getKey(), doppelgangerInfo.getValue()))
             .collect(Collectors.joining("\n", "\n", "\n"));
     log.fatal(
         "Detected {} validators doppelganger: {}", doppelgangersInfo.size(), doppelgangersLogInfo);
   }
 
+  public void doppelgangerDetectionAlert(final Set<String> doppelgangerPublicKeys) {
+    log.fatal(
```
ajsutton: This isn't fatal if we're not exiting as a result - it would just be an error level.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -235,23 +245,39 @@ public static ValidatorClientService create(
...
+      DataDirLayout dataDirLayout,
+      Optional<DoppelgangerDetector> maybeDoppelgangerDetector) {
+    RestApi validatorRestApi =
+        ValidatorRestApi.create(
+            validatorRestApiConfig,
+            proposerConfigManager,
+            activeKeyManager,
+            dataDirLayout,
+            maybeDoppelgangerDetector,
+            Optional.of(new DoppelgangerDetectionAlert()));
```
ajsutton: So for these `DoppelgangerDetector` I wouldn't use an `Optional` but use pass the class directly. In a test we can easily create a mock implementation to avoid actually shutting down and then the code doesn't have to deal with it potentially not being present.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
```diff
@@ -160,6 +165,33 @@ public DeleteRemoteKeysResponse deleteExternalValidators(List<BLSPublicKey> vali
     return new DeleteRemoteKeysResponse(deletionResults);
   }
 
+  private void deleteExternalDoppelgangers(final Set<BLSPublicKey> publicKeys) {
```
ajsutton: I'm not a fan of duplicating the deletion logic here. Given it's a rare situation and the implementation is fairly complex and important to get exactly right, we'd be better off reusing the existing code even if it performs additional computation.

Can probably call the existing `deleteExternalValidators` method and then checking if the result is successful or not and reporting from there.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
```diff
@@ -160,6 +165,33 @@ public DeleteRemoteKeysResponse deleteExternalValidators(List<BLSPublicKey> vali
...
+        if (deleteKeyResult.equals(DeleteKeyResult.success())) {
+          validatorLoader.getOwnedValidators().removeValidator(publicKey);
+          LOG.info(
+              "Removed validator doppelgangers public key {}", publicKey.toAbbreviatedString());
+
+        } else {
+          LOG.warn(
+              String.format(
+                  "Unable to remove validator doppelganger public key %s. %s",
+                  publicKey.toAbbreviatedString(), deleteKeyResult.getMessage().orElse("")));
```
ajsutton: I'd say this would be an error not a warning.  But under what situation could we be failing to remove a validator? Does it potentially mean that we are left in an unsafe situation with the validator still active and potentially getting slashed?

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
```diff
@@ -214,57 +271,141 @@ DeleteKeyResult deleteExternalValidator(final Validator activeValidator) {
...
         reloadRequired = true;
       }
     }
     if (reloadRequired) {
-      validatorTimingChannel.onValidatorsAdded();
+      if (maybeDoppelgangerDetector.isPresent()) {
+        maybeDoppelgangerDetector
+            .get()
+            .performDoppelgangerDetection(filterImportedPubKeys(importResults))
+            .thenAccept(
```
ajsutton: Haven't we already loaded the keys and added them to the `OwnedValidator` set at this point? It seems too late to be enabling doppelgänger.  The call to `validatorTimingChannel.onValidatorsAdded` is just a trigger to immediately reload duties for validators, but duties could be loaded at any point (e.g by a reorg). What stops the validator's still performing doppelgänger detection from having duties calculated?

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetectionAction.java
```diff
@@ -0,0 +1,26 @@
...
+package tech.pegasys.teku.validator.client.doppelganger;
+
+import java.util.Set;
+import tech.pegasys.teku.bls.BLSPublicKey;
+
+public interface DoppelgangerDetectionAction {
+
+  default void shutDown() {
+    System.exit(1);
+  }
```
ajsutton: Probably shouldn't be a default method given it's only used by one of two implementations. I'd just move it to the implementation.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
mehdi-aouadi: I refactored the existing deletion logic to pass on the public keys (The same for the import results) and reused it for both internal and external validators

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetectionAction.java
mehdi-aouadi: Done. I moved the shutdown to the corresponding implementation

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
mehdi-aouadi: Theoretically this shouldn't happen when importing new keys. The deletion result is more relevant when it comes to existing keys deletion. I added this log for debug purposes (If for some reasons the deletion fails). And yes if it happens it means that the validator will still stay active and hence a potential slashing (which is the downside of not shutting down the VC and rather trying to remove the keys).
I changed it to an error anyway.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -161,21 +162,46 @@ public void validatorsInitialised(final List<String> validators) {
...
+                            "Index: %s, Public key: %s",
                             doppelgangerInfo.getKey(), doppelgangerInfo.getValue()))
             .collect(Collectors.joining("\n", "\n", "\n"));
     log.fatal(
         "Detected {} validators doppelganger: {}", doppelgangersInfo.size(), doppelgangersLogInfo);
   }
 
+  public void doppelgangerDetectionAlert(final Set<String> doppelgangerPublicKeys) {
+    log.error(
+        "Detected {} validators doppelganger. The following public keys will be ignored: {}",
```
rolfyone: is it more that the keys have been removed than ignored?

I wonder if we want to say 
`Detected {} validators that appear to be already active. The following keys have been deleted after being imported: {}`

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
```diff
@@ -0,0 +1,297 @@
...
+    this.genesisDataProvider = genesisDataProvider;
+    this.checkDelay = checkDelay;
+    this.timeout = timeout;
+    this.maxEpochs = maxEpochs;
+  }
+
+  public SafeFuture<Map<UInt64, BLSPublicKey>> performDoppelgangerDetection(
+      final Set<BLSPublicKey> pubKeys) {
+    if (pubKeys.isEmpty()) {
+      LOG.info("Skipping doppelganger detection: No public keys to check");
```
rolfyone: doesn't look like it should be info?

rolfyone: if it's once on startup I could live with it though...

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
```diff
@@ -0,0 +1,297 @@
...
+      return SafeFuture.fromRunnable(
+          () -> Optional.ofNullable(doppelgangerDetectionTask).ifPresent(Cancellable::cancel));
+    }
+
+    private SafeFuture<Void> performDoppelgangerCheck(final Set<BLSPublicKey> pubKeys) {
+      if (timeProvider
+          .getTimeInSeconds()
+          .minus(startTime)
+          .isGreaterThanOrEqualTo(timeout.toSeconds())) {
+        LOG.warn("Doppelganger Detection timeout");
```
rolfyone: When i see this as a user, what do I do? As a warning, it's going to be pretty visible, and users will want to know what actions they can take...

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
```diff
@@ -0,0 +1,297 @@
...
+        return stopDoppelgangerDetector(new HashMap<>());
+      }
+
+      return genesisDataProvider
```
rolfyone, line 169: this probably needs to be broken down into something more maintainable...

rolfyone:
I think the overall flow is good, probably just a bit of cleanup etc to go...

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
mehdi-aouadi: Done. I refactored some code blocks into separate methods

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
mehdi-aouadi: Done. Added more details and moved the log to the `StatusLogger`

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
mehdi-aouadi: Yep it's not supposed to happen too often so I guess we can leave it like it is. Added a small detail though

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
mehdi-aouadi: Actually deleting was the first approach (And it doesn't work since the validators could have their duties scheduled before the DG finishes). Now it's more ignoring the DG keys. I updated the log though

validator/client/src/main/java/tech/pegasys/teku/validator/client/ActiveKeyManager.java
mehdi-aouadi: Right. I implemented a new approach: Instead of loading the validators and then deleting the detected doppelgangers, we now load the keys without adding the validators (Hence we're sure they don't get their duties scheduled). Once the DG check is finished we add the validators (and ignore the doppelganger keys)

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetectionShutDown.java
```diff
@@ -0,0 +1,27 @@
...
+
+package tech.pegasys.teku.validator.client.doppelganger;
+
+import java.util.List;
+import tech.pegasys.teku.bls.BLSPublicKey;
+
+public class DoppelgangerDetectionShutDown implements DoppelgangerDetectionAction {
+  @Override
+  public void shutDown() {
+    System.exit(1);
```
ajsutton, line 28: Do we need to log something to say why we're shutting down in this case?

ajsutton (approved):
LGTM. Just need to make sure we are logging a message before shutting down if doppelgängers are detected at startup.

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetectionShutDown.java
mehdi-aouadi: I added a log

Issue #6418, opened 2022-11-10
Replace ExecutionPayloadSchema.create methods with builder


ajsutton:
### Description
Currently `ExecutionPayload` and `ExecutionPayloadHeader` instances are created via milestone specific `create` methods in their schemas.  This makes the creation process tightly coupled to the spec milestone in use.

We need to move to a builder pattern like is used for BlockBody.

ajsutton:
As part of this we should be able to remove the `toVersionBellatrix()` method on the schema objects. At minimum we need to carefully review all the calls to it for ones that may wind up being given a Capella instance (which would return empty).

PR #6518, opened 2022-11-28
Introduce execution payload builder

ajsutton:
## PR Description
Introduces a builder for execution payload in the same vein as for BeaconBlockBody and removes the milestone specific create methods from schemas.

Also removes toVersionX() methods from ExecutionPayloadSchema as their behaviour didn't match those of ExecutionPayload.toVersionX methods (ExecutionPayloadSchema.toVersionCapella would fail for EIP4844 but ExecutionPayload.toVersionCapella would succeed for EIP4844). They are no longer used except in fuzzing which specifically needs the bellatrix schema.

Avoids the need to branch by milestone everywhere we create an ExecutionPayload.

Part of #6418 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6520, opened 2022-11-29
Implement a wrapper around c-kzg jni bindings

### LINKED ISSUE #6470: Add c-kzg java binding dependency in infrastructure crypto module
Once we have the new binding (#6389) we have to add it as a dependency in teku.
We we also need to take care about the initialization (see #6382)

zilm13:
## PR Description
I'm pretty unsure on everything here, but need to start from something
Some notes:
- I was looking at bls module while making this
- I'm not sure that we need `KZGLoader` here
- No optimizations were taken into consideration at that stage
- Test is pretty slow, but I've not found `minimal` setup anywhere
- I've isolated reset exception if nothing is loaded, I'm not sure it's useful, looks more like a bug you could get into from my side

## Fixed Issue(s)

Fixes #6470 
## Documentation

## Changelog


infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/impl/KZG4844.java
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Loads the trusted setup from a file. Once loaded, the same setup will be used for all the
+   * calls. To load a new setup, reset the current one by calling {@link #resetTrustedSetup()} and
+   * then load the new one. If no trusted setup has been loaded, all other API calls will throw an
+   * exception.
+   *
+   * @param path a path to a trusted setup file in filesystem
+   * @throws KzgException if file not found or arguments from file are incorrect
+   */
+  void loadTrustedSetup(final String path) throws KzgException;
```
tbenr: we don't use `final` in interfaces

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/impl/KZG4844.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Calculates aggregated proof for the given blobs
+   *
+   * @param blobs Blobs
+   * @return the aggregated proof
+   */
+  KZGProof computeAggregateKzgProof(List<Bytes> blobs);
```
tbenr: shouldn't we have `throws KzgException;` on all methods here?

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public static boolean verifyKzgProof(
+      final KZGCommitment kzgCommitment,
+      final Bytes32 z,
+      final Bytes32 y,
+      final KZGProof kzgProof) {
+    return kzgImpl.verifyKzgProof(kzgCommitment, z, y, kzgProof);
+  }
+
+  private static String copyResourceToTempFile(final String path) throws IOException {
```
tbenr: could we pass-through the file if it is a local path?

tbenr (commented):
Makes sense to me to leave `CkzgLoader` as a decoupling layer in case we'll different libraries in the future to choose from.

I provided some initial comments

infrastructure/kzg/src/test/java/tech/pegasys/teku/kzg/KZGTest.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Test
+  public void testKzgLoadTrustedSetup() {
+    final String resourcePath =
+        KZGTest.class.getResource(MAINNET_TRUSTED_SETUP_TEST).toExternalForm();
+    KZG.loadTrustedSetup(resourcePath);
+    KZG.resetTrustedSetup();
+  }
+
```
tbenr, line 59: We could add another test named like `testKzgLoadTrustedSetupTwice` to verify that we cannot load a setup without resetting the previous one

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/impl/KZG4844.java
zilm13: My bad, fixed

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
zilm13: yes, but I guess we will need absolute path
also how could I recognize it?

tbenr:
Worth also checking what happens if we call the kzg functions without having previously loaded any setup

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/impl/ckzg/CkzgKZG4844.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  @Override
+  public boolean verifyAggregateKzgProof(
+      final List<Bytes> blobs, final List<KZGCommitment> kzgCommitments, final KZGProof kzgProof)
+      throws KzgException {
+    try {
+      return CKzg4844JNI.verifyAggregateKzgProof(
+          blobs.stream().reduce(Bytes::wrap).orElseThrow().toArray(),
```
tbenr: empty blobs and commitments are actually a thing: https://github.com/ethereum/consensus-specs/pull/3093
I think we should cover the case here and in tests

zilm13: yeah, doing it already, nice catch!

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
tbenr: is you switch path from `String` to `URI` and then `getScheme().equalsIgnoreCase("file")` ?

zilm13: Thank you
also I decided to pass URL there instead of String, so it's clear that it comes with scheme and not just path

tbenr (approved):
LGTM.
we should do an optimization round. I'll track it in the epic backlog

PR #6526, opened 2022-11-30
BlobsSidecarAvailabilityChecker return not_required when availability check is not initiated

tbenr:
An alternative to this is to put the checker in the spec logic.
This solution follows the `ForkChoicePayloadExecutor` pattern, essentially acting as a NOOP if not engaged during block import.

related to #6522

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
@@ -277,9 +277,10 @@ public class BlockManager extends Service
                         FailedPayloadExecutionSubscriber::onPayloadExecutionFailed, block);
                     break;
                   case FAILED_BLOBS_AVAILABILITY_CHECK:
-                    // TODO: we cannot invalidate the block root since it passed the checks
-                    // we should maintain a set of invalid BlobsSidecar root OR
-                    // SignedBeaconBlockAndBlobsSidecar root
+                    // TODO:
+                    //  Trigger the fetcher in the case the coupled BeaconBlockAndBlobsSidecar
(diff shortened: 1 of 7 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -276,6 +276,11 @@ private SafeFuture<BlockImportResult> handleBlockAndBlobsSidecarImport(
                     failedPayloadExecutionSubscribers.deliver(
                         FailedPayloadExecutionSubscriber::onPayloadExecutionFailed, block);
                     break;
+                  case FAILED_BLOBS_AVAILABILITY_CHECK:
+                    // TODO: we cannot invalidate the block root since it passed the checks
```
zilm13: I think that by the logic, there should be something similar to `UNKNOWN_PARENT` block here, but as it's our `NOOP` logic, we should just pass it. So it's a bit weird stub

tbenr: Yes, good point. I'll mention the comment in the `FetchRecentBlocksService` issue description

tbenr: for the records, our NOOP will not trigger this anymore.

zilm13: I think we should reflect something like this here:
TODO: trigger the fetcher in the case the coupled `BeaconBlockAndBlobsSidecar` contains a valid block but the BlobsSidecar validation fails like we do with `pendingBlocks`

zilm13 (approved):
LGTM, just one comment

PR #6531, opened 2022-12-01
engine_getBlobsBundleV1 in engine API

### LINKED ISSUE #6387: Implement `engine_getBlobsBundleV1` in engine API
https://github.com/ethereum/execution-apis/blob/main/src/engine/blob-extension.md

- [ ] ssz object `BlobsBundleV1`
- [ ] implement `engineGetBlobsBundle(payloadId)`

zilm13:
## PR Description
- I don't see in spec that it should be ssz, made json like any other engine API
- Not sure on the way I handle possible pre-EIP4844 hits
- I've noticed that in spec Blob is `ByteVector[BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB]` and in EIP-4844 it's `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]`. We have `Blob` like in spec. Do you think it's ok? I think we don't care (at least yet) what's inside it, so we should be comfortable to black box `ByteVector`
- I see that there are some verification in https://github.com/ethereum/execution-apis/blob/main/src/engine/blob-extension.md#BlobsBundleV1 we will need to add when joining it with Payload but looks like it's another task, just not to forget

## Fixed Issue(s)

Fixes #6387 
## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/Eip4844ExecutionClientHandler.java
```diff
@@ -0,0 +1,60 @@
...
+                    SchemaDefinitionsEip4844.required(spec.atSlot(slot).getSchemaDefinitions())
+                        .getBlobSchema()),
+            BlobsBundleV1::asInternalBlobsBundle)
+        .thenPeek(
+            blobsBundle ->
+                LOG.trace(
+                    "engineGetBlobsBundle(payloadId={}, slot={}) -> {}",
+                    payloadId,
+                    slot,
+                    blobsBundle.toBriefBlobsString()));
```
tbenr: I'd move to suppliers here just to avoid calling `toBriefBlobsString()` every time

                LOG.trace(
                    "engineGetBlobsBundle(payloadId={}, slot={}) -> {}",
                        () -> payloadId,
                        () -> slot,
                        blobsBundle::toBriefBlobsString));

tbenr (approved):
LGTM with just one thing that should be changed

tbenr:
We could also plug this in `ExecutionLayerManagerImpl::create` already.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/Eip4844ExecutionClientHandler.java
zilm13: nice point!

zilm13:
We'd better plug it not on friday evening. Fixed logging

Issue #6500, opened 2022-11-24
Initial Capella ATs


rolfyone:
Former: failing to process first block after epoch transition coming in on capella gossip

```
[...]

PR #6534, opened 2022-12-01
Initial Capella ATs

lucassaldanha:
## PR Description
- Added AT for Capella upgrade
- Added AT for withdrawal credentials update

## Fixed Issue(s)
#6500

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BlsToExecutionChangeAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BlsToExecutionChangeAcceptanceTest.java
@@ -39,7 +39,7 @@ public class BlsToExecutionChangeAcceptanceTest extends AcceptanceTestBase {
     final BLSKeyPair validatorKeyPair = blsKeyPairs.get(validatorIndex);
     final Eth1Address executionAddress =
         Eth1Address.fromHexString("0xFE3B557E8Fb62b89F4916B721be55cEb828dBd73");
-    final UInt64 capellaActivationEpoch = UInt64.valueOf(1);
+    final UInt64 capellaActivationEpoch = UInt64.ONE;
 
     TekuNode primaryNode =
         createTekuNode(
@@ -50,8 +50,8 @@ public class BlsToExecutionChangeAcceptanceTest extends AcceptanceTestBase {
               c.withInteropValidators(0, 8);
               c.withStubExecutionEngine();
               c.withValidatorProposerDefaultFeeRecipient(executionAddress.toHexString());
-              c.withAltairEpoch(UInt64.valueOf(0));
-              c.withBellatrixEpoch(UInt64.valueOf(0));
+              c.withAltairEpoch(UInt64.ZERO);
+              c.withBellatrixEpoch(UInt64.ZERO);
               c.withCapellaEpoch(capellaActivationEpoch);
             });
     primaryNode.start();
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BlsToExecutionChangeAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BlsToExecutionChangeAcceptanceTest.java
@@ -18,14 +18,10 @@ import org.junit.jupiter.api.Test;
 import tech.pegasys.teku.bls.BLSKeyPair;
 import tech.pegasys.teku.ethereum.execution.types.Eth1Address;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
-import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.datastructures.interop.MockStartValidatorKeyPairFactory;
-import tech.pegasys.teku.spec.datastructures.operations.SignedBlsToExecutionChange;
-import tech.pegasys.teku.spec.logic.versions.capella.block.BlockProcessorCapella;
 import tech.pegasys.teku.test.acceptance.dsl.AcceptanceTestBase;
 import tech.pegasys.teku.test.acceptance.dsl.TekuNode;
 import tech.pegasys.teku.test.acceptance.dsl.TekuNode.Config;
-import tech.pegasys.teku.test.acceptance.dsl.tools.BlsToExecutionChangeCreator;
 
 public class BlsToExecutionChangeAcceptanceTest extends AcceptanceTestBase {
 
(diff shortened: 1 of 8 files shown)


acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BlsToExecutionChangeAcceptanceTest.java
```diff
@@ -0,0 +1,77 @@
...
+              c.withInteropValidators(0, 8);
+              c.withStubExecutionEngine();
+              c.withValidatorProposerDefaultFeeRecipient(executionAddress.toHexString());
+              c.withAltairEpoch(UInt64.ZERO);
+              c.withBellatrixEpoch(UInt64.ZERO);
+              c.withCapellaEpoch(capellaActivationEpoch);
+            });
+    primaryNode.start();
+    primaryNode.waitForOwnedValidatorCount(8);
+    primaryNode.waitForMilestone(SpecMilestone.CAPELLA);
```
ajsutton: Is there a reason we're waiting an epoch to activate capella? Seems like it's just slowing the test down but not really adding any value over just starting with capella at epoch 0.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BlsToExecutionChangeAcceptanceTest.java
```diff
@@ -0,0 +1,77 @@
...
+        new BlsToExecutionChangeCreator(
+                primaryNode.getSpec(), primaryNode.getGenesisValidatorsRoot())
+            .createAndSign(
+                UInt64.valueOf(validatorIndex),
+                validatorKeyPair.getPublicKey(),
+                executionAddress,
+                validatorKeyPair.getSecretKey(),
+                capellaActivationEpoch);
+
+    primaryNode.submitBlsToExecutionChange(signedBlsToExecutionChange);
```
ajsutton: We should push the change creator into the dsl, so I think this winds up something like:
```java
primaryNode.submitBlsToExecutionChange(validatorIndex, validatorKeyPair, executionAddress);
```

Even the detail of ensuring the capella fork is used rather than bellatrix should be handled by the DSL, not the test (though it wouldn't be an issue if we start capella at epoch 0).

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/BlsToExecutionChangeAcceptanceTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    primaryNode.waitForValidatorWithCredentials(
+        validatorIndex,
+        BlockProcessorCapella.getWithdrawalAddressFromEth1Address(executionAddress));
+  }
```
ajsutton: I'd push this into the DSL as well, so it's just `primaryNode.wairForValidatorWithWithdrawalAddress(validatorIndex, executionAddress)`.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/CapellaUpgradeAcceptanceTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    UInt64 genesisTime = primaryNode.getGenesisTime();
+
+    TekuNode lateJoiningNode =
+        createTekuNode(
+            c -> {
+              c.withGenesisTime(genesisTime.intValue());
+              c.withRealNetwork();
+              c.withPeers(primaryNode);
+              c.withInteropValidators(0, 0);
```
ajsutton: nit: we can chain these calls rather than using separate c. on each line.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -115,6 +123,10 @@ private TekuNode(final Network network, final DockerVersion version, final Confi
         .withCommand("--config-file", CONFIG_FILE_PATH);
   }
 
+  public Spec getSpec() {
+    return spec;
+  }
```
ajsutton: This doesn't feel like something that should be public... I think it's not needed if the BLS change creator gets pushed into the DSL.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -304,6 +338,19 @@ public void waitForNewFinalization() {
...
   }
 
+  public void waitForMilestone(SpecMilestone expectedMilestone) {
+    waitFor(
+        () -> {
+          final int currentEpoch = (int) getMetricValue("beacon_epoch");
+          final SpecMilestone currentMilestone =
+              spec.getForkSchedule().getSpecMilestoneAtEpoch(UInt64.valueOf(currentEpoch));
+          assertThat(currentMilestone).isEqualTo(expectedMilestone);
+          waitForNewFinalization();
```
ajsutton: Why do we need to wait for the milestone to finalize?

It may be better to wait for the log line (`Activating network upgrade: CAPELLA`) that the milestone has been activated rather than just waiting for the epoch.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -486,6 +533,18 @@ public void waitForValidators(int numberOfValidators) {
         });
   }
 
+  public void waitForValidatorWithCredentials(int validatorIndex, Bytes32 expectedCredentials) {
+    waitFor(
+        () -> {
+          Optional<BeaconState> maybeState = fetchHeadState();
```
ajsutton: Fetching the whole state seems overkill for this. We should just get the specific validator with `/eth/v1/beacon/states/{state_id}/validators/{validator_id}`

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
lucassaldanha: Don't remember why I had it there, probably something I was using during my tests and did not remove.

I was using the log line before, the problem with the log line is that it does not work when a late node is syncing. So it only works for nodes that go through the upgrade.

ajsutton: ah yeah for the other test that's an interesting case...  But getting the current epoch doesn't verify that it went through the transition at all, so there's probably no point in waiting for the transition to happen on a node that's not in sync. You'd just wait for it on the original node and know that they both have the same time so the second node must have the same current epoch anyway (then check they are actually in sync which we should already have a method for).

lucassaldanha: So even if the node spec says that the upgrade is at epoch x, the node could be at epoch x + 1 without going through the upgrade? I would expect some kind of failure if it ever happens.

ajsutton: You don't actually have the node spec from the running node though - you have the one from inside the DSL and while they should be the same, if there's a bug they may not be. So the only thing you're actually verifying is what the current epoch is and the AT already knows what epoch it scheduled the milestone to activate at and can just wait until the epoch without this method.

ajsutton (approved):
LGTM.

PR #6536, opened 2022-12-02
Updated PostBlsToExecutionChanges to reject changes prior to capella.

rolfyone:
Accepting a bellatrix bls_to_execution_change would lead to bad UX, because the user would have a good faith expectation that the change can be made (given that it was accepted), but in reality the change cannot be made even though it is valid pre-capella, because the signed object is no longer valid once Capella occurs and the signature is invalidated by the fork transition.

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

PR #6537, opened 2022-12-02
Added gossip filtering for bls_to_execution_change

rolfyone:
Gossip on channel before Capella can be ignored, as it won't be valid once capella starts, because it'll be an old signature.

## Documentation

## Changelog


ajsutton (approved):
LGTM.

Issue #4230, opened 2021-08-10
Add (experimental) light client server functionality


benjaminion:
See [these notes](https://notes.ethereum.org/@ralexstokes/S1RSe1JlF) for some background.

> A first milestone here is to fork one of the production-grade beacon clients and build out the aforementioned functionality.

I believe that Teku's architecture makes it well suited to implementing the light client server functionality and meeting this milestone.

jeyakatsa:
Hi @benjaminion !

I've began work on this here: https://github.com/jeyakatsa/teku/tree/lightclient/light-client

Will be updating and sending pull requests when necessary.

Available for any questions you or your team may have!

robzajac:
Bumping with work done in #6384.
[Endpoints](https://github.com/ethereum/beacon-APIs/tree/master/apis/beacon/light_client)
[Data types](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#containers)

- [`GetLightClientBootstrap`](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/bootstrap.yaml): Returns the data for clients to startup into the light client protocol, including the current sync committee and latest block header
- [`GetLightClientUpdatesByRange`](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/updates.yaml): Returns the LightClientUpdate for the provided sync committee range, including the attested block header and the next sync committee
- [`GetLightClientFinalityUpdate`](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/finality_update.yaml): Returns the latest known LightClientFinalityUpdate
- [`GetLightClientOptimisticUpdate`](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/optimistic_update.yaml): Returns the latest known LightClientOptimisticUpdate

Will continue to track in this issue.

robzajac:
Acceptance criteria:
- LC server endpoints are implemented, tested and deployed in teku
- Demonstrated a light client sync against a teku full node

API reference:
- [`consensus-specs`](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/full-node.md)
- [`beacon-APIs`](https://github.com/ethereum/beacon-APIs/tree/master/apis/beacon/light_client)

Progress update:

General
- [x] ~Update JSON serialization according to etan's update to the light client fields https://github.com/ethereum/beacon-APIs/pull/287~
[...]

PR #6538, opened 2022-12-02
Add utility for constructing a merkle proof against TreeNode.

robzajac:
## PR Description
Adds a utility class to construct the merkle proof for a field (represented by generalized index) in a given `TreeNode`.

Will shortly be used for verifying inclusion of `current_sync_committee` in the `BeaconState` for the `GetLightClientBootstrap` endpoint as per the [light client bootstrap spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/full-node.md#create_light_client_bootstrap).

See also [here](https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md#merkle-multiproofs) for more on merkle proofs and the `consensus-specs` implementations [setup.py](https://github.com/ethereum/consensus-specs/blob/dev/setup.py#L522) [merkle.py](https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/eth2spec/test/helpers/merkle.py#L4)

## Fixed Issue(s)
partially addresses #4230 

## Documentation

## Changelog


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/tree/MerkleUtil.java
```diff
@@ -0,0 +1,114 @@
...
+    return Collections.unmodifiableList(proof);
+  }
+
+  /**
+   * Returns the path from the root to the specified node as an array of generalized indices.
+   * path[0] is the root itself (gIndex of 1) path[path.length - 1] is the node (gIndex of
+   * `nodeIndex`)
+   */
+  @VisibleForTesting
+  static List<Long> getPathToNode(long nodeIndex) {
```
rolfyone, line 68: nit: do these map to functions in the spec? The tests are good, just wondering if there's an algorithm I'm checking them against, and if we can make that relationship a little clearer if they're just a spec function by naming them in a way we can relate them directly...

robzajac: The spec doesn't define these functions from what I can see. The light client spec defines a [`compute_merkle_proof_for_state`](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/full-node.md#compute_merkle_proof_for_state) whose implementation is not provided. The impl in this PR is more general purpose than just an interface accepting `BeaconState`. The [`Merkle multiproofs`](https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md#merkle-multiproofs) spec does not provide these functions either. The only definitions I could see came from the test cases (linked in OP).

rolfyone: ok all good, just makes it hard to verify against a spec :)

rolfyone (approved):
LGTM as a starting point.

robzajac:
Thanks @rolfyone do you mind merging in as I don't have write access? I'll followup with a PR integrating this into `GetLightClientBootstrap`

PR #6540, opened 2022-12-02
Support trusted setup configuration

### LINKED ISSUE #6382: Support trusted setup configuration
From [polynomial commitments](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md#trusted-setup):
- [ ] Support trusted setup loading. We should have a `trusted_setup.txt` backed in for testnets\mainnet with the ability to specify via CLI for devnets. We just have to use the file content as is to initialize the kzg library by calling `load_trusted_setup`.

We should have the trusted setup configured in `Eth2NetworkConfiguration` and map the corresponding setup in each `applyNetwork*` methods.

We also have to load it`KZG.loadTrustedSetup`. `BeaconChainController::initialize` is a good candidate

StefanBratanov:
## PR Description

- Create new option `--Xtrusted-setup`
- Use it to initialize `KZG` only if `EIP4844` is supported fork otherwise `NOOP`
- Initialize the KZG library with the configured field elements per blob

## Fixed Issue(s)
fixes #6382 

## Documentation

## Changelog


ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -375,6 +384,17 @@ public Builder eth1DepositContractDeployBlock(final long eth1DepositContractDepl
       return this;
     }
 
+    public Builder trustedSetup(final String trustedSetup) {
+      this.trustedSetup = Optional.of(trustedSetup);
+      return this;
+    }
+
+    public Builder trustedSetupFromClasspath(final String filename) {
+      this.genesisState =
```
zilm13: ouch

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
```diff
@@ -13,52 +13,54 @@
...
       throws IOException {
-    try {
-      if (trustedSetup.toURI().getScheme().equals(FILE_SCHEME)) {
-        // Platform-agnostic safe way to get path
-        return Paths.get(trustedSetup.toURI()).toFile().getPath();
-      }
-    } catch (final URISyntaxException ex) {
-      throw new IllegalArgumentException(trustedSetup + " is incorrect file path", ex);
+    final Optional<Path> trustedSetupFile = tryGetFile(trustedSetup);
+    if (trustedSetupFile.isPresent()) {
```
zilm13: it's present with jar

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
```diff
@@ -13,52 +13,54 @@
...
+      // Platform-agnostic safe way to get path
+      return trustedSetupFile.get().toFile().getPath();
     }
-
+    final String sanitizedTrustedSetup = UrlSanitizer.sanitizePotentialUrl(trustedSetup);
```
zilm13, line 48: we should do it before all I guess

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
StefanBratanov: ouch indeed, will fix

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
tbenr: I think we should sanitize the url before printing it out, not before using it.

zilm13: Yeah you are right, my wrong

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
```diff
@@ -13,52 +13,71 @@
...
+
+  private static Optional<Path> getFileOnFileSystem(final String resource) {
+    try {
+      return Optional.of(Paths.get(resource)).filter(Files::exists);
+    } catch (final Exception __) {
+      return Optional.empty();
+    }
+  }
+
+  private static Optional<Path> getFileOnClasspath(final String resource) {
```
zilm13: This interface and the one it's called from (`getFileOnFileSystemOrClasspath(final String resource)`) is a bit misleading. Reading it straight I'd expect that it will load any classpath resource, instead it tries to convert Uri to Path when it's local, not in jar. So I'd change this a little to be clear or add a comments

zilm13 (approved):
LGTM, only 1 thing

PR #6544, opened 2022-12-05
Change all ATs to use swift instead of less-swift.  WCPGW?

ajsutton:
## PR Description
Test what happens if we use `swift` instead of `less-swift` now that timing handling has been improved for gnosis.

## Documentation

## Changelog


ajsutton:
For the record, WCPGW = "What Could Possibly Go Wrong"

lucassaldanha (approved):

PR #6545, opened 2022-12-05
Speed up ATs by using a custom genesis state rather than processing deposits

ajsutton:
## PR Description
For ATs that need real validator keys but aren't actually testing the deposit contract interactions, switch to using a custom genesis state instead of having to spin up a besu instance and processing deposits.

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

Issue #6527, opened 2022-11-30
investigate 'no peers found on gossip topic'


rolfyone:
```
Failed to publish attestations(s) for slot NNN because no peers were available on the required gossip topic
```

rolfyone:
From Ben:
One hypothesis is that it is due to more nodes on the post-Merge network that are not running validators, making it harder to find the ones you want. Is increasing the number of peers likely to help (gut feel, yes)? If so, should we increase our default maxPeers?

tbenr:
from a user in discord:
he is experiencing some peering issues during sync committee duties. I asked him enable TRACE on `SyncCommitteeMessageGossipManager` and here it is:

```
2022-12-06 09:55:51.001 INFO  - Slot Event  *** Slot: 5291377, Block: b9ced3d8073485d1b190027e46feebc4600c1f5e5e481cdb91f7bcb520b5f015, Justified: 165354, Finalized: 165353, Peers: 100
2022-12-06 09:56:00.236 INFO  - Validator   *** Published sync_signature     Count: 1, Slot: 5291378, Root: e09755625ddaf5ba409134d0129ca612feaf7e0402eca82d5a736ccc38a65101
2022-12-06 09:56:00.236 TRACE - Failed to publish sync committee message for slot 5291378
java.util.concurrent.CompletionException: io.libp2p.pubsub.NoPeersForOutboundMessageException: No peers for message topics [/eth2/4a26c58b/sync_committee_3/ssz_snappy] found
at java.util.concurrent.CompletableFuture.encodeRelay(Unknown Source) ~[?:?]
at java.util.concurrent.CompletableFuture.completeRelay(Unknown Source) ~[?:?]
at java.util.concurrent.CompletableFuture$UniCompose.tryFire(Unknown Source) ~[?:?]
at java.util.concurrent.CompletableFuture.postComplete(Unknown Source) ~[?:?]
at java.util.concurrent.CompletableFuture$AsyncSupply.run(Unknown Source) ~[?:?]
at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) ~[?:?]
at java.util.concurrent.FutureTask.run(Unknown Source) ~[?:?]
at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source) ~[?:?]
at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) ~[?:?]
[...]

PR #6550, opened 2022-12-05
Add metric to report the number of peers subscribed to each attestation and sync committee subnet

ajsutton:
## PR Description
Add metric to report the number of peers subscribed to each attestation and sync committee subnet.

## Fixed Issue(s)
Will provide some insight into how much coverage of each topic we have is for #6527 

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/PeerSubnetSubscriptions.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/PeerSubnetSubscriptions.java
@@ -95,35 +95,38 @@ public class PeerSubnetSubscriptions {
                                           b.addSubscriber(syncCommitteeSubnet, subscriber));
                             }))
             .build();
-    updateMetrics(subnetPeerCountGauge, subscriptions);
+    updateMetrics(currentSchemaDefinitions, subnetPeerCountGauge, subscriptions);
     return subscriptions;
   }
 
   private static void updateMetrics(
+      final SchemaDefinitionsSupplier currentSchemaDefinitions,
       final SettableLabelledGauge subnetPeerCountGauge,
       final PeerSubnetSubscriptions subscriptions) {
-    updateMetrics(
-        subnetPeerCountGauge, subscriptions.attestationSubnetSubscriptions, "attestation_");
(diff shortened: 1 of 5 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/PeerSubnetSubscriptions.java
```diff
@@ -55,42 +56,69 @@ public static PeerSubnetSubscriptions create(
...
+                        .forEach(
+                            attestationSubnet -> {
+                              b.addRelevantSubnet(attestationSubnet);
+                              subscribersByTopic
+                                  .getOrDefault(
+                                      attestationTopicProvider.getTopicForSubnet(attestationSubnet),
+                                      Collections.emptySet())
+                                  .forEach(
+                                      subscriber -> b.addSubscriber(attestationSubnet, subscriber));
+                            }))
```
rolfyone, line 79: hopefully if we're not subscribed on a subnet any longer we don't retain the number from the last update... i could play with it to find out, but just a thought...

rolfyone: and i guess further, did we want to set to something to know its not subscribed... i guess that might be a bit random...

ajsutton: It's an interesting point actually. We're iterating all the attestation subnets all the time so that's not an issue, but sync committee subnets we only iterate when we actually need them, so potentially we'd fail to reset though.  Will take a look and see what we can do.

ajsutton: Fixed so now we iterate through all subnets for both attestations and sync committees when updating the metrics.

rolfyone: sounds like a good plan!

rolfyone (approved):
LGTM

PR #6552, opened 2022-12-05
Eip4844 activate KZG in MiscHelpers

zilm13:
## PR Description
I've added `isKZGNoop` instead of activating it say with empty `trusted_setup` to explicitly protect production for the case when KZG is not activated in real world
## Documentation

## Changelog


zilm13:
We'd better activate KZG in `MiscHelpersEip4844` than in `SpecLogicEip4844` I think

zilm13:
I don't like 2 comparisons with `FAR_FUTURE_EPOCH`, it's not a good way to check that milestone is supported. If someone has better idea, I'd appreciate it

tbenr (approved):
LGTM

PR #6561, opened 2022-12-06
Do one more duplication check before gossiping BlsToExecutionChange

jtraglia:
## PR Description

Noticed a small implementation difference between `SignedBlsToExecutionChangeValidator` and the other message validators. The others would perform a duplication check (for the same validator index) twice, but this one doesn't.

1. When receiving the message but before validation checks.
2. Right before returning `InternalValidationResult.ACCEPT`.

I believe the second one is important if we receive two `SignedBlsToExecutionChange` messages in close succession. If `validateForGossip` is called on both of them at almost the exact same time, the first duplication check most likely will not catch the duplication but the second duplication check would.

For example, `validateForGossip` for voluntary exits (scroll down to see all of it):

https://github.com/ConsenSys/teku/blob/22b93cb62e9326abd063d9531433ca138230e889/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/VoluntaryExitValidator.java#L50-L83

## Documentation

## Changelog


rolfyone (approved):
LGTM, nice pickup.

PR #6562, opened 2022-12-06
Change maxBlsToExecutionChanges from UInt64 to int

jtraglia:
## PR Description

Simply change the type of `maxBlsToExecutionChanges` from `UInt64` to `int`.

* The value is relatively small and doesn't need that range.
* The value isn't used as a `UInt64`, it's casted to a `long`.
* This matches the type of other max values.

For example:

https://github.com/ConsenSys/teku/blob/9f7eed52c860645a288e42a8dc9edbcff3481197/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfig.java#L123-L131

## Documentation

## Changelog


rolfyone (approved):
LGTM thanks!

PR #6565, opened 2022-12-07
Update to jc-kzg-4844 0.1.0

StefanBratanov:
## Documentation

## Changelog

### DIFF
--- a/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
+++ b/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
@@ -76,7 +76,7 @@ public final class CKZG4844 implements KZG {
   private CKZG4844(final Preset preset) {
     try {
       CKZG4844JNI.loadNativeLibrary(preset);
-      LOG.debug("Loaded C-KZG-4844");
+      LOG.debug("Loaded C-KZG-4844 with {} preset", preset);
     } catch (final Exception ex) {
       throw new KZGException("Failed to load C-KZG-4844 library", ex);
     }
--- a/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
+++ b/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
@@ -13,7 +13,9 @@
 
 package tech.pegasys.teku.kzg.ckzg4844;
 
-import ethereum.ckzg4844.CKzg4844JNI;
+import ethereum.ckzg4844.CKZG4844JNI;
+import ethereum.ckzg4844.CKZG4844JNI.Preset;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


infrastructure/kzg/src/test/java/tech/pegasys/teku/kzg/KZGTest.java
```diff
@@ -40,7 +41,17 @@ public final class KZGTest {
   private static final int RANDOM_SEED = 5566;
   private static final Random RND = new Random(RANDOM_SEED);
 
-  private final KZG kzg = CKZG4844.createInstance(FIELD_ELEMENTS_PER_BLOB);
+  private static KZG kzg;
+
+  @BeforeAll
+  public static void setUp() {
+    // test initializing with invalid fieldElementsPerBlob
```
zilm13, line 48: we'd better do it in `@Test` `testCreatingInstanceWithDifferentFieldElementsPerBlob_shouldThrowException`

infrastructure/kzg/src/test/java/tech/pegasys/teku/kzg/KZGTest.java
```diff
@@ -118,12 +129,6 @@ public void testVerifyPointEvaluationPrecompile() {
     final KZGProof kzgProof = new KZGProof(emptyCommitment);
     assertThat(kzg.verifyKzgProof(kzgCommitment, Bytes32.ZERO, Bytes32.ZERO, kzgProof)).isTrue();
     assertThat(kzg.computeAggregateKzgProof(Collections.emptyList())).isEqualTo(kzgProof);
-    assertThat(
```
zilm13, line 121: I don't understand why all tests were removed
Also it would be great if we could test new Minimal Preset somewhere

infrastructure/kzg/src/test/java/tech/pegasys/teku/kzg/KZGTest.java
zilm13: We could have separate Test class only for this case I think without setup, not that bad

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
```diff
@@ -58,8 +61,21 @@ public static CKZG4844 getInstance() {
...
+                new KZGException(
+                    String.format(
+                        "C-KZG-4844 library can't be initialized with %d fieldElementsPerBlob.",
+                        fieldElementsPerBlob)));
+  }
+
+  private CKZG4844(final Preset preset) {
     try {
+      CKZG4844JNI.loadNativeLibrary(preset);
       LOG.debug("Loaded C-KZG-4844");
```
zilm13: we could add preset here

zilm13 (approved):
LGTM

infrastructure/kzg/src/test/java/tech/pegasys/teku/kzg/KZGTest.java
zilm13: For the record: separate test class doesn't unloads native library, so this one remains until we get and idea how to achieve this

PR #6566, opened 2022-12-07
Eip4844 block production + block production refactoring

### LINKED ISSUE #6411: Implement Block Production
https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/validator.md#get_blobs_and_kzg_commitments

Update `BlockOperationSelectorFactory::createSelector` to support block creation with kzg commitments retrieved from EL via `engine_getBlobsBundleV1`

There is also a new dataflow: `engine_getBlobsBundleV1` will return blobs that needs to be stored on db once block production is completed.

We probably need a temporary cache of `BlobsSidecar` that need to be persisted once block is published.

This have an implication on the multiple beacon node endpoints for a VC. The beacon node will become stateful like we currently are when we do the blinded flow when mev\builder is enabled: the beacon node creating the block will cache the blobs associated to the block. Block publication needs to be done by the same beacon node, which is the only one who have the blobs and is able to construct `SignedBeaconBlockAndBlobsSidecar` and  publish it over the network.

- [ ] update BlockOperationSelectorFactory to call engine API `getBlobsBundle`  to provide kzg commitments to block builder
- [ ] cache `BlobsSidecar` using blobs obtained from the engine API `getBlobsBundle` 
- [ ] update block publication to construct `SignedBeaconBlockAndBlobsSidecar` with signed block coming from VC and cached `BlobsSidecar`.
- [ ] Store `BlobsSidecar` on db

zilm13:
## PR Description

Part of #6411 
- Refactored block builder to be compatible with linked fields since EIP-4844, we cannot use suppliers anymore
- Refactored block production routines from execution layer perspective, now we use `ExecutionLayerBlockProductionManager` for this and don't use lower level `ExecutionLayerManagerImpl`
- Refactored block publishing
- Implemented block production for EIP4844
- `BlockAndBlobsSidecar` is gossiped after production

Persistence is done in separate PR.

## Fixed Issue(s)

Fixes #6411 
## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -152,24 +170,90 @@
...
+          () -> {
+            final SchemaDefinitionsEip4844 schemaDefinitionsEip4844 =
+                spec.getGenesisSchemaDefinitions().toVersionEip4844().orElseThrow();
+            final MiscHelpersEip4844 miscHelpers =
+                (MiscHelpersEip4844) spec.forMilestone(SpecMilestone.EIP4844).miscHelpers();
+            return payloadCache
+                .initializer
+                .orElseThrow()
+                .thenCompose(
+                    __ ->
```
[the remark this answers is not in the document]
zilm13: That's why it's called `__`

zilm13:
Current state:
- all logic is in place, but still very dirty, especially `BlockOperationSelectorFactory`
- exception handling is missing
- I'm not sure on the cache place
- more tests should be added

zilm13:
So, we cannot return `ExecutionPayloadResult` from `Channel` interface, only `SafeFuture` and `void`s are handled there by design

tbenr:
@zilm13 I reorganized the publisher to avoid code duplication and renamed the factory class, 'cose I think it isn't an actual factory but it is acting as a factory and delegation at runtime (it actually implements BlockPublisher)

https://github.com/ConsenSys/teku/pull/6690/commits/146ea612cf09478fcd723dd0d63eba71743da7a5

feel free to checkout and cherry-pick it if you like.

teku/build.gradle
```diff
@@ -68,5 +68,6 @@ dependencies {
   testImplementation testFixtures(project(':infrastructure:serviceutils'))
 
   testImplementation project(':data:provider')
+  testImplementation project(':ethereum:executionlayer')
```
tbenr: I was able to compile everything without this. Is it really required?

zilm13: Definitely not, there are no other changes in this module, looks like some uncollected traces after refactoring

tbenr (approved):
LGTM

PR #6568, opened 2022-12-08
Updated voluntaryExitCommand confirmation wording

rolfyone:
Also modified the testing framework to allow for a bit of re-use and to test confirmations.

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
@@ -52,6 +52,7 @@ import tech.pegasys.teku.infrastructure.logging.ValidatorLogger;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.service.serviceutils.layout.DataDirLayout;
 import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.datastructures.operations.VoluntaryExit;
 import tech.pegasys.teku.spec.datastructures.state.Fork;
 import tech.pegasys.teku.spec.datastructures.state.ForkInfo;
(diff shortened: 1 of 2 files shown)


teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -157,21 +165,24 @@ public void run() {
...
     SUB_COMMAND_LOG.display("");
-    SUB_COMMAND_LOG.display(
-        "These validators won't be able to be re-activated, and withdrawals aren't likely to be possible until Phase 2 of eth2 Mainnet.");
+    SUB_COMMAND_LOG.display(WITHDRAWALS_PERMANENT_MESASGE);
+    if (!spec.isMilestoneSupported(SpecMilestone.CAPELLA)) {
+      SUB_COMMAND_LOG.display(WITHDRAWALS_NOT_ACTIVE);
+    }
     SUB_COMMAND_LOG.display("Are you sure you wish to continue (yes/no)? ");
-    Scanner scanner = new Scanner(System.in, Charset.defaultCharset().name());
+    Scanner scanner = new Scanner(System.in, Charset.defaultCharset());
```
ajsutton, line 183: oh good I saw that warning after my last PR and was going to ask you to fix it. :)

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -157,21 +165,24 @@ public void run() {
...
     final String confirmation = scanner.next();
 
     if (!confirmation.equalsIgnoreCase("yes")) {
       SUB_COMMAND_LOG.display("Cancelled sending voluntary exit.");
-      System.exit(1);
+      return false;
```
rolfyone, line 188: if we're testing confirmations, we can't really system.exit here. the result of removing this would be that it'll exit code 0 now, rather than code 1...

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -139,7 +145,9 @@ public void run() {
     try {
       initialise();
       if (confirmationEnabled) {
-        confirmExits();
+        if (!confirmExits()) {
+          return;
```
ajsutton: We lost the return code in this change.  We should be returning 1 here so we get a non-zero exit code.  Class would implement `Callable<Integer>` rather than `Runnable` as well and return 0 where it currently just exits the function without calling System.exit.

ajsutton (approved):
LGTM we just need to preserve the exit code when the user aborts (can probably update the test to check it now too).

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
rolfyone: :) I figured may as well leave it cleaner than what it was since i was there.

Issue #6406, opened 2022-11-09
beacon-apis add finalized flag to some responses


rolfyone:
https://github.com/ethereum/beacon-APIs/pull/254

PR #6569, opened 2022-12-08
Add finalized to ObjectAndMetaData

ajsutton:
## PR Description
This is the first step to making it available via the API. It was big enough it seemed worth doing the actual change of response types in a follow up PR.

## Fixed Issue(s)
#6406 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/metadata/ObjectAndMetaData.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/metadata/ObjectAndMetaData.java
@@ -60,6 +60,10 @@ public class ObjectAndMetaData<T> {
     return canonical;
   }
 
+  public boolean isFinalized() {
+    return finalized;
+  }
+
   @Override
   public boolean equals(final Object o) {
     if (this == o) {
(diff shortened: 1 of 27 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/metadata/ObjectAndMetaData.java
```diff
@@ -24,20 +24,24 @@
...
   private final boolean canonical;
+  private final boolean finalized;
 
   public ObjectAndMetaData(
       final T data,
       final SpecMilestone milestone,
       final boolean executionOptimistic,
-      final boolean canonical) {
+      final boolean canonical,
+      final boolean finalized) {
```
rolfyone, line 34: eep hopefully we don't get too many magic fields, starting to read a little random, but probably not worth a builder yet...

ajsutton: Yeah it's a bit hairy but I don't think a builder makes sense just yet given how much more verbose the construction would be and the fact that we don't actually create these in too many places in production code.

rolfyone (approved):
LGTM

PR #6573, opened 2022-12-08
Updated bls_to_execution_changes to use array processing

### LINKED ISSUE #6553: update bls_to_execution_change to allow arrays
https://github.com/ethereum/beacon-APIs/pull/275

Arguably we could be a little smart and allow a single OR array, but that may require a little trickery... it might be easier to just change all the associated areas to require the array to be added.

rolfyone:
POST bls_to_execution_changes has been updated to use array processing so that the UX is better for users that need to submit more than one change, and this will be more consistent with tooling currently available to be used.

fixes #6553

## Documentation

## Changelog


ajsutton (approved):
LGTM.

PR #6577, opened 2022-12-09
Implement GetLightClientBootstrap

robzajac:
## PR Description
Implements and tests [`GetLightClientBootstrap`](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/full-node.md#create_light_client_bootstrap), an API for a caller to initiate the [light client sync protocol](https://mycelium.xyz/research/world-of-light-clients-ethereum) against a given trusted checkpoint root.

Business logic is primarily in `LightClientUtil`, and the remainder is largely plumbing.

Followups include:
- Implementing [`calculate_merkle_root`](https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md#merkle-multiproofs) to verify the proofs created in our test cases
- Hiding the API behind a `--light-client-support-enabled` flag (nodes have the option of serving this data or not)

## Fixed Issue(s)

Partially addresses #4230

## Documentation

## Changelog


data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/beacon/GetLightClientBootstrapIntegrationTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+  final Bytes32 blockRoot = Bytes32.random();
+
+  @BeforeEach
+  void setup() {
+    startRestAPIAtGenesis(SpecMilestone.ALTAIR);
+  }
+
+  @Test
+  void shouldReturnResultIfCreatedSuccessfully() throws IOException {
```
robzajac, line 49: This is my attempt at an E2E test... would be open to suggestions on further testing. FWIW, the API functions correctly on local teku (minimal network).

ajsutton: I think this is good for this PR. As a follow up we could probably write an acceptance test that starts a real chain and then gets the light client bootstrap and checks its content matches the real data from that slot that we get from the `/eth/v2/debug/beacon/states/{state_id}` API. That would test the entire setup like it runs in the real world.

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -487,6 +489,29 @@ public boolean stateParameterMaySupportAltair(final String epochParam) {
...
+                maybeStateData.map(
+                    stateAndMetadata ->
+                        stateAndMetadata.map(
+                            state -> {
+                              Optional<LightClientUtil> lightClientUtil =
+                                  spec.getLightClientUtil(state.getSlot());
+
+                              if (lightClientUtil.isEmpty()) {
+                                // No light client bootstrap pre-altair
+                                return null;
```
ajsutton: hmm, I don't think this is going to do what we want because this is the `.map` function from `ObjectAndMetaData` which always expects to contain a value.  So I think pre-altair this would wind up returning a preset Optional containing an `ObjectAndMetaData` that holds a `null` rather than returning an empty `Optional`. I think we need to reverse the maps on getLightClientUtil and stateAndMetadata to:

```java
 public SafeFuture<Optional<ObjectAndMetaData<LightClientBootstrap>>> getLightClientBoostrap(
      final Bytes32 blockRootParam) {
    return defaultStateSelectorFactory
        .forBlockRoot(blockRootParam)
        .getState()
        .thenApply(
            maybeStateData ->
                maybeStateData.flatMap(
                    stateAndMetadata ->
                        spec.getLightClientUtil(stateAndMetadata.getData().getSlot())
                            .map(
                                clientUtil ->
                                    stateAndMetadata.map(clientUtil::getLightClientBootstrap))));
  }
```

But that's ugly as sin... So I think I'd extract a method to get:

```java
  public SafeFuture<Optional<ObjectAndMetaData<LightClientBootstrap>>> getLightClientBoostrap(
      final Bytes32 blockRootParam) {
    return defaultStateSelectorFactory
        .forBlockRoot(blockRootParam)
        .getState()
        .thenApply(maybeStateData -> maybeStateData.flatMap(this::getLightClientBootstrap));
  }

  private Optional<ObjectAndMetaData<LightClientBootstrap>> getLightClientBootstrap(
      final StateAndMetaData stateAndMetadata) {
    return spec.getLightClientUtil(stateAndMetadata.getData().getSlot())
        .map(clientUtil -> stateAndMetadata.map(clientUtil::getLightClientBootstrap));
  }
```

data/serializer/src/main/java/tech/pegasys/teku/api/response/v1/beacon/GetLightClientBootstrapResponse.java
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.api.response.v1.beacon;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import tech.pegasys.teku.api.schema.altair.LightClientBootstrap;
+
+public class GetLightClientBootstrapResponse {
```
ajsutton: We don't need to define these reflection based APIs anymore as the REST API has entirely moved over to using the `JsonTypeDefinition` approach. There are a few left around because they're still used by the validator client and things like that but we're working to remove them.

So we can move this to integrationTest which is the only place it's used or better we can just use type definitions in the integration test too. I'll push a commit that adjusts the test since it needs to add a couple of getter methods in the `LightClientBootstrap` class as well. Feel free to revert/edit/whatever that though if you disagree.

data/serializer/src/main/java/tech/pegasys/teku/api/schema/altair/LightClientBootstrap.java
```diff
@@ -0,0 +1,67 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.v3.oas.annotations.media.ArraySchema;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+import java.util.Objects;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.api.schema.BeaconBlockHeader;
+
+public class LightClientBootstrap {
```
ajsutton: We'll be able to get rid of this if we change the integration test as well.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/LightClientUtil.java
```diff
@@ -0,0 +1,83 @@
...
+
+    // Requires rehashing the state. See
+    // https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/full-node.md#create_light_client_bootstrap
+    final BeaconBlockHeader bootstrapBlockHeader =
+        new BeaconBlockHeader(
+            latestBlockHeader.getSlot(),
+            latestBlockHeader.getProposerIndex(),
+            latestBlockHeader.getParentRoot(),
+            state.hashTreeRoot(),
+            latestBlockHeader.getBodyRoot());
```
ajsutton: We can use `BeaconBlockHeader.fromState` to do this.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/LightClientUtil.java
```diff
@@ -0,0 +1,83 @@
...
+            state
+                .getSchema()
+                .getChildGeneralizedIndex(
+                    state.getSchema().getFieldIndex(BeaconStateFields.CURRENT_SYNC_COMMITTEE)));
+
+    final SszBytes32Vector currentSyncCommitteeProofSsz =
+        SszBytes32VectorSchema.create(currentSyncCommitteeProof.size())
+            .createFromElements(
+                currentSyncCommitteeProof.stream()
+                    .map(bytes32 -> SszBytes32.of(bytes32))
```
ajsutton: nit:
```suggestion
                    .map(SszBytes32::of)
```

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/LightClientUtil.java
```diff
@@ -0,0 +1,83 @@
...
+    final SyncCommittee currentSyncCommittee =
+        syncCommitteeUtil.getSyncCommittee(state, currentEpoch);
+
+    final List<Bytes32> currentSyncCommitteeProof =
+        MerkleUtil.constructMerkleProof(
+            state.getBackingNode(),
+            state
+                .getSchema()
+                .getChildGeneralizedIndex(
+                    state.getSchema().getFieldIndex(BeaconStateFields.CURRENT_SYNC_COMMITTEE)));
```
ajsutton: hmm, I'm not a huge fan of reaching into the schema like this and having separate code that depends so heavily on the internal representation of the state.  I think I'd be tempted to make this:

```java
    final List<Bytes32> currentSyncCommitteeProof =
        BeaconStateAltair.required(state).createCurrentSyncCommitteeProof();
```

And then add to `BeaconStateAltair`:
```java
  default List<Bytes32> createCurrentSyncCommitteeProof() {
    return MerkleUtil.constructMerkleProof(
        getBackingNode(),
        getSchema()
            .getChildGeneralizedIndex(
                getSchema().getFieldIndex(BeaconStateFields.CURRENT_SYNC_COMMITTEE)));
  }
```

Same code but `BeaconStateAltair` is already dealing with the tree representation and field indexes etc so it's nice to keep it encapsulated in those actual state implementation classes.

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/util/LightClientUtilTest.java
```diff
@@ -0,0 +1,48 @@
...
+        new BeaconBlockHeader(
+            latestBlockHeader.getSlot(),
+            latestBlockHeader.getProposerIndex(),
+            latestBlockHeader.getParentRoot(),
+            state.hashTreeRoot(),
+            latestBlockHeader.getBodyRoot());
+    LightClientBootstrap bootstrap = lightClientUtil.getLightClientBootstrap(state);
+
+    assertThat(bootstrap.get(0)).isEqualTo(expectedHeader);
+    assertThat(bootstrap.get(1)).isEqualTo(state.toVersionAltair().get().getCurrentSyncCommittee());
```
ajsutton:
```suggestion
    assertThat(bootstrap.get(1)).isEqualTo(BeaconStateAltair.required(state).getCurrentSyncCommittee());
```

data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -552,7 +638,9 @@ public static Stream<Arguments> getRandaoIndexCases() {
     return args.stream();
   }
 
-  private ChainDataProvider setupAltairState() {
+  private Pair<
+          ChainDataProvider, tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState>
+      setupAltairState() {
```
ajsutton: I think I'd stick with just returning the `ChainDataProvider` from here but add a method to get the internal state:

```java

  private tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState getHeadState() {
    return safeJoin(mockCombinedChainDataClient.getChainHead().orElseThrow().getState());
  }
```

It's still not really pretty but we generally avoid using generic `Pair` because it gets pretty confusing about what's actually in it given you only get generic `getLeft` and `getRight` type methods. And most places here only need the ChainDataProvider by itself.

data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -376,13 +379,96 @@ public void getStateSyncCommittees_shouldReturnEmptyListBeforeAltair() {
...
+        data.randomBeaconState();
+
+    BeaconBlockHeader latestBlockHeader = internalState.getLatestBlockHeader();
+    BeaconBlockHeader expectedBlockHeader =
+        new BeaconBlockHeader(
+            latestBlockHeader.getSlot(),
+            latestBlockHeader.getProposerIndex(),
+            latestBlockHeader.getParentRoot(),
+            internalState.hashTreeRoot(),
+            latestBlockHeader.getBodyRoot());
```
ajsutton:
```suggestion
    BeaconBlockHeader expectedBlockHeader = BeaconBlockHeader.fromState(internalState);
```

data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -376,13 +379,96 @@ public void getStateSyncCommittees_shouldReturnEmptyListBeforeAltair() {
...
+        altairState.getRight();
+
+    BeaconBlockHeader latestBlockHeader = internalState.getLatestBlockHeader();
+    BeaconBlockHeader expectedBlockHeader =
+        new BeaconBlockHeader(
+            latestBlockHeader.getSlot(),
+            latestBlockHeader.getProposerIndex(),
+            latestBlockHeader.getParentRoot(),
+            internalState.hashTreeRoot(),
+            latestBlockHeader.getBodyRoot());
```
ajsutton:
```suggestion
    BeaconBlockHeader expectedBlockHeader = BeaconBlockHeader.fromState(internalState);
```

data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -376,13 +379,96 @@ public void getStateSyncCommittees_shouldReturnEmptyListBeforeAltair() {
...
```
ajsutton:
```suggestion
    BeaconBlockHeader expectedBlockHeader = BeaconBlockHeader.fromState(internalState);
```

ajsutton (commented):
Looking good, mostly just a bunch of suggestions to simplify/tidy stuff up.

data/serializer/src/main/java/tech/pegasys/teku/api/response/v1/beacon/GetLightClientBootstrapResponse.java
robzajac: Thanks @ajsutton for the commit, I didn't see that we could parse the API response with the SSZ definition. In this case I think it's definitely cleaner to use type definitions. I'll cleanup and address comments in the upcoming day.

robzajac:
@ajsutton thanks for review, I have pushed up another round.

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/util/LightClientUtilTest.java
```diff
@@ -0,0 +1,43 @@
...
+  private final LightClientUtil lightClientUtil = spec.getLightClientUtilRequired(UInt64.ZERO);
+
+  @Test
+  public void getBoostrap_shouldReturnValidBootstrap() {
+    BeaconState state = dataStructureUtil.stateBuilderAltair().build();
+    BeaconBlockHeader expectedHeader = BeaconBlockHeader.fromState(state);
+    LightClientBootstrap bootstrap = lightClientUtil.getLightClientBootstrap(state);
+
+    assertThat(bootstrap.get(0)).isEqualTo(expectedHeader);
+    assertThat(bootstrap.get(1))
```
ajsutton: I'll push a minor tweak to change this to use the getter methods rather than the raw `get(int)` method.

ajsutton (approved):
LGTM just the one minor tweak that I'll push up so I can get this merged.  Thanks.

PR #6578, opened 2022-12-09
Use intValue method for uint64 to int conversions

jtraglia:
## PR Description

Replace `toIntExact(u64.longValue())` with `u64.intValue()`.

This is just a little nit I noticed when reading over these files. It helps readability and it will fail a bit faster if there's a problem. To be clear, there is nothing wrong with the current way things are done.

These aren't equivalent. The difference is that `intValue()` will throw an `ArithmeticException` if the value is negative, whereas `toIntExact` will not. None of these values should be negative, so I think that's better.

https://github.com/ConsenSys/teku/blob/b0b100f786b4f21a3591bfa5e8cfd0a16df722dd/infrastructure/unsigned/src/main/java/tech/pegasys/teku/infrastructure/unsigned/UInt64.java#L466-L472

Also, I added `ArithmeticException` to the handled exceptions in `safelyProcess`. I feel like this should have been there before, but maybe I'm misunderstanding something and this doesn't matter. `toIntExact` throws that too.

## Documentation

## Changelog


ajsutton (approved):
LGTM. I suspect all these cases predate our switch to using our own `UInt64` - the Tuweni  version we originally used didn't have an `intValue` method. Good to clean it up and definitely safer.

The `ArithmeticException` being in the list is a good change - it makes the rejection of the block a lot less noisy. Exceptions not in that list are treated as unexpected failures when importing the block. The block still gets rejected but with a full stack trace and big scary warning text.

PR #6589, opened 2022-12-12
Update FuzzUtil to support Capella

jtraglia:
## PR Description

This PR updates the fuzzing utilities (`FuzzUtil`) to the Capella milestone.

* Replace `Bellatrix` with `Capella` everywhere.
* Replace pyspec test files with Capella versions.
* Add new `fuzzBlsToExecutionChange` method.

### Side note

In my opinion, it's a little odd using `ExecutionPayloadCapellaImpl` instead of `ExecutionPayloadCapella`.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/SignedBlsToExecutionChange.java
```diff
@@ -35,6 +35,9 @@
     super(type, backingNode);
   }
 
+  public static final SignedBlsToExecutionChangeSchema SSZ_SCHEMA =
+      new SignedBlsToExecutionChangeSchema();
```
ajsutton: We need to get this from schema definitions - the old pattern was to have a constant for schemas, but that doesn't support data structures changing with hard forks so we've moved away from it to everything coming from `SchemaDefinitions` which is milestone aware.

ajsutton: I'll push a commit to change this - easier than explaining it in a comment. :)

fuzz/src/main/java/tech/pegasys/teku/fuzz/input/ExecutionPayloadFuzzInput.java
```diff
@@ -19,36 +19,39 @@
...
-import tech.pegasys.teku.spec.datastructures.execution.versions.bellatrix.ExecutionPayloadBellatrix;
-import tech.pegasys.teku.spec.datastructures.execution.versions.bellatrix.ExecutionPayloadSchemaBellatrix;
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.capella.BeaconBlockBodySchemaCapella;
+import tech.pegasys.teku.spec.datastructures.execution.versions.capella.ExecutionPayloadCapellaImpl;
+import tech.pegasys.teku.spec.datastructures.execution.versions.capella.ExecutionPayloadSchemaCapella;
 import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
 
 public class ExecutionPayloadFuzzInput
-    extends Container2<ExecutionPayloadFuzzInput, BeaconState, ExecutionPayloadBellatrix> {
+    extends Container2<ExecutionPayloadFuzzInput, BeaconState, ExecutionPayloadCapellaImpl> {
```
ajsutton, line 28: ah the need to use `ExecutionPayloadCapellaImpl` here is because it's referring to the concrete implementation type. There's a bit of a battle going on with Java's type system to be honest...

`ExecutionPayloadBellatrix` is also the implementation version, just that the interface in that case is `ExecutionPayload` because it's the first version of it.  The inconsistency isn't great but I do like the simple `ExecutionPayload` name and you don't see the implementation classes in many places...

ajsutton (approved):
LGTM.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/SignedBlsToExecutionChange.java
jtraglia: I understand. Thanks! 🙏

Issue #6576, opened 2022-12-09
Support pruning old blocks


ajsutton:
### Description
Now that checkpoint sync is supported by all clients, we can start moving away from storing the full block history back to genesis.

Add a (initially hidden, --X) option to specify the maximum retention period for historic block data. This should be specified in number of epochs to retain.  The minimum allowed value is [`MIN_EPOCHS_FOR_BLOCK_REQUESTS`](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#configuration). If the value is set to less than this, log a warning and use `MIN_EPOCHS_FOR_BLOCK_REQUESTS` as the retention period.

If the retention period is set to 0, retain all blocks back to genesis.  Q: Should we instead have a separate `--block-pruning-enabled` option to disable?

Also ensure that the historic block fetcher never attempts to fetch blocks prior to this retention period. It will also not be possible to rebuild historic states when block data is pruned (since it requires blocks all the way back to genesis).

benjaminion:
Would we auto-prune existing databases that have existing data back to genesis? (2 years' worth now.) Or would this apply only to newly-checkpoint-synced nodes?

I can see arguments for both, but we should choose by default whichever is least surprising to users - probably the latter. Maybe supply a tool option to prune old DBs. (Users could manually delete an old DB and do a checkpoint sync, but this comes at the cost of re-downloading 5 months worth of blocks, which is non-trivial.)

ajsutton:
Good question - ideally for existing databases the current behaviour of keeping all blocks back to genesis would be preserved so there are no surprises.  But I think we want to make it easy to enable this and just have it prune old data in the background rather than having to run a separate tool. So I think we'd just want to make the default enabled/disabled value dependent on the database so it defaults to disabled for any database created before this feature that hasn't enabled it at some point.

But the initial implementation would be disabled by default and I think we'd leave this as an opt-in feature for a fair while anyway.

ajsutton:
For initial implementation I've wound up with just a block-pruning-enabled flag and no option to customise the number of epochs it keeps. We could make it configurable if needed but it added complexity and given the spec requires about 5 months of storage, I'm not really sure why you'd want somewhere between that and everything. Maybe tax reporting but seems very very niche...

ajsutton:
Initial implementation done, there's a few remaining things to get to done:

[...]

PR #6590, opened 2022-12-12
Use new MINIMAL storage mode for pruning blocks

ajsutton:
## PR Description
Rather than using a separate flag, introduce `--data-storage-mode=MINIMAL` to prune both blocks and states.  Fits better with the CLI and there are then fewer ways the config can be conflicting.

Also changed from using switch statements to having more context dependent methods on storage mode like `storesFinalizedStates()` so we don't have to track down every switch statement when a new mode is introduced.

Since this is a new value for an existing option it's no longer a hidden toggle. It's still not the default though and has been working reliably.

## Fixed Issue(s)
#6576 

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

PR #6593, opened 2022-12-13
Limit number of validators checked during withdrawals lookup

### LINKED ISSUE #6592: Limit number of validators checked during withdrawals lookup
Based on https://github.com/ethereum/consensus-specs/pull/3095

lucassaldanha:
## PR Description

- Adding MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP to presets files and Capella SpecConfig
- Updated get_expected_withdrawals and process_withdrawals to use MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP

## Fixed Issue(s)

fixes #6592

## Documentation

## Changelog


ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/presets/minimal/capella.yaml
```diff
@@ -11,4 +11,9 @@ MAX_BLS_TO_EXECUTION_CHANGES: 16
...
 # ---------------------------------------------------------------
 # 2**2 (= 4)
-MAX_WITHDRAWALS_PER_PAYLOAD: 4
\ No newline at end of file
+MAX_WITHDRAWALS_PER_PAYLOAD: 4
+
+# Withdrawals processing
+# ---------------------------------------------------------------
+# 2**14 (= 16384) validators
+MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP: 16384
```
ajsutton: Minimal should set the max to 16.  I'd suggest setting to 16 for swift as well since it usually follows minimal spec.

ethereum/spec/src/test/resources/tech/pegasys/teku/spec/config/standard/presets/minimal/capella.yaml
```diff
@@ -9,4 +9,9 @@ MAX_BLS_TO_EXECUTION_CHANGES: 16
...
 # ---------------------------------------------------------------
 # 2**2 (= 4)
-MAX_WITHDRAWALS_PER_PAYLOAD: 4
\ No newline at end of file
+MAX_WITHDRAWALS_PER_PAYLOAD: 4
+
+# Withdrawals processing
+# ---------------------------------------------------------------
+# 2**14 (= 16384) validators
+MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP: 16384
```
ajsutton, line 17: Should be 16 here too.

ajsutton (approved):
LGTM. Just need to get the minimal config fixed.

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/presets/minimal/capella.yaml
lucassaldanha: nice catch!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/capella/block/BlockProcessorCapella.java
```diff
@@ -192,34 +193,57 @@ public void processWithdrawals(
...
+    assertWithdrawalsInExecutionPayloadMatchExpected(payloadSummary, expectedWithdrawals);
+
     for (int i = 0; i < expectedWithdrawals.size(); i++) {
       final Withdrawal withdrawal = expectedWithdrawals.get(i);
       beaconStateMutators.decreaseBalance(
           state, withdrawal.getValidatorIndex().intValue(), withdrawal.getAmount());
     }
 
-    if (expectedWithdrawals.size() > 0) {
+    if (expectedWithdrawals.size() != 0) {
```
rolfyone, line 208: this is where we'll need to be careful... 0 is definitely possible in capella...

I'll have to recheck the spec change, but my expectation is that 
 - if size == max, next = last_found+1
 - otherwise next increments prestate.next + maxValidatorsPerSweep

ajsutton: You were right here - this if structure didn't match the spec correctly and we wound up not calling `setNextWithdrawalValidatorIndex` on the state if there were no withdrawals.  The condition was right, but it should have only been around the `setNextWithdrawalIndex` part not the whole rest of the method.

Will push up a fix in a bit.

lucassaldanha: Nice catch Paul!!

PR #6597, opened 2022-12-13
Batch Block Pruning

ajsutton:
## PR Description
Be defensive and prune blocks in batches rather than always using a single transaction to delete all pruneable blocks.

## Fixed Issue(s)
#6582 

## Documentation

## Changelog


tbenr (approved):
LGTM

tbenr:
Have you done a test on a fully synced mainnet node? How much does it take to complete the pruning?

ajsutton:
hmm, actually very slow to prune because it has to load each block to prune and calculate its hash tree root. Will have to poke around and see what can be done about that but I'm not sure we have a lot of options.

ajsutton:
Ok, doing about a million deletions in 20 minutes. That's not terrible - it's hitting the CPU harder than I'd like though - basically using a CPU for pruning (but it's single threaded so will never use more than one).

PR #6603, opened 2022-12-14
Fix kzg computing proof segfault and update trusted setup

StefanBratanov:
## PR Description

- Updated trusted setup with one from the specs
- Added `testComputingProofWithZeroLengthBlobs` test. It was failing with segfault before the change and it was working after the change)
- Removed `verifyKzgProof` (only used in EL)

## Fixed Issue(s)
helps with #6596 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #6605, opened 2022-12-14
Move property test suppliers to testFixtures

jtraglia:
## PR Description

I want to move these suppliers/utilities to the test fixtures source tree for these reasons:

* The ability to re-use these classes in other components' property tests.
* The split cleans up the property test source tree, half as many files.

## Documentation

## Changelog


ajsutton (approved):
LGTM and makes sense to have them in testFixtures.

PR #6606, opened 2022-12-15
Optimise finalized block pruning

### LINKED ISSUE #6576: Support pruning old blocks

ajsutton:
## PR Description
Adjusts how finalized blocks are pruned in the unblinded database format. We now scan over the full column of blockRoot->slot entries in each batch because while that reads more data, it avoids needing to recalculate the block root for each block to be deleted.  This makes the initial pruning significantly faster and after that the number of entires is limited so scanning the full column is straight forward.

It turns out to still be important to do the deletion in batches or LevelDB chokes on having that much very old data be deleted and the whole node fails to stay in sync for quite a long period while to sorts itself out. 

Switching a fully sync'd mainnet node from `prune` to `minimal` with this version took about 80 minutes or a stock EC2 non-attached storage (not particularly fast). While CPU usage was higher during that period there were no visible impacts on the node performance, block import time etc.

## Fixed Issue(s)
fixes #6576 

## Documentation

## Changelog


tbenr (approved):
LGTM. nice!

PR #6610, opened 2022-12-15
DG Detection Tuning

mehdi-aouadi:
## PR Description
- Increased the DG check delay from 5s to 12s
- During the first check we now check the previous and the current epoch instead of starting from the current epoch (Could enhance performance since the validators could be already seen during the previous epoch)
- The DG check doesn't stop when a first DG is detected. It now continues checking the inactive keys. The DG check stops when it detects all the keys as active OR when it runs for 2 epochs
- Refactored the DG detection action
- Updated logs: Whenever a DG is detected a log is printed. When the DG finishes, a log with all the detected DGs is printed

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
```diff
@@ -181,9 +184,14 @@ private SafeFuture<Void> performDoppelgangerCheck(final Set<BLSPublicKey> pubKey
                       mapToAbbreviatedKeys(pubKeys).collect(Collectors.toSet()));
                   return stopDoppelgangerDetectorTask(new HashMap<>());
                 } else {
-                  final UInt64 previousEpoch =
-                      currentEpoch.isZero() ? currentEpoch : currentEpoch.minus(UInt64.ONE);
-                  return checkDoppelgangersAtEpoch(pubKeys, currentSlot, previousEpoch);
+                  if (firstCheck.get()) {
+                    firstCheck.set(false);
```
rolfyone: needs to be an atomic check and set - `compareAndSet`

mehdi-aouadi: Done

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/DoppelgangerDetectorAcceptanceTest.java
```diff
@@ -107,11 +107,84 @@ void shouldDetectDoppelgangerAndAlert() throws Exception {
 
     secondValidatorClient.waitForLogMessageContaining("Validator doppelganger detected...");
     secondValidatorClient.waitForLogMessageContaining("Detected 8 validators doppelganger");
+    secondValidatorClient.waitForLogMessageContaining("Doppelganger detection check finished");
+    secondValidatorClient.waitForLogMessageContaining("Detected 8 validators doppelganger");
```
zilm13, line 110: why this one is checked again?
why it's not shutdown in this case (or no check)

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -176,17 +176,22 @@ public void doppelgangerDetectionTimeout(final Set<String> publicKeys) {
...
   }
 
-  public void doppelgangerDetectionEnd(final Set<String> publicKeys) {
+  public void doppelgangerDetectionEnd(
+      final Set<String> publicKeys, final Map<UInt64, String> doppelgangersInfo) {
     log.info(
-        "No validators doppelganger detected after 2 epochs. Stopping doppelganger detection for public keys {}.",
+        "Doppelganger detection check finished. Stopping doppelganger detection for public keys {}.",
         String.join(", ", publicKeys));
+    if (!doppelgangersInfo.isEmpty()) {
```
zilm13: I'd pull it upside down for readability
```
if (doppelgangersInfo.isEmpty()) {
  log.info("No validators doppelganger detected.");
} else {
  validatorsDoppelgangersDetected(doppelgangersInfo);
}
```

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -70,7 +70,7 @@
 public class ValidatorClientService extends Service {
 
   private static final Logger LOG = LogManager.getLogger();
-  private static final Duration DOPPELGANGER_DETECTOR_CHECK_DELAY = Duration.ofSeconds(5);
+  private static final Duration DOPPELGANGER_DETECTOR_CHECK_DELAY = Duration.ofSeconds(12);
```
zilm13, line 73: Is it any sense to tie it with slot duration? (remember gnosis have 5 seconds slots)

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/DoppelgangerDetectorAcceptanceTest.java
mehdi-aouadi: In the first iteration the behaviour was to stop the DG detection whenever we detect at least one active key but now it continues running until:
All the keys are detected as active OR it has been running for 2 epochs.

Keep in mind that there are two DG detection actions: 1. Shutdown (Triggered at the VC startup) 2. Alert (Triggered when adding keys via the Key-Manager API)
In the first case (shutdown at VC startup) it could makes sense to stop the DG check at the first DG detection, but that means you're not sure it the other keys are active or not.
In the second case (alert, Key-Manager API), the detected active keys are ignored and the other ones are imported. If we stop the DG at the first active detection that means we could import keys that could be active.
That's why we implemented this change: Do not stop after the first detection, keep running for at least 2 epochs...

A log is printed whenever a DG is detected AND another log is printed at the end of the DG check with a full report of ALL the detected DGs. In this acceptance test, there are 2 DG detected at the first round: 1. First log when they're first detected 2. Second log when the check is finished

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
mehdi-aouadi: It's actually a tradeoff between DG detection speed and effectiveness.
5s is too short for 12s slots since it doesn't really make sense to check after at least one slot since we're almost sure there have been no new attestations and hence no eventual active validator caching.
12s is more suitable for both 12s slots and shorter slots...

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/DoppelgangerDetectorAcceptanceTest.java
zilm13: Getting it. So we confirmed detection on the 2nd epoch, what's next, we are going to shutdown? If yes, we should check it

mehdi-aouadi: In this use case (Loading keys via the Key Manager API) there should be an alert and the keys are ignored (not imported). I added a check for that too (`Detected 8 active validators doppelganger. The following keys have been ignored`). Good catch

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
```diff
@@ -255,6 +285,22 @@ private SafeFuture<Void> checkValidatorsLivenessAtEpoch(
           .thenApply(doppelgangerDetected -> null);
     }
 
+    private void logMissingIndices(
+        Set<BLSPublicKey> pubKeys, Map<BLSPublicKey, UInt64> validatorIndicesByPubKey) {
```
zilm13: Finals please. Totally over-complication for this method, but let it avoid catching a sight

zilm13 (approved):
LGTM
Great work!

validator/client/src/main/java/tech/pegasys/teku/validator/client/doppelganger/DoppelgangerDetector.java
mehdi-aouadi: Done

Issue #6612, opened 2022-12-15
Support locally built block value field


ajsutton:
### Description
As per https://github.com/ethereum/execution-apis/pull/314/ need to support the new field.

Will need to not fail if that field is present for next withdrawals testnet, then should actually make use of the data to use the local block if it's worth more than one offered from the builder at some point.

ajsutton:
The new syntax is now supported will leave actually using the value to https://github.com/ConsenSys/teku/issues/6327 so closing this one out.

PR #6614, opened 2022-12-15
Add blockValue to get payload V2 response

ajsutton:
## PR Description
Update expected response from get payload V2 to include the block value as per https://github.com/ethereum/execution-apis/pull/314/

Note that the value isn't currently used anywhere - just updated so that we can handle the new format.

To smooth the transition for any existing testnets, the old version without block value is supported by specifying `--Xee-version=NO_BLOCK_VALUE`.

## Fixed Issue(s)
#6612 

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #6558, opened 2022-12-06
Remove blinded block storage


ajsutton:
### Description
The attempt to store blinded blocks with separate execution payloads hasn't panned out like we hoped with reduced performance. We should remove that code to simplify the codebase.

PR #6616, opened 2022-12-16
Update database compatibility files to version 22.8.2

ajsutton:
## PR Description
This is the first version that supported Bellatrix on MainNet so all users must have updated to it and it includes the new columns added by blinded block support. Removing columns breaks compatibility so we need to ensure those columns aren't removed.

## Fixed Issue(s)
#6558 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #6618, opened 2022-12-16
Fix BlsToExecutionChangeAcceptanceTest intermittency

ajsutton:
## PR Description
Sets a long timeout for the event listener connection so that it doesn't get disconnected before an event actually arrives.

Also improves some error messages and logging to make debugging easier in the future.

## Fixed Issue(s)
fixes #6611 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #6619, opened 2022-12-16
Fix gradle dependency cache in CI

ajsutton:
## PR Description
The key used to save the cache didn't match the one being used to restore it so caching wasn't working.

## Documentation

## Changelog


ajsutton:
Wound up being merged as part of #6618

PR #6623, opened 2022-12-16
Implement API stub of GetLightClientUpdatesByRange

robzajac:
## PR Description
Implements a stub of [`GetLightClientUpdatesByRange`](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/updates.yaml) with `return 501;`

- Adds endpoint implementation and JSON reference
- Adds `LightClientUpdate` SSZ types

Note the `application/octet-stream` return type is currently missing, as it will require implementing a serialization [specific to this endpoint](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/updates.yaml#L37) that I prefer deferring to a followup PR.

## Fixed Issue(s)

partially address #4230 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
```diff
@@ -104,6 +106,15 @@ public class BeaconRestApiTypes {
...
 
+  public static final ParameterMetadata<UInt64> START_PERIOD_PARAMETER =
+      new ParameterMetadata<>(
+          START_PERIOD,
+          CoreTypes.UINT64_TYPE.withDescription("`uint64` value representing start_period"));
+
+  public static final ParameterMetadata<UInt64> COUNT_PARAMETER =
+      new ParameterMetadata<>(
+          COUNT, CoreTypes.UINT64_TYPE.withDescription("`uint64` value representing count"));
+
```
rolfyone, line 117: We should probably size constrain these for sanity... I'm not sure if the spec has anything in it about acceptable ranges....
In testing we can play with both being MAX_VALUE and watching it go boom...
I've got no firm ideas about where we constrain them, just a future idea for something we will need to look at...

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/paths/_eth_v1_beacon_light_client_updates.json
```diff
@@ -0,0 +1,96 @@
...
+                    "enum": [
+                      "phase0",
+                      "altair",
+                      "bellatrix",
+                      "capella"
+                    ]
+                  },
+                  "data": {
+                    "$ref": "#/components/schemas/LightClientUpdate"
+                  }
```
rolfyone, line 51: something's not right here in the structure... we'd have
`[{version:..., data:...}, {version:..., data:...}`
and we need 
```
{
version: ...,
data: [{...}]
}
```

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/paths/_eth_v1_beacon_light_client_updates.json
```diff
@@ -0,0 +1,96 @@
...
+                }
+              }
+            }
+          },
```
rolfyone, line 55: The structure here will be inconsistent with spec...
we've basically got 
```
[
{
 version: ...,
data: {...}
}
]
```
which i think the reference json shows too...

we actually need:
```
{
version: ...
data: [ {...}, {...}]
}
```

https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getLightClientUpdatesByRange
Screenshot 2022-12-18 at 7 31 08 am

rolfyone:
looks good, we'll need to circle back on those query parameters and find what limits exist in the spec... we can probably get away with something sane rather than potentially get arithmetic exceptions that we have to handle...

rolfyone (approved):
The return structure is weird, but it does appear consistent with the api spec... It's going to be different to everything else in the api so we'll have to keep an eye on how it evolves.

rolfyone:
looks like the paths OpenApiIntegrationTest isn't quite right...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
robzajac: Agreed. According to the [reference](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/full-node.md#create_light_client_update),

> Full nodes SHOULD provide the best derivable LightClientUpdate (according to is_better_update) for each sync committee period covering any epochs in range [max(ALTAIR_FORK_EPOCH, current_epoch - MIN_EPOCHS_FOR_BLOCK_REQUESTS), current_epoch] where current_epoch is defined by the current wall-clock time. Full nodes MAY also provide LightClientUpdate for other sync committee periods.

In other words, we could constrain it according to the range of valid epochs.

robzajac:
> The return structure is weird, but it does appear consistent with the api spec... It's going to be different to everything else in the api so we'll have to keep an eye on how it evolves.

Agreed @rolfyone, and it's additionally complicated by the [custom encoding for the `octet-stream` response on this endpoint](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/updates.yaml#L37). @ajsutton previously suggested that this may be to handle if the response spans multiple hard forks (i.e. the light client updates are from different forks). Still I can't find another endpoint with this return type, so I will probably check in with the spec authors to understand better.

PR #6628, opened 2022-12-19
[EIP-4844] Blobs pruning

tbenr:
Blocks and Blobs pruning work slightly different.

Blocks jobs run every 1h and always prune all "prunable" blocks (that's why the first time you switch MINIMAL on, takes a while to complete).

Blobs pruning are designed to run frequently and with an hard cap on the `BlobsSidecar`s that can be pruned. Currently configured to run every minute and prune maximum of 32 blobs. Under finalization, with mainnet config, we are supposed to create 5 BlobsSidecar each minute, so 32 is able to catch up relatively quickly.

The idea is to avoid those pruning jobs to overlap, so we limit the CPU\Disk pressure while node is doing its main work that's why I configured `storagePrunerAsyncRunner` to be single-threaded.

I also lowered down the thread priority.

I added the unconfirmed blobs pruning, which make sure that unconfirmed blobs (blobs its block has been received but never imported (remaining in FutureItems or in PendingPool)) are removed from DB once we finalize those slots. It shares the same prune limit (32)

Improved pruning by adding the ability to stream keys only from the KV store, so value doesn't need to be read\deserialized.

related to #6629

## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/AsyncRunnerFactory.java
+++ b/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/AsyncRunnerFactory.java
@@ -18,6 +18,7 @@ import java.util.regex.Pattern;
 public interface AsyncRunnerFactory {
 
   int DEFAULT_MAX_QUEUE_SIZE = 5000;
+  int DEFAULT_THREAD_PRIORITY = Thread.NORM_PRIORITY;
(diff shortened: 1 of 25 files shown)


ajsutton (approved):
LGTM but we should make sure we have some test coverage of iterating keys in the database somewhere.

tbenr:
It is currently tested via `pruneOldestBlobsSidecar` testing in DatabaseTest, but I generally see we don't have unit tests for the `KvStoreAccessor` layer. Something for backlog...

Issue #6327, opened 2022-10-16
Support block value in `engineGetPayload`


tbenr:
https://github.com/ethereum/execution-apis/issues/307

We could start comparing locally produced payloads with bids coming from builder endpoint and prefer local one if it has higher or similar value of the received bid. We could add percentage diff as well, so if let's say bid isn't at least 50% more valuable than local we fallback to local.

benjaminion:
I think this is a duplicate of #6140. Also see #6344.

General view seems to be that this is better done in MEV-Boost than in the consensus client. But happy to consider other views.

tbenr:
Well, not exactly. This will enable the a value comparison between local el block and builder bid. Previous issues where about applying an absolute cutoff\threshold on bids (which indeed has been recently implemented in mev-boost). The motivation behind that is to enable\encourage to choose locally produced payloads when there is "not enough rewards difference" to choose the bid. Moreover, non censored local blocks censored tx can start competing with censoring builder bids.
In the current sidecar model, mev-boost can't "see" local produced block and so the only place where it can happen is in CL.

benjaminion:
Ah, ok. Got it.

tbenr:
We can start working on this since we have #6614

PR #6632, opened 2022-12-19
Use the local block if its reported value exceed the builder bid

ajsutton:
## PR Description
Adjusts the payload selection process to actually consider the block value reported by the local Engine API.  If the local block value is equal to or greater than the block builder's block, the local block is used.

One thing to note with this change is that we now need to wait for both the builder getHeader and the local getPayload call to return before deciding which block to sign, whereas previously we'd only wait for the local EL getPayload if we actually had to fall back to it. There is a 1 second timeout on the local EL getPayload call though so the delay is strictly limited (and that 1 second timeout is in parallel with the call to the builder getHeader).

Changes `BuilderBidValidationException` to be unchecked since the only place it was being caught was using `thenComposeChecked` which just bundles it into a failed future anyway so being checked didn't provide any value anyway.

## Documentation

## Changelog


tbenr (approved):
LGTM.
referencing #6327
To close the issue we should implement a configurable percentage, not just a direct value comparison

PR #6636, opened 2022-12-20
[EIP-4844] Implement BlobsManager

### LINKED ISSUE #6473: implement `beacon_block_and_blobs_sidecar` gossip validation

### LINKED ISSUE #6629: implement `BlobsSidecar` lifecycle.

tbenr:
Introduces `BlobsSidecarManager` which is responsible for:
- storing unconfirmed _validated_ BlobsSidecars coming from gossip via `BlockManager`
- storing unconfirmed _non_validated_ BlobsSidecars coming from syncing flows via `BlockManager`
- creating a `blobsSidecarAvailabilityChecker` used in `ForkChoice`
- caching validated blobs to avoid revalidation in `ForkCoiche` for blocks\sidecar coming from gossip. 

Under kvstore and storage:
- rename kvstore method `removeUnconfirmedBlobsSidecar` -> `confirmBlobsSidecar`
- remove two unused blobs methods in `ChainStorage` and related classes
- fix SafeFuture handling on `ChainStorage` fro void returning blobs methods

In `StorageUpdate` and `StoreTransactionUpdates`:
- migrate prunedHotBlockRoots from a `Set<blockRoot>` to `Map<blockRoot,slot>` to support blobs cleanup when hot blocks are pruned

In `BlockValidator` (gossip):
- implements kzg commitments validation against blobsSidecar

In `BlockManager`:
- added `handleBlobs` method in the `doImportBlock` pipeline to store blobsSidecar via `BlobsSidecarManager`. This is done after block validity check (`handleInvalidBlock`) and after checking it is actually a new block not seen before (`handleKnownBlock`). Gossip validation result is considered to allow `BlobsSidecarManager` to know if the availability check needs to be done or not.
- it handles the blobs "discard" conditions when the corresponding block import results in a rejection of the block.

## Additional notes
- we could discard blobs sending them to the BlobsPruner which will gradually delete them from BD without queue the deletions in the storageUpdateChannel which can cause slowdown if many blobs are discarded in sequence (various pending blocks on a fork that turns out to be invalid). We could implement a BlobsDeletionChannel and block BlockPruner will consume it 

- having unconfirmed affects how we handle BlobsSidecarByRange rpc method:
  - for non finalized blobs, we should have a fast way to check if corresponding blocks are in recentChainData (via `containsBlock`)
  - for finalized we should always lookup on the unconfirmed column to filter them out (even in the case we have only one blob per slot, it might be still unconfirmed (a proposed invalid block))

- for the unconfirmed column I found an interesting approach that could make the confirmed-unconfirmed flag more efficient to store and update: https://groups.google.com/g/leveldb/c/qOMWvp7gyxg but would require some changes in our kvstore layer
  - the idea is to have a column key stored like <id, key, attributeId> where `id` is our current _prefix_ and `attributeId`s is a new _postfix_ byte from 0 to number of attributes defined + the actual value we store. The column will have multiple value serializers one for each attributeId. In our case we can have:
    - `<blobsSidecarId, <SlotAndBlockRoot>, 0>` -> `KvStoreSerializer<Boolean>` (confirmed flag)
    - `<blobsSidecarId, <SlotAndBlockRoot>, 1>` -> `KvStoreSerializer<Bytes>` (BlobsSidecar ssz data)
  so when streaming\lookingup blobs we can efficiently filter by unconfirmed flag without even try to deserialize the object if doesn't match the filter.
  This need further investigation for rocksDB.

fixes #6629
fixes #6473

## Documentation

## Changelog


ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -1754,10 +1817,20 @@ public BlobsSidecar randomBlobsSidecar(final Bytes32 blockRoot, final UInt64 slo
         SchemaDefinitionsEip4844.required(spec.getGenesisSchemaDefinitions())
             .getBlobsSidecarSchema();
 
+    return randomBlobsSidecar(
+        blockRoot, slot, randomInt((int) blobsSidecarSchema.getBlobsSchema().getMaxLength()));
+  }
+
+  public BlobsSidecar randomBlobsSidecar(
+      final Bytes32 blockRoot, final UInt64 slot, int numberOfBlobs) {
```
zilm13: final `numberOfBlobs`?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobsSidecarManagerImpl.java
```diff
@@ -0,0 +1,148 @@
...
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.eip4844.BeaconBlockBodyEip4844;
+import tech.pegasys.teku.spec.datastructures.execution.versions.eip4844.BlobsSidecar;
+import tech.pegasys.teku.spec.logic.versions.eip4844.blobs.BlobsSidecarAvailabilityChecker;
+import tech.pegasys.teku.statetransition.forkchoice.ForkChoiceBlobsSidecarAvailabilityChecker;
+import tech.pegasys.teku.storage.api.StorageQueryChannel;
+import tech.pegasys.teku.storage.api.StorageUpdateChannel;
+import tech.pegasys.teku.storage.client.RecentChainData;
+
+public class BlobsSidecarManagerImpl implements BlobsSidecarManager, SlotEventsChannel {
+  private static final int MAX_CACHED_VALIDATED_BLOBS_PER_SLOT = 10;
```
zilm13: I wouldn't shorten this as `Blobs` and `BlobsSidecar` makes much difference. Maybe `MAX_CACHED_VALIDATED_SIDECARS_PER_SLOT`?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobsSidecarManagerImpl.java
```diff
@@ -0,0 +1,148 @@
...
+    // cache already validated blobs
+    validatedPendingBlobs
+        .computeIfAbsent(blobsSidecar.getBeaconBlockSlot(), __ -> createNewMap())
+        .put(blobsSidecar.getBeaconBlockRoot(), blobsSidecar);
+
+    internalStoreUnconfirmedBlobs(blobsSidecar);
+  }
+
+  @Override
+  public void storeUnconfirmedBlobs(final BlobsSidecar blobsSidecar) {
```
zilm13: maybe use Sidecar keyword here and there too rather than have a chance to mess Blob with Blobs

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobsSidecarManagerImpl.java
```diff
@@ -0,0 +1,148 @@
...
+
+  @Override
+  public void onSlot(final UInt64 slot) {
+    validatedPendingBlobs.headMap(slot.decrement()).clear();
+  }
+
+  private void internalStoreUnconfirmedBlobs(final BlobsSidecar blobsSidecar) {
+    storageUpdateChannel
+        .onBlobsSidecar(blobsSidecar)
+        .thenRun(() -> LOG.debug("Unconfirmed BlobsSidecar stored {}", blobsSidecar))
```
zilm13: I'm still in doubts that there is any case someone want to see all raw blobs in their logs, but anyone with DEBUG logging will get a huge impact here without notice. I like the approach above this when you logged `SlotAndBlockRoot` and I also think we may need to check logging of `BlobsSidecar`'s in current master codebase to avoid logging of this

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -177,19 +182,21 @@ public void onBlockImported(final SignedBeaconBlock block) {
   }
 
   private void importBlockIgnoringResult(final SignedBeaconBlock block) {
-    doImportBlockAndBlobsSidecar(block, Optional.empty(), Optional.empty())
+    doImportBlock(block, Optional.empty(), Optional.empty(), Optional.empty())
         .ifExceptionGetsHereRaiseABug();
   }
 
-  private SafeFuture<BlockImportResult> doImportBlockAndBlobsSidecar(
+  private SafeFuture<BlockImportResult> doImportBlock(
```
zilm13, line 189: Why remove sidecar from name? It's still here

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -177,19 +182,21 @@ public void onBlockImported(final SignedBeaconBlock block) {
...
       final SignedBeaconBlock block,
       final Optional<BlobsSidecar> blobsSidecar,
+      final Optional<InternalValidationResult> validationResult,
       final Optional<BlockImportPerformance> blockImportPerformance) {
     return handleInvalidBlock(block)
         .or(() -> handleKnownBlock(block))
+        .or(() -> handleBlobs(blobsSidecar, validationResult))
```
zilm13: It's ok for now, but it's starting to look complex and unclear in functional style

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
```diff
@@ -134,12 +142,18 @@ public SafeFuture<InternalValidationResult> validate(
...
-
               if (!blockSignatureIsValidWithRespectToProposerIndex(block, postState)) {
                 return reject("Block signature is invalid");
               }
 
+              if (blobsSidecar.isPresent()) {
+                if (!blockKzgCommitmentsAreValidAgainstBlobsSidecar(
+                    block, blobsSidecar.get(), miscHelpers)) {
+                  return reject(
+                      "Block kzg commitments are invalid against the given blobs sidecars");
```
zilm13: `the given blobs sidecars` -> `sidecar` is single here I think

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
```diff
@@ -209,6 +223,27 @@ private boolean currentFinalizedCheckpointIsAncestorOfBlock(SignedBeaconBlock bl
...
 
+  private boolean blockKzgCommitmentsAreValidAgainstBlobsSidecar(
+      final SignedBeaconBlock block,
+      final BlobsSidecar blobsSidecar,
+      final MiscHelpers miscHelpers) {
+    final SszList<SszKZGCommitment> blobKzgCommitments =
+        BeaconBlockBodyEip4844Impl.required(block.getBeaconBlock().orElseThrow().getBody())
+            .getBlobKzgCommitments();
+    try {
+      return miscHelpers.isDataAvailable(
```
zilm13, line 234: I know that it's by the spec, but it's so unclear that validation is inside too. Maybe we could add Javadoc to `isDataAvailable` to clarify this?

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTest.java
```diff
@@ -145,6 +154,16 @@ public void setup() {
         .thenReturn(SafeFuture.completedFuture(PayloadValidationResult.VALID));
 
     forkChoice.subscribeToOptimisticHeadChangesAndUpdate(optimisticSyncStateTracker);
+
+    // bobs always available
```
zilm13: Blobs

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTest.java
```diff
@@ -162,13 +181,54 @@ void shouldNotTriggerReorgWhenEmptyHeadSlotFilled() {
...
+
+    importBlockWithError(blockAndState, FailureReason.FAILED_BLOBS_AVAILABILITY_CHECK);
+
+    verify(blobsSidecarManager).createAvailabilityChecker(blockAndState.getBlock());
+    verify(blobsSidecarAvailabilityChecker).initiateDataAvailabilityCheck();
+    verify(blobsSidecarAvailabilityChecker).getAvailabilityCheckResult();
+  }
+
+  @Test
+  void onBlock_shouldFailIfBlobsAreNotRequired() {
```
zilm13: Why fail?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobsSidecarManagerImpl.java
```diff
@@ -0,0 +1,148 @@
...
+    validatedPendingBlobs
+        .computeIfAbsent(blobsSidecar.getBeaconBlockSlot(), __ -> createNewMap())
+        .put(blobsSidecar.getBeaconBlockRoot(), blobsSidecar);
+
+    internalStoreUnconfirmedBlobs(blobsSidecar);
+  }
+
+  @Override
+  public void storeUnconfirmedBlobs(final BlobsSidecar blobsSidecar) {
+    // remove blobs from the already validated cache
```
zilm13: Why? We couldn't unvalidate blobs, but we could get, say, IGNORE result on the second import try and remove validation in this case. I don't understand why it's needed. From my point of view, it could only damage us in some case

zilm13 (approved):
Looks definitely like an improvement. Just a few things:

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobsSidecarManagerImpl.java
tbenr: yeah I missed renaming it from the spike

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobsSidecarManagerImpl.java
tbenr: yeah it isn't required.. it is a strange case, that should never occur, so the idea was to be conservative and force a revalidation of the sidecar against commitments during subsequent block import. But since it is strictly coupled with the block via blockroot I can remove it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
tbenr: Adding a comment there.. It might be that the validation will go away though, but not now

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
tbenr: Since everything is block centric and blobsSidecar are optional, I think it is more consistent.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobsSidecarManagerImpl.java
tbenr: k, changed to log `SlotAndBlockRoot` for now.

PR #6642, opened 2022-12-21
Add script to monitor sync committee subnet peers

ajsutton:
## PR Description
Add a test script that requests the beacon node subscribe to sync committee subnets and then monitors the number of peers available on that subnet. Primarily useful for simulating a validator in a sync committee without needing a real validator.

Also removes a pointless copy of `SyncCommitteeSubnetSubscription` to an identical object (previously that would have translated to an internal class but no need to do so with the new type definition approach).

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #6645, opened 2022-12-21
Only request peer identity once

### LINKED ISSUE #6297: Only request peer identity once
### Description
Ensure that Teku only requests the identity of peers once. It looks like currently we're requesting identity each time we received an updated status message which is unnecessary.

courtneyeh:
fixes #6297 

## Documentation

## Changelog

### DIFF
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPeer.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPeer.java
@@ -86,6 +86,9 @@ public class LibP2PPeer implements Peer {
 
   @Override
   public void checkPeerIdentity() {
+    if (maybeAgentString.isEmpty()) {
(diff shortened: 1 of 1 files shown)


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPeer.java
```diff
@@ -86,6 +86,9 @@ public LibP2PPeer(
 
   @Override
   public void checkPeerIdentity() {
+    if (maybeAgentString.isEmpty()) {
+      return;
+    }
```
ajsutton: I think that's backwards and now we'll never request the peer identity.

courtneyeh: Opps

ajsutton (approved):
LGTM.

Issue #6633, opened 2022-12-20
Implement blobs related Store methods


StefanBratanov:
As part of implementing the P2P networking, some of the Store methods were left unimplemented in `StoreTransaction`, `TestStoreImpl`, `CombinedChainDataClient`, `Store`:

- [x] getEarliestAvailableBlobsSidecarEpoch();
- [x] getBlobsSidecarBySlotAndBlockRoot(final UInt64 slot, final Bytes32 blockRoot);
- [x] retrieveSignedBlockAndBlobsSidecar(final Bytes32 blockRoot);

PR #6647, opened 2022-12-21
Add store methods required for `BlobsSidecarsByRange`

StefanBratanov:
## PR Description
- Implemented `getEarliestAvailableBlobsSidecarEpoch` and `getBlobsSidecarBySlotAndBlockRoot`
- Changed validation for `BlobsSidecarsByRange` and improved the `is blobs available` check

## Fixed Issue(s)
will complete some of the tasks in #6633 

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandler.java
@@ -137,10 +137,9 @@ public class BlobsSidecarsByRangeMessageHandler
    */
   private boolean checkBlobsSidecarsAreAvailable(
       final Optional<UInt64> earliestAvailableSidecarEpoch, final UInt64 requestEpoch) {
-    if (earliestAvailableSidecarEpoch.isEmpty()) {
-      return false;
-    }
-    return earliestAvailableSidecarEpoch.get().isLessThanOrEqualTo(requestEpoch);
+    return earliestAvailableSidecarEpoch
+        .map(earliestEpoch -> earliestEpoch.isLessThanOrEqualTo(requestEpoch))
+        .orElse(false);
   }
 
   private boolean checkRequestInMinEpochsRange(final UInt64 requestEpoch) {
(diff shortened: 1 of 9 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandler.java
```diff
@@ -228,7 +228,10 @@ SafeFuture<Optional<BlobsSidecar>> loadNextBlobsSidecar() {
               block ->
                   block
                       .map(SignedBeaconBlock::getRoot)
-                      .map(combinedChainDataClient::getBlobsSidecarByBlockRoot)
+                      .map(
+                          blockRoot ->
+                              combinedChainDataClient.getBlobsSidecarBySlotAndBlockRoot(
+                                  currentSlot.get(), blockRoot))
```
zilm13, line 221: could we modify 
```
boolean isComplete() {
      return currentSlot.get().equals(maxSlot)
```
to `greaterOrEquals`, please. it will add some safety

zilm13 (commented):
I don't understand why do we need this one now?
https://github.com/ConsenSys/teku/blob/28aa29f542e983de9f0945aa7fa8d9f88c9a9065/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandler.java#L130-L134

Ok, I think, we should keep it but it should use message.getStartSlot() epoch instead of currentEpoch, when calling `verifyBlobsSidecarsAreAvailable(earliestAvailableEpoch, currentEpoch)`.
A bit more complicated:
As I understand spec, we should throw it when message.getStartSlot() is in or after startEpoch and `earliestAvailableSidecarEpoch.get().isGreaterThan(message.getStartSlot().toEpoch()`. Am I right?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandler.java
```diff
@@ -78,32 +77,39 @@ public BlobsSidecarsByRangeMessageHandler(
...
+   */
   @Override
   public Optional<RpcException> validateRequest(
       final String protocolId, final BlobsSidecarsByRangeRequestMessage request) {
-    final UInt64 maxSlotRequested = request.getMaxSlot();
-    final SpecMilestone specMilestone =
-        spec.getForkSchedule().getSpecMilestoneAtSlot(maxSlotRequested);
-    if (!specMilestone.isGreaterThanOrEqualTo(SpecMilestone.EIP4844)) {
+    final UInt64 requestEpoch = spec.computeEpochAtSlot(request.getStartSlot());
+    if (!verifyRequestEpochIsWithinMinEpochsForBlobsSidecars(requestEpoch)) {
```
zilm13: We could have blobs outside the `must` range and I'm not sure why we shouldn't handle such kind of requests. We are not forced to prune old blobs and some user may want to run archive blob nodes. We must have blobs in range but could do anything we want outside the range

zilm13 (approved):
LGTM

PR #6664, opened 2023-01-06
TimeBasedEth1HeadTracker needs an eth1 block to complete its initiali…

rolfyone:
…sation

We are using a `getGuaranteedLatestEth1Block` elsewhere, and that will fix the problem in the init also.

Basically we can only really keep retrying until we get something but otherwise the init need not complete, as it's tracking the head block and none has been returned.

This was a problem on a devnet, causing an NPE constantly and filling up logs.

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

PR #6670, opened 2023-01-09
VC sending sync committee subscribe request to BN at every epoch

### LINKED ISSUE #6643: Send sync committee subscriptions to the beacon node more often
### Description
Currently the `/eth/v1/validator/sync_committee_subscriptions` Api is only called by the validator client when sync committee duties are recalculated which is typically a couple of epochs before the sync committee starts.

If an external VC is being used and the beacon node is restarted after that call, it will forget the subscriptions the VC may not set them up again (similar things can happen with some fail over setups etc).

The VC should notify required subscriptions every epoch.

lucassaldanha:
## PR Description
With this change, the VC will send a request to the BN to subscribe to sync committee subnets at every epoch. This is to ensure that even after a restart, the BN eventually will know what it needs to subscribe to.

## Fixed Issue(s)
fixes #6643


validator/client/src/main/java/tech/pegasys/teku/validator/client/SyncCommitteeScheduler.java
```diff
@@ -100,6 +107,13 @@ public void onSlot(final UInt64 slot) {
     }
   }
 
+  private boolean isFirstSlotOfEpoch(final UInt64 slot) {
+    final UInt64 currentEpoch = spec.atSlot(slot).miscHelpers().computeEpochAtSlot(slot);
+    final UInt64 startSlotAtEpoch =
+        spec.atEpoch(slot).miscHelpers().computeStartSlotAtEpoch(currentEpoch);
```
rolfyone: below we used `spec.computeStartSlotAtEpoch` which was a nice shortcut...

could be something like `spec.computeStartSlotAtEpoch(spec.computeEpochAtSlot(slot)).equals(slot)`

rolfyone: may be cleaner to leave currentEpoch defined and just cleaner :)
```
final UInt64 currentEpoch = spec.computeEpochAtSlot(slot);
final UInt64 startSlotAtEpoch = spec.computeStartSlotAtEpoch(currentEpoch);
return startSlotAtEpoch.equals(slot);
```

validator/client/src/main/java/tech/pegasys/teku/validator/client/PendingDuties.java
```diff
@@ -116,6 +116,11 @@ public synchronized void recalculate() {
         });
   }
 
+  @SuppressWarnings("unchecked")
+  public SafeFuture<Optional<T>> getScheduledDuties() {
+    return (SafeFuture<Optional<T>>) duties;
```
rolfyone: given we're in this class, i'm super tempted to rename `duties` -> `scheduledDutiesFuture`, might be worth seeing what it'd look like

rolfyone: i'd probably also make this do the work...
```
public Optional<T> getScheduledDutues() {
    return duties
          .filter(SafeFuture::isCompletedNormally)
          .flatMap(SafeFuture::getImmediately);
}
```
just returning empty however that works, i wont attempt to do that without playing :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/SyncCommitteeScheduler.java
```diff
@@ -205,12 +218,20 @@ public void calculateDuties() {
...
+          .ifPresent(SyncCommitteeScheduledDuties::subscribeToSubnets);
+    }
+
     public void recalculate() {
       duties.ifPresent(PendingDuties::recalculate);
     }
   }
 
   public interface EarlySubscribeRandomSource {
+
```
rolfyone: ... did you add this or spotless?

lucassaldanha: either spotless or intellij 🤷

lucassaldanha: removed!

rolfyone (approved):
LGTM

Issue #6539, opened 2022-12-02
Print configuration overview at startup


benjaminion:
Besu recently implemented [dumping the configuration](https://github.com/hyperledger/besu/pull/4451) to the console / logon startup. I like this! We should consider something similar fer Teku.

```
####################################################################################################
#                                                                                                  #
# Besu 22.10.1                                                                                     #
#                                                                                                  #
# Configuration:                                                                                   #
# Network: Mainnet                                                                                 #
# Data storage: Bonsai                                                                             #
# Sync mode: Checkpoint                                                                            #
# RPC HTTP APIs: ETH,NET,WEB3,DEBUG,ADMIN                                                          #
# RPC HTTP port: 8545                                                                              #
# Engine APIs: ENGINE,ETH                                                                          #
# Engine port: 8551                                                                                #
#                                                                                                  #
# Host:                                                                                            #
# Java: openjdk-java-17                                                                            #
# Maximum heap size: 5.00 GB                                                                       #
# OS: linux-x86_64                                                                                 #
# glibc: 2.35                                                                                      #
# Total memory: 15.50 GB                                                                           #
# CPU cores: 6                                                                                     #
#                                                                                                  #
####################################################################################################
```

benjaminion:
From stand-up discussion: I'm not wedded to a format. That Besu one is pleasant to me, but I don't use logz.io so have no idea what would work there. The important things are to (safely) enumerate any user-supplied (non-default) options and to summarise the running environment in some format.

Some of the Besu output is just the default settings - that's not required imo.

courtneyeh:
Current thoughts is for a format as below (seen in the open PR #6671):
```
Configuration | Network: mainnet, Storage Mode: PRUNE, Rest API Port: 5678
```
Open for a discussion / suggestions.
@rolfyone @lucassaldanha

PR #6671, opened 2023-01-10
Print configuration overview at startup

courtneyeh:
Partially addresses #6539 

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/AbstractNode.java
```diff
@@ -59,7 +60,18 @@ public abstract class AbstractNode implements Node {
...
+        tekuConfig
+            .eth2NetworkConfiguration()
+            .getEth2Network()
+            .map(Eth2Network::configName)
+            .orElse("empty");
+    final String storageMode = tekuConfig.storageConfiguration().getDataStorageMode().name();
+    final int restApiPort = tekuConfig.beaconChain().beaconRestApiConfig().getRestApiPort();
+
     STATUS_LOG.onStartup(VersionProvider.VERSION);
+    STATUS_LOG.startupConfigurations(network, storageMode, restApiPort);
```
lucassaldanha: I believe we should encapsulate the logic of what information goes into the log in a separate object (e.g. `StartupConfigLog`). This will give us more flexibility in testing and future changes. Might even rely on a `toString()` method to actually build the message if we decide to do it all as one string.

This way, you can update StatusLogger to have a method like `StatusLogger.startupConfiguration(StartupLogConfig config)` that we don't have to change the signature every time we want to add/remove/change the startup config.

I am not too sure about the value of having this part of StatusLogger, but if there is a reason to have them tied up, we can invert the dependency and have `StartupConfigLog` receive a Consumer<StartupConfigLog> in its constructor that will be executed when you call a `report()` method. And have the consumer be provided by StatusLogger.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -52,6 +52,10 @@ public void onStartup(final String version) {
             + "You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0");
   }
 
+  public void startupConfigurations(final StartupLogConfig config) {
+    config.report(log);
+  }
```
lucassaldanha: This is kinda creating a two-way dependency that does not add any value. 

What I meant with my previous comment was considering two things:
1. that we need to use the StatusLogger (maybe we don't need it)
2. that we want a way to test the `StartupLogConfig` msg (like check we are building the expected msg/msgs.

infrastructure/logging/src/test/java/tech/pegasys/teku/infrastructure/logging/StartupLogConfigTest.java
```diff
@@ -0,0 +1,44 @@
...
+    final CentralProcessor centralProcessor = mock(CentralProcessor.class);
+    when(hardwareInfo.getProcessor()).thenReturn(centralProcessor);
+    when(centralProcessor.getLogicalProcessorCount()).thenReturn(10);
+
+    final StartupLogConfig config = new StartupLogConfig("mainnet", "PRUNE", 5678, hardwareInfo);
+    assertThat(config.getReport())
+        .containsExactly(
+            "Configuration | Network: mainnet, Storage Mode: PRUNE, Rest API Port: 5678",
+            "Host Configuration | Maximum Heap Size: 4.00 GB, Total Memory: 16.00 GB, CPU Cores: 10");
+  }
```
lucassaldanha: @benjaminion do you have thoughts on the formatting and what info is included here?

infrastructure/logging/src/test/java/tech/pegasys/teku/infrastructure/logging/StartupLogConfigTest.java
```diff
@@ -0,0 +1,44 @@
...
+    when(memory.getTotal()).thenReturn(Long.valueOf("17179869184"));
+
+    final CentralProcessor centralProcessor = mock(CentralProcessor.class);
+    when(hardwareInfo.getProcessor()).thenReturn(centralProcessor);
+    when(centralProcessor.getLogicalProcessorCount()).thenReturn(10);
+
+    final StartupLogConfig config = new StartupLogConfig("mainnet", "PRUNE", 5678, hardwareInfo);
+    assertThat(config.getReport())
+        .containsExactly(
+            "Configuration | Network: mainnet, Storage Mode: PRUNE, Rest API Port: 5678",
```
rolfyone: could probably add the rest api listen address if listing port, and maybe the allow list if its set... I'd be tempted to have a line for rest api...
Also a line for EL configuration would be the other likely one

rolfyone:
```
Configuration | Network: mainnet, Storage Mode: PRUNE
Host Configuration | Maximum Heap Size: 4.00 GB, Total Memory: 16.00 GB, CPU Cores: 10
Rest Api Configuration | Listen address: 0.0.0.0, Port 5678, Allow: *
Validator Api Configuration | Listen address: 0.0.0.0, Port 6789, Allow: localhost
Execution Layer Configuration | Listen address: 0.0.0.0, Port 7890, Allow: localhost
```

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StartupLogConfig.java
```diff
@@ -0,0 +1,183 @@
...
+    final String general =
+        String.format("Configuration | Network: %s, Storage Mode: %s", network, storageMode);
+    final String host =
+        String.format(
+            "Host Configuration | Maximum Heap Size: %s, Total Memory: %s, CPU Cores: %d",
+            maxHeapSize, memory, cpuCores);
+    final String restApi =
+        String.format(
+            "Rest Api Configuration | Listen Address: %s, Port: %s, Allow: %s",
+            beaconChainRestApiInterface, beaconChainRestApiPort, beaconChainRestApiAllow);
```
rolfyone: how does this go when rest api isn't available?

courtneyeh: Good point, I'll add a check for whether they are enabled. We shouldn't print that line if it is not.

courtneyeh: Actually it still comes out normal when rest api isn't enabled. Do we want that? Or should we be indicating it isn't enabled? Or removing the line?

rolfyone: ok so it'll be grabbing defaults, so we want to probably add the enabled flag for those too, or not print them when enabled is false...

rolfyone (approved):
LGTM, we should probably change PR description to be 'partially addresses' and do a follow up for EL

PR #6684, opened 2023-01-13
Implement forward multipeer sync for EIP-4844

### LINKED ISSUE #6669: Implement Forward Sync for EIP-4844
Implement new sync logic PeerSync and BatchSync:
  - for pre 4844 blocks we continue using BeaconBlocksByRange
  - for post 4844 blocks:
    - if we need blocks and blobs we use BeaconBlocksByRange and BlobsSidecarsByRange (this could be done by let `SyncSourceBatch` handle blocks and blobs and check the consistency, having two different batch types seems more complicated)
    - if we need only blocks we continue using BeaconBlocksByRange

StefanBratanov:
## PR Description

- Modified `SyncSourceBatch` to cache the blobs sidecars for a batch
- Modified `BatchImported` to store blobs sidecars ahead of block import

Tested syncing to mainnet with stub EL, so current flow seems not affected.

## Fixed Issue(s)
fixes #6669 

## Documentation

## Changelog

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
@@ -19,7 +19,9 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Throwables;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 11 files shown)


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
```diff
@@ -200,9 +216,28 @@ public void requestMoreBlocks(final Runnable callback) {
     final SyncSource syncSource = currentSyncSource.orElseThrow();
     LOG.debug(
         "Requesting {} slots starting at {} from peer {}", remainingSlots, startSlot, syncSource);
```
tbenr: I think something like "Requesting blocks for {} slots starting at {} from peer {}" should be better. Similarly below.

tbenr (approved):
LGTM. just a suggestion on debug messages

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
StefanBratanov: Yeah I agree, good suggestion! I changed it and also changed the sidecar log message and level to debug. Now it's more consistent.

PR #6685, opened 2023-01-14
SSZ serialization of LightClientUpdate

robzajac:
## PR Description
The SSZ serialization of the endpoint `GetLightClientUpdatesByRange` has a custom octet encoding defined [here](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/updates.yaml#L37). This PR implements the serialization.

Note that the SSZ object is simply used for convenience, and the field names are irrelevant as it should only be used for the `application/octet-stream` response type.

## Fixed Issue(s)
partially addresses #4230 

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/lightclient/GetLightClientUpdatesByRange.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/lightclient/GetLightClientUpdatesByRange.java
@@ -21,13 +21,22 @@ import static tech.pegasys.teku.infrastructure.http.RestApiConstants.TAG_BEACON;
 import static tech.pegasys.teku.infrastructure.http.RestApiConstants.TAG_EXPERIMENTAL;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import java.util.Collections;
 import java.util.List;
+import tech.pegasys.teku.infrastructure.bytes.Bytes4;
 import tech.pegasys.teku.infrastructure.json.types.SerializableTypeDefinition;
 import tech.pegasys.teku.infrastructure.restapi.endpoints.EndpointMetadata;
 import tech.pegasys.teku.infrastructure.restapi.endpoints.RestApiEndpoint;
 import tech.pegasys.teku.infrastructure.restapi.endpoints.RestApiRequest;
+import tech.pegasys.teku.infrastructure.restapi.openapi.response.OctetStreamResponseContentTypeDefinition;
+import tech.pegasys.teku.infrastructure.restapi.openapi.response.ResponseContentTypeDefinition;
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszBytes4;
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszUInt64;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.datastructures.lightclient.LightClientUpdate;
+import tech.pegasys.teku.spec.datastructures.lightclient.LightClientUpdateResponse;
+import tech.pegasys.teku.spec.datastructures.lightclient.LightClientUpdateResponseSchema;
 import tech.pegasys.teku.spec.datastructures.metadata.ObjectAndMetaData;
 import tech.pegasys.teku.spec.schemas.SchemaDefinitionCache;
 import tech.pegasys.teku.spec.schemas.SchemaDefinitionsAltair;
@@ -45,7 +54,11 @@ public class GetLightClientUpdatesByRange extends RestApiEndpoint {
             .tags(TAG_BEACON, TAG_EXPERIMENTAL)
             .queryParamRequired(START_PERIOD_PARAMETER)
             .queryParamRequired(COUNT_PARAMETER)
-            .response(SC_OK, "Request successful", getResponseType(schemaDefinitionCache))
+            .response(
+                SC_OK,
+                "Request successful",
+                getResponseType(schemaDefinitionCache),
+                getSszResponseType(schemaDefinitionCache))
             .withNotAcceptedResponse()
             .withNotImplementedResponse()
             .build());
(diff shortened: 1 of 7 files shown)


lucassaldanha:
Thanks, @robzajac we will take a look at it asap! 👍

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/lightclient/GetLightClientUpdatesByRange.java
```diff
@@ -73,4 +86,41 @@ public void handleRequest(RestApiRequest request) throws JsonProcessingException
...
+              LightClientUpdateResponseSchema schema =
+                  SchemaDefinitionsAltair.required(
+                          schemaDefinitionCache.getSchemaDefinition(SpecMilestone.ALTAIR))
+                      .getLightClientUpdateResponseSchema();
+
+              data.stream()
+                  .forEach(
+                      lightClientUpdateObjectAndMetaData -> {
+                        // Get the fork version for the slot of `update.attested_header`
+                        Bytes4 forkDigest = Bytes4.fromHexString("TODO");
```
rolfyone: For this here we're looking at just the definition of the object for metadata purposes, would expect something similar to `v2.debug.GetState.getResponseType`, and then sszResponseType should be able to take the same 'data' object and serialize it as ssz...
It may be more complicated if the `data` is not consistent with what's sent in octet-stream, but it's a good concrete example...

robzajac: Thanks @rolfyone, I gave it another pass and simplified the interface.

Indeed this endpoint is tricky, the JSON serialization and octet serialization are effectively two different types: https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/light_client/updates.yaml#L26

For this reason I propose returning a JSON type of `List<ObjectAndMetadata<LightClientUpdate>>` and an SSZ type of `List<LightClientUpdateResponse>` (where LCUResponse is the wrapper object with fork digest, etc.) I've implemented it this way in the latest commit.

Later when we implement the endpoint, we will simply compute the fork digest in the business logic (i.e. `ChainDataProvider`) rather than trying to compute it in the endpoint... this is more sensible anyway because fork digest needs the fork version, which is located in `BeaconState`. Concretely, I will expect `ChainDataProvider` to have a function for light client updates with return type `List<ObjectAndMetadata<LightClientUpdateResponse>>`. For the JSON type, we can pull out the `version` and the `LightClientUpdate`. For the SSZ type, we simply take the `LightClientUpdateResponse`. Maybe in the end it will look slightly different, but at least this way should work with the interface we specify in this PR.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/lightclient/GetLightClientUpdatesByRange.java
```diff
@@ -71,6 +79,17 @@ public void handleRequest(RestApiRequest request) throws JsonProcessingException
...
+    return new JsonResponseContentTypeDefinition<>(
+        SerializableTypeDefinition.listOf(lightClientUpdateObjectType));
+  }
+
+  private static ResponseContentTypeDefinition<List<LightClientUpdateResponse>>
+      getSszResponseType() {
+    OctetStreamResponseContentTypeDefinition.OctetStreamSerializer<List<LightClientUpdateResponse>>
+        serializer =
+            (data, out) ->
+                data.stream().forEachOrdered(lcuResponse -> lcuResponse.sszSerialize(out));
```
robzajac, line 91: I had previously used `forEach` until I read that one does not guarantee ordering... I hope that collection streaming plays well with serialization, if not I am open to suggestions.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/lightclient/GetLightClientUpdatesByRange.java
rolfyone: Sounds like a good plan to start with, at the very least :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/lightclient/GetLightClientUpdatesByRange.java
rolfyone: I wonder if maybe we add this data type to `prettyPrint` so that we can see the output in json?

`PrettyPrintCommand`

rolfyone: Wouldnt need to be done here in this PR, just an idea that may save your sanity debugging things...

rolfyone (approved):
LGTM just a merge conflict if you get to it before i do...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/lightclient/GetLightClientUpdatesByRange.java
robzajac: That's a neat idea. How do we add it to `PrettyPrintCommand`? Is it sufficient to implement `getJsonTypeDefinition()` in `LightClientUpdateResponseSchema`?

rolfyone: Not sure to be honest for a non spec type, one of us will have to play and see what happens, you use it via debugTools
I know for SignedBeaconBlock I can just do
```
 build/install/teku/bin/teku debug-tools pp capella SignedBeaconBlock a.ssz
 ```
 
 and i think it'd be fair to expect something similar where we specify the type and it can read from the SSZ file and output the json.  The typeDefinition is the way to go for that bit, just a matter of whether we need to do some handling for the type bit in the command itself...

I'm off-site next week but can probably play after that, I might get a little extra time, but its work trip so might be nuts...
If you like, have a play and see what you think...

Issue #6655, opened 2023-01-03
VC failover when EL is offline


StefanBratanov:
VC should failover when the EL connected to a BN is offline

**Background** 

Question from user on Discord (obfuscate):

`Hi there, I am currently experimenting with fail over scenarios in my system. I have the Teku Beacon Node and Validator clients split out into separate processes. I have a primary BN connected to Besu and a secondary BN connected to Nethermind. I have configured the two BN endpoints in the VCs to examine the fail over behaviour. When I bring the primart BN down it fails over to the secondary BN almost instantly which is great! However, when I bring my primary EL client (Besu) down, the VC doesn't seem to failing over the secondary BN (with the healthy Nethermind node connected it).`

https://discord.com/channels/697535391594446898/697539289042649190/1058317129805484082

spenot09:
Hi there I've also noticed on the flip side I am seeing a failover happening under the hood it seems. As when the primary EL is down, blocks are still being produced and only when the secondary EL is down do block production failures start creeping in

PR #6696, opened 2023-01-17
Add el-offline to syncing api response

mehdi-aouadi:
## PR Description
Add an `el_offline` to the `/eth/v1/node/syncing` syncing API response

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6655

## Documentation

## Changelog

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/events/SyncStateTracker.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/events/SyncStateTracker.java
@@ -20,6 +20,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import tech.pegasys.teku.beacon.sync.forward.ForwardSync;
+import tech.pegasys.teku.ethereum.executionclient.events.ExecutionClientEventsChannel;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.logging.EventLogger;
@@ -32,7 +33,7 @@ import tech.pegasys.teku.service.serviceutils.Service;
 import tech.pegasys.teku.statetransition.forkchoice.ForkChoice.OptimisticHeadSubscriber;
 
 public class SyncStateTracker extends Service
-    implements SyncStateProvider, OptimisticHeadSubscriber {
+    implements SyncStateProvider, OptimisticHeadSubscriber, ExecutionClientEventsChannel {
   private static final Logger LOG = LogManager.getLogger();
   private final AsyncRunner asyncRunner;
   private final ForwardSync syncService;
@@ -49,6 +50,8 @@ public class SyncStateTracker extends Service
   private long peerConnectedSubscriptionId;
   private long syncSubscriptionId;
   private boolean headIsOptimistic = false;
+
+  private boolean executionLayerIsAvailable = true;
 
   private volatile SyncState currentState;
 
@@ -130,7 +133,9 @@ public class SyncStateTracker extends Service
 
   private void updateCurrentState() {
     final SyncState previousState = currentState;
-    if (headIsOptimistic) {
+    if (!executionLayerIsAvailable) {
+      currentState = SyncState.AWAITING_EL;
+    } else if (headIsOptimistic) {
       currentState = syncActive ? SyncState.OPTIMISTIC_SYNCING : SyncState.AWAITING_EL;
     } else if (syncActive) {
       currentState = SyncState.SYNCING;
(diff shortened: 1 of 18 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetSyncing.java
```diff
@@ -85,10 +91,13 @@ static class SyncStatusData {
...
 
-    public SyncStatusData(final SyncDataProvider syncProvider) {
+    public SyncStatusData(
+        final SyncDataProvider syncProvider,
+        ExecutionClientDataProvider executionClientDataProvider) {
       final SyncingStatus status = syncProvider.getSyncingStatus();
       final SyncState syncState = syncProvider.getCurrentSyncState();
-      this.isSyncing = !syncState.isInSync();
+      this.isSyncing =
+          syncState.isSyncing() || !executionClientDataProvider.isExecutionClientAvailable();
```
rolfyone: isSyncing already pulls in Optimistic_syncing and awaiting EL..., so the ELDataProvider is basically here because we don't update the syncing state properly? Maybe this is just fixing a symptom that is a problem elsewhere if that's the case?
My main concerns:
 - we seem to be overriding the results of our own SyncService with extra context, indicating maybe something is wrong with the syncService
 - this is extremely close to an api owning business logic, it does seem to blur the line a lot... I'd prefer the api functions aren't applying special business logic unless we understand really well why we need it, and potentially even then the actual answer may be adjusting the flow, not business logic in the api...

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetSyncingTest.java
```diff
@@ -49,13 +49,26 @@ public void shouldGetSyncingStatusSyncing() throws Exception {
...
+    when(executionClientDataProvider.isExecutionClientAvailable()).thenReturn(false);
+    when(syncService.getSyncStatus()).thenReturn(getSyncStatus(false, 1, 10, 11));
+    when(syncService.getCurrentSyncState()).thenReturn(SyncState.IN_SYNC);
+
+    handler.handleRequest(request);
+    assertThat(request.getResponseCode()).isEqualTo(SC_OK);
+    assertThat(request.getResponseBody())
+        .isEqualTo(new GetSyncing.SyncStatusData(true, false, 10, 1));
+  }
+
```
rolfyone: this test is probably not adequate if the handler code is to be believed, we may need an acceptance test that shuts down the EL and shows the result...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetSyncing.java
mehdi-aouadi: IMO, it depends on what the `SyncService` is supposed to do. Currently it doesn't consider the EL status (which is not necessarily incorrect). I agree with the extreme closeness of including business logic in the API and that it looks like there is something wrong with the `SyncService`.
I will give it another try and see how the EL availability could be taken into consideration directly in the `SyncService`

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetSyncingTest.java
mehdi-aouadi: OK. Actually I run some tests locally and hesitated to add an acceptance test. Will do

mehdi-aouadi:
@rolfyone I ended up making the `SyncStateTracker` listen to the `ExecutionClientEventsChannel`, removed the logic from the API and added an acceptance test

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/events/SyncStateTracker.java
```diff
@@ -130,7 +133,9 @@ public boolean unsubscribeFromSyncStateChanges(long subscriberId) {
 
   private void updateCurrentState() {
     final SyncState previousState = currentState;
-    if (headIsOptimistic) {
+    if (!executionLayerIsAvailable) {
+      currentState = SyncState.AWAITING_EL;
```
tbenr: not sure about this.
if we are syncing and EL goes down we set `AWAITING_EL` which was representing that our head is in sync but needs to be fully verified by the EL which is still syncing.

In this case we have a bad condition where EL is completely offline, so we might need another state.. have to think more about that..

mehdi-aouadi: I was thinking about adding a new `SyncState` (`EL_OFFLINE` for example). Yet that can't be really reflected in the [Sync API](https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus) since the response doesn't include such info

mehdi-aouadi:
Will be reopened/refactored once https://github.com/ethereum/beacon-APIs/pull/290 is done

mehdi-aouadi:
Replaced by https://github.com/ConsenSys/teku/pull/6770

PR #6697, opened 2023-01-17
Retrieve signed block and blobs sidecar implementation

### LINKED ISSUE #6633: Implement blobs related Store methods

tbenr:
Introduces `retrieveSignedBlockAndBlobsSidecar` method in `Store` used in the BeaconBlockAndBlobsSidecarByRoot rpc method

## Fixed Issue(s)

fixes #6633

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
@@ -497,16 +497,21 @@ class Store implements UpdatableStore {
   private SignedBeaconBlockAndBlobsSidecar createSignedBeaconBlockAndBlobsSidecar(
       final SignedBeaconBlock signedBeaconBlock, final Optional<BlobsSidecar> maybeBlobsSidecar) {
     checkState(
-        maybeBlobsSidecar.map(BlobsSidecar::getBlobs).map(List::size).orElse(0)
-            == signedBeaconBlock
-                .getBeaconBlock()
-                .orElseThrow()
-                .getBody()
-                .toVersionEip4844()
-                .orElseThrow()
-                .getBlobKzgCommitments()
-                .size(),
-        "Inconsistent number of blobs from sidecar and commitments from block");
+        maybeBlobsSidecar
+            .map(
+                blobsSidecar ->
+                    blobsSidecar.getBeaconBlockRoot().equals(signedBeaconBlock.getRoot()))
+            .orElseGet(
+                () ->
+                    signedBeaconBlock
+                        .getBeaconBlock()
+                        .orElseThrow()
+                        .getBody()
+                        .toVersionEip4844()
+                        .orElseThrow()
+                        .getBlobKzgCommitments()
+                        .isEmpty()),
+        "BlobsSidecar is inconsistent with the BeaconBlock");
 
     final SchemaDefinitionsEip4844 schemaDefinitionsEip4844 =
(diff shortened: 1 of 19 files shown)


storage/src/test/java/tech/pegasys/teku/storage/store/AbstractStoreTest.java
```diff
@@ -42,7 +43,7 @@
 import tech.pegasys.teku.storage.api.StubStorageUpdateChannel;
 
 public abstract class AbstractStoreTest {
-  protected final Spec spec = TestSpecFactory.createMinimalPhase0();
+  protected final Spec spec = TestSpecFactory.createMinimalEip4844();
```
mehdi-aouadi, line 46: Shouldn't we keep at at `Phase0` and override it if required?

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -461,10 +471,64 @@ public SafeFuture<Optional<SignedBeaconBlock>> retrieveSignedBlock(final Bytes32
...
+                () ->
+                    signedBeaconBlock
+                        .getBeaconBlock()
+                        .orElseThrow()
+                        .getBody()
+                        .toVersionEip4844()
+                        .orElseThrow()
+                        .getBlobKzgCommitments()
+                        .isEmpty()),
+        "BlobsSidecar is inconsistent with the BeaconBlock");
```
mehdi-aouadi: I would suggest to add the reason of the inconsistency (either the `BlobsSidecar` couldn't be found at all or the block root doesn't match)

mehdi-aouadi (approved):
LGTM
I left some minor changes suggestions...
Would also suggest adding a test when the `BlobsSidecar` is inconsistent (not present or block root doesn't match)

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
tbenr: point is that I mixed two different conditions here. Btw it shouldn't happen because the data comes from our own db so it is more our own consistency check instead of a validation. Btw I can split it in two, so it can be more readable.

storage/src/test/java/tech/pegasys/teku/storage/store/AbstractStoreTest.java
tbenr: It was much simpler this way.. since this is an abstract test class and we are dealing with `protected final` fields.

tbenr:
for the records, I'm exploring a different approach in the `BlobsSidecar` storage handling. I'm 99% over moving away from touching the Store completely and leave everithing at `RecentChainData` level. (and let `BlobsSidecarManager` use it without directly accessing the storage channels). If I go that route most of the things in this PR will be removed.

mehdi-aouadi (approved):
LGTM

PR #6700, opened 2023-01-18
Reference tests cleanup

zilm13:
## PR Description
Reviewing state of reference tests:
- LightClient tests - we don't support it
- `PowBlock` and `Eth1Block` tests - we don't have these ssz structures and there is no sense to create them for reference tests only because tests will verify only those structures isolated
- `PendingAttestation` for milestones other than Phase0 - I've puzzled to find a way to initialize it for other milestones but had not find it. But structures are used only in Phase0 and there is no real value to run them during other milestones.
- `ssz_generic/uints` we skip uint32 and uint128 tests. Both are not supported by Teku sizes which are not used anywhere in spec. 
- `ssz_generic/basic_vector` we skip vectors with unsupported uint32 and uint128
- `fork/fork` tests - enabled for all milestones, made it in a way we'll get failed tests with new fork
- `transition/core` tests - enabled for all milestones, made it in a way we'll get failed tests with new fork, one test was failing because we didn't update EIP4844 upgrade with historical summaries recently introduced in Capella

No other tests are ignored

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #6706, opened 2023-01-19
Deprecate v1 of beacon block methods that have been deprecated in the spec

### LINKED ISSUE #6492: Deprecate v1 of beacon block methods that have been deprecated in the spec
Reference: https://github.com/ethereum/consensus-specs/pull/3014

Deprecate v1 of the following methods:
- `/eth2/beacon_chain/req/beacon_blocks_by_root`
- `/eth2/beacon_chain/req/beacon_blocks_by_range`

**Tasks**
- Remove the methods from our production code
- Remove/Update any tests that rely on the API methods being removed
- Update changelog

courtneyeh:
fixes #6492 

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/BeaconChainMethods.java
```diff
@@ -242,9 +223,10 @@ private static Eth2RpcMethod<GoodbyeMessage, GoodbyeMessage> createGoodBye(
                   peerLookup);
 
       return VersionedEth2RpcMethod.create(
-          rpcEncoding, requestType, expectResponseToRequest, List.of(v2Method, v1Method));
+          rpcEncoding, requestType, expectResponseToRequest, List.of(v2Method));
     } else {
-      return v1Method;
+      return VersionedEth2RpcMethod.create(
+          rpcEncoding, requestType, expectResponseToRequest, List.of());
```
mehdi-aouadi: The method list should not be empty here.
I guess there is not need to check the milestone anymore. We could directly return a `VersionedEth2RpcMethod` with the `v2Method` only.
I tried that and it fixed most of the integrations tests btw

mehdi-aouadi (dismissed):

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/BeaconChainMethods.java
courtneyeh: Okay, I could do that. Is that the right approach? The [spec](https://github.com/ethereum/consensus-specs/pull/3014/files) says "Clients MAY respond with an empty list during the deprecation transition period." If that is not what we want to do, I can try your suggestions returning the `v2Method`?

mehdi-aouadi: In my understanding, that means that both `beacon_blocks_by_range` and `beacon_blocks_by_root` may return an empty list of blocks which means implementing that behaviour in the v1methods before completely removing them

beacon/sync/src/integration-test/java/tech/pegasys/teku/beacon/sync/BlockPropagationIntegrationTest.java
```diff
@@ -69,11 +69,12 @@ public void shouldFetchUnknownAncestorsOfPropagatedBlock() throws Exception {
 
     // Setup node 2
     SyncingNodeManager node2 =
-        SyncingNodeManager.create(
-            asyncRunner,
-            networkFactory,
-            validatorKeys,
-            c -> c.rpcEncoding(rpcEncoding).gossipEncoding(gossipEncoding));
+        SyncingNodeManager
+            .create( // FIXME minimal phase0 so going to get empty result for blocks (line 99)
```
rolfyone: We're not adding a fixme to master... what's the plan here? is this just a note that we need to clean up or is there still work to be done?

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/BeaconChainMethodsTest.java
```diff
@@ -93,35 +93,30 @@ public void shouldCreateVersionedBlocksByRangeMethodWithAltairEnabled() {
...
     assertThat(methods.beaconBlocksByRoot().getIds())
-        .containsExactly(
-            "/eth2/beacon_chain/req/beacon_blocks_by_root/2/ssz_snappy",
-            "/eth2/beacon_chain/req/beacon_blocks_by_root/1/ssz_snappy");
+        .containsExactly("/eth2/beacon_chain/req/beacon_blocks_by_root/2/ssz_snappy");
   }
 
   @Test
-  public void shouldCreateUnversionedBlocksByRootMethodWithAltairDisabled() {
+  public void shouldCreateUnversionedBlocksByRootMethodWithAltairDisabled() { // TODO remove test?
```
rolfyone: basically before it was testing a call to v1 not v2 was made, v1 doesnt exist (and so we can't call it...)

by that logic it's fair to say delete test

The main unanswered question is probably if we want to see that calling a function that doesn't exist results in an empty output, rather than something like things going boom...
Can we call it in this way any more? It does sound like we just won't be able to use phase0 at all without these functions? If that's the case, I think i'm happy with deletion, just wondering if there's any case where we need to actually create phase0...

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ For information on changes in released versions of Teku, see the [releases page]
 - The `validator_beacon_node_published_attestation_total`, `validator_beacon_node_published_aggregate_total`,
   `validator_beacon_node_send_sync_committee_messages_total`, `validator_beacon_node_send_sync_committee_contributions_total`
   and `validator_beacon_node_published_block_total` metrics have been replaced by the new `validator_beacon_node_requests_total` metric. An update to the [Teku Dashboard](https://grafana.com/grafana/dashboards/13457) that uses the new metric is available.
+- Deprecated v1 of beacon block methods `/eth2/beacon_chain/req/beacon_blocks_by_root` and `/eth2/beacon_chain/req/beacon_blocks_by_range`.
```
rolfyone: I think more accurately we've removed references to those functions, not deprecated their use?

rolfyone (approved):
LGTM

PR #6712, opened 2023-01-20
Implement historical blocks syncing for EIP-4844

StefanBratanov:
## PR Description

- Changed `HistoricalBatchFetcher` to import blobs sidecars if required
- Tested historical fetching with current mainnet to ensure the current workflow is not affected.
- Moved `FetchBlockTask` and all its related classses to a different package to be reused by the `HistoricalBatchFetcher`
- changed `onFinalizedBlocks` method to accept blobs sidecars

## Documentation

## Changelog


tbenr (approved):
LGTM

tbenr (changes requested):
just requesting a dummy change to prevent this to be merged.

One actual thing we may improve is this:
we currently have `isSyncDone` relying only on earliest block available, and we are not checking the blobs independently. Which means that we do not cover the conditions in which we have all the blocks but we don't have all the expected sidecars

tbenr (approved):
LGTM

PR #6740, opened 2023-01-27
Allow other users to read the /opt/teku dir when using docker

skylenet:
## PR Description

The problem is that if we try to run the teku container with any other user that has a differen uid than `1000`, we won't be able to execute the teku binary due to the permissions of the working dir ( `/opt/teku` ) being too strict 0770. I'm changing the permissions to 0775 so that other users can also read the directory and execute the teku binary.

Example on how this fails currently with other users than 1000

```sh
$ docker pull consensys/teku 
$ docker run --rm -it --entrypoint ls --user=1000 consensys/teku 
LICENSE  bin  lib  licenses  teku.autocomplete.sh
$ docker run --rm -it --entrypoint ls --user=1003 consensys/teku 
ls: cannot open directory '.': Permission denied
```

## Documentation

## Changelog


lucassaldanha:
Recreated at https://github.com/ConsenSys/teku/pull/6743 to circumvent CircleCI permission issues.

PR #6728, opened 2023-01-25
Meta 4844 interop PR [DON'T MERGE]

zilm13:
## PR Description
This branch includes all changes we do during interop event. It's a mess currently and most of them should be reworked in separate PRs after the event. Once a certain task is merged to `master` please tick the box.

- [x] Implemented historical block sync for EIP-4844 https://github.com/ConsenSys/teku/commit/8ed7a2ac4e9c57d04c2c0c0778fbb64b49740676 https://github.com/ConsenSys/teku/commit/f2f9355e17b121c104686d90374ae3937a4de809 https://github.com/ConsenSys/teku/commit/dd02ff50ad97c3c5934e92fa5fe0d6acd005cb30  (covered entirely by https://github.com/ConsenSys/teku/pull/6712) 
- [x] Fixes block casting to EIP-4844 on other forks https://github.com/ConsenSys/teku/commit/c482f043fc2914819b756ac4c138ff01a0c29aa1
- [x] Don't request `BlobsSidecars` in non EIP-4844 epochs. Say, our current is 50 slot and we have a fork on 64. Previously we requested sidecars starting from 51 to 101, instead we should request only 64-101. It's fixed with https://github.com/ConsenSys/teku/commit/f849d2dbf5673fc4144920dc9a6c91854bb88ff8
- [x] Command to dump `BlobsSidecars` from DB to zip-file https://github.com/ConsenSys/teku/commit/0de1f1383cea7217086692052eb0c87166e01cf3
- [x] Use long for a chainId,networkId instead of integer https://github.com/ConsenSys/teku/commit/8ec33baf3eb193c97be2b9f1f647d06aea62aa6d https://github.com/ConsenSys/teku/commit/1bec77b3c5be438b9b812a69c0ff7b3b1f6f1a58 https://github.com/ConsenSys/teku/commit/df0b6cf18bcbd254a08794754d227b7a153e2dc9
- [x] Using ForkDigestCodec instead of noContextCodec https://github.com/ConsenSys/teku/commit/ba9566d2f42c3f5605fc3c5ebdb590a485c83e5f https://github.com/ConsenSys/teku/commit/86495dd05aac257d1e59644b31654b3e634660b3
- [x] blobs storage metrics https://github.com/ConsenSys/teku/pull/6755
- [x] blobs pruner fix https://github.com/ConsenSys/teku/pull/6746
- [x] fix pruning of finalized blobs sidecar https://github.com/ConsenSys/teku/pull/6754

No longer needed:
[...]

PR #6748, opened 2023-01-30
Use long for Chain ID and Network ID

StefanBratanov:
## PR Description
EIP-4844 devnet had a config for DEPOSIT_CHAIN_ID and DEPOSIT_NETWORK_ID larger than what int can contain: `4844001004`. It's better to use a long to cater for such cases even though they will apply to mostly testnets.

## Fixed Issue(s)
will apply few of the commits in https://github.com/ConsenSys/teku/pull/6728 to master

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #6758, opened 2023-02-01
Blobs pruner fix

tbenr:
Fixes a bug that was causing the blobs pruner to prune sidecars at the beginning of the first epoch of the DA window.
The error was calculating the DA in slots instead of epochs.

Additionally, this implementation ensure we can frequently prune blobs by adding an epoch of "buffer" during which we can continuously prune blobs in little chunks avoiding to prune blobs of an entire epoch all at once.

## Fixed Issue(s)
will resolve one of the tasks under https://github.com/ConsenSys/teku/pull/6728

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #6667, opened 2023-01-09
Add an override for the number of asyncrunner threads for VC service


corverroos:
### Updated Description

See [this comment](https://github.com/ConsenSys/teku/issues/6667#issuecomment-1379843164) for a better description of the requested feature

### Original Description

We at Obol are running a "performance/load testing" DVT cluster using teku as the VC. The cluster has 500 active validators on goerli. The cluster is sporadically missing attestations due to "inactivity" from teku, ie, the VC doesn't perform some attestation duties. We created an internal ticket: https://github.com/ObolNetwork/charon/issues/1585

The VC logs the following for those missed slots:
```
WARN  - Skipping aggregation for slot 4694957 due to unexpected delay in slot processing
```

Digging into the teku code, I'm wondering if this due to [Serviceutils#calculateMaxThreads](https://github.com/ConsenSys/teku/blob/master/infrastructure/serviceutils/src/main/java/tech/pegasys/teku/service/serviceutils/ServiceConfig.java#L95)?
 - In a DVT cluster, some blocking API calls take much longer than in traditional single validator setups.
 - This is because we need to perform a bunch of internal p2p network hops for consensus across the whole cluster.
 - If other peers in the cluster are having issues, then the data might never be available and those queries need to timeout.
 - If teku is only using `max(Runtime.getRuntime().availableProcessors(), 5)` concurrent threads, then it seems likely that this is causing the problem.

Would it be possible to increase this number or make it configurable?

### Steps to Reproduce (Bug)

Run a Obol DVT cluster with 4 nodes and 500 active validators on goerli.

### Versions (Add all that apply)
* Software version: 22.12.0
* Java version: 17.0.5
* OS Name & Version: Ubuntu (GKE)
* Docker Version: Unknown (GKE)
* Cloud VM, type, size: GKE node, JAVA_OPTS="-Xmx4g -Xms4g

lucassaldanha:
Hi,

Thanks for your report.

We have nodes running in Goerli with 5,000 validators and we are not seeing this problem. Our nodes have close to 100% inclusion rate, and we have not seen this warning in our logs.

Our node is running on AWS, and has the following configuration:

Instance type: m6a.2xlarge

- 8 vCPUs
- 32 GB RAM
- 12.5 Gbps Network Bandwidth

We are using EBS gp3 storage with the following configuration:

- 3,000 IOPS 

I checked our configuration and we are using:

- -Xmx5g for heap size

[...]

PR #6761, opened 2023-02-01
Adds override for async runner threads in VC service

LukeHackett12:
## PR Description

This pr adds a flag `--Xvalidator-client-runner-threads` that can be used to set the number of asynchronous runner threads used in the validator client service. This will allow for a greater number of concurrent blocking requests to a BN as is required in DVT.

As suggested in the linked issue, we have built and tested a version of this with a higher configured thread count and this has resolved our issue. We are running a Teku VC as part of 1000 validator DVT cluster with no `unexpected delay in slot processing` errors.

## Fixed Issue(s)
Addresses https://github.com/ConsenSys/teku/issues/6667

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
@@ -111,6 +111,15 @@ public class ValidatorOptions {
       fallbackValue = "true")
   private boolean doppelgangerDetectionEnabled = DEFAULT_DOPPELGANGER_DETECTION_ENABLED;
 
+  @Option(
+      names = {"--Xvalidator-client-runner-threads"},
+      paramLabel = "<INTEGER>",
+      showDefaultValue = Visibility.ALWAYS,
+      description = "Set the number of threads for the validator runner",
+      hidden = true,
+      arity = "1")
+  private int runnerThreadNum = ValidatorConfig.DEFAULT_RUNNER_THREAD_NUM;
+
   public void configure(TekuConfiguration.Builder builder) {
     builder.validator(
         config ->
@@ -124,7 +133,8 @@ public class ValidatorOptions {
                         Optional.ofNullable(graffiti), Optional.ofNullable(graffitiFile)))
                 .generateEarlyAttestations(generateEarlyAttestations)
                 .executorMaxQueueSize(executorMaxQueueSize)
-                .doppelgangerDetectionEnabled(doppelgangerDetectionEnabled));
+                .doppelgangerDetectionEnabled(doppelgangerDetectionEnabled)
+                .runnerThreadNum(runnerThreadNum));
     validatorProposerOptions.configure(builder);
     validatorKeysOptions.configure(builder);
   }
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
@@ -112,13 +112,13 @@ public class ValidatorOptions {
   private boolean doppelgangerDetectionEnabled = DEFAULT_DOPPELGANGER_DETECTION_ENABLED;
 
   @Option(
-      names = {"--Xvalidator-client-runner-threads"},
+      names = {"--Xvalidator-client-executor-threads"},
       paramLabel = "<INTEGER>",
       showDefaultValue = Visibility.ALWAYS,
-      description = "Set the number of threads for the validator runner",
+      description = "Set the number of threads for the validator executor",
       hidden = true,
       arity = "1")
-  private int runnerThreadNum = ValidatorConfig.DEFAULT_RUNNER_THREAD_NUM;
+  private int executorThreads = ValidatorConfig.DEFAULT_EXECUTOR_THREADS;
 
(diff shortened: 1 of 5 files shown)


teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -111,6 +111,15 @@ public class ValidatorOptions {
       fallbackValue = "true")
   private boolean doppelgangerDetectionEnabled = DEFAULT_DOPPELGANGER_DETECTION_ENABLED;
 
+  @Option(
+      names = {"--Xvalidator-client-runner-threads"},
```
rolfyone: nit:`--Xvalidator-client-executor-threads`?

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
```diff
@@ -56,6 +56,8 @@ public class ValidatorConfig {
   public static final int DEFAULT_VALIDATOR_REGISTRATION_SENDING_BATCH_SIZE = 100;
   public static final UInt64 DEFAULT_BUILDER_REGISTRATION_GAS_LIMIT = UInt64.valueOf(30_000_000);
 
+  public static final int DEFAULT_RUNNER_THREAD_NUM = 5;
```
rolfyone: `DEFAULT_EXECUTOR_THREADS`

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
```diff
@@ -150,6 +155,7 @@ private ValidatorConfig(
     this.builderRegistrationTimestampOverride = builderRegistrationTimestampOverride;
     this.builderRegistrationPublicKeyOverride = builderRegistrationPublicKeyOverride;
     this.executorMaxQueueSize = executorMaxQueueSize;
+    this.runnerThreadNum = runnerThreadNum;
```
rolfyone: `executorThreads`

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
```diff
@@ -472,6 +484,11 @@ public Builder doppelgangerDetectionEnabled(final boolean doppelgangerDetectionE
       return this;
     }
 
+    public Builder runnerThreadNum(final int runnerThreadNum) {
+      this.runnerThreadNum = runnerThreadNum;
```
rolfyone: Might be worth validating here that we don't do something silly, even though it's a develop option...
it'd just be something like 
```
if (executorThreads < 1 || executorThreads > 5_000) {
throw new invalidConfigurationException("Invalid configuration. --Xvalidator-client-executor-threads must be greater than 0 and less than 5000.")
}
```
I picked 5000 arbitrarily, it could easily limit to something more like 1000 from your test, but something above what you found useful in your situation...
I guess the main thing is that if someone goes and puts a very large number, it may cause other issues.
Note also the check > 0 is only really window dressing, as it will be caught by the max logic. It's nicer that the user knows they've done something silly so I added it in...
The result of this validation would be an exception that stops the client running early (fail fast) so that they can correct the issue and get into a running state...

teku/src/test/java/tech/pegasys/teku/cli/subcommand/ValidatorClientCommandTest.java
```diff
@@ -173,6 +173,29 @@ public void shouldEnableDoppelgangerDetection() {
...
+
+    final String[] args = {
+      "vc", "--network", "minimal", "--Xvalidator-client-runner-threads", "1000"
+    };
+
+    final TekuConfiguration tekuConfig = getTekuConfigurationFromArguments(args);
+
+    assertThat(tekuConfig.validatorClient().getValidatorConfig().getRunnerThreadNum())
+        .isEqualTo(1000);
+  }
```
rolfyone, line 197: If we add the configuration limit, we could have an out of bounds check here too :)

rolfyone (commented):
Thanks for submitting this, it looks pretty good!

I added a couple of rename ideas, I didn't flow them through but most IDE's will recursively change things so it'll probably be pretty trivial to follow that through.
The only other suggestion was a sanity limit - it can be pretty high, just a starting point that we may tweak down the track, just to make sure no one goes and puts in MAX_INT or something bad... Even if you set it to slightly above what you've found useful, it's probably something that we can revisit if someone does find it too limiting.

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
LukeHackett12: Yeah this makes sense. I have added a check like you suggested and a corresponding test. Might make sense for consistency to do this with the `--Xrest-api-validator-threads` option at some point too.

rolfyone: Fair, I think I might have missed that PR, it's something I'd normally suggest to be a little user tolerant and fail fast :)

rolfyone (approved):
LGTM!

PR #6786, opened 2023-02-08
teku-6768 add a development cli flag for disabling listening to bls change topic

### LINKED ISSUE #6768: add a development cli flag for disabling listening to bls change topic
I want to be able to have a flag to toggle listening to bls topic.

This can be a develop only flag, I don't expect to need to use it but it'll be potentially handy in some flood scenarios, or scenarios where a larger provider wants to prioritise only their changes in their blocks.

I was thinking something like
`--Xbls_to_execution_channel_enabled=true` as default

The wrinkle will be if we need to subscribe to send messages, which i expect that we probably do.

lucassaldanha:
## PR Description
- Adds `--Xbls-to-execution-changes-subnet-enabled` option (default to true)
- When disabled, beacon node won't connect to bls_to_execution_changes subnet
- When disabled, beacon node API won't accept publishing bls_to_execution_changes messages

## Fixed Issue(s)
fixes #6768 

## Documentation

## Changelog

  - Did not update changelog because this is a developer option.


rolfyone (approved):
LGTM

rolfyone:
not doc change required for development cli btw...

PR #6790, opened 2023-02-08
Load Trusted Setup from file by reading G1 and G2 points without copying

zilm13:
## PR Description
Solves backlog entry:
- Optimize interface to Kzg library (we should minimize unnecessary data copy)

We expect less bugs with this fix and option to use any format for trusted setup in future, yaml or whatever
## Documentation

## Changelog


infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
```diff
@@ -52,6 +55,8 @@ public KZGCommitment blobToKzgCommitment(final Bytes blob) throws KZGException {
 
   void loadTrustedSetup(String trustedSetup) throws KZGException;
 
+  void loadTrustedSetup(TrustedSetup trustedSetup) throws KZGException;
```
mehdi-aouadi, line 58: Should we expose this? Could be a private method in the implementation imho. Or is this choice made to make testing easier?
I think it's an implementation detail and could be hidden (And that could impact the tests: Load the test data from a file and check that this method is called with the expected parsed data)

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
```diff
@@ -72,7 +74,7 @@ private static Preset getPreset(final int fieldElementsPerBlob) {
                         fieldElementsPerBlob)));
   }
 
-  private Optional<String> loadedTrustedSetup = Optional.empty();
+  private Optional<Integer> loadedTrustedSetupHash = Optional.empty();
```
mehdi-aouadi: Can't we move it up btw?

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
```diff
@@ -14,79 +14,77 @@
...
-    try {
-      return Optional.of(new URL(resource).toURI())
-          .filter(resourceUri -> resourceUri.getScheme().equals(FILE_SCHEME))
-          .map(Paths::get)
-          .filter(Files::exists);
-    } catch (final Exception __) {
-      return Optional.empty();
+      return new TrustedSetup(g1Points, g2Points);
+    } catch (Exception ex) {
+      throw new IOException(String.format("Failed to parse file: %s", filePath));
```
mehdi-aouadi: I'd suggest to add `trusted setup`: `Failed to parse trusted setup file`

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
```diff
@@ -14,79 +14,77 @@
...
+      final List<Bytes48> g1Points = new ArrayList<>();
+      for (int i = 0; i < g1Size; i++) {
+        final Bytes48 g1Point = Bytes48.fromHexString(reader.readLine());
+        g1Points.add(g1Point);
+      }
+      // List of G2 points, one on each new line
+      final List<Bytes> g2Points = new ArrayList<>();
+      for (int i = 0; i < g2Size; i++) {
+        final Bytes g2Point = Bytes.fromHexString(reader.readLine());
+        if (g2Point.size() != 96) {
```
mehdi-aouadi: I'd suggest using the `destinationSize` param of `fromHexString` and have the same check as for the `g1Points`

mehdi-aouadi (changes requested):
Some small changes suggested

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
zilm13: I wanted it to be a part of the API, but let me think again

tbenr:
are you testing parsing failure by passing a broken txt file?

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
zilm13: I'd rather move out `void loadTrustedSetup(String trustedSetup) throws KZGException;` out of the `KZG`'s API than that one. But I don't think we should over-complicate it before it's really needed, so I'd keep both for now.

mehdi-aouadi (approved):
LGTM

PR #6792, opened 2023-02-09
Implemented a BlsToExecutionOperationPool

rolfyone:
- Moved the 'duplicate' handling into the pool itself, out of the validator to reduce complexity

 - added a new suite of tests for the new cache, though the logic is very similar to the SimpleOperationPool.

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright ConsenSys Software Inc., 2023
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.statetransition;
+
+import static tech.pegasys.teku.statetransition.validation.ValidationResultCode.IGNORE;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.hyperledger.besu.plugin.services.MetricsSystem;
+import org.hyperledger.besu.plugin.services.metrics.Counter;
+import org.hyperledger.besu.plugin.services.metrics.LabelledMetric;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.collections.LimitedMap;
+import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
+import tech.pegasys.teku.infrastructure.ssz.SszCollection;
+import tech.pegasys.teku.infrastructure.ssz.SszList;
+import tech.pegasys.teku.infrastructure.ssz.schema.SszListSchema;
+import tech.pegasys.teku.infrastructure.subscribers.Subscribers;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.datastructures.operations.SignedBlsToExecutionChange;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.statetransition.validation.InternalValidationResult;
+import tech.pegasys.teku.statetransition.validation.OperationValidator;
+import tech.pegasys.teku.statetransition.validation.SignedBlsToExecutionChangeValidator;
+import tech.pegasys.teku.statetransition.validation.ValidationResultCode;
+
+public class BlsToExecutionOperationPool implements OperationPool<SignedBlsToExecutionChange> {
+  private static final Logger LOG = LogManager.getLogger();
+  private static final int DEFAULT_OPERATION_POOL_SIZE = 50_000;
+  final Map<UInt64, SignedBlsToExecutionChange> operations;
+  private final Function<UInt64, SszListSchema<SignedBlsToExecutionChange, ?>>
+      slotToSszListSchemaSupplier;
+  private final OperationValidator<SignedBlsToExecutionChange> operationValidator;
+  private final Subscribers<OperationAddedSubscriber<SignedBlsToExecutionChange>> subscribers =
+      Subscribers.create(true);
+  private final LabelledMetric<Counter> validationReasonCounter;
+
+  public BlsToExecutionOperationPool(
+      final String metricType,
+      final MetricsSystem metricsSystem,
+      final Function<UInt64, SszListSchema<SignedBlsToExecutionChange, ?>>
+          slotToSszListSchemaSupplier,
+      final SignedBlsToExecutionChangeValidator operationValidator) {
+
+    this.slotToSszListSchemaSupplier = slotToSszListSchemaSupplier;
+    this.operations = LimitedMap.createSynchronized(DEFAULT_OPERATION_POOL_SIZE);
+    this.operationValidator = operationValidator;
+    metricsSystem.createIntegerGauge(
+        TekuMetricCategory.BEACON,
+        OPERATION_POOL_SIZE_METRIC + metricType,
+        "Current number of operations in the pool",
+        this::size);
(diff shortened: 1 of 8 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+    return add(item, true);
+  }
+
+  @Override
+  public void addAll(final SszCollection<SignedBlsToExecutionChange> items) {
+    items.forEach(
+        item -> {
+          final int validatorIndex = getValidatorIndex(item);
+          operations.putIfAbsent(validatorIndex, item);
```
lucassaldanha, line 155: QUESTION

Do you think we should consider re-broadcasting the messages in this case? I don't think we were doing it before, but maybe it makes sense?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
```diff
@@ -0,0 +1,198 @@
...
+  private SafeFuture<InternalValidationResult> add(
+      SignedBlsToExecutionChange item, boolean fromNetwork) {
+    final int validatorIndex = getValidatorIndex(item);
+    return operationValidator
+        .validateForGossip(item)
+        .thenApply(
+            result -> {
+              validationReasonCounter.labels(result.code().toString()).inc();
+              if (result.code().equals(ValidationResultCode.ACCEPT)
+                  || result.code().equals(ValidationResultCode.SAVE_FOR_FUTURE)) {
```
lucassaldanha, line 187: I believe for bls_to_execution_change operations we should never see SAVE_FOR_FUTURE (if we do it is a bug). We could remove it from this if clause now that we have a specific pool for bls ops.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
```diff
@@ -0,0 +1,198 @@
...
+import tech.pegasys.teku.infrastructure.subscribers.Subscribers;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.datastructures.operations.SignedBlsToExecutionChange;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.statetransition.validation.InternalValidationResult;
+import tech.pegasys.teku.statetransition.validation.OperationValidator;
+import tech.pegasys.teku.statetransition.validation.SignedBlsToExecutionChangeValidator;
+import tech.pegasys.teku.statetransition.validation.ValidationResultCode;
+
+public class BlsToExecutionOperationPool implements OperationPool<SignedBlsToExecutionChange> {
```
lucassaldanha, line 46: NIT

Some methods like `subscribeOperationAdded` and `getItemsForBlock` are the same on `SimpleOperationPool` and `BlsToExecutionOperationPool`.

Not only that but the metrics initialization looks the same.

We could remove some of this duplication by having an abstract pool in the middle.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPoolTest.java
```diff
@@ -0,0 +1,265 @@
...
+    assertThat(pool.getItemsForBlock(state)).isEmpty();
+  }
+
+  @Test
+  void shouldAddEntryToPool() {
+    when(validator.validateForGossip(any()))
+        .thenReturn(SafeFuture.completedFuture(InternalValidationResult.ACCEPT));
+    final SignedBlsToExecutionChange item = dataStructureUtil.randomSignedBlsToExecutionChange();
+    final SafeFuture<?> future = pool.addLocal(item);
+    assertThat(future.isCompletedNormally()).isTrue();
```
lucassaldanha, line 83: NIT

Could use:
```
assertThat(future).isCompleted();
```

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPoolTest.java
```diff
@@ -0,0 +1,265 @@
...
+    assertThat(pool.size()).isEqualTo(1);
+    assertThat(pool.getAll()).containsExactly(item);
+  }
+
+  @Test
+  void shouldRejectDuplicateEntryFromPool() throws ExecutionException, InterruptedException {
+    final SignedBlsToExecutionChange item = initPoolWithSingleItem();
+
+    final SafeFuture<InternalValidationResult> future2 = pool.addLocal(item);
```
lucassaldanha, line 92: Where is `future1`? :P

lucassaldanha: Jokes aside, probably rename to 'future' :)

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPoolTest.java
```diff
@@ -0,0 +1,265 @@
...
+    assertThat(future2.get().code()).isEqualTo(InternalValidationResult.IGNORE.code());
```
lucassaldanha, line 93: NIT

Could use:
```
assertThat(future2).isCompletedWithValueMatching(InternalValidationResult::isIgnore);
```

Or, if we want to be certain we are ignoring the message based on the expected reason, we could test if the ignore description matches what we would expect:

```
  final InternalValidationResult expectedValidationResult =
      InternalValidationResult.ignore("BlsToExecutionChange is not the first one for validator %s.",
      item.getMessage().getValidatorIndex());
  
  assertThat(future2).isCompletedWithValue(expectedValidationResult);
```

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPoolTest.java
```diff
@@ -0,0 +1,265 @@
...
+    final int maxBlsToExecutionChanges =
+        spec.getGenesisSpecConfig().toVersionCapella().orElseThrow().getMaxBlsToExecutionChanges();
+    while (pool.size() < maxBlsToExecutionChanges) {
+      assertThat(pool.addLocal(dataStructureUtil.randomSignedBlsToExecutionChange())).isCompleted();
+    }
+
+    assertThat(pool.getItemsForBlock(state)).hasSize(maxBlsToExecutionChanges);
+    assertThat(pool.size()).isEqualTo(maxBlsToExecutionChanges);
+    assertThat(pool.getItemsForBlock(state)).contains(item);
+  }
```
lucassaldanha, line 141: I am confused by this test case. Are we testing that we can fill the pool or that `getItemsForBlock()` respects the maximum number of operations in the block?

The reason I am confused is that the pool size is defined by `DEFAULT_OPERATION_POOL_SIZE` but in the test case we are using `spec.getGenesisSpecConfig().toVersionCapella().orElseThrow().getMaxBlsToExecutionChanges()`.

It looks like we want to test `getItemsForBlock()`. Maybe it is the name of the test that is misleading?

lucassaldanha (approved):
LGTM. A couple of questions (no blockers) and some NIT suggestions.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
rolfyone: I've just left this because it was already in the queue, but i can remove it?

lucassaldanha: I don't think our bls validator will ever return SAVE_FOR_FUTURE.

And the spec for bls to execution change does not specify any path for it. So I'd say we are better with you that.

Not a biggie but removing code we don't need is always a win!

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
rolfyone: potentially, but map vs. set does complicate things... I'm not against refactoring it, it's just going to be messy ...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
rolfyone: nope, totally doesnt make sense here imo. can discuss why if you like...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/BlsToExecutionOperationPool.java
rolfyone: okies, i just was planning on being 'the same' as the simple pool...

PR #6795, opened 2023-02-09
Add Capella to `RestBuilderClientTest` spec context

StefanBratanov:
## PR Description
Just extended the test to support the Capella data structures.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM.
the only things that concerns me is that the values (especially roots) in the resources should be unique to check that we are not exchanging a value with another.

StefanBratanov:
> LGTM. the only things that concerns me is that the values (especially roots) in the resources should be unique to check that we are not exchanging a value with another.

Agree. Changed them with more random ones.

PR #6797, opened 2023-02-09
Validate `withdrawals_root` when receiving the builder bid

### LINKED ISSUE #6794: Validate `withdrawals_root` when receiving the builder bid
It would be good to perform `withdrawals_root` validation against the local execution payload before signing off on the payload obtained from the builder and fallback to local in case they don't match.

StefanBratanov:
## PR Description

- Pass the local execution payload to the `BuilderBidValidator` and add an additional check there

## Fixed Issue(s)
fixes #6794 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
+++ b/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
@@ -516,7 +516,8 @@ public final class DataStructureUtil {
         new SszPublicKey(randomPublicKey()));
   }
 
-  public ExecutionPayloadHeader randomExecutionPayloadHeader(final SpecVersion specVersion) {
+  public ExecutionPayloadHeader randomExecutionPayloadHeader(
+      final SpecVersion specVersion, final Bytes32 withdrawalsRoot) {
     final SpecConfigBellatrix specConfigBellatrix =
         SpecConfigBellatrix.required(specVersion.getConfig());
     return SchemaDefinitionsBellatrix.required(specVersion.getSchemaDefinitions())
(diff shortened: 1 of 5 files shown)


ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -538,10 +539,14 @@ public ExecutionPayloadHeader randomExecutionPayloadHeader(final SpecVersion spe
                     .baseFeePerGas(randomUInt256())
                     .blockHash(randomBytes32())
                     .transactionsRoot(randomBytes32())
-                    .withdrawalsRoot(this::randomBytes32)
+                    .withdrawalsRoot(() -> withdrawalsRoot)
```
rolfyone, line 542: oh nice - does builder listen to the normal `getPayloadV2` calls? Just wondering how they know what withdrawals to include... sorry for the totally naive question, just looking for what gaps we might have...

rolfyone (approved):
LGTM

tbenr:
so TLDR; is:
once we receive a Bid from the builder, if we have a locally produced payload:
- we do the value comparison (if local el supports it) (fallback: FallbackReason.LOCAL_BLOCK_VALUE_HIGHER)
- we make sure withdrawals_root matches (fallback: FallbackReason.BUILDER_ERROR)

if we don't have locally produced payload, we are optimistic on withdrawals_root check, because proposing using the builder bid is better than not proposing at all. In case of a mismatch the block will be rejected by the network.
 
In any case (local payload available or not), we always check the Bid wrt our CL rules (BlockProcessor::validateExecutionPayload). These are the checks that can be done against the beacon state, independently by our locally produced payload.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
StefanBratanov: Yes, we fork version the `getPayload` calls. I think builders also run nodes and know the current state and which withdrawals to include. There is a certain bit of trust that once you agree on the bid and send the signed blinded block, they will return a valid execution payload. In case of invalid payload, blocks will be rejected as bad blocks are rejected now, so builders won't get any MEV, so they are not incentivized to send you bad things. Even if they do keep sending dodgy stuff, at some point the circuit breaker will kick in and disable the builder flow.

Issue #6800, opened 2023-02-10
Unhide doppelganger detection option


mehdi-aouadi:
The DG detection option should not be hidden anymore (The feature will be released)

PR #6801, opened 2023-02-10
Unhide DG detection option

mehdi-aouadi:
## PR Description
Unhide the doppelganger detection activation option

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6800

## Documentation

## Changelog


rolfyone (approved):
LGTM. the documented flags are development flags which is not optimal, I'm going to take a release candidate given the state of the docs, before this gets merge, and then it'll be consistent.

rolfyone:
one thing that does concern me is this doesnt update any AT, which tends to suggest we have no AT for this feature...

StefanBratanov:
> one thing that does concern me is this doesnt update any AT, which tends to suggest we have no AT for this feature...

It does update `TekuNode` and `TekuValidatorNode` which are used in AT.

rolfyone:
> > one thing that does concern me is this doesnt update any AT, which tends to suggest we have no AT for this feature...
> 
> It does update `TekuNode` and `TekuValidatorNode` which are used in AT.

ah ok! That's fine then :)

PR #6802, opened 2023-02-10
Fix: too strict remote transition settings check

zilm13:
## PR Description
I've not found any confirmation of previous check in spec, maybe I'm missing something. Without fix I cannot participate in testnets with bellatrix epoch == 0, ttd == 0. 

## Documentation

## Changelog


zilm13:
Ok, I finally come to what it should look like. At first we don't need rule `remoteTransitionConfiguration.getTerminalBlockHash().isZero()
                    == remoteTransitionConfiguration.getTerminalBlockNumber().isZero()`, we need it only when terminal block hash is zero (see [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675)), other invariant is acceptable. 
```
Note: If TERMINAL_BLOCK_HASH is stubbed with 0x0000000000000000000000000000000000000000000000000000000000000000 then TERMINAL_BLOCK_HASH and TERMINAL_BLOCK_NUMBER parameters MUST NOT take an effect.
```
And it's not a error, we just ignore it, so I modified it to be warning

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/MergeTransitionConfigCheck.java
```diff
@@ -106,16 +106,14 @@ private synchronized void verifyTransitionConfiguration() {
                       .equals(remoteTransitionConfiguration.getTerminalTotalDifficulty())
                   || !localTransitionConfiguration
                       .getTerminalBlockHash()
-                      .equals(remoteTransitionConfiguration.getTerminalBlockHash())) {
+                      .equals(remoteTransitionConfiguration.getTerminalBlockHash())
+                  || !localTransitionConfiguration
+                      .getTerminalBlockNumber()
```
tbenr: I have doubts here. As far as I remember, the local transition configuration doesn't know anything about block number. There is no block number configuration on the CL side. The block number is only configured on EL side so it only appears only in `remoteTransitionConfiguration`

zilm13: OMG I forgot to push it. Fixed

tbenr (approved):
LGTM

PR #6804, opened 2023-02-10
set progressive mode to FULL when capella is defined

tbenr:
change default progressive mode for Capella-enabled networks

## Documentation

## Changelog


ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -276,8 +277,16 @@ public Eth2NetworkConfiguration build() {
...
-                          capellaForkEpoch.ifPresent(capellaBuilder::capellaForkEpoch));
+                      capellaBuilder -> {
+                        capellaForkEpoch.ifPresent(capellaBuilder::capellaForkEpoch);
+
+                        // set progressiveBalancesMode to FULL for all milestones if capella is
+                        // defined and no commandline override has been specified
+                        if (!FAR_FUTURE_EPOCH.equals(capellaBuilder.getCapellaForkEpoch())
+                            && progressiveBalancesMode == DEFAULT_PROGRESSIVE_BALANCES_MODE) {
+                          builder.progressiveBalancesMode(ProgressiveBalancesMode.FULL);
+                        }
```
rolfyone: we could just go all in and skip the if statement, no-one else has a fallback option as far as I can tell...

rolfyone: I guess my reasoning here is the default is the one we'd want to fall back to and this logic wouldn't allow that anyway.... so the CLI flag is almost not allowing us to do the only option we'd want...

tbenr: > .... so the CLI flag is almost not allowing us to do the only option we'd want...

Yep true. Going to remove it..

rolfyone (approved):
LGTM

PR #6809, opened 2023-02-13
Handling BlsToExecutionChange operations on reorgs

### LINKED ISSUE #6791: Improve bls_to_execution_change operation lifecycle in the pool
**Description**

At the moment, we are not adding/removing operations from the pool when a new block is imported or removed due to a reorg, filling our pool over time. It isn't an urgent thing because we do drop older entries as new operations come it, but it adds an unnecessary performance hit whenever we are validating rules that need to check the operations in the pool.

**Solution**

When a new block is imported:
- We need to remove the operations that are in that block from our pool

When a reorg happens:
- We need to add the operations that are being "removed" from the chain back into the operation pool
- We need to remove the operations that are being "included" in the new chain from the operation pool

lucassaldanha:
## PR Description
- Add mew non-canonical operations back into operation pool
- Remove new canonical operations from operation pool

## Fixed Issue(s)
fixes https://github.com/ConsenSys/teku/issues/6791

## Documentation

## Changelog


rolfyone (approved):
LGTM

Issue #6810, opened 2023-02-14
Unhide Xbeacon-liveness-tracking-enabled


mehdi-aouadi:
Unhide the `Xbeacon-liveness-tracking-enabled` option

PR #6812, opened 2023-02-14
Unhide liveness tracking option

mehdi-aouadi:
## PR Description
Unhide the `beacon-liveness-tracking-enabled` option

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6810

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/BeaconRestApiOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/BeaconRestApiOptions.java
@@ -112,15 +112,13 @@ public class BeaconRestApiOptions {
     this.maxUrlLength = maxUrlLength;
   }
 
-  // beacon-liveness-tracking-enabled
   @Option(
-      names = {"--Xbeacon-liveness-tracking-enabled"},
+      names = {"--beacon-liveness-tracking-enabled"},
       paramLabel = "<BOOLEAN>",
       showDefaultValue = Visibility.ALWAYS,
       description = "Track validator liveness and enable requests to the liveness rest api.",
       arity = "0..1",
-      fallbackValue = "true",
-      hidden = true)
+      fallbackValue = "true")
   private Boolean beaconLivenessTrackingEnabled =
       BeaconRestApiConfig.DEFAULT_BEACON_LIVENESS_TRACKING_ENABLED;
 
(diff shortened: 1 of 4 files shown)


mehdi-aouadi:
Reminder to update the inventories once merged. This option is used on some Goerli nodes: https://github.com/ConsenSys/protocols-aws-inventories/blob/master/teku-goerli.yml

rolfyone:
> Reminder to update the inventories once merged. This option is used on some Goerli nodes: https://github.com/ConsenSys/protocols-aws-inventories/blob/master/teku-goerli.yml

i guess my first question would be if we even need them running it in the frist place...

mehdi-aouadi:
> > Reminder to update the inventories once merged. This option is used on some Goerli nodes: https://github.com/ConsenSys/protocols-aws-inventories/blob/master/teku-goerli.yml
> 
> i guess my first question would be if we even need them running it in the frist place...

I'll probably use them for a last test after the release and then remove them

teku/src/main/java/tech/pegasys/teku/cli/options/BeaconRestApiOptions.java
```diff
@@ -112,15 +112,13 @@ public void setMaxUrlLength(int maxUrlLength) {
     this.maxUrlLength = maxUrlLength;
   }
 
-  // beacon-liveness-tracking-enabled
   @Option(
-      names = {"--Xbeacon-liveness-tracking-enabled"},
+      names = {"--beacon-liveness-tracking-enabled"},
```
rolfyone: lets make this able to accept both and avoid some of the dramas from block-timing-tracking

mehdi-aouadi: Done. But that means that the `--Xbeacon-liveness-tracking-enabled` will be printed in the help menu too (since it's not hidden anymore)

rolfyone: we can still document that its deprecated and will be removed later, just hoping that it reduces confusion - not 100% sold either way, whether we rip the bandaid off and answer all the questions, or whether better off hoping not too many people are currently using it...

rolfyone: we should probably have a team discussion around how we want to handle these generally, as a lot of people got caught out by the block timings one...

rolfyone (approved):
im confused at where this is at, but i think this is the right option. just put removing the -X as an upcoming breaking change rather than a breaking change. Because we're making them both work this can just be the note to say its available now, and an upcoming breaking.

PR #6815, opened 2023-02-14
Add `blob_sidecar` gossip topic

### LINKED ISSUE #6777: [DECOUPLING] Add blob_sidecar gossip topic
The `blob_sidecar` topic is used to propagate blobs sidecars to all nodes on the networks.

StefanBratanov:
## PR Description
Just adding everything necessary for the new gossip topic

https://github.com/arnetheduck/eth2.0-specs/blob/back-to-the-decoupled-future/specs/deneb/p2p-interface.md#blob_sidecar_index

## Fixed Issue(s)
fixes #6777 

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsDeneb.java
```diff
@@ -81,11 +86,15 @@ public GossipForkSubscriptionsDeneb(
...
         signedBlsToExecutionChangeOperationProcessor);
 
-    this.gossipedBlockAndBlobsProcessor = gossipedBlockAndBlobsProcessor;
+    this.blobSidecarProcessor = blobSidecarProcessor;
+    this.blockAndBlobsProcessor = blockAndBlobsProcessor;
   }
 
   @Override
   protected void addGossipManagers(final ForkInfo forkInfo) {
+    // TODO: change this to super.addGossipManagers(forkInfo) when the coupled block and blobs
```
mehdi-aouadi, line 95: I don't understand why we should call `super.addGossipManagers(forkInfo)` instead. The Gossip Managers are instantiated separately is each respective method.

mehdi-aouadi (changes requested):

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsDeneb.java
StefanBratanov: Because we can leverage that Deneb extends Capella and can just call super to initialize all previous forks from gossip. But because previous version of Deneb replaced the block gossip with the block_and_blobs_sidecar, we couldn't use super. Now they are separate, we can simplify it.

mehdi-aouadi (approved):

mehdi-aouadi (changes requested):
As you mentioned, the indexing is missing

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/BlobSidecarGossipManager.java
```diff
@@ -0,0 +1,168 @@
...
+public class BlobSidecarGossipManager implements GossipManager {
+
+  private final GossipNetwork gossipNetwork;
+  private final GossipEncoding gossipEncoding;
+
+  private final Int2ObjectMap<Eth2TopicHandler<SignedBlobSidecar>> indexToTopicHandler =
+      new Int2ObjectOpenHashMap<>();
+  private final Int2ObjectMap<TopicChannel> indexToChannel = new Int2ObjectOpenHashMap<>();
+
+  public BlobSidecarGossipManager(
```
tbenr: to me this is a too complex constructor. I'd move the logic to a static create

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsDeneb.java
```diff
@@ -101,9 +110,24 @@ protected void addGossipManagers(final ForkInfo forkInfo) {
...
+    blobSidecarGossipManager =
+        new BlobSidecarGossipManager(
+            recentChainData,
+            spec,
+            asyncRunner,
+            discoveryNetwork,
+            gossipEncoding,
+            forkInfo,
+            blobSidecarProcessor,
+            getMessageMaxSize());
```
tbenr, line 127: we should double check which size we should apply here

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/BlobSidecarGossipManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+  @BeforeEach
+  public void setup() {
+    storageSystem.chainUpdater().initializeGenesis();
+    // return TopicChannel mock for each blob_sidecar_<index> topic
+    doAnswer(
+            i -> {
+              final String topicName = i.getArgument(0);
+              final TopicChannel topicChannel = mock(TopicChannel.class);
+              final Pattern topicPattern = Pattern.compile("blob_sidecar_(\\d+)");
```
tbenr: you can have this pattern as a static class var

tbenr (approved):
LGTM. Nice work. Just a couple of things

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsDeneb.java
StefanBratanov: I had a look and GOSSIP_MAX_SIZE_BELLATRIX and MAX_CHUNK_SIZE_BELLATRIX haven't been changed yet for Deneb

mehdi-aouadi (approved):
LGTM

Issue #6776, opened 2023-02-07
[DECOUPLING] Add BlobSidecar container


mehdi-aouadi:
Notes from [Design notes for decoupled EIP4844](https://hackmd.io/cmYisgxkRuGe9NjX4gr97A)
Spec: https://github.com/ethereum/consensus-specs/pull/3244

```
class BlobSidecar(Container):
    beacon_block_root: Root
    beacon_block_slot: Slot
    proposer_index: ValidatorIndex
    blob_index: uint64 # We need the index of the blob, otherwise we cannot do the gossip protections
    block_parent_root: Root
    blob: Blob
    kzg_commitment: KZGCommitment
    kzg_proof: KZGProof # Allows for quick verification of kzg_commitment

class SignedBlobSidecar(Container):
    message: BlobSidecar
    signature: Signature
```

PR #6818, opened 2023-02-15
6776 blob side car container

mehdi-aouadi:
## PR Description

Add `BlobSidecar` containers and schemas

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/6776

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/schema/deneb/SignedBeaconBlockAndBlobSidecar.java
```diff
@@ -0,0 +1,54 @@
+/*
```
StefanBratanov: I don't think we need `SignedBeaconBlockAndBlobSidecar`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/versions/deneb/BlobSidecar.java
```diff
@@ -0,0 +1,96 @@
+/*
```
StefanBratanov, line 1: Would be good to add a property test similar to `BlobsSidecarPropertyTest`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/versions/deneb/BlobSidecar.java
```diff
@@ -0,0 +1,96 @@
...
+        SszKZGProof> {
+
+  BlobSidecar(final BlobSidecarSchema blobSidecarSchema, final TreeNode backingTreeNode) {
+    super(blobSidecarSchema, backingTreeNode);
+  }
+
+  public BlobSidecar(
+      BlobSidecarSchema schema,
+      Bytes32 blockRoot,
+      UInt64 index,
```
StefanBratanov, line 46: Not sure if we should create a [BlobIndex](https://github.com/arnetheduck/eth2.0-specs/blob/back-to-the-decoupled-future/specs/deneb/beacon-chain.md#custom-types) custom type instead of using UInt64, but seems fine to me. Do you have an opinion @tbenr

StefanBratanov (changes requested):
Should also add the [SignedBlobSidecar](https://github.com/arnetheduck/eth2.0-specs/blob/back-to-the-decoupled-future/specs/deneb/p2p-interface.md#signedblobsidecar) container and a property test

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/versions/deneb/BlobSidecar.java
mehdi-aouadi: Yep, I hesitated to create a new type but then thought `UInt64` could be fine...

data/serializer/src/main/java/tech/pegasys/teku/api/schema/deneb/SignedBeaconBlockAndBlobSidecar.java
mehdi-aouadi: You're right. Removed it (Replaced it by `SignedBlobSidecar`)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/versions/deneb/BlobSidecar.java
mehdi-aouadi: Done

mehdi-aouadi:
> Should also add the [SignedBlobSidecar](https://github.com/arnetheduck/eth2.0-specs/blob/back-to-the-decoupled-future/specs/deneb/p2p-interface.md#signedblobsidecar) container and a property test

Added the missing `SignedBlobSidecar` container

StefanBratanov (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/versions/deneb/BlobSidecar.java
zilm13: we don't  have `Slot`, `Epoch`, `CommitteeIndex`, `ValidatorIndex` and a lot of others. So I think we don't need dedicated class for this one too

Issue #6778, opened 2023-02-07
[DECOUPLING] Implement BlobSidecarByRange RPC method


mehdi-aouadi:
BlobsSidecarsByRange v1
Protocol ID: `/eth2/beacon_chain/req/blobs_sidecars_by_range/1/`

Request Content:

```
(
  start_slot: Slot
  count: uint64
)
Response Content:

class BlobSidecars(Container):
    sidecars: List[BlobSidecar, MAX_BLOBS_PER_BLOCK] # TODO verify constant name

(
  List[BlobSidecars, MAX_REQUEST_BLOBS_SIDECARS]
)
```

PR #6835, opened 2023-02-17
6778 blob sidecars by range

mehdi-aouadi:
## PR Description
Add `BlobSidecarsByRange` RPC method
[SPEC](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blobsidecarsbyrange-v1)

## Fixed Issue(s)

#6778

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
```diff
@@ -317,6 +322,33 @@ public SafeFuture<Void> requestBlobsSidecarsByRange(
...
   }
 
+  @Override
+  public SafeFuture<Void> requestBlobSidecarsByRange(
+      final UInt64 startSlot, final UInt64 count, final RpcResponseListener<BlobSidecar> listener) {
+    return rpcMethods
+        .blobSidecarsByRange()
+        .map(
+            method -> {
+              final UInt64 firstSupportedSlot = firstSlotSupportingBlobsSidecarsByRange.get();
```
StefanBratanov: I would rename the Supplier to `firstSlotSupportingBlobSidecarsByRange`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRangeRequestMessage.java
```diff
@@ -0,0 +1,72 @@
...
+  public UInt64 getCount() {
+    return getField1().get();
+  }
+
+  public UInt64 getMaxSlot() {
+    return getStartSlot().plus(getCount().minus(1));
+  }
+
+  @Override
+  public int getMaximumRequestChunks() {
```
zilm13: I see that it's an old interface, but I'd rename it to `getMaximumResponseChunks` for clarity. @tbenr what do you think?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -35,17 +36,20 @@ public class ThrottlingSyncSource implements SyncSource {
...
       final SyncSource delegate,
       final int maxBlocksPerMinute,
-      final int maxBlobsSidecarsPerMinute) {
+      final int maxBlobsSidecarsPerMinute,
+      final int maxBlobSidecarsPerMinute) {
     this.asyncRunner = asyncRunner;
     this.delegate = delegate;
     this.blocksRateTracker = new RateTracker(maxBlocksPerMinute, 60, timeProvider);
     this.blobsSidecarsRateTracker = new RateTracker(maxBlobsSidecarsPerMinute, 60, timeProvider);
+    this.blobSidecarsRateTracker = new RateTracker(maxBlobSidecarsPerMinute, 60, timeProvider);
```
zilm13: I think we could move this number to constant, already used 3 times

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -76,6 +80,18 @@ public SafeFuture<Void> requestBlobsSidecarsByRange(
...
 
+  @Override
+  public SafeFuture<Void> requestBlobSidecarsByRange(
+      final UInt64 startSlot, final UInt64 count, final RpcResponseListener<BlobSidecar> listener) {
+    if (blobSidecarsRateTracker.wantToRequestObjects(count.longValue()) > 0) {
+      LOG.debug("Sending request for {} blob sidecars", count);
+      return delegate.requestBlobSidecarsByRange(startSlot, count, listener);
+    } else {
+      return asyncRunner.runAfterDelay(
+          () -> requestBlobSidecarsByRange(startSlot, count, listener), Duration.ofSeconds(3));
```
zilm13: I think we could move this number to constant, already used 3 times

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+  private final Counter totalBlobSidecarsRequestedCounter;
+
+  public BlobSidecarsByRangeMessageHandler(
+      final Spec spec,
+      final UInt64 denebForkEpoch,
+      final MetricsSystem metricsSystem,
+      final RecentChainData recentChainData,
+      final CombinedChainDataClient combinedChainDataClient,
+      final UInt64 maxRequestSize,
+      final UInt64 maxBlobsPerBlock) {
```
zilm13, line 64: What if this number will be changed in some fork after Deneb? It really could. No changes are needed at this moment, just thinking out loud

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+  public Optional<RpcException> validateRequest(
+      final String protocolId, final BlobSidecarsByRangeRequestMessage request) {
+    final UInt64 finalizedEpoch = recentChainData.getFinalizedEpoch();
+    final UInt64 minEpochForBlobSidecars =
+        recentChainData
+            .getCurrentEpoch()
+            .orElse(UInt64.ZERO)
+            .minusMinZero(MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS);
+    final UInt64 minimumRequestEpoch =
+        finalizedEpoch.max(minEpochForBlobSidecars).max(denebForkEpoch);
```
zilm13: This rule with finalized_epoch cap is used only for `BlobSidecarsByRoot` from my understanding. It doesn't mentioned in `BlobSidecarsByRange` spec part and there should be some way to get Blobs at least up to `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS`
My understanding of this is that we use byRoot requests only for very recent missing blobs and byRange for the whole range.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+  @Override
+  public Optional<RpcException> validateRequest(
+      final String protocolId, final BlobSidecarsByRangeRequestMessage request) {
+    final UInt64 finalizedEpoch = recentChainData.getFinalizedEpoch();
+    final UInt64 minEpochForBlobSidecars =
+        recentChainData
+            .getCurrentEpoch()
+            .orElse(UInt64.ZERO)
+            .minusMinZero(MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS);
+    final UInt64 minimumRequestEpoch =
```
zilm13: I'd put this thing to `miscHelper` or something like this, it's really needed for everything with blobs. What do you think?
```
final UInt64 minEpochForBlobSidecars =
        recentChainData
            .getCurrentEpoch()
            .orElse(UInt64.ZERO)
            .minusMinZero(MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS);
return minEpochForBlobSidecars.max(denebForkEpoch);
```

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+        finalizedEpoch.max(minEpochForBlobSidecars).max(denebForkEpoch);
+
+    if (!verifyRequestedEpochIsInSupportedRange(request.getStartSlot(), minimumRequestEpoch)) {
```
zilm13: From spec: If any root in the request content references a block earlier than minimum_request_epoch, peers **MAY** respond with error code 3: ResourceUnavailable

I think we should leave some space here for archive nodes or smth like this

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+      final BlobSidecarsByRangeRequestMessage message,
+      final ResponseCallback<BlobSidecar> callback) {
+    final UInt64 startSlot = message.getStartSlot();
+    LOG.trace(
+        "Peer {} requested {} blob sidecars starting at slot {}.",
+        peer.getId(),
+        message.getCount(),
+        startSlot);
+
+    if (!peer.wantToMakeRequest()
```
zilm13, line 96: I don't like this old API too. Do you think it's clear? I think something like 
`if (!peer.allowedToMakeRequest()
        || !peer.allowedToReceiveBlobSidecars(
            callback, maxRequestSize.min(message.getCount()).longValue()))`
could be clearer, but not ideal too

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+  private final UInt64 maxRequestSize;
+  private final UInt64 maxBlobsPerBlock;
+  private final LabelledMetric<Counter> requestCounter;
+  private final Counter totalBlobSidecarsRequestedCounter;
+
+  public BlobSidecarsByRangeMessageHandler(
+      final Spec spec,
+      final UInt64 denebForkEpoch,
+      final MetricsSystem metricsSystem,
+      final RecentChainData recentChainData,
```
zilm13: Do we really need it here?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+      callback.completeWithUnexpectedError(error);
+    }
+  }
+
+  private boolean verifyRequestedEpochIsInSupportedRange(
+      final UInt64 requestedSlot, final UInt64 minimumRequestEpoch) {
+    return spec.computeEpochAtSlot(requestedSlot).isGreaterThanOrEqualTo(minimumRequestEpoch);
+  }
+
+  private class RequestState {
```
zilm13: This iterator-like thing is very complex and I think it could be better if we test it separately. I couldn't put all the logic together to be pretty sure it's not buggy somewhere. It really worth to be tested thoughtfully from my side.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    @NotNull
+    private SafeFuture<Optional<BlobSidecar>> retrieveBlobSidecar() {
+      SafeFuture<Optional<BlobSidecar>> blobSidecar =
+          maybeCurrentBlock
+              .map(SignedBeaconBlock::getRoot)
+              .map(
+                  blockRoot ->
+                      combinedChainDataClient.getBlobSidecarBySlotIndexAndBlockRoot(
```
zilm13: Are we getting only confirmed blobs here? @tbenr ?
I don't understand it completely yet

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerFactory.java
```diff
@@ -67,6 +67,7 @@ public Eth2Peer create(final Peer peer, final BeaconChainMethods rpcMethods) {
         PeerChainValidator.create(spec, metricsSystem, chainDataClient, requiredCheckpoint),
         new RateTracker(peerRateLimit, 60, timeProvider),
         new RateTracker(peerRateLimit, 60, timeProvider),
+        new RateTracker(peerRateLimit, 60, timeProvider),
```
zilm13: Maybe put 60 to some constant?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
```diff
@@ -317,6 +322,33 @@ public SafeFuture<Void> requestBlobsSidecarsByRange(
...
+              final BlobSidecarsByRangeRequestMessage request;
+              if (startSlot.isLessThan(firstSupportedSlot)) {
+                LOG.debug(
+                    "Requesting blob sidecars from slot {} instead of slot {} because the request is spanning the Deneb fork transition",
+                    firstSupportedSlot,
+                    startSlot);
+                final UInt64 updatedCount =
+                    count.minusMinZero(firstSupportedSlot.minusMinZero(startSlot));
+                request = new BlobSidecarsByRangeRequestMessage(firstSupportedSlot, updatedCount);
```
zilm13, line 369: What if `updatedCount` is 0? Is it possible? Will we send a request? Some peer could penalize for such behavior, we'd better not

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/generator/ChainBuilder.java
```diff
@@ -130,6 +141,10 @@ public Optional<BlobsSidecar> getBlobsSidecar(final Bytes32 blockRoot) {
     return Optional.ofNullable(blobsSidecarsByHash.get(blockRoot));
   }
 
+  public Optional<BlobSidecar> getBlobSidecar(final Bytes32 blockRoot) {
+    return Optional.ofNullable(blobSidecarsByHash.get(blockRoot));
```
zilm13, line 152: The could be more than one in every block

zilm13 (commented):
I've not looked at `BlobSidecarsByRangeMessageHandlerTest` yet but it may need some changes too in a line with spec logic

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerFactory.java
mehdi-aouadi: Done

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
mehdi-aouadi: Done

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: What would happen if it changes? It's retrieved from the spec config so we will have nothing to do

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/generator/ChainBuilder.java
mehdi-aouadi: Done. The `ChainBuilder` could be refactored further when adding integration tests later on

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
mehdi-aouadi: Actually it could be 0 and it was a potential bug (the `getMaxSlot` in `BlobsSidecarsByRangeRequestMessage` and `BlobSidecarsByRangeRequestMessage` could throw an `ArithmeticException`).
Fixed that and added some tests.
Good catch!

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: Actually this validation is not required. I removed it

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: Yep you're right. I removed the request validation

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: Not needed anymore since I removed the request validation. Plus if we export the min epoch calculation to the misc helper it won't be needed anymore.
I removed it

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: Yep that would be helpful.
I removed the request validation anyway...

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: FYI this is not yet implemented...

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: Done: Tried to simplify it and added a test for it

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: I agree. Not sure `want` is the right verb here. I'd rename `wantToMakeRequest`  to `peerRateLimitReached` and `wantToReceiveBlobSidecars` to `blobSidecarsRequestsLimitReached`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRangeRequestMessage.java
mehdi-aouadi: I think it would be better to do it in a separate PR if we agree to do so

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
StefanBratanov: I think we don't have to pass the slot since we have the root from the block so can just use getBlobSidecarByBlockRootAndIndex(). I am using the same method in the other RPC method. These could change anyways when we implement the storage so better to have one for now I think.

mehdi-aouadi: You're right! I'll rename it and remove the slot from the params

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -0,0 +1,283 @@
...
+      }
+    }
+
+    boolean isComplete() {
+      return currentSlot.get().isGreaterThan(maxSlot)
+          || maxRequestSize.isLessThanOrEqualTo(sentBlobSidecars.get());
+    }
+
+    void incrementSlotAndIndex() {
+      if (currentIndex.get().equals(maxBlobsPerBlock.minus(UInt64.ONE))) {
```
StefanBratanov, line 249: I am not sure about that actually, there could be less blobs for a block, we need to check how many kzg commitments are in the block and then decide what indexes to retrieve. I think it would be better though to retrieve all blob sidecars for block root in one method call in the combinedChainDataClient (sth like getBlobSidecarsByBlockRoot() which will return all blob sidecars we have in the DB for given root. It will avoid this index incrementation. @tbenr @zilm13 do you have a view on this?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRangeRequestMessage.java
tbenr: yeah so this is the maximum chunks expected in the response for the given the request. So I agree with Dmitry but not super important.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
tbenr: Dmitry is right. You're statically initializing this handler with deneb config, while this parameter is design to be adjusted in new milestones, so it will change at runtime when transitioning to future milestones. Not required now.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
tbenr: we can have a local private method that hides these two calls. like `doesTriggerRateLimit(callback, maxRequestSize.min(message.getCount()).longValue()))`.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
tbenr: what I don't like is always passing from the block. For finalized blobs if we stream blobs by slot we only get blobs we need (we can safely assume it).

For non finalized we still can stream like that and in case we have multiple block roots for the same slot (we have forks) than we drop non-canonical ones.
The alternative is always pass from blocks in for non finalized.


If we look at `BeaconBlocksByRangeMessageHandler` there are these logics in place.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
tbenr: @zilm13 yes the apis must filter out unconfirmed Sidecars (even if currently we don't do that)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
tbenr: but I agree that this should be addressed in a separate PR since this is a common pattern we use in other RPC methods too. I we want to change the peer interface we can log an issue and make a consistent change later.

zilm13: Yeah, it's just a first time I see code here in gossip, and I've noticed this. I agree that this should be addressed in separate RP

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
zilm13: Yeah, I've noticed it too. My opinion is leave it as is now and leave a backlog entry to make an optimization skipping non-existed blobs. I think it's too early to focus on performance

zilm13: @StefanBratanov @mehdi-aouadi 
I've tried to rework it from the other view. Not tested, but should be close to production.
https://github.com/zilm13/teku/commit/208e424ba40c56c9ad9074c1309e6e4e168b5ae5
What do you think?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -58,7 +66,7 @@ public SafeFuture<Void> requestBlocksByRange(
       return delegate.requestBlocksByRange(startSlot, count, listener);
     } else {
       return asyncRunner.runAfterDelay(
-          () -> requestBlocksByRange(startSlot, count, listener), Duration.ofSeconds(3));
+          () -> requestBlocksByRange(startSlot, count, listener), Duration.ofSeconds(DELAY));
```
zilm13: Maybe put whole `Duration.ofSeconds(3)` to constant?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -72,7 +80,19 @@ public SafeFuture<Void> requestBlobsSidecarsByRange(
...
+
+  @Override
+  public SafeFuture<Void> requestBlobSidecarsByRange(
+      final UInt64 startSlot, final UInt64 count, final RpcResponseListener<BlobSidecar> listener) {
+    if (blobSidecarsRateTracker.wantToRequestObjects(count.longValue()) > 0) {
+      LOG.debug("Sending request for {} blob sidecars", count);
+      return delegate.requestBlobSidecarsByRange(startSlot, count, listener);
+    } else {
+      return asyncRunner.runAfterDelay(
+          () -> requestBlobSidecarsByRange(startSlot, count, listener), Duration.ofSeconds(DELAY));
```
zilm13: shouldn't we decrease count here to `blobSidecarsRateTracker.wantToRequestObjects(count.longValue())` if it's smaller? I see that in blocks we ignore it, but it sounds as good behavior from my side

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSourceTest.java
```diff
@@ -129,6 +151,26 @@ void shouldDelayRequestIfBlobsSidecarsLimitAlreadyExceeded() {
...
   }
 
+  @Test
+  void shouldDelayRequestIfBlobSidecarsLimitAlreadyExceeded() {
+    final UInt64 count = UInt64.valueOf(MAX_BLOB_SIDECARS_PER_MINUTE);
+    ignoreFuture(source.requestBlobSidecarsByRange(UInt64.ZERO, count, blobSidecarsListener));
+    ignoreFuture(
+        source.requestBlobSidecarsByRange(UInt64.valueOf(100), count, blobSidecarsListener));
+
+    // Both requests happen immediately
```
zilm13: I have a gut feeling that comment "// Both requests happen immediately" was copied from first "shouldRequestImmediately" test and now is copied to all other tests including throttled. Someone should stop this :)
So I think it's correct only in immediate test cases

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRangeRequestMessage.java
```diff
@@ -0,0 +1,72 @@
...
+    extends Container2<BlobSidecarsByRangeRequestMessage, SszUInt64, SszUInt64>
+    implements RpcRequest {
+
+  public static class BlobSidecarsByRangeRequestMessageSchema
+      extends ContainerSchema2<BlobSidecarsByRangeRequestMessage, SszUInt64, SszUInt64> {
+
+    public BlobSidecarsByRangeRequestMessageSchema() {
+      super(
+          "BlobSidecarsByRangeRequestMessage",
+          namedSchema("startSlot", SszPrimitiveSchemas.UINT64_SCHEMA),
```
zilm13: Let's do it in a snake case, like in [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blobsidecarsbyrange-v1). I've fixed it in other classes to avoid confusion any more: https://github.com/ConsenSys/teku/pull/6863

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandlerTest.java
```diff
@@ -193,6 +194,27 @@ public void shouldSendToPeerRequestedNumberOfBlobsSidecars() {
     assertThat(actualSent).containsExactlyElementsOf(expectedSent);
   }
 
+  @Test
+  public void shouldIgnoreRequestWhenCountIsZero() {
```
zilm13: no need to to fix anything here I think as we are going to remove it

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerTest.java
```diff
@@ -189,6 +191,35 @@ public void shouldModifyRequestSpanningTheEip4844ForkTransition() {
...
+    final ArgumentCaptor<BlobsSidecarsByRangeRequestMessage> requestCaptor =
+        ArgumentCaptor.forClass(BlobsSidecarsByRangeRequestMessage.class);
+
+    verify(delegate, times(1)).sendRequest(any(), requestCaptor.capture(), any());
+
+    final BlobsSidecarsByRangeRequestMessage request = requestCaptor.getValue();
+
+    // Deneb starts from epoch 1, so request start slot should be 8 and the count should be 0
+    assertThat(request.getStartSlot()).isEqualTo(UInt64.valueOf(8));
+    assertThat(request.getCount()).isEqualTo(UInt64.valueOf(0));
```
zilm13: So we've sent request with 0 count to the peer, right? 
We should avoid it

networking/eth2/src/testFixtures/java/tech/pegasys/teku/networking/eth2/peers/StubSyncSource.java
```diff
@@ -99,6 +121,12 @@ public void assertRequestedBlobsSidecars(final long startSlot, final long count)
         .contains(new Request(UInt64.valueOf(startSlot), UInt64.valueOf(count)));
   }
 
+  @SuppressWarnings("unused")
+  public void assertRequestedBlobSidecars(final long startSlot, final long count) {
```
zilm13: I've not found how it's used

zilm13 (commented):
Sorry if I sound too picky on this, our network behavior could lead to bad relations with other clients and I want to have a good relations :)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
mehdi-aouadi: Done

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRangeRequestMessage.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: I'll consider this OK for now then

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: I logged an issue for that: https://github.com/ConsenSys/teku/issues/6872

networking/eth2/src/testFixtures/java/tech/pegasys/teku/networking/eth2/peers/StubSyncSource.java
mehdi-aouadi: Removed

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerTest.java
mehdi-aouadi: This test was initially added to test that handle the case when count is zero.
I refactored the code to return without sending a request when the count is 0

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRangeRequestMessage.java
mehdi-aouadi: Done

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSourceTest.java
mehdi-aouadi: Done

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
mehdi-aouadi: As discussed, I created a separate issue for the requests throttling: https://github.com/ConsenSys/teku/issues/6874

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
mehdi-aouadi: Done

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandlerTest.java
mehdi-aouadi: My bad! Mixed up.
Removed

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: @zilm13 The approach you suggested makes sense. Using a data structure could be better. I suggest keeping that for a separate PR. Would that be OK for you?
@StefanBratanov @tbenr Do we keep it like that for now and improve it in a separate PR or do you prefer considering the kzg commitments count to avoid requesting non existing blobs?

StefanBratanov (changes requested):
Btw we need to add `blobSidecarsByRange.ifPresent(allMethods::add)` in the `BeaconChainMethods` constructor

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
mehdi-aouadi: As discussed, I logged two separate issues:
https://github.com/ConsenSys/teku/issues/6880
https://github.com/ConsenSys/teku/issues/6881
I'll keep this approach for now and we could improve it later.

mehdi-aouadi:
> Btw we need to add `blobSidecarsByRange.ifPresent(allMethods::add)` in the `BeaconChainMethods` constructor

Good catch. I added it ;)

StefanBratanov (approved):
LGTM. Will improve the iteration and fetching of blob sidecars in a next PR.

PR #6840, opened 2023-02-20
Move excess_data_gas to the end of Payload

zilm13:
## PR Description
Update according to https://github.com/ethereum/consensus-specs/pull/3218 and
https://github.com/ethereum/execution-apis/pull/365 .
I guess, reference tests will fail, so we may need to postpone this one until spec release.
Also, check correctness of `TransactionStoreTest` fix
## Documentation

## Changelog


storage/src/test/java/tech/pegasys/teku/storage/store/StoreTransactionTest.java
```diff
@@ -404,24 +404,14 @@ public void getOrderedBlockRoots_withNewBlocks() {
...
-      assertThat(tx.getOrderedBlockRoots())
-          .containsExactly(
-              genesis.getRoot(),
-              forkBlock2.getRoot(),
-              mainChainBlock1.getRoot(),
-              mainChainBlock3.getRoot(),
-              mainChainBlock4.getRoot());
-    }
+    // Childrens are ordered by slot
+    assertThat(tx.getOrderedBlockRoots())
```
tbenr, line 408: Not sure about this. the order is not based on slot but on parent->child relationship.
Since `forkBlock2` and `mainChainBlock1` are siblings (both descending from genesis block) they can be picked indifferently (so the root order is applied in this case)

tbenr: I think It is possible to modify the test to hit that condition.

zilm13: I've followed the structure we are querying. The structure is TreeMap with Slot keys, so should be ordered by slots

tbenr: Ah yes, the TreeMap is used when tx data is not empty.

tbenr (approved):
LGTM

PR #6842, opened 2023-02-21
Rename remaining EIP4844 references to Deneb

StefanBratanov:
## PR Description

- Now that reference tests are updated and deneb renaming is part of the specs, can remove the remaining references to EIP-4844
- Removed the TODO from `BlobsSidecarsByRangeMessageHandler` and `FetchBlockAndBlobsSidecarTask` because these classes will be removed when implementing the blob decoupling

## Fixed Issue(s)
N/A

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobsSidecarsByRangeMessageHandler.java
```diff
@@ -131,9 +131,7 @@ public void onIncomingMessage(
   }
 
   /**
-   * TODO: Update link for Deneb
-   *
-   * <p>See <a
+   * See <a
    * href="https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#blobssidecarsbyrange-v1">BlobsSidecarsByRange
```
tbenr, line 135: you haven't updated the link here

StefanBratanov: that would be deleted and there is no reference to it now with the new spec, so just decided to leave it there

tbenr: oh but it is the pre-decoupling one. Never mind.

tbenr (approved):
LGTM

Issue #6702, opened 2023-01-18
Implement getBlockRewards API method


lucassaldanha:
**Description**

This method retrieves block reward info for a single block.

Reference: [Beacon API spec](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Rewards/getBlockRewards)

**Tasks**
- Implement API interface based on spec (returning some mock data)
- Implement real logic of fetching/calculating block rewards
- Changelog update

PR #6845, opened 2023-02-21
Create GetBlockRewards API

courtneyeh:
#6702 

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/migrated/BlockRewardData.java
```diff
@@ -0,0 +1,62 @@
...
+  private final String proposerSlashings;
+  private final String attesterSlashings;
+
+  public BlockRewardData(
+      String proposerIndex,
+      String total,
+      String attestations,
+      String syncAggregate,
+      String proposerSlashings,
+      String attesterSlashings) {
```
rolfyone: We'll ultimately want these as uint64 or long I think... i know the resulting type is string but we don't want to have to do string conversion for object creation...

rolfyone: this will alter object creation but shouldnt alter the data in the json definition

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/rewards/GetBlockRewards.java
```diff
@@ -0,0 +1,75 @@
...
+  public static final String ROUTE = "/eth/v1/beacon/rewards/blocks/{block_id}";
+
+  private static final SerializableTypeDefinition<BlockRewardData> DATA_TYPE =
+      SerializableTypeDefinition.object(BlockRewardData.class)
+          .withField("proposer_index", STRING_TYPE, BlockRewardData::getProposerIndex)
+          .withField("total", STRING_TYPE, BlockRewardData::getTotal)
+          .withField("attestations", STRING_TYPE, BlockRewardData::getAttestations)
+          .withField("sync_aggregate", STRING_TYPE, BlockRewardData::getSyncAggregate)
+          .withField("proposer_slashings", STRING_TYPE, BlockRewardData::getProposerSlashings)
+          .withField("attester_slashings", STRING_TYPE, BlockRewardData::getAttesterSlashings)
```
rolfyone: this is the bit that will change if we change the object, we should just make it correct now i think...

courtneyeh: Okay, should I make them all UInt64?

rolfyone: tempted to say long, need to know if they can be negative in this instance, im guessing htey cant but would need to look at the spec...

courtneyeh: True, another option would be to leave them strings for now. Then change when implement?

data/serializer/src/main/java/tech/pegasys/teku/api/migrated/BlockRewardData.java
```diff
@@ -0,0 +1,64 @@
...
+    return syncAggregate;
+  }
+
+  public UInt64 getProposerSlashings() {
+    return proposerSlashings;
+  }
+
+  public UInt64 getAttesterSlashings() {
+    return attesterSlashings;
+  }
```
rolfyone, line 64: nit: i'd just go ahead and add equals, toString in this object now, you'll want it for clean test reporting...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/rewards/GetBlockRewards.java
```diff
@@ -0,0 +1,75 @@
...
+            .pathParam(PARAMETER_BLOCK_ID)
+            .response(SC_OK, "Request successful", RESPONSE_TYPE)
+            .withNotFoundResponse()
+            .withInternalErrorResponse()
+            .build());
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
+    throw new NotImplementedException();
```
rolfyone, line 74: try to call this, if anyone calls it i think it'll crash... we'd need to list it as a possible response as far as i recall

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/rewards/GetBlockRewardsTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Test
+  void metadata_shouldHandle404() throws JsonProcessingException {
+    verifyMetadataErrorResponse(handler, HttpStatusCodes.SC_NOT_FOUND);
+  }
+
+  @Test
+  void metadata_shouldHandle500() throws JsonProcessingException {
+    verifyMetadataErrorResponse(handler, SC_INTERNAL_SERVER_ERROR);
+  }
```
rolfyone, line 66: one way to test would be add 501 here and see if it just dies...

rolfyone (approved):
LGTM once integration test includes 501

PR #6846, opened 2023-02-21
Updated to support 1.3.0.RC3 ref tests for capella

rolfyone:
## Documentation

## Changelog


zilm13:
Execution payload things are fixed in my PR https://github.com/ConsenSys/teku/pull/6840
I'm not sure what is a best way to commit all. Both are broken being alone

rolfyone:
> Execution payload things are fixed in my PR #6840 I'm not sure what is a best way to commit all. Both are broken being alone

I'll make ours compliant then leave it at rc2, I tried to cherry-pick yours and it didnt look clean, so will let you deal with whatever is going on there...

lucassaldanha (approved):

PR #6854, opened 2023-02-23
Update external signer block signing integration test

usmansaleem:
## PR Description
Update external signer block signing integration test. Use parameterized test for Bellatrix and Capella block signing.
## Documentation

## Changelog

### DIFF
--- a/validator/client/build.gradle
+++ b/validator/client/build.gradle
@@ -46,4 +46,6 @@ dependencies {
   testImplementation testFixtures(project(':infrastructure:serviceutils'))
   testImplementation testFixtures(project(':infrastructure:restapi'))
 
+  integrationTestRuntimeOnly "org.apache.logging.log4j:log4j-slf4j2-impl"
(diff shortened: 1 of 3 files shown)


rolfyone:
Are we planning on putting more tests in? i don't see why we'd do this restructuring currently...

i'd probably prefer its an abstract base class and we don't have to do `data.` everywhere like the `AbstractDataBackedIntegrationTest`  as a concrete example...

usmansaleem:
@rolfyone The primary intention is to use parameterized test for bellatrix, capella and onward specs for blocksigning.

rolfyone:
Have a quick look at 
```
  @ParameterizedTest
  @MethodSource("allNetworksWithAllMilestones")
```
And make sure, but I think you're better off not breaking all of that information out potentially...

usmansaleem:
Closing, will open a new PR with consolidated integration test for external signer.

Issue #6701, opened 2023-01-18
Implement getSyncCommitteeRewards API method


lucassaldanha:
**Description**

This method retrieves rewards info for sync committee members specified by array of public keys or validator index. If no array is provided, return reward info for every committee member.

Reference: [Beacon API spec](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Rewards/getSyncCommitteeRewards)

**Tasks**
- Implement API interface based on spec (returning some mock data)
- Implement real logic of fetching/calculating sync committee rewards
- Changelog update

PR #6859, opened 2023-02-23
Add calculateRewards method in ChainDataProvider

courtneyeh:
#6701

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
+++ b/data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
@@ -31,6 +31,7 @@ import static tech.pegasys.teku.infrastructure.unsigned.UInt64.ZERO;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -388,6 +390,24 @@ public void getStateSyncCommittees_shouldRejectFarFutureEpoch() {
         .isCompletedExceptionallyWith(IllegalArgumentException.class);
   }
 
+  @Test
+  public void calculateRewards_shouldGetData() {
```
rolfyone, line 394: this is good, what happens when we ask for a validator index that's not in the committee? can we demonstrate that in a unit test?

rolfyone (approved):
LGTM, suggested another unit test to push an edge case.

data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
courtneyeh: Handled in next [PR](https://github.com/ConsenSys/teku/pull/6860)

PR #6860, opened 2023-02-24
Add getCommitteeIndices method to ChainDataProvider

courtneyeh:
#6701 

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
@@ -563,15 +563,14 @@ public class ChainDataProvider {
     }
 
     final Map<Integer, Integer> output = new HashMap<>();
-    int i = 0;
-    for (BLSPublicKey key : committeeKeys) {
+    for (int i = 0; i < committeeKeys.size(); i++) {
+      final BLSPublicKey key = committeeKeys.get(i);
       final Optional<Integer> validatorIndex = spec.getValidatorIndex(state, key);
       if (validatorIndex.isPresent()
           && (validators.contains(key.toHexString())
               || validators.contains(validatorIndex.get().toString()))) {
         output.put(i, validatorIndex.get());
       }
-      i++;
     }
 
     return output;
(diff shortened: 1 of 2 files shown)


data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -390,6 +391,89 @@ public void getStateSyncCommittees_shouldRejectFarFutureEpoch() {
...
+
+    final Set<String> validators =
+        Set.of(
+            data.randomPublicKey().toHexString(), // Validator not in committee
+            "2", // Validator not in committee
+            committeeKeys.get(2).toHexString());
+    final Map<Integer, Integer> committeeIndices =
+        provider.getCommitteeIndices(committeeKeys, validators, state);
+
+    assertThat(committeeIndices).containsExactlyInAnyOrderEntriesOf(Map.of(2, 6));
```
rolfyone: if the specified validator isn't in a committee, why do we get a map entry?

courtneyeh: because one of them is in the committee

rolfyone: oh this might be way too many 2's in the example, i think you're also adding committee member 2... add committee member 7 or something random...

courtneyeh: I commented which ones were added that are not in the committee (and they were not added)

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -545,6 +546,37 @@ private StateSyncCommitteesData getSyncCommitteesFromState(
...
+              .collect(toList());
+
+      return IntStream.range(0, result.size())
+          .boxed()
+          .collect(Collectors.<Integer, Integer, Integer>toMap(Function.identity(), result::get));
+    }
+
+    final Map<Integer, Integer> output = new HashMap<>();
+    int i = 0;
+    for (BLSPublicKey key : committeeKeys) {
```
rolfyone: i think this would be cleaner if you make a for loop in a more traditional way
```
for (int i = 0; i < comitteeKeys.length(); i++) {
```
and instead of key use `committeeKeys.get(i)`
or even do final `BlsPublicKey key = committeeKeys.get(i);` as first line potentially

rolfyone (approved):
minor nits, looks good over all

data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
rolfyone: sure, but 2 everywhere becomes confusing and your brain associates it...

PR #6863, opened 2023-02-24
Make ssz containers field names snake case, like in spec

zilm13:
## PR Description
We've left some containers with camel case field names, but in spec and in other container classes we use snake case for them. Fixed.
## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #6822, opened 2023-02-16
Implement Blob Decoupling


StefanBratanov:
[EIP-4844: Free the blobs #3244](https://github.com/ethereum/consensus-specs/pull/3244)

#### SSZ Containers
- [x] #6776 

#### KZG
- [x] #6782 

#### Gossip
- [x] #6777 

#### RPC
- [x] #6778 
- [x] #6821 
- [x] #6830 
- [x] #6834
- [x] #6870 
- [x] #7246

#### Block Import and State Transition
- [x] #6848
- [x] #6849
- [x] #7080
- [x] #7110
- [x] #7112

#### Validator Coordinator
- [x] Revisit block production

#### Sync
- [x] #6948 
- [x] #6985 
- [x] #7010 
- [x] #7243
- [x] #7244

#### Engine API
- [x] https://github.com/ethereum/execution-apis/pull/402

#### Builder API
- [x] #7170
	
#### Storage
- [x] #6837
- [x] #7111
	
#### Beacon API
- [x] #6781 
- [x] #6819
- [x] #7012
- [x] #7013
- [x] #7019
- [x] #7048
- [x] #7075

#### Validator Client
- [x] #6967
- [x] #6959
- [x] #6963
- [x] #6966
- [x] Remote call from VC -> BN for creating/sending `BlockContents`

#### Cleanup
- [x] #6831
- [x] #6832
- [x] #6833
- [x] Remove old way of historical and recent block sync with coupled blobs
- [x] Remove old way of forward sync with coupled blobs
- [x] Remove `TODO: remove when blobs decoupling sync is implemented` and `TODO: remove once the block publishing is switched to a decoupling version`

#### Nice to have
- [x] #6880
- [x] #6881

StefanBratanov:
Finally closing this since all tasks are done :)

PR #6864, opened 2023-02-24
BlobSidecar gossip validation

tbenr:
- Reverts `BlockManager` to deal only with blocks (reverts our coupled deneb implementation)
- Makes `BlobSidecarManager` the handler for the gossip messages
- Introduces `BlobSidecarValidator` to implement gossip rules
- Introduces a `GossipValidationHelper` containing some common functions (block and blob sidecars) to implement gossip validation
- Use `GossipValidationHelper` in `BlockValidation`

Two tests in `BlobsSidecarsByRangeIntegrationTest` has been disabled because this PR temporary breaks the blobSidecar storage

## Documentation

## Changelog


beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/gossip/FetchRecentBlocksServiceTest.java
```diff
@@ -79,11 +79,7 @@ public void setup() {
             maxConcurrentRequests);
 
     lenient().when(fetchBlockTaskFactory.create(any(), any())).thenAnswer(this::createMockTask);
-    recentBlockFetcher.subscribeBlockFetched(
-        (block, blobsSidecar) -> {
-          blobsSidecar.ifPresent(importedBlobsSidecars::add);
```
mehdi-aouadi, line 84: Can't we get rid of `importedBlobsSidecars` since it's always empty?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarValidator.java
```diff
@@ -0,0 +1,219 @@
...
+
+  public SafeFuture<InternalValidationResult> validate(final SignedBlobSidecar signedBlobSidecar) {
+    final BlobSidecar blobSidecar = signedBlobSidecar.getBlobSidecar();
+
+    /*
+    [REJECT] The sidecar is for the correct topic -- i.e. sidecar.index matches the topic {index}.
+    */
+    /*
+    This rule is already implemented in
+    tech.pegasys.teku.networking.eth2.gossip.BlobSidecarGossipManager.TopicIndexAwareOperationProcessor
```
mehdi-aouadi, line 93: Wondering if we should double check here... Or find a way to enforce all the rules in a single place

mehdi-aouadi (commented):
We should probably rename `BlobsSidecarManager` to `BlobSidecarsManager` (and the methods signatures).
Overall LGTM. We have to be careful with the TODOs. I'd suggest logging some entries in the decoupling issue to not forget about them #6822

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarValidator.java
tbenr: the problem is that apply the rule we need the topic name which is not exposed to the OperationProcessor. I think this is a very particular case and we can accept to just reference this as implemented internally by the gossip layer.
If there are others complaining about this let's further discuss :)

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/gossip/FetchRecentBlocksServiceTest.java
tbenr: yeah.. let's clean it up for now, it will readded later if required

mehdi-aouadi (approved):
LGTM

StefanBratanov (approved):
LGTM

PR #6865, opened 2023-02-26
Create getSyncCommitteeRewardsFromBlockId logic in ChainDataProvider

courtneyeh:
#6701

## Documentation

## Changelog


data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -546,6 +547,61 @@ private StateSyncCommitteesData getSyncCommitteesFromState(
...
+    }
+
+    final tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState state =
+        maybeState.get();
+
+    final UInt64 epoch = spec.computeEpochAtSlot(block.getSlot());
+    final UInt64 slot = spec.computeStartSlotAtEpoch(epoch);
+
+    final Optional<SyncCommittee> maybeCommittee =
+        spec.getSyncCommitteeUtil(slot).map(util -> util.getSyncCommittee(state, epoch));
```
rolfyone: we can just use the slot of the block here, then delete the slot variable ...
or store the block.getSlot() as slot if you think that looks better, but ultimately no need to `spec.computeStartSlotAtEpoch`

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -546,6 +547,61 @@ private StateSyncCommitteesData getSyncCommitteesFromState(
...
+    if (maybeCommittee.isEmpty()) {
+      return Optional.of(data);
+    }
```
rolfyone: not sure how we can get here, and i doubt we have a test showing it given this method is currently private...

Is this method well tested by the other methods that call it currently?

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -546,6 +547,61 @@ private StateSyncCommitteesData getSyncCommitteesFromState(
...
+
+    final Map<Integer, Integer> committeeIndices =
+        getCommitteeIndices(
+            maybeCommittee.get().getPubkeys().stream()
+                .map(SszPublicKey::getBLSPublicKey)
+                .collect(toList()),
```
rolfyone: extract to variable might look better than doing this inline...

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -546,6 +547,61 @@ private StateSyncCommitteesData getSyncCommitteesFromState(
...
+  }
+
+  private Optional<SyncCommitteeRewardData> getSyncCommitteeRewardData(
+      Set<String> validators,
+      BeaconBlock block,
+      SyncCommitteeRewardData data,
+      Optional<tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState> maybeState) {
+    if (maybeState.isEmpty()) {
+      return Optional.empty();
+    }
```
rolfyone: we go through handling optional inside this method just because the state coming in is optional, maybe its easier to just not call this method if the state is empty?
the upshot of that would be that then this method can just return SyncCommitteeRewardData
it may just mean using `maybeState.map()` where its used...

rolfyone:
mostly looks good, just drilling into that private method a little to see if we can get it cleaner...

rolfyone (approved):
so much better :) good work!

Issue #6784, opened 2023-02-07
Implement engine_exchangeCapabilities


rolfyone:
SPEC https://github.com/ethereum/execution-apis/blob/main/src/engine/common.md#capabilities

PR #6868, opened 2023-02-27
Added method objects for engine_newPayloadV1/V2/V3

lucassaldanha:
## PR Description
- Implemented `EngineNewPayloadV1`, `EngineNewPayloadV2` and `EngineNewPayloadV3`
- Updated both Bellatrix, Capella and Deneb execution client handlers to use the new method-objects

## Fixed Issue(s)
related to #6784 

## Documentation

## Changelog


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/methods/EngineNewPayloadV1.java
```diff
@@ -0,0 +1,63 @@
...
+  public int getVersion() {
+    return 1;
+  }
+
+  @Override
+  public SafeFuture<PayloadStatus> execute(final JsonRpcRequestParams params) {
+    final ExecutionPayload executionPayload =
+        params.getRequiredParameter(0, ExecutionPayload.class);
+
+    LOG.trace("calling engineNewPayloadV1(executionPayload={})", executionPayload);
```
rolfyone, line 49: one thing we could potentially do here is add a check where if the spec exceeds bellatrix, we fail with a meaningful message...

It might be a good pattern to follow, so if v2 is called with deneb we can also fail fast - thoughts?

rolfyone (approved):
LGTM, one thought...

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/methods/EngineNewPayloadV1.java
lucassaldanha: I like this idea. I'll do it later after we add all methods in. Basically, I am avoiding anything that changes the current functionality so we can still use all test as they are to ensure correctness.

PR #6876, opened 2023-02-28
Implemented withdrawals calculation for builder endpoint

### LINKED ISSUE #6858: [Capella] add an endpoint to get withdrawals from a state
https://github.com/ethereum/beacon-APIs/issues/301

rolfyone:
Added test cases, most of the complication is actually in getting a `preState` which generally we don't deal with, but is important with withdrawals, particularly if the initial state is on the other side of an epoch boundary.

Fixes #6858

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

PR #6877, opened 2023-02-28
Handle request for GetSyncCommitteeRewards

courtneyeh:
#6701 

## Documentation

## Changelog


rolfyone:
overall pretty good, lets add an integrationTest based on AbstractDataBackedIntegrationTest (from memory) that has an actual altair chain, and we query it for results...
The test cases:
 - empty list (`[]`) -> all -> just a count would be fine to check
 - empty body -> 400 error
 - `["asdf"]` -> 400 error
 - `["99999999999999"]` (out of validator range) -> empty list returned
 - `["X","Y"]` and X in sync committee, Y not -> x returned in list

something like this anyway... just some negative tests

The point here is we're not mocking, we're fetching data from a chain and seeing it end-2-end being fetched etc.

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/rewards/GetSyncCommitteeRewardsIntegrationTest.java
```diff
@@ -0,0 +1,137 @@
...
+    Response response =
+        post(
+            GetSyncCommitteeRewards.ROUTE.replace("{block_id}", "head"),
+            jsonProvider.objectToJSON(requestBody));
+
+    assertThat(response.code()).isEqualTo(SC_BAD_REQUEST);
+    assertThat(response.body().string())
+        .isEqualTo(
+            "{\"code\":400,\"message\":\"'asdf' is not a valid "
+                + "hex encoded public key or validator index in the committee\"}");
```
rolfyone, line 107: oh i thought it was just index, we should probably check a pubkey works too...

rolfyone (approved):
super close! just a test case for pubkey and we can call this done!

PR #6878, opened 2023-02-28
Introduce option to use local payload instead of builder 

### LINKED ISSUE #6327: Support block value in `engineGetPayload`

zilm13:
## PR Description
Introduces option to use local payload instead of builder if local matches X% of builder's value. X is configured:
- I have some doubts I did it in appropriate options class
- I'm not sure I should unhide it right from the beginning, plus if I do it, I will need to add docs and changelog entry
- I don't like I couldn't pass default value to tests

## Fixed Issue(s)

Fixes #6327 
## Documentation

## Changelog


ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImplTest.java
```diff
@@ -234,9 +235,46 @@ public void builderGetHeaderGetPayload_shouldReturnHeaderAndPayloadViaEngineWhen
...
+        dataStructureUtil.randomPayloadExecutionContext(false, true);
+    final UInt64 slot = executionPayloadContext.getForkChoiceState().getHeadBlockSlot();
+    final BeaconState state = dataStructureUtil.randomBeaconState(slot);
+
+    final UInt256 builderValue =
+        prepareBuilderGetHeaderResponse(executionPayloadContext, false).getValue();
     final ExecutionPayload localExecutionPayload =
-        prepareEngineGetPayloadResponse(executionPayloadContext, UInt256.MAX_VALUE, slot);
+        prepareEngineGetPayloadResponse(
+            executionPayloadContext, builderValue.multiply(51).divide(100), slot);
```
rolfyone, line 277: 50 would be the boundary right? could just multiply by boundary here... i guess point being that we're implementing >=, so boundary is 50 not 51, and we wouldnt be testing boundary at 51, just every other value...

zilm13: It's in percents

zilm13: I wanted to have 2 tests cases side by side one where value is a bit higher and another where it's a bit smaller. And I converted the old one to default value, but it's not passed from the same place in tests like in main, so it's not correct say it's default, which I don't like

rolfyone: ok but the boundary is 50%, so it'd be nice to also test the boundary and make sure >= doesnt turn into >

zilm13: ahhh got your comment now
the reason I didn't make it because it could change with even/odd number
Of course, I could use not random to be sure

zilm13: and we test == on old test, which is left in place

zilm13:
I decided to add feature to use builder's bid whenever is possible not comparing with local value to the same flag with keyword "NEVER". If it totally doesn't work for us, I made it with a single commit, so it could be easily reverted

teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionLayerOptions.java
```diff
@@ -96,6 +97,19 @@ public class ExecutionLayerOptions {
   private int builderCircuitBreakerAllowedConsecutiveFaults =
       DEFAULT_BUILDER_CIRCUIT_BREAKER_ALLOWED_CONSECUTIVE_FAULTS;
 
+  @Option(
+      names = {"--validators-builder-bid-challenge-percentage"},
+      paramLabel = "<STRING>",
+      showDefaultValue = Visibility.ALWAYS,
+      description =
+          "Fallback to local payload, when builder bid is available but its value could be beat by local payload. "
```
tbenr: we can simply say "Fallback to local payload whenever its value beats the builder bid."

teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionLayerOptions.java
```diff
@@ -96,6 +97,19 @@ public class ExecutionLayerOptions {
...
+              + "Value is whole, percent (e.g. 100, default value, means use local payload when it at least matches builder bid, "
+              + "80 means use local payload when its value is at least 80% of builder bid).\n"
+              + "NEVER: ignore local value, use builder's bid whenever it passed validation",
```
tbenr: "passes"?

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -164,9 +168,21 @@ public SafeFuture<HeaderWithFallbackData> builderGetHeader(
...
+                        .lessOrEqualThan(localPayloadValue.multiply(HUNDRED_PERCENTS))) {
+                  LOG.info(
+                      "Using local Execution Payload instead of Builder Bid as it's challenged. "
+                          + "Builder bid value: {}, local block value: {}, builderBidChallengePercentage: {}%",
+                      signedBuilderBid.getMessage().getValue().toDecimalString(),
+                      localPayloadValue.toDecimalString(),
+                      builderBidChallengePercentage);
                   return getResultFromLocalExecutionPayload(
-                      localExecutionPayload, slot, FallbackReason.LOCAL_BLOCK_VALUE_HIGHER);
+                      localExecutionPayload, slot, FallbackReason.LOCAL_BLOCK_VALUE_WIN);
```
tbenr: what about `LOCAL_BLOCK_VALUE_WON` ? similarly to `CIRCUIT_BREAKER_ENGAGED`

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -164,9 +168,21 @@ public SafeFuture<HeaderWithFallbackData> builderGetHeader(
                         .map(ExecutionPayloadWithValue::getValue)
                         .orElse(UInt256.ZERO);
                 logReceivedBuilderBid(signedBuilderBid.getMessage());
-                if (signedBuilderBid.getMessage().getValue().lessOrEqualThan(localPayloadValue)) {
+                // 1 ETH is 10^18 wei, Uint256 max is more than 10^77
+                if (builderBidChallengePercentage.isPresent()
```
tbenr: what about moving this in a `isLocalPayloadValueWinning` function?

tbenr (approved):
LGTM with some nits. And maybe an english native guy can review user-facing strings 😉

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
zilm13: it was the first name. don't know why I reverted it

zilm13:
@tbenr updated up to your feedback. Let's keep it till native speaker for final proof-reading

zilm13:
I've remembered what was wrong when we posted about this feature last time. It will be enabled only since Capella, so we'd better keep it hidden to avoid misleading. Docs etc should be added with Capella release I think. What do you think? (but we may have a lot of debts for this release)

tbenr:
> I've remembered what was wrong when we posted about this feature last time. It will be enabled only since Capella, so we'd better keep it hidden to avoid misleading. Docs etc should be added with Capella release I think. What do you think? (but we may have a lot of debts for this release)

OH yes. very good catch. I was about to fall under the same issue. Even if when we start using the new engine api system that Loucas is working on, we will detach from the Capella fork.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -50,6 +50,7 @@
 public class ExecutionBuilderModule {
 
   private static final Logger LOG = LogManager.getLogger();
+  private static final int HUNDRED_PERCENTS = 100;
```
rolfyone: we shouldnt need to define 100%, but if we do, maybe just `HUNDRED_PERCENT`, as the `s` is confusing :)

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -164,10 +168,18 @@ public SafeFuture<HeaderWithFallbackData> builderGetHeader(
                         .map(ExecutionPayloadWithValue::getValue)
                         .orElse(UInt256.ZERO);
                 logReceivedBuilderBid(signedBuilderBid.getMessage());
-                if (signedBuilderBid.getMessage().getValue().lessOrEqualThan(localPayloadValue)) {
+
+                if (isLocalPayloadValueWinning(signedBuilderBid, localPayloadValue)) {
+                  LOG.info(
+                      "Using local Execution Payload instead of Builder Bid as it's challenged. "
+                          + "Builder bid value: {}, local block value: {}, builderBidChallengePercentage: {}%",
```
rolfyone: `The local execution payload value ({}) was awarded the block over the builder payload ({}), {}% challenge configured.`

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImplTest.java
rolfyone: ok fair :)

rolfyone (approved):
some tiny NIT's, but looks good :)

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
zilm13: Oh, could you explain this, please, I know you are right, but I don't understand this. Is percent always plural?

PR #6884, opened 2023-03-01
Added EthGetBlockByHash and EthGetBlockByNumber

lucassaldanha:
## PR Description
- Implemented `EthGetBlockByHash` and `EthGetBlockByNumber`
- Updated Bellatrix client handler to use the new method-objects

## Fixed Issue(s)
related to #6784 

## Documentation

## Changelog


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/methods/EthGetBlockByHash.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  public String getName() {
+    return "eth_getBlockByHash";
+  }
+
+  @Override
+  public int getVersion() {
+    return Integer.MAX_VALUE;
+  }
```
rolfyone: lets just return 1... or better still, 0....

if version 0 then we could just have the versionedName not include and woulldnt have to override...

rolfyone (approved):
yep that looks good now!

PR #6887, opened 2023-03-01
Add MAX_REQUEST_BLOB_SIDECARS to SpecConfigDeneb

StefanBratanov:
## PR Description
I think this way is cleaner. Also follows a pattern used in SpecConfig (`getMinEpochsForBlocksRequests`). It allows to be overriden in the future for next milestone without having to change the logic in other places.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java
@@ -14,7 +14,6 @@
 package tech.pegasys.teku.networking.eth2.rpc.beaconchain.methods;
 
 import static tech.pegasys.teku.networking.eth2.rpc.core.RpcResponseStatus.INVALID_REQUEST_CODE;
-import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS_DENEB;
 import static tech.pegasys.teku.spec.config.Constants.MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS;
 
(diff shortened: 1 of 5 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java
```diff
@@ -84,15 +83,14 @@ public BlobSidecarsByRootMessageHandler(
   @Override
   public Optional<RpcException> validateRequest(
       final String protocolId, final BlobSidecarsByRootRequestMessage request) {
-    // MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK
-    final int maxRequestBlobSidecars = calculateMaxRequestBlobSidecars();
-    if (request.size() > maxRequestBlobSidecars) {
+    final UInt64 maxRequestBlobSidecars = getMaxRequestBlobSidecars();
+    if (maxRequestBlobSidecars.isLessThan(request.size())) {
```
zilm13: I think `request.size.isGreaterThan(maxRequestBlobSidecars)` is more straightforward way of comparison. Do we break a limit? Not a "is limit smaller than we"?

zilm13 (approved):
LGTM

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java
StefanBratanov: I agree, it's less natural. Changed it.

PR #6891, opened 2023-03-02
Adds sync committee metrics

### LINKED ISSUE #6270: Expose current sync committee and next sync committee per pubkey in the validator prometheus metrics
# 🚀 Feature Request

### Description

Mainly need to know two things (from sync committee perspective):
- which pubkey is in a sync committee currently
- which pubkey will be in a sync committee next

If we're really concerned about data, then as a baseline, even just knowing that the validator has one of the keys in sync committee currently and/or next would be useful to expose via prometheus, so we can plan maintenance at the validator level (which is managing for us ~200-300 pubkeys) accordingly.  

### Describe the solution you'd like

Expose prometheus metric per pubkey in the validator metrics

### Describe alternatives you've considered

I've looked at https://www.coincashew.com/coins/overview-eth/guide-or-how-to-setup-a-validator-on-eth2-mainnet/part-ii-maintenance/checking-my-eth-validators-sync-committee-duties already, but would like to have this via prometheus

### Links:
- [ethstaker teku-helpdesk discord conversation](https://discord.com/channels/694822223575384095/782670046736285697/1025413246217375814)
- similar to https://github.com/prysmaticlabs/prysm/issues/11522
- [Prysmatic Labs code-review channel discord conversation with potuz and prestonvanloon](https://discord.com/channels/476244492043812875/483017808658169866/1025450792112631848)

rolfyone:
`validator_scheduled_sync_committee_duties_current` This new metric is a visualisation to see how many validators are currently assigned to performing sync committee duties.

There were already metrics showing the attestation duties and block production duties, named in the same way, so this was just filling in the gap, although the other duties are accounted in a slightly different way, as they work differently.

This new metric is effectively the number of validators that you have that are in a sync committee currently.

Also added
`validator_current_sync_committee_last_epoch`, which will be the last epoch of the current sync committee.

This will allow validator maintainers to better determine if they are in a sync committee currently, and when the sync committee will be changing.

Fixes #6270

## Documentation

## Changelog


validator/client/src/test/java/tech/pegasys/teku/validator/client/SyncCommitteeDutyLoaderTest.java
```diff
@@ -109,6 +114,58 @@ void shouldRetrieveDuties() {
...
+
+    assertThat(
+            metricsSystem
+                .getGauge(TekuMetricCategory.VALIDATOR, "current_sync_committee_last_epoch")
+                .getValue())
+        .isEqualTo(63.0);
+  }
+
+  @Test
+  void shouldGetCountOfValidatorsInSyncCommitteeThroughMetrics2() {
```
lucassaldanha: Is this method a duplicate of `shouldGetCountOfValidatorsInSyncCommitteeThroughMetrics`?
If yes, then please remove it.
If no, then we need a better name for it because I swear they look exactly the same! :P

lucassaldanha (approved):
LGTM. Just need to deal with the duplicated test.

validator/client/src/test/java/tech/pegasys/teku/validator/client/SyncCommitteeDutyLoaderTest.java
rolfyone: oh yep i remember this, i need to rename it heh

Issue #6870, opened 2023-02-27
Change maximum number of blocks in an RPC request


StefanBratanov:
MAX_REQUEST_BLOCKS_DENEB should be used for Deneb

https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md

This will require changes in the `byRange` and `byRoot` block RPC methods

PR #1061, opened 2019-12-13
Limit outgoing block range request size

cemozerr:
## PR Description
Limit outgoing block range request size

PR #6892, opened 2023-03-02
limit deneb block requests (by range and by root)

mehdi-aouadi:
## PR Description
No more than `MAX_REQUEST_BLOCKS_DENEB` may be requested at a time (`BeaconBlocksByRange` and `BeaconBlocksByRoot`)

## Fixed Issue(s)

#6870 
## Documentation

## Changelog


StefanBratanov (commented):
Hmm think this is not correct. MAX_BLOCK_BY_RANGE_REQUEST_SIZE is used for rate limiting. Think can leave this value as it is for Deneb. We should see the usage of MAX_REQUEST_BLOCKS instead and change it accordingly for the Deneb fork.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
```diff
@@ -73,14 +81,19 @@ public void onIncomingMessage(
...
+              callback, maxRequestSize.min(UInt64.valueOf(message.size())).longValue())) {
         requestCounter.labels("rate_limited").inc();
         return;
       }
-
       requestCounter.labels("ok").inc();
       totalBlocksRequestedCounter.inc(message.size());
-      for (SszBytes32 blockRoot : message) {
+
+      final int count = Math.min(maxRequestSize.intValue(), message.size());
```
zilm13: we can do it before `if (!peer.wantToMakeRequest()` and calculate it only once if I'm not wrong

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/BeaconChainMethods.java
```diff
@@ -13,7 +13,7 @@
 
 package tech.pegasys.teku.networking.eth2.rpc.beaconchain;
 
-import static tech.pegasys.teku.spec.config.Constants.MAX_BLOCK_BY_RANGE_REQUEST_SIZE;
+import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS_DENEB;
```
zilm13, line 16: Do we still need old constant in `PeerSync`?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/BeaconChainMethods.java
```diff
@@ -197,7 +197,8 @@ private static Eth2RpcMethod<GoodbyeMessage, GoodbyeMessage> createGoodBye(
           final PeerLookup peerLookup,
           final RpcEncoding rpcEncoding) {
     final BeaconBlocksByRootMessageHandler beaconBlocksByRootHandler =
-        new BeaconBlocksByRootMessageHandler(spec, metricsSystem, recentChainData);
+        new BeaconBlocksByRootMessageHandler(
+            spec, metricsSystem, recentChainData, MAX_REQUEST_BLOCKS_DENEB);
```
zilm13, line 201: It's a Deneb constant but it will be applied much earlier. However smaller is not bigger, so it looks safe. Just thoughts aloud. I think it's ok

zilm13:
@StefanBratanov is `MAX_BLOCK_BY_RANGE_REQUEST_SIZE` anywhere in specs? I've not find. And 128 is smaller than 200 so we just rate limiting even more

mehdi-aouadi:
> @StefanBratanov is `MAX_BLOCK_BY_RANGE_REQUEST_SIZE` anywhere in specs? I've not find. And 128 is smaller than 200 so we just rate limiting even more

Actually `MAX_BLOCK_BY_RANGE_REQUEST_SIZE` is defined internally for peers sync limitation. Not needed anymore

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/BeaconChainMethods.java
mehdi-aouadi: I think we don't need it anymore. Replaced it with the new DENEB limit

mehdi-aouadi:
> Hmm think this is not correct. MAX_BLOCK_BY_RANGE_REQUEST_SIZE is used for rate limiting. Think can leave this value as it is for Deneb. We should see the usage of MAX_REQUEST_BLOCKS instead and change it accordingly for the Deneb fork.

Actually `MAX_REQUEST_BLOCKS_DENEB` < `MAX_BLOCK_BY_RANGE_REQUEST_SIZE` as @zilm13 mentioned so it doesn't make sense to use different limits in the method and in the peer, the smaller one will be enforced anyway. I added a unit test for the `BeaconBlockByRange` method when the count is greater than `MAX_REQUEST_BLOCKS`.
There is no `MAX_REQUEST_BLOCKS` check in the `BeaconBlockByRoot` because the `BeaconBlocksByRootRequestMessage` cannot contain more than `MAX_REQUEST_BLOCKS` elements anyway.

zilm13 (approved):
LGTM

StefanBratanov (commented):
Hmm, not sure why MAX_BLOCK_BY_RANGE_REQUEST_SIZE was introduced. But there are usages of MAX_REQUEST_BLOCKS still which are definied in phase0 specs https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md. In my mind this constant is getting overriden for Deneb https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md so all usages of MAX_REQUEST_BLOCKS should be replaced by MAX_REQUEST_BLOCKS_DENEB once Deneb is active. Maybe @tbenr can jump in and have a look as well if he agrees with me.

tbenr:
@mehdi-aouadi @StefanBratanov 

MAX_BLOCK_BY_RANGE_REQUEST_SIZE  has been introduced way back here https://github.com/ConsenSys/teku/pull/1061

I agree with @StefanBratanov that we should change at runtime the limit from `MAX_REQUEST_BLOCKS` to `MAX_REQUEST_BLOCKS_DENEB`

mehdi-aouadi:
Done in: https://github.com/ConsenSys/teku/pull/6938

PR #6894, opened 2023-03-02
Fix Gradle warnings

zilm13:
## PR Description
Fix gradle warnings.
There is also a bunch of warnings like 
```
Resolution of the configuration :acceptance-tests:runtimeClasspath was attempted from a context different than the project context. Have a look at the documentation to understand why this is a problem and how it can be resolved. This behaviour has been deprecated and is scheduled to be removed in Gradle 8.0. See https://docs.gradle.org/7.6.1/userguide/viewing_debugging_dependencies.html#sub:resolving-unsafe-configuration-resolution-errors for more details.
```
Looks like not a real bug and every gradle submodule produces one of these errors
https://github.com/gradle/gradle/issues/21492
So, ignoring warnings like this

## Documentation

## Changelog


lucassaldanha (approved):
LGTM. We might need to adjust this if/when we upgrade to Gradle 8.

PR #6896, opened 2023-03-02
Update eventsource

### LINKED ISSUE #6890: Update okhttp-eventsource
### Description
The following dependencies have later milestone versions:
> com.launchdarkly:okhttp-eventsource [3.0.0 -> 4.1.0]
     https://github.com/launchdarkly/okhttp-eventsource

But it has breaking changes, so it requires usage update

zilm13:
## PR Description
The was a major update of the library, trying to applying it to our code. From [release notes](https://github.com/launchdarkly/okhttp-eventsource/releases/tag/4.0.0):
> This release revises the implementation of `EventSource` so that it does not create its own worker threads. The methods for starting and reading a stream now perform synchronous I/O, and the caller is responsible for choosing what thread to start the stream and read the stream on. This makes the core implementation simpler and more efficient, and reduces possibilities for deadlock. If you want to continue using the old asynchronous push model, see the `com.launchdarkly.eventsource.background` package.

So for this PR I've used `com.launchdarkly.eventsource.background` API which also changed but not too much. Do we need to use synchronous API with our own event thread instead? 
Also giving the sensitive part of code this update is touching, I'd better test it on our nodes before merging
## Fixed Issue(s)

Fixes #6890 

## Documentation

## Changelog


StefanBratanov:
LGTM to me after looking at the release notes and the APIs changed. There is also usage of `EventSource` in the acceptance tests so code needs to be changed there as well. Otherwise, I like using the `background` package since it behaves similar way as before and won't be a more drastic change for already well-behaving component.

rolfyone (approved):
LGTM

PR #6902, opened 2023-03-03
Improvements when using VC with failovers

StefanBratanov:
## PR Description

- Rename `RemoteBeaconNodeSyncingChannel` to `BeaconNodeReadinessChannel`
- Don't perform readiness check when no failovers are configured. (it is not needed)
- Immediately going to failover if BN is not ready. (It was doing that before as well but it was using exceptions which was triggering logging and it could be confusing). I think it is more readable now.
- Improve logging a little bit.
- Add timeout for the syncing status call

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
@@ -30,7 +30,6 @@ import org.apache.tuweni.bytes.Bytes32;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import org.hyperledger.besu.plugin.services.metrics.Counter;
 import org.hyperledger.besu.plugin.services.metrics.LabelledMetric;
-import tech.pegasys.teku.api.exceptions.RemoteServiceNotAvailableException;
 import tech.pegasys.teku.api.migrated.ValidatorLivenessAtEpoch;
 import tech.pegasys.teku.api.response.v1.beacon.ValidatorStatus;
 import tech.pegasys.teku.bls.BLSPublicKey;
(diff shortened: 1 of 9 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
```diff
@@ -312,38 +311,58 @@ private <T> SafeFuture<T> relayRequest(
       final String method,
       final boolean relayRequestToFailovers) {
     if (failoverDelegates.isEmpty() || !relayRequestToFailovers) {
-      return runPrimaryRequestWithNoFailovers(request, method);
+      return runPrimaryRequest(request, method);
     }
-    final SafeFuture<T> primaryResponse = runPrimaryRequestWithConfiguredFailovers(request, method);
     final Map<RemoteValidatorApiChannel, Throwable> capturedExceptions = new ConcurrentHashMap<>();
-    final List<SafeFuture<T>> failoversResponses =
+    final List<SafeFuture<T>> failoverResponses =
```
mehdi-aouadi, line 317: Shouldn't we check if the `primaryDelegate` is ready and try to `runPrimaryRequest` first before trying to redirect the request to the `failoverDelegates`?
If the primary node is ready and its request doesn't fail we'd have asked the failovers to run the request for nothing.
In my understanding the failovers should be asked to run the request as a last resort...

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
```diff
@@ -352,88 +371,91 @@ private <T> SafeFuture<T> relayRequest(
...
-          }
-          recordSuccessfulRequest(delegate, method);
-          return SafeFuture.completedFuture(response);
-        });
+    return request
+        .run(delegate)
+        .handleComposed(
+            (response, throwable) -> {
+              if (throwable != null) {
+                recordFailedRequest(delegate, method);
```
mehdi-aouadi, line 456: I'd suggest adding a log here

mehdi-aouadi (changes requested):

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
StefanBratanov: We relay to all failovers for some methods (like sending attestations, validator registrations, etc) regardless if the primary returns a correct response. There is `relayRequest` and `tryRequestUntilSuccess` methods. One of them sends to all, the others tries the primary and then one by one the failovers sorted by readiness

mehdi-aouadi: I see you point. But in my understanding, if we'd want to relay the request to all nodes, either we do not prioritise the primary node and hence send the request to all nodes, or we prioritise it and in that case we send the request primarily to the primary node and we relay to the failovers if it fails only.
In the current implementation, we relay to the request to all failovers, we keep the responses aside, we then send it to the primary node, if it fails we consider the failovers responses otherwise we "ignore" them and return the primary node response. In the last case we'd have engaged the failovers for nothing.
If that's the desired behaviour, that's OK otherwise I think we could optimise it

StefanBratanov: Yeah that's the behaviour. It's not for nothing though. The messages sent to all nodes are subscriptions for subnets, attestations, blocks which will be gossiped by all nodes, validator registrations, beacon proposers information. We ignore their responses unless we don't get a response from the primary node, but we still need to send them to make sure all nodes are prepared to be a failover.

mehdi-aouadi (approved):
LGTM
Thanks for the explanation 🙏

PR #6903, opened 2023-03-05
Create calculateProposerSyncAggregateBlockRewards method in ChainDataProvider

courtneyeh:
#6702 

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
@@ -677,7 +677,7 @@ public class ChainDataProvider {
   }
 
   @VisibleForTesting
-  protected int calculateSyncAggregateBlockRewards(UInt64 proposerReward, SyncAggregate aggregate) {
+  protected int calculateProposerSyncAggregateBlockRewards(UInt64 proposerReward, SyncAggregate aggregate) {
     final SszBitvector syncCommitteeBits = aggregate.getSyncCommitteeBits();
     int total = 0;
     for (int i = 0; i < syncCommitteeBits.size(); i++) {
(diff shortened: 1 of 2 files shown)


data/provider/src/test/java/tech/pegasys/teku/api/ChainDataProviderTest.java
```diff
@@ -548,6 +549,33 @@ public void calculateRewards_shouldGetData() {
...
+    final UInt64 reward = UInt64.valueOf(1234);
+    final Spec spec = TestSpecFactory.createMinimalAltair();
+    final DataStructureUtil data = new DataStructureUtil(spec);
+    final ChainDataProvider provider = setupAltairState();
+    final SyncAggregate syncAggregate =
+        data.randomSyncAggregate(0, 3, 4, 7, 16, 17, 20, 23, 25, 26, 29, 30);
+
+    final int syncAggregateBlockRewards =
+        provider.calculateSyncAggregateBlockRewards(reward, syncAggregate);
+    assertThat(syncAggregateBlockRewards).isEqualTo(reward.times(12).intValue());
```
lucassaldanha: Where does the value 12 come from? Is it the number of participant validators? It is probably worth assigning it to a variable with a name explaining the rationale to make the test more readable.

Another option would be something like this:
```
int[] participantsValidatorIndexes = new int[] {0, 3, 4, 7, 16, 17, 20, 23, 25, 26, 29, 30};
final SyncAggregate syncAggregate = data.randomSyncAggregate(participantsValidatorIndexes);

[...]

assertThat(syncAggregateBlockRewards).isEqualTo(reward.times(participantsValidatorIndexes.length).intValue());
```

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -675,6 +676,18 @@ protected SyncCommitteeRewardData calculateRewards(
     return data;
   }
 
+  @VisibleForTesting
+  protected int calculateSyncAggregateBlockRewards(UInt64 proposerReward, SyncAggregate aggregate) {
```
lucassaldanha: Would be clearer if this method was named `calculateProposerSyncAggregateBlockRewards`?

lucassaldanha (approved):
LGMT. Just some readability suggestions.

PR #6906, opened 2023-03-06
Reduce default allowed faults to 5 for builder circuit breaker

tbenr:
As per some discussion, let's reduce the default default allowed faults to 5 for builder circuit breaker.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #6874, opened 2023-02-27
Peers rate limiting


mehdi-aouadi:
In the `ThrottlingSyncSource` we're supposed to consider the `peerRateLimit` more accurately:
Currently we just check if `wantToRequestObjects(count.longValue()) > 0` and proceed with the request.
A more accurate behaviour would be to respond with `min(count, wantToRequestObjects(count.longValue()))`

zilm13:
Also the same with answer. Currently if we are allowed to answer with at least 1 object we provide full response whatever size is it. It would be more precise to provide only objects left in throttle allowance

zilm13:
Plus I'd refactor `wantToSmth` names. It's now clear. Something like `allowedToSmth` or similar could be better

PR #6907, opened 2023-03-06
6874 peer rate limiting

mehdi-aouadi:
## PR Description
- Rename the peers rate limiting methods
- Calculate the blob sidecars rate limiting at runtime (`MAX_REQUEST_BLOCKS_DENEB * spec.getMaxBlobsPerBlock()`)

## Fixed Issue(s)

#6874 

## Documentation

## Changelog

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
@@ -13,8 +13,8 @@
 
 package tech.pegasys.teku.beacon.sync.forward.multipeer.chains;
 
-import static tech.pegasys.teku.spec.config.Constants.MAX_BLOB_SIDECARS_PER_MINUTE;
-import static tech.pegasys.teku.spec.config.Constants.MAX_BLOCKS_PER_MINUTE;
+import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS;
+import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS_DENEB;
 import static tech.pegasys.teku.spec.config.Constants.SYNC_BATCH_SIZE;
 import static tech.pegasys.teku.spec.config.Constants.SYNC_BLOB_SIDECARS_SIZE;
 
@@ -22,8 +22,11 @@ import java.util.HashMap;
 import java.util.Map;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.time.TimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.networking.eth2.peers.Eth2Peer;
 import tech.pegasys.teku.networking.eth2.peers.SyncSource;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.config.SpecConfigDeneb;
 
 public class SyncSourceFactory {
 
@@ -36,11 +39,20 @@ public class SyncSourceFactory {
     this.timeProvider = timeProvider;
   }
 
-  public SyncSource getOrCreateSyncSource(final Eth2Peer peer) {
+  public SyncSource getOrCreateSyncSource(final Eth2Peer peer, final Spec spec) {
     // Limit request rate to just a little under what we'd accept
-    final int maxBlocksPerMinute = MAX_BLOCKS_PER_MINUTE - SYNC_BATCH_SIZE.intValue() - 1;
+    final int maxBlocksPerMinute = MAX_REQUEST_BLOCKS - SYNC_BATCH_SIZE.intValue() - 1;
+    final int maxBlobsPerBlock =
+        spec.atEpoch(peer.finalizedEpoch())
+            .getConfig()
+            .toVersionDeneb()
+            .map(SpecConfigDeneb::getMaxBlobsPerBlock)
+            .orElse(0);
     final int maxBlobSidecarsPerMinute =
-        MAX_BLOB_SIDECARS_PER_MINUTE - SYNC_BLOB_SIDECARS_SIZE.intValue() - 1;
+        MAX_REQUEST_BLOCKS_DENEB
+            .times(UInt64.valueOf(maxBlobsPerBlock))
+            .minusMinZero(SYNC_BLOB_SIDECARS_SIZE.intValue() - 1)
+            .intValue();
     return syncSourcesByPeer.computeIfAbsent(
         peer,
         source ->
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
@@ -14,16 +14,13 @@
 package tech.pegasys.teku.beacon.sync.forward.multipeer.chains;
 
 import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS;
-import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS_DENEB;
 import static tech.pegasys.teku.spec.config.Constants.SYNC_BATCH_SIZE;
-import static tech.pegasys.teku.spec.config.Constants.SYNC_BLOB_SIDECARS_SIZE;
 
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.time.TimeProvider;
-import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.networking.eth2.peers.Eth2Peer;
 import tech.pegasys.teku.networking.eth2.peers.SyncSource;
 import tech.pegasys.teku.spec.Spec;
@@ -44,18 +41,17 @@ public class SyncSourceFactory {
     // Limit request rate to just a little under what we'd accept
     final int maxBlocksPerMinute = MAX_REQUEST_BLOCKS - SYNC_BATCH_SIZE.intValue() - 1;
 
-    final int maxBlobsPerBlock =
+    final int maxBlobSidecarsPerMinute =
         Optional.ofNullable(spec.forMilestone(SpecMilestone.DENEB))
             .map(
                 specDeneb ->
-                    specDeneb.getConfig().toVersionDeneb().orElseThrow().getMaxBlobsPerBlock())
+                    specDeneb
+                        .getConfig()
+                        .toVersionDeneb()
+                        .orElseThrow()
+                        .getMaxRequestBlobSidecars()
+                        .intValue())
             .orElse(0);
-
-    final int maxBlobSidecarsPerMinute =
-        MAX_REQUEST_BLOCKS_DENEB
-            .times(UInt64.valueOf(maxBlobsPerBlock))
-            .minusMinZero(SYNC_BLOB_SIDECARS_SIZE.intValue() - 1)
-            .intValue();
 
     return syncSourcesByPeer.computeIfAbsent(
         peer,
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.beacon.sync.forward.multipeer.chains;
 
+import static tech.pegasys.teku.spec.config.Constants.MAX_BLOBS_PER_BLOCK;
 import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS;
 import static tech.pegasys.teku.spec.config.Constants.MAX_REQUEST_BLOCKS_DENEB;
 import static tech.pegasys.teku.spec.config.Constants.SYNC_BATCH_SIZE;
@@ -47,7 +48,7 @@ public class SyncSourceFactory {
             .getConfig()
             .toVersionDeneb()
             .map(SpecConfigDeneb::getMaxBlobsPerBlock)
-            .orElse(0);
+            .orElse(MAX_BLOBS_PER_BLOCK);
     final int maxBlobSidecarsPerMinute =
         MAX_REQUEST_BLOCKS_DENEB
             .times(UInt64.valueOf(maxBlobsPerBlock))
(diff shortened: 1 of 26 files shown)


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -56,9 +56,17 @@ public SafeFuture<Void> requestBlocksByRange(
       final UInt64 startSlot,
       final UInt64 count,
       final RpcResponseListener<SignedBeaconBlock> listener) {
-    if (blocksRateTracker.wantToRequestObjects(count.longValue()) > 0) {
-      LOG.debug("Sending request for {} blocks", count);
-      return delegate.requestBlocksByRange(startSlot, count, listener);
+    final UInt64 updatedCount =
+        count.min(UInt64.valueOf(blocksRateTracker.allowedRequestObjectsCount(count.longValue())));
```
Nashatyrev: Wouldn't it be better to update `RateTracker` implementation to return min of requested and remaining requests?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -33,19 +33,19 @@ public RateTracker(
     this.timeProvider = timeProvider;
   }
 
-  // boundary: if a request comes in and remaining capacity is at least 1, then
-  // they can have the objects they request otherwise they get none.
-  public synchronized long wantToRequestObjects(final long objectCount) {
+  public synchronized long allowedRequestObjectsCount(final long objectCount) {
```
Nashatyrev: nit: is there probably any better method name which emphasizes that object state is changed and requests are actually subtracted by this method? Kind of `take/poll/popRequests`  (the similar for `EthPeer` class)
Else the method invocation is pretending to be just a getter and the state mutation is lost inside `if` expression like: 
```java
    if (!peer.allowedToMakeRequest()
        || !peer.allowedToReceiveBlocks(
```
It would be quite easy to make a mistake in future updates IMO

Nashatyrev (commented):
> Then I guess we need to replace that `500` by `MAX_REQUEST_BLOCKS`

Should we also do it in this PR?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
mehdi-aouadi: Actually it already does that. I refactored the code, there were an unnecessary double `min` operation...

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Right, the mutation isn't clear from the methods names. I renamed them to `popObjectRequests`, `popBlockRequests`, `popBlobSidecarRequests`, and `popRequest`

mehdi-aouadi:
> > Then I guess we need to replace that `500` by `MAX_REQUEST_BLOCKS`
> 
> Should we also do it in this PR?

Replaced the 500 limit by `MAX_REQUEST_BLOCKS` for blocks and `MAX_REQUEST_BLOB_SIDECARS` (`MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK`) for blob sidecars. Not sure about the `MAX_BLOBS_PER_BLOCK` though since it's from the spec. I used the peer's finalized epoch to determine the spec config...

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
```diff
@@ -36,11 +39,20 @@ public SyncSourceFactory(final AsyncRunner asyncRunner, final TimeProvider timeP
     this.timeProvider = timeProvider;
   }
 
-  public SyncSource getOrCreateSyncSource(final Eth2Peer peer) {
+  public SyncSource getOrCreateSyncSource(final Eth2Peer peer, final Spec spec) {
     // Limit request rate to just a little under what we'd accept
-    final int maxBlocksPerMinute = MAX_BLOCKS_PER_MINUTE - SYNC_BATCH_SIZE.intValue() - 1;
+    final int maxBlocksPerMinute = MAX_REQUEST_BLOCKS - SYNC_BATCH_SIZE.intValue() - 1;
+    final int maxBlobsPerBlock =
+        spec.atEpoch(peer.finalizedEpoch())
```
zilm13: Assuming we are going to start Deneb soon, or started and not finalized yet, we will have 0 `maxBlobsPerBlock` for all already joined peers. I'd put instead firstDenebMilestoneValue in `orElse`

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -68,9 +76,17 @@ public SafeFuture<Void> requestBlocksByRange(
...
   public SafeFuture<Void> requestBlobSidecarsByRange(
       final UInt64 startSlot, final UInt64 count, final RpcResponseListener<BlobSidecar> listener) {
-    if (blobSidecarsRateTracker.wantToRequestObjects(count.longValue()) > 0) {
-      LOG.debug("Sending request for {} blob sidecars", count);
-      return delegate.requestBlobSidecarsByRange(startSlot, count, listener);
+    final UInt64 updatedCount =
+        UInt64.valueOf(blobSidecarsRateTracker.popObjectRequests(count.longValue()));
+    if (updatedCount.isGreaterThan(UInt64.ZERO)) {
+      if (updatedCount.isLessThan(count)) {
+        LOG.info(
```
zilm13: I'm not sure we really need info for it. Same for the next one in 5 lines

zilm13 (commented):
Few comments

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
```diff
@@ -36,11 +40,20 @@ public SyncSourceFactory(final AsyncRunner asyncRunner, final TimeProvider timeP
     this.timeProvider = timeProvider;
   }
 
-  public SyncSource getOrCreateSyncSource(final Eth2Peer peer) {
+  public SyncSource getOrCreateSyncSource(final Eth2Peer peer, final Spec spec) {
     // Limit request rate to just a little under what we'd accept
-    final int maxBlocksPerMinute = MAX_BLOCKS_PER_MINUTE - SYNC_BATCH_SIZE.intValue() - 1;
+    final int maxBlocksPerMinute = MAX_REQUEST_BLOCKS - SYNC_BATCH_SIZE.intValue() - 1;
```
Nashatyrev: It's not clear why we use `MAX_REQUEST_BLOCKS` as a `maxBlockPerMinute`? May be it makes sense to change `MAX_BLOCKS_PER_MINUTE` constant?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
```diff
@@ -47,7 +48,7 @@ public SyncSource getOrCreateSyncSource(final Eth2Peer peer, final Spec spec) {
             .getConfig()
             .toVersionDeneb()
             .map(SpecConfigDeneb::getMaxBlobsPerBlock)
-            .orElse(0);
+            .orElse(MAX_BLOBS_PER_BLOCK);
```
zilm13: I think one of the things we target now is to reduce number of constants, especially while this one should come from config. I think something like this could be better `final int initMaxBlobs = Optional.ofNullable(spec.forMilestone(SpecMilestone.DENEB)).map(specDeneb -> specDeneb.getConfig().toVersionDeneb().orElseThrow().getMaxBlobsPerBlock()).orElse(0);
` (not tested)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
zilm13: We want tor reduce number of magic constants

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
mehdi-aouadi: Updated the log level to debug

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
mehdi-aouadi: Right! Added a default blobs per block constant set to 4

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
mehdi-aouadi: Sounds right to me. Added it instead of a constant

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
Nashatyrev: nit: that way it looks even more 'magic' to me than before since denotes different things and even different units of measurement.

Nashatyrev (approved):
LGTM

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
mehdi-aouadi: A slight simplification:

```
final int maxBlobSidecarsPerMinute =
        Optional.ofNullable(spec.forMilestone(SpecMilestone.DENEB))
            .map(
                specDeneb ->
                    specDeneb
                        .getConfig()
                        .toVersionDeneb()
                        .orElseThrow()
                        .getMaxRequestBlobSidecars()
                        .intValue())
            .orElse(0);
```

zilm13: @mehdi-aouadi this way we will not change this number in the next fork if it's changed. The idea was to leave `spec.atEpoch()` thing but do a fallback not on 0, on Deneb number instead.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
zilm13: @Nashatyrev @mehdi-aouadi ok, I think we should discuss how to deal with constants like this with the team. @StefanBratanov don't like some current constants things also

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
mehdi-aouadi: Now I understand what you meant. Put the `spec.atEpoch(peer.finalizedEpoch())` back and used your suggestion to calculate the fallback value

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
mehdi-aouadi: I already created a separate issue for that #6927. It requires a design review and more changes

zilm13 (approved):
Looks great!

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
```diff
@@ -36,11 +38,18 @@ public SyncSourceFactory(final AsyncRunner asyncRunner, final TimeProvider timeP
...
-  public SyncSource getOrCreateSyncSource(final Eth2Peer peer) {
+  public SyncSource getOrCreateSyncSource(final Eth2Peer peer, final Spec spec) {
     // Limit request rate to just a little under what we'd accept
     final int maxBlocksPerMinute = MAX_BLOCKS_PER_MINUTE - SYNC_BATCH_SIZE.intValue() - 1;
+
+    final int maxBlobsPerBlock = spec.getMaxBlobsPerBlock().orElse(0);
+
     final int maxBlobSidecarsPerMinute =
-        MAX_BLOB_SIDECARS_PER_MINUTE - SYNC_BLOB_SIDECARS_SIZE.intValue() - 1;
+        MAX_REQUEST_BLOCKS_DENEB
```
StefanBratanov: this should be MAX_BLOCKS_PER_MINUTE i think

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
```diff
@@ -36,11 +38,18 @@ public SyncSourceFactory(final AsyncRunner asyncRunner, final TimeProvider timeP
...
+            .times(UInt64.valueOf(maxBlobsPerBlock))
+            .minusMinZero(SYNC_BLOB_SIDECARS_SIZE.intValue() - 1)
```
StefanBratanov: I think we should make this SYNC_BATCH_SIZE.times(maxBlobsPerBlock).intValue() - 1. I have removed SYNC_BLOB_SIDECARS_SIZE entirely in my PR.

StefanBratanov: Basically SYNC_BATCH_SIZE is what we use for syncing blocks and we will make request spanning same slots, but blobs sidecars are more than one per slot so can just multiply this value by the max blobs per block

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
mehdi-aouadi: done

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
mehdi-aouadi: Sounds right to me. Done

StefanBratanov (approved):
LGTM

PR #6910, opened 2023-03-06
Added an earliest available block slot application level cache

rolfyone:
By default, this will only reduce in-flight querries, but can also use `--Xnetwork-earliest-available-block-slot-cache-seconds` to only query the earliest block slot every N seconds.

## Documentation

## Changelog

### DIFF
--- a/ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
+++ b/ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
@@ -17,6 +17,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.Arrays.asList;
 import static tech.pegasys.teku.spec.config.SpecConfig.FAR_FUTURE_EPOCH;
+import static tech.pegasys.teku.spec.constants.NetworkConstants.DEFAULT_EARLIEST_AVAILABLE_BLOCK_SLOT_QUERY_FREQUENCY;
 import static tech.pegasys.teku.spec.constants.NetworkConstants.DEFAULT_SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY;
 import static tech.pegasys.teku.spec.networks.Eth2Network.GNOSIS;
 import static tech.pegasys.teku.spec.networks.Eth2Network.KILN;
@@ -75,6 +76,7 @@ public class Eth2NetworkConfiguration {
   private final Optional<UInt256> totalTerminalDifficultyOverride;
   private final Optional<UInt64> terminalBlockHashEpochOverride;
   private final Optional<Eth2Network> eth2Network;
+  private final int earliestAvailableBlockSlotFrequency;
 
   private Eth2NetworkConfiguration(
       final Spec spec,
@@ -96,6 +98,7 @@ public class Eth2NetworkConfiguration {
       final Optional<Bytes32> terminalBlockHashOverride,
       final Optional<UInt256> totalTerminalDifficultyOverride,
       final Optional<UInt64> terminalBlockHashEpochOverride,
+      int earliestAvailableBlockSlotFrequency,
       final Optional<Eth2Network> eth2Network) {
     this.spec = spec;
     this.constants = constants;
@@ -119,6 +122,7 @@ public class Eth2NetworkConfiguration {
     this.terminalBlockHashOverride = terminalBlockHashOverride;
     this.totalTerminalDifficultyOverride = totalTerminalDifficultyOverride;
     this.terminalBlockHashEpochOverride = terminalBlockHashEpochOverride;
+    this.earliestAvailableBlockSlotFrequency = earliestAvailableBlockSlotFrequency;
     this.eth2Network = eth2Network;
   }
 
@@ -210,6 +214,10 @@ public class Eth2NetworkConfiguration {
     return totalTerminalDifficultyOverride;
   }
 
+  public int getEarliestAvailableBlockSlotFrequency() {
+    return earliestAvailableBlockSlotFrequency;
+  }
+
   public Optional<UInt64> getTerminalBlockHashEpochOverride() {
     return terminalBlockHashEpochOverride;
   }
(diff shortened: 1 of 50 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/constants/NetworkConstants.java
```diff
@@ -18,4 +18,6 @@ public class NetworkConstants {
   public static final int SYNC_COMMITTEE_SUBNET_COUNT = 4;
   public static final int INTERVALS_PER_SLOT = 3;
   public static final int DEFAULT_SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY = 128;
+
```
tbenr: It is a bit weird that we have this in NetworkConstants.. Even if effects rpc, seems more related to Storage

storage/src/main/java/tech/pegasys/teku/storage/client/EarliestAvailableBlockSlot.java
```diff
@@ -0,0 +1,76 @@
...
+        .isLessThanOrEqualTo(timeProvider.getTimeInSeconds())) {
+      return actuallyGetEarliestBlockSlot();
+    }
+    return earliestAvailableBlockSlotFuture;
+  }
+
+  private synchronized SafeFuture<Optional<UInt64>> actuallyGetEarliestBlockSlot() {
+    if (queryTime
+        .plus(earliestAvailableBlockSlotFrequency)
+        .isLessThanOrEqualTo(timeProvider.getTimeInSeconds())) {
```
tbenr: since this is synchronized we can call timeProvider.getTimeInSeconds() only once before the if?

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
```diff
@@ -130,6 +131,16 @@ public class Eth2NetworkOptions {
       arity = "1")
   private UInt64 terminalBlockHashEpochOverride;
 
+  @Option(
```
tbenr: like before, just wonder is it is more storage related than network

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -665,5 +676,11 @@ public Builder applyGnosisNetworkDefaults() {
               "enr:-Ly4QKJ5BzgFyJ6BaTlGY0C8ROzl508U3GA6qxdG5Gn2hxdke6nQO187pYlLvhp82Dez4PQn436Fts1F0WAm-_5l2LACh2F0dG5ldHOIAAAAAAAAAACEZXRoMpCCS-QxAgAAZP__________gmlkgnY0gmlwhA-YLVKJc2VjcDI1NmsxoQI8_Lvr6p_TkcAu8KorKacfUEnoOon0tdO0qWhriPdBP4hzeW5jbmV0cwCDdGNwgiMog3VkcIIjKA",
               "enr:-Ly4QJMtoiX2bPnVbiQOJCLbtUlqdqZk7kCJQln_W1bp1vOHcxWowE-iMXkKC4_uOb0o73wAW71WYi80Dlsg-7a5wiICh2F0dG5ldHOIAAAAAAAAAACEZXRoMpCCS-QxAgAAZP__________gmlkgnY0gmlwhDbP3KmJc2VjcDI1NmsxoQNvcfKYUqcemLFlpKxl7JcQJwQ3L9unYL44gY2aEiRnI4hzeW5jbmV0cwCDdGNwgiMog3VkcIIjKA");
     }
+
+    public Builder earliestAvailableBlockSlotFrequency(
```
tbenr: I'd move this up next to the other getters :)

tbenr (approved):
LGTM with some comments.

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
rolfyone: it's a builder.. but sure heh.

tbenr: *getters* :facepalm:

storage/src/main/java/tech/pegasys/teku/storage/client/EarliestAvailableBlockSlot.java
```diff
@@ -0,0 +1,75 @@
...
+    this.earliestAvailableBlockSlotFrequency = earliestAvailableBlockSlotFrequency;
+  }
+
+  public SafeFuture<Optional<UInt64>> get() {
+    if (!earliestAvailableBlockSlotFuture.isDone()) {
+      // if the current query is not done, we should just return the same future
+      return earliestAvailableBlockSlotFuture;
+    }
+
+    if (queryTime
```
zilm13, line 51: I couldn't sleep with it. Why didn't you replace next 6 lines to just `actuallyGetEarliestBlockSlot();`?
You don't want fall into `synchronized` when it's not needed? But assuming most users will run it with default settings, they will fall in it and run this part of code twice

rolfyone: It's basically re-using the same cached query that's already running...

rolfyone: we dont even need to go into the synchronized section, because we have a future thats incomplete, so can just skip all the processing completely.

zilm13: > It's basically re-using the same cached query that's already running...

I understand it, the question was, why to calculate this twice, with default 0 value, we will count `queryTime
        .plus(earliestAvailableBlockSlotFrequency)
        .isLessThanOrEqualTo(timeProvider.getTimeInSeconds())` here and next in synchronized section the same again
```
    if (queryTime
        .plus(earliestAvailableBlockSlotFrequency)
        .isLessThanOrEqualTo(timeProvider.getTimeInSeconds())) {
      return actuallyGetEarliestBlockSlot();
    }
    return earliestAvailableBlockSlotFuture;
  }

  private synchronized SafeFuture<Optional<UInt64>> actuallyGetEarliestBlockSlot() {
    if (queryTime
        .plus(earliestAvailableBlockSlotFrequency)
        .isLessThanOrEqualTo(timeProvider.getTimeInSeconds())) {
      queryTime = timeProvider.getTimeInSeconds();
      LOG.trace("Setting earliestBlockSlot query time to {}", queryTime);
    } else {
      return earliestAvailableBlockSlotFuture;
    }
```

rolfyone: There's no guarantee it hasn't happened in the meantime... and it's a trivial computation...

PR #6917, opened 2023-03-08
Moved block pruner fetching outside the try with resources of an update

rolfyone:
This should mean the fetch query is no longer blocking an update.

Also extended frequency to 4-hourly, which is around 1200 blocks.

I had considered removing the need for the index scan by utilising the block column, but this may be worse, as it has to compute the block root for every block returned... In the first cleanup instance it would likely mean a large amount of computation.

## Documentation

## Changelog


lucassaldanha (commented):
The change looks ok to me. However, we deleted a test for the functionality. Do we have other tests to ensure we are not breaking the desired behaviour?

rolfyone:
> The change looks ok to me. However, we deleted a test for the functionality. Do we have other tests to ensure we are not breaking the desired behaviour?

We deleted the function... Now it's just a search for blocks and a delete blocks. I can check we cover those two elsewhere...

lucassaldanha (approved):
LGTM

PR #6922, opened 2023-03-08
Adds info log when pruning completes

tbenr:
With this we can see when the block pruning ends.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #6936, opened 2023-03-13
Avoid non-core flows to saturate `StorageQueryChannel`


tbenr:
As internally discussed we could better separate DB accesses between core-flows and non-core flows.
The idea is to guarantee DB query to core beaconchain functionalities, on the other hand limit DB accesses to non-core functionality (ie. Network RPC methods)

PR #6937, opened 2023-03-13
Implements `ThrottlingCombinedChainDataClient` part2

tbenr:
partially address: https://github.com/ConsenSys/teku/issues/6936

## Documentation

## Changelog


tbenr:
`RecentChainData` should be throttled as well since it uses the query channel too... and it isn't straight forward..

So this currently throttle only access to finalized data

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -357,6 +360,9 @@ protected SafeFuture<?> initialize() {
         eventChannels.getPublisher(CombinedStorageChannel.class, beaconAsyncRunner);
     storageQueryChannel = combinedStorageChannel;
     storageUpdateChannel = combinedStorageChannel;
+    throttlingStorageQueryChannel =
+        new ThrottlingStorageQueryChannel(storageQueryChannel, 4, metricsSystem);
```
rolfyone: so we'd go from 10 down to 4?

I wonder if we want 10 in the throttled instance, and headroom like 3-5 outside of it...

tbenr:
demoting to Draft since there were other ideas to address this

Issue #6242, opened 2022-09-22
Default behavior should not send packets to RFC1918 addresses


dtto:
### Description
Normal operation should not send traffic to rfc1918 addresses.

### Steps to Reproduce (Bug)
Run teku. Watch for traffic on rfc1918 addresses, e.g. 

```
# while ! ss -p state syn-sent | grep 192.168; do sleep .1; done
tcp   0      1      [::ffff:XXXX]:51566 [::ffff:192.168.0.158]:13000 users:(("java",pid=16341,fd=429))
```

**Expected behavior:** [What you expect to happen]
No traffic is sent to rfc1918.
I suggest to add a flag to enable sending traffic to rfc1918 as needed.

**Actual behavior:** [What actually happens]
On occasion, traffic (likely p2p discovery?) is sent to RFC1918 addresses. 

**Frequency:** [How regularly does it occur?]
Varies.

### Versions (Add all that apply)
* Software version: [`teku --version` or look for log lines starting with `Teku version:`]
```
$ ./teku --version
teku/v22.9.1+3-g3d7dd4f/linux-x86_64/-debian-openjdk64bitservervm-java-17
```
* Java version: [`java -version`]
```
$ java --version
openjdk 17.0.4 2022-07-19
OpenJDK Runtime Environment (build 17.0.4+8-Debian-1deb11u1)
OpenJDK 64-Bit Server VM (build 17.0.4+8-Debian-1deb11u1, mixed mode, sharing)
```
* OS Name & Version: [`cat /etc/*release`]

Debian 11

* Docker Version: [`docker version`]

N/A

* Cloud VM, type, size: [Amazon Web Services I3-large]
N/A

taggart:
I am seeing this too. It would be good if these were off by default.

ajsutton:
For the record, these ranges to avoid map to `InetAddress.isSiteLocalAddress` in Java. https://www.rfc-editor.org/rfc/rfc1918 

```
   The Internet Assigned Numbers Authority (IANA) has reserved the
   following three blocks of the IP address space for private internets:

     10.0.0.0        -   10.255.255.255  (10/8 prefix)
     172.16.0.0      -   172.31.255.255  (172.16/12 prefix)
     192.168.0.0     -   192.168.255.255 (192.168/16 prefix)
```

ajsutton:
Finishing this off will require:
* getting #6631 merged
* deploying and monitoring a box with this enabled to check there aren't other places that need filtering
[...]

PR #6955, opened 2023-03-16
Unhide `--p2p-discovery-site-local-addresses-enabled`

zilm13:
## PR Description
It's follow-up for the #6242 
How do I see it:
- this one PR: we unhide the feature but left it toggled off to give some time to update it
- before the next release we toggle it to the `false` by default and add appropriate CHANGELOG entry 

```
Syntax: --p2p-discovery-site-local-addresses-enabled[=<BOOLEAN>]
Example: --p2p-discovery-site-local-addresses-enabled=true
Environment variable: P2P_DISCOVERY_SITE_LOCAL_ADDRESSES_ENABLED=true
Configuration file: p2p-discovery-site-local-addresses-enabled: true
```

Whether discovery accepts messages and peer records with site local (RFC1918) addresses. Default is `false`.

Normal Teku operation should not send traffic to local networks (RFC1918) addresses.
```
   The Internet Assigned Numbers Authority (IANA) has reserved the
   following three blocks of the IP address space for private internets:

 	10.0.0.0    	-   10.255.255.255  (10/8 prefix)
 	172.16.0.0  	-   172.31.255.255  (172.16/12 prefix)
 	192.168.0.0 	-   192.168.255.255 (192.168/16 prefix)
```	 
For some reasons operators may need to enable discovery of local addresses, for example when you run multiple consensus layer nodes in one local network, these nodes are not discovered in public Internet and are advertised with local IPs. Such configurations are usual for test or private networks. When running in such configuration this flag should be toggled (set to true).

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ For information on changes in released versions of Teku, see the [releases page]
 - Send validator registrations to the Beacon node when the Validator client has reconnected to the event stream
 - Added optional query parameters `require_prepared_proposers` and `require_validator_registrations` to the `teku/v1/admin/readiness` endpoint
 - Added experimental feature `--Xdeposit-snapshot-enabled` to use bundled deposit contract tree snapshot and persist it after finalization to decrease EL pressure and speed up node startup
-- Added `--p2p-discovery-site-local-addresses-enabled` option to allow discovery connections to local network addresses 
+- Added `--p2p-discovery-site-local-addresses-enabled` option to allow discovery connections to local (RFC1918) addresses.
 
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -19,5 +20,7 @@ For information on changes in released versions of Teku, see the [releases page]
 - Send validator registrations to the Beacon node when the Validator client has reconnected to the event stream
 - Added optional query parameters `require_prepared_proposers` and `require_validator_registrations` to the `teku/v1/admin/readiness` endpoint
 - Added experimental feature `--Xdeposit-snapshot-enabled` to use bundled deposit contract tree snapshot and persist it after finalization to decrease EL pressure and speed up node startup
+- Added `--p2p-discovery-site-local-addresses-enabled` option to allow discovery connections to local network addresses
```
StefanBratanov: I would change `local network` to `local (RFC1918)`.

StefanBratanov (approved):
LGTM, apart from a nit

PR #6962, opened 2023-03-20
Updated jackson libraries to switch to using dataformats-text

rolfyone:
dataformats-yaml had some issues, and I also noticed that dataformats-text supports toml.

This should replace both libs with the current version.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #6965, opened 2023-03-20
add decoupled blob sidecars signing

mehdi-aouadi:
## PR Description
Add decoupled blob sidecars signing

## Fixed Issue(s)

#6959 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/LocalSigner.java
```diff
@@ -122,6 +123,11 @@ public SafeFuture<BLSSignature> signValidatorRegistration(
     return sign(signingRootUtil.signingRootForValidatorRegistration(validatorRegistration));
   }
 
+  @Override
+  public SafeFuture<BLSSignature> signBlobSidecar(BlobSidecar blobSidecar, ForkInfo forkInfo) {
```
StefanBratanov: final 🗡️

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/generator/signatures/NoOpSigner.java
```diff
@@ -46,6 +47,11 @@ public SafeFuture<BLSSignature> signBlock(final BeaconBlock block, final ForkInf
     return new SafeFuture<>();
   }
 
+  @Override
+  public SafeFuture<BLSSignature> signBlobSidecar(BlobSidecar blobSidecar, ForkInfo forkInfo) {
```
StefanBratanov: final  🗡️

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
```diff
@@ -112,6 +116,30 @@ public SafeFuture<SignedBeaconBlock> signBlock(
         .thenApply(signature -> SignedBeaconBlock.create(spec, unsignedBlock, signature));
   }
 
+  @SuppressWarnings("unused")
+  public SafeFuture<SignedBlobSidecar> signBlobSidecar(
```
StefanBratanov, line 120: hmm actually not sure this method belongs here. We will probably need something like `BlobSidecarProductionDuty` similar to `AttestationProductionDuty`, but can do it in a PR which will deal with the validator flow.

StefanBratanov (changes requested):
LGTM, just couple of final nits and one method is not needed for now

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
tbenr: humm I'm actually ok with that. The duty is the block production, which now requires signing and posting several objects but I see it as the "extended" `ProductionDuty`.  `createUnsignedBlock` will give us `BlockContent` object wrapping a (blinded)block and (blinded)sidecars.
Happy to discuss it if you have arguments against it.

StefanBratanov: yeah I see now from the specs, It's one flow true. I am ok with the method there now. We can later change to something like BlockAndSidecarProductionDuty which extends the `BlockProductionDuty`, but in any case that method won't change, so it's good to have it.

mehdi-aouadi: We could add `BlockAndSidecarProductionDuty` which extends the `BlockProductionDuty` and leave it to the `BlockDutyFactory` to determine which one to create based on the slot...

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/generator/signatures/NoOpSigner.java
mehdi-aouadi: Done

ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/LocalSigner.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
StefanBratanov: Sounds good! Think it would be cleaner to tackle it in the PR which will work on the block  production

StefanBratanov (approved):

StefanBratanov:
LGTM

Issue #6679, opened 2023-01-12
Log start of block production


StefanBratanov:
Inspired by the discussion in https://github.com/ConsenSys/teku/issues/6658

Could be implemented in `ValidatorApiHandler` where currently there is a trace log: 
```
LOG.trace("Creating unsigned block for slot {}", slot);
```

It would be better to use the `EventLogger` to make the formatting nicer and to make sure it is included in the console log. Can test by starting Interop BN with multiple validators.

PR #6970, opened 2023-03-21
Added block production logging

rolfyone:
Partially addresses #6679

## Documentation

## Changelog


tbenr:
This is a bit off :) let's have a conversation tomorrow so we can quickly fix it together

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceUpdateData.java
```diff
@@ -196,6 +190,44 @@ public SafeFuture<Optional<ForkChoiceUpdatedResult>> send(
...
+
+  private void logSendingForkChoiceUpdated() {
+    if (LOG.isDebugEnabled()) {
+      LOG.debug(
+          "send - calling forkChoiceUpdated({}, {})", forkChoiceState, payloadBuildingAttributes);
+    } else {
+      if (payloadBuildingAttributes.isEmpty()) {
+        LOG.info(
+            "Calling local execution layer to start block production (parent slot: {})",
+            forkChoiceState.getHeadBlockSlot());
```
rolfyone: So this is why i said parent slot, because if we're building on the headBlockSlot, then that would be the parent of what we're building, so i guess this is just overall badly worded...

tbenr (approved):
LGTM

PR #6971, opened 2023-03-21
Fix local execution payload value challenge won log

zilm13:
## PR Description

Currently logs local value on a place of bid value and vice versa. Plus readability improved

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
@@ -81,7 +81,7 @@ public class ExecutionBuilderModule {
     if (builderClient.isPresent() && builderBidChallengePercentage.isPresent()) {
       LOG.info(
           "During block production, the builder bid will be considered at {}% of its value when compared to locally produced payload."
-              + "Configure with --Xvalidators-builder-bid-challenge-percentage",
+              + " Configure with --Xvalidators-builder-bid-challenge-percentage",
           builderBidChallengePercentage.get());
     }
   }
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
@@ -78,6 +78,12 @@ public class ExecutionBuilderModule {
     this.builderClient = builderClient;
     this.eventLogger = eventLogger;
     this.builderBidChallengePercentage = builderBidChallengePercentage;
+    if (builderClient.isPresent() && builderBidChallengePercentage.isPresent()) {
+      LOG.info(
+          "During block production, the builder bid will be considered at {}% of its value when compared to locally produced payload."
+              + "Configure with --validators-builder-bid-compare-factor",
+          builderBidChallengePercentage.get());
+    }
   }
 
   private Optional<SafeFuture<HeaderWithFallbackData>> validateBuilderGetHeader(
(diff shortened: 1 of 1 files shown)


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -170,12 +171,7 @@ public SafeFuture<HeaderWithFallbackData> builderGetHeader(
                 logReceivedBuilderBid(signedBuilderBid.getMessage());
 
                 if (isLocalPayloadValueWinning(signedBuilderBid, localPayloadValue)) {
```
tbenr, line 179: why dont we extract the builderBidValue once here and pass it around instead of the entire bid object?

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -412,4 +408,14 @@ private void logReceivedBuilderBid(final BuilderBid builderBid) {
         payloadHeader.getGasLimit(),
         payloadHeader.getGasUsed());
   }
+
+  private void logLocalPayloadWin(
+      final SignedBuilderBid signedBuilderBid, final UInt256 localPayloadValue) {
+    LOG.info(
+        "The local execution payload value ({} Gwei) was awarded the block over the builder payload ({} Gwei), "
```
tbenr: Since we are changing it, the sentence doesn't sound very good.
something like "the local execution payload value () has been chosen over the the builder bid value ()."

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -412,4 +408,14 @@ private void logReceivedBuilderBid(final BuilderBid builderBid) {
...
+            + "{}% challenge configured.",
```
tbenr: I'd print this out only if !=100%. And we could rephrase by saing "Only {}% of the builder bid value has been considered". Or something like that.

tbenr:
moreover do we want to allow >100% on the challenge config? we have `NEVER` but does make any sense to configure something like 150%?
I'm worried about user errors, that may interpret the configuration in a wrong way. We could prevent those values or at lest print a warning saying "is this really what you intended to do?"

zilm13:
@tbenr 
I have this draft for docs, I think it's clear or could be improved to be clear. I think users are smart and they really may need >100%
```
### validators-builder-bid-compare-factor

<!--tabs-->

# Syntax

``bash
--validators-builder-bid-compare-factor=<INTEGER>
``

# Example

``bash
--vvalidators-builder-bid-compare-factor=50
``

# Environment variable

``bash
TEKU_VALIDATORS_BUILDER_BID_COMPARE_FACTOR=50
``

# Configuration file

``bash
validators-builder-bid-compare-factor: 50
``

<!--/tabs-->

Changes validator's builder bid compare factor. The default value is 100 (100%).

Since Capella fork local execution payload contains its value, the validator client could use it over builder bid when it's higher or equal. Operator could override the factor, in this case 50 means builder bid is considered at 50% value, so it should be at least twice bigger to win over local payload and used for block production, contra 200 would mean use builder bid over local payload even if it's smaller up to 2 times.
```

tbenr:
@zilm13 
```
``bash
--vvalidators-builder-bid-compare-factor=50
``
```

`--validators-builder-bid-compare-factor=50`

zilm13:
thanks, and we need a section with NEVER also

tbenr:
@zilm13 what about closing this in favour of #6974

PR #6974, opened 2023-03-21
Unhide builder bid compare factor and fix win logging

zilm13:
## PR Description
This PR is based on #6971, which should be merged before this one
- Option is unhided, plus name was changed and code all the way updated to refer to the new name
- Changelog entry added

Documentation PR will be made after this one merged

UPDATE: as #6971 is closed with all changes applied here, copying its contents, it's fixed here too:
- Fix local execution payload value challenge won log. Currently logs local value on a place of bid value and vice versa. Plus readability improved
## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ For information on changes in released versions of Teku, see the [releases page]
 - Added mainnet configuration for CAPELLA network fork due epoch 194048, April 12, 2023; 10:27:35pm UTC
 - Added experimental feature `--Xdeposit-snapshot-enabled` to use bundled deposit contract tree snapshot and persist it after finalization to decrease EL pressure and speed up node startup
 - Added `--p2p-discovery-site-local-addresses-enabled` option to allow discovery connections to local (RFC1918) addresses.
-- Added `--validators-builder-bid-compare-factor` to override factor of fallback to local payload when its value beats the builder bid (starting on Shapella).
+- Added `--builder-bid-compare-factor` to be applied to the builder bid value when comparing it with locally produced payload. (starting on Shapella).
 
 
 ### Bug Fixes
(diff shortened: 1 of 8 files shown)


tbenr:
moving it here. Likely needs to be updated

```
### builder-bid-compare-factor

<!--tabs-->

# Syntax

``bash
--builder-bid-compare-factor=<INTEGER>
``

# Example

``bash
--builder-bid-compare-factor=50
``

# Environment variable

``bash
TEKU_BUILDER_BID_COMPARE_FACTOR=50
``

# Configuration file

``bash
builder-bid-compare-factor: 50
``

<!--/tabs-->

Changes validator's builder bid compare factor. The default value is 100 (100%).

Since Capella fork local execution payload contains its value, the validator client could use it over builder bid when it's higher or equal. Operator could override the factor, in this case 50 means builder bid is considered at 50% value, so it should be at least twice bigger to win over local payload and used for block production, contra 200 would mean use builder bid over local payload even if it's smaller up to 2 times.
```

zilm13:
Was working on this. Check it

```

### builder-bid-compare-factor

<!--tabs-->

# Syntax

``bash
--builder-bid-compare-factor=<STRING>
``

# Example

``bash
--builder-bid-compare-factor=50
``

# Environment variable

``bash
TEKU_BUILDER_BID_COMPARE_FACTOR=50
``

# Configuration file

``bash
builder-bid-compare-factor: 50
``

<!--/tabs-->

Changes validator's builder bid compare factor. The default value is 100 (100%).

Starting from Shapella fork local execution payload contains its value, so beacon node could compare it against the builder bid in order to maximize validators profit or improve network censorship at low or no cost. 

Set the compare factor applied to the builder bid value when comparing it with locally produced payload. Factor is expressed in percentage (e.g. `100` means locally produced payload will be chosen when its value is equal or greater than the entire builder bid value, `80` means local payload will be chosen when its value is at least 80% of builder bid value).

Set it to `BUILDER_ALWAYS` to always use builder bid. Locally produced payload will be used only when the bid is invalid.
```

tbenr (approved):
LGTM

benjaminion:
> Starting from Shapella fork

Ugh - can we never ever say "Shapella", please 🤮

"Shanghai–Capella" (yes, that's an [en dash](https://www.grammarly.com/blog/en-dash/) between them).

PR #6982, opened 2023-03-22
Multipeer forward sync for decoupled blobs

### LINKED ISSUE #6948: Implement Forward Sync for Decoupled Blobs
Will require changes to both `SinglePeerSyncService` and `MultipeerSyncService`

StefanBratanov:
## PR Description

- Main changes are in `SyncSourceBatch` and `BatchImporter`.
- Essentially when creating the batch, simultaneously block and blob sidecars are downloaded and then stored together. Then for every block, its blob sidecars are imported and after that the block is imported.

## Fixed Issue(s)
fixes #6948 

## Documentation

## Changelog

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
@@ -69,14 +69,14 @@ public class BatchImporter {
         () -> {
           final SignedBeaconBlock firstBlock = blocks.get(0);
           SafeFuture<BlockImportResult> importResult =
-              importBlobSidecarsAndBlock(firstBlock, blobSidecars, source.orElseThrow());
+              importBlockAndBlobSidecars(firstBlock, blobSidecars, source.orElseThrow());
           for (int i = 1; i < blocks.size(); i++) {
             final SignedBeaconBlock block = blocks.get(i);
             importResult =
                 importResult.thenCompose(
                     previousResult -> {
                       if (previousResult.isSuccessful()) {
-                        return importBlobSidecarsAndBlock(
+                        return importBlockAndBlobSidecars(
                             block, blobSidecars, source.orElseThrow());
                       } else {
                         return SafeFuture.completedFuture(previousResult);
@@ -100,7 +100,7 @@ public class BatchImporter {
         });
   }
 
-  private SafeFuture<BlockImportResult> importBlobSidecarsAndBlock(
+  private SafeFuture<BlockImportResult> importBlockAndBlobSidecars(
       final SignedBeaconBlock block,
       final Map<Bytes32, List<BlobSidecar>> blobSidecars,
       final SyncSource source) {
@@ -114,11 +114,6 @@ public class BatchImporter {
         blobSidecarsForBlock.size(),
         blockRoot);
     return importBlobSidecars(blobSidecarsForBlock).thenCompose(__ -> importBlock(block, source));
-  }
-
-  private SafeFuture<Void> importBlobSidecars(final List<BlobSidecar> blobSidecars) {
-    return SafeFuture.allOfFailFast(
-        blobSidecars.stream().map(blobsSidecarManager::importBlobSidecar));
   }
 
   private SafeFuture<BlockImportResult> importBlock(
@@ -139,6 +134,11 @@ public class BatchImporter {
               }
               return result;
             });
+  }
+
+  private SafeFuture<Void> importBlobSidecars(final List<BlobSidecar> blobSidecars) {
+    return SafeFuture.allOfFailFast(
+        blobSidecars.stream().map(blobsSidecarManager::importBlobSidecar));
   }
 
   public enum BatchImportResult {
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
@@ -108,9 +108,12 @@ public class BatchImporter {
     if (!blobSidecars.containsKey(blockRoot)) {
       return importBlock(block, source);
     }
-    LOG.debug("Importing {} blob sidecars for block with root {}", blobSidecars, blockRoot);
-    return importBlobSidecars(blobSidecars.get(blockRoot))
-        .thenCompose(__ -> importBlock(block, source));
+    final List<BlobSidecar> blobSidecarsForBlock = blobSidecars.get(blockRoot);
+    LOG.debug(
+        "Importing {} blob sidecars for block with root {}",
+        blobSidecarsForBlock.size(),
+        blockRoot);
+    return importBlobSidecars(blobSidecarsForBlock).thenCompose(__ -> importBlock(block, source));
   }
 
   private SafeFuture<Void> importBlobSidecars(final List<BlobSidecar> blobSidecars) {
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
@@ -17,31 +17,40 @@ import static com.google.common.base.Preconditions.checkState;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes32;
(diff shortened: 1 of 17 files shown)


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
```diff
@@ -299,6 +348,58 @@ public String toString() {
...
+  }
+
+  private boolean validateNewBlobSidecars(
+      final List<SignedBeaconBlock> newBlocks,
+      final Map<Bytes32, List<BlobSidecar>> newBlobSidecarsByBlockRoot) {
+    for (final SignedBeaconBlock block : newBlocks) {
+      final Bytes32 blockRoot = block.getRoot();
+      final List<BlobSidecar> blobSidecars =
+          newBlobSidecarsByBlockRoot.getOrDefault(blockRoot, List.of());
+      final int numberOfKzgCommitments =
```
mehdi-aouadi, line 382: You can reuse the `getBlobSidecarsCount` from `MiscHelpersDeneb`

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
```diff
@@ -88,6 +100,27 @@ public SafeFuture<BatchImportResult> importBatch(final Batch batch) {
         });
   }
 
+  private SafeFuture<BlockImportResult> importBlobSidecarsAndBlock(
```
mehdi-aouadi: I'd rather rename it to `importBlockAndBlobSidecars`

mehdi-aouadi (approved):
LGTM.
Some small changes suggestions though...

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
StefanBratanov: yeah I would have done if the `spec` object was passed to this class, but unfortunately it isn't

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
```diff
@@ -299,6 +348,58 @@ public String toString() {
...
+          "Marking batch invalid because new blocks do not form a chain with previous blocks");
+      return false;
+    }
+    return true;
+  }
+
+  private boolean validateNewBlobSidecars(
+      final List<SignedBeaconBlock> newBlocks,
+      final Map<Bytes32, List<BlobSidecar>> newBlobSidecarsByBlockRoot) {
+    for (final SignedBeaconBlock block : newBlocks) {
```
tbenr, line 378: I have doubts here. Seems like you are validating blobs from the blocks we have.

what if:
blockByRange(10,12) returns:
`[blockAt10,BlockAt12]`

BlobsByRange(10,12) returns:
`[BlobsAt10[],BlobsAt11[],BlobsAt12[]]`

and BlobsAt11 contains a sidecar with slot set to `100000`

that sidecar isn't validated and end up being imported and never pruned (slot `100000`)

Maybe we just need to check that `newBlobSidecarsByBlockRoot.size() == newBlocks.size()`
and we need to make sure we have newBlobSidecarsByBlockRoot populated also for blocks with no sidecars. Or, you should track the sidecars you visited and make sure you haven't left any out.

tbenr (changes requested):
blocking before we fully agree on validation

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
tbenr: Wait, you then import sidecars by block so they wont be considered. Sorry :-)
there is still a  weird situation in which the peer sent you random blobs uncorrelated with blocks and we don't care. We could make an argument that we should downscore it, but it is non critical...

StefanBratanov: Good point, It makes sense. I added a commit with this change and a test. Debug log looks like this:
```
2023-03-23 18:53:03.458+00:00 | ForkJoinPool-1-worker-1 | DEBUG | SyncSourceBatch | Marking batch invalid because unexpected blob sidecars with roots [0x7e3c83fd2c8dab66478a606a6f1a996132f655886c22be380f7b9eac6631682d] were received
```

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
```diff
@@ -299,6 +352,71 @@ public String toString() {
...
+        }
+      }
+    }
+    final SetView<Bytes32> unexpectedBlobSidecarsRoots =
+        Sets.difference(newBlobSidecarsByBlockRoot.keySet(), blockRootsWithKzgCommitments);
+    if (!unexpectedBlobSidecarsRoots.isEmpty()) {
+      LOG.debug(
+          "Marking batch invalid because unexpected blob sidecars with roots {} were received",
+          unexpectedBlobSidecarsRoots);
+      return false;
```
mehdi-aouadi: Should we mark the whole batch as invalid when we receive additional unexpected blob sidecars? If we received all the block's blob sidecars we could simply log a warning, ignore them and maybe downrate the peer

mehdi-aouadi (changes requested):

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
StefanBratanov: yeah I agree, marking batch as invalid is bit harsh, when we have already received what we need, added a commit for that

mehdi-aouadi (approved):
LGTM

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
```diff
@@ -299,6 +355,76 @@ public String toString() {
...
+              "Marking batch invalid because blob sidecar for root {} was received with slot {} which is different than the block slot {}",
+              blockRoot,
+              blobSidecar.getSlot(),
+              blockSlot);
+          return false;
+        }
+      }
+    }
+
+    final SetView<Bytes32> unexpectedBlobSidecarsRoots =
```
tbenr: unexpectedBlobSidecarsBlockRoots :-)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
```diff
@@ -299,6 +355,76 @@ public String toString() {
...
+        Sets.difference(newBlobSidecarsByBlockRoot.keySet(), blockRootsWithKzgCommitments);
+    if (!unexpectedBlobSidecarsRoots.isEmpty()) {
+      LOG.debug(
+          "Applying minor penalty to peer {} because unexpected blob sidecars with roots {} were received",
+          syncSource,
+          unexpectedBlobSidecarsRoots);
+      syncSource.adjustReputation(ReputationAdjustment.SMALL_PENALTY);
```
tbenr: to follow the design, shouldn't we have it inside the conflictResolutionStrategy?

tbenr (approved):
LGTM, just a couple of things

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/batches/SyncSourceBatch.java
StefanBratanov: Yeah, but reportInvalidBatch results in a LARGE_PENALTY for the peer, is that what we would like in this scenario?

tbenr: I see like another method exposed there to cover this specific situation..

StefanBratanov: Kk, did a change, not 100% sure about naming, but think it makes sense.

tbenr (approved):
LGTM

PR #6990, opened 2023-03-24
Avoid blob array data copy

tbenr:
`toArrayUnsafe` avoids blobs array to be copied into a new array. We trust kzg lib.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
+++ b/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
@@ -166,7 +166,7 @@ public final class CKZG4844 implements KZG {
       throws KZGException {
     try {
       final byte[] proof =
-          CKZG4844JNI.computeBlobKzgProof(blob.toArrayUnsafe(), kzgCommitment.toArray());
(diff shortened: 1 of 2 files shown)


infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
```diff
@@ -165,7 +165,8 @@ public KZGCommitment blobToKzgCommitment(final Bytes blob) throws KZGException {
   public KZGProof computeBlobKzgProof(final Bytes blob, final KZGCommitment kzgCommitment)
       throws KZGException {
     try {
-      final byte[] proof = CKZG4844JNI.computeBlobKzgProof(blob.toArray(), kzgCommitment.toArray());
+      final byte[] proof =
+          CKZG4844JNI.computeBlobKzgProof(blob.toArrayUnsafe(), kzgCommitment.toArray());
```
StefanBratanov: We can also do the same for the commitment I guess

StefanBratanov (approved):
Agree with it. Binding makes sure that the passed arrays are not modified.

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844.java
tbenr: ah yes, it is a layer below..

PR #6993, opened 2023-03-27
use progressive balance and don't check anymore

tbenr:
just a spike
should improve epoch transition by fully using progressive balance and not recalculating balance from scratch to verify progressive balance

## Documentation

## Changelog


tbenr:
we can work on this when we peek the relative issue. closing for now.

PR #6996, opened 2023-03-28
Percentage sign escape in `--builder-bid-compare-factor` description

tbenr:
fixes:
```
[picocli WARN] Could not format 'Set the compare factor applied to the builder bid value when comparing it with locally produced payload. Factor is expressed in percentage (e.g. '100' means locally produced payload will be chosen when its value is equal or greater than the entire builder bid value, '80' means local payload will be chosen when its value is at least 80% of builder bid value).
Set it to 'BUILDER_ALWAYS' to always use builder bid. In this configuration locally produced payload will be used only when the bid is invalid.' (Underlying error: Format specifier '% o'). Using raw String: '%n' format strings have not been replaced with newlines. Please ensure to escape '%' characters with another '%'.
```

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #6803, opened 2023-02-10
Top-up blocks voluntary exit?


heeckhau:
### Description and steps to Reproduce (Bug)
I have set up a few validators on zhejiang and the voluntary exit does not work for validators where I did a "top up"

Note that the voluntary exit did work for a regular zhejiang validator I also started: https://zhejiang.beaconcha.in/validator/0x87093e5995aa89184ce9fd151ebc1a787804c54ecd62d22ec8879af7c54c02e95af1550351633a127d18146c203e8239

For the other validator, I did some extra topups with https://zhejiang.launchpad.ethereum.org/en/top-up
This validator I can not exit. Whatever I try, the exit epoch remains on 18446744073709551615
https://zhejiang.beaconcha.in/validator/0xa743b1bcc41fa493a44a3c362a1f9a8699ffb35a2d6333bc04427b559cf44b2649d9df79bcec49e746f9345085999a07#deposits
```
Loading configuration...
Exits are going to be generated for validators:
98e2bf9
Epoch: 1381

These validators won't be able to be re-activated once this operation is complete.
Are you sure you wish to continue (yes/no)?
yes
Exit for validator 98e2bf9 submitted.
```

I tried the same with another validator that was topped-up. Same result.

<img width="979" alt="image" src="https://user-images.githubusercontent.com/735288/218102739-d96a5c9b-838a-4c8f-9f57-3a196e8427ff.png">

⚠️ **Update** ⚠️ : when I tried again a day later, it just worked. Does a top-up reset a timer?

https://zhejiang.beaconcha.in/validator/0x98e2bf94fe3db418b3d7aa47930a672e1ad35d971928c198d997b31276d654867fd91f1aa66a71ce0ee3ff2d39175a6a
<img width="905" alt="image" src="https://user-images.githubusercontent.com/735288/218102643-66074035-043d-4b9d-be27-596f418ea2f0.png">

-------

**Expected behavior:**

A voluntary exit should work or an error message should be shown.

**Actual behavior:** 

The exit command is accepted, but the exit epoch is not changed

### Versions (Add all that apply)
* Software version: teku/v23.1.0+10-gac10368/osx-aarch_64/-eclipseadoptium-openjdk64bitservervm-java-17
* Java version: OpenJDK 64-Bit Server VM Temurin-17.0.2+8 (build 17.0.2+8, mixed mode)
* OS Name & Version: macOS Ventura 13.1

benjaminion:
[...]

PR #6999, opened 2023-03-28
Refactor BLS Operation pool for re-use.

rolfyone:
Generify the class so that it can be employed by voluntary exit operation pool.

The advantage of this class is that it allows differentiation between locally and remotely submitted operations, which allows us in turn to select preferentially, and potentially in future perform other operations on local objects.

This does substantially increase the potential pool size for voluntary exits, but we can reduce that if we need to.

Partially addresses #6803

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/MessageWithValidatorId.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.spec.datastructures.operations;
+
+import tech.pegasys.teku.infrastructure.ssz.SszData;
+
+public interface MessageWithValidatorId extends SszData {
```
lucassaldanha, line 18: Should we call it `OperationWithValidatorId`?

lucassaldanha (approved):
LGTM. Just a nit naming suggestion.

PR #7004, opened 2023-03-30
Fix blobsidecar gossip validation

tbenr:
I spotted a bug in the gossip validation rule for "first seen for tuple index-blockroot"

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarValidator.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarValidator.java
@@ -18,6 +18,7 @@ import static tech.pegasys.teku.spec.config.Constants.VALID_BLOCK_SET_SIZE;
 import static tech.pegasys.teku.statetransition.validation.InternalValidationResult.reject;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Objects;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarValidator.java
```diff
@@ -201,13 +201,48 @@ private boolean isSignatureValidWithRespectToProposerIndex(
...
   }
 
-  private boolean blobSidecarIsFirstWithValidSignatureForSlot(BlobSidecar blobSidecar) {
+  private boolean isFirstWithValidSignatureForIndexAndBlockRoot(BlobSidecar blobSidecar) {
     return !receivedValidBlobSidecarInfoSet.contains(
-        new SlotAndBlockRoot(blobSidecar.getSlot(), blobSidecar.getBlockRoot()));
+        new IndexAndBlockRoot(blobSidecar.getIndex(), blobSidecar.getBlockRoot()));
+  }
+
+  static class IndexAndBlockRoot {
```
StefanBratanov, line 214: I think would be useful to move this object to spec.blobs similar to `SlotAndBlockRoot`, so can reuse it for the future.

StefanBratanov (approved):
LGTM with one comment

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarValidator.java
StefanBratanov: Actually we can also just use `BlobIdentifier`? (ignore since it is ssz object)

tbenr: If i have to move it next to `SlotAndBlockRoot` it'd be identical to it..I'd be tempted to name it `UInt64AndBytes32` and having the meaningful information on the var name...
edit:
having the equality being false between the two it's a good safety feature though.

StefanBratanov: let's leave it as it is in this PR, can move it if we need it somewhere else at some point

PR #7007, opened 2023-04-03
Added option for adding non-critical methods on Web3JClient

lucassaldanha:
## PR Description

Added option for adding non-critical methods on Web3JClient.
Non-critical methods won't trigger availability updates.

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
+++ b/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
@@ -42,6 +42,9 @@ public abstract class Web3JClient {
   private final ExecutionClientEventsChannel executionClientEventsPublisher;
   private Web3jService web3jService;
   private Web3j eth1Web3j;
+  /*
+   Non-critical methods won't trigger availability updates when they succeed/fail
+  */
   private final Collection<String> nonCriticalMethods = new HashSet<>();
 
   // Default to the provider having a previous failure at startup so we log when it is first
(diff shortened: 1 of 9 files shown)


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
```diff
@@ -77,39 +82,50 @@ public <T> SafeFuture<Response<T>> doRequest(
             (response, exception) -> {
               if (exception != null) {
                 final boolean couldBeAuthError = isAuthenticationException(exception);
-                handleError(exception, couldBeAuthError);
+                handleError(isCriticalRequest(web3jRequest), exception, couldBeAuthError);
```
StefanBratanov: I would initialise beforehand `boolean isCritical = isCriticalRequest(web3jRequest)` since it is used in all calls

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
```diff
@@ -49,10 +52,12 @@ public abstract class Web3JClient {
   protected Web3JClient(
       final EventLogger eventLog,
       final TimeProvider timeProvider,
-      final ExecutionClientEventsChannel executionClientEventsPublisher) {
+      final ExecutionClientEventsChannel executionClientEventsPublisher,
+      final Collection<String> nonCriticalMethods) {
     this.eventLog = eventLog;
     this.timeProvider = timeProvider;
     this.executionClientEventsPublisher = executionClientEventsPublisher;
+    this.nonCriticalMethods.addAll(nonCriticalMethods);
```
StefanBratanov, line 63: Why not just `this.nonCriticalMethods = nonCriticalMethods` to be consistent

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3jClientBuilder.java
```diff
@@ -31,8 +33,8 @@ public class Web3jClientBuilder {
   private URI endpoint;
   private Optional<JwtConfig> jwtConfigOpt = Optional.empty();
   private Duration timeout;
-
   private ExecutionClientEventsChannel executionClientEventsPublisher;
+  private Collection<String> nonCriticalMethods = new HashSet<>();
```
StefanBratanov: Think we can add already engine_exchangeTransitionConfigurationV1 and engine_exchangeCapabilities. Wdyt?

StefanBratanov (approved):
logic LGTM! Added few comments

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
lucassaldanha: Just to ensure it can't be modified externally. So instead of accepting a reference from the outside, I add the values into my own internal collection. Not really critical here but why not. :)

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3jClientBuilder.java
lucassaldanha: I will add them where we use the constructor instead of as default values. But thanks for the reminder! :)

PR #7008, opened 2023-04-03
Add unit test on wrong bid milestone

tbenr:
Added a test to ensure we fail the response parsing when builder sends wrong milestone bid (so we will fallback to local)

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7009, opened 2023-04-03
Add builder API response version checking

tbenr:
Makes version field actually parsed and verified when calling builder APIs.

It is actually important in future milestone where we could have no changes in the Payload. So we can early detect an inconsistency with builder milestone and fallback to locally immediately

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/schema/Response.java
```diff
@@ -41,15 +45,39 @@ public static <T> Response<T> withErrorMessage(final String errorMessage) {
...
+    final boolean milestonesAreMatching = receivedMilestone.equals(expectedMilestone);
+
+    if (strictVersionCheck && !milestonesAreMatching) {
+      throw new IllegalArgumentException(
+          "Wrong response version: expected "
+              + expectedMilestone
+              + ", received "
+              + receivedMilestone);
+    } else {
+      LOG.warn(
```
StefanBratanov: Isn't this logic always going to log warning, regardless if milestones are matching or not for `getPayload`?

tbenr: 🤦

StefanBratanov (approved):
LGTM

PR #7011, opened 2023-04-03
Implement historical sync for Deneb

### LINKED ISSUE #7010: Implement Historical Syncing for Deneb
Import blob sidecars (if required) for a block, validate and save in database

StefanBratanov:
## PR Description
When historically syncing blocks: 
- retrieve blobs which are required to be available
- validate them
- save them to database along with the blocks

## Fixed Issue(s)
fixes #7010 

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
```diff
@@ -178,53 +193,136 @@ private boolean batchIsComplete() {
...
   }
 
-  private boolean shouldRetryBlocksByRangeRequest() {
+  private boolean shouldRetryBlocksAndBlobSidecarsByRangeRequest() {
     return !batchIsComplete() && requestCount.incrementAndGet() < maxRequests;
   }
 
-  private SafeFuture<Void> requestBlockByHash() {
-    LOG.trace("Request next historical block directly by hash {}", lastBlockRoot);
+  private SafeFuture<Void> requestBlockByRoot() {
```
tbenr, line 252: We where discussing with @zilm13 that jumping gaps with by root works only for recent blocks, peers are not expected to serve the byRoot rpc for old blocks. Which means:
1- we should remove it and use only by range.
2- do another fallback to byRange (if by root fails) and from that moment use only that

this could be a dedicated PR :)

tbenr (approved):
LGTM. just a comment for a future PR

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
StefanBratanov: Cool, so use byRange but only for one slot in this case? It will make things slightly easier actually as well since can just request blob sidecars rather than calculate number of commitments and then create BlobIdentifier list.

PR #7015, opened 2023-04-04
Capella upgrade AT using real EL

lucassaldanha:
## PR Description
- This is the first time we are going through testing a milestone using real EL after the merge.
- Using the option of overriding EL genesis to set `shanghaiTime`
- Forcing Teku to use a chosen `genesisTime`
- Calculating `shanghaiTime` from `genesisTime`


zilm13:
Closing as similar test (made from this one) is added to #7178

PR #7016, opened 2023-04-04
Perform readiness check when starting the VC

StefanBratanov:
## PR Description
Perform readiness check on startup when failovers are configured to avoid calling `eth/v1/beacon/states/head/validators`  (that endpoint does not return 503 when node is still syncing) on a node which is not synced.

Reported by a Discord user: https://discordapp.com/channels/697535391594446898/697539289042649190/1092731844841185360

Avoids behaviour where initial validator statuses are retrieved on a node which is not synced. (`Unable to retrieve status for validator`) This can cause a validator to not attest on the epoch the VC starts:
image
image

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
@@ -28,6 +28,7 @@ import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.logging.ValidatorLogger;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.service.serviceutils.Service;
 import tech.pegasys.teku.validator.api.ValidatorTimingChannel;
 
 /**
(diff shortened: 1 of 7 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -96,19 +107,18 @@ public void onAttestationCreationDue(final UInt64 slot) {}
 
   @Override
   public void onAttestationAggregationDue(final UInt64 slot) {
+    performReadinessCheckAgainstAllNodes().ifExceptionGetsHereRaiseABug();
```
tbenr, line 110: is this actually a Bug if it happens?

tbenr: seems to be yes :)

StefanBratanov: yeah this should never happen unless something very bad is happening, the futures handle `exceptionally` so they should not fail.

tbenr (approved):
LGTM

PR #7018, opened 2023-04-05
Implement `BlobSidecarsAvailabilityChecker`

StefanBratanov:
## PR Description

- Introduced `BlobSidecarsAvailabilityChecker` and an implementation of `ForkChoiceBlobSidecarsAvailabilityChecker`
- Changed `MiscHelpersDeneb` to use `verifyBlobKzgProofBatch`

## Fixed Issue(s)
resolves a task in #6822 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -63,32 +63,59 @@ private static KZG initKZG(final SpecConfigDeneb config) {
...
+              blobSidecar.getSlot(),
+              slot);
+          checkArgument(
+              beaconBlockRoot.equals(blobSidecar.getBlockRoot()),
+              "Blob sidecar block root %s does not match block root %s",
+              blobSidecar.getBlockRoot(),
+              beaconBlockRoot);
+        });
+    checkArgument(
+        kzgCommitments.size() == blobSidecars.size(),
```
zilm13: I think we'd better follow spec stricter here:
```python
    assert len(expected_kzg_commitments) == len(blobs)
    assert len(blobs) == len(proofs)
```
So, I'd prefer check matching after unpacking blobs from blobSidecars. Also it would be good to check that number of proofs match. There is no place where it could get  not equal in current code, but I'm a supporter of stricter spec following in such cases. Spec could be changed in the future and it'll be easier to follow changes when we just copy it.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/blobs/BlobSidecarsAvailabilityChecker.java
```diff
@@ -0,0 +1,190 @@
...
+   *     validation
+   */
+  boolean initiateDataAvailabilityCheck();
+
+  SafeFuture<BlobSidecarsAndValidationResult> getAvailabilityCheckResult();
+
+  /** Only perform the {@link MiscHelpersDeneb#isDataAvailable} check */
+  SafeFuture<BlobSidecarsAndValidationResult> validate(List<BlobSidecar> blobSidecars);
+
+  enum BlobSidecarsValidationResult {
```
zilm13: I think `BlobSidecarsAvailabilityChecker` will be more readable if we move this enum outside in separate class

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/blobs/BlobSidecarsAvailabilityChecker.java
```diff
@@ -0,0 +1,190 @@
...
+    NOT_REQUIRED,
+    NOT_AVAILABLE,
+    INVALID,
+    VALID
+  }
+
+  SafeFuture<BlobSidecarsAndValidationResult> NOT_REQUIRED_RESULT_FUTURE =
+      SafeFuture.completedFuture(BlobSidecarsAndValidationResult.NOT_REQUIRED);
+
+  class BlobSidecarsAndValidationResult {
```
zilm13: I think `BlobSidecarsAvailabilityChecker` will be more readable if we move this class from inner to separate class. The constant above could be moved with it too

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private int getNumberOfKzgCommitmentsInBlock() {
+    return block
+        .getMessage()
+        .getBody()
+        .toVersionDeneb()
+        .orElseThrow()
+        .getBlobKzgCommitments()
+        .size();
+  }
```
zilm13: We could define instead such method and use it twice
```java
  private int getKzgCommitments() {
    return block
        .getMessage()
        .getBody()
        .toVersionDeneb()
        .orElseThrow()
        .getBlobKzgCommitments();
  }
```

zilm13 (commented):
Some nit picks

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
StefanBratanov: I agree. Changing it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
StefanBratanov: you are right, thanks! this makes it easier

tbenr (approved):
LGTM

PR #7021, opened 2023-04-08
Add BlobSidecar getters in CombinedChainDataClient

### LINKED ISSUE #6821: [DECOUPLING] Implement getBlobSidecarBySlotAndBlockRoot
Implement `getBlobSidecarBySlotAndBlockRoot` in `CombinedChainDataClient`

zilm13:
## PR Description
There were 3 possible approaches to implement storage for BlobSidecars:
Using  `Slot+BlockRoot+BlobIndex` key, `BlockRoot+BlobIndex` key or `Slot+BlobIndex` key. Currently we have these type of queries:
 - A. By Slot range RPC, Slot query
 - B. By BlockIdentifiers RPC, BlockRoot + BlobIndex query 
 - C. Earliest available epoch query
 - D. By Slot for pruning, similar to `A`

`A` could be used both for recent missed slots and historical sync. `B` is used only for recent slots but we assume it could be changed in the future, because it's very useful to fill several small gaps in one query. `C` is used once per query along with `A`. `D` is happen once in epoch. We could get following breakdown of operation complexity by chosen key type:
  1. `Slot+BlockRoot+BlobIndex` key
  	A+D. For each slot query slot -> Block to get root and indexes
  	B. For each blobsidecar search for a block to get slot
  	C. First element
  2. `BlockRoot+BlobIndex` key
  	A+D. For each slot query slot -> Block to get root and indexes
  	B. No additional queries
  	C. Extra query for a block or we need to use DB variable with some in memory cache for it
  3. `Slot+BlobIndex` key
  	A. No additional queries
  	B. For each blobSidecar search for a block to get slot
  	C+D. First element

For current usage we assume getting a slot for `B` will be very fast as it's memory operation (for recent slots). Plus for `A` and `D` we could use slot order streaming. It makes `1` and `3` preferrable with not much difference between `1` and `3` when using DB streaming. For clarity we choose `Slot+BlockRoot+BlobIndex` key.

We cannot use DB stream in `BlobSidecarsByRangeMessageHandler` because of the application design and strict requirements for stream closing. Also we couldn't pass streams through async interfaces. There are 2 possible approaches to get this around:
  1. Stream keys and collect it before closing. Key weight should be about 16 (object header) + 24 (UInt64) + 48 (Bytes32) + 24 (UInt64) + 8 bytes list reference = 120 bytes per `SlotAndBlockRootAndBlobIndex`. So maximum 61440 bytes for 128 blocks with 4 blobSidecars in each. 16 peers with maximum possible queries will consume under 1Mb memory for it which sounds acceptable for me
  2. Something tricky like `getNextKeyAfter(V value)` will void opportunity of stream by calling for a lot times plus follows to the weird code.

So I've chosen the 1st

Confirmed/unconfirmed logic, `BlobSidecar` persistence and pruning is not covered by this PR.
Also I'm not sure on the placement of `SlotAndBlockRootAndBlobIndex`.

## Fixed Issue(s)

Fixes #6821 
## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
@@ -327,8 +327,7 @@ public class Spec {
     return atSlot(slot)
         .getSchemaDefinitions()
         .toVersionDeneb()
-        .orElseThrow(
-            () -> new RuntimeException("Deneb milestone is required to load execution payload"))
+        .orElseThrow(() -> new RuntimeException("Deneb milestone is required to load blob sidecar"))
         .getBlobSidecarSchema()
         .sszDeserialize(serializedBlobSidecar);
   }
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
@@ -323,6 +323,16 @@ public class Spec {
         .sszDeserialize(serializedPayload);
   }
 
+  public BlobSidecar deserializeBlobSidecar(final Bytes serializedBlobSidecar, final UInt64 slot) {
+    return atSlot(slot)
+        .getSchemaDefinitions()
+        .toVersionDeneb()
+        .orElseThrow(
+            () -> new RuntimeException("Deneb milestone is required to load execution payload"))
+        .getBlobSidecarSchema()
+        .sszDeserialize(serializedBlobSidecar);
+  }
+
   public BlobsSidecar deserializeBlobsSidecar(
       final Bytes serializedBlobsSidecar, final UInt64 slot) {
     return atSlot(slot)
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
@@ -45,6 +45,10 @@ public class SlotAndBlockRootAndBlobIndex implements Comparable<SlotAndBlockRoot
     return blobIndex;
   }
 
+  public boolean hasNoBlobs() {
+    return blobIndex.equals(UInt64.MAX_VALUE);
+  }
+
   @Override
   public boolean equals(final Object o) {
     if (this == o) {
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
@@ -33,7 +33,8 @@ public class SlotAndBlockRootAndBlobIndex implements Comparable<SlotAndBlockRoot
     this.blobIndex = blobIndex;
   }
 
-  public static SlotAndBlockRootAndBlobIndex createWithNoBlobs(final UInt64 slot, final Bytes32 blockRoot) {
+  public static SlotAndBlockRootAndBlobIndex createWithNoBlobs(
+      final UInt64 slot, final Bytes32 blockRoot) {
     return new SlotAndBlockRootAndBlobIndex(slot, blockRoot, UInt64.MAX_VALUE);
   }
 
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
@@ -22,6 +22,8 @@ import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 
 /** Key for storing blobs in DB */
 public class SlotAndBlockRootAndBlobIndex implements Comparable<SlotAndBlockRootAndBlobIndex> {
+  public static final UInt64 NO_BLOBS_INDEX = UInt64.MAX_VALUE;
+
   private final UInt64 slot;
   private final Bytes32 blockRoot;
   private final UInt64 blobIndex;
@@ -33,9 +35,9 @@ public class SlotAndBlockRootAndBlobIndex implements Comparable<SlotAndBlockRoot
     this.blobIndex = blobIndex;
   }
 
-  public static SlotAndBlockRootAndBlobIndex createWithNoBlobs(
+  public static SlotAndBlockRootAndBlobIndex createNoBlobsKey(
       final UInt64 slot, final Bytes32 blockRoot) {
-    return new SlotAndBlockRootAndBlobIndex(slot, blockRoot, UInt64.MAX_VALUE);
+    return new SlotAndBlockRootAndBlobIndex(slot, blockRoot, NO_BLOBS_INDEX);
   }
 
   public UInt64 getSlot() {
@@ -50,8 +52,8 @@ public class SlotAndBlockRootAndBlobIndex implements Comparable<SlotAndBlockRoot
     return blobIndex;
   }
 
-  public boolean hasNoBlobs() {
-    return blobIndex.equals(UInt64.MAX_VALUE);
+  public boolean isNoBlobsKey() {
+    return blobIndex.equals(NO_BLOBS_INDEX);
   }
 
   @Override
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright ConsenSys Software Inc., 2022
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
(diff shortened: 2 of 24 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -323,6 +323,16 @@ public ExecutionPayload deserializeExecutionPayload(
         .sszDeserialize(serializedPayload);
   }
 
+  public BlobSidecar deserializeBlobSidecar(final Bytes serializedBlobSidecar, final UInt64 slot) {
+    return atSlot(slot)
+        .getSchemaDefinitions()
+        .toVersionDeneb()
+        .orElseThrow(
+            () -> new RuntimeException("Deneb milestone is required to load execution payload"))
```
tbenr: `... to load blob sidecar`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -213,103 +201,52 @@ class RequestState {
...
-                && currentIndex.get().equals(getMaxBlobSidecarIndex())) {
-              maybeCurrentBlock = Optional.empty();
-            }
-          });
-    }
-
-    private UInt64 getMaxBlobSidecarIndex() {
-      return UInt64.valueOf(blobSidecarsCount).minusMinZero(UInt64.ONE);
+    boolean isComplete() {
+      return endSlot.isLessThan(startSlot)
```
tbenr, line 248: this is strange. Shouldn't we check the message on `validateRequest` so this never occurs?

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandlerTest.java
```diff
@@ -191,18 +188,14 @@ public void shouldSendResourceUnavailableIfBlobSidecarsAreNotAvailable() {
 
   @Test
   public void shouldCompleteSuccessfullyIfRequestNotWithinRange() {
```
tbenr: the name of this test is a bit generic. Are we testing "no blob sidecars in the requested range"?

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
```diff
@@ -478,25 +477,40 @@ public SafeFuture<Optional<SignedBeaconBlock>> getBlockByBlockRoot(final Bytes32
...
-  }
-
-  @SuppressWarnings("unused")
   public SafeFuture<Optional<BlobSidecar>> getBlobSidecarByBlockRootAndIndex(
       final Bytes32 blockRoot, final UInt64 index) {
-    return SafeFuture.failedFuture(new UnsupportedOperationException("Not yet implemented"));
+    return recentChainData
+        .getSlotForBlockRoot(blockRoot)
+        .map(slot -> getBlobSidecarByKey(new SlotAndBlockRootAndBlobIndex(slot, blockRoot, index)))
+        .orElseGet(
```
tbenr: I don't like this `orElseGet` after a `map` can be confusing to read. (if becomes a`flatMap` it'd behave completely differently) . I'd clearly separate the two cases.
(and there will be less nesting level of the second case)

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/serialization/SlotAndBlockRootAndBlobIndexKeySerializer.java
```diff
@@ -0,0 +1,63 @@
...
+  static final int BLOB_INDEX_OFFSET = Bytes32.SIZE + UINT64_SIZE;
+
+  @Override
+  public SlotAndBlockRootAndBlobIndex deserialize(final byte[] data) {
+    final UInt64 slot =
+        UInt64.fromLongBits(
+            Longs.fromBytes(
+                data[0], data[1], data[2], data[3], data[4], data[5], data[6], data[7]));
+    final Bytes rootBytes = Bytes.wrap(data, UINT64_SIZE, Bytes32.SIZE);
+    final Bytes blobIndexBytes = Bytes.wrap(data, BLOB_INDEX_OFFSET, UINT64_SIZE);
```
tbenr, line 40: do we want to allocate 8 bytes for the index? seem like an overkill. How many blobs cold we ever have even in full sharding?

tbenr (commented):
Added some comments. Overall good btw.

tbenr:
We need to consider the fact that, in the current design, we need to write something on DB when blobs are missing (zero commitments) for the slot-blockroot, so the availability  window check isn't broken (especially for testnet). 
An option is to store a 0 byte (like `VoidSerializer`) for the 0 `BlobIndex`. And then handle it in `deserializeBlobSidecar` which simply return empty if you try to deserialize a 0 byte.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/serialization/SlotAndBlockRootAndBlobIndexKeySerializer.java
zilm13: it's uint64 by the spec

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
zilm13: copycat error

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
zilm13: Request is `startSlot, count`, we consider request (19, 0) as legal and don't count it as error. Internally we interpret it as `startSlot, endSlot` as (19, 18) because (19, 19) should return blobs for 19th slot. I don't want penalize peers for 0 count request by the policy "forgive everything not criminal,  but don't act yourself to be punished (even if you don't punish for such things)"

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
zilm13: Actually, such constructions are not my favorites too, will get pleasure from refactoring it

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/serialization/SlotAndBlockRootAndBlobIndexKeySerializer.java
zilm13: I'm not sure we could get any real economy here even if not following spec, everything is 64bit in java for example. even if we make it smaller. I'm pretty sure RocksDB will allocate more than 1byte for it even if we make it 1byte, which could be risk for the spec

zilm13:
@tbenr I've used MAX_VALUE for `blobIndex` for empty blobs, it looks clearer for me, what do you think?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/serialization/SlotAndBlockRootAndBlobIndexKeySerializer.java
tbenr: I was thinking on more on level DB side rather than java. At most we could shave at most like 6 bytes, from 48 to 42, if we will allow 2^16 (65k) blobs. Maybe it doesn't count much.. for the spec i think they are just reusing Uint64 for ssz convenience.

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
```diff
@@ -485,22 +485,27 @@ public SafeFuture<Optional<UInt64>> getEarliestAvailableBlobSidecarEpoch() {
...
-                                                blockOptional.get().getSlot(), blockRoot, index)))
-                                .orElse(SafeFuture.completedFuture(Optional.empty()))));
+    if (index.equals(UInt64.MAX_VALUE)) {
+      return SafeFuture.completedFuture(Optional.empty());
+    }
+    final Optional<UInt64> hotSlotForBlockRoot = recentChainData.getSlotForBlockRoot(blockRoot);
+    if (hotSlotForBlockRoot.isPresent()) {
+      return getBlobSidecarByKey(
+          new SlotAndBlockRootAndBlobIndex(hotSlotForBlockRoot.get(), blockRoot, index));
+    } else {
```
tbenr: you can skip this `else` and indent down even more :)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/SlotAndBlockRootAndBlobIndex.java
```diff
@@ -45,6 +45,10 @@ public UInt64 getBlobIndex() {
     return blobIndex;
   }
 
+  public boolean hasNoBlobs() {
```
tbenr: I'd put a `static Uint64 NO_BLOBS_INDEX = UInt64.MAX_VALUE;`
and call this method `isNoBlobsKey()`

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
```diff
@@ -485,22 +485,27 @@ public SafeFuture<Optional<UInt64>> getEarliestAvailableBlobSidecarEpoch() {
...
-                    .thenCompose(
-                        blockOptional ->
-                            blockOptional
-                                .map(
-                                    block ->
-                                        getBlobSidecarByKey(
-                                            new SlotAndBlockRootAndBlobIndex(
-                                                blockOptional.get().getSlot(), blockRoot, index)))
-                                .orElse(SafeFuture.completedFuture(Optional.empty()))));
+    if (index.equals(UInt64.MAX_VALUE)) {
```
tbenr: then use `NO_BLOBS_INDEX` here for clarity

zilm13:
Good points, refactored a bit

tbenr:
I'll give a final check tomorrow

tbenr (approved):
LGTM

PR #7024, opened 2023-04-11
Create getBlockRewardsFromBlockId method in ChainDataProvider

courtneyeh:
#6702 

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/migrated/BlockRewardData.java
```diff
@@ -17,16 +17,16 @@
...
   private final Long attestations;
   private final UInt64 syncAggregate;
   private final UInt64 proposerSlashings;
   private final UInt64 attesterSlashings;
 
   public BlockRewardData(
-      final UInt64 proposerIndex,
-      final UInt64 total,
+      final Integer proposerIndex,
+      final Long total,
```
rolfyone: should be able to use primitives rather than Long / Integer

data/serializer/src/main/java/tech/pegasys/teku/api/migrated/BlockRewardData.java
```diff
@@ -39,11 +39,11 @@ public BlockRewardData(
     this.attesterSlashings = attesterSlashings;
   }
 
-  public UInt64 getProposerIndex() {
+  public Integer getProposerIndex() {
```
rolfyone: same with return types, can use primitives

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -684,14 +687,80 @@ protected SyncCommitteeRewardData calculateRewards(
...
+  protected UInt64 calculateProposerSyncAggregateBlockRewards(
       UInt64 proposerReward, SyncAggregate aggregate) {
     final SszBitvector syncCommitteeBits = aggregate.getSyncCommitteeBits();
-    int total = 0;
+    UInt64 total = ZERO;
     for (int i = 0; i < syncCommitteeBits.size(); i++) {
       if (syncCommitteeBits.getBit(i)) {
-        total += proposerReward.intValue();
+        total = total.plus(proposerReward);
       }
```
rolfyone: so really here we're counting the bits set, which we could just do with an int count, then `return proposerReward.times(count)`

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -684,14 +687,80 @@ protected SyncCommitteeRewardData calculateRewards(
...
+    final long attestationsBlockRewards = calculateAttestationRewards(state);
+    final UInt64 syncAggregateBlockRewards =
+        calculateProposerSyncAggregateBlockRewards(proposerReward, aggregate);
+    final UInt64 proposerSlashingsBlockRewards = calculateProposerSlashingsRewards(block, state);
+    final UInt64 attesterSlashingsBlockRewards = calculateAttesterSlashingsRewards(block, state);
+    final long total =
+        attestationsBlockRewards
+            + syncAggregateBlockRewards.longValue()
+            + proposerSlashingsBlockRewards.longValue()
+            + attesterSlashingsBlockRewards.longValue();
```
rolfyone: this looks very much like we should be computing longs and returning longs from functions...

courtneyeh: I guess I did it that way, because thats just adding up the total, and the other values are supposed to be UInt64? I could change that if we wanted to change the structure of the output data.

rolfyone: there's .longValue everywhere which suggests maybe worth taking a look... if we start with primitives it may be simpler in some ways

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -719,30 +719,31 @@ private ObjectAndMetaData<BlockRewardData> getBlockRewardData(
...
+    final long proposerSlashingsBlockRewards = calculateProposerSlashingsRewards(block, state);
+    final long attesterSlashingsBlockRewards = calculateAttesterSlashingsRewards(block, state);
     final long total =
         attestationsBlockRewards
-            + syncAggregateBlockRewards.longValue()
-            + proposerSlashingsBlockRewards.longValue()
-            + attesterSlashingsBlockRewards.longValue();
+            + syncAggregateBlockRewards
+            + proposerSlashingsBlockRewards
+            + attesterSlashingsBlockRewards;
```
rolfyone: nit, could just let the BlockRewardData do the increment given its summing the fields you pass to it, and its a reasonable expectation for it to be able to do that...

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/rewards/GetBlockRewardsTest.java
```diff
@@ -35,14 +35,11 @@
...
 import tech.pegasys.teku.infrastructure.http.HttpStatusCodes;
-import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.datastructures.metadata.ObjectAndMetaData;
 
 public class GetBlockRewardsTest extends AbstractMigratedBeaconHandlerTest {
-  private final BlockRewardData data =
-      new BlockRewardData(
-          123, 123L, 123L, UInt64.valueOf(123), UInt64.valueOf(123), UInt64.valueOf(123));
+  private final BlockRewardData data = new BlockRewardData(123, 123L, 123L, 123L, 123L, 123L);
```
rolfyone: better to use unique values so you can identify where they're going... can just be `121,122,123...`

data/serializer/src/main/java/tech/pegasys/teku/api/migrated/BlockRewardData.java
```diff
@@ -14,23 +14,22 @@
...
 
   public BlockRewardData(
-      final UInt64 proposerIndex,
-      final UInt64 total,
-      final Long attestations,
-      final UInt64 syncAggregate,
-      final UInt64 proposerSlashings,
-      final UInt64 attesterSlashings) {
+      final int proposerIndex,
+      final long total,
```
rolfyone: the more i think about it the more im convinced we just do our addition inside the class and don't initialise with total, it removes the chance of it being accumulated to the wrong value at some random point in code..

rolfyone (approved):
looks good, pending how it compares to other implemented responses...

PR #7026, opened 2023-04-12
Introduces BlobSidecarPool with block-blobs tracking

### LINKED ISSUE #6849: Implement `BlobSidecar` pooling

tbenr:
The basic idea is to have a pool of trackers for each `blockRoot` we see (coming from a block or from a sidecar)

- each tracker have the responsibility to signal when a block-blobs bundle is completed
- the pool is currently responsible of signal to fetchers (block and blobs) to retrieve missing content. This logic is triggered only when dealing with object related to current slot (following the head, content coming from gossip)

- the pool implements a fetch delay logic. There are three constants:
  - `TARGET_WAIT_MILLIS`: ideal delay before fetching (currently set to 1s)
  - `MIN_WAIT_MILLIS`: minimum delay before fetching (currently set to 0.5s). The rationale is that we don't want to issue fetches if we are most probably still receiving block and blobs.
  - `MAX_WAIT_RELATIVE_TO_ATT_DUE_MILLIS`: maximum "distance" from attestation due (4s) when we really want to have block and blobs to process (currently set to 1.5s). The rationale is that, if we are close to attestation due, even if we may still receive something from gossip, we are in rush and we can waste some bandwidth to have more chance to import before the 4s mark (and attest to correct head)

- currently if we try to fetch blobsidecars and we don't have the corresponding block yet, we try to fetch all possible sidecars (considering maxBlobsPerBlock). If we receive the block after fetch occurs, we drop the (eventually) non existing blobs.

fixes #6849

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/SignedBlobSidecar.java
```diff
@@ -31,11 +32,19 @@ public SignedBlobSidecar(
     super(schema, blobSidecar, new SszSignature(signature));
   }
 
+  public BlobSidecar getMessage() {
```
zilm13, line 35: Do we really need this? Because it's Signed*? Maybe introduce interface in this case. I have not find where it's used

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -2327,18 +2379,19 @@ public SignedBeaconBlockAndBlobsSidecar randomConsistentSignedBeaconBlockAndBlob
   }
 
   public SszList<SszKZGCommitment> randomSszKzgCommitmentList() {
+    final int count = randomInt(spec.getMaxBlobsPerBlock().orElseThrow()) + 1;
+    return randomSszKzgCommitmentList(count);
+  }
+
+  public SszList<SszKZGCommitment> randomSszKzgCommitmentList(final int conut) {
```
zilm13: `count`

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -875,6 +875,25 @@ public SignedBeaconBlock randomSignedBeaconBlockWithEmptyCommitments() {
     return signedBlock(beaconBlock);
   }
 
+  public SignedBeaconBlock randomSignedBeaconBlockWithCommitments(int count) {
```
zilm13: final please

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/gossip/blocks/FetchRecentBlocksService.java
```diff
@@ -36,28 +37,37 @@
 
   private final ForwardSync forwardSync;
   private final PendingPool<SignedBeaconBlock> pendingBlocksPool;
+  private final BlobSidecarPool blobSidecarPool;
```
zilm13, line 40: very nit comment: there is one called with plural above, so it's noticed. Maybe `blobSidecarsPool`.
UPDATE: checked other pools and looks like we'd better rename `pendingBlocksPool`

beacon/sync/src/testFixtures/java/tech/pegasys/teku/beacon/sync/SyncingNodeManager.java
```diff
@@ -101,6 +103,7 @@ public static SyncingNodeManager create(
       Consumer<Eth2P2PNetworkBuilder> configureNetwork)
       throws Exception {
     final Spec spec = TestSpecFactory.createMinimalPhase0();
+    final TimeProvider timeProvider = new SystemTimeProvider();
```
zilm13: I think it's better to move it close to the first usage

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobSidecarPool.java
```diff
@@ -15,21 +15,31 @@
...
+        @Override
+        public void onNewBlock(final SignedBeaconBlock block) {}
 
         @Override
         public boolean containsBlobSidecar(final BlobIdentifier blobIdentifier) {
           return false;
         }
 
+        @Override
+        public BlockBlobSidecarsTracker getBlockBlobsSidecarsTracker(SignedBeaconBlock block) {
```
zilm13, line 44: `getBlockBlobSidecarsTracker`? and base method

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  public boolean containsBlobSidecar(final BlobIdentifier blobIdentifier) {
+    return Optional.ofNullable(blobSidecars.get(blobIdentifier.getIndex()))
+        .map(blobSidecar -> blobSidecar.getBlockRoot().equals(blobIdentifier.getBlockRoot()))
+        .orElse(false);
+  }
+
+  public Stream<BlobIdentifier> getMissingBlobSidecars() {
+    blockBody.get();
```
zilm13: looks like traces

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -0,0 +1,143 @@
...
+        "Wrong blobSidecar block root");
+
+    if (blobSidecarsComplete.isDone()) {
+      // already completed
+      return false;
+    }
+
+    boolean addedNew = blobSidecars.put(blobSidecar.getIndex(), blobSidecar) == null;
+
+    if (addedNew) {
```
zilm13, line 111: !addedNew? Do we need to check completion in this case?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -0,0 +1,143 @@
...
+      new AtomicReference<>(Optional.empty());
+
+  private final Map<UInt64, BlobSidecar> blobSidecars = new ConcurrentHashMap<>();
+  private final SafeFuture<Void> blobSidecarsComplete = new SafeFuture<>();
+
+  public BlockBlobSidecarsTracker(final SlotAndBlockRoot slotAndBlockRoot) {
+    this.slotAndBlockRoot = slotAndBlockRoot;
+  }
+
+  public Map<UInt64, BlobSidecar> getBlobSidecars() {
```
zilm13: How do you think to remove `public` and add VisibleForTesting to clarify usage

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
```diff
@@ -0,0 +1,105 @@
...
+    final UInt64 slotLimit = calculateItemAgeLimit();
+    return slot.compareTo(slotLimit) <= 0;
+  }
+
+  private boolean isSlotFromAFinalizedSlot(final UInt64 slot) {
+    return slot.compareTo(latestFinalizedSlot) <= 0;
+  }
+
+  private UInt64 calculateItemAgeLimit() {
+    return currentSlot.compareTo(historicalSlotTolerance.plus(UInt64.ONE)) > 0
```
zilm13: `isGreaterThan` looks more clear

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
```diff
@@ -0,0 +1,105 @@
...
+    return slot.isGreaterThan(slotLimit);
+  }
+
+  private boolean isSlotOutsideOfHistoricalLimit(final UInt64 slot) {
+    final UInt64 slotLimit = calculateItemAgeLimit();
+    return slot.compareTo(slotLimit) <= 0;
+  }
+
+  private boolean isSlotFromAFinalizedSlot(final UInt64 slot) {
+    return slot.compareTo(latestFinalizedSlot) <= 0;
```
zilm13: `isLessThanOrEqual`?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private boolean isSlotFromFarFuture(final UInt64 slot) {
+    final UInt64 slotLimit = calculateFutureItemLimit();
+    return slot.isGreaterThan(slotLimit);
+  }
+
+  private boolean isSlotOutsideOfHistoricalLimit(final UInt64 slot) {
+    final UInt64 slotLimit = calculateItemAgeLimit();
+    return slot.compareTo(slotLimit) <= 0;
```
zilm13: `isLessThanOrEqual`?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
```diff
@@ -0,0 +1,105 @@
...
+import static tech.pegasys.teku.spec.config.SpecConfig.GENESIS_SLOT;
+
+import com.google.common.annotations.VisibleForTesting;
+import tech.pegasys.teku.ethereum.events.SlotEventsChannel;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.datastructures.state.Checkpoint;
+import tech.pegasys.teku.storage.api.FinalizedCheckpointChannel;
+
+abstract class AbstractIgnoringFutureHistoricalSlot
```
zilm13, line 33: Maybe add some brief description for it?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
```diff
@@ -0,0 +1,369 @@
...
+
+  @Override
+  public synchronized BlockBlobSidecarsTracker getBlockBlobsSidecarsTracker(
+      final SignedBeaconBlock block) {
+    return getOrCreateBlobSidecarsTracker(block.getSlotAndBlockRoot(), __ -> {});
+  }
+
+  @VisibleForTesting
+  @Override
+  synchronized void prune(final UInt64 slotLimit) {
```
zilm13, line 241: `slotLimit` sounds like some delta for me. Maybe use something like `maxSlot` or similar here?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
```diff
@@ -0,0 +1,369 @@
...
+      return;
+    }
+
+    final SlotAndBlockRoot slotAndBlockRoot = blobSidecar.getSlotAndBlockRoot();
+
+    final BlockBlobSidecarsTracker blobSidecarsTracker =
+        getOrCreateBlobSidecarsTracker(
+            slotAndBlockRoot,
+            newTracker -> {
+              if (!syncing) {
```
zilm13: I don't understand why are we checking syncing here and not checking for BlobSidecar. If it's intentional, it worth a comment

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
```diff
@@ -0,0 +1,369 @@
...
+
+    if (blockBlobSidecarsTracker.checkCompletion()) {
+      return;
+    }
+
+    if (blockBlobSidecarsTracker.getBlockBody().isEmpty()) {
+      requiredBlockRootSubscribers.deliver(
+          RequiredBlockRootSubscriber::onRequiredBlockRoot,
+          blockBlobSidecarsTracker.getSlotAndBlockRoot().getBlockRoot());
+      return;
```
zilm13: I'd add comment something that we don't know number of blobs yet

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
```diff
@@ -0,0 +1,369 @@
...
+
+    final UInt64 upperLimitRelativeToAttDue =
+        attestationDueMillis.minus(MAX_WAIT_RELATIVE_TO_ATT_DUE_MILLIS);
+
+    final UInt64 targetMillis = nowMillis.plus(TARGET_WAIT_MILLIS);
+
+    final UInt64 finalTime =
+        targetMillis.min(upperLimitRelativeToAttDue).max(nowMillis.plus(MIN_WAIT_MILLIS));
+
+    return Optional.of(Duration.ofMillis(finalTime.minus(nowMillis).intValue()));
```
zilm13, line 372: I'm not sure on all this logic if we take Gnosis. Do you know their plans for Blobs?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -0,0 +1,143 @@
...
+    return Optional.ofNullable(blobSidecars.get(blobIdentifier.getIndex()))
+        .map(blobSidecar -> blobSidecar.getBlockRoot().equals(blobIdentifier.getBlockRoot()))
+        .orElse(false);
+  }
+
+  public Stream<BlobIdentifier> getMissingBlobSidecars() {
+    blockBody.get();
+    if (blockBody.get().isEmpty()) {
+      // TODO: block is still unknown.
+      //  Should we return all potential maxBlobsPerBlock BlobIdentifiers?
```
zilm13: I think we don't need it at least at the beginning. Let's see, how it works without it. If we want to add something smart here, we could add following: if we have a BlobSidecar with non-0 index, we could expect blobSidecar's to the left

zilm13 (commented):
Looks good, but I'm a bit paranoid on synchronization pattern. I think we have some difference here with other pending pools which may lead to dead locks or slow behaviour. I will take a look again tomorrow with a fresh head

tbenr:
@zilm13 yes you're right, I promised my self to review synchronization. I'll do a second pass tomorrow

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
tbenr: good point... I think at the end it may hold even in gnosis timings.. i'll think about it

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobSidecarPool.java
tbenr: @zilm13 what do you mean by that? please explain :)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
tbenr: good points

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
tbenr: no because the plan is to use method in the availability checker to get current blobSidecars and perform the batch check on those while waiting for completion.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
tbenr: yeah, those comparisons were just copy\paste of the old implementation. changing them.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
tbenr: to me sounds ok, at least I prefer it compared to `maxSlot` or other.
In the context of pruning, is the limit the pruning has when doing its pruning job

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
tbenr: it is currently intentional but, as I mention in the PR description, it may go away at the end (just let the `firstSeen` checks go, and will not trigger any fetch by design since, by definition, when we sync we are dealing with old slots)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/SignedBlobSidecar.java
tbenr: yep.. I did it ages ago for consistency with other signed objects and remained there.. we don't have a `SignedMessage` interface so it isn't really a thing..

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
tbenr: ahh.. yeah...

tbenr: @zilm13 implemented

tbenr: implemented!

tbenr: even if, thinking about it again, I'm not sure it adds any value. If the objective is to try our best to import the block in time, this is not enough. If the block fetch is successful we may realize that we don't have requested all we need, so we will not be able to complete block import.

zilm13: Assume we fill only one blob once in a 2 blocks in the beginning in Mainnet. If every Teku requests all possible variants in advance, it will look very spammy. My proposal will not work all the times but at least it will not spam network

tbenr: true, but I don't like the probabilistic approach: if you're lucky that you missed the "left side" you can successfully recover, otherwise you cannot. So, for know I added the full "spammy" approach, hoping that we can tune constants in a way that we fall in that situation sufficiently rarely that we can afford to be "spammy". We will see in real devnets\testnets if worth to have it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/BlobSidecarPoolImpl.java
tbenr: I think for fast slot networks, we would end up being always on the `MIN_WAIT_MILLIS` which makes sense to me.
And I think it is acceptable that value being constant instead of being dependant by slot time, since that number is more related to network latency than actual slot time. We could argue that on fast nets, blocks (and blobs?) could be smaller so the expected latency could be lower on average, but for now seems like we can leave with a constant.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -0,0 +1,172 @@
...
+    return blobSidecars.size();
+  }
+
+  public boolean setBlock(final SignedBeaconBlock block) {
+    checkArgument(block.getSlotAndBlockRoot().equals(slotAndBlockRoot), "Wrong block");
+    final Optional<BeaconBlockBodyDeneb> oldBlock =
+        blockBody.getAndSet(
+            Optional.of(BeaconBlockBodyDeneb.required(block.getMessage().getBody())));
+    if (oldBlock.isPresent()) {
+      LOG.debug("block was already set!");
```
zilm13: It's useless without any extra info (block info or whatever), and we could hit it very often. I'd remove it

zilm13 (approved):
LGTM

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
tbenr: going to remove it for now, will put it in the debugging PR to show timings

PR #7028, opened 2023-04-13
BlobSidecar's persistence

zilm13:
## PR Description
Main part of `BlobSidecar` persistence
- `BlobsSidecar` removal from `ChainBuilder` started
- `BlobSidecarManager` renamed because it's already more of `BlobSidecar` to avoid confusion
- Pruning needed to be done to enable some tests, I will make pruning in separate PR
- `Store` and `StoreTest` should be reworked with `SignedBlockContents` instead of `SignedBeaconBlockAndBlobsSidecar`, not done in this PR, will do in separate PR
## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
```diff
@@ -158,11 +166,26 @@ public SafeFuture<Optional<SignedBeaconBlock>> getFinalizedBlockAtSlot(final UIn
     return SafeFuture.of(() -> database.getFinalizedBlockAtSlot(slot));
   }
 
+  @Override
+  public SafeFuture<Void> onNoBlobsSlot(final SlotAndBlockRoot slotAndBlockRoot) {
```
StefanBratanov, line 162: Wonder if this could be named better since technically It is noBlobsSlotAndBlockRoot, can we just say onNoBlobSidecars(...) and also storeNoBlobSidecars(...)

StefanBratanov (changes requested):
LGTM overall. But think It's better to rename back BlobSidecarManager to BlobsSidecarManager since it covers most of the change. Then we can do final cleanup later on. I would prefer to keep the important commits to master small.

storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
zilm13: I don't want something like NoBlobSidecars because we could have several BlobSidecars in block, so such name is less clear for me. 
Renaming back manager, ok, I will make separate rename PR for it.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -303,6 +304,25 @@ public List<SignedBeaconBlock> getNonCanonicalBlocksAtSlot(final UInt64 slot) {
...
+      final Collection<SignedBeaconBlock> blocks,
+      final Map<UInt64, List<BlobSidecar>> finalizedBlobSidecarsBySlot) {
+    try (final FinalizedUpdater updater = finalizedUpdater()) {
+      blocks.forEach(
+          block -> {
+            updater.addFinalizedBlock(block);
+            final List<BlobSidecar> blobSidecars =
+                finalizedBlobSidecarsBySlot.getOrDefault(block.getSlot(), Collections.emptyList());
+            if (blobSidecars.isEmpty()) {
+              updater.addNoBlobsSlot(new SlotAndBlockRoot(block.getSlot(), block.getRoot()));
```
StefanBratanov, line 319: I am not sure of the implications of this since this will start to be persisted pre-Deneb. Is there an overhead of doing this operation. Maybe could be an idea to make the map an optional pre-Deneb?

zilm13: should be fixed with https://github.com/ConsenSys/teku/pull/7028/commits/81d070b1662df0cd87b3da93813a5f610cd237b1

zilm13: not that simple
with slots without blocks

zilm13:
All is fixed now and should work correctly with 0 blobs blocks or even empty blocks inside the availability range

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
```diff
@@ -239,10 +241,28 @@ private SafeFuture<Boolean> requestBlocksAndBlobSidecarsByRange() {
         .thenApply(__ -> shouldRetryBlocksAndBlobSidecarsByRangeRequest());
   }
 
+  // We explicitly mark post-Deneb slots in availability range as some could contain no BlobSidecars
+  private void fillBlobSidecarBySlotToImport(final UInt64 startSlot, final UInt64 endSlot) {
```
StefanBratanov: Hmm that works yeah, but I wonder if logic can be simplified to avoid all these loops pre-Deneb. Shall we discuss on Monday?

zilm13: it's called only when we are inside if `    if (blobsSidecarManager.isAvailabilityRequiredAtSlot(endSlot)) {`

StefanBratanov: kk just rename to `fillBlobSidecarsBySloToImport`

StefanBratanov (approved):
LGTM

PR #7033, opened 2023-04-14
Double check g2 points are 96 bytes

jtraglia:
## PR Description

This is checked in `CKZG4844Utils.parseTrustedSetupFile` and in the C library, but I think it's a good idea to check this in the `TrustedSetup` class too, just in case. It would probably give a better error message here.

https://github.com/ConsenSys/teku/blob/c4bac6ca85b43b232d63f16054daae399a90368b/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java#L73-L78

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7034, opened 2023-04-15
Optimization of validator exit processing

### LINKED ISSUE #7041: Slow processing of blocks with multiple voluntary exit operations
### Description
We discovered this during our investigations post-Capella. The voluntary exit code path has never been in the critical path of block processing, now due to the growing number of validators exiting, more blocks with multiple exits exist.

This behaviour has been observed in our own nodes and multiple discord users have reported it as well.

e.g.
```
21:34:04.195 WARN  - Late Block Import *** Block: 17995607bcfee97f8f7ab71a0096d4272c60342a54fbb57352dbe84672ba5f4d (6216468) proposer 473561 arrival 2673ms, pre-state_retrieved +14ms, processed +2496ms, execution_payload_result_received +0ms, begin_importing +0ms, transaction_prepared +1ms, transaction_committed +0ms, completed +11ms
```

### Steps to Reproduce (Bug)
Mainnet issue.

**Expected behavior:** Stable block processing time, no late blocks warnings.

**Actual behavior:** Increased block processing time. Lots of late block import warnings.

Often. Hard to predict if the volume of exits will drop or not. In any case we need to fix this.

* 23.3.1 and previous versions

tbenr:
State transition voluntary exit timings:

With this PR:
```
 ** exits: 0 ms: 0
 ** exits: 16 ms: 97
 ** exits: 0 ms: 0
 ** exits: 16 ms: 71
 ** exits: 16 ms: 23
 ** exits: 16 ms: 24
```

Previous implementation:
```
 ** exits: 0 ms: 0
 ** exits: 16 ms: 1571
 ** exits: 0 ms: 0
 ** exits: 16 ms: 1399
 ** exits: 16 ms: 1442
 ** exits: 16 ms: 1325
```

fixes: #7041

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
```diff
@@ -103,6 +104,43 @@ public void initiateValidatorExit(MutableBeaconState state, int index) {
                     exitQueueEpoch.plus(specConfig.getMinValidatorWithdrawabilityDelay())));
   }
 
+  public Optional<UInt64> initiateValidatorExit(
+      MutableBeaconState state,
+      int index,
+      final UInt64 exitQueueEpoch,
+      List<Validator> exitedValidators) {
+    Validator validator = state.getValidators().get(index);
```
rolfyone: i would make this variable store state.getValidators, and should only have to call that once hopefully... can improve things a lot the less we call state.get functions...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
```diff
@@ -793,8 +795,29 @@ public void processVoluntaryExits(
...
+            return;
+          }
+
+          AtomicReference<UInt64> maxEpoch = new AtomicReference<>(UInt64.ZERO);
+
+          List<Validator> exitedValidators =
+              state.getValidators().stream()
+                  .filter(validator -> !validator.getExitEpoch().equals(FAR_FUTURE_EPOCH))
+                  .peek(validator -> maxEpoch.set(maxEpoch.get().max(validator.getExitEpoch())))
+                  .collect(toList());
```
rolfyone: i'll look at this more closely in the morning... I think I agree with the direction, just some of the details I think we need to be careful with, esp like here where we're passing lists constantly, that'll be a big memory impact potentially.... this list will get bigger for ever...

tbenr: During the we I thought about an additional improvement and a way to make it cleaner..

tbenr: Not sure we can avoid to collect that list. Take a look to my last version

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
tbenr: ah.. we can put Validators in the context now.

tbenr:
Btw there is also the slashing codepath going there.. maybe we should find something more global that covers that case too.. or initiate the exit context before the slashing loop too and pass it along the way...

tbenr:
actually i'm not super sure about putting BeaconState things in the context.. seems like a bit less clean...

tbenr:
I'm also thinking that this approach could be generalized for other operations (mutators) if thay could benefit in the same way: image at the beginning of a state transition we lazily create these contexts that are available for each state mutators. But the assumption that there are no interactions, a given state field can be changed by only one mutator so they do not interfere, so contexts contain completely different set of values.

Just an attempt to generalize the approach,if make sense.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
```diff
@@ -103,6 +107,83 @@ public void initiateValidatorExit(MutableBeaconState state, int index) {
...
+              UInt64.valueOf(
+                  exitedValidators.stream()
+                      .filter(v -> v.getExitEpoch().equals(validatorExitContext.exitQueueEpoch))
+                      .count());
+
+          return validatorExitContext;
+        });
+  }
+
+  public static class ValidatorExitContext {
```
rolfyone, line 174: no public getters so guessing this doesn't need to be public...

rolfyone:
I guess the real question would be if this improved numbers further? :)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
tbenr: the class must be known to `AbstractBlockProcessor`

tbenr:
A different approach to this would be to do something similar to `TransitionCaches` on the state.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
```diff
@@ -98,18 +110,85 @@ public void initiateValidatorExit(MutableBeaconState state, int index) {
...
+          final ValidatorExitContext validatorExitContext =
+              new ValidatorExitContext(beaconStateAccessors.getValidatorChurnLimit(state));
+
+          validatorExitContext.exitQueueEpoch = UInt64.ZERO;
+
+          final List<Validator> exitedValidatorsInExitQueueEpoch = new ArrayList<>();
+
+          for (Validator validator : state.getValidators()) {
+            final UInt64 validatorExitEpoch = validator.getExitEpoch();
+            if (validatorExitEpoch.equals(FAR_FUTURE_EPOCH)) {
```
zilm13: How about this? it could be clearer I think
```java
            if (validatorExitEpoch.equals(validatorExitContext.exitQueueEpoch)) {
              exitedValidatorsInExitQueueEpoch.add(validator);
            } else if (!validatorExitEpoch.equals(FAR_FUTURE_EPOCH)
                && validatorExitEpoch.isGreaterThan(
                    validatorExitContext.exitQueueEpoch)) { // New biggest exit epoch found
              validatorExitContext.exitQueueEpoch = validatorExitEpoch;
              exitedValidatorsInExitQueueEpoch.clear();
              exitedValidatorsInExitQueueEpoch.add(validator);
            }
```

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
```diff
@@ -98,18 +110,85 @@ public void initiateValidatorExit(MutableBeaconState state, int index) {
...
+  }
+
+  /**
+   * This function implements an optimized version of exitQueueEpoch and exitQueueChurn calculation,
+   * compared to the `initiateValidatorExit` reference implementation.
+   *
+   * @param state
+   * @return
+   */
+  public Supplier<ValidatorExitContext> createValidatorExitContextSupplier(
```
zilm13, line 125: I think it's a good candidate for some testing

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
```diff
@@ -98,18 +110,85 @@ public void initiateValidatorExit(MutableBeaconState state, int index) {
...
+      final BeaconState state) {
+    return Suppliers.memoize(
```
zilm13: I don't clearly understand for how long is it cached. If it runs only once when client starts, it could be a risk for some reasons. Say, in new fork we will change `churnLimit`. Instead I'd rewrite code to rerun it once in a X (slot? epoch should be good) but omitting such risks

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
```diff
@@ -98,18 +110,85 @@ public void initiateValidatorExit(MutableBeaconState state, int index) {
...
+          } else {
+            validatorExitContext.exitQueueChurn =
+                UInt64.valueOf(exitedValidatorsInExitQueueEpoch.size());
+          }
+
+          return validatorExitContext;
+        });
+  }
+
+  public static class ValidatorExitContext {
```
zilm13, line 174: I may sound little boring, but I don't like direct value access both for writing and reading here. Not sure this should be improved, just some eye catching for me

zilm13 (commented):
Looks very promising!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
tbenr: It can be reused only within the same state transition, so the slot (and `churnLimit`) is definitely not changing.

zilm13: ahh, got now, thank you for explanation.
A bit unclear anyway

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
tbenr: My brain considers that version less clear :)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
tbenr: well, reference tests are there to test state transitions.. we may add tests to prove equivalence with the old code.. 🤔

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
tbenr: I tend to be not super strict on that in cases like this where the fields are private, I see them like private fields in the outer class actually. Well, not exactly..

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
```diff
@@ -98,18 +110,91 @@ public void initiateValidatorExit(MutableBeaconState state, int index) {
...
+   * @return
+   */
+  public Supplier<ValidatorExitContext> createValidatorExitContextSupplier(
+      final BeaconState state) {
+    return Suppliers.memoize(() -> createValidatorExitContext(state));
+  }
+
+  /**
+   * This function implements an optimized version of exitQueueEpoch and exitQueueChurn calculation,
+   * compared to the `initiateValidatorExit` reference implementation.
```
rolfyone: I would probably use the python function name when talking about reference implementation...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
rolfyone: Yeh i think where the code is already at is more clear...

rolfyone (approved):
LGTM just one thing on one of the comments...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/BeaconStateMutators.java
rolfyone: arguably, 
```
    for (Validator validator : state.getValidators()) {
      final UInt64 validatorExitEpoch = validator.getExitEpoch();
      if (!validatorExitEpoch.equals(FAR_FUTURE_EPOCH)
          && validatorExitEpoch.isGreaterThanOrEqualTo(validatorExitContext.exitQueueEpoch)) {
        if (validatorExitEpoch.isGreaterThan(validatorExitContext.exitQueueEpoch)) {
          exitedValidatorsInExitQueueEpoch.clear();
          validatorExitContext.exitQueueEpoch = validatorExitEpoch;
        }
        exitedValidatorsInExitQueueEpoch.add(validator);
      }
    }
```
clearest i could come up with...

rolfyone: hopefully you like that because I committed it :)

tbenr: @rolfyone  Honestly, I don't! :) my brain still prefers my version: it follows the early return pattern, which has no nested ifs and clear separated conditions. Moreover this version is doing an additional comparison.

PR #7039, opened 2023-04-16
Unhide `--deposit-snapshot-enabled`

zilm13:
## PR Description
Unhides  `--deposit-snapshot-enabled` options. There are some questions with no clear decisions
- We had `--Xdeposit-snapshot-enabled` in testing by some users. As I removed it, Teku will fail to start until flag is updated. If I leave it, we have it in help along with flag without `X`, which I dislike.
- Should we make it default to true immediately? Why not? Why yes? I'm not sure

Doc draft. Will be submitted after `default` decision:

    Syntax: --deposit-snapshot-enabled[=<BOOLEAN>]
    Example: --deposit-snapshot-enabled=false
    Environment variable: TEKU_DEPOSIT_SNAPSHOT_ENABLED=false
    Configuration file: deposit-snapshot-enabled: false

Use bundled deposit contract tree snapshot and persist tree after finalization.

Currently at sync Teku requests all deposit logs from the execution layer up to the head. Next, at each startup Teku loads all deposits from the disk and replays them to recreate the merkle tree. Both operations consumes peer resources and delays node availability on restart. This feature dramatically decreases the time of both operations by bundling deposit tree snapshots in Teku distribution for all major networks including testnets (Mainnet, Gnosis, Goerli, Sepolia) and persisting current tree after finalization. Instead of replaying thousands of deposits on startup, Teku will load bundled tree or a saved one, whichever is the latest.

Security considerations: if a malicious agent changes bundled tree, Teku will throw `InvalidDepositEventsException` on the next deposit received from Execution Layer and will be unable to follow up the chain, so you will not be able to propose with incorrect deposit tree snapshot. Overall, it means network couldn't be attacked with malicious bundle, the consequences will be inability to propose like with any other serious bug in proposal flow.

## Fixed Issue(s)

Not yet or maybe...
## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/options/DepositOptions.java
```diff
@@ -60,10 +60,9 @@ public class DepositOptions {
   private String depositSnapshotPath;
 
   @Option(
-      names = {"--Xdeposit-snapshot-enabled"},
+      names = {"--deposit-snapshot-enabled"},
```
rolfyone, line 65: we need a better story around how we do these options i think.

I won't want this to be a breaking change for everyone, potentially we need a strategy...  Maybe what we do is accept the -X still, have it print a deprecation warning and then do the same functionality. This would mean that we have 2 flags until we remove the -X flag in a couple of releases.. .thoughts?

I guess my main concern is i'd prefer to have extra work than break user configuration...

rolfyone: If we do it this way though, it'll need to be listed as a breaking change... the alternative i suggested could be an 'upcoming breaking change'

zilm13: @rolfyone I'm on your side :joy: I dislike breaking too
I will try to do it in this way

zilm13: @rolfyone not super smart but clear in code and good for users
https://github.com/ConsenSys/teku/pull/7039/commits/3008c3a71a9e2b9849cd8cd47102b7da29bae1f0
I've not find any way to make it work on one PicoCli `Option` item

zilm13:
So any thoughts on making this default to true? We tested it for months internally

tbenr:
I'm ok with the default true

zilm13:
Updated default to `true` with changelog update

teku/src/test/java/tech/pegasys/teku/cli/options/DepositOptionsTest.java
```diff
@@ -129,9 +133,17 @@ public void shouldSetDefaultBundleSnapshotPathForSupportedNetwork(final String n
                 .toExternalForm());
   }
 
+  public static Stream<Arguments> getNetworksAndFlagsForDepositSnapshot() {
+    final Set<String> networks = Set.of("mainnet", "goerli", "prater", "gnosis", "sepolia");
+    final Set<String> flags =
+        Set.of("--deposit-snapshot-enabled", "--Xdeposit-snapshot-enabled"); // new and an old one
```
tbenr, line 139: I don't get the purpose of testing the flags like this if they are enabled by default

tbenr: Maybe we want to test that the --X continues having the precedence over the new one?

zilm13: we check that it results to the appropriate resource in configuration

tbenr: Ok got it

teku/src/main/java/tech/pegasys/teku/cli/options/DepositOptions.java
lucassaldanha: This approach sounds reasonable to me. And the --X option won't show up in the standard help option.

rolfyone: > I've not find any way to make it work on one PicoCli `Option` item

I was thinking 2 options to have the separation... and then we can just delete the -X in a future release

rolfyone (approved):
LGTM

PR #7056, opened 2023-04-18
Improve wording and variable names for update-snapshots script

zilm13:
## PR Description
When using we realized that ssh credentials like `user@host` are not actually an urls, so improving script wording and variable names
## Documentation

## Changelog

- [x I thought about adding a changelog entry, and added one if I deemed necessary.


tbenr (approved):
LGTM

PR #7057, opened 2023-04-18
Refactoring GetStateTest to use mocked chainDataProvider

### LINKED ISSUE #6869: investigate GetStateTest `shouldHandle_200`
When returning other states outside of phase0, the test is failing, and we need to understand if its a problem.

lucassaldanha:
## PR Description
Removing the dependency on a real ChainDataProvider makes this test quicker w/o losing its value.

## Fixed Issue(s)

fixes #6869

## Documentation

## Changelog


courtneyeh (approved):
LGTM

PR #7059, opened 2023-04-19
Allow local operations to republish on the network periodically

### LINKED ISSUE #6803: Top-up blocks voluntary exit?

rolfyone:
- If a local operation has been in the pool for 2 hours, socialise the operation again, assuming it's been lost from pools. Any operations using the mapped pool will start using this functionality.
 - reduce the BLS operation pool size to default (10k)

 Re-submission of local operations would still error as in the ticket, but I think that's correct as we can't re-add them. This will at least allow that stale operation to be re-published, which is the intent of re-running the command.

 fixes #6803

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/MappedOperationPool.java
```diff
@@ -89,6 +100,52 @@ public MappedOperationPool(
...
             "result");
+
+    this.timeProvider = timeProvider;
+    asyncRunner.runWithFixedDelay(
+        this::updateLocalSubmissions,
+        Duration.ofMinutes(10),
+        this::updateLocalSubmissionsErrorHandler);
+  }
+
+  private void updateLocalSubmissionsErrorHandler(Throwable throwable) {}
```
lucassaldanha: Not sure if the intent was to do nothing with the error in this handler. If that is the case I would suggest renaming it to make it clear that the intent is to do nothing (or maybe adding a `//DO NOTHING` comment in the body).

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/MappedOperationPool.java
```diff
@@ -89,6 +100,52 @@ public MappedOperationPool(
...
+          "Re-publishing {} operations that are still in the local {} operation pool",
+          staleLocalOperations.size(),
+          metricType);
+    }
+    int i = 0;
+    for (OperationPoolEntry<T> entry : staleLocalOperations) {
+      if (++i % 200 == 0) {
+        try {
+          // if we're processing large numbers, don't flood the network...
+          Thread.sleep(100);
```
lucassaldanha, line 134: Just confirming: this will sleep the thread from `operationPoolAsyncRunner`, right?

lucassaldanha (approved):
LGMT. A few nits and a question.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/MappedOperationPool.java
rolfyone: should really log it, may leave at debug initially

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/MappedOperationPool.java
rolfyone: yep will sleep the async runner, we have 2 threads.. .and it'll only happen if we get more than 200 entries...

tbenr: what about putting those constants in a `private static` var?

rolfyone: to what end? it's not really helping in any way...

PR #7061, opened 2023-04-19
Perform KZG commitments checks for blob transactions in reference tests

zilm13:
## PR Description
I'm not sure that we should link it with `bls_setting` but it sounds reasonable for me
## Documentation

## Changelog


tbenr (approved):
LGTM

PR #6148, opened 2022-09-01
Revert "Make it possible to disable BLS verification for everything except Deposits (#6116)"

ajsutton:
## PR Description
Revert "Make it possible to disable BLS verification for everything except Deposits (#6116)"

This reverts commit 35eb4759e2571c96feb397337c87dbc0cc4a41ce.

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

Issue #6147, opened 2022-09-01
Prater sync failure with invalid public key.


mcdee:
### Description
Teku stops syncing on goerli/prater at slot 3047065.  Truncated error logs are:

```
Caused by: tech.pegasys.teku.bls.impl.BlsException: Deserialization of public key bytes failed: 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
at tech.pegasys.teku.bls.impl.blst.BlstPublicKey.fromBytes(BlstPublicKey.java:43) ~[teku-infrastructure-bls-22.8.2.jar:22.8.2]
...
Caused by: java.lang.RuntimeException: BLST_ERROR: bad point encoding
at supranational.blst.blstJNI.new_P1_Affine__SWIG_1(Native Method) ~[jblst-0.3.8.jar:?]
...
2022-09-01 00:00:23.263 WARN  - Failed to process block
tech.pegasys.teku.bls.impl.BlsException: Failed to fastAggregateVerify
at tech.pegasys.teku.bls.BLS.fastAggregateVerify(BLS.java:203) ~[teku-infrastructure-bls-22.8.2.jar:22.8.2]
...
2022-09-01 00:00:23.264 WARN  - State transition error while importing block 3047065 (0xb1e4fad4c8b6042ecab18f6e0fbd815d8f3d69496837132dc2b69763323cd4e6)
```

Full log after a restart at [teku.log.gz](https://github.com/ConsenSys/teku/files/9468248/teku.log.gz)

### Steps to Reproduce (Bug)
This is reproducible on the given instance on restart, however I have not seen this on any other instance (either prater or other networks) so is not reproducible outside of this one instance (db available on request).

### Versions (Add all that apply)
Teku 22.8.2

asanso:
@ajsutton  @mcdee was this fixed or why was it closed ?

benjaminion:
I don't have full context any more, but the timing seems consistent with [this PR](https://github.com/ConsenSys/teku/pull/6148) - likely there was a bug that we quietly fixed via that reversion without advertising its existence. The others can comment if they can still remember.

ajsutton:
Yeah Ben's spot on. The BLS library changed from returning `false` when verification failed to throwing an exception which caused the issue.

PR #7065, opened 2023-04-21
Make it possible to disable BLS verification for everything except Deposits (attempt 2)

Nashatyrev:
## PR Description

Replace `AbstractBlockProcessor.blsVerifyDeposit` static flag with `BLSSignatureVerifier` reference to be able to set the verifier to the working instance even if the BLS verification is globally disabled with `BLSConstants.disableBLSVerification()`

This is the second attempt to enable this feature. 
The first one #6116 had an issue #6147 and had to be reverted: #6148

The problem was in inconsistent `BLS.verify*` methods behavior with respect to invalid `BLSPublicKey`: some methods returned `false`, some thrown `BLSException`. This PR also aligns these methods behavior: they are now all return `false`. Additionally `catch(BLSException)` clauses were added around `verify*` calls in `AbstractBlockProcessor` before (in revert PR #6148)

## Documentation

## Changelog


tbenr:
Today @rolfyone was actually working on the BLS verification area. You may need some coordination :)

rolfyone:
> Today @rolfyone was actually working on the BLS verification area. You may need some coordination :)

I'll keep digging and see if theres reasons I can't do what I'm thinking, but I'd really like to be avoiding using BlsSignatureVerifier.SIMPLE, or somehow making sure we minimise individual verify calls where possible...

It's just that in production we seem to be spending a significant amount of time doing individual verifies rather than batching... I'll drill into uses of the single verify function when i get a sec...

rolfyone (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
```diff
@@ -81,11 +81,16 @@
...
+  @VisibleForTesting
+  public static final BLSSignatureVerifier DEFAULT_DEPOSIT_SIGNATURE_VERIFIER =
+      BLSSignatureVerifier.SIMPLE;
   /**
    * For debug/test purposes only enables/disables {@link DepositData} BLS signature verification
    * Setting to <code>false</code> significantly speeds up state initialization
    */
-  public static boolean blsVerifyDeposit = true;
+  @VisibleForTesting
+  public static BLSSignatureVerifier depositSignatureVerifier = DEFAULT_DEPOSIT_SIGNATURE_VERIFIER;
```
rolfyone, line 93: are we planning on differentiating? why not just reference the default_deposit_signature_verifier?

Nashatyrev: `DEFAULT_DEPOSIT_SIGNATURE_VERIFIER` was intended as a default value for resetting the `depositSignatureVerifier` field after testing with disabled verifier. But you are right, it's probably an over-complication for the production code.

PR #7067, opened 2023-04-21
Send client information in User-Agent header when making builder bid request

### LINKED ISSUE #7030: Send teku and version in User-Agent header in builder client
In case of incompatibility issues, builder relayers can leverage user-agent header to identity specific client\versions that has problems and not release blocks to them, so client fallbacks to local EL.

We could use `VersionProvider.CLIENT_IDENTITY`/`VersionProvider.IMPLEMENTATION_VERSION` and avoid releasing too many info using `VersionProvider.VERSION`

User should be able to toggle this on/off.

StefanBratanov:
## PR Description

- Set User-Agent header to `VersionProvider.CLIENT_IDENTITY/VersionProvider.IMPLEMENTATION_VERSION` for builder bid requests
- By default okhttp seems to be sending User-Agent in the format `okhttp/4.10.0`. Wondering if we should remove this using an interceptor?

## Fixed Issue(s)
fixes #7030 

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
+++ b/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
@@ -26,6 +26,8 @@ import java.time.Duration;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Consumer;
+import java.util.regex.Pattern;
 import okhttp3.OkHttpClient;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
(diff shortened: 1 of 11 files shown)


ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
```diff
@@ -225,6 +236,31 @@ void registerValidators_failures() {
     verifyPostRequest("/eth/v1/builder/validators", signedValidatorRegistrationsRequest);
   }
 
+  @TestTemplate
```
lucassaldanha, line 239: Are we missing the test when `setUserAgentHeader = true`?

lucassaldanha (approved):
LGMT. Just a question re tests.

ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
StefanBratanov: By default it is true, so have added USER_AGENT_HEADER_ASSERTIONS to the other tests.

PR #7070, opened 2023-04-24
Use the blob sidecar pool when forward syncing

StefanBratanov:
## PR Description

- Introduce `void onBlobSidecarsFromSync(Bytes32 blockRoot, List<BlobSidecar> blobSidecars)` in BlobSidecarPool
- Modified `PeerSync` (single peer sync) and `BatchImporter` (multipeer sync) to use this method

## Fixed Issue(s)
N/A

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/PeerSync.java
```diff
@@ -195,19 +193,19 @@ private SafeFuture<PeerSyncResult> executeSync(
...
-              final SafeFuture<Void> blocksRequest =
-                  peer.requestBlocksByRange(
-                      requestContext.startSlot, requestContext.count, blockListener);
-
-              return SafeFuture.allOfFailFast(blocksRequest, blobSidecarsRequest)
+              return blobSidecarsRequest
+                  .thenCompose(
+                      __ ->
+                          peer.requestBlocksByRange(
+                              requestContext.startSlot, requestContext.count, blockListener))
```
zilm13, line 209: I think it's not a very safe way to pass whole mutable map along several consumers. Also I think the last consumer (importBlock) shouldn't get anything but its block only blobSidecars. Maybe it could be cleaner if we pass only related blobSidecars down from `blobSidecarsRequest`?

StefanBratanov: Thanks, I will have a double look to see how to improve it.

zilm13 (approved):
LGTM

PR #7071, opened 2023-04-24
Make maxConcurrentRequests in the blobs fetcher based on MAX_BLOBS_PER_BLOCK

StefanBratanov:
## PR Description
Value shouldn't be hardcoded to 3 as in the block fetcher, but should be calculated based on MAX_BLOBS_PER_BLOCK

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/gossip/blobs/FetchRecentBlobSidecarsService.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/gossip/blobs/FetchRecentBlobSidecarsService.java
@@ -60,7 +60,7 @@ public class FetchRecentBlobSidecarsService
       final FetchTaskFactory fetchTaskFactory,
       final Spec spec) {
     final int maxConcurrentRequests =
-        FetchRecentBlocksService.MAX_CONCURRENT_REQUESTS * spec.getMaxBlobsPerBlock().orElse(0);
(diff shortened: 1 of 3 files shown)


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/gossip/blobs/FetchRecentBlobSidecarsService.java
```diff
@@ -57,9 +57,12 @@ public static FetchRecentBlobSidecarsService create(
       final AsyncRunner asyncRunner,
       final BlobSidecarPool blobSidecarPool,
       final ForwardSync forwardSync,
-      final FetchTaskFactory fetchTaskFactory) {
+      final FetchTaskFactory fetchTaskFactory,
+      final Spec spec) {
+    final int maxConcurrentRequests =
+        FetchRecentBlocksService.MAX_CONCURRENT_REQUESTS * spec.getMaxBlobsPerBlock().orElse(0);
```
zilm13: I'm a bit paranoid on such `orElse`'s. Imagine we get here 0 by some bug and when and how we could realize it? I'd rather have `,orElse(1)`, so we still get `MAX_CONCURRENT_REQUESTS` or add some logging on service start to quickly catch a bug there if it is 0

StefanBratanov: 1 makes sense tbh, will fix

zilm13 (approved):
LGTM

PR #7083, opened 2023-04-26
Cleaned up blockValidator a little

### LINKED ISSUE #6893: Gossip validation improvements
1. when we validate signature in blocks (and blobSidecar) we add them to the validity set in this way:
```java
private boolean blockSignatureIsValidWithRespectToProposerIndex(
....

return signatureValid && receivedValidBlockInfoSet.add(new SlotAndProposer(block));
```
which cause the check to fail and triggers a `reject` like the signature was invalid

2. in blobSidecar gossip validation we haven't added the check we do on block, which is:
```java
    if (gossipValidationHelper.isBlockAvailable(block.getRoot())) {
      LOG.trace("Block is already imported");
      return completedFuture(InternalValidationResult.IGNORE);
    }
```
This should be always caught by `blockIsFirstBlockWithValidSignatureForSlot`. The only exception is when the limited cache doesn't contain it anymore because has been evicted (flooded). Should we remove it from block validation?

rolfyone:
- made some variables final more in line with our coding standards
 - cleaned up the `blockSignatureIsValidWithRespectToProposerIndex` fn so that it's not internally adding to the cache, as that's not really what the function describes that it's doing.

fixes #6893

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
```diff
@@ -135,16 +136,20 @@ public SafeFuture<InternalValidationResult> validate(final SignedBeaconBlock blo
                 return reject("Block signature is invalid");
               }
 
+              if (!receivedValidBlockInfoSet.add(new SlotAndProposer(block))) {
+                return ignore("Block has already been imported");
```
tbenr: here I'd say something similar to the first check: "Block is not the first with valid signature for its slot. It will be dropped"

tbenr (commented):
Looks good. I'd just change the message and apply the same change to `BlobSidecarValidator`

tbenr (approved):
LGTM

PR #7085, opened 2023-04-27
BlobSidecars pruning

zilm13:
## PR Description
BlobSidecar pruning + tests + some extra
- several BlobsSidecar traces removal
- fixes for deletedIds in db columns
- fixes for no blobs slots handling

`Store.retrieveBlobsSidecars...` was removed. We will add `Store.retrieveBlobSidecars...` when moving store flow from manager to transactionStore. I propose to add in the same PR replacements for tests: 
- `retrieveSignedBlockAndBlobsSidecar_shouldReturnEmptyIfBlockNotPresent`
- `retrieveSignedBlockAndBlobsSidecar_withInconsistentSidecar`
- `retrieveSignedBlockAndBlobsSidecar_withEmptyBlobs`
- `retrieveSignedBlockAndBlobsSidecar_withBlobs`

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -813,49 +793,23 @@ public void removeBlobSidecars(final UInt64 slot) {
...
-                  })
-              .count();
+      int remaining = pruneLimit;
+      int pruned = 0;
+      for (final Iterator<SlotAndBlockRootAndBlobIndex> it = prunableBlobKeys.iterator();
+          it.hasNext(); ) {
+        --remaining;
+        final boolean finished = remaining < 0;
+        final SlotAndBlockRootAndBlobIndex key = it.next();
+        if (finished && (key.isNoBlobsKey() || key.getBlobIndex().equals(ZERO))) {
```
tbenr, line 807: So, IIUC, we want to continue pruning blobs for the current slot so we don't partially delete blobs for a given slot.
Correct?

if yes it means that `pruned` can go over `pruneLimit` and the return statement should be

`return pruned >= pruneLimit`

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
```diff
@@ -407,325 +402,29 @@ public int dumpBlobsSidecar(
...
-      }
-    }
-
-    System.out.println(
-        "Wrote " + index + " unconfirmed blobs sidecars to " + outputFile.toAbsolutePath());
-    return 0;
-  }
-
-  @Command(
-      name = "validate-blobs-sidecars",
```
tbenr, line 451: Why you completely removed this command? It was useful to me to check that we have all blobs we expect to have

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
zilm13: Nice catch

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
zilm13: We'd better to start it from the blank, there are soo many changes in categories

storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlobSidecarPruner.java
```diff
@@ -47,52 +43,39 @@ public class BlobsSidecarPruner extends Service implements FinalizedCheckpointCh
...
     this.asyncRunner = asyncRunner;
     this.pruneInterval = pruneInterval;
     this.pruneLimit = pruneLimit;
     this.timeProvider = timeProvider;
-    this.blobsSidecarStorageCountersEnabled = blobsSidecarStorageCountersEnabled;
+    this.blobSidecarStorageCountersEnabled = blobSidecarStorageCountersEnabled;
 
-    if (blobsSidecarStorageCountersEnabled) {
+    if (blobSidecarStorageCountersEnabled) {
       LabelledGauge labelledGauge =
```
tbenr, line 72: We don't need a labelled gague anymore, we could go with `metricsSystem.createGauge` now.
OR we can add a label to track also the earliest slot so we can see the availability window moving.

tbenr (approved):
LGTM. Just a nit

storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlobSidecarPruner.java
tbenr: because now we have the total blob counter, which doesn't track directly the number of slots anymore

tbenr (approved):
LGTM

PR #7104, opened 2023-05-05
Activate storing of blobs on block import + tests

zilm13:
## PR Description
Actually store blobSidecars along with block in `ForkChoice`.
Discovered several bugs on the way.

As for `initialAnchor`: we don't need `earliestBlobSidecar` when initialized with weak subjectivity, because only header is there and we don't have blobSidecar for start slot. We will download blobSidecars for the next slot and set `earliestBlobSidecar` in the next slot if it's in availability window. When `initialAnchor` is genesis and is Deneb, we set `earliestBlobSidecar` to its slot.
I don't like hack I made on `AnchorPoint` loading but it looks safe.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/BlobsUtil.java
```diff
@@ -42,7 +42,7 @@ public class BlobsUtil {
   // actual list of versioned hashes representing the commitments
   private static final Bytes MAGIC_BLOB_TRANSACTION_PREFIX =
       Bytes.fromHexString(
-          "0x05450000000000000000000000000000000000000000000000000000000000000000"
+          "0x03450000000000000000000000000000000000000000000000000000000000000000"
```
tbenr: oh.. I spotted this too and already merged today..

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -590,7 +592,9 @@ public BuilderBid randomBuilderBid(final BLSPublicKey builderPublicKey) {
   }
 
   public BuilderBid randomBuilderBid(final Bytes32 withdrawalsRoot) {
-    return SchemaDefinitionsBellatrix.required(spec.getGenesisSchemaDefinitions())
+    return SchemaDefinitionsBellatrix.required(
+            spec.forMilestone(spec.getForkSchedule().getHighestSupportedMilestone())
```
tbenr: adding a `getSchemaDefinitionBellatrix` like the deneb one?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -464,6 +476,26 @@
...
   }
 
+  private Optional<UInt64> computeEarliestBlobSidecarsSlot(
+      final BlobSidecarsAndValidationResult blobSidecarsAndValidationResult,
+      final BeaconBlock block) {
+    if (!blobSidecarsAndValidationResult.isValid()) {
+      return Optional.empty();
+    }
+    // Earliest empty slot before this block or slot of this block if no empty (without blocks)
+    // slots before. If it's inside Deneb, it's `earliestBlobSidecarsSlot`
```
tbenr: This comment is driving me crazy. We need to describe why we need this code instead of writing the code in english :)

something like that, as comment for the entire method

\\ In order to keep track of DataAvailability Window, we need to compute the earliest slot we can consider data available for the given block. It needs to take in account possible empty slots so the actual slot should be parent's block slot + 1. Moreover, we need to manage the case in which the calculated slot is before Deneb activation, so it should never be before the first deneb slot.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -409,20 +412,29 @@
...
+    if (blobSidecarsAndValidationResult.isNotRequired()) {
+      // Outside availability window or pre-Deneb
+      blobSidecars = Collections.emptyList();
+    } else if (blobSidecarsAndValidationResult.isValid()) {
+      blobSidecars = blobSidecarsAndValidationResult.getBlobSidecars();
+    } else {
+      throw new IllegalStateException(
+          String.format(
+              "Unexpected attempt to store invalid blob sidecars (%s) for block: %s",
+              blobSidecarsAndValidationResult, block));
```
tbenr: I think codescan is right here. Doesn't make sense to print `blobSidecarsAndValidationResult`. Maybe number of blobs? For block I'd use `toLogString`

tbenr (commented):
some additional comments on top of the one in DM

tbenr:
this should close #7111

tbenr (approved):
LGTM

PR #7105, opened 2023-05-07
Added support for environment variables to ManualReferenceTestRunner

lucassaldanha:
## PR Description
Added support to filter reference tests using environment variables. This allow for more customization when running reference tests in CI pipelines and also makes devs lives easier creating specific IDE run config for specific use-cases.

## Fixed Issue(s)

N/A

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
@@ -15,6 +15,7 @@ package tech.pegasys.teku.reference;
 
 import com.google.errorprone.annotations.MustBeClosed;
 import java.io.IOException;
+import java.util.Optional;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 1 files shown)


eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
```diff
@@ -38,16 +39,30 @@ public class ManualReferenceTestRunner extends Eth2ReferenceTestCase {
...
+   */
   private static final String SPEC = "";
 
-  /** Filter test to run only those for a specific milestone. Use values from TestFork. */
-  private static final String MILESTONE = null;
+  /**
+   * Filter test to run only those for a specific milestone. Use values from TestFork.
+   *
+   * <p>May be overridden by the ENV_MILESTONE environment variable.
+   */
```
rolfyone, line 58: i like this, but the one other thing i'd add is the default to run all... so milestone would be null to run all, but some are empty string etc...

lucassaldanha: `MILESTONE = ""` will run all milestones by default 👍

rolfyone (approved):
LGTM

Issue #7047, opened 2023-04-17
update java minimum supported?


rolfyone:
With a dependency on besu, who now supports minimum java 17, we're going to have to probably follow suit, as we include besu libraries which we can no longer update, because we support java 11.

lucassaldanha:
Just pointing out some of the changes in Besu that could be similar to ours:
- https://github.com/hyperledger/besu/pull/3953
- https://github.com/hyperledger/besu/pull/4532
- https://github.com/hyperledger/besu/pull/4977

Don't be disheartened by the number of changes on https://github.com/hyperledger/besu/pull/4977 as most of them are related to PKI infrastructure support in Besu so it does not relate to Teku.

Here is the tracking ticket in Besu: https://github.com/hyperledger/besu/issues/3320

PR #7107, opened 2023-05-08
Upgrade to Java 17

siladu:
## PR Description

Upgrade source compatibility and CI to use Java 17

## Fixed Issue(s)

https://github.com/ConsenSys/teku/issues/7047

## Documentation

## Changelog


lucassaldanha:
We need to update our README as well:

https://github.com/ConsenSys/teku/blob/master/README.md#install-prerequisites

StefanBratanov:
Think it would be good to upgrade all the `besu` libraries to the latest ones as part of this PR as well

build.gradle
```diff
@@ -131,8 +131,8 @@ allprojects {
     jar.preserveFileTimestamps = false
```
StefanBratanov, line 130: We also need to remove jdk16 from `def dockerJdkVariants = [ "jdk17", "jdk16", ]`

lucassaldanha: Line 503 of build.gradle @siladu

siladu: done

siladu:
> Think it would be good to upgrade all the `besu` libraries to the latest ones as part of this PR as well

Might be worth waiting until just before merging this as likely to be more besu releases in between.

StefanBratanov:
@siladu when you are free, can you rebase this? The current CHANGELOG.md in master is fine, so no need to have changes there.

StefanBratanov:
> @siladu when you are free, can you rebase this? The current CHANGELOG.md in master is fine, so no need to have changes there.

Don't worry. Already rebased it.

StefanBratanov:
@siladu getting `Caused by: org.apache.logging.log4j.LoggingException: log4j-slf4j-impl cannot be present with log4j-to-slf4j` when updating the besu libraries. (you can see build) Any idea? I tried fixing it by excluding this dependency in the `evm` module, but it doesn't work.

StefanBratanov:
Upgrading **circleci/windows@2.4.1** to **circleci/windows@5.0.0** breaks 4 tests on windowsBuild consistently:
```
onDeneb_shouldStoreBlobSidecarsAlongWithBlock()
onDeneb_shouldStoreEarliestBlobSidecarSlotCorrectlyWhenThereIsGap()
onDeneb_shouldStoreEarliestBlobSidecarSlotCorrectlyWhenItsDenebGenesis()
retrieveBlobSidecars_shouldReturnBlobSidecarsWhenExists()
```

These tests actually fail on my Windows laptop as well consistently with same assertion error, so can investigate.

Upgrading besu libraries from **22.10.4** to **23.4.3** break a lot of tests with an exception: `org.apache.logging.log4j.LoggingException: log4j-slf4j-impl cannot be present with log4j-to-slf4j`

Probably should exclude some dependency?

zilm13:
I've pushed the fix https://github.com/Consensys/teku/pull/7107/commits/0889bebf4060c0e2a868dc23050bfd5ab8c435af
Detailed explanation is available in `log-captor` library https://github.com/Hakky54/log-captor#using-log-captor-alongside-with-other-logging-libraries
My little worry about it is that it could affect logging of some tests, but I ran few and didn't notice any changes.
As an alternative we could introduce new task/configuration dedicated to the tests with usage of the `log-captor`, it will look similar to those we have for integration or acceptance tests. So it will run as a separate task here in CI and require separated sourceSet, dedicated directories for these tests, dedicated dependencies set etc.

build.gradle
```diff
@@ -773,6 +773,12 @@ subprojects {
     acceptanceTestRuntimeOnly.extendsFrom testRuntimeOnly
     propertyTestRuntimeOnly.extendsFrom testRuntimeOnly
     referenceTestRuntimeOnly.extendsFrom testRuntimeOnly
+
+    // LogCaptor uses log4j-to-slf4j which couldn't be used along with log4j-slf4j-impl going to infinite loop
```
StefanBratanov: Instead of this comment, I may just add the link itself as a comment: https://github.com/Hakky54/log-captor#using-log-captor-alongside-with-other-logging-libraries

zilm13: Updated

StefanBratanov:
Wonder if we can enable `CapellaUpgradeAcceptanceTest` with upgraded besu libraries?

zilm13:
@StefanBratanov it's not related, we already had latest docker there. We need a feature to toggle Capella not on time condition like it was done for mainnet. As far as I know it's not implemented in Besu yet

StefanBratanov:
> @StefanBratanov it's not related, we already had latest docker there. We need a feature to toggle Capella not on time condition like it was done for mainnet. As far as I know it's not implemented in Besu yet

Yeah got it. All good. Not related yeah.

siladu:
> @siladu getting `Caused by: org.apache.logging.log4j.LoggingException: log4j-slf4j-impl cannot be present with log4j-to-slf4j` when updating the besu libraries. (you can see build) Any idea? I tried fixing it by excluding this dependency in the `evm` module, but it doesn't work.

Sorry @StefanBratanov I missed this. Thanks @zilm13 for the fix. Let me know if there's any other besu questions, maybe better to ping me on Slack.

zilm13:
@siladu all good, thank you for the help! Sorry, we've not touched it for more than a month, just started testing.

tbenr (approved):
LGTM

bgravenorst:
Removing the doc-change-required label because the docs already mention Java 17+ is required:
https://docs.teku.consensys.net/get-started/install/build-from-source

PR #7114, opened 2023-05-08
Extract process attestation proposer reward calculation

courtneyeh:
#6702


rolfyone (approved):
LGTM, let reftests run before merging pls...

PR #7121, opened 2023-05-09
Default optional SyncingStatus fields to `Optional.empty()`

### LINKED ISSUE #7100: Syncing status call fails when elOffline does not exist
Reported on Discord:  https://discordapp.com/channels/697535391594446898/697539289042649190/1103404864656113664

Was able to reproduce locally as well. 

When failovers are used in VC, there is a syncing status request call (`/eth/v1/node/syncing`) made in `BeaconNodeReadinessManager`.If the elOffline (optional field) is not returned in the response json, then Teku VC fails with:

```
2023-05-04 10:27:48.700+01:00 | validator-async-5 | DEBUG | BeaconNodeReadinessManager | http://127.0.0.1:5051/ is NOT ready to accept requests because the syncing status request failed: java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "java.util.Optional.orElse(Object)" because "this.elOffline" is null
```

StefanBratanov:
## PR Description

- Default `isOptimistic` and `elOffline` to Optional.empty() in `SyncingStatus` to avoid the optional being a null.

I decided to simplify this PR only to deal with the fixed issue.

A previous commit was reflecting the changes below:  https://github.com/ConsenSys/teku/pull/7121/commits/75e84f5edd561ce7121d44e19f314d696adcca8d but I thought more thought should be put in before making this changes, however happy to reconsider if others have a look.

Changes `DeserializableObjectTypeDefinition` to set fields who are defined as `withOptionalField` to Optional.empty() if the fields are missing in a json. Current behaviour was that they were null unless explicitly defined, which resulted in issues such as the `SyncingStatus` request.

## Fixed Issue(s)
fixes #7100 

## Documentation

## Changelog

### DIFF
--- a/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableObjectTypeDefinition.java
+++ b/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableObjectTypeDefinition.java
@@ -80,21 +80,8 @@ class DeserializableObjectTypeDefinition<TObject, TBuilder>
 
     final List<String> missingRequiredFields =
         deserializableFields.keySet().stream()
-            .filter(key -> !presentFields.contains(key))
             .filter(
-                key -> {
-                  final DeserializableFieldDefinition<TObject, TBuilder> objectField =
-                      deserializableFields.get(key);
-                  if (objectField.isRequired()) {
-                    return true;
-                  }
-                  // set missing fields to Optional.empty() instead of null
-                  if (objectField instanceof OptionalDeserializableFieldDefinition) {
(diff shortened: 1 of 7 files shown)


tbenr:
Not super sure, why you rolled back. If we want the framework to handle it correctly we shouldn't care about the specific case...

StefanBratanov:
> Not super sure, why you rolled back. If we want the framework to handle it correctly we shouldn't care about the specific case...

Wasn't sure about solution and there were couple of other places where it was initializing the optional fields to Optional.empty(). Can do a separate PR and can have discussion/review there. Think would be cleaner.

tbenr (approved):
LGTM

PR #7127, opened 2023-05-10
Set optional missing fields to empty when deserializing

StefanBratanov:
## PR Description
Currently if deserializing using `DeserializableTypeDefinition.object(...)` and there is a missing field in the json which is defined as `withOptionalField()` It won't be set to Optional.empty(). Instead the default would be used which could be null unless explicitly defined. It could lead to misuses of the framework and null optional fields which is not ideal.

`DeserializableObjectTypeDefinitionTest` has test cases, which cover optional fields so removed the initialization to Optional.empty() to test the change.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableObjectTypeDefinition.java
+++ b/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableObjectTypeDefinition.java
@@ -18,6 +18,7 @@ import static java.util.stream.Collectors.toSet;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -67,10 +68,10 @@ class DeserializableObjectTypeDefinition<TObject, TBuilder>
     for (; t == JsonToken.FIELD_NAME; t = p.nextToken()) {
       String fieldName = p.getCurrentName();
       p.nextToken();
-      final DeserializableFieldDefinition<TObject, TBuilder> objectField =
+      final DeserializableFieldDefinition<TObject, TBuilder> fieldDefinition =
           deserializableFields.get(fieldName);
-      if (objectField != null) {
-        objectField.readField(builder, p);
+      if (fieldDefinition != null) {
+        fieldDefinition.readField(builder, p);
         presentFields.add(fieldName);
       } else {
         LOG.debug("Unknown field: {}", fieldName);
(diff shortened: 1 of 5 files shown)


tbenr (commented):
I think this needs to be done, but I don't like the implementation. 'filter' with a side effect.
And the 'instanceof' with the cast. Any way to avoid that?

StefanBratanov:
> I think this needs to be done, but I don't like the implementation. 'filter' with a side effect. And the 'instanceof' with the cast. Any way to avoid that?

Yeah, I wasn't sure about the implementation as well, thus left it as a separate PR. Will think about it.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableObjectTypeDefinition.java
```diff
@@ -80,8 +80,21 @@ public TObject deserialize(final JsonParser p) throws IOException {
...
+                  if (objectField.isRequired()) {
+                    return true;
+                  }
+                  // set optional missing fields to Optional.empty()
+                  if (objectField instanceof OptionalDeserializableFieldDefinition) {
+                    ((OptionalDeserializableFieldDefinition<TObject, TBuilder, ?>) objectField)
+                        .setFieldToEmpty(builder);
+                  }
+                  return false;
+                })
```
rolfyone: for the sake of readability, can we do this explicitly and intentionally?
It works here, its just obfuscated a bit, and we'd be better off leaving the filter from the original and doing another pass through for all the fields that are optional that aren't present... same result just more easy to digest..
```
deserializableFields.keySet().stream()
     .filter(key -> !presentFields.contains(key) && !deserializableFields.get(key).isRequired())
    .forEach(key -> {
       final DeserializableFieldDefinition<TObject, TBuilder> objectField =
                      deserializableFields.get(key);
      if (objectField instanceof OptionalDeserializableFieldDefinition) {
          ((OptionalDeserializableFieldDefinition<TObject, TBuilder, ?>) objectField)
                        .setFieldToEmpty(builder);
      }
   })
```
 * guarantee thats not working currently, but for demonstration purposes :)
Then the old code dealing with missing required fields can be back to previous, it'll just make it easier to reason about.

rolfyone: might even be worth a private fn that `initializeMissingOptioalFields`, and maybe `reportMissingRequiredFields` for future us...

tbenr: maybe we can add a `Optional<OptionalDeserializableFieldDefinition> DeserializableFieldDefinition:: toOptionalDeserializableFieldDefinition` , like other `toVersion` for data structures. Maybe shorter names.. :)

StefanBratanov: Did a small refactor in latest commit, think it's more readable

tbenr (approved):
LGTM

PR #7128, opened 2023-05-10
Move away from no blobs placeholders to storing earliestBlobSidecar separately

zilm13:
## PR Description
TODO:
- ~deal with TODOs~
- ~add tests~

## Documentation

## Changelog


zilm13:
Ready for review
I'm a bit messed with all of our storage types: Store, RecentChainData, HotDao, FinalizedDao, Database, so be careful on review

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
```diff
@@ -74,8 +75,9 @@ public class HistoricalBatchFetcher {
   private final BlobSidecarManager blobSidecarManager;
   private final SafeFuture<BeaconBlockSummary> future = new SafeFuture<>();
   private final Deque<SignedBeaconBlock> blocksToImport = new ConcurrentLinkedDeque<>();
-  private final Map<UInt64, List<BlobSidecar>> blobSidecarsBySlotToImport =
+  private final Map<SlotAndBlockRoot, List<BlobSidecar>> blobSidecarsBySlotToImport =
       new ConcurrentHashMap<>();
+  private Optional<UInt64> maybeEarliestBlobSidecarSlot = Optional.empty();
```
tbenr, line 80: better to put it as `volatile`? can different threads access this non-final field?

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransaction.java
```diff
@@ -92,10 +93,15 @@ public void putBlockAndState(
...
-      final Optional<List<BlobSidecar>> maybeBlobSidecars) {
+      final List<BlobSidecar> blobSidecars,
+      final Optional<UInt64> maybeEarliestBlobSidecarSlot) {
     blockData.put(block.getRoot(), new TransactionBlockData(block, state, blockCheckpoints));
-    maybeBlobSidecars.ifPresent(
-        blobSidecars -> this.blobSidecars.put(block.getSlotAndBlockRoot(), blobSidecars));
+    if (!blobSidecars.isEmpty()) {
+      this.blobSidecars.put(block.getSlotAndBlockRoot(), blobSidecars);
+    }
+    if (maybeEarliestBlobSidecarTransactionSlot.isEmpty()) {
```
tbenr: shouldn't be `maybeEarliestBlobSidecarSlot` from method param?
If I am right, why no tests are failing?

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransaction.java
```diff
@@ -92,10 +93,15 @@ public void putBlockAndState(
...
+      maybeEarliestBlobSidecarTransactionSlot = maybeEarliestBlobSidecarSlot;
```
tbenr: `this.` just to be clear?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -164,6 +164,11 @@ public int getBlobSidecarsCount(final Optional<SignedBeaconBlock> signedBeaconBl
         .orElse(0);
   }
 
+  public UInt64 computeDenebStartSlot() {
```
tbenr: I think we should use a feature not a milestone in the name. Something like `computeFirstSlotWithBlobSupport`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -408,10 +409,20 @@ private BlockImportResult importBlockAndState(
     }
 
     final StoreTransaction transaction = recentChainData.startStoreTransaction();
+    final UInt64 earliestAffectedSlot =
```
tbenr, line 412: I'd definitely put a comment here to explain why we are doing this way. Or move the computation on a dedicated private method with a self-explanatory name

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -304,23 +304,25 @@ public List<SignedBeaconBlock> getNonCanonicalBlocksAtSlot(final UInt64 slot) {
...
-            }
+            finalizedBlobSidecarsBySlot
+                .get(block.getSlotAndBlockRoot())
+                .forEach(updater::addBlobSidecar);
           });
+      final Optional<UInt64> maybeEarliestBlobSidecarSlotDb = dao.getEarliestBlobSidecarSlot();
+      if (maybeEarliestBlobSidecarSlotDb.isEmpty() && maybeEarliestBlobSidecar.isPresent()) {
+        updater.setEarliestBlobSidecarSlot(maybeEarliestBlobSidecar.get());
+      }
       updater.commit();
```
tbenr: what about

```
maybeEarliestBlobSidecar.ifPresent(earliestBlobSidecar -> {
if(dao.getEarliestBlobSidecarSlot().isEmpty()) {
updater.setEarliestBlobSidecarSlot(earliestBlobSidecar);
}
});
```

so we query only if need?

tbenr (commented):
Some comments

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
zilm13: I've changed to the same idea but I like this code more
```
      if (maybeEarliestBlobSidecar.isPresent() && dao.getEarliestBlobSidecarSlot().isEmpty()) {
        updater.setEarliestBlobSidecarSlot(maybeEarliestBlobSidecar.get());
      }
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
zilm13: Added comment to `applyBlockToStore` method, do you think it could be enough?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
zilm13: I think if `alwaysRun` fragment is executed in the same time (and it's the only other usage) when we do import, we definitely have problems in the whole class. I'm not sure we need volatile here

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransaction.java
zilm13: No
We don't set it again if it's already set, that's an idea. 
We set it only once.
We could override it later only when pruning.
I think it's a correct logic as we don't want to move it to the right on every batch

zilm13: As an option, we could check if new value is smaller and change it only in this condition.
Also we could consider moving earliestBlobSidecar to the separate method but I dislike it as I want to force anyone saving blobSidecars to provide it. I think it's safer

tbenr (approved):
LGTM

PR #7132, opened 2023-05-10
Data unavailable block pool

tbenr:
Implements a "wait more and retry" mechanism for blocks that failed to import due to data unavailability.

This behaviour is only active when the node is in "in-sync" mode.

fixes: #7112

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #7134, opened 2023-05-11
Ensure DA pending blocks are not re-fetched

tbenr:
When we have blocks pending in `DataUnavailableBlockPool`, we don't want that blocks or attestations depending on it, being added to corresponding pendingPools, triggering the recentBlockFetcher, cause the node to redownload and attempt a block import.

builds on top of #7132

diff commit: 218ab4d25fa8220985f74a8d2b04ecce38e93311

## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #7142, opened 2023-05-13
Process attestations when the state from their target block state is available in cache

ajsutton:
## PR Description
When selecting a state to use to process attestation gossip, check if the attestations `beaconBlockRoot` state is already available in the cache before deciding if it's worth regenerating a state. This allows us to process significantly more attestations without requiring additional state generations - particularly if attestations are received for blocks we just imported on a long fork.

Also potentially reduces the number of generations we'd do during periods of non-finality by better leveraging what we have in cache.

## Documentation

## Changelog


rolfyone (approved):
LGTM! Nice to see you :)

PR #7149, opened 2023-05-16
Some optimizations for epoch transition

tbenr:
avoid some Bytes wrapping, especially during epoch transition, when `shuffleList` is called, which currently calls `intToBytes` 25 million times!

use thread local `sha256` where still not used.

old:
image

new:
image

## Documentation

## Changelog


rolfyone (approved):
LGTM, you've shown memory used, is it actually taking less time to process also?

tbenr:
A bit less, but hashing here is the big cpu consumption

tbenr:
A bit less, but hashing here is the big cpu.

I'll check it again cose it should be better to have a local sha256 here to avoid millions of lookups to thread local

tbenr:
actually there isn't much difference in cpu even switching back to local sha256.
but the overall `shuffleList` memory allocation is nicely reduced:

new
image

old
image

PR #7157, opened 2023-05-18
Use thread local on locally created sha256 

tbenr:
Just a tiny one to follow up with #7149 to reduce TL lookups when hashing in loops. We can still get the digest from the TL but wrap it into the sha256 class we can safely do hashings with.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #6992, opened 2023-03-27
improve messaging when adjusting upper and lower peer limits


rolfyone:
scenario, set upper-bound to 25, leave lower-bound default
log output:

```
2023-03-27 03:57:05.718 INFO  - Adjusting target number of peers lower bound to equal upper bound, which is 25
2023-03-27 03:57:05.721 INFO  - Adjusting target number of peers upper bound to equal lower bound, which is 64
```

potentially something more appropriate like:
```
WARN: Target number of peers upper bound cannot be set below the default lower bound.  Increasing target to 64.
```

VictoriaAde:
Hello @rolfyone, please I would like to be assigned this issue.
I have started working on it.

PR #7162, opened 2023-05-20
enhancement:improve messaging when adjusting upper and lower peer limits

l646505418:
## PR Description
improve messaging when adjusting upper and lower peer limits
## Fixed Issue(s)
https://github.com/ConsenSys/teku/issues/6992

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -383,7 +383,7 @@ public class StatusLogger {
 
   public void adjustingP2pUpperBoundToLowerBound(final int p2pLowerBound) {
     log.warn(
-        "Target number of peers upper bound cannot be set below the default lower bound.  Increasing target to {}.",
+        "Target number of peers upper bound cannot be set below the peers lower bound.  Increasing target to {}.",
         p2pLowerBound);
   }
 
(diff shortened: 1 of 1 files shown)


infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -382,8 +382,8 @@ public void adjustingP2pLowerBoundToUpperBound(final int p2pUpperBound) {
   }
 
   public void adjustingP2pUpperBoundToLowerBound(final int p2pLowerBound) {
-    log.info(
-        "Adjusting target number of peers upper bound to equal lower bound, which is {}",
+    log.warn(
+        "Target number of peers upper bound cannot be set below the default lower bound.  Increasing target to {}.",
```
rolfyone:
```suggestion
        "Target number of peers upper bound cannot be set below the peers lower bound.  Increasing target to {}.",
```

rolfyone: The very slight difference I'm suggesting is that it's going to increase to the lower bound, whether you've set one or it's the default, so if you've overridden the lower bound it'll use your suggested setting...

Because of this removing 'default' is probably less misleading :) hope that makes sense...

l646505418: I think it do make sense

rolfyone (approved):
LGTM thanks for the suggestion!

PR #7163, opened 2023-05-22
fix blob transaction prefix in EL stub

tbenr:
## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7164, opened 2023-05-22
Let tracker return independent completion futures

tbenr:
We want `getCompletionFuture` to return a new `SafeFuture` so that if we use `.orTimeout` on it doesn't complete exceptionally the internal completion future, which is design to track completion no matter the usage of the returned future will be.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7165, opened 2023-05-22
[Decoupling] Unblinding of SignedBlockContainer (part 3)

StefanBratanov:
## PR Description

- Implement `unblindSignedBlockIfBlinded` in `BlockFactoryDeneb` which deals with unblinding of the list of `BlindedBlobSidecar`
- Introduce `SignedBlobSidecarsUnblinder`
- Introduce `unblindSignedBlindedBlobSidecars` under spec in similar pattern as `unblindSignedBeaconBlock`

Next **final** part:
- Signing of blob sidecars in BlockProductionDuty

## Fixed Issue(s)
next **final** part will fix couple of tasks in https://github.com/ConsenSys/teku/issues/6822

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
```diff
@@ -394,24 +395,24 @@ private SafeFuture<Map<UInt64, List<Attestation>>> getAttestationsIncludedInEpoc
...
   @Override
-  public void saveProducedBlock(SignedBeaconBlock block) {
-    UInt64 epoch = spec.computeEpochAtSlot(block.getSlot());
-    Set<SlotAndBlockRoot> blocksInEpoch =
+  public void saveProducedBlock(final SignedBlockContainer blockContainer) {
+    final UInt64 epoch = spec.computeEpochAtSlot(blockContainer.getSlot());
+    final Set<SlotAndBlockRoot> blocksInEpoch =
         producedBlocksByEpoch.computeIfAbsent(epoch, __ -> concurrentSet());
-    blocksInEpoch.add(new SlotAndBlockRoot(block.getSlot(), block.getRoot()));
+    blocksInEpoch.add(new SlotAndBlockRoot(blockContainer.getSlot(), blockContainer.getRoot()));
```
tbenr: I added `getSlotAndBlockRoot()` deep in the block hierarchy. Might be good to expose it via container too. Or simply use `getSignedBlock().getSlotAndBlockRoot()` as you prefer

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/AbstractBlockFactory.java
```diff
@@ -27,14 +25,4 @@ protected AbstractBlockFactory(
     this.spec = spec;
     this.operationSelector = operationSelector;
   }
-
-  @Override
```
tbenr, line 31: Do we actually need the AbstractBlockFactory anymore?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/SignedBlobSidecarsUnblinderDeneb.java
```diff
@@ -0,0 +1,85 @@
...
+            blindedBlobSidecar.getBlockParentRoot(),
+            blindedBlobSidecar.getProposerIndex(),
+            blob,
+            blindedBlobSidecar.getKZGCommitment(),
+            blindedBlobSidecar.getKZGProof());
+    return signedBlobSidecarSchema.create(
+        unblindedBlobSidecar, signedBlindedBlobSidecar.getSignature());
+  }
+
+  private Blob findBlobWithRoot(final BlobsBundle blobsBundle, final Bytes32 blobRoot) {
```
tbenr: I'm not super convinced here. I know that this is safer but shouldn't we have bundle and blindedSidecars aligned? so we can lookup by index instead of searching a matching root? We could lookup by index and then check that roots corresponds. If index is out of bounds then it's a sign of misalignment

tbenr (commented):
some comments before checking it at runtime

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/AbstractBlockFactory.java
StefanBratanov: good point not needed

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
StefanBratanov: good idea

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/SignedBlobSidecarsUnblinderDeneb.java
StefanBratanov: yeah I like that approach more, fixed it

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
StefanBratanov: Btw tried that but there was a clash with MinimalBeaconSummary default method, which is the same, so removed it and just changed it to `blockContainer.getSignedBlock().getSlotAndBlockRoot()` in this class

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/SignedBlobSidecarsUnblinderDeneb.java
```diff
@@ -71,15 +70,22 @@ private SignedBlobSidecar unblindSignedBlindedBlobSidecar(
...
-        .orElseThrow(
-            () ->
-                new IllegalArgumentException(
-                    String.format(
-                        "Blob with root %s was not found in the BlobsBundle: %s",
-                        blobRoot, blobsBundle.toBriefString())));
+  private Blob findBlobWithRoot(
+      final BlobsBundle blobsBundle, final BlindedBlobSidecar blindedBlobSidecar) {
+    final int blindedBlobSidecarIndex = blindedBlobSidecar.getIndex().intValue();
+    if (blobsBundle.getNumberOfBlobs() < blindedBlobSidecarIndex + 1) {
```
tbenr: `blindedBlobSidecarIndex >= blobsBundle.getNumberOfBlobs()` ?

StefanBratanov: Makes more sense. I changed it to:
```
final UInt64 blindedBlobSidecarIndex = blindedBlobSidecar.getIndex();
if (blindedBlobSidecarIndex.isGreaterThanOrEqualTo(blobsBundle.getNumberOfBlobs())) 
```
to ensure there can't be any Integer overflow when calling `intValue()` on the index. Technically it should never happen but better to be safe.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/SignedBlobSidecarsUnblinderDeneb.java
```diff
@@ -0,0 +1,92 @@
...
+            blindedBlobSidecar.getBlockParentRoot(),
+            blindedBlobSidecar.getProposerIndex(),
+            blob,
+            blindedBlobSidecar.getKZGCommitment(),
+            blindedBlobSidecar.getKZGProof());
+    return signedBlobSidecarSchema.create(
+        unblindedBlobSidecar, signedBlindedBlobSidecar.getSignature());
+  }
+
+  private Blob findBlobWithRoot(
```
tbenr: this method name should be changed. `findBlobByIndex` or `findBlobByBlindedBlobSidecar`?

tbenr (approved):
LGTM. just a nit

PR #7169, opened 2023-05-23
[Decoupling] Signing BlockContainer (part 4)

StefanBratanov:
## PR Description

- Introduce `BlockContainerSigner` and have `Phase0` and `Deneb` implementations
- Add ability to sign blinded blob sidecars
- Cleanup of some unnecessary overwrites in `BeaconBlockBody` (essentially because they are same as the super class and Intellij was highlighting that)
- Remove some unused methods in sidecar schemas

## Fixed Issue(s)
will fix couple of tasks in #6822 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
@@ -50,27 +50,23 @@ public class SigningRootUtil {
   }
 
   public Bytes signingRootForBlobSidecar(final BlobSidecar blobSidecar, final ForkInfo forkInfo) {
-    final SpecVersion specVersion = spec.atSlot(blobSidecar.getSlot());
     final Bytes32 domain =
         spec.getDomain(
             Domain.DOMAIN_BLOB_SIDECAR,
             spec.computeEpochAtSlot(blobSidecar.getSlot()),
             forkInfo.getFork(),
             forkInfo.getGenesisValidatorsRoot());
-    specVersion.miscHelpers().computeSigningRoot(blobSidecar, domain);
     return spec.computeSigningRoot(blobSidecar, domain);
   }
 
   public Bytes signingRootForBlindedBlobSidecar(
       final BlindedBlobSidecar blindedBlobSidecar, final ForkInfo forkInfo) {
-    final SpecVersion specVersion = spec.atSlot(blindedBlobSidecar.getSlot());
     final Bytes32 domain =
         spec.getDomain(
             Domain.DOMAIN_BLOB_SIDECAR,
             spec.computeEpochAtSlot(blindedBlobSidecar.getSlot()),
             forkInfo.getFork(),
             forkInfo.getGenesisValidatorsRoot());
-    specVersion.miscHelpers().computeSigningRoot(blindedBlobSidecar, domain);
     return spec.computeSigningRoot(blindedBlobSidecar, domain);
   }
 
@@ -81,13 +77,11 @@ public class SigningRootUtil {
   }
 
   private Bytes32 getDomainForSignBlock(UInt64 slot, ForkInfo forkInfo) {
-    final Bytes32 domain =
-        spec.getDomain(
-            Domain.BEACON_PROPOSER,
-            spec.computeEpochAtSlot(slot),
-            forkInfo.getFork(),
-            forkInfo.getGenesisValidatorsRoot());
-    return domain;
+    return spec.getDomain(
+        Domain.BEACON_PROPOSER,
+        spec.computeEpochAtSlot(slot),
+        forkInfo.getFork(),
+        forkInfo.getGenesisValidatorsRoot());
   }
 
   public Bytes signingRootForSignAttestationData(
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
@@ -19,6 +19,7 @@ import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.SpecVersion;
 import tech.pegasys.teku.spec.constants.Domain;
+import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.BlindedBlobSidecar;
 import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.BlobSidecar;
 import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlock;
 import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlockHeader;
@@ -58,6 +59,19 @@ public class SigningRootUtil {
             forkInfo.getGenesisValidatorsRoot());
     specVersion.miscHelpers().computeSigningRoot(blobSidecar, domain);
     return spec.computeSigningRoot(blobSidecar, domain);
+  }
+
+  public Bytes signingRootForBlindedBlobSidecar(
+      final BlindedBlobSidecar blindedBlobSidecar, final ForkInfo forkInfo) {
+    final SpecVersion specVersion = spec.atSlot(blindedBlobSidecar.getSlot());
+    final Bytes32 domain =
+        spec.getDomain(
+            Domain.DOMAIN_BLOB_SIDECAR,
+            spec.computeEpochAtSlot(blindedBlobSidecar.getSlot()),
+            forkInfo.getFork(),
+            forkInfo.getGenesisValidatorsRoot());
+    specVersion.miscHelpers().computeSigningRoot(blindedBlobSidecar, domain);
+    return spec.computeSigningRoot(blindedBlobSidecar, domain);
   }
 
   public Bytes signingRootForSignBlockHeader(
--- a/ethereum/spec/src/test/java/tech/pegasys/teku/spec/signatures/LocalSignerTest.java
+++ b/ethereum/spec/src/test/java/tech/pegasys/teku/spec/signatures/LocalSignerTest.java
@@ -83,11 +83,15 @@ class LocalSignerTest {
     final UInt64 slot = UInt64.valueOf(10);
     final BeaconBlock block = dataStructureUtilDeneb.randomBlindedBeaconBlock(slot);
     final BlobSidecar blobSidecar =
-        dataStructureUtilDeneb.randomBlobSidecar(slot, block.getRoot(), UInt64.valueOf(2));
+        dataStructureUtilDeneb
+            .createRandomBlobSidecarBuilder()
+            .forBlock(block)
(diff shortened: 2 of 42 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
```diff
@@ -60,6 +61,19 @@ public Bytes signingRootForBlobSidecar(final BlobSidecar blobSidecar, final Fork
...
+  public Bytes signingRootForBlindedBlobSidecar(
+      final BlindedBlobSidecar blindedBlobSidecar, final ForkInfo forkInfo) {
+    final SpecVersion specVersion = spec.atSlot(blindedBlobSidecar.getSlot());
+    final Bytes32 domain =
+        spec.getDomain(
+            Domain.DOMAIN_BLOB_SIDECAR,
+            spec.computeEpochAtSlot(blindedBlobSidecar.getSlot()),
+            forkInfo.getFork(),
+            forkInfo.getGenesisValidatorsRoot());
+    specVersion.miscHelpers().computeSigningRoot(blindedBlobSidecar, domain);
```
zilm13: If I'm not wrong, this line is redundant. The one similar from method above too

ethereum/spec/src/test/java/tech/pegasys/teku/spec/signatures/LocalSignerTest.java
```diff
@@ -94,6 +95,28 @@ public void shouldSignBlobSidecar() {
...
 
+  @Test
+  public void shouldSignBlindedBlobSidecar() {
+    final UInt64 slot = UInt64.valueOf(10);
+    final BeaconBlock block = dataStructureUtilDeneb.randomBlindedBeaconBlock(slot);
+    final BlindedBlobSidecar blindedBlobSidecar =
+        dataStructureUtilDeneb
+            .createRandomBlobSidecarBuilder()
+            .slot(slot)
+            .blockRoot(block.getRoot())
```
zilm13: we could add `forBlock` method to builder which sets blockRoot, parent block root and slot

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/BlockContainerSignerDeneb.java
```diff
@@ -0,0 +1,137 @@
...
+        .map(
+            blindedBlockContainer -> {
+              final List<BlindedBlobSidecar> blindedBlobSidecars =
+                  blindedBlockContainer.getBlindedBlobSidecars().orElse(Collections.emptyList());
+              return signBlock(unsignedBlock, validator, forkInfo)
+                  .thenCombine(
+                      signBlindedBlobSidecars(blindedBlobSidecars, validator, forkInfo),
+                      this::createSignedBlindedBlockContents);
+            })
+        .orElseGet(
```
zilm13, line 61: I'd add comment for clarity that // when blinded is not available. Why it's not available btw? It could be useful info for comment too

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
```diff
@@ -101,47 +103,17 @@ public SafeFuture<BLSSignature> createRandaoReveal(final ForkInfo forkInfo) {
...
-    final BeaconBlock unsignedBlock =
-        maybeBlock
-            .map(BlockContainer::getBlock)
-            .orElseThrow(
-                () -> new IllegalStateException("Node was not syncing but could not create block"));
+  public SafeFuture<SignedBlockContainer> signBlockContainer(
+      final ForkInfo forkInfo, final Optional<BlockContainer> maybeBlockContainer) {
+    final BlockContainer unsignedBlockContainer =
+        maybeBlockContainer.orElseThrow(
+            () -> new IllegalStateException("Node was not syncing but could not create block"));
```
zilm13, line 110: I don't understand this comment. Maybe we could improve it if we are here?

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
```diff
@@ -70,19 +71,20 @@ public SafeFuture<DutyResult> performDuty() {
   public SafeFuture<DutyResult> produceBlock(final ForkInfo forkInfo) {
     return createRandaoReveal(forkInfo)
         .thenCompose(this::createUnsignedBlock)
```
zilm13, line 73: `createUnsignedBlockContainer`?

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
```diff
@@ -70,19 +71,20 @@ public SafeFuture<DutyResult> performDuty() {
...
+        .thenCompose(unsignedBlock -> signBlockContainer(forkInfo, unsignedBlock))
         .thenCompose(this::sendBlock)
         .exceptionally(error -> DutyResult.forError(validator.getPublicKey(), error));
   }
 
-  private SafeFuture<DutyResult> sendBlock(final SignedBeaconBlock signedBlock) {
+  private SafeFuture<DutyResult> sendBlock(final SignedBlockContainer signedBlockContainer) {
     return validatorApiChannel
-        .sendSignedBlock(signedBlock)
+        .sendSignedBlock(signedBlockContainer)
```
zilm13, line 81: `sendSignedBlockContainer`?

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/BlockProductionDutyTest.java
```diff
@@ -51,11 +67,15 @@
 import tech.pegasys.teku.validator.api.ValidatorApiChannel;
 import tech.pegasys.teku.validator.client.ForkProvider;
 import tech.pegasys.teku.validator.client.Validator;
+import tech.pegasys.teku.validator.client.signer.BlockContainerSigner;
+import tech.pegasys.teku.validator.client.signer.MilestoneBasedBlockContainerSigner;
 
 class BlockProductionDutyTest {
   private static final String TYPE = "block";
   private static final UInt64 SLOT = UInt64.valueOf(498294);
```
zilm13: Maybe CAPELLA_SLOT for clarity? or just capellaSlot, we are not very strict on case here usually. It's not important that it's a Capella, but we should clarify that it's pre-Deneb and we are not actually testing it on other forks

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/BlockProductionDutyTest.java
```diff
@@ -108,8 +138,178 @@ public void shouldCreateAndPublishBlock(final boolean isBlindedBlocksEnabled) {
...
+            denebSlot,
+            forkProvider,
+            validatorApiChannel,
+            blockContainerSigner,
+            false,
+            spec);
+
+    final BLSSignature randaoReveal = dataStructureUtil.randomSignature();
+    final BLSSignature blockSignature = dataStructureUtil.randomSignature();
+    final BlockContents unsignedBlockContents = dataStructureUtil.randomBlockContents(denebSlot);
```
zilm13, line 171: Could you please add comment to `randomBlockContents` first line, why we have some hardcoded number here, so noone will later remove it from there or change to random.

zilm13:
Damn IDEA, lost review comment. Everything looks good, just a few nits above

ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/SigningRootUtil.java
StefanBratanov: good catch, I just copied the method without having a more thorough look

ethereum/spec/src/test/java/tech/pegasys/teku/spec/signatures/LocalSignerTest.java
StefanBratanov: Like it, makes sense

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
StefanBratanov: I am not 100% sure about changing the method name. Technically the beacon api is still `/blocks/` so I classify `BlockContainer` more as an internal Teku thing and would like the methods name to be closer to the spec. Do you have any opinion @tbenr @rolfyone ?

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
StefanBratanov: Same comment as above.

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
StefanBratanov: Hmm technically this happens either if we can't deserialize the response from the BN endpoint from VC or we don't have the state for a slot in the `Store`. But then again different BNs handle this differently. In any case, this is a serious issue, which shouldn't happen, but unfortunately we don't have information of exact cause. Not sure how it can be expressed better.

zilm13: What are the relations with syncing? There are a lot of focus on it in the message.
Maybe describe what we really have: Block container from BeaconNode is empty. BeaconNode provided incorrect unsigned block container. Something around it

zilm13 (approved):
LGTM
I'd just wait for thoughts on `createUnsignedBlockContainer` and `sendSignedBlockContainer` from guys and also consider improving exception message

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
tbenr: the container is just a box, what we want to do here is to create a block and what is needed around it.. so I'd remain on current naming.

StefanBratanov: cool, I will merge then if you are both happy?

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
tbenr: if the node is syncing it return an HTTP status with error and will be handled differently. In this case you received an empty response. I think we can leave it, has been there since beacon launch :)

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/BlockProductionDuty.java
tbenr: GO

PR #7171, opened 2023-05-24
Sync committee cleanup

rolfyone:
- Improved trace logging

 - resolved the spelling of 'Validatable' in a couple of classes.

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/synccommittee/SyncCommitteeContributionPool.java
```diff
@@ -77,6 +87,13 @@ private SafeFuture<InternalValidationResult> add(
                     subscriber ->
                         subscriber.onOperationAdded(
                             signedContributionAndProof, result, fromNetwork));
+              } else if (result.isUnknownBlock()) {
+                LOG.trace(
+                    "Deferring contribution and proof {} for slot {} as required block {} is not yet present",
+                    signedContributionAndProof.hashTreeRoot(),
```
tbenr: Not so important but, not sure if at this stage we ever calculated the root of this object. I feel more comfortable by using lambda params

tbenr: the pending pool will trigger it but not for ignored messages.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobSidecarManagerImpl.java
```diff
@@ -93,6 +93,7 @@ public SafeFuture<InternalValidationResult> validateAndPrepareForBlockImport(
               invalidBlobSidecarRoots.put(
                   signedBlobSidecar.getBlobSidecar().hashTreeRoot(), result);
               break;
+            case UNKNOWN_BLOCK: // TODO is this ok?
```
tbenr: it is not relevant but might be misleading. The blob validation never returns UNKNOWN_BLOCK. So maybe moving it as last case with a comment might be better

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/PoolFactory.java
```diff
@@ -32,6 +35,9 @@ public class PoolFactory {
   private static final UInt64 DEFAULT_HISTORICAL_SLOT_TOLERANCE = UInt64.valueOf(320);
   private static final int DEFAULT_MAX_ITEMS = 5000;
   private static final int DEFAULT_MAX_BLOCKS = 5000;
+
+  private static final int DEFAULT_MAX_SYNC_COMMITTEE_MESSAGES = 512;
```
tbenr: should we reference VALID_SYNC_COMMITTEE_MESSAGE_SET_SIZE constant instead of heaving another constant?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/PoolFactory.java
```diff
@@ -32,6 +35,9 @@ public class PoolFactory {
...
+  private static final int DEFAULT_MAX_CONTRIBUTION_AND_PROOFS = 256;
```
tbenr: could be a function of the corresponding value in `Constants`?

tbenr (approved):
LGTM with some minor comments.

Additionally, we could inject `invalidBlockRoots` in the validators to distinguish from a simple IGNORE to a REJECT in case the voted block turns out to be a known invalid block.

rolfyone:
This is less ambitious now, testing showed that the unknown block processing was firing a lot more than I thought it should, so have removed that aspect. The log messages weren't great as they gave no context, so I added context.

PR #7172, opened 2023-05-24
Test fixtures updated to use actual schema definitions instead of gen…

zilm13:
…esis

## PR Description
We have a number of usages of genesis schema definitions in test fixtures which could fail in complex test cases where several forks are changed during the test. This update fixes this
## Documentation

## Changelog


tbenr (approved):
LGTM

PR #7173, opened 2023-05-24
Clarify Java 17 upcoming breaking change

StefanBratanov:
## PR Description
related to #7107 

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Upcoming Breaking Changes
 
-- Upgrading source code to Java 17 meaning users will need to upgrade their Java install to at least 17, or use the jdk17 variant of the docker image.
+- Upgrading source code to Java 17 which means users will need to upgrade their Java installation to at least 17. Docker users will not be affected, apart from the change that `jdk16` images will no longer be available for new versions.
 
 ## Current Releases
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Upcoming Breaking Changes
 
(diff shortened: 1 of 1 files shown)


zilm13:
Is it a reason jdk16 images should be specially covered?

StefanBratanov:
> Is it a reason jdk16 images should be specially covered?

Don't know, it's just something that was available that won't be anymore. Doubt there are many users using it but would be worth to mention it.

zilm13:
Ahh I got it, if you use something like `develop-jdk16` and pull it regularly to update, it will be no more supported. We could change it to smth like "that `jdk16` images will no longer be available, use `jdk17` instead", I like suggestions more than prohibitions :)

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Upcoming Breaking Changes
 
-- Upgrading source code to Java 17 meaning users will need to upgrade their Java install to at least 17, or use the jdk17 variant of the docker image.
+- Upgrading source code to Java 17 which means users will need to upgrade their Java installation to at least 17. Docker users will not be affected, apart from the change that `jdk16` images will no longer be available for new Teku versions.
```
rolfyone:
```suggestion
- Upgrading the minimum Java version to Java 17, which means users will need to upgrade their Java installation to at least `Java 17`.  The docker versions relying on `jdk16` will no longer be published, so docker users explicitly referencing the `jdk16` build need to update their package to reference `jdk17`, as tags `develop-jdk16`, `develop-jdk16-arm64`, `latest-jdk16` will no longer be able to be updated. 
```

StefanBratanov: Thanks that's good. Done.

rolfyone (approved):
LGTM

PR #7177, opened 2023-05-25
Set mainnet trusted setup as default for unknown networks

tbenr:
For convenience we can always set mainnet trusted setup as default

This fixes my initial problem which is `genesis` command is not working when creating a Deneb genesis, because it is not able to get trustedSetup from parameter but then network config requires it.

## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #7178, opened 2023-05-25
Updating reference tests to 1.4.0.alpha [Part 2, completed]

### LINKED ISSUE #7220: Remove `is_merge_transition_block` and `is_execution_enabled` since Capella
https://github.com/ethereum/consensus-specs/pull/3232
https://github.com/ethereum/consensus-specs/pull/3350

zilm13:
## PR Description
WIP
Fixing new execution payload reference tests
Includes following updates/fixes:
- `mergeTransitionComplete` should return true w/o checks for all forks after Bellatrix (Capella etc) + MergeTransitionBlockValidator should skip validation (I think) when it's post-Bellatrix
- 1 test disabled until resolution of the issue https://github.com/ethereum/consensus-specs/issues/3390 .UPDATE: enable considering it's always execution on Capella
- #7220 

## Fixed Issue(s)

Fixes #7220 
## Documentation

## Changelog


zilm13:
kzg tests are broken with update, investigating

zilm13:
`CapellaUpgradeAcceptanceTest` runs ok locally but I worry that it could be flaky on CI until Besu gets `shanghaiBlock`, so I disabled it. Time-based it has different behavior even on laptop and desktop.

zilm13:
Also I think we could close #7015 with merging this one

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/MergeTransitionBlockValidator.java
```diff
@@ -139,6 +140,9 @@ private SafeFuture<PayloadValidationResult> retrieveAndVerifyNonFinalizedTransit
 
   private SafeFuture<PayloadStatus> verifyTransitionPayload(
       final UInt64 slot, final ExecutionPayloadSummary executionPayload) {
+    if (spec.atSlot(slot).getMilestone().isGreaterThanOrEqualTo(SpecMilestone.CAPELLA)) {
```
StefanBratanov, line 143: Problem i see is `verifyTransitionPayload` is called in `verifyFinalizedTransitionBlock` where 
```
final Optional<SlotAndExecutionPayloadSummary> finalizedOptimisticTransitionPayload =
        recentChainData.getStore().getFinalizedOptimisticTransitionPayload();
```

The slot retrieved would always be bellatrix one for mainnet and we would still end up calling `validateMergeBlock()` so spec wouldn't be met.

zilm13: I'm not sure if it was removed in Capella. Actually I'm a bit confused to understand what part of spec it resembles. The shortcut was made for Devnets with genesis starting from Capella or later forks, not targeting Mainnet

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/ForkChoiceUtil.java
```diff
@@ -547,9 +547,17 @@ private boolean isBellatrixBlockOld(final ReadOnlyStore store, final UInt64 bloc
         .isLessThanOrEqualTo(getCurrentSlot(store));
   }
 
-  private boolean isExecutionBlock(final ReadOnlyStore store, final Bytes32 blockRoot) {
+  private boolean isExecutionBlock(
+      final ReadOnlyStore store, final UInt64 blockSlot, final Bytes32 parentBlockRoot) {
```
tbenr: would be simpler to test block body version instead of slot?
`block.getBeaconBlock().map(BeaconBlock::getBody).map(BeaconBlockBody::toVersionCapella)`

StefanBratanov: Hmm actually not sure, can it be blinded block?

zilm13: yes it could be both blinded and unblined

tbenr:
I wonder if we just override `isExecutionEnabled` too in `miscHelperCapella`, we can avoid having a capella version of `processBlock`. So the solution will only be in the mischelper. On the other hand, heaving a cleaner `processBlock` seems nice too. Leveraging inheritance seems not wrong, even if introduces duplication. But is quite safe since that codepath is havely stressed by reference tests.

tbenr:
I'd improve the miscHelpers api with this: https://github.com/tbenr/teku/commit/be4f4854e7e12196cbc526895a7a63385fca70a2

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/ForkChoiceUtil.java
tbenr:
```java
    BeaconBlockBody body = block.getMessage().getBody();
    if(body.toVersionCapella().isPresent() || body.toBlindedVersionCapella().isPresent()) {
      return true;
    }
```
?

StefanBratanov: looks good

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/TerminalPowBlockMonitor.java
```diff
@@ -278,17 +278,20 @@ private void checkTerminalBlockByTTD() {
...
+    // if it's genesis validate fast way
+    if (terminalBlock.getParentHash().isZero()) {
+      return SafeFuture.completedFuture(true);
+    }
     return executionLayer
         .eth1GetPowBlock(terminalBlock.getParentHash())
         .thenApply(
             powBlock -> {
-              UInt256 totalDifficulty =
+              UInt256 parentDifficulty =
```
tbenr: `parentTotalDifficulty` :-)

tbenr (approved):
LGTM with a nit

PR #7186, opened 2023-05-29
Added test cases for RewardsCalculator

rolfyone:
Can now build blocks with the test BeaconBlockBuilder with random data that might be needed for testing things like attester slashings and sync aggregates, which improves the tests we can write for unit testing.

## Documentation

## Changelog


rolfyone:
now that the problem I was looking for is identified, will close this and merge with original PR.

PR #7193, opened 2023-05-30
Fix attestation performance log message issue

### LINKED ISSUE #7175: attestation performance log message issue
```
Attestation performance: epoch 202574, expected 1, produced 1, included 0 (0%), distance 2147483647 / 0,00 / -2147483648, correct target 1 (100%), correct head 1 (100%)
```

from code in `DefaultPerformanceTracker`:
```
    // IntSummaryStatistics returns Integer.MIN and MAX when the summarized integer list
    // is empty.
```
our included is 0, but we're checking produced != 0 after the performance details are gathered...

we probably need to be a little careful, we've produced an attestation but it wasn't included and we want a sensible log message for that.

courtneyeh:
fixes #7175 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7200, opened 2023-05-30
Rpc blobs bug fix + blobsUtil endianness + integration tests

### LINKED ISSUE #6834: Add BlobSidecarsByRangeIntegrationTest
Add `BlobSidecarsByRangeIntegrationTest` when the syncing and the storage are done

### LINKED ISSUE #7183: BlobSidecarsByRange does not work
Running interop, testing forward sync I realize blobs by range RPC doesn't work.

- the request is sent from the client.
- it is received from the server and the server prepares the response.
- As soon as the server start streaming the result to client, client closes the connection falling in this condition:
in `Eth2OutgoingRequestHandler#processData`
image

- Server early stops the streaming this way:
image

### LINKED ISSUE #7216: `BlobSidecarByRoot` RPC should not fail if block root is not found.
```
Requests sidecars by block root and index.
The response is a list of BlobSidecar whose length is less than or equal to the number of requests.
It may be less in the case that the responding peer is missing blocks or sidecars.
```

tbenr:
fixes: #7216
fixes: #7183
fixes: #6834

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
@@ -65,6 +65,7 @@ import tech.pegasys.teku.spec.datastructures.state.Checkpoint;
 class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
   private static final Logger LOG = LogManager.getLogger();
 
+  private final Spec spec;
   private final BeaconChainMethods rpcMethods;
   private final StatusMessageFactory statusMessageFactory;
   private final MetadataMessagesFactory metadataMessagesFactory;
@@ -92,6 +93,7 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
       final RateTracker blobSidecarsRequestTracker,
       final RateTracker requestTracker) {
     super(peer);
+    this.spec = spec;
     this.rpcMethods = rpcMethods;
     this.statusMessageFactory = statusMessageFactory;
     this.metadataMessagesFactory = metadataMessagesFactory;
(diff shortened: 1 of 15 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java
```diff
@@ -164,25 +156,29 @@ private UInt64 getFinalizedEpoch() {
    *   <li>The block root references a block greater than or equal to the minimum_request_epoch
```
StefanBratanov, line 155: I would just remove the `<li>A block for the block root is available` above

StefanBratanov (approved):
LGTM with one javadoc nit

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
```diff
@@ -290,9 +293,19 @@ public SafeFuture<Void> requestBlobSidecarsByRange(
                 if (updatedCount.isZero()) {
                   return SafeFuture.COMPLETE;
                 }
-                request = new BlobSidecarsByRangeRequestMessage(firstSupportedSlot, updatedCount);
+
+                request =
+                    new BlobSidecarsByRangeRequestMessage(
+                        firstSupportedSlot,
+                        updatedCount,
+                        spec.getMaxBlobsPerBlock(firstSupportedSlot.plus(updatedCount))
```
StefanBratanov: should be technically` .plus(updatedCount).minus(1)`

StefanBratanov (commented):
LGTM with one javadoc nit

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
tbenr: right.. good point

tbenr: put it on the other place too

PR #7202, opened 2023-05-31
GetAttestationRewards integration tests

lucassaldanha:
## PR Description

- Added integration tests for GetAttestationRewards
- Fixed isExecutionOptimistic and isFinalized flags

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
@@ -40,6 +40,7 @@ import tech.pegasys.teku.api.exceptions.ServiceUnavailableException;
 import tech.pegasys.teku.api.migrated.AttestationRewardsData;
 import tech.pegasys.teku.api.migrated.BlockHeadersResponse;
 import tech.pegasys.teku.api.migrated.BlockRewardData;
+import tech.pegasys.teku.api.migrated.GetAttestationRewardsResponse;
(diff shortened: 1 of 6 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -623,6 +663,20 @@ private UInt64 findSlotAtEndOfNextEpoch(final UInt64 epoch) {
...
   }
 
+  private Optional<UInt64> findLatestAvailableEpochForRewardCalculation() {
+    final int minNumOfSlots = 2;
+    return recentChainData
+        .getCurrentEpoch()
+        .flatMap(
+            epoch -> {
+              if (epoch.isGreaterThanOrEqualTo(minNumOfSlots)) {
+                return Optional.of(epoch.minus(minNumOfSlots));
```
rolfyone: oh so we have to be 2 epochs behind the head epoch?

also `minNumberOfSlots` is not what this is if we're decrementing epoch by it...

rolfyone (approved):
LGTM generally with the caveat that a variable name is confusing...

data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
lucassaldanha: Yeah, I messed up the name...My brain isn't keeping up with rewards API anymore...

Issue #7179, opened 2023-05-25
Move p2p networking setting from Constants to SpecConfig


lucassaldanha:
Spec change: https://github.com/ethereum/consensus-specs/pull/3375

P2P options will depend on chain configuration (e.g. Mainnet can have different values than Goerli).

zilm13:
Resolved by #7250

PR #7209, opened 2023-05-31
Moved gossip_max_size into out of constants into spec.

rolfyone:
All of the move constants are in the mainnet.yaml and minimal.yaml, just that they are all disabled except for gossip_max_size. This will be the 'todo' list of #7179.

partially addresses #7179

## Changelog


lucassaldanha (approved):
LGTM

PR #7211, opened 2023-06-01
Remove ignore from trivyignore file

### LINKED ISSUE #6978: Update dependencies using snakeyaml < 2.0
We use snakeyaml transitively through web3j, swagger-core and jackson-dataformat-yaml.

They have release snakeyaml 2.0 that addresses the vulnerability, however, this would require us to exclude the transitive dependency and rely on providing our own snakeyaml in the classpath.

There are a few breaking changes in snakeyaml 2.0 but nothing that directly impacts us that I could find. But it is hard to know how it could impact the libraries using it.

From what I can see, only jackson is working on an update 2.15.0-rc to use snakeyaml 2.0.

We should update our dependencies for versions using snakeyaml 2.0 and consider ignoring the transitive dependency for the dependencies that don't update.

For now, we are ignoring the alert as it only applies from parsing yaml from untrusted sources (#6977). **Once this is update we should stop ignoring it [here](https://github.com/ConsenSys/teku/pull/6976/files#diff-3eb22c00f55476e288641d2f35cfc384ecf77a765c18eb15e7c57e731af05fc5R2) .**

Example: https://github.com/ConsenSys/teku/commit/2fc0c0c370952123362b6e164b628c9c7647cfb4

lucassaldanha:
## PR Description

Removing unnecessary CVE from Trivy's ignore file.

## Fixed Issue(s)
fixes #6978 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7212, opened 2023-06-01
Fixed the request body of validator liveness endpoint

### LINKED ISSUE #7204: `liveness` endpoint not conform to the specification?
The Beacon API `v2.4.0` defines the [getLiveness endpoint](https://ethereum.github.io/beacon-APIs/#/Validator/getLiveness) as following:

- **Verb:** POST
- **Endpoint:**  `/eth/v1/validator/liveness/{epoch}`
- **Body example:** 
```
["1", "42", "1664"]
```
(List of indexes)

However, the Teku implements the [getLiveness endpoint](https://docs.teku.consensys.net/api#tag/Experimental/operation/postValidatorLiveness) in a slightly different way:
- **Verb:** POST (same)
- **Endpoint:**  `/eth/v1/validator/liveness/{epoch}` (same)
- **Body example:** 
```
{
   "indices": [
      "1",
      "42",
      "1663"
   ]
}
```
(Map with `indices` as key and with a list of indexes as value)

Is there any specific reason not to conform to the specification?
If no, would it be possible to adapt it?
If you don't want do break retro-compatibility, I guess it's possible for the endpoint to accept both current Teku body and ETH beacon API specification body.

rolfyone:
Also removed the `experimental` tag as it's used in doppelganger feature.

fixes #7204

## Documentation

## Changelog


StefanBratanov (changes requested):
I think we need to also change `OkHttpValidatorRestApiClient` in this PR to conform to the spec, since `ValidatorLivenessRequest` is still using the other input. Otherwise VC and BN compatibility would break.

mehdi-aouadi (dismissed):
We could also remove `ValidatorLivenessRequest` since it's not used anymore

rolfyone:
> We could also remove `ValidatorLivenessRequest` since it's not used anymore

ppl - comments are fine, not sure i'm a huge fan of spamming 'changes requested'

rolfyone:
> I think we need to also change `OkHttpValidatorRestApiClient` in this PR to conform to the spec, since `ValidatorLivenessRequest` is still using the other input. Otherwise VC and BN compatibility would break.

this was done at the bottom of the changes list, are there more changes you were expecting? because the change currently there is what you're talking about as far as I can see...

StefanBratanov (approved):
Oh yes, sorry. I somehow missed on these changed files.

rolfyone:
@mehdi-aouadi you've taken ownership to merge this with your changes-requested flag.

PR #7219, opened 2023-06-02
Pass versioned hashes to EL for validation when processing block

### LINKED ISSUE #7203: Use engine_newPayloadV3 to pass versioned_hashes to EL for validation
https://github.com/ethereum/consensus-specs/pull/3359

StefanBratanov:
## PR Description

- Removed KZGCommitmentsProcessor since `process_blob_kzg_commitments` in spec no longer required
- Introduced `computeNewPayloadRequest` in `BlockProcessor` and logic is different in Bellatrix and Deneb as per spec
- Removed unused methods in MiscHelpersDeneb

## Fixed Issue(s)
fixes #7203 

## Documentation

## Changelog


ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/deneb/block/BlockProcessorDenebTest.java
```diff
@@ -34,17 +39,43 @@ protected Spec createSpec() {
...
+    final NewPayloadRequest newPayloadRequest =
+        spec.getBlockProcessor(UInt64.ONE).computeNewPayloadRequest(blockBody);
+
+    assertThat(newPayloadRequest.getExecutionPayload())
+        .isEqualTo(blockBody.getOptionalExecutionPayload().orElseThrow());
+    assertThat(newPayloadRequest.getVersionedHashes()).isPresent();
+    assertThat(newPayloadRequest.getVersionedHashes().get())
+        .hasSize(3)
+        .allMatch(
+            versionedHash -> versionedHash.isVersion(SpecConfigDeneb.VERSIONED_HASH_VERSION_KZG));
```
zilm13: Maybe also add a check that hashes match?

zilm13 (commented):
I see that spec now contains method
`verify_and_notify_new_payload` which has 1 or 2 verification pre-checks (2 for Deneb) before calling EL. I think we should add it too or I missed its implementation

StefanBratanov:
> I see that spec now contains method `verify_and_notify_new_payload` which has 1 or 2 verification pre-checks (2 for Deneb) before calling EL. I think we should add it too or I missed its implementation

My understanding of the spec is that `verify_and_notify_new_payload` is essentially an abstraction of the EL communication. In case of Teku, `is_valid_block_hash`, `is_valid_versioned_hashes` and `notify_new_payload` are done by EL (newPayloadV3), since we don't have the information to do these verficiations.

StefanBratanov:
> > I see that spec now contains method `verify_and_notify_new_payload` which has 1 or 2 verification pre-checks (2 for Deneb) before calling EL. I think we should add it too or I missed its implementation
> 
> My understanding of the spec is that `verify_and_notify_new_payload` is essentially an abstraction of the EL communication. In case of Teku, `is_valid_block_hash`, `is_valid_versioned_hashes` and `notify_new_payload` are done by EL (newPayloadV3 -`OptimisticExecutionPayloadExecutor`), since we don't have the information to do these verficiations.

zilm13:
Sorry, my misunderstanding, rechecked spec again and you are right, everything should be done on el side

zilm13 (approved):

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/deneb/block/BlockProcessorDenebTest.java
StefanBratanov: done

PR #7224, opened 2023-06-05
Fixes historical sync

tbenr:
Fix a bug when computing  FirstSlotWithBlobSupport that was keeping genesis spec.
Moved in spec and it will stream active milestones and will keep the first supporting deneb.

Fixed another bug which was an edge case: if the batch we have downloaded contains only the genesis block, we will exclude it from signature verification, ending up with an empty signature verification sets, which fails. So we keep downloading this genesis block forever.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
@@ -884,9 +884,8 @@ public class Spec {
     return forkSchedule.getSupportedMilestones().stream()
         .flatMap(milestone -> forMilestone(milestone).getConfig().toVersionDeneb().stream())
         .findFirst()
-        .map(
-            specConfigDeneb ->
-                specConfigDeneb.getDenebForkEpoch().times(specConfigDeneb.getSlotsPerEpoch()));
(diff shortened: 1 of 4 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -880,6 +881,19 @@ public UInt64 computeSubnetForBlobSidecar(final SignedBlobSidecar signedBlobSide
     return signedBlobSidecar.getBlobSidecar().getIndex().mod(BLOB_SIDECAR_SUBNET_COUNT);
   }
 
+  public Optional<UInt64> computeFirstSlotWithBlobSupport() {
+    return forkSchedule.getActiveMilestones().stream()
+        .map(
```
StefanBratanov: You can just `flatMap` here and avoid the next two `filter` and `map`

StefanBratanov (approved):

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
tbenr: yeah

PR #7226, opened 2023-06-06
Log block value in ETH instead of wei

### LINKED ISSUE #6972: Log block value in ETH instead of wei
Currently, we log the block values in wei in `ExecutionBuilderModule`. It would be nicer if we can log these value in ETH instead. We can use the code below as a base to convert wei to readable ETH value.

```java
private String weiToEth(final UInt256 wei) {
    final String weiString = wei.toDecimalString();
    final StringBuilder stringBuilder = new StringBuilder(20);

    if(weiString.length() > 18) {
      final int pointPosition =  weiString.length()-18;
      final String ethInteger =weiString.substring(0, pointPosition);
      final String ethDecimals = weiString.substring(pointPosition);

      stringBuilder.append(ethInteger).append(".").append(ethDecimals);
    } else {
      final int paddingZeros =  18-weiString.length();
      stringBuilder.append("0.");
      IntStream.range(0,paddingZeros).forEach(__ -> stringBuilder.append("0"));
      stringBuilder.append(weiString);
    }

    return stringBuilder.toString();
  }
```

courtneyeh:
fixes #6972 

## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -404,26 +404,44 @@ private void logReceivedBuilderExecutionPayload(final ExecutionPayload execution
...
+      stringBuilder.append(ethInteger).append(".").append(ethDecimals);
+    } else {
+      final int paddingZeros = 18 - weiString.length();
+      stringBuilder.append("0.");
+      IntStream.range(0, paddingZeros).forEach(__ -> stringBuilder.append("0"));
+      stringBuilder.append(weiString);
+    }
+
+    return stringBuilder.toString();
+  }
```
rolfyone: so we'll want to think about this...

 - we want it to be fast, and we want tests.... Not sure whether we should be somehow doing this mathematically, or some kind of hybrid approach... but unit tests would let us see the ins and outs...

lucassaldanha: Maybe we should consider using something like https://github.com/web3j/web3j/blob/master/utils/src/main/java/org/web3j/utils/Convert.java ?

lucassaldanha: From their tests it looks like it does exactly what we want: https://github.com/web3j/web3j/blob/master/utils/src/test/java/org/web3j/utils/ConvertTest.java#L38-L40

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -404,26 +405,28 @@ private void logReceivedBuilderExecutionPayload(final ExecutionPayload execution
   private void logReceivedBuilderBid(final BuilderBid builderBid) {
     final ExecutionPayloadHeader payloadHeader = builderBid.getExecutionPayloadHeader();
     LOG.info(
-        "Received Builder Bid (Block Number = {}, Block Hash = {}, MEV Reward (wei) = {}, Gas Limit = {}, Gas Used = {})",
+        "Received Builder Bid (Block Number = {}, Block Hash = {}, MEV Reward (ETH) = {}, Gas Limit = {}, Gas Used = {})",
         payloadHeader.getBlockNumber(),
         payloadHeader.getBlockHash(),
-        builderBid.getValue().toDecimalString(),
+        weiToEth(builderBid.getValue()),
```
lucassaldanha, line 410: We need to be careful when converting BigDecimal to strings. Depending on how small the value is, it might stringify it using scientific notation...

Example:
```
BigDecimal number = new BigDecimal("0.0000001");

System.out.println(number); 
// prints 1E-7

System.out.println(number.toPlainString());
// prints 0.0000001
```

I suggest we use `toPlainString()` when logging BigDecimals to be consistent.

rolfyone (approved):
LGTM

PR #7227, opened 2023-06-06
Refactor ReputationManager to allow for multiple implementations

### LINKED ISSUE #7130: Allow multiple implementation of `ReputationManager`
We should make `ReputationManager` be an interface and current implementation be `DefaultReputationManager`.

This is the initial step to being able to have multiple versions and being able to swap them easily.

courtneyeh:
fixes #7130 

## Documentation

## Changelog


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/ReputationManager.java
```diff
@@ -13,156 +13,32 @@
...
   // software update
-  static final UInt64 BAN_PERIOD = UInt64.valueOf(TimeUnit.HOURS.toSeconds(12));
+  UInt64 BAN_PERIOD = UInt64.valueOf(TimeUnit.HOURS.toSeconds(12));
 
-  static final int LARGE_CHANGE = 10;
-  static final int SMALL_CHANGE = 3;
-  private static final int DISCONNECT_THRESHOLD = -LARGE_CHANGE;
-  private static final int MAX_REPUTATION_SCORE = 2 * LARGE_CHANGE;
+  int LARGE_CHANGE = 10;
+  int SMALL_CHANGE = 3;
```
courtneyeh: Do I need to make it so that these static values can change with each implementation?

StefanBratanov: I would just move all of these values to DefaultReputationManager same as before, since they seem more like implementation details.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/ReputationManager.java
```diff
@@ -13,156 +13,32 @@
...
-import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
-import tech.pegasys.teku.infrastructure.time.TimeProvider;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.networking.p2p.network.PeerAddress;
 import tech.pegasys.teku.networking.p2p.peer.DisconnectReason;
-import tech.pegasys.teku.networking.p2p.peer.NodeId;
 
-public class ReputationManager {
+public interface ReputationManager {
   // This is not a big ban, we expect the peer could be usable very soon
```
StefanBratanov: We can add a `NOOP` value here similar to BlobSidecarManager. Even though it won't be used anywhere, It could be used easily to switch implementations.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/ReputationManager.java
StefanBratanov: I would just move all static finals back to the DefaultReputationManager instead of the interface since they look like implementation details.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/DefaultReputationAdjustment.java
```diff
@@ -13,15 +13,15 @@
 
 package tech.pegasys.teku.networking.p2p.reputation;
 
-public enum ReputationAdjustment {
-  LARGE_PENALTY(-ReputationManager.LARGE_CHANGE),
-  SMALL_PENALTY(-ReputationManager.SMALL_CHANGE),
-  SMALL_REWARD(ReputationManager.SMALL_CHANGE),
-  LARGE_REWARD(ReputationManager.LARGE_CHANGE);
+public enum DefaultReputationAdjustment {
```
StefanBratanov: I would leave this named `ReputationAdjustment`

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/ReputationManager.java
```diff
@@ -13,156 +13,45 @@
...
 
-public class ReputationManager {
-  // This is not a big ban, we expect the peer could be usable very soon
-  static final UInt64 COOLDOWN_PERIOD = UInt64.valueOf(TimeUnit.MINUTES.toSeconds(2));
-  // It's a big ban, we expect that the peer is not useful until major changes, for example,
-  // software update
-  static final UInt64 BAN_PERIOD = UInt64.valueOf(TimeUnit.HOURS.toSeconds(12));
+public interface ReputationManager {
+  ReputationManager NOOP =
+      new ReputationManager() {
```
StefanBratanov, line 22: I think returning true for the NOOP operations makes more sense aka allowing everything.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/ReputationManager.java
```diff
@@ -13,156 +13,44 @@
...
 
-    @Override
-    public String toString() {
-      return MoreObjects.toStringHelper(this)
-          .add("suitableAfter", suitableAfter)
-          .add("score", score)
-          .toString();
-    }
-  }
+  boolean adjustReputation(final PeerAddress peerAddress, final ReputationAdjustment effect);
```
rolfyone, line 56: this is a little confusing, we might want a comment on the interface to describe the what the boolean means, because the name isn't indicative of it...

it looks like the original returns `shouldDisconnect`, and our noop returns true, for example... but if the output is whether it should disconnect, our noop would be expected to return false i think...

rolfyone (approved):
LGTM

Issue #7170, opened 2023-05-23
Implement updated builder apis


tbenr:
https://github.com/ethereum/builder-specs/pull/61

PR #7230, opened 2023-06-06
Add Deneb builder containers

StefanBratanov:
## PR Description
Add/extend containers according to spec: https://github.com/ethereum/builder-specs/blob/main/specs/deneb/builder.md#containers

- Versioned `BuilderBid` (similar to `ExecutionPayload`)
- Moved the engine api `BlobsBundle` to `execution` package
- Property tests for new containers
- Random generations of new containers in DataStructuresUtil

## Fixed Issue(s)
will help with #7170 

## Documentation

## Changelog


zilm13 (commented):
What I noticed:
- BlindedBlobsBundle, BlobsBundle: we usually provided unpacked (from ssz) getters, if I'm not wrong you'll need it in implementation to build a block
- BlindedBlobsBundleSchema, BlobsBundleSchema: we usually pass specConfig as an input, not exact variable and it sounds reasonable, especially when you could need several variables and it varies from one specConfig to another

StefanBratanov:
> What I noticed:
> 
> * BlindedBlobsBundle, BlobsBundle: we usually provided unpacked (from ssz) getters, if I'm not wrong you'll need it in implementation to build a block
> * BlindedBlobsBundleSchema, BlobsBundleSchema: we usually pass specConfig as an input, not exact variable and it sounds reasonable, especially when you could need several variables and it varies from one specConfig to another

Agree about the SpecConfig passing to the schema. Fixed that. Erm not sure about the first. For example `BeaconBlockBody` the getters are ssz stuff. I will convert to an internal BlobsBundle when required in next PRs.

zilm13 (approved):
LGTM

PR #7231, opened 2023-06-06
Blobs pool improvements

tbenr:
1. This PR improves historical sync which now creates a `BlockBlobSidecarsTracker` that doesn't go in the BlobsPool anymore. Since there is no actual import, that Tracker would have been forgotten there filling up the pool.
Now there is a specialized flow that creates a tracker on the fly just to be used with the on availability checker.

2. When we forward sync, gossip is started when we are close to the head while we are still importing blobs and blocks via syncing. Added a filter on the pool to avoid tracking blobs related to already imported blocks.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7236, opened 2023-06-07
Some deneb nits

tbenr:
- generate random numbers of blobs in stub
- fix logging bug in tracker

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
```diff
@@ -288,8 +290,9 @@ public SafeFuture<PayloadStatus> engineNewPayload(final NewPayloadRequest newPay
         Optional.ofNullable(knownPosBlocks.get(executionPayload.getBlockHash()))
             .orElse(payloadStatus);
     LOG.info(
-        "newPayload: executionPayload blockHash: {} -> {}",
+        "newPayload: executionPayload blockHash: {}  versionedHashes: {} -> {}",
         executionPayload.getBlockHash(),
+        newPayloadRequest.getVersionedHashes().map(List::size),
```
StefanBratanov: Why not just print the hashes? Would be more valuable.

StefanBratanov (approved):
LGTM with one nit

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
tbenr: ye, my intent was to avoid adding too many noise from the stub, but makes sense.

PR #7242, opened 2023-06-08
Builder REST API changes for Deneb

StefanBratanov:
## PR Description

Implements changes described in https://ethereum.github.io/builder-specs/?urls.primaryName=dev

- Changed **BuilderClient** interface to pass a `SignedBlindedBlockContainer` which could be either `SignedBlindedBlockContents` or `SignedBeaconBlock`
- Changed return type of `getPayload` to `BuilderPayload` which can contain `BlobsBundle` as well as the `ExecutionPayload`
- Introduced `BuilderPayloadSchema` to be able to deserialize the `BuilderPayload`
- Added deneb examples to `RestBuilderClientTest`
- Made some renamings to be consistent

_TODO_ in next PR: use the `BuilderPayload` response to create BlobSidecar(s) and fallback to local if needed.

## Fixed Issue(s)
will help with #7170 

## Documentation

## Changelog


zilm13 (approved):
LGTM

But I don't like a bit that now `ExecutionPayload` extends `BuilderPayload`.
I understand that it's a way to make all of this work in Java, but intuitively `ExecutionPayload` is base spec thing and Builder is 3rd party thing, so this hierarchy touches an eye. If we could find some better domain implementation for this, it will be better I think. Just my opinion

StefanBratanov:
> LGTM
> 
> But I don't like a bit that now `ExecutionPayload` extends `BuilderPayload`. I understand that it's a way to make all of this work in Java, but intuitively `ExecutionPayload` is base spec thing and Builder is 3rd party thing, so this hierarchy touches an eye. If we could find some better domain implementation for this, it will be better I think. Just my opinion

I see what you mean. I can't think of cleaner approach though at the moment and still take advantage of same schemas, type definitions etc.

PR #7248, opened 2023-06-09
BlobSidecarsByRange send only canonical blobs

### LINKED ISSUE #7246: Make sure `BlobSidecarByRange` RPC responds with only canonical blobs
A logic similar to the one implemented here `BeaconBlocksByRangeMessageHandler#sendMatchingBlocks` should make sure we send only canonical blobs for non-finalized slot range

tbenr:
This PR makes sure that BlobSidecarsByRange RPC responds only with blobsidecars belonging to canonical chain.

It is implemented with the following logic:

- Splits the range in two logical sub ranges of slots: finalized and non finalized.
  - For the finalized subrange, It assumes the storage is already pruned from non canonical blobs, so we keep them as they are.
  - For non-finalized subrange, it obtains the canonical  `Map<Slot,BlockRoot>` from combinedChainDataClient, and it filters out blobs that do not refer to any of those SlotAndBlockRoot.

fixes #7246 

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
@@ -146,7 +146,7 @@ public class BlobSidecarsByRangeMessageHandler
                     combinedChainDataClient.getAncestorRoots(startSlot, UInt64.ONE, hotSlotsCount);
 
                 // refresh finalized slot to avoid race condition that can occur if we finalize just
-                // after getting hot canonical roots
+                // before getting hot canonical roots
                 finalizedSlot = combinedChainDataClient.getFinalizedBlockSlot().orElse(UInt64.ZERO);
               } else {
                 canonicalHotRoots = ImmutableSortedMap.of();
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
@@ -204,7 +204,8 @@ public class BlobSidecarsByRangeMessageHandler
     private final UInt64 startSlot;
     private final UInt64 endSlot;
 
-    private Optional<Iterator<SlotAndBlockRootAndBlobIndex>> iterator = Optional.empty();
+    private Optional<Iterator<SlotAndBlockRootAndBlobIndex>> blobSidecarKeysIterator =
+        Optional.empty();
 
     RequestState(
         final ResponseCallback<BlobSidecar> callback,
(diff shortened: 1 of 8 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -222,31 +223,54 @@ SafeFuture<Void> sendBlobSidecar(final BlobSidecar blobSidecar) {
...
     }
 
-    private SafeFuture<Optional<BlobSidecar>> getNextElement(
+    private SafeFuture<Optional<BlobSidecar>> getNextBlobSidecar(
         final Iterator<SlotAndBlockRootAndBlobIndex> iterator) {
       if (iterator.hasNext()) {
-        return combinedChainDataClient.getBlobSidecarByKey(iterator.next());
-      } else {
-        return SafeFuture.completedFuture(Optional.empty());
+        final SlotAndBlockRootAndBlobIndex slotAndBlockRootAndBlobIndex = iterator.next();
```
mehdi-aouadi: You could rename this iterator as well 😄

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -127,12 +132,31 @@ public void onIncomingMessage(
...
-              final BlobSidecarsByRangeMessageHandler.RequestState initialState =
-                  new BlobSidecarsByRangeMessageHandler.RequestState(
-                      callback, maxRequestBlobSidecars, startSlot, endSlot);
+
+              final UInt64 finalizedSlot =
+                  combinedChainDataClient.getFinalizedBlockSlot().orElse(UInt64.ZERO);
+
+              final SortedMap<UInt64, Bytes32> canonicalHotRoots;
+              if (endSlot.isGreaterThan(finalizedSlot)) {
+                final UInt64 count = endSlot.minus(finalizedSlot);
```
zilm13: i would call it something like hotSlotsCount to not be confused with request count

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -203,50 +227,76 @@ class RequestState {
...
+        }
+
+        // non-canonical, try next one
+        return getNextBlobSidecar(blobSidecarKeysIterator);
       }
+
+      return SafeFuture.completedFuture(Optional.empty());
+    }
+
+    private boolean isCanonical(final SlotAndBlockRootAndBlobIndex slotAndBlockRootAndBlobIndex) {
```
zilm13: I'd clarify, something like `isCanonicalHotBlobSidecar`, maybe you get something shorter on a mind

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -203,50 +227,76 @@ class RequestState {
     private final UInt64 maxRequestBlobSidecars;
     private final UInt64 startSlot;
     private final UInt64 endSlot;
+    private final UInt64 finalizedSlot;
+    private final Map<UInt64, Bytes32> canonicalHotRoots;
 
-    private Optional<Iterator<SlotAndBlockRootAndBlobIndex>> iterator = Optional.empty();
+    private Optional<Iterator<SlotAndBlockRootAndBlobIndex>> blobSidecarKeysIterator =
```
zilm13, line 238: Maybe add a comment, something like includes both canonical and non-canonical blob sidecars on a hot part?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -127,12 +132,31 @@ public void onIncomingMessage(
...
+
+              final UInt64 finalizedSlot =
+                  combinedChainDataClient.getFinalizedBlockSlot().orElse(UInt64.ZERO);
+
+              final SortedMap<UInt64, Bytes32> canonicalHotRoots;
+              if (endSlot.isGreaterThan(finalizedSlot)) {
+                final UInt64 count = endSlot.minus(finalizedSlot);
+                final UInt64 hotBlocksStartSlot = finalizedSlot.plus(1);
+
+                canonicalHotRoots =
```
zilm13, line 145: I'm a little worried on concurrency here, but looking at code, if we have new finalized epoch between this line and finalizedSlot above, we just get empty hotRoots and return smaller set to the requester, maybe even empty. Not a fatal, just noticing. Though we should reply with at least one element as I remember by spec. Not a big bug I think anyway

zilm13 (approved):
LGTM, just some boring comments

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
tbenr: Good point!
I should have improved that.

tbenr: I suspect there could be still some edge cases but conditions in which we may responds with empty blobs are at least lower now.

mehdi-aouadi: It's better this way! Even though we could still finalise while processing the request and hence fail sending some blobs but I think it's not a big issue.
I was thinking if it's worth rechecking the `finalizedSlot` within `getNextBlobSidecar` (instead of fixing it for the whole request) and check if it moved before retrieving the blobs? In that case we could update the `canonicalHotRoots` accordingly (remove the new blocks with slot <= new finalized slot)... I understand that this could be an overkill though so just thinking out loud :)

mehdi-aouadi (approved):
LGTM

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
tbenr: Sorry I saw your comment after selecting auto merge. What i'm more concerned about is to get a consistent finalizedSlot-canonicalRoots pair instead of being concerned about get them updated.
In theory the definitive solution is to have a single method that atomically get hotRoots and the finalized slot.
If things change while we iterate it can always happen but i actually accepted that for performance reason (it was actually my first draft and i moved away from it)

mehdi-aouadi: No problem! I totally like it as a first iteration. Let's see how it goes

PR #7250, opened 2023-06-10
Use networking constants from the spec config

### LINKED ISSUE #7179: Move p2p networking setting from Constants to SpecConfig

zilm13:
## PR Description
WIP 
- separate interface NetworkingSpecConfig made for some domain separation
- Shotrcut made in `spec` because networking spec is part of the main config should not be changed in forks with presets
- MIN_EPOCHS_FOR_BLOCK_REQUESTS kept commented to omit spec config hardcofing instead of calculations for other configs (not mainnet/minimal)
- I've noticed that `SUBNETS_PER_NODE` is 2 in spec, and was 1 in our code
- I've not found subnet calculation logic from the spec in our code (https://github.com/ethereum/consensus-specs/blob/68ce45b3058186ff9e8bdecb19d51327851a4259/specs/phase0/p2p-interface.md#attestation-subnet-subcription), looks like we do simple random. So `ATTESTATION_SUBNET_PREFIX_BITS` and `ATTESTATION_SUBNET_EXTRA_BITS` looks unused

TODO:
- polishing
- add deneb constants
- assuming some people are using custom devnet configs which may miss networking constants, what we could do for them to make better experience with this change?
- will do `GOSSIP_MAX_SIZE_BELLATRIX` and `MAX_CHUNK_SIZE_BELLATRIX` 

## Fixed Issue(s)

Fixes #7179 
+ adds deneb constants from the 1.4.0.alpha.2

## Documentation

## Changelog


rolfyone:
In the other one, theres a bunch of constants in the mainnet.yaml to uncomment as you go through... I wonder if closing this, then working through 1-2 at a time would be a good approach... that was my plan before you did this huge sweep...

zilm13:
@rolfyone I'm repairing this and I will split it to the smaller parts

zilm13:
Closed in favor of #7300 #7302 #7303

PR #7254, opened 2023-06-12
Add deneb networking constants

zilm13:
## PR Description
Draft.
I don't like inconsistency of networking constants among spec:
- Max gossip and max chunk constants are in main config
- Deneb networking constants are in main config
- Bellatrix max gossip and chunk constants are in spec (neither config nor presets)

In our code previously:
- Max gossip and max chunk constants - phase0 preset
- Bellatrix max gossip and max chunk - bellatrix preset

What I think it should be:
- either put all in config, so in spec we will need to include 2 Bellatrix gossip constants in config
- or keep base networking constants in base config, move deneb constants to deneb preset, add bellatrix gossip constants to bellatrix preset

I think ideologically it should be in each preset, but names not clashes and it will be simpler for us to load them all from the main config. In the code I've tried both usage when we need `specVersion` for epoch/milestone and when we could use base config with `getNetworkingConfig()` shortcut to see how it's simpler.

**UPDATE:** Questions are resolved with the following outcome:
- Bellatrix networking constants are abolished, based constants are aligned with its numbers. We will do the same (in separate PR)
- We put Deneb networking constants in main config like it's done in consensus specs

## Documentation

## Changelog


zilm13:
All done except failing `SpecConfigLoaderTest`. I'm not sure what is the best way to hack this, looking for advices.

- Current consensus specs configuration is to put all networking constants to the main config, so I adopted it (not adding other constants yet though) and unified networking constants usage via `getNetworkingConfig()` from spec or specVersion. Big number of files changed is mostly due to this unification.
- `SpecConfig` is updated to look not so messy
- Moves `GOSSIP_MAX_SIZE` and `MAX_CHUNK_SIZE` to main configs like in Consensus specs
- Adds deneb networking constants in configs: `MAX_REQUEST_BLOCKS_DENEB`, `MAX_REQUEST_BLOB_SIDECARS`, `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS`

Next PR will remove `GOSSIP_MAX_SIZE_BELLATRIX` and `MAX_CHUNK_SIZE_BELLATRIX` (already removed in consensus specs)

zilm13:
So, I'm stuck. Describing the issue in details

There are a number of networking constants added to consensus specs including [Deneb specific constants](https://github.com/ethereum/consensus-specs/blob/dev/configs/mainnet.yaml#L128-L136).
I'm trying to put original consensus specs configs/presets in our repository and it's not easy. So for these deneb networks we had several options:
- put it in `SpecConfig`. As we must have all fields defined for `SpecConfig` it will not work for other configs like `prater` or `kiln`. It looks not good to add them to those configs too.

So I left constants in `config` but added setters to `SpecConfigDeneb`. It raises another issue: we fill all configs for any network independently whether fork is defined in it or not. So if keeping it `config` we are unable to fill it for networks where Deneb is not defined. Thinking that it's not good to fill full forks settings for networks where such forks are not defined, I've tried to make some kind of NoOpConfigs for each of forks and not fill anything in it except fork version and epoch on criteria of `FAR_FUTURE_EPOCH`. This approach is not working too because we has minimal config which presets are the base for many networks and has all transition epochs set to `FAR_FUTURE_EPOCH`.

Here I stuck.
So I see 2 options to solve this:
- put Deneb networking constants in presets unlike the consensus-specs and follow our old config/presets independently of consensus-specs
- put Deneb networking constants all configs including prater, kiln, whatever

I strongly dislike 2nd option, so I will try the first one. Additionally I could try to propose similar changes in `consensus-specs`

zilm13:
> put Deneb networking constants in presets unlike the consensus-specs and follow our old config/presets independently of consensus-specs

This solution doesn't work too as we create even `SpecConfigDeneb` for all networks including those that doesn't have deneb fork. It was passing previously because primitives like `int` are automatically filled with 0 so, there were no issues. There is an option to switch fields to `int` but it stinks too. If we are ok with it, I will do it. 
My idea to make stub configs will not work as we use `minimal` config widely including reference tests and it has `FAR_FUTURE_EPOCH` forks while is used for forks other than phase0.

zilm13:
Closed in favor of #7304

PR #7256, opened 2023-06-13
Improve calculation of total bytes of blobs

StefanBratanov:
## PR Description
Don't stream over blob bytes unnecessarily to get total size, when size of blobs is known.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #7260, opened 2023-06-14
Updated gossip max size and max_chunk_size in line with spec.

rolfyone:
In line with https://github.com/ethereum/consensus-specs/pull/3394 moved the fields into mainnet.yaml and minimal.yaml and updated the phase0 values to be the old bellatrix values, since that's basically in line with implementation anyway (bellatrix enabled meant use bellatrix values)

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

PR #7261, opened 2023-06-14
Refactor to help with Deneb builder flow

StefanBratanov:
## PR Description

- Change `blobsBundle` parameter in `ExecutionPayloadResult` to `Optional<SafeFuture<Optional<BlobsBundle>>> blobsBundleFuture` and changed the order to make more sense. In normal non-blinded flow, it is present, in blinded flow it is empty.
- Added `BlobsBundle` to `FallbackData` and `BlindedBlobsBundle` to `HeaderWithFallbackData`
- Rename nits
- Changed `GetPayloadResponse` to store `BlobsBundle` as optional as it makes more sense. 
- Changed `builderGetPayload` in `ExecutionLayerChannel` to accept `SignedBlockContainer`
- Changed `getUnblindedPayload` in `ExecutionLayerBlockProductionManager` to return `BuilderPayload` which would most likely be cached for X amount of slots, probably same value as `EXECUTION_RESULT_CACHE_RETENTION_SLOTS`
- Changed the callers of `getUnblindedPayload` to actually pass a `SignedBlockContainer` instead of the block (this required changes in BlindBlockUtil and SignedBeaconBlockUnblinder
- Added `Optional<BuilderPayload> getCachedUnblindedPayload(UInt64 slot)` in `ExecutionLayerBlockProductionManager` to help with getting a cached builder payload. WIll be used when unblinding the blob sidecars, which happens after block unblinding.
- Changed the `createBlobSidecarsUnblinderSelector` to get the cached `BuilderPayload`
- Changed the `createBlindedBlobSidecarsSelector` to use the `BlindedBlobsBundle` from the `HeaderWithFallbackData`
- Implemented setting the kzg commitments in a block for a blinded flow using the blinded blobs bundle from `HeaderWithFallbackData`

**TODO:**

- Pretty much all is left is changes in `ExecutionBuilderModule` to construct a `BuilderPayload` and `HeaderWithFallbackData` when local fallback is needed.
- And enable/fix disabled tests

## Fixed Issue(s)
will help with #7170 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockFactoryPhase0.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockFactoryPhase0.java
@@ -16,7 +16,6 @@ package tech.pegasys.teku.validator.coordinator;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import java.util.Optional;
-import java.util.function.Function;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.bls.BLSSignature;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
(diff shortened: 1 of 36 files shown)


tbenr (approved):
LGTM

PR #7268, opened 2023-06-17
Disabling stuff that should not run when Deneb or Capella are not supported

zilm13:
## PR Description
Maybe I missed something but it's all I caught disabling deneb constants. Also we should be careful with things like this in our code, it will not work as expected
```
forMilestone(highestSupportedMilestone)
        .getConfig()
        .toVersionDeneb()
        .map(SpecConfigDeneb::getMaxBlobsPerBlock);
```
It actually doesn't work like it should (being an Optional) because we have always some `SpecConfigDeneb`. 
## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -867,7 +871,12 @@ public boolean isAvailabilityOfBlobSidecarsRequiredAtEpoch(
   }
 
   public Optional<Integer> getMaxBlobsPerBlock() {
-    return forMilestone(getForkSchedule().getHighestSupportedMilestone())
+    final SpecMilestone highestSupportedMilestone =
```
tbenr, line 897: what about

    final SpecMilestone highestSupportedMilestone =
      getForkSchedule().getHighestSupportedMilestone();
    return Optional.ofNullable(forMilestone(highestSupportedMilestone).getConfig())
        .flatMap(SpecConfig::toVersionDeneb)
        .map(SpecConfigDeneb::getMaxBlobsPerBlock);

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/DefaultSyncServiceFactory.java
```diff
@@ -127,9 +129,14 @@ public SyncService create(final EventChannels eventChannels) {
...
-    final FetchRecentBlobSidecarsService fetchRecentBlobSidecarsService =
-        FetchRecentBlobSidecarsService.create(
-            asyncRunner, blobSidecarPool, forwardSyncService, fetchTaskFactory, spec);
+    final RecentBlobSidecarFetcher fetchRecentBlobSidecarsService;
+    if (spec.isMilestoneSupported(SpecMilestone.DENEB)) {
+      fetchRecentBlobSidecarsService =
+          FetchRecentBlobSidecarsService.create(
+              asyncRunner, blobSidecarPool, forwardSyncService, fetchTaskFactory, spec);
+    } else {
+      fetchRecentBlobSidecarsService = RecentBlobSidecarFetcher.NOOP;
```
tbenr: this naming difference is blowing up my mind.

tbenr (commented):
in general looks good. But the only thing I'm not super convinced about is the general bjective. I dont like haveing all these ifs.
A part of me says we should minimize those spaghetti and enable what is safe to be enabled in advance.

zilm13:
@tbenr did my best at refactoring this. Anyway I'm ok to NO MERGE if it's not needed

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -36,19 +37,21 @@ public class ThrottlingSyncSource implements SyncSource {
   private final SyncSource delegate;
 
   private final RateTracker blocksRateTracker;
-  private final RateTracker blobSidecarsRateTracker;
+  private final Optional<RateTracker> blobSidecarsRateTracker;
```
tbenr: In this case I'd like to follow the `NOOP` pattern. Proliferation of Optionals is a form of spaghetti

tbenr (approved):
LGTM. I suggested one more improvement

tbenr (approved):
LGTM

PR #7270, opened 2023-06-17
Sanitize SpecConfigBuilder on primitive field setters

zilm13:
## PR Description
Ensures Builders have actually non-primitive setters, because primitive setters are silently filled with 0's and doesn't lead to client startup failure with clear missing field error. This PR includes check only for `SpecConfigBuilder` as fork builders are more complicated to deal with.

Disabled test is not actually working because swift config in release doesn't contain MAX_GOSSIP and MAX_CHUNK_SIZE. It pretends to work before this release but was actually filled values with 0s and would fail in production environment without any clear evidence
## Documentation

## Changelog


rolfyone (approved):
LGTM

rolfyone:
> Disabled test is not actually working because swift config in release doesn't contain MAX_GOSSIP and MAX_CHUNK_SIZE. It pretends to work before this release but was actually filled values with 0s and would fail in production environment without any clear evidence


I find this curious because I added these two fields to swift and less-swift... admittedly they're the old phase0 values because an AT was breaking if i updated it, but I'm surprised they came out as 0...

zilm13:
@rolfyone but old release docker image doesn't contain it

PR #7273, opened 2023-06-19
Removed ProgressiveBalancesMode as no longer required

### LINKED ISSUE #6998: Remove progressive balance modes
Since now progressive balance is by default in FULL, we can do a code cleanup, remove the CLI option and different code paths.
#6993 was a spike in this direction

rolfyone:
Now that progressiveBalancesMode is always 'full', we no longer need to toggle it, so cleaning up.

partially addresses #6998

## Documentation

## Changelog


tbenr:
I think what is missing here is the removal of the progressiva balance check and the actual usage of it.

It happens in the `AbstractEpochProcessor` where we call `progressiveTotalBalances.checkResult`.

```java
    final ProgressiveTotalBalancesUpdates progressiveTotalBalances =
        transitionCaches.getProgressiveTotalBalances();
    progressiveTotalBalances.checkResult(specConfig, state.getSlot(), totalBalances);
    progressiveTotalBalances.onEpochTransition(validatorStatuses.getStatuses());
    processJustificationAndFinalization(state, validatorStatuses.getTotalBalances());
 ```
 
 IIRC, from the spike mentioned in the issue, it requires some changes.
 If we want to implement it in a separate PR, i'm ok with it, just  don't close the issue :)

rolfyone:
> IIRC, from the spike mentioned in the issue, it requires some changes.
> If we want to implement it in a separate PR, i'm ok with it, just don't close the issue :)

Yep I can look into the spike next, changed description to 'partially addresses'

lucassaldanha (approved):

PR #7274, opened 2023-06-19
Don't disconnect static peers

### LINKED ISSUE #7155: don't disconnect static peers
Regardless of the scoring of a static peer, it should not be disconnected...

Possibly shouldn't get killed if they get told to back off either?

courtneyeh:
fixes #7155 

## Documentation

## Changelog


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/DefaultReputationManager.java
```diff
@@ -92,6 +99,9 @@ public void reportDisconnection(
   @Override
   public boolean adjustReputation(
       final PeerAddress peerAddress, final ReputationAdjustment effect) {
+    if (peerPools.getPool(peerAddress.getId()).equals(PeerConnectionType.STATIC)) {
```
rolfyone: i think `getPool` would be better off as `getPeerConnectionType`

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/DefaultReputationManager.java
```diff
@@ -92,6 +99,9 @@ public void reportDisconnection(
...
+      return false;
+    }
```
rolfyone, line 104: we should definitely be able to have a unit test for this bit

rolfyone (approved):
LGTM

PR #7282, opened 2023-06-20
Version 1 of SpecConfig changes: Don't create SpecConfigs which are not supported

zilm13:
## PR Description
Built on top of #7268.
Version 1 of strict field checks for SpecConfig. Doesn't create SpecConfigs for Milestones that are not supported
Pros:
- cleaner solution. We don't create incorrect entities with filled 0's. We don't create things that are not used in network
- `toVersionXXX()` works as it was designed (currently it is always non-empty for any fork and network)

Cons:
- requires very careful merging, it's possible to miss some things that are still created for higher milestones, especially for networks, for which we don't have throughful testing (other than Mainnet, Goerli, Gnosis)

TODO:
- TODOs in code
- Skip creating endpoints for non-supported networks
- Actually make all values in all fork builders non-primitive and add a tests verifying this
- Add primitive fields check test for all builders

I recommend to check only commmit https://github.com/ConsenSys/teku/commit/6f8feb3f16d9f473d118bd49e316e9d91331ad27  , especially `builder` folder to understand what's made, other code are fixes of the things that started to fail after these changes.
## Documentation

## Changelog


StefanBratanov:
I guess we can close this one?

zilm13:
@StefanBratanov yeah closing this, I think it will be too difficult to made and not safer for main functionality

PR #7283, opened 2023-06-20
Version 2 of SpecConfig changes: Fill 0 values in SpecConfigs intentionally

zilm13:
## PR Description
Allows to perform strict fields checks for all builders for non-FAR_FUTURE_EPOCH forks. Filling 0 values when Fork is in FAR_FUTURE_EPOCH, so it will pass validation and fork `build()` will not fail. I think we should start with this one as it's safe to merge and apply Version 1 later. Maybe some hybrid approach for Version 1, starting only since Capella, as it will be difficult to not run pre-Capella things in unsupported networks and as actually all current networks support Bellatrix.

Pros:
- Less risk to merge comparing to Version 1 (#7282)

Cons:
- Not so clean as Version 1 (#7282)
- Keeping the debt of using non-supported things with some strange values

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7284, opened 2023-06-21
Use progressive balances during epoch transition

### LINKED ISSUE #6998: Remove progressive balance modes

rolfyone:
Also cleaned up some of the processing logic, but only very small tweak.

Basically it's cheap to check if a delta is zero, and if it's zero there's no need to adjust balance. This will occur for every exited validator that has 0 balance, and any exited validator outside of any slashing period, so it's actually about 15% of the validator set, at which point it becomes quite worthwhile to check if the delta is zero, as we save a read+write to the balance set.

This is not a huge speed improvement, but it does round out the progressive balances change.

Fixes #6998

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AggregatedRewardAndPenalty.java
```diff
@@ -57,6 +57,11 @@ public UInt64 getPenalty() {
     return penalty;
   }
 
+  @Override
+  public boolean isZero() {
```
tbenr: I don't think we actually need this override. Leaving only the default is equivalent.

Also a very tiny improvement would be to check getRewards first, 'cose most of the time penalties are 0 and rewards are > 0

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
```diff
@@ -101,9 +103,8 @@ protected void processEpoch(final BeaconState preState, final MutableBeaconState
     final TransitionCaches transitionCaches = BeaconStateCache.getTransitionCaches(state);
     final ProgressiveTotalBalancesUpdates progressiveTotalBalances =
         transitionCaches.getProgressiveTotalBalances();
-    progressiveTotalBalances.checkResult(specConfig, state.getSlot(), totalBalances);
```
tbenr, line 104: We could remove `checkResult` method completely

tbenr (approved):
LGTM, just minor things

PR #7289, opened 2023-06-22
Increased the pending pool size for attestations

rolfyone:
This pool is used when an attestation is seen for a block we're not yet aware of.

The two main cases this is going to be useful is
 - when we're falling behind and listening to all subnets
 - if clients are attesting on receipt of block

 The pool size, worst case, would need to handle the number of attestations in a slot (all subnets would have to be enabled)

 With 32 slots, there's 32 segments of the validator count, so at 1 million validators, that'd be 31_250 attestations per slot.

 An upcoming testnet will be larger again, will have to re-evaluate if we're hitting the limit at that point, but we're currently hitting the 5k size of this pool on some nodes several times an hour.

 Having some head room on mainnet for a while should give us a good idea of sizing this cache going forward, and it may need to be periodically tweaked.

## Documentation

## Changelog


tbenr (approved):
LGTM. lets keep this monitored

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/PoolFactory.java
```diff
@@ -30,7 +30,10 @@
 public class PoolFactory {
 
   private static final UInt64 DEFAULT_HISTORICAL_SLOT_TOLERANCE = UInt64.valueOf(320);
-  private static final int DEFAULT_MAX_ITEMS = 5000;
+  // should fit attestations for a slot given validator set size
+  // so DEFAULT_MAX_ITEMS * slots_per_epoch should be >= validator set size ideally
+  // on all subnets, you may receive and have to cache that number of messages
+  private static final int DEFAULT_MAX_ITEMS = 30_000;
```
tbenr: I'd be tempted to start naming it referencing `ATTESTATIONS` since there are some reasoning strictly related to them

PR #7293, opened 2023-06-22
Employ one milestone one method approach for engine api calls

### LINKED ISSUE #7287: Employ one method one structure approach for engine_newPayloadV3
https://github.com/ethereum/execution-apis/pull/418

As part of this change can just change the whole `ExecutionClientHandler` to be milestone based.

StefanBratanov:
## PR Description

- Make `engine_forkchoiceUpdated` , `engine_newPayload` and `engine_getPayload` milestone based. (look at `MilestoneBasedExecutionJsonRpcMethodsResolver` and `ExecutionClientHandlerImpl`)
- Added `getMilestone` to `ExecutionPayload`

Essentially Bellatrix will call V1, Capella will call V2 and Deneb will call V3

- Removed `EngineApiCapabilitiesProvider` and its implementation since no longer needed
- Removed `--exchange-capabilities-enabled` option since it is no longer relevant

**TODO:** raise another issue to make use of `engine_exchangeCapabilitiesEnabled` to monitor and raise warnings if a required method is not supported for a given fork.

## Fixed Issue(s)
fixes #7287 

## Documentation

## Changelog


tbenr:
I wander if we should more strongly couple the versions in our code for past milestone too..
like https://github.com/Consensys/teku/pull/7319/commits/58148d7d21a5fba9e71d5eb23418dfbb8f2593e4

StefanBratanov:
> I wander if we should more strongly couple the versions in our code for past milestone too.. like [58148d7](https://github.com/Consensys/teku/commit/58148d7d21a5fba9e71d5eb23418dfbb8f2593e4)

Yeah we can. I was trying to more closely follow the spec, which does follow this only for V3 methods.

tbenr:
Ye, but we practically won't do that anymore.. and we can remove some tests..

StefanBratanov:
> Ye, but we practically won't do that anymore.. and we can remove some tests..

Could do that. I agree.

StefanBratanov:
> > Ye, but we practically won't do that anymore.. and we can remove some tests..
> 
> Could do that. I agree. Will apply your changes.

ethereum/executionclient/src/test/java/tech/pegasys/teku/ethereum/executionclient/methods/EngineNewPayloadV3Test.java
```diff
@@ -76,90 +74,43 @@ public void executionPayloadParamIsRequired() {
...
     final String errorResponseFromClient = "error!";
 
     when(executionEngineClient.newPayloadV3(any(), any()))
         .thenReturn(dummyFailedResponse(errorResponseFromClient));
 
     final JsonRpcRequestParams params =
-        new JsonRpcRequestParams.Builder().add(executionPayload).build();
+        new JsonRpcRequestParams.Builder()
+            .add(executionPayload)
+            .addOptional(Optional.of(blobVersionedHashes))
```
tbenr: shouldn't be `.add(blobVersionedHashes)` ?

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionClientHandlerImpl.java
```diff
@@ -66,7 +70,10 @@ public SafeFuture<ForkChoiceUpdatedResult> engineForkChoiceUpdated(
             .build();
 
     return methodsResolver
-        .getMethod(EngineApiMethods.ENGINE_FORK_CHOICE_UPDATED, ForkChoiceUpdatedResult.class)
+        .getMilestoneMethod(
+            EngineApiMethod.ENGINE_FORK_CHOICE_UPDATED,
+            () -> spec.atSlot(forkChoiceState.getHeadBlockSlot()).getMilestone(),
```
tbenr: I think should use `PayloadBuildingAttributes.blockSlot` if present, otherwise this slot. Because if we have the first slot of the new milestone missed, this will break if we are about to produce a block for the upcoming milestone (the V2 is actually due to the V2 on the payload attributes param)

tbenr (commented):
Seems good but i'd update the slot selection in the FCU call (and update tests accordingly)

ethereum/executionclient/src/test/java/tech/pegasys/teku/ethereum/executionclient/methods/EngineNewPayloadV3Test.java
StefanBratanov: Yes. Will fix.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionClientHandlerImpl.java
StefanBratanov: That makes sense. Will change tomorrow.

tbenr (approved):
LGTM

PR #7294, opened 2023-06-22
Refactor SpecConfig, detaching NetworkingSpecConfig with network constants

zilm13:
## PR Description
This is a beginning of splitting #7250 to the smaller PRs.

This PR detaches current networking constants in the interface `NetworkingSpecConfig`. There are currently only 3 such constants but future PRs will increase this number to around 15 so I decided to separate them in one dedicated domain. Some notes about this type of constants: they doesn't change on forks, all of them are available from the genesis and no constant names are ever reused in preset settings. If something should be changed, new constant name is used. So it's safe to always query them from the genesis.

Start review from classes: `NetworkingSpecConfig`, `SpecConfig`, `Spec`. Everything else are one line refactoring changes of using `NetworkingSpecConfig` instead of `SpecConfig`.

## Documentation

## Changelog


tbenr (approved):
LGTM. Nice cleanup

Issue #6788, opened 2023-02-08
Revisit rate limiting of blobs sidecars


StefanBratanov:
There are issues while syncing related to rate limiting of blobs sidecars. It is related to #6767 

- We rate limit "in advance" on `wantToReceiveBlobsSidecars` so we do not consider if there are empty blobs sidecar in the batch. To be evaluated, the PR in spec says "MAY".
- Related to the first point. we may want to consider applying the rate limit only from the responses we get from the peer rather than in advance.
- Revisit the different `RateTrackers` for blocks and blobs. Do we want one or multiple, change the number of accepted requests, etc.

PR #7295, opened 2023-06-22
rate limit based on the served objects instead of the requested objects count

mehdi-aouadi:
## PR Description

Update objects requests rate limiting based on the served objects count instead of the requested objects count:
`BeaconBlockByRange`, `BeaconBlockByRoot`, `BlobSidecarsByRange` and `BlobSidecarsByRoot` are all rate limited based on the requested objects count (The rate limiter is triggered when the request is received). In most case the real served objects count is lower (For blocks request, some blocks could not be returned and BlobSidecars we estimate the requested objects count by assuming that every block will contains `MAX_BLOB_SIDECARS_BY_BLOCK` Blob Sidecars which is not always/rarely the case).
The proposed solution is to add an object count adjustment to the rate limiter:
- `popBlockRequests` and `popBlobSidecarsRequests` will update the rate limiter with the max estimated objects counts.
- `adjustBlockRequests` and `adjustBlobSidecarsRequests` will adjust the initially estimated objects count and replace it with the served objects count (only if the adjustment happens during the same rate limiting time window).
- `cancelBlockRequests` and `cancelBlobSidecarRequests` will reset the initially estimated objects if the request is cancelled for some reason (mainly errors when serving the requests)

## Fixed Issue(s)

#6788 

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
@@ -24,6 +24,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
@@ -340,31 +341,31 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
   }
 
   @Override
-  public void popBlockRequests(final long blocksCount) {
-    popObjectRequests(blockRequestTracker, blocksCount);
-  }
-
-  @Override
-  public boolean wantToRequestBlocks(
+  public Pair<UInt64, Boolean> popBlockRequests(
       final ResponseCallback<SignedBeaconBlock> callback, long blocksCount) {
-    return wantToRequestObjects("blocks", blockRequestTracker, blocksCount, callback);
+    return popObjectRequests("blocks", blockRequestTracker, blocksCount, callback);
   }
 
   @Override
-  public void popBlobSidecarRequests(final long blobSidecarsCount) {
-    popObjectRequests(blobSidecarsRequestTracker, blobSidecarsCount);
+  public void adjustBlockRequests(final long blocksCount, final UInt64 time) {
+    adjustObjectRequests(blockRequestTracker, blocksCount, time);
   }
 
   @Override
-  public boolean wantToRequestBlobSidecars(
+  public Pair<UInt64, Boolean> popBlobSidecarRequests(
       final ResponseCallback<BlobSidecar> callback, long blobSidecarsCount) {
-    return wantToRequestObjects(
+    return popObjectRequests(
         "blob sidecars", blobSidecarsRequestTracker, blobSidecarsCount, callback);
   }
 
   @Override
+  public void adjustBlobSidecarRequests(final long blobSidecarsCount, final UInt64 time) {
+    adjustObjectRequests(blobSidecarsRequestTracker, blobSidecarsCount, time);
+  }
+
+  @Override
   public boolean popRequest() {
-    if (requestTracker.popObjectRequests(1L) == 0L) {
+    if (requestTracker.popObjectRequests(1L).getRight() == 0L) {
       LOG.debug("Peer {} disconnected due to request rate limits", getId());
       disconnectCleanly(DisconnectReason.RATE_LIMITING).ifExceptionGetsHereRaiseABug();
       return false;
@@ -395,23 +396,25 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
         .thenCompose(__ -> responseHandler.getResult());
   }
 
-  private void popObjectRequests(final RateTracker requestTracker, final long objectCount) {
-    requestTracker.popObjectRequests(objectCount);
+  private void adjustObjectRequests(
+      final RateTracker requestTracker, final long objectCount, final UInt64 time) {
+    requestTracker.adjustRequestObjects(objectCount, time);
   }
 
-  private <T> boolean wantToRequestObjects(
+  private <T> Pair<UInt64, Boolean> popObjectRequests(
       final String requestType,
       final RateTracker requestTracker,
       final long objectCount,
       final ResponseCallback<T> callback) {
-    if (requestTracker.wantToRequestObjects(objectCount) == 0L) {
+    Pair<UInt64, Long> requestTrackerResponse = requestTracker.popObjectRequests(objectCount);
+    if (requestTrackerResponse.getRight() == 0L) {
       LOG.debug("Peer {} disconnected due to {} rate limits", getId(), requestType);
       callback.completeWithErrorResponse(
           new RpcException(INVALID_REQUEST_CODE, "Peer has been rate limited"));
       disconnectCleanly(DisconnectReason.RATE_LIMITING).ifExceptionGetsHereRaiseABug();
-      return false;
+      return Pair.of(requestTrackerResponse.getLeft(), false);
     }
-    return true;
+    return Pair.of(requestTrackerResponse.getLeft(), true);
   }
 
   private <I extends RpcRequest, O extends SszData> SafeFuture<Optional<O>> requestOptionalItem(
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
@@ -353,6 +353,11 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
   }
 
   @Override
+  public void cancelBlockRequests(final long initialBlockCount, final UInt64 time) {
+    adjustObjectRequests(blockRequestTracker, initialBlockCount, time);
+  }
+
+  @Override
   public Pair<UInt64, Boolean> popBlobSidecarRequests(
       final ResponseCallback<BlobSidecar> callback, long blobSidecarsCount) {
     return popObjectRequests(
@@ -364,6 +369,11 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
       final long returnedBlobSidecarsCount, final long initialBlobSidecarCount, final UInt64 time) {
     adjustObjectRequests(
         blobSidecarsRequestTracker, returnedBlobSidecarsCount, initialBlobSidecarCount, time);
+  }
+
+  @Override
+  public void cancelBlobSidecarRequests(final long initialBlobSidecarCount, final UInt64 time) {
+    adjustObjectRequests(blobSidecarsRequestTracker, initialBlobSidecarCount, time);
   }
 
   @Override
@@ -405,6 +415,11 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
       final long initialObjectCount,
       final UInt64 time) {
     requestTracker.adjustRequestObjects(returnedObjectCount, initialObjectCount, time);
+  }
+
+  private void adjustObjectRequests(
+      final RateTracker requestTracker, final long initialObjectCount, final UInt64 time) {
+    requestTracker.adjustRequestObjects(0, initialObjectCount, time);
   }
 
   private <T> Pair<UInt64, Boolean> popObjectRequests(
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
@@ -340,34 +340,34 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
   }
 
   @Override
-  public Optional<RequestApproval> popBlockRequests(
+  public Optional<RequestApproval> approveBlocksRequest(
       final ResponseCallback<SignedBeaconBlock> callback, long blocksCount) {
-    return popObjectRequests("blocks", blockRequestTracker, blocksCount, callback);
+    return approveObjectsRequest("blocks", blockRequestTracker, blocksCount, callback);
   }
 
   @Override
-  public void adjustBlockRequests(
+  public void adjustBlocksRequest(
       final RequestApproval blocksRequest, final long returnedBlocksCount) {
-    adjustObjectRequests(blockRequestTracker, blocksRequest, returnedBlocksCount);
+    adjustObjectsRequest(blockRequestTracker, blocksRequest, returnedBlocksCount);
   }
 
   @Override
-  public Optional<RequestApproval> popBlobSidecarRequests(
+  public Optional<RequestApproval> approveBlobSidecarsRequest(
       final ResponseCallback<BlobSidecar> callback, long blobSidecarsCount) {
-    return popObjectRequests(
+    return approveObjectsRequest(
         "blob sidecars", blobSidecarsRequestTracker, blobSidecarsCount, callback);
   }
 
   @Override
-  public void adjustBlobSidecarRequests(
+  public void adjustBlobSidecarsRequest(
       final RequestApproval blobSidecarsRequest, final long returnedBlobSidecarsCount) {
-    adjustObjectRequests(
+    adjustObjectsRequest(
         blobSidecarsRequestTracker, blobSidecarsRequest, returnedBlobSidecarsCount);
   }
 
   @Override
-  public boolean popRequest() {
-    if (requestTracker.popObjectRequests(1L).isEmpty()) {
+  public boolean approveRequest() {
+    if (requestTracker.approveObjectsRequest(1L).isEmpty()) {
       LOG.debug("Peer {} disconnected due to request rate limits", getId());
       disconnectCleanly(DisconnectReason.RATE_LIMITING).ifExceptionGetsHereRaiseABug();
       return false;
@@ -398,26 +398,27 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
         .thenCompose(__ -> responseHandler.getResult());
   }
 
-  private void adjustObjectRequests(
+  private void adjustObjectsRequest(
       final RateTracker requestTracker,
       final RequestApproval requestApproval,
       final long returnedObjectsCount) {
-    requestTracker.adjustObjectRequests(requestApproval, returnedObjectsCount);
+    requestTracker.adjustObjectsRequest(requestApproval, returnedObjectsCount);
   }
 
-  private <T> Optional<RequestApproval> popObjectRequests(
+  private <T> Optional<RequestApproval> approveObjectsRequest(
       final String requestType,
       final RateTracker requestTracker,
-      final long objectCount,
+      final long objectsCount,
       final ResponseCallback<T> callback) {
-    final Optional<RequestApproval> objectsRequest = requestTracker.popObjectRequests(objectCount);
-    if (objectsRequest.isEmpty()) {
+    final Optional<RequestApproval> requestApproval =
+        requestTracker.approveObjectsRequest(objectsCount);
+    if (requestApproval.isEmpty()) {
       LOG.debug("Peer {} disconnected due to {} rate limits", getId(), requestType);
       callback.completeWithErrorResponse(
           new RpcException(INVALID_REQUEST_CODE, "Peer has been rate limited"));
       disconnectCleanly(DisconnectReason.RATE_LIMITING).ifExceptionGetsHereRaiseABug();
     }
-    return objectsRequest;
+    return requestApproval;
   }
 
   private <I extends RpcRequest, O extends SszData> SafeFuture<Optional<O>> requestOptionalItem(
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
@@ -410,7 +410,7 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
       final RateTracker requestTracker,
       final long objectCount,
       final ResponseCallback<T> callback) {
-    Optional<RequestApproval> objectsRequest = requestTracker.popObjectRequests(objectCount);
+    final Optional<RequestApproval> objectsRequest = requestTracker.popObjectRequests(objectCount);
     if (objectsRequest.isEmpty()) {
       LOG.debug("Peer {} disconnected due to {} rate limits", getId(), requestType);
       callback.completeWithErrorResponse(
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
@@ -347,8 +347,9 @@ class DefaultEth2Peer extends DelegatingPeer implements Eth2Peer {
   }
 
   @Override
-  public void adjustBlockRequests(final long blocksCount, final UInt64 time) {
-    adjustObjectRequests(blockRequestTracker, blocksCount, time);
+  public void adjustBlockRequests(
+      final long returnedBlocksCount, final long initialBlockCount, final UInt64 time) {
+    adjustObjectRequests(blockRequestTracker, returnedBlocksCount, initialBlockCount, time);
   }
 
   @Override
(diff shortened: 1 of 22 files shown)


zilm13:
My first feeling on this: time especially in seconds couldn't be unique identifier, so I'd change
`Pair<UInt64, Long> popObjectRequests(long objectCount)`
to
`Optional<UUID> popObjectRequests(objectCount)`
if it's provided, we are allowed, and we next pass this uuid along with updated count in
`adjustRequestObjects(long returnedObjectCount, long initialObjectCount, UUID requestId)`

For UUID we could use `AtomicInteger` counter in `RateTracker`, we probably don't need real UUID. But I'd still wrap it in some `RequestId` class for clarity.
`NavigableMap<UInt64, Long> requestCount` is changed to `NavigableMap<UInt64, Map<UUID, count>> requestCount`
Something like this. 
Ok, I missed search by UUID. We could either make 2 total maps to handle this or decide which one is used more often. Also we could make combined key in this map, if we want maximum optimization. But as everything is synchronized there, 2 maps looks like the simplest approach

Also I think we need to always toggle some number, say 2 or 5 to avoid DoSing with empty requests. Or did I miss total request tracking? There was one but I cannot find it now

What do you think on this?

mehdi-aouadi:
> My first feeling on this: time especially in seconds couldn't be unique identifier, so I'd change `Pair<UInt64, Long> popObjectRequests(long objectCount)` to `Optional<UUID> popObjectRequests(objectCount)` if it's provided, we are allowed, and we next pass this uuid along with updated count in `adjustRequestObjects(long returnedObjectCount, long initialObjectCount, UUID requestId)`
> 
> For UUID we could use `AtomicInteger` counter in `RateTracker`, we probably don't need real UUID. But I'd still wrap it in some `RequestId` class for clarity. `NavigableMap<UInt64, Long> requestCount` is changed to `NavigableMap<UInt64, Map<UUID, count>> requestCount` Something like this. Ok, I missed search by UUID. We could either make 2 total maps to handle this or decide which one is used more often. Also we could make combined key in this map, if we want maximum optimization. But as everything is synchronized there, 2 maps looks like the simplest approach
> 
> Also I think we need to always toggle some number, say 2 or 5 to avoid DoSing with empty requests. Or did I miss total request tracking? There was one but I cannot find it now
> 
> What do you think on this?

In the current implementation, time is not used as a unique identifier but as "pointer" to a time window. When different requests happen during the same second, the object count is summed up and hence the need of the initial count when adjusting in order to know how much to subtract and how much to add instead.

> and we next pass this uuid along with updated count in
> adjustRequestObjects(long returnedObjectCount, long initialObjectCount, UUID requestId)

If we use a unique identifier, there will be no need to send the initial request object count during the adjustment since the entry is unique and we have the uuid: we only need to update the existing value with the new one. We could simply do `adjustRequestObjects(long objectCount, UUID requestId)`

> NavigableMap<UInt64, Long> requestCount is changed to NavigableMap<UInt64, Map<UUID, count>> requestCount

The benefit of this (and the whole unique id suggestion) is to directly access/update the concerned entry instead of retrieving the object count sum per second, subtract the old value and add the new real count. Is that how you see it?
I want to make sure that you don't miss any detail because I'd prefer to avoid adding complexity here (uuid plus Map of a Map).
I suspected that you missed the object count sum up logic when I saw that you suggested to use `adjustRequestObjects(long returnedObjectCount, long initialObjectCount, UUID requestId)` still with the `initialObjectCount`

zilm13:
My miss, I see a sum now.
So what I don't like:
- Pair return 
- Usage of time reference in client class

So I see 2 approaches:
1. The one I already suggested and we could skip requestObjects in this case I see
2. Wrap this time in something like `RequestSlot`, I don't like the word `slot`, confusing with `BeaconChain` slot, maybe another word... but the idea is to pass the same thing but wrap it to isolate logic from the client. So you change return type to `Optional<RequestSlot>` which stores the same value. When it's provided, you are allowed to pop. And it doesn't look like some strange Uint64 passing to client. You could keep almost the same code but it will be more elegant on the client side.

mehdi-aouadi:
@zilm13 Here is a summary of what I changed:
- Removed the `Pair` usage: Replaced it with an `Optional<ObjectsRequestResponse>` which is empty is the request couldn't be fulfilled and contains the requests detail otherwise
- Wrapped the request details in `ObjectsRequestResponse` which contains the time in seconds, the objects count and a `UUID`
- Used a `UUID` as id for the requests. This could be replaced by an `AtomicInteger` if required (I guess we won't have that much requests per second). I used the `UUID` because I find it easier
- Updated the `RateTracker` data structure from `NavigableMap<UInt64, Long>` to `NavigableMap<ObjectRequestsEntryKey, Long>` (`ObjectRequestsEntryKey` is a combination of the `UUID` and the time in seconds). I think this is more readable and performant compared to using 2 separate maps.
- The  `popObjectRequests` now returns an `Optional<ObjectsRequestResponse>` which can be used later to adjust the counter by calling `adjustObjectRequests`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +39,108 @@
...
-  public synchronized long popObjectRequests(final long objectCount) {
+  public synchronized Optional<ObjectsRequestResponse> popObjectRequests(final long objectCount) {
     pruneRequests();
+    final UInt64 currentTime = timeProvider.getTimeInSeconds();
     if ((peerRateLimit - requestsWithinWindow) <= 0) {
-      return 0L;
+      return Optional.empty();
     }
 
     requestsWithinWindow += objectCount;
```
zilm13: I'd call it `objectsWithinWindow` for clarity

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +39,108 @@
...
+      return Objects.hash(timeSeconds, requestId);
+    }
+
+    @Override
+    public int compareTo(@NotNull ObjectRequestsEntryKey other) {
+      return timeSeconds.compareTo(other.timeSeconds);
+    }
+  }
+
+  public static class ObjectsRequestResponse {
```
zilm13: I think we'd better move it as separate class, so usage will be more elegant without outer path. Also let's find better name. How about `RequestApproval`?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +39,108 @@
...
+    public UInt64 getTimeSeconds() {
+      return timeSeconds;
+    }
+
+    public long getObjectsCount() {
+      return objectsCount;
+    }
+
+    public static final class ObjectsRequestBuilder {
+      private UInt64 timeSeconds = UInt64.ZERO;
```
zilm13: Let's remove default value, it would be saferr and could be easily caught in test

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +39,108 @@
...
-        requestCount.headMap(currentTime.minus(timeoutSeconds), false);
+    final NavigableMap<ObjectRequestsEntryKey, Long> headMap =
+        requestCount.headMap(
+            new ObjectRequestsEntryKey(currentTime.minus(timeoutSeconds), UUID.randomUUID()),
+            false);
     headMap.values().forEach(value -> requestsWithinWindow -= value);
     headMap.clear();
   }
+
+  public static class ObjectRequestsEntryKey implements Comparable<ObjectRequestsEntryKey> {
```
zilm13: Github scan looks helpful

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +39,108 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(timeSeconds, requestId);
+    }
+
+    @Override
+    public int compareTo(@NotNull ObjectRequestsEntryKey other) {
+      return timeSeconds.compareTo(other.timeSeconds);
```
zilm13: You forgot requestId which means the two keys with different requestId is the same, which is not true. How about something like this? `Comparator.comparing(ObjectRequestsEntryKey::getTimeSeconds).thenComparing(ObjectRequestsEntryKey::getRequestId).compare(this, other);
`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +39,108 @@
...
   void pruneRequests() {
     final UInt64 currentTime = timeProvider.getTimeInSeconds();
     if (currentTime.isLessThan(timeoutSeconds)) {
       return;
     }
-    final NavigableMap<UInt64, Long> headMap =
-        requestCount.headMap(currentTime.minus(timeoutSeconds), false);
+    final NavigableMap<ObjectRequestsEntryKey, Long> headMap =
+        requestCount.headMap(
+            new ObjectRequestsEntryKey(currentTime.minus(timeoutSeconds), UUID.randomUUID()),
```
zilm13: And here's the issue taking UUID. We don't know what is minimal uuid.
Let's use AtomicInteger instead.
So add `private final AtomicInteger newRequestId = new AtomicInteger(0)` field to RateTracker, add `requestId(int requrstId)` to builder and use it with `.requestId(newRequestId.getAndIncrement())` when building.
After that init key here with (currentime -...,, 0)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +39,108 @@
...
+
+    @Override
+    public int compareTo(@NotNull ObjectRequestsEntryKey other) {
+      return timeSeconds.compareTo(other.timeSeconds);
+    }
+  }
+
+  public static class ObjectsRequestResponse {
+
+    private final UUID requestId;
```
zilm13: How about storing a  key here for better interface?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
```diff
@@ -121,17 +123,27 @@ public void onIncomingMessage(
...
-    totalBlocksRequestedCounter.inc(message.getCount().longValue());
+
     sendMatchingBlocks(message, callback)
         .finish(
-            callback::completeSuccessfully,
+            requestState -> {
+              if (requestState.sentBlocks.get() != message.getCount().longValue()) {
+                peer.adjustBlockRequests(blockRequests.get(), requestState.sentBlocks.get());
+              }
+              totalBlocksRequestedCounter.inc(requestState.sentBlocks.get());
```
zilm13: `totalBlocksRequestedCounter`
I maybe wrong but from the name it sounds like it's counting the number of requested only , not adjusted

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
```diff
@@ -121,17 +123,27 @@ public void onIncomingMessage(
...
+              callback.completeSuccessfully();
+            },
             error -> {
+              peer.adjustBlockRequests(blockRequests.get(), 0);
```
zilm13: I don't think it's true, the stream could be broken anywhere. It doesn't look that anyone could exploit it but anyway...Just thoughts aloud

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
```diff
@@ -112,10 +120,27 @@ public void onIncomingMessage(
...
+                                              .respond(signedBeaconBlock)
+                                              .thenRun(sentBlocks::incrementAndGet))
+                                  .orElse(SafeFuture.COMPLETE);
                             }));
       }
-      future.finish(callback::completeSuccessfully, err -> handleError(callback, err));
+      future.finish(
+          () -> {
+            if (sentBlocks.get() != message.size()) {
+              peer.adjustBlockRequests(blockRequests.get(), sentBlocks.get());
```
zilm13: maybe `blocksRequest`?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java
```diff
@@ -107,16 +109,20 @@ public void onIncomingMessage(
...
       requestCounter.labels("rate_limited").inc();
       return;
     }
 
     requestCounter.labels("ok").inc();
-    totalBlobSidecarsRequestedCounter.inc(message.size());
 
     SafeFuture<Void> future = SafeFuture.COMPLETE;
 
+    final AtomicInteger sentBlobSidecars = new AtomicInteger(0);
```
zilm13, line 127: I'd compact these 4 lines (remove some newlines, even -1 would be better :)

zilm13 (commented):
Looks good and solves the issue in a pretty good way. So let's improve several things before the merge.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Done, renamed it to `RequestApproval`, sounds OK to me

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Yep! My bad forgot to override `equals`. Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Actually I didn't really forget, since we use `headMap` to retrieve sub maps based on time only, I explicitly ignored the request ids... Added it now (even though it's not useful for the current implementation, that wouldn't be aligned with `compareTo` definition)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Done. Kept the `ObjectRequestsEntryKey` class within the `RateTracker` class though. We could extract if needed...

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
mehdi-aouadi: Done: renamed them to `blockRequestsApproval` and `blobSidecarRequestsApproval`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java
mehdi-aouadi: Done, removed the new lines if that's what you meant 😄

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Done: Used an `AtomicInteger` to generate request ids instead. Even though when calling `headMap` we only need a sub map based on the time and not the request id. It's better now anyway.
The id generator reset happens when we try to pop requests and we have reached a new time in seconds

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
mehdi-aouadi: Moved these counters back before the request processing. Yeah these look like requests count metrics (whether they succeed or not)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
mehdi-aouadi: Yeah, I thought at least we could adjust the counter when we know about the error...

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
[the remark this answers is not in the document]
mehdi-aouadi: Fixed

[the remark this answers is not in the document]
mehdi-aouadi: Fixed

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
zilm13: I mean it's a single request for several blocks, so naming for these and similar methods confuses

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +42,94 @@ public RateTracker(
...
+  public synchronized Optional<RequestApproval> popObjectRequests(final long objectCount) {
     pruneRequests();
-    if ((peerRateLimit - requestsWithinWindow) <= 0) {
-      return 0L;
+    final UInt64 currentTime = timeProvider.getTimeInSeconds();
+    if ((peerRateLimit - objectsWithinWindow) <= 0) {
+      return Optional.empty();
     }
+    objectsWithinWindow += objectCount;
+    resetRequestId(currentTime);
```
zilm13: I think we don't need this, we have pretty big margin to be sure we are not running out of numbers, especially as tracker is tied to a single connection

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
```diff
@@ -35,27 +42,94 @@ public RateTracker(
...
-        requestCount.headMap(currentTime.minus(timeoutSeconds), false);
-    headMap.values().forEach(value -> requestsWithinWindow -= value);
+    final NavigableMap<ObjectRequestsEntryKey, Long> headMap =
+        requestCount.headMap(
+            new ObjectRequestsEntryKey(currentTime.minus(timeoutSeconds), 0), false);
+    headMap.values().forEach(value -> objectsWithinWindow -= value);
     headMap.clear();
   }
+
+  public static class ObjectRequestsEntryKey implements Comparable<ObjectRequestsEntryKey> {
```
zilm13: Maybe RequestKey? Or something other more elegant

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: At the end I renamed the `ObjectRequestsEntryKey` to `ObjectRequestsKey` and moved it to a separate class

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
```diff
@@ -365,19 +378,26 @@ public <I extends RpcRequest, O extends SszData> SafeFuture<O> requestSingleItem
...
+
+  private <T> Optional<RequestApproval> popObjectRequests(
       final String requestType,
       final RateTracker requestTracker,
-      final ResponseCallback<T> callback,
-      final long objectCount) {
-    if (requestTracker.popObjectRequests(objectCount) == 0L) {
+      final long objectCount,
+      final ResponseCallback<T> callback) {
+    Optional<RequestApproval> objectsRequest = requestTracker.popObjectRequests(objectCount);
```
zilm13: final

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
```diff
@@ -121,7 +123,10 @@ public void onIncomingMessage(
         message.getStartSlot(),
         message.getStep());
 
-    if (!peer.popRequest() || !peer.popBlockRequests(callback, message.getCount().longValue())) {
+    final Optional<RequestApproval> blockRequestsApproval =
```
zilm13: I think it's a single request for several blocks, so such names confuse

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
```diff
@@ -112,10 +120,27 @@ public void onIncomingMessage(
...
+                                  .orElse(SafeFuture.COMPLETE);
                             }));
       }
-      future.finish(callback::completeSuccessfully, err -> handleError(callback, err));
+      future.finish(
+          () -> {
+            if (sentBlocks.get() != message.size()) {
+              peer.adjustBlockRequests(blockRequestsApproval.get(), sentBlocks.get());
+            }
+            totalBlocksRequestedCounter.inc(sentBlocks.get());
```
zilm13: You forgot to move counter here outside to requested, not sent values

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Done. Renamed it to `RequestKey`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRootMessageHandler.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/RateTracker.java
mehdi-aouadi: Hmm, yes I see the rate tracker is initialised when `onConnect`. We could avoid resetting it

mehdi-aouadi: Done

zilm13 (approved):
LGTM

PR #7300, opened 2023-06-23
Add networking subnet constants from the spec to NetworkingSpecConfig

zilm13:
## PR Description
The part of the #7250 
Adding subnet constants. 
I'm not sure about `GetSpecResponse`. Whether we should pass new constants through GetSpecResponse or load them from local config like it was done for `MAX_GOSSIP_SIZE` and `MAX_CHUNK_SIZE`.
Start review from `NetworkingSpecConfig`. Everything else is using the values from it instead of `Constants`

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigReader.java
```diff
@@ -61,6 +61,8 @@ public class SpecConfigReader {
           // Unsupported, upcoming fork-related keys
           "SHARDING_FORK_VERSION",
           "SHARDING_FORK_EPOCH",
+          "EIP6110_FORK_VERSION",
```
tbenr, line 64: do we need `WHISK_FORK_*` too?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
```diff
@@ -68,7 +68,7 @@ public SchemaDefinitionsAltair(final SpecConfigAltair specConfig) {
         ContributionAndProofSchema.create(syncCommitteeContributionSchema);
     this.signedContributionAndProofSchema =
         SignedContributionAndProofSchema.create(contributionAndProofSchema);
-    this.metadataMessageSchema = new MetadataMessageSchemaAltair();
+    this.metadataMessageSchema = new MetadataMessageSchemaAltair(specConfig.getNetworkingConfig());
```
tbenr, line 71: since `specConfig` _IS_ networkingConfig too, there will be some inconsistencies here and there when accessing the configs or one of those fields. While I get the nice separation, there will always be inconsistency using it until we actually separate them (if we ever will) and you'll be forced to use `getNetworkingConfig()` by then

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/minimal.yaml
```diff
@@ -46,9 +46,12 @@ BELLATRIX_FORK_EPOCH: 18446744073709551615
 # Capella
 CAPELLA_FORK_VERSION: 0x03000001
 CAPELLA_FORK_EPOCH: 18446744073709551615
-# Deneb
+# DENEB
 DENEB_FORK_VERSION: 0x04000001
 DENEB_FORK_EPOCH: 18446744073709551615
+# EIP6110
+EIP6110_FORK_VERSION: 0x05000001
+EIP6110_FORK_EPOCH: 18446744073709551615
```
tbenr, line 54: `# WHISK` ?

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/MetadataMessageTest.java
```diff
@@ -66,7 +67,7 @@ public void shouldRejectOutOfBoundsAttnets() {
   public void shouldRejectOutOfBoundsSyncnets() {
     assertThatThrownBy(
             () ->
-                ALTAIR_SCHEMA.create(
+                altairSchema.create(
```
tbenr, line 70: I wonder if SYNC_COMMITTEE_SUBNET_COUNT has been forgotten on spec side..

tbenr (approved):
LGTM. just some comments and a question

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigReader.java
zilm13: Added

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaDefinitionsAltair.java
zilm13: Yeah, I want some separation by domain, especially in some networking only packages. I don't think it's a big issue it's the same class, until we cast `NetworkingConfig` to `SpecConfig`. And also it improves some organizing, less things in one class, you could consider it like some implementation of mixin pattern

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/MetadataMessageTest.java
zilm13: Yeah, makes sense to add it consensus-specs, we could propose it later

Nashatyrev: Guys what do you mean by 'forgotten on spec side' ?
Looks like the constant is in the spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/validator.md#misc

I've just failed to change that constant for my minimal setup 😆

PR #7304, opened 2023-06-25
Add Deneb networking constants to spec config

zilm13:
## PR Description
Added deneb networking constants like in `consensus-specs`. I don't like a lot of `orElseThrow()` but didn't get better idea, may be you have good suggestion how to organize this.
## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
```diff
@@ -0,0 +1,36 @@
...
+package tech.pegasys.teku.spec.datastructures.networking.libp2p.rpc;
+
+import tech.pegasys.teku.infrastructure.ssz.schema.impl.AbstractSszListSchema;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNode;
+import tech.pegasys.teku.spec.config.SpecConfigDeneb;
+
+public class BlobSidecarsByRootRequestMessageSchema
+    extends AbstractSszListSchema<BlobIdentifier, BlobSidecarsByRootRequestMessage> {
+
+  // Size validation according to the spec (MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK) is
```
tbenr, line 23: I wander if now we can start using the actual `getMaxRequestBlobSidecars`. If i'm not mistaken, now those values are bonded to milestones, so if they will be changed it will be done via another fork, which will have its own schema definition with a new `BlobSidecarsByRootRequestMessageSchema` with updated size

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -197,11 +196,24 @@ public SpecConfig getGenesisSpecConfig() {
...
+   * <p>These constants are unified among forks and are not overriden, new constant name is used if
+   * it's changed in the new fork
+   */
   public NetworkingSpecConfig getNetworkingConfig() {
     // Networking config is constant along forks
     return getGenesisSpec().getConfig().getNetworkingConfig();
   }
 
+  /** Networking config with Deneb constants */
+  public Optional<NetworkingSpecConfigDeneb> getNetworkingConfigDeneb() {
```
tbenr, line 214: Even if this is looks bad, we have to deal with it. As mentioned in the previous method javadoc, there is no override. Fork specific code will refer to fork specific configs. At the moment this still holds, but maybe in the future we could have some ugly things where the same logic will need to switch config (it will be a de-facto config param override). ATM I wouldn't do anything different than this.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -884,7 +897,10 @@ public Optional<Integer> getMaxBlobsPerBlock(final UInt64 slot) {
   }
 
   public UInt64 computeSubnetForBlobSidecar(final SignedBlobSidecar signedBlobSidecar) {
-    return signedBlobSidecar.getBlobSidecar().getIndex().mod(BLOB_SIDECAR_SUBNET_COUNT);
+    return signedBlobSidecar
+        .getBlobSidecar()
+        .getIndex()
+        .mod(getNetworkingConfigDeneb().orElseThrow().getBlobSidecarSubnetCount());
```
tbenr: ahh.. here we are.
So, an idea could be to have  `getBlobSidecarSubnetCount` on the base `SpecConfig` and do the override-like game in SpecConfig.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
StefanBratanov, line 214: In my opinion if we always do `orElseThrow()` when using this we better make it not an optional, we can have `NetworkingSpecConfigDeneb.required()` in similar pattern we use for the SpecConfig and just throw exception if it's called.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
tbenr: Bu I don't actually know how the transition would be handled in our code... so maybe we should do something when we will have to

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/topics/GossipTopics.java
```diff
@@ -65,19 +64,19 @@ public static String getBlobSidecarSubnetTopic(
...
-    for (int i = 0; i < networkingConfig.getAttestationSubnetCount(); i++) {
+    for (int i = 0; i < spec.getNetworkingConfig().getAttestationSubnetCount(); i++) {
       topics.add(getAttestationSubnetTopic(forkDigest, i, gossipEncoding));
     }
     for (int i = 0; i < NetworkConstants.SYNC_COMMITTEE_SUBNET_COUNT; i++) {
       topics.add(getSyncCommitteeSubnetTopic(forkDigest, i, gossipEncoding));
     }
-    for (int i = 0; i < Constants.BLOB_SIDECAR_SUBNET_COUNT; i++) {
-      topics.add(getBlobSidecarSubnetTopic(forkDigest, i, gossipEncoding));
+    if (spec.getNetworkingConfigDeneb().isPresent()) {
```
StefanBratanov, line 76: We can use something like `spec.getNetworkingConfig().toVersionDeneb().ifPresent(networkingConfigDeneb ->.....)` potentially? if we make spec.getNetworkingConfigDeneb() not an optional

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -523,7 +522,10 @@ private Optional<UInt64> getEarliestAvailabilityWindowSlotBeforeBlock(
     }
     final UInt64 firstAvailabilityWindowSlot =
         spec.computeStartSlotAtEpoch(
-                currentEpoch.minusMinZero(MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS))
+                currentEpoch.minusMinZero(
```
tbenr: An idea to remove these spaghetti could be to move methods like this to ForkChoiceUtil  and we could start having multiple milestone-based implementations.

tbenr (commented):
just throwing some comments for the moment

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
zilm13: I don't want to do anything in `SpecConfig`, I've removed deneb networking constants getter from there because we cannot be sure there that Deneb is supported and could do some error

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
zilm13: Doesn't look very good passing spec there https://github.com/Consensys/teku/pull/7304/commits/c580aa7f3d94e77783b2bc412b7c942238757248

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/topics/GossipTopics.java
zilm13: We could refactor it in functional style but I don't think it looks less ugly
https://github.com/Consensys/teku/pull/7304/commits/da26990115c042cd349140040f5b326447ff58ce . I'd revert this

teku/src/main/java/tech/pegasys/teku/config/TekuConfiguration.java
```diff
@@ -232,10 +230,14 @@ public TekuConfiguration build() {
...
-              spec.getNetworkingConfig().getMaxRequestBlocks(),
-              MAX_REQUEST_BLOCKS_DENEB.intValue());
+      final long maxAllowedBatchSize =
+          spec.getNetworkingConfigDeneb()
+              .map(
+                  networkingSpecConfigDeneb ->
+                      Math.min(
+                          spec.getNetworkingConfig().getMaxRequestBlocks(),
+                          networkingSpecConfigDeneb.getMaxRequestBlocksDeneb()))
+              .orElse(spec.getNetworkingConfig().getMaxRequestBlocks());
```
StefanBratanov, line 237: I would just do. Implementation before was wrong anyways. It depends on the fork what maximum we allow, no need to do the min comparison.
```java
spec.getNetworkingConfigDeneb()
  .map(NetworkingSpecConfigDeneb::getMaxRequestBlocksDeneb)
  .orElse(spec.getNetworkingConfig().getMaxRequestBlocks());
```

zilm13: you are right, it will work

tbenr (approved):
LGTM

PR #7305, opened 2023-06-26
Increase max attestation inclusion slot (p2p changes)

### LINKED ISSUE #7299: Increase max attestation inclusion slot
https://github.com/ethereum/consensus-specs/pull/3360

StefanBratanov:
## PR Description

- Created `AttestationUtilDeneb` to deal with slot inclusion gossip validation changes for attestations in Deneb. Moved the previous slot inclusion validations in `AttestationUtilPhase0`
- Split `AttestationValidatorTest` into Deneb and Phase0 one
- Other small nits/refactors

[p2p-interface.md changes](https://github.com/ethereum/consensus-specs/pull/3360/files#diff-21cd1dde2a4815e714ef42d0cefa1bbd2997a4358ecf6e9e51025332c1d642fd)

## Fixed Issue(s)
fixes #7299 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttestationData.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttestationData.java
@@ -82,7 +82,7 @@ public class AttestationData
   public UInt64 getEarliestSlotForForkChoice(final UInt64 targetEpochStartSlot) {
     // Attestations can't be processed by fork choice until their slot is in the past and until we
     // are in the same epoch as their target.
-    return getSlot().plus(1).max(targetEpochStartSlot);
+    return getSlot().plus(UInt64.ONE).max(targetEpochStartSlot);
   }
 
   public UInt64 getSlot() {
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttestationData.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttestationData.java
@@ -70,15 +70,19 @@ public class AttestationData
   }
 
   public UInt64 getEarliestSlotForForkChoice(final Spec spec) {
-    // Attestations can't be processed by fork choice until their slot is in the past and until we
-    // are in the same epoch as their target.
-    return getSlot().plus(UInt64.ONE).max(getTarget().getEpochStartSlot(spec));
+    final UInt64 targetEpochStartSlot = getTarget().getEpochStartSlot(spec);
+    return getEarliestSlotForForkChoice(targetEpochStartSlot);
   }
 
   public UInt64 getEarliestSlotForForkChoice(final MiscHelpers miscHelpers) {
-    return getSlot()
-        .plus(UInt64.ONE)
-        .max(miscHelpers.computeStartSlotAtEpoch(getTarget().getEpoch()));
+    final UInt64 targetEpochStartSlot = miscHelpers.computeStartSlotAtEpoch(getTarget().getEpoch());
+    return getEarliestSlotForForkChoice(targetEpochStartSlot);
+  }
+
+  public UInt64 getEarliestSlotForForkChoice(final UInt64 targetEpochStartSlot) {
+    // Attestations can't be processed by fork choice until their slot is in the past and until we
+    // are in the same epoch as their target.
+    return getSlot().plus(1).max(targetEpochStartSlot);
   }
 
   public UInt64 getSlot() {
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
@@ -122,13 +122,13 @@ public class MiscHelpers {
     return epoch.times(specConfig.getSlotsPerEpoch());
   }
 
-  public UInt64 computeSlotAtTime(final UInt64 genesisTime, final UInt64 time) {
-    return time.minusMinZero(genesisTime).dividedBy(specConfig.getSecondsPerSlot());
+  public UInt64 computeTimeAtSlot(final UInt64 genesisTime, final UInt64 slot) {
+    return genesisTime.plus(slot.times(specConfig.getSecondsPerSlot()));
   }
 
-  public UInt64 computeEpochAtTime(final UInt64 genesisTime, final UInt64 time) {
-    final UInt64 slot = computeSlotAtTime(genesisTime, time);
-    return computeEpochAtSlot(slot);
+  public UInt64 computeStartTimeAtEpoch(final UInt64 genesisTime, final UInt64 epoch) {
+    final UInt64 epochStartSlot = computeStartSlotAtEpoch(epoch);
+    return computeTimeAtSlot(genesisTime, epochStartSlot);
   }
 
   public UInt64 computeTimeAtSlot(final BeaconState state, final UInt64 slot) {
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
@@ -114,15 +114,24 @@ public class MiscHelpers {
     }
   }
 
-  public UInt64 computeEpochAtSlot(UInt64 slot) {
+  public UInt64 computeEpochAtSlot(final UInt64 slot) {
     return slot.dividedBy(specConfig.getSlotsPerEpoch());
   }
 
-  public UInt64 computeStartSlotAtEpoch(UInt64 epoch) {
+  public UInt64 computeStartSlotAtEpoch(final UInt64 epoch) {
     return epoch.times(specConfig.getSlotsPerEpoch());
   }
 
-  public UInt64 computeTimeAtSlot(BeaconState state, UInt64 slot) {
+  public UInt64 computeSlotAtTime(final UInt64 genesisTime, final UInt64 time) {
+    return time.minusMinZero(genesisTime).dividedBy(specConfig.getSecondsPerSlot());
+  }
+
+  public UInt64 computeEpochAtTime(final UInt64 genesisTime, final UInt64 time) {
+    final UInt64 slot = computeSlotAtTime(genesisTime, time);
+    return computeEpochAtSlot(slot);
+  }
+
+  public UInt64 computeTimeAtSlot(final BeaconState state, final UInt64 slot) {
     UInt64 slotsSinceGenesis = slot.minus(SpecConfig.GENESIS_SLOT);
     return state.getGenesisTime().plus(slotsSinceGenesis.times(specConfig.getSecondsPerSlot()));
   }
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
@@ -123,7 +123,7 @@ public class MiscHelpers {
   }
 
   public UInt64 computeEndSlotAtEpoch(final UInt64 epoch) {
-    return computeStartSlotAtEpoch(epoch.plus(1)).minusMinZero(1);
+    return computeStartSlotAtEpoch(epoch).plus(specConfig.getSlotsPerEpoch() - 1);
   }
 
   public UInt64 computeSlotAtTime(final UInt64 genesisTime, final UInt64 currentTime) {
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
@@ -127,12 +127,8 @@ public class MiscHelpers {
   }
 
   public UInt64 computeTimeAtSlot(final UInt64 genesisTime, final UInt64 slot) {
-    return genesisTime.plus(slot.times(specConfig.getSecondsPerSlot()));
-  }
-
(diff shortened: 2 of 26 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/AttestationValidator.java
```diff
@@ -130,6 +132,20 @@ SafeFuture<InternalValidationResultWithState> singleOrAggregateAttestationChecks
                 return completedFuture(InternalValidationResultWithState.ignore());
               }
               final BeaconState state = maybeState.get();
+
+              final Optional<SlotInclusionGossipValidationResult>
+                  slotInclusionGossipValidationWithStateResult =
+                      attestationUtil.performSlotInclusionGossipValidation(attestation, state);
```
rolfyone: If we were going to do this, we'd be needing to pass the head state, not just any state chosen for gossip topic inclusion. Head state is also sketchy (eg. future slot errors where we haven't completed epoch transitions).

Because it's using the state to compute the current epoch and previous epoch, if we end up using something like a fork state from previous epoch as the best state, then we'd actually be using current-1, current-2, and that'd be bad...
My suggestion if they're only using state to check against epoch is we compute the epoch against wall time inside that function, and then even if the state machine is behind we're accepting gossip correctly...

Then `performSlotInclusionGossipValidation` would only take the attestation... unless we need a state because of test cases or something...

If we did we could probably just have a wrapper interface that takes attestation and state and computes the epoch of the state then calls a private function that takes `attestation, headEpoch`, and the default public interface could derive headEpoch from wallTime...

Anyway, that's a couple of ideas. happy to talk further.

StefanBratanov: Did change implementation to use the wall time instead.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/util/AttestationUtilDeneb.java
```diff
@@ -0,0 +1,99 @@
...
+   * MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. attestation.data.slot <= current_slot (a
+   * client MAY queue future attestation for processing at the appropriate slot).
+   *
+   * <p>[IGNORE] the epoch of attestation.data.slot is either the current or previous epoch (with a
+   * MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. compute_epoch_at_slot(attestation.data.slot)
+   * in (get_previous_epoch(state), get_current_epoch(state))
+   */
+  @Override
+  public Optional<SlotInclusionGossipValidationResult> performSlotInclusionGossipValidation(
+      final Attestation attestation, final UInt64 genesisTime, final UInt64 currentTimeMillis) {
```
rolfyone, line 50: Do we guard against pre-genesis somewhere ? we'll be pre-genesis a lot in test nets is my main concern...

StefanBratanov: Yeah, when we initialise the store, we set the time to at least genesis and don't update it until we hit it. You can look in `StoreBuilder` class:

image

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
```diff
@@ -114,15 +114,28 @@ public int computeProposerIndex(
...
     return epoch.times(specConfig.getSlotsPerEpoch());
   }
 
-  public UInt64 computeTimeAtSlot(BeaconState state, UInt64 slot) {
+  public UInt64 computeStartTimeAtEpoch(final UInt64 genesisTime, final UInt64 epoch) {
+    final UInt64 epochStartSlot = computeStartSlotAtEpoch(epoch);
+    return computeStartTimeAtSlot(genesisTime, epochStartSlot);
+  }
+
+  public UInt64 computeStartTimeAtSlot(final UInt64 genesisTime, final UInt64 slot) {
```
tbenr: Shouldn't be just another version of `computeTimeAtSlot`, otherwise it should be `computeSlotStartTime` (same for epoch). According to my hears.

tbenr (commented):
we discussed a couple of things in a call. This is one of them

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/phase0/util/AttestationUtilPhase0.java
```diff
@@ -91,7 +79,19 @@ private boolean isFromFarFuture(
...
+  private boolean isCurrentTimeBeforeMinimumAttestationBroadcastTime(
+      final UInt64 attestationSlot, final UInt64 genesisTime, final UInt64 currentTimeMillis) {
+    final UInt64 minimumBroadcastTimeMillis =
+        minimumBroadcastTimeMillis(attestationSlot, genesisTime);
+    return currentTimeMillis.isLessThan(minimumBroadcastTimeMillis);
+  }
+
+  private boolean isCurrentTimeAfterAttestationPropagationSlotRange(
+      final UInt64 attestationSlot, final UInt64 genesisTime, final UInt64 currentTimeMillis) {
+    return maximumBroadcastTimeMillis(attestationSlot, genesisTime).isLessThan(currentTimeMillis);
```
rolfyone: can we check code coverage, and if its not immaculate, then we should consider making tests for these - they could easily be package private static, and we could feed in values to validate with great coverage..

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/util/AttestationUtilDeneb.java
```diff
@@ -0,0 +1,92 @@
...
+        miscHelpers.computeStartSlotAtEpoch(previousEpoch).minusMinZero(slotDisparity);
+    final UInt64 maxSlot = miscHelpers.computeEndSlotAtEpoch(currentEpoch).plus(slotDisparity);
+
+    return attestationSlot.isGreaterThanOrEqualTo(minSlot)
+        && attestationSlot.isLessThanOrEqualTo(maxSlot);
+  }
+
+  private int calculateMaximumGossipClockDisparityInSlots() {
+    return (specConfig.getMaximumGossipClockDisparity() / specConfig.getMillisPerSlot()) + 1;
+  }
```
rolfyone: likewise here, could definitely do some unit testing if we need to check any of this... it would show us our business logic is correct for `isAttestationSlotInCurrentOrPreviousEpoch` for example, and we could easily boundary test etc...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
```diff
@@ -114,15 +114,27 @@ public int computeProposerIndex(
...
+
+  public UInt64 computeSlotAtTime(final UInt64 genesisTime, final UInt64 currentTime) {
+    return currentTime.minusMinZero(genesisTime).dividedBy(specConfig.getSecondsPerSlot());
+  }
+
+  public UInt64 computeTimeAtSlot(final UInt64 genesisTime, final UInt64 slot) {
+    return genesisTime.plus(slot.times(specConfig.getSecondsPerSlot()));
+  }
+
+  public UInt64 computeTimeAtSlot(final BeaconState state, final UInt64 slot) {
```
rolfyone: test... they're simple in this case but we really do need basic tests so that when we refactor we can find issues.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/phase0/util/AttestationUtilPhase0.java
StefanBratanov: Sure, I will extend current tests for Phase0. I didn't touch them but good idea to see we are covering all methods.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttestationData.java
```diff
@@ -70,15 +70,19 @@ public AttestationData(UInt64 slot, AttestationData data) {
...
-        .plus(UInt64.ONE)
-        .max(miscHelpers.computeStartSlotAtEpoch(getTarget().getEpoch()));
+    final UInt64 targetEpochStartSlot = miscHelpers.computeStartSlotAtEpoch(getTarget().getEpoch());
+    return getEarliestSlotForForkChoice(targetEpochStartSlot);
+  }
+
+  public UInt64 getEarliestSlotForForkChoice(final UInt64 targetEpochStartSlot) {
+    // Attestations can't be processed by fork choice until their slot is in the past and until we
+    // are in the same epoch as their target.
+    return getSlot().plus(1).max(targetEpochStartSlot);
```
rolfyone: should use the constant here for less conversions (UInt64.ONE) like initial implementation.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
```diff
@@ -114,15 +114,23 @@ public int computeProposerIndex(
...
-  public UInt64 computeTimeAtSlot(BeaconState state, UInt64 slot) {
+  public UInt64 computeSlotAtTime(final UInt64 genesisTime, final UInt64 currentTime) {
+    return currentTime.minusMinZero(genesisTime).dividedBy(specConfig.getSecondsPerSlot());
+  }
+
+  public UInt64 computeTimeAtSlot(final UInt64 genesisTime, final UInt64 slot) {
+    return genesisTime.plus(slot.times(specConfig.getSecondsPerSlot()));
+  }
+
+  public UInt64 computeTimeAtSlot(final BeaconState state, final UInt64 slot) {
```
rolfyone: This interface i really don't like and its not used much, maybe we just change the couple of occurrences to pass in genesis time...

If we were passing in state i'd expect that we just pass state (which has slot)

Since we're passing in state just to get the genesis time, it's just confusing things, and it'd be better to just not have it and let ppl call interface at 129.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/util/AttestationUtilDeneb.java
```diff
@@ -0,0 +1,95 @@
...
+  public Optional<SlotInclusionGossipValidationResult> performSlotInclusionGossipValidation(
+      final Attestation attestation, final UInt64 genesisTime, final UInt64 currentTimeMillis) {
+    final UInt64 attestationSlot = attestation.getData().getSlot();
+    final UInt64 attestationSlotTimeMillis =
+        secondsToMillis(miscHelpers.computeTimeAtSlot(genesisTime, attestationSlot));
+    if (isAttestationSlotAfterCurrentTime(attestationSlotTimeMillis, currentTimeMillis)) {
+      if (isFromFarFuture(attestation, genesisTime, currentTimeMillis)) {
+        return Optional.of(SlotInclusionGossipValidationResult.IGNORE);
+      }
+      return Optional.of(SlotInclusionGossipValidationResult.SAVE_FOR_FUTURE);
```
rolfyone:
```
return Optional.of(
 isFromFarFuture(attestation, genesisTime, currentTimeMillis)
 ? SlotInclusionGossipValidationResult.IGNORE
 : SlotInclusionGossipValidationResult.SAVE_FOR_FUTURE);

File: ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/deneb/util/AttestationUtilDenebTest.java [hunk 10]
```diff
@@ -0,0 +1,68 @@
...
+            UInt64.valueOf(attestationSlot), genesisTime, currentTimeMillis);
+    assertThat(actualResult).isEqualTo(expectedResult);
+  }
+
+  // minimal is 6 seconds per slot, 8 slots per epoch
+  private static Stream<Arguments> provideAttestationSlotInCurrentOrPreviousEpochArguments() {
+    return Stream.of(
+        Arguments.of(6, getTimeForSlot(10), true),
+        Arguments.of(6, getTimeForSlot(3), true),
+        Arguments.of(6, getTimeForSlot(26), false),
```
rolfyone: Boundary type tests can be more readable testing the slot we fail on vs the slot we pass on...
so for example, at 8 slots per epoch, (so 0-7, 8-15, 16-23), test slot 15,16 etc... (one in second epoch, one the slot after) - helps see what we're testing sometimes, or pass a name where you could describe is sometimes done, so then you could go `upper_bound`, `lower_bound` kind of idea...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttestationData.java
StefanBratanov: For some reason in my head thought using a primitive uses less conversion. Will fix tomorrow.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
StefanBratanov: Agree. Will change.

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/deneb/util/AttestationUtilDenebTest.java
StefanBratanov: makes sense, I added comments and improved test cases across the two test classes. Think it's more clear now and covers more cases.

rolfyone (approved):
LGTM

PR #7308, opened 2023-06-27
Add inactivity leak log

### LINKED ISSUE #7139: Add log lines for the inactivity leak
After four epochs of non-finalisation the beacon chain enters an "inactivity leak". It would be useful to signal this in the logs/console. Perhaps at every epoch transition where the leak is in effect print an `INFO` level message to say the the inactivity leak is in operation. It could be `WARN`, but there's not a lot the operator can do about it, so `INFO` is probably OK.

courtneyeh:
fixes #7139 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/phase0/statetransition/epoch/RewardsAndPenaltiesCalculatorPhase0.java
```diff
@@ -166,8 +169,9 @@ public void applyInactivityPenaltyDelta(
       final UInt64 baseReward,
       final UInt64 finalityDelay,
       final RewardAndPenalty delta) {
-
     if (isInactivityLeak(finalityDelay)) {
+      LOG.info("Beacon chain has entered inactivity leak");
```
lucassaldanha: We probably want to say: "Beacon chain is in activity leak" or something like that. "entered" only makes sense for the first time it happens.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/phase0/statetransition/epoch/RewardsAndPenaltiesCalculatorPhase0.java
```diff
@@ -30,6 +32,7 @@
 import tech.pegasys.teku.spec.logic.common.statetransition.epoch.status.ValidatorStatuses;
 
 public class RewardsAndPenaltiesCalculatorPhase0 extends RewardsAndPenaltiesCalculator {
```
lucassaldanha: I think the log message would make more sense on AbstractEpochProcessor#processEpoch. We could check at the end of the processEpoch method.

Having it deep into the reward calculation isn't very intuitive and having it in the Phase0 implementation means that in the future, if someone overrides the method with Deneb, for example, we would lose this functionality

lucassaldanha (approved):

PR #7313, opened 2023-06-28
Increase blobs pruning limit

tbenr:
Rises the default pruning limit from 32 to 48, considering the new 3/6 max blobs configuration. We could implement some calculation in the future.

## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #7323, opened 2023-07-04
Use `Eth-Consensus-Version` when deserializing ssz

lucassaldanha:
- Passing header data into getRequestBody method
- Consensus version based body selector


data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/AbstractDataBackedRestAPIIntegrationTest.java
```diff
@@ -381,7 +382,12 @@ protected Response post(final String route, final String postData) throws IOExce
 
   protected Response postSsz(final String route, final byte[] postData) throws IOException {
     final RequestBody body = RequestBody.create(postData, SSZ);
-    final Request request = new Request.Builder().url(getUrl() + route).post(body).build();
+    final Request request =
+        new Request.Builder()
+            .url(getUrl() + route)
+            .header(HEADER_CONSENSUS_VERSION, "deneb")
```
lucassaldanha: Added this here temporarily just to see if the plumbing of the header was done right. Won't be part of the final code.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -156,6 +158,29 @@ public <T> T getRequestBody(final InputStream body, final Optional<String> maybe
     }
   }
 
+  @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
+  public <T> T getRequestBody(final InputStream body, final Map<String, String> headers)
```
StefanBratanov, line 162: We may want to deprecate the other `getRequestBody` method to avoid using it in tests from now on.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -326,26 +328,47 @@ public SignedBeaconBlock deserializeSignedBeaconBlock(final Bytes serializedSign
...
+
+    return schemaDefinition
         .getSignedBlindedBlockContainerSchema()
         .sszDeserialize(serializedSignedBlindedBlockContainer);
   }
 
+  private Optional<SchemaDefinitions> getSchemaDefinitionsForMilestone(
+      final Optional<String> milestone) {
+    return milestone
+        .map(v -> SpecMilestone.valueOf(v.toUpperCase(Locale.ROOT)))
```
StefanBratanov: I think it would be useful to have a static method in SpecMilestone like:

```java
public static SpecMilestone fromEthConsensusVersion(final String ethConsensusVersion) {
    return SpecMilestone.valueOf(ethConsensusVersion.toUpperCase(Locale.ROOT));
}
```

Could be reused in tests, etc.

StefanBratanov (approved):
LGTM with couple of suggestions.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -326,26 +328,47 @@ public SignedBeaconBlock deserializeSignedBeaconBlock(final Bytes serializedSign
...
+    final SchemaDefinitions schemaDefinition =
+        getSchemaDefinitionsForMilestone(milestone)
+            .orElseGet(getSchemaDefinitionForSignedBlockSSZ(serializedSignedBlindedBlockContainer));
+
+    return schemaDefinition
         .getSignedBlindedBlockContainerSchema()
         .sszDeserialize(serializedSignedBlindedBlockContainer);
   }
 
+  private Optional<SchemaDefinitions> getSchemaDefinitionsForMilestone(
```
StefanBratanov, line 353: maybe can rename this to `getSchemaDefinitionsForEthConsensusVersion(final Optional<String> ethConsensusVersion)` just to clarify it a bit more

lucassaldanha: EthConsensusVersion is a concept of the API layer. In this case, what we are searching for is a schema for a milestone, in the spec/domain layer. So I think my name makes a bit more sense. I'll keep it as is but we can always revisit it later if that's the case.

PR #7325, opened 2023-07-04
fix PubKeyLoader from externalSigner

### LINKED ISSUE #7301: Failed to load public keys from Web3Signer URL when TLS enabled
### Description
When using the Web3Signer with Teku and specifying a URL to load public keys from, when TLS is enabled Teku will fail to load the public keys due to a SSL exception:
```
javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

When the public keys are explicitly defined in the argument rather than the URL, Teku will correctly boot. As seen in the attached logs, Teku will state that the external signer is reachable but throw an exception upon the retrieval of public keys.

[validator.log](https://github.com/ConsenSys/teku/files/11848741/validator.log)

### Steps to Reproduce (Bug)
- Run Web3Signer with TLS enabled
- Run Teku with Web3Signer enabled and `--validators-external-signer-public-keys` set to a https URL 

**Expected behavior:** Teku will fetch all public keys from Web3Signer

**Actual behavior:** Exception is thrown

**Frequency:** Always, at boot

* Software version: Teku 23.5.0 & 23.6.0, Web3Signer 23.5.0 & 23.6.0
* Java version: Java version built into container `consensys/teku:23.6.0`

tbenr:
`--validators-external-signer-public-keys` accepts `external-signer` value. external signer HttpClient (which includes TLS configuration) will be used when querying the standard `/api/v1/eth2/publicKeys` endpoint against the same url specified via `--validators-external-signer-url`.
 
 I decided to go this way to avoid possible breaking changes to people using https URLs relaying on default JVM trusted store. The benefit is also that, since it is a standard API, user doesn't need to specify the full URL.

fixes #7301 

## Documentation

## Changelog

### DIFF
--- a/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
+++ b/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
@@ -603,7 +603,7 @@ public class ValidatorConfig {
                 final URL url = new URL(source);
                 if (hostAndPortMatching(url, validatorExternalSignerUrl)) {
                   LOG.warn(
-                      "'--validators-external-signer-public-keys' contains an URL matching the external-signer-url host and port. Use 'external-signer' instead if you want to use all public keys exposed by the external-signer");
+                      "'--validators-external-signer-public-keys' contains an URL matching the external-signer-url host and port. Use 'external-signer' instead if you want to use all public keys exposed by the external signer");
                 }
               } catch (MalformedURLException e) {
                 throw new InvalidConfigurationException(
--- a/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
+++ b/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
@@ -15,12 +15,14 @@ package tech.pegasys.teku.validator.api;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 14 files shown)


StefanBratanov (approved):
LGTM

tbenr:
@StefanBratanov can you give another look? I added what Lucas has suggested this morning.

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
```diff
@@ -593,6 +595,36 @@ private void validateExternalSignerUrlAndPublicKeys() {
...
       }
+
+      validatorExternalSignerPublicKeySources.forEach(
+          source -> {
+            if (source.contains(":")) {
+              try {
+                final URL url = new URL(source);
+                if (hostAndPortMatching(url, validatorExternalSignerUrl)) {
+                  LOG.warn(
+                      "'--validators-external-signer-public-keys' contains an URL matching the external-signer-url host and port. Use 'external-signer' instead if you want to use all public keys exposed by the external-signer");
```
StefanBratanov: Super nit :D `exposed by external signer` (without the dash)

StefanBratanov:
> @StefanBratanov can you give another look? I added what Lucas has suggested this morning.

It's good. I like it.

PR #7330, opened 2023-07-05
Fix leveldb open iterators metrics counter

tbenr:
counter increment in the wrong place, not capturing the iterators opened outside the `withIterator` method

## Documentation

## Changelog


zilm13 (approved):
Nice catch!

PR #7331, opened 2023-07-05
Fix ReconstructHistoricalStatesService not starting when initial anchor slot is 0

StefanBratanov:
## PR Description

- This fixes an uint64 underflow issue when initial anchor slot is 0 (no --initial-state or genesis state specified) and `reconstruct-historic-states` is set to true. Technically in that case after this change, the ReconstructHistoricalStatesService does nothing and just completes with log message `ReconstructHistoricalStatesService reconstruction complete` but it is better than throwing an error.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #7333, opened 2023-07-05
Add feature to override min epochs for blob sidecars requests

### LINKED ISSUE #6395: make MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS overridable

zilm13:
## PR Description
It's a bit tricky as we should keep spec value in some places. And also we may need adjust sync behavior in the future, but I decided to make it simpler at the first version until we get into some trouble.

## Fixed Issue(s)

Fixes #6395 

## Documentation

## Changelog


StefanBratanov (commented):
The more I think about this, I would rather name the option something like `--Xepochs-store-blobs` and then description says it should be at least MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS spec value. Thing is we don't really override the value, since the fork choice, rpc methods still use the spec value and it can get confusing. We rather just use this value for the storage part. And syncing/pruner use this value if present instead of the config one. What do you think?

benjaminion:
> I would rather name the option something like `--Xepochs-store-blobs`

Yeah, I like that. Clean, with separation of concerns. We should print a warning, though, if it is set to less than `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS`. I assume in that case we would ignore the user's value, right?

```
WARN: Ignoring Xepochs-store-blobs value N since it is less than MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS = M
```

tbenr:
>  rpc methods still use the spec value and it can get confusing.

I'm not sure about RPC. Thinking about an altruistic node that serves more blobs than the minimal required ones, like an archival node for blobs. And you want to let blobs archival nodes to be able to sync.

zilm13:
@tbenr 
> I'm not sure about RPC.

We will serve it with this code. But we should follow the spec error requirement in case we don't have the blobSidecar only inside spec min epochs

zilm13:
@StefanBratanov yeah, I don't like current huge name, I will follow your suggestions. By the logic I think I've done it in the way you described

zilm13:
@benjaminion It will throw with the current code if the value is set to be smaller than the one from the spec. Do you think the warning would be a better way? We usually are strict on the config values to make sure the user is aware of config errors.

tbenr:
> We will serve it with this code. But we should follow the spec error requirement in case we don't have the blobSidecar only inside spec min epochs

yeah, right!

benjaminion:
> Do you think the warning would be a better way?

Yes, I think I'd prefer a warning. We know how to recover if the parameter is set too small - just ignore it. It won't cause issues for the node operator, since doing what they asked would make their node non-compliant. So I don't think we should crash or `ERROR` - a `WARNING` would be fine. Similar to what we do if max peers is set less than min peers, for example (or what we should do, anyway).

zilm13:
I've applied all feedback, hope, didn't miss anything.
I've kept the user input in case minEpochs spec value will change in the next forks, so we could compare it again in the next forks. But actually in this case we will need to update all usages and they will become very complicated at fork transition zones.
@StefanBratanov and @tbenr ready for review.

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -215,12 +219,17 @@ public Optional<Eth2Network> getEth2Network() {
...
+    return epochsStoreBlobs;
+  }
+
   @Override
   public String toString() {
     return constants;
   }
 
   public static class Builder {
+    private static final String EPOCHS_MAX_KEYWORD = "MAX";
```
StefanBratanov: EPOCHS_STORE_BLOBS_MAX_KEYWORD could be more clarifying name I suppose

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/util/ForkChoiceUtilDeneb.java
```diff
@@ -54,7 +54,7 @@ private UInt64 computeFirstAvailabilityWindowSlot(final ReadOnlyStore store, fin
     final SpecConfigDeneb specConfigDeneb = SpecConfigDeneb.required(specConfig);
     final UInt64 maybeAvailabilityWindowStartSlot =
         spec.computeStartSlotAtEpoch(
-            currentEpoch.minusMinZero(specConfigDeneb.getMinEpochsForBlobSidecarsRequests()));
+            currentEpoch.minusMinZero(specConfigDeneb.getEpochsStoreBlobs()));
```
StefanBratanov: Hmm think `ForkChoice` should use the spec value `getMinEpochsForBlobSidecarsRequests`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigDeneb.java
```diff
@@ -52,6 +52,11 @@ static SpecConfigDeneb required(final SpecConfig specConfig) {
 
   boolean isKZGNoop();
 
+  int getEpochsStoreBlobs();
+
+  /** Maybe user input, use {@link #getEpochsStoreBlobs()} instead */
+  Optional<Integer> getMaybeEpochsStoreBlobs();
```
StefanBratanov: We don't need to expose that method though right, can just store the optional variable in the SpecConfigDenebImpl class and use it in `getEpochsStoreBlobs`

StefanBratanov:
I think in syncing and pruner I would explicitly use something like `isStorageOfBlobSidecarsRequiredAtSlot` to distinguish between availiability (spec) and storage (user preference).

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigDeneb.java
zilm13: I've added it in case next fork could change minEpochs spec value, in this case `getEpochsStoreBlobs` may need to be changed. Actually it's over-complication at this stage and I'm ok to remove it for now

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -632,5 +653,27 @@ public Builder applyChiadoNetworkDefaults() {
               // GnosisDAO Bootnode: 3.69.35.13
               "enr:-Ly4QLgn8Bx6faigkKUGZQvd1HDToV2FAxZIiENK-lczruzQb90qJK-4E65ADly0s4__dQOW7IkLMW7ZAyJy2vtiLy8Bh2F0dG5ldHOIAAAAAAAAAACEZXRoMpAxNnBDAgAAb___________gmlkgnY0gmlwhANFIw2Jc2VjcDI1NmsxoQMa-fWEy9UJHfOl_lix3wdY5qust78sHAqZnWwEiyqKgYhzeW5jbmV0cwCDdGNwgiMog3VkcIIjKA");
     }
+
+    private Optional<Integer> validateAndParseEpochsStoreBlobs(final String epochsStoreBlobs) {
+      if (epochsStoreBlobs == null || epochsStoreBlobs.isBlank()) {
+        return Optional.empty();
+      }
+      if (epochsStoreBlobs.toUpperCase(Locale.ROOT).equals(EPOCHS_STORE_BLOBS_MAX_KEYWORD)) {
```
tbenr: `equalsIgnoreCase` ?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
```diff
@@ -224,7 +224,7 @@ private SafeFuture<Boolean> requestBlocksAndBlobSidecarsByRange() {
             requestParams.getStartSlot(), requestParams.getCount(), requestManager::processBlock);
 
     final SafeFuture<Void> blobSidecarsRequest;
-    if (blobSidecarManager.isAvailabilityRequiredAtSlot(endSlot)) {
+    if (blobSidecarManager.isStorageOfBlobSidecarsRequiredAtSlot(endSlot)) {
```
tbenr: The problem I see here is that we could stop backfilling blocks too if we struggle finding peers serving old blobs. But this is a limitation of current sync design, which bonds block and blobs sync.

Secondly we should make sure we don't downscore peers not serving blobs outside spec DA.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/PeerSync.java
```diff
@@ -171,7 +171,8 @@ private SafeFuture<PeerSyncResult> executeSync(
               final PeerSyncBlobSidecarListener blobSidecarListener =
                   new PeerSyncBlobSidecarListener(requestContext.startSlot, requestContext.endSlot);
 
-              if (blobSidecarManager.isAvailabilityRequiredAtSlot(requestContext.endSlot)) {
+              if (blobSidecarManager.isStorageOfBlobSidecarsRequiredAtSlot(
```
tbenr: This has the same problem of the historical sync, but I think is more critical here: we could stall in forward sync if we are way outside the standard DA and very few\no peers are serving this blobs.

tbenr (commented):
The more I look at this the more seems like a bit more complicated.

My current gut feeling says that sync logic should stick to spec values.

Ideally we should have a blob-dedicated historical sync that is responsible to search for old blobs outside spec DA.

zilm13:
@tbenr changed the behavior, created an issue for archive blobs sync https://github.com/Consensys/teku/issues/7350

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -347,8 +347,8 @@ private BlobSidecarsAndValidationResult computeFinalValidationResult(
         Collections.unmodifiableList(completeValidatedBlobSidecars));
   }
 
-  private boolean isBlockOutsideDataAvailabilityWindow() {
-    return !spec.isAvailabilityOfBlobSidecarsRequiredAtSlot(
+  private boolean isStorageOfBlobSidecarsRequiredAtSlot() {
```
tbenr: We should go back to `isBlockOutsideDataAvailabilityWindow` (note that this is negative of `isStorageOfBlobSidecarsRequiredAtSlot`)

tbenr: The only place I see we can use the new Storage param is in the BlobSidecarPruner

tbenr: The `validateImmediately` flow is used for the historical backfill. But I'd do that in the decoupling backfill PR, otherwise we make backfill of blocks stall if the value is used and set too big.

But IMO that method name should be renamed to something more tight to the historical batch sync, otherwise it won't be clear (and risky) that we want to use the storage limit in that case.

StefanBratanov:
I wonder if it's not an idea to basically only have `--Xepochs-store-blobs` and then let it only be used by the blobs archiver in the new issue if set larger than min epochs and it kicks off only after historical sync for the blobs has finished. (with some callback) It can also be used by the RPC methods if set. This way other changes in the code will be avoided.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -78,6 +78,26 @@ public void warnBellatrixParameterChanged(final String parameterName, final Stri
             Color.YELLOW));
   }
 
+  public void warnDenebEpochsStoreBlobsParameterSet(
+      final String epochsStoreBlobs, final boolean isOverriden, final String specValue) {
+    if (isOverriden) {
+      final boolean isMax = String.valueOf(Integer.MAX_VALUE).equals(epochsStoreBlobs);
+      log.warn(
+          print(
+              "Xepochs-store-blobs sets the number of epochs blob sidecars are stored and requested during the sync to: "
```
tbenr: strange log line IMHO. You re-explain what the param does and you print the actual value at the end. I'd split things in two sentences: "Xepochs-store-blobs has been set to X. It overrides the standard number of epochs blob sidecars are stored and requested during the sync. "

Knd of

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -78,6 +78,26 @@ public void warnBellatrixParameterChanged(final String parameterName, final Stri
             Color.YELLOW));
   }
 
+  public void warnDenebEpochsStoreBlobsParameterSet(
+      final String epochsStoreBlobs, final boolean isOverriden, final String specValue) {
+    if (isOverriden) {
+      final boolean isMax = String.valueOf(Integer.MAX_VALUE).equals(epochsStoreBlobs);
```
tbenr: Since you are using `Integer.MAX_VALUE` here again, I'd suggest tu put it as a constant like `final static Integer MAX_EPOCHS_STORE_BLOBS = Integer.MAX;` and use it everywhere.

tbenr (approved):
LGTM.
Even if I dont like the name of the option.
It is in the Networking config but it refers to storage, seems like inconsistent. And the name itself if quite obscure. We could thinkg about a better name when we unhide it, if we cant think about a better one. From an internal standpoint, since it is a Networking config, we could explicitly refer to t spec var override. But from a user prespective it is definitely a storage option.

tbenr (approved):
LGTM

Issue #6580, opened 2022-12-11
Create CLI option for minimum persistent attestation subnet subscriptions


lucassaldanha:
Currently, Teku has two modes to choose how many attestation subnets it is subscribed to:
- Validator based (validator att subnet + 1 random)
- All subnets (all 64 att subnets)

We want a 3rd mode that creates a minimum number of attestation subnets subscriptions. Keep in mind that if the minimum number is lower than the number of validators, we need to ensure we don't end up subscribing to fewer subnets than we need to.

Example:
- If a node has 10 validators, it is expected to subscribe to 20 subnets. If `min_att_subscriptions = 10`, we should have 20 subscriptions (10 needed + 10 random).
- If a node has 10 validators, it is expected to subscribe to 20 subnets. If `min_att_subscriptions = 30`, we should have 30 subscriptions (10 needed + 20 random)
- If a node has 0 validators, it is expected to subscribe to 0 subnets. If `min_att_subscriptions = 20`, we should have 20 subscriptions (20 random)

ajsutton:
Added as a hidden option and closing as there's no plan to make this a supported thing.

PR #7341, opened 2023-07-07
subnets subscription revamp

### LINKED ISSUE #7296: Update subnet subscriptions according to the spec
### Description
Spec is updated with subscriptions moving from random to deterministic behavior. We should follow and implement this change
https://github.com/ethereum/consensus-specs/pull/3312
The proposed solution is to get beacon nodes to subscribe to long-lived subnets based on their node-id (can think peer-id). Some benefits of this are:
- Enforceable (a peer-id should now be subscribed to a pre-determined subnet, if not we can penalize)

mehdi-aouadi:
## PR Description
Subscribe to fixed [`SUBNETS_PER_NODE`](https://github.com/ethereum/consensus-specs/blob/2cd967e2bb4f8437cdf70ecd1a8a119a005330a9/configs/mainnet.yaml#L127C7-L127C7) attestation subnets based on the node id for [`EPOCHS_PER_SUBNET_SUBSCRIPTION`](https://github.com/ethereum/consensus-specs/blob/2cd967e2bb4f8437cdf70ecd1a8a119a005330a9/configs/mainnet.yaml#L112C4-L112C4) epochs:
- Calculate the subnet ids based on the node id using the new `compute_subscribed_subnets(node_id, epoch)` function
- Remove the random subnet subscriptions / un-subscriptions
- Remove the `--Xp2p-minimum-subnet-subscriptions` cli option added in #6580 

The spec can be found [here](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#attestation-subnet-subscription)
## Fixed Issue(s)

fixes #7296 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/constants/ValidatorConstants.java
```diff
@@ -16,7 +16,7 @@
 public class ValidatorConstants {
   // Phase0
   public static final int TARGET_AGGREGATORS_PER_COMMITTEE = 16;
-  public static final int RANDOM_SUBNETS_PER_VALIDATOR = 1;
+  public static final int NODE_ID_BITS = 256;
```
zilm13: It shouldn't be there, the major change in this revamp is that not only validators will subscribe, any node

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/ValidatorBasedStableSubnetSubscriber.java
```diff
@@ -95,9 +114,8 @@ public void onSlot(final UInt64 slot, final int validatorCount) {
   private Set<SubnetSubscription> adjustNumberOfSubscriptionsToNumberOfValidators(
       UInt64 currentSlot, int validatorCount) {
 
-    final int randomSubnetsPerValidator = ValidatorConstants.RANDOM_SUBNETS_PER_VALIDATOR;
     final int requiredSubnetSubscriptions =
-        min(attestationSubnetCount, randomSubnetsPerValidator * validatorCount);
+        min(attestationSubnetCount, subnetsPerNode * validatorCount);
```
zilm13: similar here, it's not validator based since this change, so we should not multiply it and we should rename the class

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/ValidatorBasedStableSubnetSubscriber.java
```diff
@@ -120,8 +145,43 @@ private Set<SubnetSubscription> adjustNumberOfSubscriptionsToNumberOfValidators(
...
+    final ECPoint ecPoint =
+        Functions.publicKeyToPoint(
+            discoveryNodeId.orElseThrow(
+                () -> new IllegalArgumentException("Unable to get discovery node id")));
+    final Bytes pubKeyUncompressed = Bytes.wrap(ecPoint.getEncoded(false)).slice(1);
+    BigInteger nodeId = new BigInteger(pubKeyUncompressed.toArray());
+    return UInt256.valueOf(nodeId);
+  }
+
+  private List<UInt64> computeSubscribedSubnets(final UInt256 nodeId, final UInt64 epoch) {
```
zilm13: I'd move these and other calculation helpers to something like MiscHelpers, so we could test it carefully

ethereum/spec/src/main/java/tech/pegasys/teku/spec/constants/ValidatorConstants.java
mehdi-aouadi: I'll put it `NetworkConstants` then

mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/ValidatorBasedStableSubnetSubscriber.java
mehdi-aouadi: Moved to `MiscHelpers`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.tuweni.units.bigints.UInt256;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.datastructures.validator.SubnetSubscription;
+
+public class NodeBasedStableSubnetSubscriber implements StableSubnetSubscriber {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private final AttestationTopicSubscriber persistentSubnetSubscriber;
+  private final IntSet availableSubnetIndices = new IntOpenHashSet();
```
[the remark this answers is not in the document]
mehdi-aouadi: Fixed

CHANGELOG.md
```diff
@@ -14,11 +14,13 @@ For information on changes in released versions of Teku, see the [releases page]
 
 - `--exchange-capabilities-enabled` option has been removed since it is no longer applicable because of [execution-apis](https://github.com/ethereum/execution-apis/pull/418) spec change.
 - Removed kiln and ropsten as `--network` options
+- The Development option `--Xp2p-minimum-subnet-subscriptions` has been removed and will no longer be recognised as a command line option.
 
 ### Additions and Improvements
 
 - `--validators-external-signer-public-keys` parameter now accepts `external-signer` value. It will enable public key retrieval from external signer standard API, making sure that configured keystore and trustStore will be used, if any.
 - Stopped calling `engine_exchangeTransitionConfigurationV1` Engine API method, since the method will be deprecated in the future.
+- Subscribe to a fixed number of attestation subnets ([`SUBNETS_PER_NODE`](https://github.com/ethereum/consensus-specs/blob/2cd967e2bb4f8437cdf70ecd1a8a119a005330a9/configs/mainnet.yaml#L127C7-L127C7)) based on the node id for a period of [`EPOCHS_PER_SUBNET_SUBSCRIPTION`](https://github.com/ethereum/consensus-specs/blob/2cd967e2bb4f8437cdf70ecd1a8a119a005330a9/configs/mainnet.yaml#L112C4-L112C4) epochs
```
zilm13: Maybe something like ```Update attestation subnet subscriptions strategy according to [the spec changes](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#attestation-subnet-subscription). Since the update all peers (including non-validator nodes) will subscribe to 2 subnets regardless of the number of validators served.```
@rolfyone @tbenr  what do you think?

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ActiveValidatorTracker.java
```diff
@@ -54,7 +54,7 @@ public void onSlot(final UInt64 slot) {
     final UInt64 epoch = spec.computeEpochAtSlot(slot);
     final int validatorCount = getNumberOfValidatorsForEpoch(epoch);
     LOG.debug("{} active validators counted for epoch {}", validatorCount, epoch);
-    stableSubnetSubscriber.onSlot(slot, validatorCount);
+    stableSubnetSubscriber.onSlot(slot);
```
zilm13: Looks like not a correct place for this, it's no more related to validators. Probably we need to put it in `GossipForkSubscriptionsPhase0`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
```diff
@@ -177,6 +182,52 @@ private IntList computeCommitteeShuffle(
...
+                    .dividedBy(specConfig.getNetworkingConfig().getEpochsPerSubnetSubscription())));
+
+    final int permutedPrefix =
+        computeShuffledIndex(
+            nodeIdPrefix,
+            1 << specConfig.getNetworkingConfig().getAttestationSubnetPrefixBits(),
+            permutationSeed);
+
+    return UInt64.valueOf(
+        permutedPrefix + index % specConfig.getNetworkingConfig().getAttestationSubnetCount());
```
zilm13: I'd add parentheses like in spec here: `(permutated_prefix + index) %` for clarity

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpersTest.java
```diff
@@ -164,6 +166,18 @@ public void isSlotAtNthEpochBoundary_invalidNParameter_negative() {
         .hasMessageContaining("Parameter n must be greater than 0");
   }
 
+  @ParameterizedTest
+  @MethodSource("provideSubnetsForNodeIds")
+  public void testDiscoveryNodeBasedSubnetIds(
+      final String nodeId, final String epoch, int firstSubnetId, int secondSubnetId) {
```
zilm13: It maybe better to make it compatible with any number of subnets (passing ids in listOf) and we will have simpler assertion in the end

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
tbenr: I'd use parentheses here, without relying just on operator precedence

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -720,14 +719,12 @@ protected void initAttestationTopicSubscriber() {
...
+
     final StableSubnetSubscriber stableSubnetSubscriber =
         beaconConfig.p2pConfig().isSubscribeAllSubnetsEnabled()
             ? AllSubnetsSubscriber.create(attestationTopicSubscriber, spec.getNetworkingConfig())
-            : new ValidatorBasedStableSubnetSubscriber(
-                attestationTopicSubscriber,
-                new Random(),
-                spec,
-                beaconConfig.p2pConfig().getMinimumSubnetSubscriptions());
+            : new NodeBasedStableSubnetSubscriber(
```
tbenr, line 740: since this is no more dependent on number of active validator, should we have it injected in the active validator tracker? since it is node based, seems to me that should leave independently by the tracker.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  /**
+   * Adjusts the number of subscriptions to SUBNETS_PER_NODE. Returns the set of new subscriptions
+   * that were added, if there were no new subscriptions, it returns an empty set.
+   */
+  private Set<SubnetSubscription> adjustNumberOfSubscriptionsToNodeRequirement(
+      final UInt64 currentSlot) {
+
+    if (subnetSubscriptions.size() >= subnetsPerNode) {
```
zilm13, line 80: I'm not sure on >. Previously we were going to unsubscribe in this case. The only case I see with > is fork transition, still unclear how could we follow spec on it.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,121 @@
...
+      return emptySet();
+    }
+    LOG.info(
+        "Updating number of persistent subnet subscriptions from {} to {}",
+        subnetSubscriptions.size(),
+        subnetsPerNode);
+
+    final List<UInt64> nodeSubscribedSubnets =
+        spec.getGenesisSpec()
```
zilm13: We have slot here. I'd use it instead of using genesis spec in case we have changes on forks

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,121 @@
...
+                discoveryNodeId.orElseThrow(
+                    () -> new IllegalArgumentException("Unable to get discovery node id")),
+                spec.computeEpochAtSlot(currentSlot));
+
+    final Iterator<UInt64> nodeSubscribedSubnetsIterator = nodeSubscribedSubnets.iterator();
+
+    final Set<SubnetSubscription> newSubnetSubscriptions = new HashSet<>();
+
+    while (subnetSubscriptions.size() < subnetsPerNode) {
+      newSubnetSubscriptions.add(
```
zilm13: Assuming we have 1 active subscription and we have 2 in list from compute, we have no guarantee we are not making duplicate subscription here

zilm13:
I stuck a bit on reviewing this and tried to figure out the questions, for which I'm not sure in answers. Though my gut feeling is "yes" for all 3:
1. How do we subscribe in advance before fork, I have not find this code. Spec requires to subscribe to 2 additional subnets in addition to 2 usual before the forks. It's clear for sync committees when subscribing to the next committee subnets, but it starts from duties for it. Required subnet subscription doesn't start from any duties so it may need additional logic for this? Or did I just miss it and we can reuse?
2. PR implementation subscribes to attnet in the middle for EPOCHS_PER_SUBNET_SUBSCRIPTION, the issue is a compute function will not output these epochs for whole period of our subscription, we are guaranteed to pass the check only during the epoch of the subscription, on the next epoch check could fail. Assuming in Rationale there is some kind of the penalization for not following the rules (not yet proposed in spec and actually it could never happen) it may be more correct to subscribe during the period when we match the output of computation. Should we arrange subscription end epoch so it will match computation at any time of comparing with computation function?
3. PR implementation subscribes in the middle of the epoch, should we do it on start of the epoch? Subscribing in the middle could be safer for network but the check of correctness of the subscription could fail for several slots in this case. Also we usually do other subscriptions on the first slot of the epoch and so it looks like a usual behavior in this case. Should we move the subcriptions of required subnets to the first slot of the epoch?

Some additional info which may be needed. Computation function have input of nodeId and epoch. It will output the same subnetIds for a period of EPOCHS_PER_SUBNET_SUBSCRIPTION moving the beginning of the period using the function of the last bits of the nodeId, so if all nodes would strictly follow the rules, actual switching epochs per computation are equally distributed during EPOCHS_PER_SUBNET_SUBSCRIPTION.
Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#attestation-subnet-subscription
PR (there were few changes already after this PR merged): https://github.com/ethereum/consensus-specs/pull/3312

zilm13:
Ok, I have clarification on my concerns and could finish the review. Answers on my questions:
1. Looking at the spec we must follow "fork versioning" which is already done in our code, [check this lines](https://github.com/ConsenSys/teku/blob/fb9fd5a6fa1a74a7675ae0fc86f65b704dd1b3fb/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/GossipForkManager.java#L119-L129). The spec requires to do this in advance of 256 epochs, which should be just a misspell, I made an [issue for clarification](https://github.com/ethereum/consensus-specs/issues/3451)
2. The short answer is yes, especially considering even not only penalization but the fact we will stop ENR advertisment in the future, we have a transition period currently. So nodes will expect to get certain subnetId by computation using nodeId and would fail on this. So we should strictly follow computation epoch range.
3. The same yes, we usually do it at 1st slot of the epoch and as it's split among 256 shares of all nodes in the network it shouldn't be a problem doing it almost simultaneously.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/DiscoveryNetwork.java
```diff
@@ -105,6 +107,18 @@ public Optional<String> getEnr() {
...
   }
 
+  @Override
+  public Optional<UInt256> getDiscoveryNodeId() {
+    BigInteger nodeId =
+        new BigInteger(
+            1,
+            discoveryService
+                .getNodeId()
+                .orElseThrow(() -> new IllegalArgumentException("Unable to get discovery node id"))
```
zilm13: we'd better keep optional flow here, why to throw if we have optional return

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/DiscoveryNetwork.java
```diff
@@ -105,6 +107,18 @@ public Optional<String> getEnr() {
     return discoveryService.getEnr();
   }
 
+  @Override
+  public Optional<UInt256> getDiscoveryNodeId() {
```
zilm13, line 111: Could you please also add a test checking that UInt256 will match BigInteger with 256bit and smaller numbers. somthing with when(discoveryService.getNodeId).then return new(BigInteger("12345)) -> UInt256.toString match 12345

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
```diff
@@ -13,51 +13,56 @@
...
+    subscriber.onSlot(UInt64.ONE);
+    verify(network, times(spec.getNetworkingConfig().getSubnetsPerNode()))
+        .subscribeToAttestationSubnetId(intThat(i -> i >= 0 && i < attestationSubnetCount));
+    subscriber.onSlot(
+        UInt64.ONE.plus(
+            UInt64.valueOf(
+                (long) spec.getNetworkingConfig().getEpochsPerSubnetSubscription()
+                    * spec.getSlotsPerEpoch(UInt64.ONE))));
+    verify(network, times(spec.getNetworkingConfig().getSubnetsPerNode() * 2))
+        .subscribeToAttestationSubnetId(intThat(i -> i >= 0 && i < attestationSubnetCount));
```
zilm13: I'd also check that 2 lists doesn't match. (I think they could have 1 equal subnet but definitely not both)

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/StableSubnetSubscriberTest.java
```diff
@@ -157,14 +112,15 @@ void shouldReplaceExpiredSubscriptionsWithNewOnes() {
   void shouldGenerateLargeNumberOfSubscriptionsAndCheckTheyreAllCorrect() {
```
zilm13: This one could be dropped, I'm not sure we could test here anything with the changes

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/StableSubnetSubscriberTest.java
```diff
@@ -60,89 +62,42 @@ void shouldNotNotifyAnyChangeWhenNumberOfValidatorsDecrease() {
     ArgumentCaptor<Set<SubnetSubscription>> subnetSubscriptions =
         ArgumentCaptor.forClass(Set.class);
 
-    stableSubnetSubscriber.onSlot(ZERO, 2);
-    verify(validatorApiChannel).subscribeToPersistentSubnets(subnetSubscriptions.capture());
+    stableSubnetSubscriber.onSlot(ZERO);
```
zilm13: we are not changing number of validators in these test since this PR, looks not relevant

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
```diff
@@ -177,6 +182,52 @@ private IntList computeCommitteeShuffle(
...
+  }
+
+  public UInt64 calculateNodeSubnetUnsubscriptionSlot(final UInt64 currentSlot) {
```
zilm13: Ok, as I finally got my answers and clear on this, how should we calculate this slot:
It should be the last slot of the epoch prior to change in subnetIds computation result
So this method should have logic something like this:
```
(nodeId, currentSlot)
final UInt64 nodeOffset =
        UInt64.valueOf(
            nodeId.mod(specConfig.getNetworkingConfig().getEpochsPerSubnetSubscription()).toLong());
currentEpoch = f(currentSlot)
nextPeriodEpoch = ((currentEpoch + specConfig.getNetworkingConfig().getEpochsPerSubnetSubscription()) % specConfig.getNetworkingConfig().getEpochsPerSubnetSubscription()) - nodeOffset
nextPeriodEpochFirstSlot = f(nextPeriodEpoch)
return nextPeriodEpochFirstSlot - 1
```
We need some testing for this, my first idea is iterate over 1000 sequential epochs with several random nodeIds and the results of compute and this function will match. Any extra tests on this will be good too

zilm13 (commented):
As we finally are cleared how this should work, it needs some rework and some more tests would be good

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
mehdi-aouadi: Done

mehdi-aouadi: Done

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
mehdi-aouadi: Makes sense. I separated it from the validator tracker and made it listen to the `SlotEventsChannel`.
@zilm13, this joins your [comment](https://github.com/Consensys/teku/pull/7341#discussion_r1261281048) where you suggested to move it to the `GossipForkSubscriptionsPhase0`. The problem is that we need the `Eth2P2PNetwork` to get the `discoveryNodeId` in order to calculate the subnet ids and the `GossipForkSubscriptionsPhase0` is initialised within the `Eth2P2PNetwork` builder (circular dependency, need the `Eth2P2PNetwork` in order to initialise the `StableSubnetSubscriber`...). I made it completely independant from those. Let me know if that's ok for you or if I'm missing something here

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ActiveValidatorTracker.java
mehdi-aouadi: See [my comment above](https://github.com/Consensys/teku/pull/7341#discussion_r1266942738)

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/DiscoveryNetwork.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Done

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/StableSubnetSubscriberTest.java
mehdi-aouadi: Done

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
zilm13: Ok, I see your changes.
I think we could definitely subscribe not in validator related block. I will think whether we could move it out of `BeaconChainController` (and should we?)

mehdi-aouadi: My bad, I moved the subscription to `initSubnetSubscriber`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java
mehdi-aouadi: Done. I implemented the slot calculation slightly differently and added a test. Will see if I can test it further. Let me know what you think about it now.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/DiscoveryNetwork.java
mehdi-aouadi: Done

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpersTest.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Yep I was thinking of the fork transition case. That's the only one where we could subscribe to more than `subnetsPerNode`. Plus we don't need to unsubscribe manually any more in the new subscription approach...

mehdi-aouadi: So I guess we could leave as it is?!

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/StableSubnetSubscriberTest.java
mehdi-aouadi: Done

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
mehdi-aouadi: Done, added an assertion to check that we have at least one different subnet id after two rounds

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Good catch!
I updated it so we update the subnet un-subscription slot with the new calculated one if we're already subscribed to that subnet (Subscribed to subnet 12 until slot 12, if we're proposed to subscribe again to subnet 12, we keep it and just update the un-subscription slot). I also added a test for that

CHANGELOG.md
mehdi-aouadi: Updated it

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,122 @@
...
+              spec.atSlot(currentSlot)
+                  .miscHelpers()
+                  .calculateNodeSubnetUnsubscriptionSlot(nodeId, currentSlot)));
+    }
+    return newSubnetSubscriptions;
+  }
+
+  private SubnetSubscription subscribeToSubnet(
+      final int subnetId, final UInt64 unsubscriptionSlot) {
+    SubnetSubscription subnetSubscription = new SubnetSubscription(subnetId, unsubscriptionSlot);
```
zilm13: final, please

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,122 @@
...
+    this.discoveryNodeId = discoveryNodeId;
+  }
+
+  @Override
+  public void onSlot(final UInt64 slot) {
+    // Iterate through current subscriptions to remove the ones that have expired
+    final Iterator<SubnetSubscription> iterator = subnetSubscriptions.iterator();
+    while (iterator.hasNext()) {
+      final SubnetSubscription subnetSubscription = iterator.next();
+      if (subnetSubscription.getUnsubscriptionSlot().compareTo(slot) > 0) {
```
zilm13: isGreaterThan could be more readable

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,122 @@
...
+        break;
+      }
+      iterator.remove();
+    }
+
+    // Adjust the number of subscriptions
+    // If there are new subscriptions, pass the new subscription set to BeaconNode
+    Set<SubnetSubscription> newSubnetSubscriptions =
```
zilm13: final please

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
zilm13: On transition we will have same same 2 subnets, so I think we should switch here to `==` to avoid possible errors

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,122 @@
...
+   * Adjusts the number of subscriptions to SUBNETS_PER_NODE. Returns the set of new subscriptions
+   * that were added, if there were no new subscriptions, it returns an empty set.
+   */
+  private Set<SubnetSubscription> adjustNumberOfSubscriptionsToNodeRequirement(
+      final UInt64 currentSlot) {
+
+    if (subnetSubscriptions.size() >= subnetsPerNode) {
+      return emptySet();
+    }
+    LOG.info(
```
zilm13: I'm not sure we need this in info. Do we log anything like this in INFO? It's not very useful for most users. Plus if everything is correct it will log not very useful info - "from 0 to 2" always. I'd add instead DEBUG logging on subscription/unsubscription, something like Subscribed to persistent subnet 12. Unsubscribed from persistent  subnet 45.

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/StableSubnetSubscriberTest.java
```diff
@@ -17,132 +17,83 @@
...
-    assertSubnetsAreDistinct(subnetSubscriptions.getValue());
-    assertThat(subnetSubscriptions.getValue()).hasSize(attestationSubnetCount);
-
-    stableSubnetSubscriber.onSlot(UInt64.valueOf(2), attestationSubnetCount);
-
-    verifyNoMoreInteractions(validatorApiChannel);
-  }
-
   @Test
   void shouldReplaceExpiredSubscriptionsWithNewOnes() {
```
zilm13, line 72: I think the comment in the end is no more relevant and could be removed, since it's calculated now. I'd add following checks to these test: 1. SubnetsSet1 != SubnetsSet2 2. unsubscriptionSlot2  - unsubscriptionSlot1 == 256 epochs

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
```diff
@@ -0,0 +1,153 @@
...
+    return new NodeBasedStableSubnetSubscriber(
+        new AttestationTopicSubscriber(spec, network),
+        spec,
+        Optional.of(dataStructureUtil.randomUInt256()));
+  }
+
+  private void assertSubscribedToAtLeastOneNewSubnet(List<Integer> subscribedSubnetIds) {
+    IntSet subnetIds =
+        IntOpenHashSet.toSet(subscribedSubnetIds.stream().mapToInt(Integer::intValue));
+    assertThat(subnetIds.size()).isGreaterThanOrEqualTo(2);
```
zilm13: isGreaterThanOrEqualTo(1) according to the name or maybe change the name?

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
```diff
@@ -0,0 +1,153 @@
...
+  private NodeBasedStableSubnetSubscriber createSubscriber() {
+    return new NodeBasedStableSubnetSubscriber(
+        new AttestationTopicSubscriber(spec, network),
+        spec,
+        Optional.of(dataStructureUtil.randomUInt256()));
+  }
+
+  private void assertSubscribedToAtLeastOneNewSubnet(List<Integer> subscribedSubnetIds) {
+    IntSet subnetIds =
+        IntOpenHashSet.toSet(subscribedSubnetIds.stream().mapToInt(Integer::intValue));
```
zilm13: why so difficult? it breaks the brain a little looking for extra logic except converting List to the set

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
```diff
@@ -0,0 +1,153 @@
...
+    subscriber.onSlot(UInt64.ONE);
+    subscriber.onSlot(
+        UInt64.ONE.plus(
+            UInt64.valueOf(
+                (long) spec.getNetworkingConfig().getEpochsPerSubnetSubscription()
+                    * spec.getSlotsPerEpoch(UInt64.ONE))));
+
+    final ArgumentCaptor<Integer> subnetIdsCaptor = ArgumentCaptor.forClass(Integer.class);
+
+    verify(network, times(spec.getNetworkingConfig().getSubnetsPerNode() * 2))
```
zilm13, line 79: I'd rather split it to verification after each subscribers onSlot call. Plus it would be great to verify unsubscribe calls too

zilm13 (commented):
Super!
Everything looks good, just few minor nits in `NodeBasedStableSubnetSubscriber` and tests. So I think we need debug lines in `NodeBasedStableSubnetSubscriber` on subscription/unsubscription with subnetId. I'd make it INFO for a while and deploy this build to some test node to see how it behaves for 2 - 3 days (256 epochs is ~27 hours), so it will be ~3 subscription switches in logs, one for less 256 epochs probably, and others for 256. If it's all good, we could move log lines to DEBUG and merge. What do you think on this?

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
mehdi-aouadi: Done. Split it and added un-subscription checks

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
mehdi-aouadi: Actually this test is not relevant. Removed it

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriberTest.java
mehdi-aouadi: Removed it. Doesn't look relevant to me

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/subnets/StableSubnetSubscriberTest.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Done. Added subscription and un-subscription INFO log. Will replace them by DEBUG after testing

tbenr:
@mehdi-aouadi @zilm13 I think we don't need to have a temporary `INFO` level. We can have like it should be and login on the box and go to `/opt/log4j/teku-log-config.xml` and add a Logger there like the following to see how it behaves

```xml
<Logger name="tech.pegasys.teku.networking.eth2.gossip.subnets" level="TRACE"/>
```

tbenr:
or even more specific to a class

```xml
<Logger name="tech.pegasys.teku.networking.eth2.gossip.subnets.NodeBasedStableSubnetSubscriber" level="TRACE"/>
```

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,124 @@
...
+import tech.pegasys.teku.spec.datastructures.validator.SubnetSubscription;
+
+public class NodeBasedStableSubnetSubscriber implements StableSubnetSubscriber {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private final AttestationTopicSubscriber persistentSubnetSubscriber;
+  private final NavigableSet<SubnetSubscription> subnetSubscriptions =
+      new TreeSet<>(
+          Comparator.comparing(SubnetSubscription::getUnsubscriptionSlot)
+              .thenComparing(SubnetSubscription::getSubnetId));
```
tbenr: what about explicitly implementing `Comparable<SubnetSubscription>` in `SubnetSubscription`?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,124 @@
...
+  @Override
+  public void onSlot(final UInt64 slot) {
+    // Iterate through current subscriptions to remove the ones that have expired
+    final Iterator<SubnetSubscription> iterator = subnetSubscriptions.iterator();
+    while (iterator.hasNext()) {
+      final SubnetSubscription subnetSubscription = iterator.next();
+      if (subnetSubscription.getUnsubscriptionSlot().isGreaterThan(slot)) {
+        break;
+      }
+      LOG.info(
```
tbenr: I think this is a logline bit misleading. The actual unsubscription will be done by `persistendSubnetSubscriber`. Here we are just detecting that an unsubscription should happen.

Which make me think about the separation of responsibility between this class and persistentSubnetSubscriber.

Ok, I understand that the local `subnetSubscriptions` data structure is a way to cache the computed subscriptions to avoid run `computeSubscribedSubnets` on each slot.

tbenr (commented):
Looks good to me. I just added some comments. I leave the final review to @zilm13 since he knows the details.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Done. That's better!

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Actually we have a TRACE log in the `AttestationTopicSubscriber` when calling `eth2P2PNetwork.unsubscribeFromAttestationSubnetId(subnetId)` and I guess that should be enough so I removed this one

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Done

mehdi-aouadi:
> or even more specific to a class
> 
> ```
> <Logger name="tech.pegasys.teku.networking.eth2.gossip.subnets.NodeBasedStableSubnetSubscriber" level="TRACE"/>
> ```

This is definitely be a better approach. I updated the logs to a `DEBUG` level as needed. Will play with the log configs on the box when testing

zilm13:
@tbenr thanks for logging advice, I have always troubles with logger, would be a better approach for sure!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/validator/SubnetSubscription.java
```diff
@@ -35,6 +37,13 @@ public UInt64 getUnsubscriptionSlot() {
     return unsubscriptionSlot;
   }
 
+  @Override
+  public int compareTo(@NotNull SubnetSubscription o) {
```
zilm13: final, please

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
```diff
@@ -0,0 +1,118 @@
...
+    final Set<SubnetSubscription> newSubnetSubscriptions = new HashSet<>();
+
+    while (subnetSubscriptions.size() < subnetsPerNode) {
+      final SubnetSubscription newSubnetSubscription =
+          subscribeToSubnet(
+              nodeSubscribedSubnetsIterator.next().intValue(),
+              spec.atSlot(currentSlot)
+                  .miscHelpers()
+                  .calculateNodeSubnetUnsubscriptionSlot(nodeId, currentSlot));
+      LOG.debug(
```
zilm13: Let's move it to AttestationTopicSubscriber:71 and change to TRACE for consistency. So both subscription and unsubscription log lines will be in one class, so we are able to enable it in logs with one logger config line

zilm13 (commented):
Looks great, just 2 more nits. Let's deploy it to testnet server and see how it behaves. After that we could merge it

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/NodeBasedStableSubnetSubscriber.java
mehdi-aouadi: Done

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/validator/SubnetSubscription.java
mehdi-aouadi: Done

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpersTest.java
```diff
@@ -164,6 +166,37 @@ public void isSlotAtNthEpochBoundary_invalidNParameter_negative() {
...
+  public void testDiscoveryNodeBasedSubnetIds(
+      final UInt256 nodeId, final UInt64 epoch, List<UInt64> subnetIds) {
+    final List<UInt64> nodeSubnetIds = miscHelpers.computeSubscribedSubnets(nodeId, epoch);
+    assertThat(nodeSubnetIds).hasSize(subnetIds.size());
+    assertThat(nodeSubnetIds).isEqualTo(subnetIds);
+  }
+
+  @ParameterizedTest
+  @MethodSource("provideNodeIdsAndSlots")
+  public void unsubsciptionEpochMustMatchSubnetsCalculationResultChange(
```
zilm13: unsubsc`r`iptionEpochMustMatchSubnetsCalculationResultChange(

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/AttestationTopicSubscriber.java
```diff
@@ -63,17 +63,21 @@ public synchronized void subscribeToPersistentSubnets(
...
-          subnetIdToUnsubscribeSlot.computeIfAbsent(
-              subnetId,
-              (key) -> {
-                eth2P2PNetwork.subscribeToAttestationSubnetId(subnetId);
-                return ZERO;
-              });
-
-      subnetIdToUnsubscribeSlot.put(
-          subnetId, existingUnsubscriptionSlot.max(subnetSubscription.getUnsubscriptionSlot()));
+      LOG.trace("Subscribing to persistent subnet {}", subnetId);
```
zilm13: let's add ` with unsubscribe due at slot {}` similar to aggregation committee log line

zilm13 (approved):
LGTM, just final touch

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpersTest.java
mehdi-aouadi: Done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/subnets/AttestationTopicSubscriber.java
mehdi-aouadi: Done

Nashatyrev:
Just adding a reference to the spec pr: https://github.com/ethereum/consensus-specs/pull/3312

Issue #6943, opened 2023-03-14
SPIKE: what's next for storage?


rolfyone:
- Leveldb seems to be not supported any longer
 - Rocksdb wasn't very stable previously which is why we moved to leveldb

Moving forward we need to decide what we're doing in this area, we need to get a good idea of a direction.

zilm13:
For the record: we've halted updates of RocksDB at `7.7.3` after #6930, while it's regularly updated.
We may consider to begin updating it again once we are sure we have some testing for it

tbenr:
I did another search session and I found this https://github.com/jankotek/mapdb/

Update: seems like storage isn't mature enough: https://mapdb.org/blog/mapdb_storage_and_design_change/
Pure java implementation would have been interesting though.

tbenr:
there are interesting java projects referenced here:
https://stackoverflow.com/questions/260804/alternative-to-berkeleydb

tbenr:
https://www.speedb.io/
seems like an alternative to rocksdb. We should check with besu guys if they had a chance to evaluate, since it is data-file compatible.

tbenr:
for the records, Besu has integrated the use of BlobDB which is part of RocksDB. They put static data in blobs to reduce compactions.

https://github.com/hyperledger/besu/pull/5471
https://github.com/hyperledger/besu/pull/5475

resources:
https://github.com/facebook/rocksdb/wiki/BlobDB
https://rocksdb.org/blog/2021/05/26/integrated-blob-db.html

rolfyone:
- articulate state of where we're at (problem statement)

rolfyone:
@tbenr close?

tbenr:
ok let's close it for now

PR #7345, opened 2023-07-10
update dependencies

mehdi-aouadi:
## PR Description
Dependencies update

## Documentation

## Changelog


StefanBratanov:
It seems like swagger UI updated its major version. It would be worth starting a node locally and testing out swagger works as expected.

mehdi-aouadi:
> It seems like swagger UI updated its major version. It would be worth starting a node locally and testing out swagger works as expected.

Yep did that already. Swagger looked working fine locally 🚀
Would be cool if you try it too to be sure 🙏

zilm13:
2 questions:
- Will we ignore RocksDB updates forever? Maybe add some comment there?
- It shows javalin 5.6.1 is available for me

zilm13 (approved):
LGTM

mehdi-aouadi:
> 2 questions:
> 
> * Will we ignore RocksDB updates forever? Maybe add some comment there?
> * It shows javalin 5.6.1 is available for me

Regarding javalin 5.6.1, [javalin-rendering](https://github.com/javalin/javalin-rendering/tags) 5.6.1 is not published yet. Since we're grouping the javalin dependencies it would break unless we ungroup javalin core and rendering (which doesn't seem to be the best solution). So I preferred to update it once the rendering lib is published.

Regarding Rocksdb, since we didn't update it for a while, I thought a separate update would be better to evaluate its impact. Plus the last time we downgraded from 7.9.2 back to 7.7.3, so I think it's worth being more careful about it

zilm13:
@mehdi-aouadi added comment there about RocksDB https://github.com/Consensys/teku/issues/6943#issuecomment-1629069947
Not sure it's an appropriate place but anyway

mehdi-aouadi:
> @mehdi-aouadi added comment there about RocksDB [#6943 (comment)](https://github.com/Consensys/teku/issues/6943#issuecomment-1629069947)
> Not sure it's an appropriate place but anyway

Cool thanks! I was thinking I'd suggest to create an internal issue, didn't know we already had a spike about it... Good to know

StefanBratanov:
> > It seems like swagger UI updated its major version. It would be worth starting a node locally and testing out swagger works as expected.
> 
> Yep did that already. Swagger looked working fine locally 🚀 Would be cool if you try it too to be sure 🙏

Tested locally as well against your branch. Confirmed swagger looks good

PR #7352, opened 2023-07-11
Fix description for `delete-hot-blocks` command

StefanBratanov:
## PR Description
`delete-hot-blocks` deletes non-justified blocks which can't be parsed from the db. If you pass `--delete-all`, all non-justified blocks are deleted. The description in this PR should fit better.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


rolfyone (approved):
LGTM, hopefully this is just devs using it

PR #7354, opened 2023-07-12
update openapidoc node dependencies

tbenr:
updates dependencies for the nodejs app used to publish openapi docs.

updates node executor to node 16

I generated a lockfile version 3 (that's why so many deletions) which should be backward compatible to npm version 8 (used by node 16).

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #7290, opened 2023-06-22
getBlobSidecars improvement


tbenr:
I think we could avoid in most cases to lookup blocks in order to lookup the requested blobs.

for instance:

if the id is a finalized slot, we can directly lookup blobs by slot, assuming only one the finalized ones are present.
if the id is a block root, we can lookup in our `slotsByFinalizedRoot` column to get the slot we need.

PR #7358, opened 2023-07-13
Refactor retrieval of blobs in GetBlobSidecars endpoint

StefanBratanov:
## PR Description

- Introduce `AbstractBlockIdSelectorFactory` which returns a selector based on `block_id` or `state_id` rest api parameter
- Created `BlobSidecarSelectorFactory` (currently we were using the `BlockSelectorFactory` to retrieve block and then we retrieve the blob sidecars) which will help with more easily define how to retrieve blobs, which don't need retrieval of blocks. Current implementation retrieves blocks anyways, but this will change with future PRs.
- Made `BlockSelectorFactory` and `StateSelectorFactory` use the new abstract class
- Quite a few renamings etc.

## Fixed Issue(s)
will help with #7290 

## Documentation

## Changelog


data/provider/src/main/java/tech/pegasys/teku/api/blockselector/BlockSelectorFactory.java
```diff
@@ -29,49 +29,27 @@
...
-    try {
-      return forSlot(UInt64.valueOf(selectorMethod));
-    } catch (NumberFormatException ex) {
-      throw new BadRequestException("Invalid block: " + selectorMethod);
-    }
   }
 
+  @Override
+  public BlockSelector stateRootSelector(final Bytes32 stateRoot) {
+    throw new UnsupportedOperationException();
```
tbenr: as mentioned in DM, this will cause a 500 instead of a 400 status code we return without the change

tbenr (approved):
LGTM

PR #7359, opened 2023-07-14
Added epoch state cache configuration

rolfyone:
Now that attestations aren't being validated against states so often, we need a cache closer aligned to what we're doing when we need to regenerate - basically we want to avoid regeneration of state if at all possible.

One epoch transition the finalized and sometimes justified state get read, and were causing regenerations, so adding a cache to handle storing those, but leaving it size 0 by default for now.

The idea would be to have the current state cache reduced significantly in size, potentially to about half an epoch or so, and have an epoch cache with a maximum size over 2, and could probably be below 10 by default.

This would mean the vast majority of the time, no state regeneration is required, and our states in memory can drop to a much lower number than currently stored in memory (32 * 5 max by default currently).

Tested in minimal network, needs to be deployed to a more complex setup to see if it behaves well in a real environment.

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
@@ -505,7 +505,7 @@ class Store implements UpdatableStore {
   }
 
   @Override
-  public SafeFuture<Optional<SignedBlockAndState>> retrieveBlockAndState(Bytes32 blockRoot) {
+  public SafeFuture<Optional<SignedBlockAndState>> retrieveBlockAndState(final Bytes32 blockRoot) {
     return getAndCacheBlockAndState(blockRoot);
   }
 
@@ -516,7 +516,7 @@ class Store implements UpdatableStore {
   }
 
   @Override
-  public SafeFuture<Optional<BeaconState>> retrieveBlockState(Bytes32 blockRoot) {
+  public SafeFuture<Optional<BeaconState>> retrieveBlockState(final Bytes32 blockRoot) {
     return getAndCacheBlockAndState(blockRoot)
         .thenApply(
             maybeStateAndBlockSummary ->
@@ -524,13 +524,14 @@ class Store implements UpdatableStore {
   }
 
   @Override
-  public SafeFuture<Optional<BeaconState>> retrieveCheckpointState(Checkpoint checkpoint) {
+  public SafeFuture<Optional<BeaconState>> retrieveCheckpointState(final Checkpoint checkpoint) {
     return checkpointStates.perform(
         new StateAtSlotTask(spec, checkpoint.toSlotAndBlockRoot(spec), this::retrieveBlockState));
   }
 
   @Override
-  public SafeFuture<Optional<BeaconState>> retrieveStateAtSlot(SlotAndBlockRoot slotAndBlockRoot) {
+  public SafeFuture<Optional<BeaconState>> retrieveStateAtSlot(
+      final SlotAndBlockRoot slotAndBlockRoot) {
     return checkpointStates.perform(
         new StateAtSlotTask(spec, slotAndBlockRoot, this::retrieveBlockState));
   }
@@ -561,7 +562,7 @@ class Store implements UpdatableStore {
 
   @Override
   public SafeFuture<Optional<BeaconState>> retrieveCheckpointState(
-      Checkpoint checkpoint, final BeaconState latestStateAtEpoch) {
+      final Checkpoint checkpoint, final BeaconState latestStateAtEpoch) {
     return checkpointStates.perform(
         new StateAtSlotTask(
             spec,
@@ -589,7 +590,7 @@ class Store implements UpdatableStore {
     }
   }
 
-  VoteTracker getVote(UInt64 validatorIndex) {
+  VoteTracker getVote(final UInt64 validatorIndex) {
     readVotesLock.lock();
     try {
       if (validatorIndex.intValue() >= votes.length) {
@@ -847,13 +848,13 @@ class Store implements UpdatableStore {
         });
   }
 
-  void updateFinalizedAnchor(AnchorPoint latestFinalized) {
+  void updateFinalizedAnchor(final AnchorPoint latestFinalized) {
     pruneOldFinalizedStateFromEpochCache(this.finalizedAnchor);
     finalizedAnchor = latestFinalized;
     cacheFinalizedAnchorPoint(latestFinalized);
   }
 
-  private void cacheFinalizedAnchorPoint(AnchorPoint latestFinalized) {
+  private void cacheFinalizedAnchorPoint(final AnchorPoint latestFinalized) {
     maybeEpochStates.ifPresent(
         epochStates -> {
           final BeaconState state = latestFinalized.getState();
@@ -878,7 +879,7 @@ class Store implements UpdatableStore {
         });
   }
 
-  void updateJustifiedCheckpoint(Checkpoint checkpoint) {
+  void updateJustifiedCheckpoint(final Checkpoint checkpoint) {
     this.justifiedCheckpoint = checkpoint;
     maybeEpochStates.ifPresent(
         epochStates -> {
@@ -891,7 +892,7 @@ class Store implements UpdatableStore {
         });
   }
 
-  void updateBestJustifiedCheckpoint(Checkpoint checkpoint) {
+  void updateBestJustifiedCheckpoint(final Checkpoint checkpoint) {
     this.bestJustifiedCheckpoint = checkpoint;
   }
 }
--- a/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
@@ -27,7 +27,6 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
@@ -47,13 +46,11 @@ import tech.pegasys.teku.dataproviders.lookup.StateAndBlockSummaryProvider;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.collections.LimitedMap;
-import tech.pegasys.teku.infrastructure.collections.LimitedSet;
 import tech.pegasys.teku.infrastructure.metrics.SettableGauge;
 import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.BlobSidecar;
-import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlockSummary;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBlockAndState;
 import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
@@ -89,7 +86,7 @@ class Store implements UpdatableStore {
 
   private Optional<SettableGauge> epochStatesCountGauge = Optional.empty();
 
-  private final Optional<Set<StateAndBlockSummary>> maybeEpochStates;
+  final Optional<Map<Bytes32, StateAndBlockSummary>> maybeEpochStates;
 
   private final Spec spec;
   private final StateAndBlockSummaryProvider stateProvider;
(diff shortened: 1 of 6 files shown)


storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -609,14 +640,43 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
...
     }
+
+    // is it an epoch boundary?
+    if (maybeEpochStates.isPresent()) {
+      final Optional<StateAndBlockSummary> maybeEpochState =
+          maybeEpochStates.get().stream()
+              .filter(
+                  stateAndBlockSummary ->
+                      stateAndBlockSummary.getBlockSummary().getRoot().equals(blockRoot))
+              .findFirst();
```
ajsutton: Might be worth making `maybeEpochStates` a map keyed by block root to make lookups faster?

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -609,14 +640,43 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
...
+                    ? states.perform(maybeTask.get()).thenPeek(this::cacheIfEpochState)
                     : EmptyStoreResults.EMPTY_STATE_AND_BLOCK_SUMMARY_FUTURE);
   }
 
+  private void cacheIfEpochState(Optional<StateAndBlockSummary> maybeStateAndBlockSummary) {
+    if (maybeStateAndBlockSummary.isPresent() && maybeEpochStates.isPresent()) {
+      final UInt64 slot = maybeStateAndBlockSummary.get().getSlot();
+      final int slotsPerEpoch =
+          spec.atSlot(maybeStateAndBlockSummary.get().getSlot()).getConfig().getSlotsPerEpoch();
+      if (slot.mod(slotsPerEpoch).isZero()) {
```
ajsutton: This is going to miss epochs where the first slot is empty. I suspect the key question is whether we'd want to cache the first block from the epoch or if it's better to make this a cache of checkpoint states where the state is from the last block of the previous epoch with the state transition applied. The nice thing about using that checkpoint state approach is that happens to be what we pre-cache prior to the epoch transition so we'd always generate that state and then it can be used to apply the first block of the epoch without redoing the transition regardless of how many empty slots are present at the start of the epoch. It can also be used as-is for any duty calculations that needs a state from that epoch.

But it probably depends which lookups are trying to hit this cache - it wouldn't really work for getting a state by block root in that case which I think is more what you're targeting for getting the justified/finalized states.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -747,4 +807,24 @@ private void putBlock(final SignedBeaconBlock block) {
...
+
+  @VisibleForTesting
+  Optional<Set<StateAndBlockSummary>> getEpochStates() {
+    return maybeEpochStates;
+  }
+
+  void removeStateAndBlock(final Bytes32 root) {
+    blocks.remove(root);
+    states.remove(root);
+  }
```
ajsutton: This probably should remove from epoch states too which would work well if it's keyed by block root (not so well if it winds up being the checkpoint state cache I mention above...). In long periods of non-finalisation you may have some relatively long chains invalidated when finalisation invalidates a fork.

ajsutton (commented):
Other thing to consider here is that this epoch state cache probably should be setup to ensure it works well with the logic to persist hot states periodically which is also looking for states at the epoch boundary. I'm not sure if that's dependent on what's in the cache to be able to persist it or not.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
rolfyone: that was my initial intent, it would have been so simple...
pruned hot blocks listed the finalized block so it deleted the associated state, so i handled the epochs separately initially so that it remains...

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
rolfyone: given how the 'hot state persistence' is done, i might be able to more closely integrate with that - will take another pass at it...

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -609,14 +637,46 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
...
     return createStateGenerationTask(blockRoot)
         .thenCompose(
             maybeTask ->
                 maybeTask.isPresent()
-                    ? states.perform(maybeTask.get())
+                    ? states.perform(maybeTask.get()).thenPeek(this::cacheIfEpochState)
                     : EmptyStoreResults.EMPTY_STATE_AND_BLOCK_SUMMARY_FUTURE);
   }
 
+  private void cacheIfEpochState(Optional<StateAndBlockSummary> maybeStateAndBlockSummary) {
```
tbenr: `final`

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -575,7 +599,11 @@ VoteTracker getVote(UInt64 validatorIndex) {
 
   private SafeFuture<Optional<BeaconState>> getAndCacheBlockState(final Bytes32 blockRoot) {
     return getOrRegenerateBlockAndState(blockRoot)
-        .thenApply(res -> res.map(StateAndBlockSummary::getState));
+        .thenApply(
+            res -> {
```
tbenr: I'd rename to `maybeStateAndBlockSummary` for readability

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -609,14 +637,46 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
...
             maybeTask ->
                 maybeTask.isPresent()
-                    ? states.perform(maybeTask.get())
+                    ? states.perform(maybeTask.get()).thenPeek(this::cacheIfEpochState)
                     : EmptyStoreResults.EMPTY_STATE_AND_BLOCK_SUMMARY_FUTURE);
   }
 
+  private void cacheIfEpochState(Optional<StateAndBlockSummary> maybeStateAndBlockSummary) {
+    if (maybeStateAndBlockSummary.isPresent() && maybeEpochStates.isPresent()) {
+      final StateAndBlockSummary summary = maybeStateAndBlockSummary.get();
```
tbenr: I'd name it as `stateAndBlockSummary`

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -609,14 +637,46 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
...
+      final UInt64 slot = summary.getSlot();
+      if (!isSlotAtNthEpochBoundary(slot, summary.getParentRoot(), 1)) {
+        return;
+      }
+
+      final Map<Bytes32, StateAndBlockSummary> epochStates = maybeEpochStates.get();
+      if (!slot.mod(spec.getSlotsPerEpoch(slot)).isZero()) {
+        // pre-epoch transition state
+        // This will be referenced during epoch transition if the first slot of the epoch is empty
+        final Optional<StateAndBlockSummary> maybeParent =
```
tbenr: `maybeParentStateAndBlockSummary`

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -609,14 +637,46 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
...
+            states.getIfAvailable(summary.getParentRoot());
+        maybeParent.ifPresent(
+            stateAndBlockSummary -> epochStates.put(summary.getParentRoot(), stateAndBlockSummary));
```
tbenr: `parentStateAndBlockSummary`

tbenr (commented):
for now just commenting some nits. I want to run some tests too before approving

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
tbenr: do we break anything if we move `cacheIfEpochState` from here to directly in `getOrRegenerateBlockAndState:638` ?
seems like more consistent and we won't try to put result in `epochState` cache when there is a cache hit from `epochState` itself.

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransactionUpdates.java
```diff
@@ -118,16 +118,24 @@ public void applyToStore(final Store store, final UpdateResult updateResult) {
 
     // Update finalized data
     finalizedChainData.ifPresent(
-        finalizedData -> store.finalizedAnchor = finalizedData.getLatestFinalized());
+        finalizedData -> {
+          store.maybeEpochStates.ifPresent(
```
tbenr: can we add some comments to cover the reasoning? or move this in a method with a meaningful name

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -609,14 +637,46 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
     if (cachedResult.isPresent()) {
       return SafeFuture.completedFuture(cachedResult);
     }
+
+    // is it an epoch boundary?
+    if (maybeEpochStates.isPresent()) {
```
tbenr: what about?

```java
    final Optional<StateAndBlockSummary> maybeEpochState =
        maybeEpochStates.flatMap(epochStates -> Optional.ofNullable(epochStates.get(blockRoot)));
    if (maybeEpochState.isPresent()) {
      return SafeFuture.completedFuture(maybeEpochState);
    }
```

tbenr:
I think I hit a bug.

My setup:
- a 2 node interop with 8 slots per epoch - 3 seconds per slot
- Both nodes have an hardcoded failure on _first-slot-of-every-epoch_ block production.
- Node 1  have an hardcoded 2 minutes delay on block publishing when Deneb fork enables.
- Node 2 always publishes on time.

so turns out that Node 1 created its first deneb block at slot 45, and released 2 minutes late to Node 2.

Node 2 exception:
image

seems like it was trying to apply the block over a state taken from `epochStates` which was from slot 46? 🤔

tbenr:
tried to replicate with no luck

rolfyone:
> for now just commenting some nits. I want to run some tests too before approving

just remember that its off by default ;)

rolfyone:
> seems like it was trying to apply the block over a state taken from `epochStates` which was from slot 46? 🤔

checkpoint states was untouched so thats unusual... would be interesting to see what happens without the epoch-states flag set, epoch-states would be storing slot 46 if it was populated, and if finalized was 45 then it'd end up with that too...
were there regenerating messages?

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
rolfyone: wish I had left some trace in - should be ok, i'll instrument it again... good pick up maybe...

rolfyone: the problem is the other getBlockState, will refactor but its currently separate code paths.

rolfyone:
> I think I hit a bug.

pretty sure it's a general teku bug if it's a bug... weird things can happen on hard fork that we've seen occasionally and not been able to reproduce

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
rolfyone: i really dont mind, can change to that format i guess.

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransactionUpdates.java
rolfyone: i pretty much rewrote all of this should be clearer now

tbenr:
> were there regenerating messages?

I think it broke before printing one.

In general yes, this test triggers regeneration and i used:
Xstore-state-cache-size: 16
Xstore-epoch-state-cache-size: 10

tbenr:
> pretty sure it's a general teku bug if it's a bug... weird things can happen on hard fork that we've seen occasionally and not been able to reproduce

could be definitely a pre-existing problem, simply coming from a different cache.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -819,8 +839,59 @@ void removeStateAndBlock(final Bytes32 root) {
...
-            epochStates.remove(root);
+            final StateAndBlockSummary stateAndBlockSummary = epochStates.remove(root);
+            if (stateAndBlockSummary != null) {
+              LOG.trace("epochCache REM {}", stateAndBlockSummary::getSlot);
+            }
+          }
+        });
+  }
+
+  void updateFinalizedAnchor(final AnchorPoint latestFinalized) {
```
tbenr, line 850: `private`

tbenr (approved):
LGTM

tbenr:
I'm currently thinking that the exception above could be due to a race condition on gossip validation of blobSidecars and block. The code was publishing them together, which explains the double "epochStates cache hit".

but the strange thing is that both hits resulted in state at slot 46, starting from a query asking for slot 45 parent root!

I currently don't see how epochState can have an entry like: `blockAt45.parentBlockRoot` -> `stateAtSlot46`

PR #7360, opened 2023-07-14
Improve retrieval of blobs in GetBlobSidecars

### LINKED ISSUE #7290: getBlobSidecars improvement

StefanBratanov:
## PR Description

- `blockRootSelector` tries to get finalized slot from the root if it is not present, it falls back to getting a block by the root.
- `slotSelector` checks if the slot is finalized, if it is, then it gets all blob sidecars for it (assumes that DB does not contain any additional blob sidecars for finalized slots). If slot is not finalized, it falls back to getting a block at a given slot. 
- Refactor of the `getBlobSidecars` method in `CombinedChainDataClient`

## Fixed Issue(s)
fixes #7290 

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
@@ -36,8 +36,18 @@ public class BlobSidecarSelectorFactory extends AbstractSelectorFactory<BlobSide
   public BlobSidecarSelector blockRootSelector(final Bytes32 blockRoot) {
     return indices ->
         client
-            .getBlockByBlockRoot(blockRoot)
-            .thenCompose(maybeBlock -> getBlobSidecarsForBlock(maybeBlock, indices));
+            .getFinalizedSlotByBlockRoot(blockRoot)
+            .thenCompose(
+                maybeSlot -> {
+                  if (maybeSlot.isPresent()) {
+                    final SlotAndBlockRoot slotAndBlockRoot =
+                        new SlotAndBlockRoot(maybeSlot.get(), blockRoot);
+                    return getBlobSidecars(slotAndBlockRoot, indices);
+                  }
+                  return client
+                      .getBlockByBlockRoot(blockRoot)
+                      .thenCompose(maybeBlock -> getBlobSidecarsForBlock(maybeBlock, indices));
+                });
   }
 
   @Override
@@ -45,12 +55,7 @@ public class BlobSidecarSelectorFactory extends AbstractSelectorFactory<BlobSide
     return indices ->
         client
             .getChainHead()
-            .map(
-                head -> {
-                  final SlotAndBlockRoot slotAndBlockRoot =
-                      new SlotAndBlockRoot(head.getSlot(), head.getRoot());
-                  return client.getBlobSidecars(slotAndBlockRoot, indices).thenApply(Optional::of);
-                })
+            .map(head -> getBlobSidecars(head.getSlotAndBlockRoot(), indices))
             .orElse(SafeFuture.completedFuture(Optional.empty()));
   }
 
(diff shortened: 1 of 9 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/CombinedStorageChannelSplitter.java
```diff
@@ -218,6 +223,11 @@ public SafeFuture<Optional<BlobSidecar>> getBlobSidecar(final SlotAndBlockRootAn
     return asyncRunner.runAsync(() -> queryDelegate.getBlobSidecar(key));
   }
 
+  @Override
+  public SafeFuture<List<SlotAndBlockRootAndBlobIndex>> getBlobSidecarKeys(final UInt64 slot) {
```
zilm13, line 227: I'd add a comment to use it carefully for non-finalized data as it  could return non-canonical blob sidecars

storage/api/src/main/java/tech/pegasys/teku/storage/api/StorageQueryChannel.java
```diff
@@ -87,6 +89,8 @@ SafeFuture<List<BlobSidecar>> getBlobSidecarsBySlotAndBlockRoot(
 
   SafeFuture<Optional<BlobSidecar>> getBlobSidecar(SlotAndBlockRootAndBlobIndex key);
 
+  SafeFuture<List<SlotAndBlockRootAndBlobIndex>> getBlobSidecarKeys(UInt64 slot);
```
zilm13, line 93: I'd add a comment to use it carefully for non-finalized data as it  could return non-canonical blob sidecars

storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
```diff
@@ -285,6 +287,17 @@ public SafeFuture<Optional<BlobSidecar>> getBlobSidecar(final SlotAndBlockRootAn
     return SafeFuture.of(() -> database.getBlobSidecar(key));
   }
 
+  @Override
+  public SafeFuture<List<SlotAndBlockRootAndBlobIndex>> getBlobSidecarKeys(final UInt64 slot) {
```
zilm13, line 291: I'd add a comment to use it carefully for non-finalized data as it  could return non-canonical blob sidecars

data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
```diff
@@ -84,6 +94,16 @@ private SafeFuture<Optional<List<BlobSidecar>>> getBlobSidecarsForBlock(
...
-    return client.getBlobSidecars(block.getSlotAndBlockRoot(), indices).thenApply(Optional::of);
+    return getBlobSidecars(block.getSlotAndBlockRoot(), indices);
+  }
+
+  private SafeFuture<Optional<List<BlobSidecar>>> getBlobSidecars(
+      final SlotAndBlockRoot slotAndBlockRoot, final List<UInt64> indices) {
+    return client.getBlobSidecars(slotAndBlockRoot, indices).thenApply(Optional::of);
+  }
+
+  private SafeFuture<Optional<List<BlobSidecar>>> getBlobSidecars(
```
zilm13, line 105: maybe `getFinalizedBlobSidecars` for more clarity?

data/provider/src/test/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactoryTest.java
```diff
@@ -87,8 +90,27 @@ public void genesisSelector_shouldGetBlobSidecarsForSlotZero()
...
+    final Optional<List<BlobSidecar>> result =
+        blobSidecarSelectorFactory
+            .blockRootSelector(block.getRoot())
+            .getBlobSidecars(indices)
+            .get();
+    assertThat(result).hasValue(blobSidecars);
+  }
+
+  @Test
+  public void blockRootSelector_shouldGetBlobSidecarsByRetrievingBlock()
```
zilm13, line 110: +WhenSlotNotFinalized in name?

zilm13 (approved):
Looks good, I'd just clarify all the way down in by slot methods, that it could return non-canonical blobSidecars as someone could want to reuse it in the future and we all will forget about it

data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
StefanBratanov: I added comment instead

storage/src/main/java/tech/pegasys/teku/storage/server/CombinedStorageChannelSplitter.java
StefanBratanov: Makes sense. I added a comment on the interface level.

PR #7362, opened 2023-07-16
Implement produce signed voluntary exit Keymanager API

### LINKED ISSUE #7002: [Keymanager API] implement produce signed voluntary exit
https://github.com/ethereum/keymanager-APIs/pull/58

courtneyeh:
fixes #7002 

## Documentation

## Changelog

### DIFF
--- a/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorApiChannel.java
+++ b/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorApiChannel.java
@@ -46,126 +46,126 @@ public interface ValidatorApiChannel extends ChannelInterface {
       new ValidatorApiChannel() {
         @Override
         public SafeFuture<Optional<GenesisData>> getGenesisData() {
-          return null;
+          return SafeFuture.completedFuture(Optional.empty());
         }
 
         @Override
         public SafeFuture<Map<BLSPublicKey, Integer>> getValidatorIndices(
             Collection<BLSPublicKey> publicKeys) {
-          return null;
+          return SafeFuture.completedFuture(Map.of());
         }
 
         @Override
         public SafeFuture<Optional<Map<BLSPublicKey, ValidatorStatus>>> getValidatorStatuses(
             Collection<BLSPublicKey> validatorIdentifiers) {
-          return null;
+          return SafeFuture.completedFuture(Optional.empty());
         }
 
         @Override
         public SafeFuture<Optional<AttesterDuties>> getAttestationDuties(
             UInt64 epoch, IntCollection validatorIndices) {
-          return null;
+          return SafeFuture.completedFuture(Optional.empty());
         }
 
         @Override
         public SafeFuture<Optional<SyncCommitteeDuties>> getSyncCommitteeDuties(
             UInt64 epoch, IntCollection validatorIndices) {
-          return null;
+          return SafeFuture.completedFuture(Optional.empty());
         }
 
         @Override
         public SafeFuture<Optional<ProposerDuties>> getProposerDuties(UInt64 epoch) {
-          return null;
+          return SafeFuture.completedFuture(Optional.empty());
         }
 
         @Override
         public SafeFuture<Optional<BlockContainer>> createUnsignedBlock(
             UInt64 slot, BLSSignature randaoReveal, Optional<Bytes32> graffiti, boolean blinded) {
-          return null;
+          return SafeFuture.completedFuture(Optional.empty());
         }
 
         @Override
         public SafeFuture<Optional<AttestationData>> createAttestationData(
             UInt64 slot, int committeeIndex) {
-          return null;
+          return SafeFuture.completedFuture(Optional.empty());
         }
 
         @Override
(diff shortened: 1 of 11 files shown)


validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorApiChannel.java
```diff
@@ -42,6 +42,133 @@
 import tech.pegasys.teku.spec.datastructures.validator.SubnetSubscription;
 
 public interface ValidatorApiChannel extends ChannelInterface {
+  ValidatorApiChannel NO_OP =
```
lucassaldanha, line 45: I think returning `null` in the NOOP methods implementations breaks the `ValidatorApiChannel` interface. This can lead to errors and frustration in the future. My suggestion would be to respect the async and optional expectations in the interface. Something like this:

```
      new ValidatorApiChannel() {
        @Override
        public SafeFuture<Optional<GenesisData>> getGenesisData() {
          return SafeFuture.completedFuture(Optional.empty());
        }

        @Override
        public SafeFuture<Map<BLSPublicKey, Integer>> getValidatorIndices(final Collection<BLSPublicKey> publicKeys) {
          return SafeFuture.completedFuture(Map.of());
        }

        [...]
```

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/apis/PostVoluntaryExit.java
```diff
@@ -58,17 +84,79 @@ public PostVoluntaryExit() {
...
+                  final UInt64 slot =
+                      genesisSpec.miscHelpers().computeSlotAtTime(genesisTime, currentTime);
+                  return spec.computeEpochAtSlot(slot);
+                });
+
+    final SafeFuture<SignedVoluntaryExit> future = getVoluntaryExit(publicKey, epoch);
+    request.respondAsync(future.thenApply(AsyncApiResponse::respondOk));
+  }
+
+  private SafeFuture<SignedVoluntaryExit> getVoluntaryExit(
```
lucassaldanha: I think this method could be better named to represent the fact that we are creating a signed voluntary exit. Maybe `createSignedVoluntaryExit()` or something like that?

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/apis/PostVoluntaryExit.java
```diff
@@ -58,17 +84,79 @@ public PostVoluntaryExit() {
...
+      final BLSPublicKey publicKey, final UInt64 epoch) {
+    return validatorApiChannel
```
lucassaldanha: NIT: consider breaking some of this logic into smaller private methods to make this easier to read.

rolfyone: better still, package private that you can test if appropriate :) indices map and how that's handled is a good example of a level where we could feed data in rather than wait for futures etc...

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/apis/PostVoluntaryExit.java
```diff
@@ -58,17 +84,79 @@ public PostVoluntaryExit() {
...
   public void handleRequest(RestApiRequest request) throws JsonProcessingException {
-    request.respondError(SC_NOT_IMPLEMENTED, "Not implemented");
+    final BLSPublicKey publicKey = request.getPathParameter(PARAM_PUBKEY_TYPE);
+    final UInt64 epoch =
+        request
+            .getOptionalQueryParameter(EPOCH_PARAMETER)
+            .orElseGet(
+                () -> {
+                  final SpecVersion genesisSpec = spec.getGenesisSpec();
+                  final UInt64 genesisTime = genesisSpec.getConfig().getMinGenesisTime();
```
rolfyone: not sure we should use min genesis time here... that's not guaranteed to be genesis time

rolfyone: could potentially hit `/eth/v1/beacon/genesis` and memoize that so its only done once, or something like that...

validator/client/src/main/java/tech/pegasys/teku/validator/client/restapi/apis/PostVoluntaryExit.java
```diff
@@ -58,17 +84,79 @@ public PostVoluntaryExit() {
...
+                        final int validatorIndex = indicesMap.get(publicKey);
+                        final Validator validator =
+                            keyManager.getActiveValidatorKeys().stream()
+                                .filter(v -> v.getPublicKey().equals(publicKey))
+                                .findFirst()
+                                .orElseThrow(
+                                    () ->
+                                        new BadRequestException(
+                                            "Validator not found with public key value."));
+                        final VoluntaryExit message =
```
rolfyone: maybe something like "Validator {} is not in the list of keys managed by this service."

PR #7363, opened 2023-07-17
Use `engine_exchangeCapabilities` to log warnings

### LINKED ISSUE #7327: Use `engine_exchangeCapabilities` to log warnings
We no longer call engine methods based on the  `engine_exchangeCapabilities` response after https://github.com/Consensys/teku/pull/7293 . However, we can still use the response from this method to make sure EL supports the methods we expect to be available.

For example when DENEB_FORK_EPOCH is configured for mainnet in CL, we would like ideally EL to return `engine_newPayloadV3` in the exchangeCapabilities response. We could log warnings if not so. This could only be done once on startup to avoid constant polling.

StefanBratanov:
## PR Description

- Enhanced EngineJsonRpcMethodsResolver to return capabilities for all milestones. If a method has been deprecated (added default method in interface) or deleted of course from the code, It will not be returned.
- Create EngineCapabilitiesMonitor which does an exchange capabilities check every 10 epochs (could be too much) against EL.
- It checks engine capabilites against the local capabilities. If some are not supported by EL, a warning will be logged.
- Introduce new parameter `--exchange-capabilities-monitoring-enabled`

## Fixed Issue(s)
fixes #7327 

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/EngineCapabilitiesMonitor.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/EngineCapabilitiesMonitor.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright ConsenSys Software Inc., 2023
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.ethereum.executionlayer;
+
+import com.google.common.base.Suppliers;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import tech.pegasys.teku.ethereum.events.SlotEventsChannel;
+import tech.pegasys.teku.ethereum.executionclient.ExecutionEngineClient;
+import tech.pegasys.teku.ethereum.executionclient.response.ResponseUnwrapper;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.logging.EventLogger;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+
+public class EngineCapabilitiesMonitor implements SlotEventsChannel {
+
+  private static final Logger LOG = LogManager.getLogger();
+
+  private static final int EPOCHS_PER_MONITORING = 10;
+  private static final UInt64 SLOT_IN_THE_EPOCH_TO_RUN_MONITORING = UInt64.valueOf(3);
+
+  private final AtomicReference<UInt64> lastRunEpoch = new AtomicReference<>();
+
+  private final Spec spec;
+  private final EventLogger eventLogger;
+  private final Supplier<List<String>> capabilitiesSupplier;
+  private final ExecutionEngineClient executionEngineClient;
+
+  public EngineCapabilitiesMonitor(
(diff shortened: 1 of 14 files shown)


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/EngineCapabilitiesMonitor.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  private SafeFuture<Void> monitor() {
+    final List<String> capabilities = capabilitiesSupplier.get();
+    return executionEngineClient
+        .exchangeCapabilities(capabilities)
+        .thenApply(ResponseUnwrapper::unwrapExecutionClientResponseOrThrow)
+        .thenAccept(
+            engineCapabilities -> {
+              LOG.debug("Engine API capabilities: " + engineCapabilities);
```
lucassaldanha: Might be worth specifying that these are the method returned from the remote peer. Something like:

```
LOG.debug("Engine API capabilities response: " + engineCapabilities);
```

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/EngineCapabilitiesMonitor.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.logging.log4j.Logger;
+import tech.pegasys.teku.ethereum.events.SlotEventsChannel;
+import tech.pegasys.teku.ethereum.executionclient.ExecutionEngineClient;
+import tech.pegasys.teku.ethereum.executionclient.response.ResponseUnwrapper;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.logging.EventLogger;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+
+public class EngineCapabilitiesMonitor implements SlotEventsChannel {
```
lucassaldanha, line 32: It looks like we don't have a feature toggle for this. Has this been discussed?

lucassaldanha (approved):
LGTM. Just a nit comment and a question.

teku/src/main/java/tech/pegasys/teku/cli/options/ExecutionLayerOptions.java
```diff
@@ -127,6 +128,18 @@ public class ExecutionLayerOptions {
...
 
+  @Option(
+      names = {"--exchange-capabilities-monitoring-enabled"},
+      paramLabel = "<BOOLEAN>",
+      description =
+          "Enables querying EL periodically for the Engine API methods it supports. If incompatibility is detected, there will be a warning raised in the logs.",
+      arity = "1",
+      showDefaultValue = Visibility.ALWAYS,
+      fallbackValue = "true",
+      hidden = true)
```
zilm13: if it's not X, we shouldn't hide it?

StefanBratanov: good catch. I just copied above carelessly.

StefanBratanov: fixed in last commit + fixed arity as well in this and other classes. There were couple of wrong ones.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/MilestoneBasedEngineJsonRpcMethodsResolver.java
```diff
@@ -119,4 +122,13 @@ public <T> EngineJsonRpcMethod<T> getMethod(
     }
     return foundMethod;
   }
+
+  @Override
+  public List<String> getCapabilities() {
```
tbenr: I wonder if in theory we should return `Set` to eliminate potential duplication (if a method is used by multiple milestones)

tbenr (approved):
LGTM. just one comment

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/MilestoneBasedEngineJsonRpcMethodsResolver.java
StefanBratanov: Yes. Makes sense. Will change.

bgravenorst:
Removed the doc-change-required flag because the option is documented here: https://docs.teku.consensys.net/reference/cli#exchange-capabilities-monitoring-enabled

PR #7364, opened 2023-07-17
Openapi node update fix

zilm13:
## PR Description
My guess is obsolete rimraf behavior is changed with node 16 and it doesn't fall silently anymore on this with node 16 (like it did in 14), so this check should fix it

## Documentation

## Changelog


StefanBratanov:
Maybe we can also update lockfile version to 3 as per last PR which was having issues: https://github.com/Consensys/teku/commit/8eb737df45d9b22649f02cdc4d26387cdf30dbce

zilm13:
lockfile was not updated, we don't need to change it probably

StefanBratanov (approved):
happy to try this in CI

Issue #7349, opened 2023-07-11
Add new block publishing API with `broadcast_validation` param


tbenr:
https://github.com/ethereum/beacon-APIs/pull/317

Note: we need to take care of respecting `202` http status code condition.

PR #7371, opened 2023-07-19
Allow different kinds of validation when publishing blocks

tbenr:
It is the easiest approach I come up with to extract the "consensus validation" stage from the block import process.

In summary:
- `ForkChoice.onBlock` now accepts an optional `SafeFuture` to be completed by the method when state transition and blobs availability check succede
- `BlockManager` implementing `BlockImportChannel` now expose the import method with an optional `BroadcastValidation` param. Based on the value, the import flow behaves so that the returned `SafeFuture` completes when the broadcast can be performed. If `empty` is passed, the `SafeFuture` will resolve when import is fully executed (same flow as today)
- `BlockValidator` (now `BlockGossipValidator`) responsible for performing gossip validation, now:
  - exposes the equivocation check, so that it can be done at the end of the consensus check, if required.
  - `validation` method now expect a `isLocal` parameter to know whenever the equivocation data should only be checked (`isLocal == true`, locally produced) or need to be updated as well (`isLocal == true`, block actually coming from a gossip)
- new `BlockValidator` uses `BlockGossipValidator` and consensus validation result to compose the validation flow.

What I don't like much:
- a safe future as parameter, that needs to be resolved by the method itself
alternatives: 
    - wrapping the return in a class that holds the two stages of resolution. But i feel like there is more impact in the whole code base. 
    - use a subscription pattern to subscribe to a consensus validation result even, but i think it makes the coordination of the flow more complex. 
    - pass a `Consumer` instead of a future, as a callback pattern. 

- New block import `FailureReason`s that are not relevant for the actual block import process. But having both consensus SafeFuture and full block import SafeFuture returning the same type makes the flow easier to handle.

needed for #7349

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportChannel.java
```diff
@@ -13,11 +13,20 @@
...
 package tech.pegasys.teku.statetransition.block;
 
+import java.util.Optional;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.events.ChannelInterface;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.logic.common.statetransition.results.BlockImportResult;
 
 public interface BlockImportChannel extends ChannelInterface {
-  SafeFuture<BlockImportResult> importBlock(SignedBeaconBlock block);
```
StefanBratanov, line 22: Could possibly leave this as default and pass Optional.empty()?

tbenr: I thought about that but having a default implementation essentially used for testing, i remember @zilm13 complaining about that. but for sure it will reduce the noise.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -99,9 +99,65 @@ protected SafeFuture<?> doStop() {
...
+    final Optional<SafeFuture<BlockImportResult>> consensusValidationResult;
+    if (broadcastValidationLevel > BroadcastValidation.GOSSIP.ordinal()) {
+      consensusValidationResult = Optional.of(new SafeFuture<>());
+    } else {
+      consensusValidationResult = Optional.empty();
+    }
+
+    // GOSSIP
+    SafeFuture<BlockImportResult> importPipeline =
+        validator
```
StefanBratanov: Hmm wondering if it is an idea to make the BlockValidator accept different types of Validation (gossip, consensus, ...) enum as a parameter and then do the logic there. Can be a separate method broadcastValidate() which returns BroadcastValidationResult, can leave the already existing gossip one untouched.

StefanBratanov: And the gossip validation will just point to the already existing validate inside that class, will just map `InternalValidationResult` to `BroadcastValidationResult`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportChannel.java
rolfyone: I guess the work is done now, and it does kinda make it cleaner to have the 1 interface to support..

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
tbenr: yes it is a nice idea, better separation of duties

tbenr: I think at the end a cleaner solution is to have a second method exposed in `BlockImportChannel`, like `importBlockWithBroadcastValidation` and this method returns a `BroadcastValidationResult`

tbenr: I implemented this spike https://github.com/Consensys/teku/pull/7371/commits/81b0af25a94900c71946237b323e0ba81ff147b0

But I think there is room for improvements following my previous comment. Need also check if my spike is actually correct. (it currently uses the existing `InternalValidationResult`)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
```diff
@@ -136,7 +141,14 @@ public SafeFuture<InternalValidationResult> validate(final SignedBeaconBlock blo
                 return reject("Block signature is invalid");
               }
 
-              if (!receivedValidBlockInfoSet.add(new SlotAndProposer(block))) {
+              // We want to add blocks in the infoSet only when they come from gossip.
+              // When dealing with locally produced block, we only want to check them against seen
+              // block coming from gossip.
```
zilm13: As I understand it (correct me if I'm wrong), you are not adding local block to the seen set here because it will pass Validator twice and it will be actually added to the seen set in another run. I'd clarify this in the comment

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -383,6 +389,7 @@ private SafeFuture<BlockImportResult> onBlock(
       final SignedBeaconBlock block,
       final Optional<BeaconState> blockSlotState,
       final Optional<BlockImportPerformance> blockImportPerformance,
+      final Optional<SafeFuture<BlockImportResult>> consensusValidation,
```
zilm13: Here and on all the way of this var I think we should better clarify it's purpose. Maybe something like `consensusValidationResultCallback`?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidator.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private final BlockGossipValidator blockGossipValidator;
+
+  public BlockBroadcastValidator(final BlockGossipValidator blockGossipValidator) {
+    this.blockGossipValidator = blockGossipValidator;
+  }
+
+  public SafeFuture<BroadcastValidationResult> validate(
+      final SignedBeaconBlock block,
+      final BroadcastValidation broadcastValidation,
```
zilm13: Maybe `BroadcastValidationMode` and the same for the var name? It's not clear that it's just enum and not an action class.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -105,9 +110,40 @@ protected SafeFuture<?> doStop() {
...
+    final SafeFuture<BlockImportResult> consensusValidationResultOrImportFailure =
+        consensusValidationResult.or(importResult);
+
+    final SafeFuture<BlockImportResult> finalImportResult =
+        blockBroadcastValidator
+            .validate(block, broadcastValidation.get(), consensusValidationResultOrImportFailure)
+            .thenCompose(
+                broadcastValidationResult ->
+                    switch (broadcastValidationResult) {
+                      case SUCCESS, CONSENSUS_FAILURE -> importResult;
```
zilm13: Could you, please, add a comment why we are ok to continue with `CONSENSUS_FAILURE` here not returning some immediate result.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -414,6 +421,19 @@ private SafeFuture<BlockImportResult> onBlock(
 
     blobSidecarsAvailabilityChecker.initiateDataAvailabilityCheck();
 
+    final SafeFuture<BlobSidecarsAndValidationResult> blobSidecarsAvailabilityResult =
```
zilm13: My main concern on this that we had block processing, which is not fast and only after that were checking the state of blobSidecar validation. So we had a chance it was run simultaneously (correct me if it's on one thread), now we are blocked by blobSidecar validation.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -130,12 +166,13 @@ public SafeFuture<InternalValidationResult> validateAndImportBlock(
           InternalValidationResult.reject("Block (or its parent) previously marked as invalid"));
     }
 
-    SafeFuture<InternalValidationResult> validationResult = validator.validate(block);
+    final SafeFuture<InternalValidationResult> validationResult =
+        blockGossipValidator.validate(block);
```
zilm13: I'd pass this one through `blockBroadcastValidator`  too, so all production code is using it. It would be cleaner to my mind

zilm13 (commented):
Looks like a good direction for improvement. Some thoughts on it:

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -186,17 +223,18 @@ public void onBlockValidated(SignedBeaconBlock block) {
...
   private void importBlockIgnoringResult(final SignedBeaconBlock block) {
-    doImportBlock(block, Optional.empty()).ifExceptionGetsHereRaiseABug();
+    doImportBlock(block, Optional.empty(), Optional.empty()).ifExceptionGetsHereRaiseABug();
   }
 
   private SafeFuture<BlockImportResult> doImportBlock(
       final SignedBeaconBlock block,
-      final Optional<BlockImportPerformance> blockImportPerformance) {
+      final Optional<BlockImportPerformance> blockImportPerformance,
+      final Optional<SafeFuture<BlockImportResult>> consensusValidation) {
```
lucassaldanha: We need a better name for this variable to clarify that it acts as a "probe" or a callback of consensus validation rules.

Maybe something like `consensusValidationListener` or `onConsensusValidationCompleted`.

lucassaldanha:
I think we are gossiping blocks we receive from the Beacon API w/o any validation rules:
```
@Override
  protected SafeFuture<BlockImportResult> gossipAndImportUnblindedSignedBlock(
      final SignedBlockContainer blockContainer) {
    final SignedBeaconBlock block = blockContainer.getSignedBlock();
    blockGossipChannel.publishBlock(block);
    return blockImportChannel.importBlock(block);
  }
```

If this is the case, all the logic within the `BlockManager` won't prevent the block from being broadcast if invalid. Is this something you know about but is planning to address in a future PR?

tbenr:
> is planning to address in a future PR?

yes, this PR provide the backend for the API change in a following PR.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
tbenr: Consensus validation IS a successful state transition. So I think the bug here is that the consensus validation future can be completed before the state transition starts.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidator.java
zilm13: `BroadcastValidationLevel` another name

lucassaldanha: I like `BroadcastValidationLevel` because it matches the Beacon API spec description for the parameter.
> broadcast_validation: Level of validation that must be applied to a block before it is broadcast.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
tbenr: I think you're actually right that there is a but here. Fixed in the upcoming commit.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
tbenr: done

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockImportChannel.java
```diff
@@ -13,11 +13,19 @@
...
 import tech.pegasys.teku.infrastructure.events.ChannelInterface;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.logic.common.statetransition.results.BlockImportResult;
+import tech.pegasys.teku.statetransition.validation.BlockValidator.BroadcastValidationLevel;
 
 public interface BlockImportChannel extends ChannelInterface {
-  SafeFuture<BlockImportResult> importBlock(SignedBeaconBlock block);
+
+  SafeFuture<BlockImportResult> importBlock(
+      SignedBeaconBlock block, Optional<BroadcastValidationLevel> broadcastValidation);
```
zilm13: Let's name var `broadcastValidationLevel` too

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -105,9 +106,43 @@ protected SafeFuture<?> doStop() {
   }
 
   @Override
-  public SafeFuture<BlockImportResult> importBlock(final SignedBeaconBlock block) {
+  public SafeFuture<BlockImportResult> importBlock(
+      final SignedBeaconBlock block, final Optional<BroadcastValidationLevel> broadcastValidation) {
```
zilm13: I'd update var name too to `broadcastValidationLevel`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
```diff
@@ -0,0 +1,210 @@
...
+      final Spec spec,
+      final RecentChainData recentChainData,
+      final GossipValidationHelper gossipValidationHelper) {
+    this.spec = spec;
+    this.recentChainData = recentChainData;
+    this.gossipValidationHelper = gossipValidationHelper;
+  }
+
+  SafeFuture<InternalValidationResult> validate(
+      final SignedBeaconBlock block, final boolean isLocal) {
```
zilm13: I think `isLocal` could be a bit misleading here, because we don't pass standard locally produced blocks here, maybe we could somehow highlight it in the name of this flag. I don't know, `isLocalUntrusted`?

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTest.java
```diff
@@ -233,6 +238,86 @@ void onBlock_shouldFailIfBlobsAreNotAvailable() {
...
+    consensusValidationResult = Optional.of(new SafeFuture<>());
+    final SignedBlockAndState blockAndState = chainBuilder.generateBlockAtSlot(ONE);
+    storageSystem.chainUpdater().advanceCurrentSlotToAtLeast(blockAndState.getSlot());
+
+    final BlockProcessor blockProcessor = mock(BlockProcessor.class);
+    when(spec.getBlockProcessor(blockAndState.getSlot())).thenReturn(blockProcessor);
+    when(blockProcessor.processAndValidateBlock(any(), any(), any(), any()))
+        .thenThrow(new StateTransitionException("error!"));
+
+    importBlockWithError(blockAndState, FailureReason.FAILED_STATE_TRANSITION);
```
zilm13, line 288: Just noticed the name is a bit confusing. Not a scope of this PR but maybe something like `assertBlockImportedWithError`

beacon/sync/src/testFixtures/java/tech/pegasys/teku/beacon/sync/SyncingNodeManager.java
```diff
@@ -124,9 +125,10 @@ public static SyncingNodeManager create(
             transitionBlockValidator,
             new StubMetricsSystem());
 
-    final BlockValidator blockValidator =
-        new BlockValidator(
+    final BlockGossipValidator blockGossipValidator =
+        new BlockGossipValidator(
             spec, recentChainData, new GossipValidationHelper(spec, recentChainData));
+    final BlockValidator blockBroadcastValidator = new BlockValidator(blockGossipValidator);
```
zilm13: let's rename the var back as you renamed the class

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
```diff
@@ -13,187 +13,91 @@
...
+  public SafeFuture<InternalValidationResult> validateGossip(final SignedBeaconBlock block) {
+    return blockGossipValidator.validate(block, false);
+  }
 
-    if (parentBlockSlot.isGreaterThanOrEqualTo(block.getSlot())) {
-      return completedFuture(reject("Parent block is after child block."));
+  public SafeFuture<BroadcastValidationResult> validateBroadcast(
+      final SignedBeaconBlock block,
+      final BroadcastValidationLevel broadcastValidationLevel,
+      final SafeFuture<BlockImportResult> consensusValidationResult) {
```
zilm13, line 35: Are you sure on this interface? In case of GOSSIP validation we don't need this SafeFuture but still need to provide it.

zilm13 (commented):
a bit more

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
zilm13: But from the other side, BlockGossipValidator shouldn't know too much over his job. So maybe it's ok to leave it as is. Just want to remember what's happening there after 2 months.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
tbenr: ye, you're actually true that we could avoid tracking consensus if validation level is GOSSIP, but the internal validation logic will percolate outside the `BlockValidator` class, with more `if`s in `BlockManager`, and with imperceptible perf benefits. So I'm not sure..

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
tbenr: I don't get `isLocalUntrusted` but maybe `isLocallyProduced`?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockValidator.java
lucassaldanha: IMHO the way we have the rules "pipeline" in here with each step next to each other helps readability. Even if the consensus validations result is not needed for one case, at least things are contained here.

I think we should keep it this way :)

zilm13: The question is whether we have consensusValidationResult when we call this method with `GOSSIP` validation level. If yes - this interface is ok. If we need to pass some SafeFuture stub instead when we have GOSSIP validation, it will stink a bit.

zilm13 (approved):
LGTM

PR #7373, opened 2023-07-20
Create calculate epoch logic in VoluntaryExitDataProvider

courtneyeh:
#7002 

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
@@ -27,7 +27,7 @@ public class VoluntaryExitDataProvider {
     this.timeProvider = timeProvider;
   }
 
-  protected UInt64 calculateCurrentEpoch(final UInt64 genesisTime) {
+  UInt64 calculateCurrentEpoch(final UInt64 genesisTime) {
     final SpecVersion genesisSpec = spec.getGenesisSpec();
     final UInt64 currentTime = timeProvider.getTimeInSeconds();
     final UInt64 slot = genesisSpec.miscHelpers().computeSlotAtTime(genesisTime, currentTime);
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
@@ -13,7 +13,6 @@
 
 package tech.pegasys.teku.validator.client;
 
-import java.util.Optional;
 import tech.pegasys.teku.infrastructure.time.TimeProvider;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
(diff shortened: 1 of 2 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
```diff
@@ -0,0 +1,41 @@
...
+public class VoluntaryExitDataProvider {
+  private final Spec spec;
+  private final TimeProvider timeProvider;
+
+  VoluntaryExitDataProvider(final Spec spec, final TimeProvider timeProvider) {
+    this.spec = spec;
+    this.timeProvider = timeProvider;
+  }
+
+  protected UInt64 getOrCalculateCurrentEpoch(
```
lucassaldanha: I think you don't need this optional here. I'd leave the optional check to whoever is calling this function.

Example:
```
final UInt64 epoch = maybeEpoch.orElse(calculateCurrentEpoch(genesisData.getGenesisTime()));
```

validator/client/src/test/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProviderTest.java
```diff
@@ -0,0 +1,42 @@
...
+  private final StubTimeProvider timeProvider = StubTimeProvider.withTimeInSeconds(100000);
+  private VoluntaryExitDataProvider provider;
+
+  @BeforeEach
+  void setUp() {
+    provider = new VoluntaryExitDataProvider(spec, timeProvider);
+  }
+
+  @Test
+  void calculateExpectedEpochWhenNotProvided() {
```
lucassaldanha: We need to update the name of this method as we don't have the optional parameter anymore.

lucassaldanha (approved):

validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
```diff
@@ -0,0 +1,36 @@
...
+  private final Spec spec;
+  private final TimeProvider timeProvider;
+
+  VoluntaryExitDataProvider(final Spec spec, final TimeProvider timeProvider) {
+    this.spec = spec;
+    this.timeProvider = timeProvider;
+  }
+
+  protected UInt64 calculateCurrentEpoch(final UInt64 genesisTime) {
+    final SpecVersion genesisSpec = spec.getGenesisSpec();
```
rolfyone: This is very close to the misc helper result. I htink the value in this interface that we were discussing was to have it handling the optional epoch, and doing building out that complexity...
so something to the effect of 
```
SafeFuture<Uint64> getRequiredEpoch(Optional<UInt64> maybeEpochParameter) {
 return maybeEpoch.map(SafeFuture::completedFuture).orElse(() -> 
this::calculateCurrentEpoch // which can use the provider i guess if we have it...
}

SafeFuture<UInt64> calculateCurrentEpoch() {
return genesisDataProvider.getGenesisTime()
.thenApply( genesisTime -> 
final UInt64 slot = miscHelpers.computeSlotAtTime(genesisTime, timeProvider.getCurrentTimeMIllis...)
return spec.computeEpochAtSlot(slot);
}
```

that definitely wont compile... anyway was just illustrative but i thought that was kind of idea was the goal, otherwise after this PR all we've done is implemented a call to the misc helper...

lucassaldanha: My comment was to remove the optional parameter because I don't see the value in handling the optional logic here.

Not a blocker for me either way though.

PR #7377, opened 2023-07-21
Add pruning metrics

tbenr:
To early detect possible IO issues, added 2 metrics to track when pruning is active and how long was the last session.

For both blocks and blob sidecars.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7378, opened 2023-07-21
Add performance calculation metrics

tbenr:
We saw several times performance calculation lagging significantly. This PR adds a timing tracking metrics for it.

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
@@ -126,6 +126,7 @@ public class DefaultPerformanceTracker implements PerformanceTracker {
       return;
     }
 
+    final long startTime = System.currentTimeMillis();
     final List<SafeFuture<?>> reportingTasks = new ArrayList<>();
     // Output attestation performance information for current epoch - 2 since attestations can be
     // included in both the epoch they were produced in or in the one following.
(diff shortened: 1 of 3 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
```diff
@@ -140,7 +144,11 @@ public void onSlot(final UInt64 slot) {
       reportingTasks.add(reportSyncCommitteePerformance(currentEpoch));
     }
 
-    SafeFuture.allOf(reportingTasks.toArray(SafeFuture[]::new)).handleException(LOG::error).join();
+    final long startTime = System.currentTimeMillis();
```
StefanBratanov: Just to be more precise I may add the start time before the line `final List<SafeFuture<?>> reportingTasks = new ArrayList<>();` to avoid some of the futures already starting or some additional logic before returning the futures.

StefanBratanov (approved):
LGTM but with one comment

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
tbenr: I don't think what you said is true. I'll argue when I'll be back on the desk.

tbenr: Ok so, future wont start until there is something asking for the result of it (which is ultimately the `join` down here).
But there there are also some synchronous calculations as you mentioned so make sense to start counting time earlier.

PR #7383, opened 2023-07-24
Create getSignedVoluntaryExit method in VoluntaryExitDataProvider

courtneyeh:
#7002 

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
@@ -14,27 +14,66 @@
 package tech.pegasys.teku.validator.client;
 
 import java.util.Optional;
+import java.util.Set;
+import org.apache.tuweni.bytes.Bytes32;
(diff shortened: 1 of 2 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/VoluntaryExitDataProvider.java
```diff
@@ -14,29 +14,68 @@
...
+              final int validatorIndex =
+                  Optional.ofNullable(indicesMap.get(publicKey))
+                      .orElseThrow(
+                          () ->
+                              new BadRequestException(
+                                  String.format(
+                                      "Public key %s is not in the map of validator indices.",
+                                      publicKey)));
+
+              return createSignedVoluntaryExit(validatorIndex, publicKey, epoch, forkInfo);
```
rolfyone, line 75: one of the failure modes is likely to be around the exit epoch being too early. eg. if i get a new validator, and as soon as it becomes active i attempt to process an exit - do we handle that as an error? Is it something we still need to build out?

rolfyone (approved):
LGTM, may currently be a gap in implementation to address, left an open question...

PR #7385, opened 2023-07-25
Fix BlobSidecarsByRange rpc method edge cases

StefanBratanov:
## PR Description
Edge cases captured by @jtraglia on Discord https://discord.com/channels/697535391594446898/1050616638497640548/1133165930575315027

- There is mixed usage of retrieving maxBlobsPerBlock either by slot or getting it directly from spec around the code. Given currently, there is no difference and it is cleaner to get it from the spec directly and avoids the pre deneb 0 count edge case.
- Fixed an arithmetic exception possibility + unit test

## Fixed Issue(s)
N/A

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #7388, opened 2023-07-25
Enable Epoch-states cache by default

rolfyone:
After successful testing on goerli and mainnet, enabling the use of epoch-states cache by default.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7393, opened 2023-07-27
Implement PostVoluntaryExit API

courtneyeh:
#7002 

## Documentation

## Changelog


validator/client/src/test/java/tech/pegasys/teku/validator/client/restapi/apis/PostVoluntaryExitTest.java
```diff
@@ -14,27 +14,73 @@
...
   private final Spec spec = TestSpecFactory.createMinimal(SpecMilestone.CAPELLA);
   private final DataStructureUtil dataStructureUtil = new DataStructureUtil(spec);
+  private final StubRestApiRequest request = new StubRestApiRequest(handler.getMetadata());
+
+  final VoluntaryExit message = new VoluntaryExit(UInt64.valueOf(123), UInt64.ZERO);
+  final SignedVoluntaryExit signedVoluntaryExit =
+      new SignedVoluntaryExit(message, dataStructureUtil.randomSignature());
+
+  @Test
+  void noEpochQueryParamProvided_shouldGiveValidSignedVoluntaryExit()
```
rolfyone: i think really these are just testing that we should call through to the provider with the parameter values.

The 'validSignedVoluntaryExit' is coming just from the fact that we've generated one and passed one out... so all we're really testing here is that the params are correctly passed to the provider...

`shouldCallProviderWhenEpochNotProvided`
`shouldCallProviderWhenEpochProvided`

These might be better names...

rolfyone:
I think we're at the point where we should be adding a changelog entry, as this will be feature complete. I need to still play with the functionality but arguably we can remove experimental flag from the endpoint also.

rolfyone (approved):
LGTM

PR #7401, opened 2023-07-31
Updated the default storage mode for new databases to minimal

### LINKED ISSUE #7392: Change the default storage mode to minimal
This will be easiest in a couple of stages.

For this release we'll create a file in the database folder that stores the mode that the database was opened in. This will allow us to refer to that in future as 'current' mode the database is in.

Future release will change from 'prune' data-storage-mode as default, probably to be null.

Then we'll need logic to reconcile if this is not set, use the one stored to file etc.

Once we've got that logic in place, we'll be able to change the 'default' to be minimal, and existing databases with prune as default will continue to load that from disk and retain there data.

First PR for persisting the current storage mode is #7391.

rolfyone:
Previously initialised databases will remain on prune mode if their storage is not specified.

Makes use of the new data-storage-mode file to allow teku to remember the last storage requested on the subsequent opening.

Can still be specified by data-storage-mode and the storage mode will be updated if this is used.

Fixes #7392

## Documentation

## Changelog


storage/src/test/java/tech/pegasys/teku/storage/server/StorageConfigurationTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public static Stream<Arguments> getStateStorageDefaultScenarios() {
+    ArrayList<Arguments> args = new ArrayList<>();
+    args.add(Arguments.of(false, Optional.empty(), NOT_SET, MINIMAL));
+    args.add(Arguments.of(false, Optional.empty(), ARCHIVE, ARCHIVE));
+    args.add(Arguments.of(false, Optional.empty(), MINIMAL, MINIMAL));
+    args.add(Arguments.of(false, Optional.empty(), PRUNE, PRUNE));
+    args.add(Arguments.of(true, Optional.empty(), NOT_SET, PRUNE));
+    args.add(Arguments.of(true, Optional.of(ARCHIVE), NOT_SET, ARCHIVE));
+    args.add(Arguments.of(true, Optional.of(PRUNE), StateStorageMode.ARCHIVE, ARCHIVE));
```
lucassaldanha: NIT: is this full qualification `StateStorageMode.ARCHIVE` needed? It does not match other entries.

lucassaldanha (approved):

storage/src/test/java/tech/pegasys/teku/storage/server/StorageConfigurationTest.java
rolfyone: looks like something didn't clean up... weird - looking.

PR #7404, opened 2023-07-31
Update BlobSidecarValidator to better match spec

jtraglia:
## PR Description

When comparing the checks in `BlobSidecarValidator#validate` I noticed a few small differences from the spec.

* Some of the validation checks were rephrased since this was originally implemented.
* The reference to `TopicIndexAwareOperationProcessor` was out-of-date.
* The spec has introduced a new helper function `verify_blob_sidecar_signature`.
  * Renamed `isSignatureValidWithRespectToProposerIndex` to `verifyBlobSidecarSignature`.
  * Swapped order of parameters to match spec too.

Reference to checks in the spec:

* https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blob_sidecar_subnet_id

## Documentation

## Changelog


StefanBratanov (approved):
Thanks  🙏 LGTM

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarValidator.java
```diff
@@ -189,22 +188,22 @@ current shuffling (defined by block_parent_root/slot). If the proposer_index can
             });
   }
 
-  private boolean isSignatureValidWithRespectToProposerIndex(
-      SignedBlobSidecar signedBlobSidecar, BeaconState postState) {
+  private boolean verifyBlobSidecarSignature(
+      BeaconState state, SignedBlobSidecar signedBlobSidecar) {
```
rolfyone: i might just tweak these to be final before merge, but looks good...

jtraglia: Good call. Thanks! 👍

PR #7413, opened 2023-08-05
Avoid exception in stable subnet subscriber when discovery is disabled

zilm13:
## PR Description
In some occasions we have discovery disabled, for example in acceptance tests. Let's move to `NOOP`  pattern with stable subnet subscription to avoid exceptions on every slot.
## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7415, opened 2023-08-07
Don't use ByteBuffer when flattening bytes

StefanBratanov:
## PR Description
Fixes observation/comment by @jtraglia https://discord.com/channels/697535391594446898/1050616638497640548/1137117541249060924

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
+++ b/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
@@ -33,7 +33,7 @@ import tech.pegasys.teku.kzg.TrustedSetup;
 
 public class CKZG4844Utils {
 
-  private static final int MAX_BYTES_TO_FLATTEN = 100_000_000; // 100 MB
+  private static final int MAX_BYTES_TO_FLATTEN = 100_663_296; // ~100.66 MB or 768 blobs
 
   public static byte[] flattenBlobs(final List<Bytes> blobs) {
     return flattenBytes(blobs, CKZG4844JNI.getBytesPerBlob() * blobs.size());
--- a/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
+++ b/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
@@ -42,13 +42,13 @@ public class CKZG4844Utils {
   public static byte[] flattenCommitments(final List<KZGCommitment> commitments) {
     return flattenBytes(
         commitments,
-        KZGCommitment::getBytesCompressed,
+        KZGCommitment::toArrayUnsafe,
         CKZG4844JNI.BYTES_PER_COMMITMENT * commitments.size());
   }
 
   public static byte[] flattenProofs(final List<KZGProof> kzgProofs) {
     return flattenBytes(
-        kzgProofs, KZGProof::getBytesCompressed, CKZG4844JNI.BYTES_PER_PROOF * kzgProofs.size());
+        kzgProofs, KZGProof::toArrayUnsafe, CKZG4844JNI.BYTES_PER_PROOF * kzgProofs.size());
   }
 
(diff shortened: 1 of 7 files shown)


infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
```diff
@@ -88,9 +83,19 @@ public static TrustedSetup parseTrustedSetupFile(final String filePath) throws I
...
+  private static <T> byte[] flattenBytes(
+      final List<T> toFlatten, final Function<T, Bytes> bytesConverter, final int capacity) {
+    final byte[] flattened = new byte[capacity];
+    int destPos = 0;
+    for (final T data : toFlatten) {
+      final Bytes bytes = bytesConverter.apply(data);
+      System.arraycopy(bytes.toArray(), 0, flattened, destPos, bytes.size());
+      destPos += bytes.size();
+    }
+    return flattened;
```
zilm13, line 118: Maybe add exception here if capacity doesn't match destPos? It could lead to some weird errors if missed in advance

StefanBratanov: I agree, good point. Added.

jtraglia:
Looks great to me! Thanks!

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/ckzg4844/CKZG4844Utils.java
```diff
@@ -70,27 +70,43 @@ public static TrustedSetup parseTrustedSetupFile(final String filePath) throws I
...
-    final ByteBuffer buffer = ByteBuffer.allocate(capacity);
-    bytes.map(Bytes::toArray).forEach(buffer::put);
-    return buffer.array();
+  private static byte[] flattenBytes(final List<Bytes> toFlatten, final int expectedSize) {
+    return flattenBytes(toFlatten, Function.identity(), expectedSize);
+  }
+
+  private static <T> byte[] flattenBytes(
+      final List<T> toFlatten, final Function<T, Bytes> bytesConverter, final int expectedSize) {
+    final byte[] flattened = new byte[expectedSize];
```
jtraglia, line 105: Should we catch the potential `OutOfMemoryError` here and throw an `IllegalArgumentException` instead?

```java
try {
    final byte[] flattened = new byte[expectedSize];
    // ... rest of the code ...
} catch (OutOfMemoryError e) {
    throw new IllegalArgumentException("expectedSize is too big");
}
```

StefanBratanov: I am not a fan of catching OOM exceptions. I added though a max size of 100 million bytes (100MB) which should be a decent default.

jtraglia: So this isn't enough to batch verify a full BlobSidecarsByRange response at once.

* Right now, `MAX_REQUEST_BLOB_SIDECARS` is 768.
* A blob is 128KiB (4096 * 32) or 131072 bytes.
* 131072 * 768 is 100_663_296, which is just a hair more than 100MB.

The BlobSidecarsByRange handler doesn't do this verification, but I think it's a reasonable limit.

But eh, I'm not sure you'd ever verify all 768 at once. Feel free to ignore.

StefanBratanov: I think it's nicer default to be honest in any case. Changed it to:

```
private static final int MAX_BYTES_TO_FLATTEN = 100_663_296; // ~100.66 MB or 768 blobs
```

jtraglia: Okay, thanks 😄

rolfyone (approved):
LGTM

PR #7419, opened 2023-08-14
tweaked low values of libp2p config

rolfyone:
I was investigating memory and noticed some clients have lower defaults on D_LOW, D_LAZY, HISTORY.

These values reflect that, needs to be built and run on some nodes.

## Documentation

## Changelog


StefanBratanov:
I am not sure exactly why Lighthouse uses values different than default (seems like they do some bandwith optimizations), but Prysm uses the same as us as well as the spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#the-gossip-domain-gossipsub

rolfyone:
all good, we can leave defaults from the config I guess, just figured it was worth looking at :)

PR #7423, opened 2023-08-15
Avoid registration of non-active validators

### LINKED ISSUE #6290: Improve `validatorRegistration` signing process by excluding unknown validators
We currently sign `validatorRegistration` objects for all loaded keys, without excluding validators that are currently unknown (yet without an associated index)

### LINKED ISSUE #7382: Teku VC asks Web3Signer to sign for inactive validators
### Description
Teku VC asks Web3Signer to sign for inactive validators. Let's say I load 10 keys into Teku; 5 that have received 32 ETH and become active, and another 5 that have yet to receive 32 ETH, I see errors in the logs for Web3Signer like:

> `2023-07-24 09:42:15.029+00:00 | vert.x-eventloop-thread-0 | ERROR | LogErrorHandler | Failed request: http://web3signer:9000/api/v1/eth2/sign/0xa60345b3482359171a7aae637b93fcd9245126a07bc58717cbc55502a197bf838e10adb6825f52420994c9f29c913b7d`

Other validator clients correctly recognize the validator is not active and do not ask Web3Signer to sign.

### Steps to Reproduce (Bug)

**Expected behavior:**

Teku should not ask Web3Signer to sign for inactive keys.

**Actual behavior:**

Teku asks Web3Signer to sign for inactive keys.

**Frequency:**

Every epoch.

* Teku v23.6.2
* Web3Signer v23.6.0

zilm13:
## PR Description
- I don't like `OwnedValidators.getActiveValidators()` name because it's misleading. I'd rename to exclude word active, but it will lead to changes in several classes, so I skipped it to omit unnecessary changes in this PR and could do a separate PR with this change.
- Cache looks like not an issue. We will not cache not yet active validator. For exited validator we will not receive any duty but once we try to renew registration, it will be filtered out and will not lead to sign attempt. Cache will not be pruned in this case but it doesn't look an issue too. As validator remains in owned, we will try to further attempts anyway.
- What I don't like is that we do everything on the 3rd slot, so if validator becomes active ongoing in this epoch and luckily become a proposer in first 3 slots, it will not use builder registration for it

I appreciate any ideas how could I test it thoroughly, because it's a sensitive part of the code.

## Fixed Issue(s)

Fixes #7382 
Fixes #6290 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorRegistrator.java
```diff
@@ -177,28 +183,46 @@ private SafeFuture<Void> registerValidators(final List<Validator> validators) {
...
+                  validatorRegistrationBatchSender::sendInBatches);
             });
   }
 
-  private Stream<SafeFuture<SignedValidatorRegistration>> createValidatorRegistrations(
+  private SafeFuture<List<Validator>> filterActiveValidators(final List<Validator> validators) {
+    final Map<BLSPublicKey, Validator> validatorMap =
+        validators.stream().collect(Collectors.toMap(Validator::getPublicKey, Function.identity()));
+    return validatorApiChannel
+        .getValidatorStatuses(validators.stream().map(Validator::getPublicKey).toList())
```
rolfyone: hopefully on a remote validator client this results in 1 api call?
It'd be worth setting up a BN/VC combo and just making sure we only do the 1, or we might end up doing 10k rest api calls, which would be not fun...

zilm13: it was one and I'm currently remaking it to do in batches, because it will fail with 5k validators.
`getValidatorStatuses` allows array of publicKeys as input by API spec

zilm13: I will double check the code that it's a one query when called with the list

rolfyone: might be worth having some timings around it if we don't... This might be why theres a lot of timeouts on goerli...

rolfyone: given it doesnt do that now, happy to leave that out of this review :) might be a few bits in that refactor...

zilm13: I'd check how it works in this style, we will already get a gap between registration calls due to statuses query

zilm13:
Finished reworking it. Now everything in batches. Tried to keep all existing logic.

StefanBratanov (approved):
LGTM

rolfyone (approved):
LGTM

zilm13:
Still testing, don't merge

tbenr:
as discussed on standup:

`ValidatorTimingActions::onSlot` calls `statusLogger.checkValidatorStatusChanges();` at second slot of each epoch

it also happens on startup via `statusLogger.printInitialValidatorStatuses();`.

StefanBratanov:
> 

Hmm good point, so we make duplicate calls.

zilm13:
Yeah, nice catch, detaching it to `ValidatorStatusProvider` so both logger and registrator could use it

zilm13:
Rework finished, some notes:
- start from `ValidatorStatusProvider`
- updated to register pending validators according to the spec [[1]](https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/register_validator.yaml#L13), [[2]](https://ethereum.github.io/builder-specs/#/Builder/registerValidator)
- I think it's a cleaner approach that validator statuses are not queried in `ValidatorApiChannel` when we do registration, though it could be changed and we don't call for statuses twice when we run vc locally

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -300,7 +301,36 @@ public SafeFuture<Void> prepareBeaconProposer(
 
   public SafeFuture<Void> registerValidators(
       SszList<SignedValidatorRegistration> validatorRegistrations) {
-    return validatorApiChannel.registerValidators(validatorRegistrations);
+    return validatorApiChannel
```
tbenr, line 303: I have 2 doubts here:

1- putting filtering here means that it will only be applied when registering from a remote  VC, and it will not be applied in single process.
I see this ValidatorDataProvider as a very thin layer. We should not have business logic unless we really want to differentiate between in process and remote VC handling.

2- if we filter upfront it means that validators in unknown status will not go in our BN data structure. Which means that, if VC doesn't keep calling the registration endpoint, the validator could become known at some point in future. It is an edge case but in theory we loose this info.

tbenr: sorry for `2-`. It is totally irrelevant considering what we are trying to achieve with this PR :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/validator/PostRegisterValidator.java
```diff
@@ -69,8 +69,11 @@ public void handleRequest(final RestApiRequest request) throws JsonProcessingExc
             .handle(
                 (__, error) -> {
                   if (error != null) {
-                    return AsyncApiResponse.respondWithError(
-                        SC_INTERNAL_SERVER_ERROR, error.getMessage());
+                    final String message =
```
tbenr: what about using `ExceptionUtil::getRootCauseMessage` or add another method there if there is nothing currently fitting the need

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorStatusLogger.java
```diff
@@ -13,11 +13,96 @@
...
+      new AtomicReference<>();
+
+  public ValidatorStatusLogger(OwnedValidators validators) {
+    this.validators = validators;
+  }
+
+  @Override
+  public void onNewValidatorStatuses(
+      final Map<BLSPublicKey, ValidatorStatus> newValidatorStatuses) {
+    Map<BLSPublicKey, ValidatorStatus> oldValidatorStatuses =
```
tbenr: `final`

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorStatusLogger.java
```diff
@@ -13,11 +13,96 @@
...
+      }
+      return;
+    }
+
+    // updates
+    for (Map.Entry<BLSPublicKey, ValidatorStatus> entry : newValidatorStatuses.entrySet()) {
+      BLSPublicKey key = entry.getKey();
+      ValidatorStatus newStatus = entry.getValue();
+      ValidatorStatus oldStatus = oldValidatorStatuses.get(key);
+
```
tbenr: `entry` and other locals -> `final`

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorStatusLogger.java
```diff
@@ -13,11 +13,96 @@
...
+          Optional.ofNullable(validatorStatuses.get(publicKey));
+      maybeValidatorStatus.ifPresentOrElse(
+          validatorStatus ->
+              STATUS_LOG.validatorStatus(publicKey.toAbbreviatedString(), validatorStatus.name()),
+          () -> STATUS_LOG.unableToRetrieveValidatorStatus(publicKey.toAbbreviatedString()));
+    }
+  }
+
+  private void printValidatorStatusSummary(Map<BLSPublicKey, ValidatorStatus> validatorStatuses) {
+    Map<ValidatorStatus, AtomicInteger> validatorStatusCount = new HashMap<>();
```
tbenr: `final`

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorStatusLogger.java
```diff
@@ -13,11 +13,96 @@
...
+      if (oldStatus.equals(newStatus)) {
+        continue;
+      }
+      STATUS_LOG.validatorStatusChange(
+          oldStatus.name(), newStatus.name(), key.toAbbreviatedString());
+    }
+  }
+
+  private void printValidatorStatusesOneByOne(
+      Map<BLSPublicKey, ValidatorStatus> validatorStatuses) {
```
tbenr: `final`

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorStatusLogger.java
```diff
@@ -13,11 +13,96 @@
...
+    for (BLSPublicKey publicKey : validators.getPublicKeys()) {
```
tbenr: `final`s

tbenr (commented):
releasing some pending comments while waiting for the fixed version

zilm13:
Reworked to use subscribers model, so `ValidatorStatusProvider` actively checks and updating validator statuses, including new additions. One of subscriber is `ValidatorRegistrator` which checks if recent status update requires registration according to its rules and if it's required, registers validators without checking for statuses again. Also I've promoted log level of some events in `ValidatorRegistrator`, makes sense for me. If it's not, could revert it.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorRegistrationSigningService.java
```diff
@@ -0,0 +1,158 @@
...
+import tech.pegasys.teku.ethereum.execution.types.Eth1Address;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.datastructures.builder.SignedValidatorRegistration;
+import tech.pegasys.teku.spec.datastructures.builder.ValidatorRegistration;
+import tech.pegasys.teku.spec.schemas.ApiSchemas;
+import tech.pegasys.teku.spec.signatures.Signer;
+
+public class ValidatorRegistrationSigningService {
```
tbenr: seems like not a real Service. How about `ValidatorRegistrationCreator`?

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -455,9 +467,9 @@ private void scheduleValidatorsDuties(
       validatorRegistrator.ifPresent(validatorTimingChannels::add);
     }
     addValidatorCountMetric(metricsSystem, validators);
-    this.validatorStatusLogger =
-        new DefaultValidatorStatusLogger(
-            metricsSystem, validators, validatorApiChannel, asyncRunner);
+    final ValidatorStatusLogger validatorStatusLogger = new ValidatorStatusLogger(validators);
+    validatorStatusProvider.subscribeNewValidatorStatuses(
+        validatorStatusLogger::onNewValidatorStatuses);
```
tbenr: what about calling the method `onRefreshedValidatorStatuses`? otherwise it could seem that it is called only when you have a new status. The same applies to the subscription interface

validator/client/src/test/java/tech/pegasys/teku/validator/client/ValidatorStatusProviderTest.java
```diff
@@ -115,11 +128,44 @@ void shouldUpdateMetricsForValidatorStatuses() {
     assertThat(gauge.getValue(ValidatorStatus.active_ongoing.name()))
         .isEqualTo(OptionalDouble.of(0));
 
-    logger.checkValidatorStatusChanges();
+    provider.onSlot(spec.computeStartSlotAtEpoch(UInt64.ONE).plus(1));
```
tbenr, line 135: maybe you should change the name of this test. It is testing that onSlot we do the job of updating statuses, and you could add the check that the subscriber is called.

tbenr (commented):
some comments waiting `onPossibleMissedEvents` fix

tbenr (approved):
LGTM

PR #7436, opened 2023-08-21
Block builder should fail when schema is null

mehdi-aouadi:
## PR Description
Checks if the schema or blinded schema is not null according to the block builder `blinded` property.
An issue was reported on [discord](https://discord.com/channels/697535391594446898/1050616638497640548/1142112922168995900) by @jtraglia:
With the previous implementation, `builder.blindedSchema(validSchema).schema(null).build()` would pass

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
@@ -71,9 +71,11 @@ public class BeaconBlockBodyBuilderBellatrix extends BeaconBlockBodyBuilderAltai
   @Override
   protected void validateSchema() {
     if (isBlinded()) {
-      checkState(blindedSchema != null, "blindedSchema must be set");
+      checkState(
+          blindedSchema != null && schema == null, "blindedSchema must be set with no schema");
     } else {
-      checkState(schema != null, "schema must be set");
+      checkState(
+          schema != null && blindedSchema == null, "schema must be set with no blindedSchema");
     }
   }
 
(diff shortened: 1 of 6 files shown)


jtraglia (approved):
LGTM 👍 thanks for making this PR!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
```diff
@@ -70,7 +70,11 @@ public BeaconBlockBodyBuilder executionPayloadHeader(
 
   @Override
   protected void validateSchema() {
-    checkState(schema != null || blindedSchema != null, "schema or blindedSchema must be set");
+    if (isBlinded()) {
+      checkState(blindedSchema != null, "blindedSchema must be set");
```
rolfyone: to me the 'nit' was we should be doing something like
```
checkState(blindedSchema != null && schema == null, "The blindedSchema must be set, with no schema set.")
```

rolfyone: @jtraglia ?

rolfyone (approved):
marking approved, its a small change if one is required at all...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
mehdi-aouadi: Makes sense. I updated the checks and the tests accordingly

jtraglia: Sorry for the slow reply. Yes, this what I originally had in mind. Thanks.

PR #7437, opened 2023-08-21
Add experimental yamux support (networking)

StefanBratanov:
## PR Description

- Upgrade `jvm-libp2p` to 1.0.1 with improved yamux support in comparison to 1.0.0
- Increase `maxBufferedConnection` writes to 150 MiB as a band-aid fix untill more optimized yamux backpressure is implemented in jvm-libp2p
- Introduce `--Xp2p-yamux-enabled` to enable/disable yamux. Default will be **false** until tested extensively and newer version of libp2p is released with improved backpressure.
- Rename `MplexFirewall` to `MuxFirewall` and made it compatible with yamux frames.
- Add `jitpack.io` repo. It is required for transitive dependency in libp2p (`multibase-java` )
- Explicitly use `netty-codec-http` and `netty-handler` because they are runtime dependencies in libp2p (not exposed by the library)

## Fixed Issue(s)
N/A

## Documentation

## Changelog


gradle/versions.gradle
```diff
@@ -58,10 +58,8 @@ dependencyManagement {
     }
 
     dependencySet(group: 'io.netty', version: '4.1.96.Final') {
```
Nashatyrev: nit: Libp2p has a bit older version `4.1.90` but it seems ok. May upgrade Libp2p on the next release

Nashatyrev (approved):
LGTM

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
```diff
@@ -156,6 +156,15 @@ protected Host createHost() {
           b.getIdentity().setFactory(() -> privKey);
           b.getTransports().add(TcpTransport::new);
           b.getSecureChannels().add(NoiseXXSecureChannel::new);
+
+          // yamux MUST take precedence during negotiation
+          if (config.isYamuxEnabled()) {
+            // TODO: setting maxBufferedConnectionWrites to 150 MiB to handle overflowing the write
```
rolfyone: outstanding todo? is there a ticket?

rolfyone (approved):
LGTM just a todo question

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
lucassaldanha: @StefanBratanov  It looks like the default config is already 150MiB? https://github.com/libp2p/jvm-libp2p/pull/325/files

StefanBratanov: good point, added one https://github.com/Consensys/teku/issues/7532 will reference it in the comment

StefanBratanov: > @StefanBratanov It looks like the default config is already 150MiB? https://github.com/libp2p/jvm-libp2p/pull/325/files

We closed that PR, agreed it's better to override it in Teku instead for the time being, since 150 MiB was excessive as a default.

lucassaldanha: Ah my bad, didn't notice it was closed instead of merged ;)

PR #7449, opened 2023-08-24
load `MIN_EPOCHS_FOR_BLOCK_REQUESTS` from config

tbenr:
This way we:
1. support official configs when running tests
2. support values that are not inline with mainnet calculation

## Documentation

## Changelog


rolfyone (approved):
LGTM

zilm13:
I've checked computed values, while `MIN_VALIDATOR_WITHDRAWABILITY_DELAY` is the same for all (it was overridden in first Holesky test, but looks like the standard number is back in final config), `CHURN_LIMIT_QUOTIENT` varies across configs. Unless we have official config which overrides this, we should keep calculated values I think
```
# mainnet

# 2**8 (= 256) epochs ~27 hours
MIN_VALIDATOR_WITHDRAWABILITY_DELAY: 256

# mainnet:

# 2**16 (= 65,536)
CHURN_LIMIT_QUOTIENT: 65536

# swift:
CHURN_LIMIT_QUOTIENT: 32
# gnosis:
CHURN_LIMIT_QUOTIENT: 4096
# chiado:
CHURN_LIMIT_QUOTIENT: 4096

So calculated:
# swift:
MIN_EPOCHS_FOR_BLOCK_REQUESTS: 272
# gnosis:
MIN_EPOCHS_FOR_BLOCK_REQUESTS: 2304
# chiado:
MIN_EPOCHS_FOR_BLOCK_REQUESTS: 2304
```

tbenr:
@zilm13 thanks for checking. I should have fixed it.
I also temporary disabled the broken ATs and added changelog info.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ For information on changes in released versions of Teku, see the [releases page]
 ### Breaking Changes
 
 - The Development option `--Xp2p-minimum-subnet-subscriptions` has been removed and will no longer be recognised as a command line option.
+- If running separated Beacon Node and Validator node it is required to upgrade Beacon Node first and than Validator Node. A Validator Node running this release won't start if connecting to an older Beacon Node version.
```
zilm13: `then` instead of `than`?

zilm13 (approved):
LGTM

Issue #7422, opened 2023-08-15
Teku fails if web3signer isn't up yet


yorickdowne:
### Description

Teku `v23.8.0` fails but does not exit with `ERROR - PLEASE FIX OR REPORT | Unexpected exception thrown for validator-async-0` when starting it alongside web3signer and web3signer is not yet up. This keeps validators from being loaded, and Teku from executing their duties.

The parameters for remote signing are `--validators-external-signer-url http://web3signer:9000 --validators-external-signer-public-keys external-signer`

### Steps to Reproduce (Bug)

- Configure Teku and web3signer
- Start both at the same time, so Teku looks for web3signer before it can respond

**Expected behavior:** [What you expect to happen]

Teku continues trying web3signer until it can be reached

**Frequency:** [How regularly does it occur?]

Reproducible

### Versions (Add all that apply)
* Software version: v23.8.0
* web3signer version: v23.6.0

### Logs

```
eth-docker-devel-consensus-1  | 2023-08-15 11:03:39.454 INFO  - Eth1 endpoint not provided, using execution engine endpoint for eth1 data
eth-docker-devel-consensus-1  | 2023-08-15 11:03:39.896 ERROR - External signer is currently not reachable at http://web3signer:9000
eth-docker-devel-consensus-1  | 2023-08-15 11:03:39.914 ERROR - External signer is currently not reachable at http://web3signer:9000
eth-docker-devel-consensus-1  | 2023-08-15 11:03:39.945 INFO  - Storing beacon chain data in: /var/lib/teku/beacon
eth-docker-devel-consensus-1  | 2023-08-15 11:03:39.991 INFO  - Using portable BLST library.
eth-docker-devel-consensus-1  | 2023-08-15 11:03:39.998 INFO  - Starting metrics http service on 0.0.0.0:8008
eth-docker-devel-consensus-1  | 2023-08-15 11:03:39.991 ERROR - PLEASE FIX OR REPORT | Unexpected exception thrown for validator-async-0
eth-docker-devel-consensus-1  | java.util.concurrent.CompletionException: tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException: Failed to load public keys from external signer.
eth-docker-devel-consensus-1  | at java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source) ~[?:?]
eth-docker-devel-consensus-1  | at java.util.concurrent.CompletableFuture.completeThrowable(Unknown Source) ~[?:?]
[...]

PR #7452, opened 2023-08-25
Fix failure if web3signer isn't running

courtneyeh:
#7422 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -243,6 +245,17 @@ public static ValidatorClientService create(
                   .propagateTo(validatorClientService.initializationComplete);
               return SafeFuture.COMPLETE;
             })
+        .exceptionally(
+            er -> {
+              Optional<Throwable> maybeCause =
```
rolfyone: final, and lets call 'er' something else? `error` and `err` seem to be common, i think `error` is better...

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -243,6 +245,17 @@ public static ValidatorClientService create(
...
+                  ExceptionUtil.getCause(er, InvalidConfigurationException.class);
+              if (maybeCause.isPresent()) {
+                LOG.warn(maybeCause.get().getMessage());
+              } else {
+                LOG.error("Error was encountered during validator client service start up.", er);
+              }
+              return null;
+            })
         .ifExceptionGetsHereRaiseABug();
```
rolfyone: this cant stay here iirc, we need to remove `ifExceptionGetsHereRaiseABug` - we've handled the error before that...

courtneyeh: Yes just working that one out. The compiler complains if we don't check the result of the `exceptionally()`.

rolfyone: can probably just do a .always wtih log.trace 'finished starting validator client service'

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -243,7 +245,18 @@ public static ValidatorClientService create(
                   .propagateTo(validatorClientService.initializationComplete);
               return SafeFuture.COMPLETE;
             })
-        .ifExceptionGetsHereRaiseABug();
+        .exceptionally(
+            error -> {
+              Optional<Throwable> maybeCause =
```
rolfyone: final...

rolfyone (approved):
LGTM, one nit.

Issue #7438, opened 2023-08-21
Update GetBlobSidecars to check non canoncal blob sidecars when available


mehdi-aouadi:
As per #7258, non canonical blob sidecars could be available when `--data-storage-non-canonical-blocks-enabled` is used.
The `GetBlobSidecars` should check non canonical blob sidecars if available.

PR #7453, opened 2023-08-25
add non canonical blob sidecars fallback

mehdi-aouadi:
## PR Description
Add non canonical blob sidecars fallback in the `GetBlobSidecars` API.
Do not look for blob sidecars for pre Deneb blocks and when there is no kzg commitments in the block.

## Fixed Issue(s)

#7438 
## Documentation

## Changelog


data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
```diff
@@ -90,7 +90,16 @@ public BlobSidecarSelector slotSelector(final UInt64 slot) {
 
   private SafeFuture<Optional<List<BlobSidecar>>> getBlobSidecarsForBlock(
       final Optional<SignedBeaconBlock> maybeBlock, final List<UInt64> indices) {
-    if (maybeBlock.isEmpty()) {
+    if (maybeBlock.isEmpty()
```
zilm13: How about these, would be more readable:
```java
    if (maybeBlock.isEmpty()) {
      return SafeFuture.completedFuture(Optional.empty());
    }

    final Optional<BeaconBlockBodyDeneb> maybeDenebBlock =
        maybeBlock.get().getMessage().getBody().toVersionDeneb();
    if (maybeDenebBlock.isEmpty() || maybeDenebBlock.get().getBlobKzgCommitments().isEmpty()) {
      return SafeFuture.completedFuture(Optional.empty());
    }

```

storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
```diff
@@ -284,7 +284,8 @@ public SafeFuture<Optional<UInt64>> getEarliestAvailableBlobSidecarSlot() {
 
   @Override
   public SafeFuture<Optional<BlobSidecar>> getBlobSidecar(final SlotAndBlockRootAndBlobIndex key) {
-    return SafeFuture.of(() -> database.getBlobSidecar(key));
+    return SafeFuture.of(
+        () -> database.getBlobSidecar(key).or(() -> database.getNonCanonicalBlobSidecar(key)));
```
zilm13: This method is used for example in byRange handler, though it will not iterate by non-canonical keys, it's unnecessary penalty to performance. Plus we have non-canonical method below. I would rather put this `or` to the level higher in `CombinedChainDataClient`

zilm13 (approved):
LGTM just 2 nits

data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
mehdi-aouadi: Can't say no for more readability :)

storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
mehdi-aouadi: Makes sense. Was worried about that use case and missed the RPC handler. We could have a slot without blob sidecars and in that case we shouldn't check the non canonical ones...

zilm13: So it will be called in byRange anyway, I was not right in comment, but byRange will not query non canonical keys, so shouldn't be an issue, just for consistency as far as we have two methods for this in `ChainStorage`, let's combine them on higher level, so we still have choice in `ChainStorage`

data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
mehdi-aouadi: Done

storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java
mehdi-aouadi: Done

PR #7463, opened 2023-08-28
Disconnect EventSubscriber after successive failures.

rolfyone:
If a peer continually fails to accept events from the stream, disconnect them, they can join again when they're ready.

Also reduced the noise on the 'first' failure, as looking at a log this seemed a little excessive.

Finally, in the case where we do get too many IllegalStateExceptions attempting to share information, print a better log message.

## Documentation

## Changelog


tbenr:
I wonder why in `onEvent` we close using `sseClient.ctx().req().getAsyncContext().complete();` and here we are doing simply via `sseClient.close();`.

Is there any technical reason for having it implemented in two different ways?

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/events/EventSubscriber.java
```diff
@@ -124,18 +126,40 @@ private void processEventQueue() {
...
+              if (!queuedEvents.isEmpty()) {
                 processEventQueue();
               }
             })
         .finish(
-            error ->
-                LOG.error(
-                    "Failed to process event queue for client " + sseClient.hashCode(), error));
+            error -> {
+              successiveFailureCounter++;
```
tbenr: I suspect there might be some concurrency problems here.
the previous `.alwasyRun` opens the floodgate again before `.finish` runs, potentially resetting the counter to 0?

rolfyone:
My theory was that at this point the connection is dead, we just haven't figured that out yet ... we just need to terminate it so the client can reconnect

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/events/EventSubscriber.java
rolfyone: that'd be fine, if it resets to 0 we can continue and not nuke the connection... but if it goes over that limit we should really give up and not keep trying to send on blocked ports continuously...

tbenr: k so I'd suggest to add `volatile` to it, just to make sure threads always read the last updated value.

tbenr (approved):
LGTM with just a `volatile`

tbenr (approved):
LGTM. last version seems better. Tried to replicate some connection issues manually, not that easy.

PR #7469, opened 2023-08-29
Fix proposer config refresh

tbenr:
Refactor the proposer config reload logic, removing the (odd) case in which the provider could stuck providing always the same config skipping the refresh.

fixed a unit test too

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/proposerconfig/AbstractProposerConfigProvider.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/proposerconfig/AbstractProposerConfigProvider.java
@@ -35,7 +35,8 @@ public abstract class AbstractProposerConfigProvider implements ProposerConfigPr
   private final TimeProvider timeProvider;
   private Optional<ProposerConfig> lastProposerConfig = Optional.empty();
   private UInt64 lastProposerConfigTimeStamp = UInt64.ZERO;
-  private Optional<SafeFuture<Optional<ProposerConfig>>> futureProposerConfig = Optional.empty();
+  private SafeFuture<Optional<ProposerConfig>> futureProposerConfig =
+      SafeFuture.completedFuture(Optional.empty());
 
   AbstractProposerConfigProvider(
       final AsyncRunner asyncRunner,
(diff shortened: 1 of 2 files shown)


zilm13 (approved):
LGTM
could add to fixed test
```
    verify(proposerConfigLoader, times(2)).getProposerConfig(any());
```
to clarify it's fixed

validator/client/src/main/java/tech/pegasys/teku/validator/client/proposerconfig/AbstractProposerConfigProvider.java
```diff
@@ -67,39 +68,33 @@ public synchronized SafeFuture<Optional<ProposerConfig>> getProposerConfig() {
...
+
+  private synchronized Optional<ProposerConfig> handleException(final Throwable throwable) {
+    if (lastProposerConfig.isPresent()) {
+      LOG.warn("An error occurred while obtaining config, providing last loaded config", throwable);
+      return lastProposerConfig;
+    }
+    throw new RuntimeException(
+        "An error occurred while obtaining config and there is no previously loaded config",
+        throwable);
+  }
```
rolfyone, line 88: why synchronized? and do we have tests showing why the runtimeException should be there? it seems like this might be a problem...

tbenr: > why synchronized? 

we read `lastProposerConfig` there (I want to make sure the thread reads the last value) and I become paranoid due to the unexplained bug :)

tbenr: > runtimeException should be there?

I'm ok with that, because we cannot even start the node if we never loaded a config. So having a valid `lastProposerConfig` present is a must.

tbenr: `ProposerConfigManager::initialize` wants it to complete exceptionally to prevent node to startup

tbenr:
> could add to fixed test

💯 
@zilm13 if it had been there it would have caught the issue

PR #7470, opened 2023-08-30
Added a new signal for when a block is validated but not imported

rolfyone:
This allows event streams to show block messages before head messages.

## Documentation

## Changelog


tbenr:
To me looks good. The only thing we should address it the optimistic flag.

I also noticed that we do not support `payload_attributes` event which has been merged https://github.com/ethereum/beacon-APIs/blob/8b52d949965d2294d3f0191c6b15644535a933fe/apis/eventstream/index.yaml#L35

and lightclient events...

rolfyone:
> and lightclient events...

we've not really addressed lightclient anywhere (basically), so i can't see us doing that until we allocate an epic...

rolfyone:
> I also noticed that we do not support `payload_attributes` event which has been merged https://github.com/ethereum/beacon-APIs/blob/8b52d949965d2294d3f0191c6b15644535a933fe/apis/eventstream/index.yaml#L35

raised #7477

tbenr:
do we want to be set independently? Does make sense to set them both to `true`? I think more that the param should be associated to an `enum` that could be `NONE, VALIDATION, IMPORT`. WDYT?

rolfyone:
> do we want to be set independently? Does make sense to set them both to `true`? I think more that the param should be associated to an `enum` that could be `NONE, VALIDATION, IMPORT`. WDYT?

I think i was going to just let it set separately in case there's another weird use-case...  If they want to get notified for none, or all, or just validation, its all possible with the 2 flags, and simple to flow through but the enumeration would end up ,none, validation, import,validation_and_import' so the checks end up more complicated on the code side because you'd need to check that its in the set that allows validation, and i wasn't sure it was worth it...

tbenr:
I see, but I don't see why there should be the need to get notified twice for the same block. To me the `true, true` combination is odd.

rolfyone:
> I see, but I don't see why there should be the need to get notified twice for the same block. To me the `true, true` combination is odd.

sure it's odd, but I'm not against it if people want to see the block and then actually get the optimistic flag properly once its fully imported.

tbenr:
> sure it's odd, but I'm not against it if people want to see the block and then actually get the optimistic flag properly once its fully imported.

but to get that you just subscribe to `head` too.

Btw its a nit at the end.

tbenr:
so we get this messages for each block when subscribing to block and head, with both flags at true

```
event: block
data: {"slot":"7270564","block":"0xbf34b0c85dd8d56d64db5fc6f3a39a3a35ca1ffa17b9e32c961375ee5049ebe4","execution_optimistic":false}

event: head
data: {"slot":"7270564","block":"0xbf34b0c85dd8d56d64db5fc6f3a39a3a35ca1ffa17b9e32c961375ee5049ebe4","state":"0x1edbea7443751452b723418f882842407018176d74228104449ae6182478108e","epoch_transition":false,"previous_duty_dependent_root":"0x84ac0d9cfc2ab85ad892cc3f26b257f8b89c462cc08ea4364a83efd5c26d4c80","current_duty_dependent_root":"0x01639a86dda3e8b058cadda82a23318b983b9ac3e92ed9c4f965402d84514d46","execution_optimistic":false}

event: block
data: {"slot":"7270564","block":"0xbf34b0c85dd8d56d64db5fc6f3a39a3a35ca1ffa17b9e32c961375ee5049ebe4","execution_optimistic":false}
```

tbenr (approved):
LGTM

rolfyone:
> so we get this messages for each block when subscribing to block and head, with both flags at true

yep the less reliable thing is that first block notification when the head is optimistic etc, but i've got limited options other than searching for the blocks parent...

PR #7471, opened 2023-08-30
Avoid db iteration when it's not needed

zilm13:
## PR Description
In all fixed cases we were looking for latest block in effect and later filtering it by slot, so actually looking for exact slot block, but acquiring db iterator while it's not really needed in these cases. Optimizing a bit. Not sure on the proper checks in the beginning, actually both tries to lookup from store, but if the block is finalized it's not needed.

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
@@ -731,13 +731,9 @@ public class ChainDataProvider {
   }
 
   public SafeFuture<Optional<Bytes32>> getFinalizedBlockRoot(final UInt64 slot) {
-    final SafeFuture<Optional<SignedBeaconBlock>> futureFinalizedBlock =
-        combinedChainDataClient.getFinalizedBlockInEffectAtSlot(slot);
-    return futureFinalizedBlock.thenApply(
-        maybeFinalizedBlock ->
-            maybeFinalizedBlock
-                .filter(block -> block.getSlot().equals(slot))
-                .map(SignedBeaconBlock::getRoot));
+    return combinedChainDataClient
+        .getFinalizedBlockAtSlotExact(slot)
+        .thenApply(maybeFinalizedBlock -> maybeFinalizedBlock.map(SignedBeaconBlock::getRoot));
   }
 
   public SpecMilestone getMilestoneAtHead() {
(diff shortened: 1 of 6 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/ChainDataProvider.java
```diff
@@ -731,13 +731,9 @@ List<Withdrawal> getExpectedWithdrawalsFromState(
...
-    final SafeFuture<Optional<SignedBeaconBlock>> futureFinalizedBlock =
-        combinedChainDataClient.getFinalizedBlockInEffectAtSlot(slot);
-    return futureFinalizedBlock.thenApply(
-        maybeFinalizedBlock ->
-            maybeFinalizedBlock
-                .filter(block -> block.getSlot().equals(slot))
-                .map(SignedBeaconBlock::getRoot));
+    return combinedChainDataClient
+        .getFinalizedBlockAtSlotExact(slot)
+        .thenApply(maybeFinalizedBlock -> maybeFinalizedBlock.map(SignedBeaconBlock::getRoot));
```
rolfyone, line 736: we'd need to be careful if we go down changing these low level functions, hopefully theres god tests...

 - `*atSlotExact` is 'if theres a block at this slot exactly, then i want a block, otherwise i want empty
 - `*inEffectAtSlot` means, i want whatever is the 'current' block at a given slot, so if im at slot N, and there was no block at slot N, but there was one at slot N-1, then my expectation is that i get the block at slot N-1.

By this reasoning, i'm fairly sure that we don't want to make this specific change proposed...

zilm13: we were filtering it lines below:
```
                .filter(block -> block.getSlot().equals(slot))
```
we need exact slot in this function. I understand the difference

zilm13: actually I've broken tests, I don't understand something here, rechecking

zilm13: RecentChainData.getBlockRootBySlot is actually in effect, renamed for clarity, fixed tests

rolfyone: i've actually got a rather ugly stack trace in this area at the moment to look at, so looking if they're related

tbenr (approved):
LGTM

Issue #7464, opened 2023-08-28
Clean up stack traces when execution client is not available


benjaminion:
I started Besu and Teku at the same time. While Besu was starting Teku dumped ugly stack traces. It would be much more friendly to see a one line error or warning message in these circumstances.

```
2023-08-28 07:57:10.319 WARN  - Not loading specified initial state as chain data already exists.
2023-08-28 07:57:10.326 WARN  - Rest API is enabled but no default fee recipient has been configured via the validators-proposer-default-fee-recipient option! It is strongly recommended to configure it to avoid po>
2023-08-28 07:57:11.369 INFO  - BLS: loaded BLST library
2023-08-28 07:57:16.422 INFO  - Completed regeneration of state for block 0xf45ffefc077f61c5dff18f8ef9351577027031604744068124f848c645179891 at slot 7198720 by replaying 32 blocks. Took 6078ms
2023-08-28 07:57:16.605 ERROR - Execution Client request failed. Make sure the Execution Client is online and can respond to requests.
java.net.ConnectException: Failed to connect to /127.0.0.1:8551
at okhttp3.internal.connection.RealConnection.connectSocket(RealConnection.kt:297) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.connection.RealConnection.connect(RealConnection.kt:207) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.kt:226) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.kt:106) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.kt:74) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.connection.RealCall.initExchange$okhttp(RealCall.kt:255) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:32) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109) ~[okhttp-4.11.0.jar:?]
at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95) ~[okhttp-4.11.0.jar:?]
[...]

PR #7472, opened 2023-08-30
Clean up Execution Client request failed error message

courtneyeh:
#7464 

## Documentation

## Changelog


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/web3j/Web3JClient.java
```diff
@@ -167,7 +167,7 @@ private void logExecutionClientError(final Throwable error, final boolean couldB
     if (error instanceof TimeoutException) {
       eventLog.executionClientRequestTimedOut();
     } else {
-      eventLog.executionClientRequestFailed(error, couldBeAuthError);
+      eventLog.executionClientRequestFailed(couldBeAuthError);
```
rolfyone: maybe what we want to do is more like
```
if(Throwables.hasCause(error, ConnectException)) {
eventLog.executionClientConnectFailure();
} else {
eventLog.executionClientRequestFailed(error, couldBeAuthError);
}
```

so that unknown errors still get the trace - we know this one specifically is a connection error ,but no guarantee they are all the same thing at that point...

rolfyone (approved):
LGTM

PR #7482, opened 2023-09-05
Use Java16 `Stream::toList`

tbenr:
I translated `.collect(Collectors.toUnmodifiableList())` and `.collect(Collectors.toList())` to `.toList()`

The assumption is that whenever we used `.collect(Collectors.toList())` was in any case intended to be an immutable list.

https://blogs.oracle.com/javamagazine/post/the-hidden-gems-in-java-16-and-java-17-from-streammapmulti-to-hexformat

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
@@ -23,7 +23,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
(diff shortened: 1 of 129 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
```diff
@@ -399,15 +398,15 @@ private <T> SafeFuture<T> makeRequestToFailoversUntilSuccess(
...
+    final Iterator<? extends RemoteValidatorApiChannel> failoverDelegates =
         beaconNodeReadinessManager.getFailoversInOrderOfReadiness();
     return makeRequestToFailoversUntilSuccess(
         failoverDelegates.next(), failoverDelegates, request, method, capturedExceptions);
   }
 
   private <T> SafeFuture<T> makeRequestToFailoversUntilSuccess(
       final RemoteValidatorApiChannel currentFailoverDelegate,
-      final Iterator<RemoteValidatorApiChannel> failoverDelegates,
+      final Iterator<? extends RemoteValidatorApiChannel> failoverDelegates,
```
StefanBratanov, line 409: Is there a reason for having few changes of those `? extends`

StefanBratanov (approved):
LGTM apart from a question

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/FailoverValidatorApiHandler.java
tbenr: Yes, the root of it is
```java
final List<? extends RemoteValidatorApiChannel> failoverValidatorApis =
    failoverEndpoints.stream()
        .map(
            endpoint ->
                RemoteValidatorApiHandler.create(
                    endpoint,
                    okHttpClient,
                    spec,
                    validatorConfig.isValidatorClientUseSszBlocksEnabled(),
                    asyncRunner))
        .toList();
```

tbenr: This is not the only place it happens, when we want to use the higher level interface as type

StefanBratanov: Got it. I don't mind it. Makes sense.

PR #7483, opened 2023-09-05
Decrease block pruning size

zilm13:
## PR Description
In case of changing storage mode or min epochs for block requests we get thousands of blocks to remove. Removal may freeze node for a minutes and after that DB compaction starts, which could freeze node even more. To avoid such situation we may want to prune old blocks without hurry, only 1 batch per run. Also added configuration option, so the value could be easily overriden.

Some concerns still on this:

0. Maybe I've over engineered it and we don't need boolean return, it will make all the things way simpler
1. Am I missing the case when it could be empty while finalized blocks are really stored? (see old code)
2. We are not pruning non-canonical blocks, should we?
3. Calculations:
  
  blobs: 48 every 1 minutes
  avg 3 in block => 30 epochs in hour, 3.2 times faster than arrival
  1 year of blobs pruned in 4 months
  
  blocks: 5000 every 15 minutes
  20000 in hour => 625 epochs in hour, 66.6 times faster than arrival
  1 year of blocks pruned in 5.5 days

5.5 days for a year of blocks when you change DB storage mode, for example, looks reliable, though we may consider to run in more often in smaller sets. You keep node performance, which could be production node with validators, not a big issue that disk space will be freed only after few days. We should prioritize node stability in this case.
Regarding the blobs, I understand the difference: most of the nodes will keep only required epochs of blobs, but I think we may need to increase this number.

## Documentation

## Changelog


tbenr:
> Regarding the blobs, I understand the difference: most of the nodes will keep only required epochs of blobs, but I think we may need to increase this number.

Considering the fact that almost all users will start by immediately keep the required blobs, I don't think we should change the default since we already have `--Xdata-storage-blobs-pruning-limit` to cover that little percentage of users that may require a change. WDYT?

zilm13:
@tbenr nice point. We may ask to override this value for few users that may need it. It will not affect performance, just free space will not expand as expected. Let's keep blobs as is

storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlockPruner.java
```diff
@@ -100,8 +103,16 @@ private void pruneBlocks() {
...
+      final boolean allPruned =
+          database.pruneFinalizedBlocks(earliestSlotToKeep.decrement(), pruneLimit);
+      if (allPruned) {
+        LOG.info("Finalized blocks before slot {} have been pruned.", earliestSlotToKeep);
+      } else {
+        LOG.info(
+            "{} finalized blocks before slot {} have been pruned, pruning not finished.",
+            pruneLimit,
+            earliestSlotToKeep);
+      }
```
rolfyone: i wonder if a better way to lay this out, is pruneFinalizedBlocks changes to return a UInt64 of the last block deleted...
Basically then we can say something like
'Blocks up to slot N have been deleted', and compare the `earliestSlotToKeep.decrement()` value to the last block deleted to determine if all blocks got deleted (actually thats less easy because empty slots, but if its within an epoch i'd say all were deleted)

thoughts?

tbenr: @zilm13 I like the idea, we can also improve info logs. And seems feasible on both blocks and blobs.

We can also have it on a separate PR to change it on both, if this one become too big.

zilm13: @rolfyone how about this? https://github.com/Consensys/teku/pull/7483/commits/7a41c3a9b6fbbd248dd8140b57764e76b3661521
Even simpler than your suggestion plus we don't really care much how correct is the return value. The only payback is 1 log.info line correctness once in the ages

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -379,68 +380,52 @@ public void deleteHotBlocks(final Set<Bytes32> blockRootsToDelete) {
...
-    } while (blocksToPrune.size() > 0);
-  }
-
-  private void deleteFinalizedBlocks(final Map<UInt64, Bytes32> blocksToPrune) {
+  private void deleteFinalizedBlocks(final List<Pair<UInt64, Bytes32>> blocksToPrune) {
     if (blocksToPrune.size() > 0) {
       if (blocksToPrune.size() < 20) {
-        LOG.debug("Received blocks ({}) to delete", blocksToPrune.keySet());
+        LOG.debug(
+            "Received blocks ({}) to delete", blocksToPrune.stream().map(Pair::getLeft).toList());
```
tbenr: lambda here!

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -379,68 +380,52 @@ public void deleteHotBlocks(final Set<Bytes32> blockRootsToDelete) {
   }
 
   @Override
-  public void pruneFinalizedBlocks(final UInt64 lastSlotToPrune) {
+  public boolean pruneFinalizedBlocks(final UInt64 lastSlotToPrune, final int pruneLimit) {
     final Optional<UInt64> earliestBlockSlot =
         dao.getEarliestFinalizedBlock().map(SignedBeaconBlock::getSlot);
     LOG.debug(
         "Earliest block slot stored is {}",
         earliestBlockSlot.isEmpty() ? "EMPTY" : earliestBlockSlot.get().toString());
```
tbenr: I'd put a lambda here too

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -379,68 +380,52 @@ public void deleteHotBlocks(final Set<Bytes32> blockRootsToDelete) {
...
+    if (blocksToPrune.isEmpty()) {
+      LOG.debug("No finalized blocks to prune up to {} slot", lastSlotToPrune);
+      return true;
     }
     LOG.debug(
-        "Pruning {} finalized blocks, target last slot is {}",
+        "Pruning {} finalized blocks, last slot is {}",
         blocksToPrune.size(),
-        lastSlotToPrune);
+        blocksToPrune.get(blocksToPrune.size() - 1));
```
tbenr: and here?

zilm13: this one is irrelevant anymore with change in return type

tbenr (approved):
LGTM

storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlockPruner.java
rolfyone: that works

PR #7484, opened 2023-09-06
Cleaned up test data generation a little

rolfyone:
- there's now CLI arg tools to generate key lists and blocks.
 - key data has been split out into paginated files, and expanded to 3.2 million keys.

My hope is that the generator apps can be run outside of intellij so that I can generate some larger data sets.

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #7444, opened 2023-08-23
Teku Maven artifacts issues when including as dependency


Nashatyrev:
### Description

When using different versions of `io.spring.dependency-management` gradle plugin including any Teku module artifact as a maven dependency to a gradle build the following error may occur:

- version `1.1.0`: no errors
- version `1.1.1`: build fails with the following error (plugin issue [here](https://github.com/spring-gradle-plugins/dependency-management-plugin/issues/365))
```
> Task :startScripts FAILED
2 actionable tasks: 1 executed, 1 up-to-date
Errors occurred while build effective model from C:\Users\Admin\.gradle\caches\modules-2\files-2.1\tech.pegasys.teku.internal\unsigned\23.8.0\c0e66a134880efb10f2882a13b6cfd8d69b2fe7f\unsigned-23.8.0.pom:

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':startScripts'.
> Error while evaluating property 'relativeClasspath' of task ':startScripts'
   > Cannot invoke "io.spring.gradle.dependencymanagement.org.apache.maven.model.Model.getGroupId()" because "effectiveModel" is null
```
- version `1.1.3`: build succeeds bu the the following error(s) are printed:
```
Errors occurred while building effective model from C:\Users\Admin\.gradle\caches\modules-2\files-2.1\tech.pegasys.teku.internal\unsigned\23.8.0\c0e66a134880efb10f2882a13b6cfd8d69b2fe7f\unsigned-23.8.0.pom:
	'dependencies.dependency[tech.pegasys.teku.internal:unsigned:23.8.0]' for tech.pegasys.teku.internal:unsigned:23.8.0 is referencing itself. in tech.pegasys.teku.internal:unsigned:23.8.0
```

The above errors indicate another problem: all dependencies of `test-fixture` scope are also picked up when only module runtime dependencies are required. Here is some more insights: https://stackoverflow.com/questions/69877418/why-does-every-pom-file-published-by-gradle-has-a-self-referential-dependency

### Steps to Reproduce (Bug)
run `gradle build` for the following `build.gradle`

```gradle
plugins {
    id 'io.spring.dependency-management' version '1.1.1'
    id "java"
    id "application"
}

repositories {
        mavenCentral()
        maven { url "https://artifacts.consensys.net/public/teku/maven/" }
}

dependencies {
    implementation 'tech.pegasys.teku.internal:unsigned:23.8.0'
}
```

[...]

PR #7485, opened 2023-09-06
Fix self-referential dependencies when publishing test fixtures

StefanBratanov:
## PR Description
As suggested by @jframe 

https://github.com/bratkartoffel/security-jwt/commit/48a7e8421aa15c18baab5816d9ddfd75132ae719

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -150,9 +150,6 @@ allprojects {
     }
   }
 
-  components.java.withVariantsFromConfiguration(configurations.testFixturesApiElements) { skip() }
-  components.java.withVariantsFromConfiguration(configurations.testFixturesRuntimeElements) { skip() }
-
   dependencies {
     errorprone("com.google.errorprone:error_prone_core")
     errorprone("tech.pegasys.tools.epchecks:errorprone-checks")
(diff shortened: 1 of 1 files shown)


rolfyone (approved):
LGTM

Nashatyrev:
Relates to #7444 

There are two points about this fix: 
1. It returns back publishing `test-fixture` artifacts which is good 
2. It returns back needless extra dependencies (derived from `test-fixture` scope) for the main artifact 

I don't actually see any crime regarding the (2) except a bunch of unused dependencies are to be downloaded and appended to the classpath.

StefanBratanov:
> Relates to #7444
> 
> There are two points about this fix:
> 
> 1. It returns back publishing `test-fixture` artifacts which is good
> 2. It returns back needless extra dependencies (derived from `test-fixture` scope) for the main artifact
> 
> I don't actually see any crime regarding the (2) except a bunch of unused dependencies are to be downloaded and appended to the classpath.

Didn't realise about (2) annoying really, we will see if there is another workaround, but given we actually have downstream consumers using the `test-fixtures` it is preferrable to be published.

PR #7491, opened 2023-09-07
Increased speed of block generation with some caching

rolfyone:
Wrote some attestation caching for when we're generating good blocks with large block counts.

Reduced time to use block generator for 1.5 mil validators from 330 seconds per block to around 17 seconds per block on my machine.

Renamed the block files to have better information in the file, need to revisit how existing files are used.

Finally, the genesis generation doesnt handle 2 million validators at the current heap size, so need to investigate that, but will as a separate PR.

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

PR #7499, opened 2023-09-12
Updated response of the liveness endpoint

### LINKED ISSUE #7498: liveness endpoint does not conform to the specification
https://github.com/ethereum/beacon-APIs/blob/205bb822e920fcbbf6db0ffa021b83c505299490/apis/validator/liveness.yaml#L39

```
            title: PostLivenessResponseBody
            type: object
            properties:
              data:
                type: array
                items:
                  type: object
                  properties:
                    index:
                      $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
                    is_live:
                      type: boolean
```

Teku seems to throw in in an extra `epoch` field which is not part of the standard.

rolfyone:
Removed epoch, which was not in the spec. Listed as a breaking change because our liveness endpoint has always returned it, despite it being off-spec.

fixes #7498

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM
We could maybe rename `ValidatorLivenessAtEpoch` to `ValidatorLiveness`

PR #7506, opened 2023-09-13
Raised error if a rest api fails to start

rolfyone:
This will allow us to fail fast rather than potentially silently swallow errors.

at least partially addresses #7502 which saw no 'listening on' for validator api.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/RestApi.java
+++ b/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/RestApi.java
@@ -65,6 +65,18 @@ public class RestApi extends Service {
                 "TCP Port %d is already in use. "
                     + "You may need to stop another process or change the HTTP port for this process.",
                 app.port()));
+      } else if (e instanceof IllegalStateException
+          || Throwables.getRootCause(e) instanceof IllegalStateException) {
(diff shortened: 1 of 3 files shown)


mehdi-aouadi (approved):
LGTM

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/RestApi.java
```diff
@@ -65,6 +65,18 @@ protected SafeFuture<?> doStart() {
                 "TCP Port %d is already in use. "
                     + "You may need to stop another process or change the HTTP port for this process.",
                 app.port()));
+      } else if (e instanceof IllegalStateException
+          || Throwables.getRootCause(e) instanceof IllegalStateException) {
+        // IllegalStateException is a sign that something needed has failed to be initialised.
+        // throwing it here will terminate the process effectively.
+        LOG.error("Failed to start Rest API", e);
+        throw e;
+      } else if (app.jettyServer() == null || !app.jettyServer().started) {
```
tbenr, line 74: we could move the `app.jettyServer() == null` check right after `app.start()` so we can remove all intelliJ's complains

rolfyone: i did try this but its still messy in intellij :(

PR #7508, opened 2023-09-13
Other `Stream::toList` conversions

tbenr:
Found other `Collector.toList()`s when we were importing it as a static function

## Documentation

## Changelog


zilm13 (approved):
LGTM

rolfyone (approved):
LGTM - build not happy though... :)

tbenr:
there was another subtle situation where we need a mutable list but the UT was not catching the exception

PR #7510, opened 2023-09-14
Fix issue where remote config is missing data

### LINKED ISSUE #7492: interop issue with constants that have moved to the network configuration
ref. #7489

When we are loading from a spec url, if the moved constants aren't available, we should still continue, and just reference the old constant values.

This is likely to need to be done until Deneb, because we can't guarantee currently the version of a client that someone is connecting to has these same settings.

courtneyeh:
fixes #7492 

## Documentation

## Changelog


ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/TestSpecFactory.java
```diff
@@ -343,19 +343,30 @@ private static SpecConfigDeneb getDenebSpecConfig(final Eth2Network network) {
...
+        network, altairForkEpoch, bellatrixForkEpoch, capellaForkEpoch, denebForkEpoch, true);
+  }
+
+  public static SpecConfigDeneb getDenebSpecConfig(
+      final Eth2Network network,
+      final UInt64 altairForkEpoch,
+      final UInt64 bellatrixForkEpoch,
+      final UInt64 capellaForkEpoch,
+      final UInt64 denebForkEpoch,
+      final boolean kzgNoop) {
```
rolfyone: seems very curious we've had to make this change? doesnt seem related to anything we're doing...

teku/src/test/java/tech/pegasys/teku/cli/subcommand/RemoteSpecLoaderTest.java
```diff
@@ -55,6 +64,29 @@ void shouldFailWhenRequiredItemsAreMissing() {
...
+        Resources.toString(
+            Resources.getResource(RemoteSpecLoaderTest.class, "spec_config.json"),
+            StandardCharsets.UTF_8);
+    final ObjectMapper objectMapper = new ObjectMapper();
+    TypeReference<Map<String, String>> typeReference = new TypeReference<>() {};
+    Map<String, String> data = objectMapper.readValue(jsonConfig, typeReference);
+
+    final SpecConfig specConfig = SpecConfigLoader.loadRemoteConfig(data);
+
+    final SpecConfig expectedConfig =
```
rolfyone: basically instead of doing this, you'd just create a mainnet spec, and get the genesis spec config from that...

teku/src/test/java/tech/pegasys/teku/cli/subcommand/RemoteSpecLoaderTest.java
```diff
@@ -55,6 +64,29 @@ void shouldFailWhenRequiredItemsAreMissing() {
...
+        TestSpecFactory.getDenebSpecConfig(
+            Eth2Network.MAINNET,
+            UInt64.valueOf(74240),
+            UInt64.valueOf(144896),
+            UInt64.valueOf(194048),
+            UInt64.valueOf("18446744073709551615"),
+            false);
+    assertThat(specConfig).isEqualTo(expectedConfig);
```
rolfyone: rather than comparing the entire spec, all we need to know is that the load didnt cause an exception (it couldnt have to get this far) and that the fields that were missing are now set with the correct values. so we can check max_chunk_size etc explicitly from our mainnet spec we internally generate against the one we load remotely.

teku/src/test/resources/tech/pegasys/teku/cli/subcommand/spec_config.json
```diff
@@ -0,0 +1,97 @@
+{
+  "CONFIG_NAME": "mainnet",
+  "PRESET_BASE": "mainnet",
```
rolfyone: we might name this better... something like `config_missing_network_fields.json` so that we know what it is

rolfyone (approved):
LGTM

Issue #7503, opened 2023-09-12
web3signer remote registration, unable to delete keys


yorickdowne:
### Description

I am starting Teku with `--validators-external-signer-url http://web3signer:9000 --validators-external-signer-public-keys external-signer` so that keys loaded into web3signer come in on startup, and users do not need to manually register these keys if they switched consensus layer clients.

I also register the keys with Teku when loading them into web3signer, so that they are immediately live.

This creates an issue when trying to delete the remote registration.

- Load keys into web3signer and register as remote key with Teku, via keymanager API
- Restart Teku, it loads keys from the web3signer URL
- Attempt to delete keys and delete remote registration, this fails. Likely because the remote registration was created upon startup from the web3signer URL:

```
Remote registration for validator 0x8fe5b2c0ce05093e3c6bbe1db0363b542709dfb4b8433e13ce23f1e7a7d840cc09a2ad97c5fd03980988df48bbe840f2 was found but an error was encountered trying to delete it:
Cannot remove read-only validator
```

- Restart Teku again. It would now not get a key list from web3signer, but it still sees the remote registration that was created on key import:

```
No keys loaded into web3signer
Remote keys registered with consensus
0x9642bbe4853617cb603863e4c90398304724ad192c4d6337fded814a6ab0c04ede5dafd9d9148ae4c0d1a51354a17687 http://web3signer:9000
0x8fe5b2c0ce05093e3c6bbe1db0363b542709dfb4b8433e13ce23f1e7a7d840cc09a2ad97c5fd03980988df48bbe840f2 http://web3signer:9000
```

This is worrisome for users: They deleted keys but still see the registration. One user came to me with slashing fears.

**Expected behavior:**

I'd like to be able to remove registrations that I created, or that were created from web3signer when Teku starts, so key deletion also removes the remote registration from Teku.

I'd like to keep the behavior where keys are immediately live on import, which is why they are imported into web3signer and then registered with Teku, both via keymanager API.

**Frequency:**

Reproducible

### Versions (Add all that apply)
* Software version: `teku/v23.9.0/linux-x86_64/-eclipseadoptium-openjdk64bitservervm-java-17`

rolfyone:
any Key that is added outside of keymanager is 'read only', because each startup it's going to load from the configuration and we have no control to remove it.
It can be very confusing attempting to load keys via config AND via keymanager, so my initial suggestion would be avoiding attempting to do that.

[...]

PR #7511, opened 2023-09-14
Added a feature to remove keys on reload

rolfyone:
Behind a feature toggle. Removes keys in the same way that the keymanager does, to ensure that the signer can't be referenced once the key has been removed.

Previously the HUP signal was purely additive, so you couldn't actually remove keys without a restart.

Made a script that allows removal of keys, so that i can add, reload, remove, reload with that interop folder.

Offers a solution to #7503

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
@@ -13,6 +13,8 @@
 
 package tech.pegasys.teku.validator.client.loader;
 
+import static tech.pegasys.teku.infrastructure.logging.StatusLogger.STATUS_LOG;
+
 import com.google.common.annotations.VisibleForTesting;
 import java.net.URL;
 import java.net.http.HttpClient;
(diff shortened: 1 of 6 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -253,14 +264,47 @@ static ValidatorLoader create(
...
+  }
+
+  private void removeKeysNoLongerValid() {
+    final List<BLSPublicKey> currentKeySet =
+        validatorSources.stream()
+            .map(ValidatorSource::getAvailableValidators)
+            .flatMap(List::stream)
+            .map(ValidatorProvider::getPublicKey)
+            .toList();
+    final List<BLSPublicKey> removeKeys =
```
tbenr: I'd call it `keysToBeRemoved`

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -253,14 +264,47 @@ static ValidatorLoader create(
...
         mutableLocalValidatorSource,
         mutableExternalValidatorSource,
         graffitiProvider,
         maybeDataDirLayout,
-        slashingProtectionLogger);
+        slashingProtectionLogger,
+        removeValidatorsOnReload);
+  }
+
+  private void removeKeysNoLongerValid() {
```
tbenr: `valid` seems like too strong but I don't have a better suggestion

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -55,24 +57,32 @@ public class ValidatorLoader {
...
+
   private ValidatorLoader(
       final List<ValidatorSource> validatorSources,
       final Optional<ValidatorSource> mutableLocalValidatorSource,
       final Optional<ValidatorSource> mutableExternalValidatorSource,
       final GraffitiProvider graffitiProvider,
       final Optional<DataDirLayout> maybeDataDirLayout,
-      final SlashingProtectionLogger slashingProtectionLogger) {
+      final SlashingProtectionLogger slashingProtectionLogger,
+      final boolean remooveValidatorsOnReload) {
```
tbenr: double `oo`

tbenr (approved):
LGTM. Just nits. Haven't tried the script.
Tested locally with keys on FS and works.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
tbenr: maybe `removeKeysNoLongerInConfig`, similarly to the printed log

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
rolfyone: oopsie!

rolfyone:
I broke this fairly easily, closing for now, will have to re-evaluate.

PR #7514, opened 2023-09-14
Use `safeJoin` in all UT

tbenr:
Let's use `safeJoin` instead of `join` in all Unit Tests.

Also added explicit mutable collection when we need a **mutable** List from a Stream (`Collectors.toList()` doesn't guarantee it, in theory). Checkout https://github.com/Consensys/teku/pull/7514/commits/373650a858342b37202f0fa301b7ecc30ec6bf90

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7523, opened 2023-09-18
Add Boolean.class in SpecConfigReader

usmansaleem:
## PR Description
Add Boolean.class in SpecConfigReader to allow parsing of custom network config files that use boolean options such as `KZG_NOOP: true`

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/SpecConfigReader.java
```diff
@@ -103,6 +103,7 @@ public class SpecConfigReader {
           .put(Bytes4.class, fromString(Bytes4::fromHexString))
           .put(Bytes32.class, fromString(Bytes32::fromHexStringStrict))
           .put(boolean.class, fromString(Boolean::valueOf))
```
rolfyone: do we need both this and 105? was 105 just wrong?

usmansaleem: I am not 100% sure whether `boolean.class` can be invoked by the jackson/yaml parser? Since leaving it there would cover both cases i.e. Boolean and boolean.

StefanBratanov: I think we should make it similar to how we do for other primitives:
```
.put(Boolean.TYPE, fromString(Boolean::valueOf))
.put(Boolean.class, fromString(Boolean::valueOf))
```

rolfyone: that makes perfect sense, there did seem to be Long above that we could use as the basis for this change, potentially move it up with the other primitives...

usmansaleem: done 👍🏼

rolfyone (approved):
LGTM

PR #7524, opened 2023-09-18
Disable Kzg when it's not needed

zilm13:
## PR Description
When Deneb is enabled and vc is started separately it throws on `trustedSetupPath` missing. Actually it's not needed on vc side and for the most subcommands. Disabled

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/subcommand/ValidatorClientCommand.java
```diff
@@ -158,6 +158,6 @@ public TekuConfiguration tekuConfiguration() {
     loggingOptions.configureWireLogs(builder);
     interopOptions.configure(builder);
     metricsOptions.configure(builder);
-    return builder.build();
+    return builder.build(false);
```
rolfyone: I'm not a huge fan of passing in context into build() functions. I'd prefer that we have parameters on the builder that we can set, then call build with no parameters, which is a much more 'normal' builder pattern...

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -254,6 +254,10 @@ public void spec(Spec spec) {
     }
 
     public Eth2NetworkConfiguration build() {
+      return build(true);
+    }
+
+    public Eth2NetworkConfiguration build(final boolean forBeaconNode) {
```
rolfyone: i wrote the other comment below when i was looking for this, but basically, have 
`Eth2NetworkConfigurationBuilder.forBeaconNode(true/false)` and default it to whatever makes sense, then build can continue to have no parameter.

zilm13: yeah I passed it explicitly in some cases to clarify it

teku/src/main/java/tech/pegasys/teku/cli/subcommand/ValidatorClientCommand.java
tbenr: ye I agree, another option could be to have two different `build`, `buildForBeaconNode` `buildForValidatorClient` and drop the simple `build`

zilm13: My point on this was that options are the same for bn and vc but we don't need the same config for vc and most subcommands as we need for bn, but I'm not 100% on this, yeah, especialy if it'll branch in the future, this approach is definitely bad, I will check how this could look like in atomic kzg disabling for this case

zilm13:
@rolfyone I've reworked the PR to avoid passing anything to the builder.
Plus, @tbenr discovered that PR still fails with `--network=auto` which surfaced another failing area - spec building. Fixed it too.

tbenr (approved):
LGTM. Just `DENEB_KZG_STUB` could be `DENEB_KZG_NOOP`, unless you have an argument :)

rolfyone:
actually is this the actual kzg_noop config item? because if it is we should maybe leave it?

zilm13:
@tbenr done
@rolfyone what do you mean by leave it?

tbenr (approved):
LGTM

PR #7522, opened 2023-09-18
Add max epoch activation churn limit for Deneb

StefanBratanov:
## PR Description
- Introduced `getValidatorActivationChurnLimit` in `BeaconStateAccessors` and overrode it in `BeaconStateAccessorsDeneb`
- Added `MAX_PER_EPOCH_ACTIVATION_CHURN_LIMIT` to `SpecConfigDeneb` to all config files, which have deneb config params already.
- Enabled `--Xfork-choice-proposer-boost-uniqueness-enabled` to true by default
- Bumped up ref test version to `v1.4.0-beta.2-hotfix`
- Disabled running the new Deneb fork-choice tests. Raised https://github.com/Consensys/teku/issues/7539 to support those tests.
- Changed one of the post.ssz in FuzzUtilTest because the post state has changed with the new release - https://github.com/ethereum/consensus-spec-tests/tree/master/tests/minimal/capella/operations/attester_slashing/pyspec_tests/basic_surround

## Fixed Issue(s)
fixes #7520 

## Documentation

- [X] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [X] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #7528, opened 2023-09-19
Apply proposer boost to first block in case of equivocation

StefanBratanov:
## PR Description
Implements changes as per https://github.com/ethereum/consensus-specs/pull/3352

- Introduces a temporary feature flag `--Xfork-choice-proposer-boost-uniqueness-enabled` defaulting to false. Will enable in #7522  which will update the ref tests.
- Manually set the flag to true in ForkChoiceTest instead of relying on the static const to make sure unit tests are modified and passing with the flag set to true. Will use the static const in #7522 PR as well.
- Edited couple of unit tests which were checking reorg logic but they were based on having chains with same slots, which is not typically how reorgs happen unless shuffling changes and test was not working because proposer boost was applied only to first block for that slot. So changed tests to create forks with one empty slot difference.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -587,7 +587,7 @@ public class ForkChoice implements ForkChoiceUpdatedResultSubscriber {
       if (isBeforeAttestingInterval(millisPerSlot, timeIntoSlotMillis)) {
         if (forkChoiceProposerBoostUniquenessEnabled) {
           // is_first_block
-          return transaction.getProposerBoostRoot().isEmpty();
+          return recentChainData.getStore().getProposerBoostRoot().isEmpty();
         }
         return true;
       }
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -543,7 +543,7 @@ public class ForkChoice implements ForkChoiceUpdatedResultSubscriber {
         blobSidecars,
         earliestBlobSidecarsSlot);
 
-    if (applyProposerBoost(block, transaction)) {
+    if (shouldApplyProposerBoost(block, transaction)) {
       transaction.setProposerBoostRoot(block.getRoot());
     }
 
@@ -578,21 +578,25 @@ public class ForkChoice implements ForkChoiceUpdatedResultSubscriber {
     return result;
   }
 
-  private boolean applyProposerBoost(
+  // from consensus-specs/fork-choice:
+  // get_current_slot(store) == block.slot and is_before_attesting_interval and is_first_block
+  private boolean shouldApplyProposerBoost(
       final SignedBeaconBlock block, final StoreTransaction transaction) {
-    if (spec.getCurrentSlot(transaction).equals(block.getSlot())) {
-      final UInt64 millisPerSlot = spec.getMillisPerSlot(block.getSlot());
-      final UInt64 timeIntoSlotMillis = getMillisIntoSlot(transaction, millisPerSlot);
-
-      if (isBeforeAttestingInterval(millisPerSlot, timeIntoSlotMillis)) {
-        if (forkChoiceProposerBoostUniquenessEnabled) {
-          // is_first_block
-          return recentChainData.getStore().getProposerBoostRoot().isEmpty();
-        }
-        return true;
-      }
+    if (!spec.getCurrentSlot(transaction).equals(block.getSlot())) {
+      return false;
     }
-    return false;
+
+    final UInt64 millisPerSlot = spec.getMillisPerSlot(block.getSlot());
+    final UInt64 timeIntoSlotMillis = getMillisIntoSlot(transaction, millisPerSlot);
+    if (!isBeforeAttestingInterval(millisPerSlot, timeIntoSlotMillis)) {
+      return false;
+    }
+
+    if (forkChoiceProposerBoostUniquenessEnabled) {
+      return transaction.getProposerBoostRoot().isEmpty();
+    }
+
+    return true;
   }
 
   /**
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -93,6 +93,7 @@ public class ForkChoice implements ForkChoiceUpdatedResultSubscriber {
   private final ForkChoiceNotifier forkChoiceNotifier;
   private final MergeTransitionBlockValidator transitionBlockValidator;
   private final boolean forkChoiceUpdateHeadOnBlockImportEnabled;
+  private final boolean forkChoiceProposerBoostUniquenessEnabled;
   private final AttestationStateSelector attestationStateSelector;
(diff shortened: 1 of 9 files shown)


teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
```diff
@@ -166,6 +166,19 @@ public class Eth2NetworkOptions {
   private boolean forkChoiceUpdateHeadOnBlockImportEnabled =
       Eth2NetworkConfiguration.DEFAULT_FORK_CHOICE_UPDATE_HEAD_ON_BLOCK_IMPORT_ENABLED;
 
+  // can be removed after all clients have rolled out the spec change:
+  // https://github.com/ethereum/consensus-specs/pull/3352
```
lucassaldanha: Do we have a ticket to review this?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -582,6 +578,23 @@ private BlockImportResult importBlockAndState(
...
+        if (forkChoiceProposerBoostUniquenessEnabled) {
+          // is_first_block
+          return recentChainData.getStore().getProposerBoostRoot().isEmpty();
+        }
+        return true;
+      }
+    }
+    return false;
+  }
+
```
lucassaldanha: NIT: This is a suggestion on how to rewrite this logic w/o the nested if statements. I know that not everyone is a fan of multiple exit points in a method but this makes it easier to add/remove rules to the proposer boost logic (and to me also clearly shows the distinct rules).

```
  private boolean applyProposerBoost(
      final SignedBeaconBlock block, final StoreTransaction transaction) {
    if (!spec.getCurrentSlot(transaction).equals(block.getSlot())) {
      return false;
    }

    final UInt64 millisPerSlot = spec.getMillisPerSlot(block.getSlot());
    final UInt64 timeIntoSlotMillis = getMillisIntoSlot(transaction, millisPerSlot);
    if (!isBeforeAttestingInterval(millisPerSlot, timeIntoSlotMillis)) {
      return false;
    }

    final boolean isFirstBlock = recentChainData.getStore().getProposerBoostRoot().isEmpty();
    if (forkChoiceProposerBoostUniquenessEnabled && !isFirstBlock) {
      return false;
    }

    return true;
  }
```

Potentially, we could move `isFirstBlock` into the if statement to avoid evaluating it w/o the need to do so, for the sake of readability. But the toggle will be removed soon so we will need to perform this check all the time anyway.

lucassaldanha (commented):
LGTM. I'd like to have a second pair of eyes looking into this.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
lucassaldanha: Here is another suggestion if we want to stay closer to the spec. Less optimized but easier to read imho:
```
  private boolean applyProposerBoost(
      final SignedBeaconBlock block, final StoreTransaction transaction) {
    final UInt64 millisPerSlot = spec.getMillisPerSlot(block.getSlot());
    final UInt64 timeIntoSlotMillis = getMillisIntoSlot(transaction, millisPerSlot);

    final boolean isBlockForCurrentSlot = spec.getCurrentSlot(transaction).equals(block.getSlot());
    final boolean isBlockBeforeAttestingInterval = isBeforeAttestingInterval(millisPerSlot, timeIntoSlotMillis);
    final boolean isFirstBlock =
        forkChoiceUpdateHeadOnBlockImportEnabled ? recentChainData.getStore().getProposerBoostRoot().isEmpty() : true;

    return isBlockForCurrentSlot && isBlockBeforeAttestingInterval && isFirstBlock;
  }
```
I don't love the ternary condition there, but it will probably be removed once we don't need that booster flag anymore.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -150,6 +153,7 @@ public ForkChoice(
         new TickProcessor(spec, recentChainData),
         transitionBlockValidator,
         true,
+        true,
```
rolfyone, line 156: assuming this is just a testing constructor if we're hard coding here?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -582,6 +578,23 @@ private BlockImportResult importBlockAndState(
     return result;
   }
 
+  private boolean applyProposerBoost(
```
rolfyone: given this returns boolean, i'd prefer to rename it... something like `isApplyProposerBoost`

`apply` suggests its changing something, but what this function doing is determine whether we *should* apply proposer boost.

I'd potentially go as far as referring to the spec formula we're computing by putting a comment above it like
```
// from consensus spec (maybe deneb/fork_choice?):
// if get_current_slot(store) == block.slot and is_before_attesting_interval and is_first_block:
//        store.proposer_boost_root = hash_tree_root(block)
private boolean isApplyProposerBoost(final SignedBeaconBlock block, final StoreTransaction transaction) {
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
StefanBratanov: yeah this is just old deprecated constructor used in some old tests

StefanBratanov:
Thanks for the feedback from both. :) Did changes with a mix of all of them.

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
StefanBratanov: Thanks. Raised one https://github.com/Consensys/teku/issues/7537 and changed the comment to point to issue instead.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -582,6 +578,27 @@ private BlockImportResult importBlockAndState(
...
+    if (!isBeforeAttestingInterval(millisPerSlot, timeIntoSlotMillis)) {
+      return false;
+    }
+
+    if (forkChoiceProposerBoostUniquenessEnabled) {
+      return transaction.getProposerBoostRoot().isEmpty();
+    }
+
+    return true;
+  }
```
lucassaldanha: NIT: I like this idea, but maybe we can have the comment on top of each condition?

e.g.

```
// get_current_slot(store) == block.slot
if (!spec.getCurrentSlot(transaction).equals(block.getSlot())) {

[...]

// is_before_attesting_interval
if (!isBeforeAttestingInterval(millisPerSlot, timeIntoSlotMillis)) {

[...]

// is_first_block
if (forkChoiceProposerBoostUniquenessEnabled) {
  return transaction.getProposerBoostRoot().isEmpty();
}
```

lucassaldanha (approved):
LGTM. Just another NIT suggestion.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
StefanBratanov: I like it. Done.

PR #7529, opened 2023-09-19
add block v3 API

### LINKED ISSUE #7390: Add block v3 endpoint
Spec: https://github.com/ethereum/beacon-APIs/pull/339

- [x] Add Get Block V3 API
- [x] Add `RemoteValidatorApiChannel` implementation
- [x] Add [`Eth-Consensus-Block-Value`](https://github.com/ethereum/beacon-APIs/pull/358)
- [x] Consider `SKIP_RANDAO_VERIFICATION_PARAMETER`
- [x] Add a `--Xuse-block-v3` flag
- [x] Add a block v2 fallback when block v3 return 404
- [x] #7851

mehdi-aouadi:
## PR Description
Add `ProduceBlock` API `/eth/v3/validator/blocks/{slot}`
[Link to the spec](https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/block.v3.yaml)

- Add new `/eth/v3/validator/blocks/{slot}` endpoint
- Deprecate `GetNewBlock` `/eth/v2/validator/blocks/{slot}`
- Deprecate `GetNewBlindedBlock` `/eth/v1/validator/blinded_blocks/{slot}`

## Fixed Issue(s)

fixes #7390 
## Documentation

## Changelog


mehdi-aouadi:
Missing parts to implement in a separate PR:
- Implement `ProduceBlockRequest` (Like the current `CreateBlockRequest`) and add the new endpoint response deserialisation
- Make the `BlockProductionDuty` use the new `createUnsignedBlock(UInt64 slot, BLSSignature randaoReveal, Optional<Bytes32> graffiti)` (without passing the blind parameter)

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockFactoryPhase0.java
```diff
@@ -66,6 +72,34 @@ public SafeFuture<BlockContainer> createUnsignedBlock(
...
+    final Bytes32 parentRoot = spec.getBlockRootAtSlot(blockSlotState, slotBeforeBlock);
+
+    return spec.createNewUnsignedBlock(
+            newSlot,
+            spec.getBeaconProposerIndex(blockSlotState, newSlot),
+            blockSlotState,
+            parentRoot,
+            operationSelector.createSelector(
+                parentRoot, blockSlotState, randaoReveal, optionalGraffiti),
+            produceBlindedBlocks)
```
tbenr: AS per discussion, I think we should avoid having this flag attached to the existing validator config.
I can see it hardcoded to `true` or eventually driven by another beacon-node-specific `--X` parameter

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
```diff
@@ -0,0 +1,151 @@
...
+                + "The beacon node must return an unblinded block if it obtains the execution payload from its\n"
+                + "paired execution node. It must only return a blinded block if it obtains the execution payload\n"
+                + "header from an MEV relay.\n"
+                + "Metadata in the response indicates the type of block produced, and the supported types of block\n"
+                + "will be added to as forks progress.")
+        .tags(TAG_VALIDATOR, TAG_VALIDATOR_REQUIRED)
+        .pathParam(SLOT_PARAMETER.withDescription(SLOT_PATH_DESCRIPTION))
+        .queryParamRequired(RANDAO_PARAMETER)
+        .queryParam(GRAFFITI_PARAMETER)
+        .queryParam(SKIP_RANDAO_VERIFICATION_PARAMETER)
```
tbenr, line 94: `SKIP_RANDAO_VERIFICATION_PARAMETER` is also present in V2 but apparently we never supported it.
Indeed we haven't implemented the logic, and in fact you don't read the param in `handleRequest`.
But we could leave it as a separate PR.

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImplTest.java
```diff
@@ -842,7 +843,8 @@ private void verifyFallbackToLocalEL(
                             executionPayloadContext,
                             Optional.empty(),
                             Optional.empty(),
-                            Optional.of(SafeFuture.completedFuture(headerWithFallbackData))))))
+                            Optional.of(SafeFuture.completedFuture(headerWithFallbackData)),
+                            Optional.empty()))))
```
tbenr, line 958: I think here we need to handle the case of payload being produced via blinded flow.

So I think we need to expose the winning value in the `HeaderWithFallbackData` created in `ExecutionBuilderModule::builderGetHeader`.

I'd rename the field to `executionPayloadValueFuture` instead of `blockValue`

tbenr (commented):
some other comments added

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockFactoryPhase0.java
mehdi-aouadi: Done, removed the validator config option

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImplTest.java
mehdi-aouadi: Done and updated the tests accordingly

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
mehdi-aouadi: Yep, I'll do that in a separate PR

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+@TestSpecContext(allMilestones = true)
+public class GetNewBlockV3IntegrationTest extends AbstractDataBackedRestAPIIntegrationTest {
+
+  private DataStructureUtil dataStructureUtil;
+  private SpecMilestone specMilestone;
+  private final UInt256 executionPayloadValue = UInt256.valueOf(12345);
+
+  @BeforeEach
+  void setup(TestSpecInvocationContextProvider.SpecContext specContext) {
```
zilm13: final, please

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
```diff
@@ -0,0 +1,176 @@
...
+        .response(
+            SC_OK, "Request successful", getResponseType(schemaDefinitionCache), sszResponseType())
+        .build();
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
```
zilm13: final please

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
```diff
@@ -0,0 +1,176 @@
...
+                              blockContainerAndMetaData
+                                  .getExecutionPayloadValue()
+                                  .toDecimalString());
+                          return AsyncApiResponse.respondOk(blockContainerAndMetaData);
+                        })
+                    .orElseThrow(ChainDataUnavailableException::new)));
+  }
+
+  private static SerializableTypeDefinition<BlockContainerAndMetaData> getResponseType(
+      SchemaDefinitionCache schemaDefinitionCache) {
```
zilm13: final please

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
```diff
@@ -0,0 +1,176 @@
...
+
+    final Map<
+            Pair<BiPredicate<BlockContainer, SpecMilestone>, SpecMilestone>,
+            Function<SchemaDefinitions, SszSchema<? extends BlockContainer>>>
+        schemaGetters = new LinkedHashMap<>();
+
+    schemaGetters.put(
```
zilm13: It's very hard to read. We probably need to refactor it to some kind of builder. As an easy fix, I'd suggest adding a comment here something like // (test predicate, earliest milestone) -> schema getter.

Better fast approach could be to add record to Milestone utils (you could add create method also if you like it): 
```java
  public record ConditionalSchemaGetter<T> (BiPredicate<T, SpecMilestone> testPredicate, SpecMilestone earliestSupportedMilestone, Function<SchemaDefinitions, SszSchema<? extends T>> schemaGetter) {}
```
refactor a bit usage in util method 
and then populate it here:
```java
    final List<MilestoneDependentTypesUtil.ConditionalSchemaGetter<BlockContainer>> schemaGetters =
        new ArrayList<>();

    schemaGetters.add(
        new MilestoneDependentTypesUtil.ConditionalSchemaGetter<>(
            (blockContainer, milestone) ->
                schemaDefinitionCache.milestoneAtSlot(blockContainer.getSlot()).equals(milestone)
                    && !blockContainer.isBlinded(),
            SpecMilestone.PHASE0,
            SchemaDefinitions::getBlockContainerSchema));
```

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -763,10 +766,14 @@ protected void initSubnetSubscriber() {
     eventChannels.subscribe(SlotEventsChannel.class, stableSubnetSubscriber);
   }
 
+  public void initExecutionLayerBlockProductionManager() {
+    LOG.debug("BeaconChainController.initExecutionLayerBlockProductionManager()");
+    executionLayerBlockProductionManager =
```
zilm13: `this.` please

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/BlockProposalUtil.java
```diff
@@ -110,6 +110,16 @@ public SafeFuture<BeaconBlockAndState> createNewUnsignedBlock(
...
   }
 
+  public SafeFuture<BeaconBlockAndState> createNewUnsignedBlock(
+      final UInt64 newSlot,
+      final int proposerIndex,
+      final BeaconState blockSlotState,
+      final Bytes32 parentBlockSigningRoot,
+      final Consumer<BeaconBlockBodyBuilder> bodyBuilder) {
+    return createNewUnsignedBlock(
+        newSlot, proposerIndex, blockSlotState, parentBlockSigningRoot, bodyBuilder, true);
```
zilm13, line 120: Could you please explain this, I don't understand why it is always blinded?
Maybe add a comment?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/metadata/BlockContainerAndMetaData.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("data", getData())
+        .add("milestone", getMilestone())
+        .add("executionOptimistic", isExecutionOptimistic())
+        .add("canonical", isCanonical())
+        .add("executionPayloadValue", executionPayloadValue.toDecimalString())
```
zilm13: missed `finalized` field

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadResult.java
```diff
@@ -25,15 +26,19 @@ public class ExecutionPayloadResult {
   private final Optional<SafeFuture<Optional<BlobsBundle>>> blobsBundleFuture;
   private final Optional<SafeFuture<HeaderWithFallbackData>> headerWithFallbackDataFuture;
 
+  private final Optional<SafeFuture<UInt256>> executionPayloadValueFuture;
```
zilm13, line 28: worth updating equals, hashCode and toString too

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
```diff
@@ -132,10 +142,17 @@ private ExecutionPayloadResult builderGetHeader(
     final SafeFuture<HeaderWithFallbackData> headerWithFallbackDataFuture =
         executionLayerChannel.builderGetHeader(executionPayloadContext, state);
 
+    final SafeFuture<GetPayloadResponse> getPayloadResponseFuture =
```
zilm13: I'm not sure it's the right thing calling it here, we do it somewhere inside `builderGetHeader` already, right @StefanBratanov ?
You should probably pass it inside HeaderWithFallbackData from there

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -100,6 +108,46 @@
...
+        executionLayerBlockProductionManager
+            .getCachedPayloadResult(slot)
+            .orElseThrow(
+                () -> new IllegalStateException("ExecutionPayloadResult is not available"));
+    return payloadResult
+        .getExecutionPayloadValueFuture()
+        .orElseThrow(() -> new IllegalStateException("Execution Payload Value is not available"));
+  }
+
+  private void checkBlockProducingParameters(UInt64 slot, BLSSignature randao) {
```
zilm13: finals plesae

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -100,6 +108,46 @@
...
+              maybeBlockContainer.get(),
+              spec.atSlot(maybeBlockContainer.get().getSlot()).getMilestone(),
+              false,
+              false,
+              false,
+              executionPayloadValue));
+    }
+  }
+
+  private SafeFuture<UInt256> retrieveExecutionPayloadValue(UInt64 slot) {
```
zilm13: final please

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -100,6 +108,46 @@
...
+
+  public SafeFuture<Optional<BlockContainerAndMetaData>> produceBlock(
+      final UInt64 slot, final BLSSignature randao, final Optional<Bytes32> graffiti) {
+    checkBlockProducingParameters(slot, randao);
+    return validatorApiChannel
+        .createUnsignedBlock(slot, randao, graffiti)
+        .thenCombine(retrieveExecutionPayloadValue(slot), this::lookUpData);
+  }
+
+  private Optional<BlockContainerAndMetaData> lookUpData(
```
zilm13: finals please

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -100,6 +108,46 @@
...
+      Optional<BlockContainer> maybeBlockContainer, UInt256 executionPayloadValue) {
+    if (maybeBlockContainer.isEmpty()) {
+      return Optional.empty();
+    } else {
+      return Optional.of(
+          new BlockContainerAndMetaData(
+              maybeBlockContainer.get(),
+              spec.atSlot(maybeBlockContainer.get().getSlot()).getMilestone(),
+              false,
+              false,
```
zilm13: I'm not sure it's not canonical. It looks like a bad sign when reusing ObjectAndMetadata that we actually need only 2 of its 5 fields. Maybe it worth setting up a separate class for this (record?)

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImplTest.java
```diff
@@ -111,6 +114,9 @@ public void preDeneb_builderOffline() throws Exception {
         .isEqualTo(executionPayloadContext);
     assertThat(executionPayloadResult.getExecutionPayloadFuture()).isEmpty();
     assertThat(executionPayloadResult.getBlobsBundleFuture()).isEmpty();
+    assertThat(executionPayloadResult.getExecutionPayloadValueFuture()).isPresent();
```
zilm13, line 116: I'd add `verify(executionClientHandler).engineGetPayload(any(), any())` here (which is good idea to clarify when it's fixed), and I'm getting error, because it was 2 times. The reason is we are calling it 2 times (see last comment to `ExecutionLayerBlockProductionManagerImpl`) which is not correct I think.

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImplTest.java
```diff
@@ -163,10 +169,13 @@ public void preDeneb_builderOnline() throws Exception {
...
         executionPayloadResult.getHeaderWithFallbackDataFuture().orElseThrow();
     assertThat(headerWithFallbackDataFuture.get()).isEqualTo(expectedResult);
+    final SafeFuture<UInt256> executionPayloadValueFuture =
+        executionPayloadResult.getExecutionPayloadValueFuture().orElseThrow();
+    assertThat(executionPayloadValueFuture.get()).isEqualTo(executionPayloadValue);
 
     // we expect both builder and local engine have been called
     verifyBuilderCalled(slot, executionPayloadContext);
-    verifyEngineCalled(executionPayloadContext, slot);
+    verifyEngineCalled(executionPayloadContext, slot, 2);
```
zilm13: it shouldn't be called for 2 times here and in the other cases, or may be I got it wrong. Let's clarify with @StefanBratanov

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
```diff
@@ -0,0 +1,328 @@
...
+  private String getExpectedBlockAsJson(
+      final SpecMilestone specMilestone, final boolean blinded, final boolean blockContents)
+      throws IOException {
+    final String fileName =
+        String.format(
+            "new%s%s%s.json",
+            blinded ? "Blinded" : "",
+            blockContents ? "BlockContents" : "Block",
+            specMilestone.name());
+    return Resources.toString(
```
zilm13, line 378: I'd better parse json. Every time we change something in dataStructureUtil order could be changed and all tests with hardcoded random expectation starts to failing. Block structures are big and this means more chances to affect it. Plus we have the actual json data by the hand. I'd propose to refactor this to json parsing. @tbenr what's your feeling?

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
```diff
@@ -0,0 +1,328 @@
...
+    assertThat(response.header(HEADER_EXECUTION_PAYLOAD_BLINDED)).isEqualTo(Boolean.toString(true));
+    assertThat(response.header(HEADER_EXECUTION_PAYLOAD_VALUE))
+        .isEqualTo(executionPayloadValue.toDecimalString());
+    final BlindedBlockContents result =
+        (BlindedBlockContents)
+            spec.getGenesisSchemaDefinitions()
+                .getBlindedBlockContainerSchema()
+                .sszDeserialize(Bytes.of(response.body().bytes()));
+    assertThat(result).isEqualTo(blindedBlockContents);
+  }
```
zilm13, line 308: We may need to add error cases test to this, but I'd separate it

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3Test.java
```diff
@@ -0,0 +1,193 @@
...
+    request.setQueryParameter(RANDAO_REVEAL, signature.toBytesCompressed().toHexString());
+    when(validatorDataProvider.getMilestoneAtSlot(UInt64.ONE)).thenReturn(SpecMilestone.ALTAIR);
+  }
+
+  @TestTemplate
+  void shouldHandleBlindedBeaconBlocks() throws Exception {
+    assumeThat(specMilestone).isGreaterThanOrEqualTo(BELLATRIX).isLessThanOrEqualTo(CAPELLA);
+    dataStructureUtil = new DataStructureUtil(spec);
+    final BeaconBlock blindedBeaconBlock = dataStructureUtil.randomBlindedBeaconBlock(ONE);
+    BlockContainerAndMetaData blockContainerAndMetaData =
```
zilm13: finals for `blockContainerAndMetaData` here and in next cases please

zilm13 (commented):
The main question for me is am I getting it right that we do extra call with this, which should not be done. If I get it correctly, we need to fix this. Other comments are minor like boring finals plus one refactoring suggestion in handler which could improve readability to my mind

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImplTest.java
StefanBratanov: Hmm yeah it should only be once I think. What is the reason of this change @mehdi-aouadi

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3Test.java
mehdi-aouadi: done

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
mehdi-aouadi: done

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
mehdi-aouadi: done

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
mehdi-aouadi: done

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadResult.java
mehdi-aouadi: done

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/metadata/BlockContainerAndMetaData.java
mehdi-aouadi: added it

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
mehdi-aouadi: done

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
mehdi-aouadi: done

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
mehdi-aouadi: done

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
mehdi-aouadi: done

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v3/validator/GetNewBlockV3.java
mehdi-aouadi: Done

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/BlockProposalUtil.java
mehdi-aouadi: This has been discussed [here](https://github.com/Consensys/teku/pull/7529#discussion_r1335928037)
We might add a -X option to switch to unblinded block production but that could be done in a separate if we agree to do so...

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
mehdi-aouadi: This was suggested in [this previous discussion](https://github.com/Consensys/teku/pull/7529#discussion_r1336013723) and it makes sense to me

data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
mehdi-aouadi: Done. I refactored the `BlockContainerAndMetadata`, created a record and added the ssz serialisation

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/BlockProposalUtil.java
tbenr: @zilm13
The point is that our current flow requires to know "in advance" which  type (blinded, or full) of object we are about to produce. The decision is made by the VC. This new API enables BN to decide the type based on the flow, and VC will be passive and should handle both types dynamically.

The problem is that, to allow us to change type dynamically on BN side it will require to move the type selection deep into the flow (where we know if we did a builder flow, or a fallback to local) and requires some changes. So, for now, the only way to be initially compatible with this API is to always return blinded objects (which is the only option we have to remain compatible with all flows) and later on provide a dynamic version that could be under a feature flag (like `blockCreationRespondWithFullContentWhenPossible`)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/metadata/BlockContainerAndMetaData.java
```diff
@@ -0,0 +1,56 @@
...
+
+package tech.pegasys.teku.spec.datastructures.metadata;
+
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
+import org.apache.tuweni.units.bigints.UInt256;
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.datastructures.blocks.BlockContainer;
+
+public class BlockContainerAndMetaData extends ObjectAndMetaData<BlockContainer> {
```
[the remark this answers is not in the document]
mehdi-aouadi: Fixed. Replaced `class BlockContainerAndMetaData` by a `record BlockContainerAndMetaData`

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImplTest.java
mehdi-aouadi: Done, added a callback when calling `builderGetHeader` in order to get `executionPayloadValue` at the same time and avoid calling the EL a second time

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
mehdi-aouadi: Done. As discussed, added a fallback to get the `executionPayloadValue`

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImplTest.java
StefanBratanov: Nice

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImplTest.java
mehdi-aouadi: Done

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
mehdi-aouadi: Added some integration test for empty block production result, missing execution payload result in cache and missing execution payload value

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
mehdi-aouadi: As discussed, we didn't add the annotated classes for the json parser and hence can't parse the generated block objects. We could refactor the tests in a separate PR if we agree to do so

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/BlockProposalUtil.java
zilm13: Could we add some comment to it, it would be valuable

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
```diff
@@ -0,0 +1,382 @@
...
+        .thenReturn(
+            Optional.of(
+                new ExecutionPayloadResult(
+                    mock(ExecutionPayloadContext.class),
+                    Optional.empty(),
+                    Optional.empty(),
+                    Optional.empty(),
+                    Optional.of(SafeFuture.completedFuture(executionPayloadValue)))));
+    Response response = get(signature, ContentTypes.JSON);
+    assertThat(response.code()).isEqualTo(SC_NO_CONTENT);
```
zilm13: I'm not sure on this. API spec has only 200, 400, 500, 503 response codes. We probably should go with the same 500 here

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannel.java
```diff
@@ -126,5 +129,7 @@ SafeFuture<BuilderPayload> builderGetPayload(
    * BeaconState, boolean)} instead
    */
   SafeFuture<HeaderWithFallbackData> builderGetHeader(
-      ExecutionPayloadContext executionPayloadContext, BeaconState state);
+      ExecutionPayloadContext executionPayloadContext,
+      BeaconState state,
+      SafeFuture<UInt256> localExecutionPayloadValue);
```
zilm13: Let's call it `localPayloadValueResult` here and in all childs to highlight that it's callback.  Also I'd add Javadoc with param description that it's callback

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImplTest.java
```diff
@@ -206,7 +208,11 @@ public void builderGetHeaderGetPayload_shouldReturnHeaderAndPayloadViaBuilder()
     prepareEngineGetPayloadResponse(executionPayloadContext, UInt256.ZERO, slot);
 
     // we expect result from the builder
-    assertThat(executionLayerManager.builderGetHeader(executionPayloadContext, state))
+    assertThat(
+            executionLayerManager.builderGetHeader(
+                executionPayloadContext,
+                state,
+                SafeFuture.completedFuture(localExecutionPayloadValue)))
```
zilm13: I think it's an inappropriate way to use callback in test. It's better to change it here and in next cases to something like:
```java
    prepareEngineGetPayloadResponse(executionPayloadContext, UInt256.ZERO, slot);
    final SafeFuture<UInt256> blockValueResult = new SafeFuture<>();
    // we expect result from the builder
    assertThat(
            executionLayerManager.builderGetHeader(
                executionPayloadContext,
                state,
                blockValueResult))
        .isCompletedWithValue(HeaderWithFallbackData.create(header));
    assertThat(blockValueResult).isCompletedWithValue(UInt256.ZERO);
```

zilm13 (commented):
Few comments plus this one please https://github.com/Consensys/teku/pull/7529#discussion_r1367728063

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v3/GetNewBlockV3IntegrationTest.java
mehdi-aouadi: Done, updated the API interface to return 500 when no block is produced and updated the tests accordingly

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannel.java
mehdi-aouadi: done, renamed it and added a java doc

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImplTest.java
mehdi-aouadi: Done

zilm13 (approved):
LGTM

PR #7531, opened 2023-09-19
Store only most recent blocks in cache

zilm13:
## Fixed Issue(s)

Not sure in this but let's try.
So by my measurement on mainnet node it was something like this:
2200 hits in hour
- 1500 available in current master
- 700 misses

All requests in interval [current-94, current], so last 3 epochs
After changes I get something like
- 1650 available
- 550 misses

Just a 10% improvement. Also looking at misses, we will spend 64 more (96 total) `signedBlock` memory slots for 1/4 of hits, while currently using 32 for 3/4 of hits. Doesn't worth increasing the size of the cache.

## Documentation

## Changelog


tbenr:
Yes, I agree seem like with 32 we cover a decent amount and, from the numbers you posted, increasing it would just be an attempt to cover "the long tail".

rolfyone:
i'd be happy with just going straight to storage (not caching) anything finalized, and non finalized can go through the cache, which probably solves the same thing.... i guess basically hot blocks column has a cache and finalized blocks column arguably doesnt need one...

thats a sweeping statement, sorry... but yeh i think given the numbers thats probably the avenue i'd be investigating...

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransactionUpdates.java
```diff
@@ -109,7 +110,11 @@ public void applyToStore(final Store store, final UpdateResult updateResult) {
     tx.genesisTime.ifPresent(value -> store.genesisTime = value);
     tx.justifiedCheckpoint.ifPresent(store::updateJustifiedCheckpoint);
     tx.bestJustifiedCheckpoint.ifPresent(store::updateBestJustifiedCheckpoint);
-    hotBlocks.forEach((root, value) -> store.blocks.put(root, value.getBlock()));
+    hotBlocks.values().stream()
+        .sorted(Comparator.comparing(BlockAndCheckpoints::getSlot))
+        .map(BlockAndCheckpoints::getBlock)
+        .forEach(block -> store.blocks.put(block.getRoot(), block));
+    store.blockCountGauge.ifPresent(gauge -> gauge.set(store.blocks.size()));
```
rolfyone: this is weird, we've made the gauge non private just to access here and pass it data from the store anyway...

we'd be better off having a function on the store to set that gauge if its present, this is a weird flow.

storage/src/test/java/tech/pegasys/teku/storage/store/StoreTest.java
```diff
@@ -298,6 +298,40 @@ public void retrieveFinalizedCheckpointAndState() {
...
+
+    // Add blocks
+    final StoreTransaction tx = store.startTransaction(new StubStorageUpdateChannel());
+    chainBuilder
+        .streamBlocksAndStates()
+        .forEach(
+            blockAndState ->
+                tx.putBlockAndState(
+                    blockAndState, spec.calculateBlockCheckpoints(blockAndState.getState())));
+    tx.commit().join();
```
rolfyone: iirc we needed to do safeJoin in tests... @tbenr was looking at something around joins recently...

tbenr: Yes, we should always use `safeJoin` 'cose in some cases we may not catch issues (may not apply in this specific case but we should have the good habit)

zilm13: Got it, fixed

storage/src/main/java/tech/pegasys/teku/storage/store/StoreTransactionUpdates.java
zilm13: I'm a bit maximalist there, how about this? https://github.com/Consensys/teku/pull/7531/commits/47c516afba98e30534ebe980fa601027397cd88a

rolfyone: yep i think that's cleaner.

rolfyone (approved):
LGTM

PR #7540, opened 2023-09-22
Test gzip compression in rest API

### LINKED ISSUE #7459: add gzip compression on rest api into a test
https://github.com/Consensys/teku/pull/7457

could just be an acceptance or integration test, just so that we can verify it continues to work

courtneyeh:
fixes #7459 

## Documentation

## Changelog


data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/AbstractDataBackedRestAPIIntegrationTest.java
```diff
@@ -340,6 +340,17 @@ protected void assertBodyEquals(final Response response, final String body) thro
     assertThat(response.body().string()).isEqualTo(body);
   }
 
+  protected Response getGzipResponse(final String path, final String contentType)
```
lucassaldanha: I think we could stick to the DSL convention and have an override of the `getResponse` method that accepts the value for the Accept-Encoding header.

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/beacon/GetBlockV2IntegrationTest.java
```diff
@@ -82,10 +83,25 @@ public void shouldGetAltairBlockAsSsz() throws IOException {
     assertThat(response.header(HEADER_CONSENSUS_VERSION)).isEqualTo(Version.altair.name());
   }
 
+  @Test
+  public void shouldGetAltairBlockAsGzip() throws IOException {
+    startRestAPIAtGenesis(SpecMilestone.ALTAIR);
+    createBlocksAtSlots(10);
+    final Response response = getGzip("head", OCTET_STREAM);
```
lucassaldanha: My suggestion is to get the response as JSON just to make deserializing it into an object a bit easier. This way all we need to do is decompress the Gzip into a string and parse it into a `GetBlockResponseV2` object.

Not that it can't be done with SSZ. So it is up to you :)

lucassaldanha (approved):

Issue #7554, opened 2023-09-27
Add `inactivity` and `inclusion_delay` to `ideal_rewards` in Attestation Rewards API response


courtneyeh:
### Description
Add the `inactivity` and `inclusion_delay` field to `ideal_rewards` in the `AttestationsRewards` response.

Related to [ethereum/beacon-APIs#340](https://github.com/ethereum/beacon-APIs/pull/340)

Broken down from #7414

PR #7564, opened 2023-10-02
Adding inclusion_delay and inactivity to ideal_rewards object

lucassaldanha:
## PR Description

Added two new fields to the `ideal_rewards` object accordingly to https://github.com/ethereum/beacon-APIs/pull/340.

## Fixed Issue(s)
partially addresses #7554 

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/migrated/TotalAttestationReward.java
```diff
@@ -69,7 +69,8 @@ public TotalAttestationReward(long validatorIndex, final RewardAndPenalty reward
     this.inactivity =
         detailedRewardAndPenalty
             .getReward(RewardComponent.INACTIVITY)
-            .minus(detailedRewardAndPenalty.getPenalty(RewardComponent.INACTIVITY));
+            .minus(detailedRewardAndPenalty.getPenalty(RewardComponent.INACTIVITY))
+            .longValue();
```
courtneyeh: Now that we are using a `long` value, we should match the previous calculation's formats above.
```
this.inactivity =
        detailedRewardAndPenalty.getReward(RewardComponent.INACTIVITY).longValue()
            - detailedRewardAndPenalty.getPenalty(RewardComponent.INACTIVITY).longValue();
```

data/beaconrestapi/src/test/resources/tech/pegasys/teku/beaconrestapi/handlers/v1/rewards/getAttestationRewards.json
```diff
@@ -1 +1 @@
-{"execution_optimistic":false,"finalized":false,"data":{"ideal_rewards":[{"effective_balance":"1000000000","head":"2500","target":"5000","source":"5000"}],"total_rewards":[{"validator_index":"0","head":"2000","target":"2000","source":"4000","inclusion_delay":"2000","inactivity":"300"}]}}
\ No newline at end of file
+{"execution_optimistic":false,"finalized":false,"data":{"ideal_rewards":[{"effective_balance":"1000000000","head":"2500","target":"5000","source":"5000","inactivity":"300"}],"total_rewards":[{"validator_index":"0","head":"2000","target":"2000","source":"4000","inclusion_delay":"2000","inactivity":"300"}]}}
```
courtneyeh, line 1: Did we want to test `inclusion_delay` is correct in the handle `200` result test?

data/serializer/src/main/java/tech/pegasys/teku/api/migrated/TotalAttestationReward.java
lucassaldanha: Actually, we should probably update the other calculations to have a single `.longValue()` so we only do the unboxing once. I'll update the PR.

data/beaconrestapi/src/test/resources/tech/pegasys/teku/beaconrestapi/handlers/v1/rewards/getAttestationRewards.json
lucassaldanha: This only applies to phase0. Given we don't support phase0 at the moment I think it is okay to leave it out.

data/serializer/src/main/java/tech/pegasys/teku/api/migrated/TotalAttestationReward.java
lucassaldanha: Actually, I introduced a bug doing it that way because we would have issues when the penalties for a component is greater than the rewards and we end up with a negative value causing a UInt underflow.

courtneyeh (approved):
LGTM

PR #7561, opened 2023-09-29
Implement `blob_sidecar` event

tbenr:
Implements `blob_sidecar` SSE event.

Note that this diverges from block implementation: to capture locally produced blobs as well as the one arriving via gossip, I intercept blobs at `BlobSidecarPool` instead of `BlobSidecarManager`

Partially implements #7513

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #7569, opened 2023-10-03
Fix example for KZG_COMMITMENT_TYPE

jtraglia:
## PR Description

Noticed that the example for `KZG_COMMITMENT_TYPE`, which was added yesterday (#7561), was wrong. A (compressed) KZG commitment is 48 bytes. The example was 96 bytes. I replaced the example with a value from the spec tests:

* https://github.com/ethereum/consensus-spec-tests/blob/master/tests/general/deneb/kzg/blob_to_kzg_commitment/kzg-mainnet/blob_to_kzg_commitment_case_valid_blob_19b3f3f8c98ea31e/data.yaml

I also changed the description from plural to singular. I think this was the intention.

## Documentation

## Changelog


tbenr (approved):
LGTM good catch

PR #7575, opened 2023-10-04
Improve `isDataAvailable` checks

tbenr:
~~This PR is mainly focused on making sure that commitments coming from block match the commitments contained in the `BlobSidecar`s~~

Updated strategy by moving away extra checks from `isDataAvailable` method to a dedicated method in miscHelper.

Since we check that blobsSidecar content are consistent with the block before calling `isDataAvailable`, we can now pass them only to perform data availability.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -65,26 +67,43 @@ private static KZG initKZG(final SpecConfigDeneb config) {
...
+        .forEach(
+            index -> {
+              final BlobSidecar blobSidecar = blobSidecars.get(index);
+
+              checkArgument(
+                  slot.equals(blobSidecar.getSlot()),
+                  "Blob sidecar slot %s does not match block slot %s",
+                  blobSidecar.getSlot(),
+                  slot);
+              checkArgument(
```
zilm13: You could also add parent root check (but it will require changes in interface)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -65,26 +67,43 @@ private static KZG initKZG(final SpecConfigDeneb config) {
...
-              beaconBlockRoot);
-        });
-    final List<Bytes> blobs =
-        blobSidecars.stream().map(BlobSidecar::getBlob).map(Blob::getBytes).toList();
-    final List<KZGProof> proofs = blobSidecars.stream().map(BlobSidecar::getKZGProof).toList();
-
-    return kzg.verifyBlobKzgProofBatch(blobs, kzgCommitments, proofs);
+    checkArgument(
+        blobSidecars.size() == kzgCommitmentsFromBlock.size(),
+        "Blob sidecars count %s does not match KZG commitments count %s",
```
zilm13: maybe `match block KZG commitments` to be more clear

zilm13 (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
rolfyone: would this be more helpful to report if we had the actual block slot or something associating it to the block in the message?
It'll be very hard for anyone to understand what this relates to because there's nothing tying it into data we can understand, so from the perspective of a node op, this will be a 'useless message' as it stands.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -65,26 +67,43 @@ private static KZG initKZG(final SpecConfigDeneb config) {
...
+              checkArgument(
+                  beaconBlockRoot.equals(blobSidecar.getBlockRoot()),
+                  "Blob sidecar block root %s does not match block root %s",
+                  blobSidecar.getBlockRoot(),
+                  beaconBlockRoot);
+              checkArgument(
+                  kzgCommitmentsFromBlock.get(index).equals(blobSidecar.getKZGCommitment()),
+                  "Blob sidecar KZG commitment %s does not match block KZG commitment %s",
+                  blobSidecar.getKZGCommitment(),
+                  kzgCommitmentsFromBlock.get(index));
```
rolfyone: similar to above, adding slot at a minimum would be helpful

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
tbenr: right!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
tbenr: I started thinking that these checks could leave in the `BlockBlobSidecarsTracker`. If the tracker guarantees that block and sidecars field are consistent, we can then have this method only dedicated to KZG check, which is the original intention. The good thing on discarding inconsistent blobSidecars at `BlockBlobSidecarsTracker` level is that we could try to refetch them early.

tbenr: thinking more about it, an early refetch might be undesirable here. If the proposer is doing something strange, is better to do nothing, skip the block import, vote on previous head, and then lookup only if next block builds on top of this (which might be not the case most times, unless next proposer is under control of the same malicious\buggy entity)

tbenr: so `ForkChoiceBlobSidecarsAvailabilityChecker::validateBatch` might be a better place to have this checks

tbenr: I decided to move the validation back to miscHelper but on a dedicated method. This simplified the tests in `ForkChoiceBlobSidecarsAvailabilityCheckerTest`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -60,29 +61,22 @@ private static KZG initKZG(final SpecConfigDeneb config) {
...
+   *
+   * <p>In Deneb we don't need to retrieve data, everything is already available via blob sidecars.
    */
   @Override
-  public boolean isDataAvailable(
-      final UInt64 slot,
-      final Bytes32 beaconBlockRoot,
-      final List<KZGCommitment> kzgCommitments,
-      final List<BlobSidecar> blobSidecars) {
+  public boolean isDataAvailable(final List<BlobSidecar> blobSidecars) {
```
zilm13: What I don't like here, both methods, this and `validateBlobSidecarsAgainstBlock` could be called now with part of blobs, which is not clear. Next,  `isDataAvailable` is a spec thing, so I expect it will verify it completely. Plus the last piece of puzzle, that we call this method with blobSidecars with indexes matching all indexes of `kzgCommitments` from block is somewhere else in the 3rd place.    

I'm not sure what's the best solution to make this better. My suggestion (not the best one, need to think more on this) is:
1. rename this method, so it's actually 2nd round of verification
2. add comments to both methods that they could be used for partial verification
3. add actual isDataAvailable(kzgCommitmentsFromBlock, verifiedBlobKzgCommitments) here, which will finalize the action by just equal matching 2 lists

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -2123,24 +2124,38 @@ public Bytes randomBlobBytes() {
   }
 
   public List<BlobSidecar> randomBlobSidecarsForBlock(final SignedBeaconBlock block) {
-    return randomSignedBlobSidecarsForBlock(block).stream()
+    return randomBlobSidecarsForBlock(block, (__, builder) -> builder);
+  }
+
+  public List<BlobSidecar> randomBlobSidecarsForBlock(
+      final SignedBeaconBlock block,
+      BiFunction<Integer, RandomBlobSidecarBuilder, RandomBlobSidecarBuilder> modifier) {
```
zilm13: boring final

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -2123,24 +2124,38 @@ public Bytes randomBlobBytes() {
...
+    return randomSignedBlobSidecarsForBlock(block, modifier).stream()
         .map(SignedBlobSidecar::getBlobSidecar)
         .collect(toList());
   }
 
-  public List<SignedBlobSidecar> randomSignedBlobSidecarsForBlock(final SignedBeaconBlock block) {
-    final int numberOfKzgCommitments =
+  public List<SignedBlobSidecar> randomSignedBlobSidecarsForBlock(
+      final SignedBeaconBlock block,
+      BiFunction<Integer, RandomBlobSidecarBuilder, RandomBlobSidecarBuilder> modifier) {
```
zilm13: another one

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
zilm13: or better isDataAvailable(block, verifiedBlobIdentifiers)
return IntRange.of(0, block.getKzgCommitments.size()).map(index -> verifiedBlobIdentifiers.contains(new BlobIdentifier(block.getRoot, index))).filter(false).findFirst().isEmpty()

tbenr: I agree that it isn't clear that those methods can be called with partial blobs. I'm thinking if it is better to go back to single method.

Consider that `ForkChoiceBlobSidecarsAvailabilityChecker::computeFinalValidationResult` is the method performing the checks on the final set.

zilm13: Yeah, yeah, I understand it's happened there but we'd better have it here it and by my understanding it should be called `isDataAvailable` (for block, all data)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -318,21 +307,34 @@ private BlobSidecarsAndValidationResult computeFinalValidationResult(
...
+  }
+
+  /**
+   * make sure that final blob sidecar list is complete
+   *
+   * @param additionalBlobSidecarsAndValidationResult
+   * @return
+   */
+  private BlobSidecarsAndValidationResult performCompletenessValidation(
+      final BlobSidecarsAndValidationResult additionalBlobSidecarsAndValidationResult) {
```
zilm13: it's probably not additional, all instead, especially if it could pass completeness check

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -318,21 +307,34 @@ private BlobSidecarsAndValidationResult computeFinalValidationResult(
...
+      return additionalBlobSidecarsAndValidationResult;
+    }
+
+    try {
+      spec.atSlot(blockBlobSidecarsTracker.getSlotAndBlockRoot().getSlot())
+          .miscHelpers()
+          .verifyBlobSidecarCompleteness(
+              additionalBlobSidecarsAndValidationResult.getBlobSidecars(),
+              kzgCommitmentsFromBlockSupplier.get());
+    } catch (final IllegalArgumentException ex) {
```
zilm13, line 342: It's not clear by `verifyBlobSidecarCompleteness` signature that we should catch `IllegalArgumentException` from it

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -106,34 +110,34 @@ public SafeFuture<BlobSidecarsAndValidationResult> getAvailabilityCheckResult()
...
+      return validateBatch(blobSidecars);
     }
 
     if (isBlockOutsideDataAvailabilityWindow()) {
       return BlobSidecarsAndValidationResult.NOT_REQUIRED;
     }
 
-    if (kzgCommitments.isEmpty()) {
+    if (kzgCommitmentsFromBlock.isEmpty()) {
       return BlobSidecarsAndValidationResult.validResult(List.of());
```
zilm13, line 124: We are not doing completeness check in this case, or I missed it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -106,34 +110,34 @@ public SafeFuture<BlobSidecarsAndValidationResult> getAvailabilityCheckResult()
   @Override
   public BlobSidecarsAndValidationResult validateImmediately(final List<BlobSidecar> blobSidecars) {
 
-    final List<KZGCommitment> kzgCommitments = kzgCommitmentsSupplier.get();
+    final List<KZGCommitment> kzgCommitmentsFromBlock = kzgCommitmentsFromBlockSupplier.get();
 
     if (!blobSidecars.isEmpty()) {
-      return validateBatch(blobSidecars, kzgCommitments);
+      return validateBatch(blobSidecars);
```
zilm13: We are not doing completeness check in this case or I missed it.

zilm13 (commented):
Maybe worth adding a test especially on completeness?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
tbenr: yes, good catch, I wasn't clearly in a good shape!

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -106,34 +111,39 @@ public SafeFuture<BlobSidecarsAndValidationResult> getAvailabilityCheckResult()
...
   public BlobSidecarsAndValidationResult validateImmediately(final List<BlobSidecar> blobSidecars) {
 
-    final List<KZGCommitment> kzgCommitments = kzgCommitmentsSupplier.get();
+    final List<KZGCommitment> kzgCommitmentsFromBlock = kzgCommitmentsFromBlockSupplier.get();
 
     if (!blobSidecars.isEmpty()) {
-      return validateBatch(blobSidecars, kzgCommitments);
+      final BlobSidecarsAndValidationResult blobSidecarsAndValidationResult =
+          validateBatch(blobSidecars);
+      performCompletenessValidation(blobSidecarsAndValidationResult);
```
zilm13: You ignore the result of completeness check, there should be probably return on this line

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -318,22 +313,35 @@ private BlobSidecarsAndValidationResult computeFinalValidationResult(
...
+    }
+
+    try {
+      spec.atSlot(blockBlobSidecarsTracker.getSlotAndBlockRoot().getSlot())
+          .miscHelpers()
+          .verifyBlobSidecarCompleteness(
+              blobSidecarsAndValidationResult.getBlobSidecars(),
+              kzgCommitmentsFromBlockSupplier.get());
+    } catch (final IllegalArgumentException ex) {
+      checkArgument(
```
zilm13: I have some concerns on checkArgument vs checkState here, anyway we are already not super consistent with it in code. The more important is that we throw it all the way up, while we have  BlobSidecarsAndValidationResult.invalidResult(blobSidecars, throwable) which looks better in this case. Or we could at least add a comment that this could throw. Though I like .invalidResult more.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
tbenr: `performCompletenessValidation` is designed to throw only

zilm13: It has the line `return BlobSidecarsAndValidationResult.NOT_REQUIRED;`
Unless it's not `void` return I assume we care on return

tbenr: you're right.

zilm13 (approved):
LGTM

PR #7576, opened 2023-10-04
Add BlobSidecarsByRangeListenerWrapper with BlobSidecar p2p verification

### LINKED ISSUE #6727: [BlobSidecarsByRange] Verify sidecar before consuming next response chunk
[As per spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blobsidecarsbyrange-v1)

Before consuming the next response chunk, the response reader SHOULD verify the blob sidecar is well-formatted and correct w.r.t. the expected KZG commitments through validate_blobs.

zilm13:
## PR Description
So we will do verification actually 2 times and when receiving using `byRange` verification is run one by one without using opportunity of batch verification. It's a bit of extra resources consumed, but it's per spec.
Also I've added checks similar to those we do for blocks by range. Hope, I haven't made it too strict and it will not break our integration with other clients.

## Fixed Issue(s)

Fixes #6727 

## Documentation

## Changelog


tbenr:
Looks good with new latest name. To be fair, in the spec it is a `SHOULD`, not a `MUST`. With that said, I'd to have #7575 merged before this.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeListenerValidatingProxy.java
```diff
@@ -0,0 +1,141 @@
...
+    this.peer = peer;
+    this.blobSidecarResponseListener = blobSidecarResponseListener;
+    this.maxBlobsPerBlock = maxBlobsPerBlock;
+    this.miscHelpersDeneb = miscHelpersDeneb;
+    this.startSlot = startSlot;
+    this.endSlot = startSlot.plus(count);
+  }
+
+  @Override
+  public SafeFuture<?> onResponse(final BlobSidecar response) {
```
tbenr: param name `blobSidecar`?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeListenerValidatingProxy.java
```diff
@@ -0,0 +1,141 @@
...
+    implements RpcResponseListener<BlobSidecar> {
+
+  private final Peer peer;
+  private final RpcResponseListener<BlobSidecar> blobSidecarResponseListener;
+  private final UInt64 startSlot;
+  private final UInt64 endSlot;
+  private final Integer maxBlobsPerBlock;
+  private final MiscHelpersDeneb miscHelpersDeneb;
+
+  private Optional<BlobSidecarSummary> maybeLastBlobSidecarSummary = Optional.empty();
```
tbenr: I'd declare this `volatile` just in case the thread changes from one call to the other.

tbenr (approved):
LGTM. some nits

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -236,4 +248,15 @@ public int getBlobKzgCommitmentsCount(final SignedBeaconBlock signedBeaconBlock)
...
   }
+
+  public boolean verifyBlobSidecar(final BlobSidecar blobSidecar) {
+    try {
+      return kzg.verifyBlobKzgProofBatch(
+          List.of(blobSidecar.getBlob().getBytes()),
+          List.of(blobSidecar.getKZGCommitment()),
+          List.of(blobSidecar.getKZGProof()));
+    } catch (final KZGException ex) {
+      return false;
```
tbenr, line 279: Another thing is that I'm not a fan of swallowing the exception here without even a log line (debug?)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeListenerValidatingProxy.java
zilm13: nice point

PR #7591, opened 2023-10-13
Force installing libc deps on docker image

lucassaldanha:
## PR Description
- Forces installation of libc-bin and libc6

## Fixed Issue(s)
N/A

## Documentation

## Changelog


rolfyone (approved):
we should raise an issue so we can remember to remove this at some point...

PR #7593, opened 2023-10-13
Add blob sidecars cache in Store + usage

zilm13:
## PR Description
Some considerations when making this:
1. `Store` is tied with `RecentChainData` and it was actually incorrect when we were actually using it as historical storage for blobSidecars. Now it will return only cached blobSidecars which is not on par with blocks where we have hotBlocks, but at least we somehow separate recent and historical chain data.
2. To clarify presence of `blobSidecars` I was forced to return back to `Optional` wrap which we removed previously for simplicity, we'd better keep it to avoid unneeded db lookups.
3. I've used block size cache configuration, I think we could go without separate settings, at least in the beginning
4. I've added some usage of cache, where it could be useful, not other clear places were found for using this cache
5. `earliestBlobSidecarSlot` definitely needs caching too, will detach in separate PR, should be not small because pruning is on Database layer and we should handle this somehow
6. `BlobSidecarsProvider` was removed, it's unnecessary when we don't serve recentChainData as historical

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/MutableStore.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/MutableStore.java
@@ -13,7 +13,6 @@
 
 package tech.pegasys.teku.spec.datastructures.forkchoice;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import org.apache.tuweni.bytes.Bytes32;
@@ -43,7 +42,7 @@ public interface MutableStore extends ReadOnlyStore {
       SignedBeaconBlock block,
       BeaconState state,
       BlockCheckpoints checkpoints,
-      List<BlobSidecar> blobSidecars,
+      Optional<List<BlobSidecar>> blobSidecars,
       Optional<UInt64> earliestBlobSidecarSlot);
 
   default void putBlockAndState(
@@ -52,7 +51,7 @@ public interface MutableStore extends ReadOnlyStore {
         blockAndState.getBlock(),
         blockAndState.getState(),
         checkpoints,
-        Collections.emptyList(),
+        Optional.empty(),
         Optional.empty());
   }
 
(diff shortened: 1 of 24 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/MutableStore.java
```diff
@@ -43,7 +42,7 @@ void putBlockAndState(
       SignedBeaconBlock block,
       BeaconState state,
       BlockCheckpoints checkpoints,
-      List<BlobSidecar> blobSidecars,
+      Optional<List<BlobSidecar>> blobSidecars,
```
tbenr, line 45: it requires javadoc to be updated

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -308,6 +307,13 @@ public void startMetrics() {
                   TekuMetricCategory.STORAGE,
                   "memory_block_count",
                   "Number of beacon blocks held in the in-memory store"));
+      blobSidecarsBlocksCountGauge =
+          Optional.of(
+              SettableGauge.create(
+                  metricsSystem,
+                  TekuMetricCategory.STORAGE,
+                  "memory_blob_sidecars_blocks_count",
+                  "Number of complete blob sidecars blocks held in the in-memory store"));
```
tbenr: this is a bit confusing. We could say "Number of beacon blocks with complete blobs held in the in-memory store" or similar.

maybe the metrics could change as well, like "memory_blocks_with_cached_blobs_count" ?

tbenr (approved):
LGTM. Just a couple of nits

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/forkchoice/MutableStore.java
zilm13: Good point!

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
zilm13: Changed, sounds better

PR #7594, opened 2023-10-13
Make blocks cache locked in Store when used + minor cleanup

zilm13:
## PR Description
I've noticed that blocks are not actually locked if we go in block provider, after inMemory lookup in `retrieveSignedBlock` (it could actually changed between) and when directly using `blockProvider`, so tried to fix it.
## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -178,7 +178,7 @@ private Store(
                         .getSignedBeaconBlock()
                         .map((b) -> Map.of(b.getRoot(), b))
                         .orElseGet(Collections::emptyMap)),
-            fromMap(this.blocks),
```
rolfyone, line 181: not sure if we added fromMap for the blocks access, does it need to be removed?

zilm13: I don't understand you
I'm changing `fromMap` to `fromMapWithLock` in this PR
do you mean that we should completely remove blocks from map source from provider?

rolfyone: was more of a question of if fromMap can be cleanly removed from code altogether now...

looks like its used elsewhere still, so we can't - all good.

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
```diff
@@ -37,6 +38,22 @@ static BlockProvider fromDynamicMap(Supplier<Map<Bytes32, SignedBeaconBlock>> ma
     return (roots) -> fromMap(mapSupplier.get()).getBlocks(roots);
   }
 
+  static BlockProvider fromMapWithLock(
```
rolfyone, line 41: should probably add a test case

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
zilm13: Ahh yes, it's used by other providers

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
zilm13: Thanks you asked for it, I made all stupidly wrong, fixed

zilm13: I did the wrong thing again, thinking how to do it correctly

zilm13: So my understanding on this is:
1. We should remove `SafeFuture.supplyAsync()` to avoid such errors, noone was using it, and we should use our own AsyncRunner whenever it's needed
2. In this case we should use our AsyncRunner 
3. We should also use it in Store methods that use locks and return SafeFuture otherwise they are synchronously locked (`retrieveFinalizedCheckpointAndState`, `createStateGenerationTask`)
4. It will require rework of many `Store` tests

What do you think?

ethereum/dataproviders/src/test/java/tech/pegasys/teku/dataproviders/lookup/BlockProviderTest.java
```diff
@@ -132,4 +141,50 @@ void fromDynamicMap() {
...
+                      provider.getBlock(block.getRoot());
+                  notBlockedOnFuture.set(true);
+                  // Attempt to read in other thread - will stuck here
+                  assertThat(blockFuture.get()).contains(block);
+                } finally {
+                  isNotBlocked.countDown();
+                }
+              })
+          .ifExceptionGetsHereRaiseABug();
+      assertThat(isNotBlocked.await(1, TimeUnit.SECONDS)).isFalse();
```
rolfyone: even something small like 100ms should be plenty here?

zilm13: Sure

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
rolfyone: i think we could raise a ticket for that cleanup, but probably not worth it now, it'd be a big distraction if its used a bunch in tests...
Using async here though would make sense so we dont add to the problem...

zilm13:
@rolfyone finished, ready for review
1. Somehow test failed on Windows (at least once, on `notBlockedOnFuture`) with 100ms delay. I don't want to add any flakiness, pushed it to 500ms. 
2. Final fixes for other methods are in separate PR, based on this one, need to merge this one at first

rolfyone (approved):
LGTM

zilm13:
Disabled windows test for a while, will check on my local windows

zilm13:
Fixed flakiness with https://github.com/Consensys/teku/pull/7594/commits/50a3be1dc1cb71d284991e1357848e0d8616a359

PR #7596, opened 2023-10-15
added a sleep while doing large bls batches

rolfyone:
## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
@@ -144,6 +144,14 @@ public class NodeDataProvider {
 
     for (int i = 0; i < blsToExecutionChanges.size(); i++) {
       maybeFutureErrors.add(addBlsToExecutionChange(i, blsToExecutionChanges.get(i)));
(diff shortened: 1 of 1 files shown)


rolfyone:
basically we were adding way too many signatures over a 20k batch, and overflowing buffers... this is a low tech solution which I used for network updates.

lucassaldanha:
I'd rather have a limit on the number of BlsToExecutionChanges operations we can receive via a single API request. I think this idea of throttling the inclusion on the API side adds unnecessary complexity to our API layer.

tbenr:
I'm with Lucas here.. forcing user to split in multiple calls seems better. I guess there is no `limit` mention in API spec but in this particular case I don't think it's a big deal to have some "client diversity".

PS: `Thread.sleep` in production code is difficult to accept :)

rolfyone:
well the sleeps are machine dependent as to the size, but the main point of this was we were getting 500 errors on the api, which is way worse... and I didnt want to lose track of it.

We can put in a limit, our queue limit is 10k, and i don't have 10k to test with now but hopefully that won't overflow the signing queue again...

rolfyone:
raised #7601

Issue #7570, opened 2023-10-03
ProtoArray quadratic complexity when chain is not finalized 


Nashatyrev:
### Description

`ProtoArray.applyScoreChanges()/findOptimisticHead()` methods has quadratic complexity when the chain if not finalized 

### Steps to Reproduce (Bug)

The following test running times depending on `nonFinalizedBlockCount`:
- 10000 - 1s
- 20000 - 5s
- 40000 - 25s

```diff
Index: storage/src/test/java/tech/pegasys/teku/storage/protoarray/ProtoArrayTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/storage/src/test/java/tech/pegasys/teku/storage/protoarray/ProtoArrayTest.java b/storage/src/test/java/tech/pegasys/teku/storage/protoarray/ProtoArrayTest.java
--- a/storage/src/test/java/tech/pegasys/teku/storage/protoarray/ProtoArrayTest.java	(revision 6adad2a394243b0013b6031af2ebb9d69019abdc)
+++ b/storage/src/test/java/tech/pegasys/teku/storage/protoarray/ProtoArrayTest.java	(date 1696495894392)
@@ -67,6 +67,44 @@
     addOptimisticBlock(0, Bytes32.ZERO, Bytes32.ZERO);
   }
 
+
+  @Test
+  void testNoQuadraticComplexity() {
+    int nonFinalizedBlockCount = 40000;
+    Bytes32 parentBlockRoot = GENESIS_CHECKPOINT.getRoot();
+
+    Checkpoint justifiedCheckpoint = GENESIS_CHECKPOINT;
+    Checkpoint finalizedCheckpoint = GENESIS_CHECKPOINT;
+    int epoch = 0;
+    for (; epoch < nonFinalizedBlockCount; epoch++) {
+      UInt64 slot = UInt64.valueOf(8L * epoch + 1);
+
+      Bytes32 newBlockRoot = dataStructureUtil.randomBytes32();
+
+      if (epoch == 1) {
+        justifiedCheckpoint = new Checkpoint(UInt64.valueOf(epoch), newBlockRoot);
+        finalizedCheckpoint = new Checkpoint(UInt64.valueOf(epoch), parentBlockRoot);
+      }
+
+      protoArray.onBlock(
+              slot,
+              newBlockRoot,
+              parentBlockRoot,
[...]

PR #7598, opened 2023-10-16
Updated ProtoArray#nodeLeadsToViableHead to check own node viability before best descendant

lucassaldanha:
## PR Description

When checking the viability of a node, we consider a node to be viable if either itself or its best descendant is viable. The current implementation checks the viability of the best descendant before checking the viability of the current node.

Part of the viability check involves traversing the array, following the node's ancestry (see ProtoArray#hasAncestorAtSlot).

When we need to update descendants, we visit every node and check its viability. Consider a scenario of long non-finality and no forks, where we have n nodes, where n[i].parent = n[i-1]. When we are checking the viability of node n[i] (where i is the index of the node and 0 <= i < n), we are going to traverse the array for its best descendant which is always going to be n[n-1] (the worst case traversing the whole array).

With this change, in those scenarios, we are moving from the worst case towards the best case as we visit the nodes. This change improves the time it takes to perform the operations.

## Fixed Issue(s)
related to #7570

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
@@ -597,13 +597,14 @@ public class ProtoArray {
    * the head.
    */
   private boolean nodeLeadsToViableHead(ProtoNode node) {
-    boolean bestDescendantIsViableForHead =
-        node.getBestDescendantIndex()
-            .map(this::getNodeByIndex)
-            .map(this::nodeIsViableForHead)
-            .orElse(false);
+    if (nodeIsViableForHead(node)) {
+      return true;
+    }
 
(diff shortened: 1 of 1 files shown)


storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -597,13 +597,14 @@ private void changeToNone(ProtoNode parent) {
    * the head.
    */
   private boolean nodeLeadsToViableHead(ProtoNode node) {
```
rolfyone: i know this wasn't you, but can param be final :)

rolfyone (approved):
given the original function was a logical or, this looks non-controversial...

one nit just outside of what's been changed...

PR #7599, opened 2023-10-16
Don't send empty list of validator registrations to the Beacon node

StefanBratanov:
## PR Description
An issue discovered by a discord user: https://discord.com/channels/697535391594446898/697539289042649190/1163444580415455283

Because we are checking if builder is enabled for a specific validator after creating the batches of validators, we can end up calling `validatorApiChannel.registerValidators` with an empty list. This PR filters the validators which are not enabled in the beginning of the flow.

Also changed `SignedValidatorRegistrationFactory` to not return an Optional. I think this class should be agnostic to whether builder is enabled or not for a validator. Also fee recipient should pretty much never be not specified so changed it to throw an exception and raise a warning in the logs. This also removes the need of `Pair` usage in the `ValidatorRegistrator`.

- Moved `proposerConfigPropertiesProvider.refresh()` on top because it is used before the `registerValidators` method as well.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/SignedValidatorRegistrationFactory.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/SignedValidatorRegistrationFactory.java
@@ -62,7 +62,7 @@ public class SignedValidatorRegistrationFactory {
             .orElseGet(
                 () -> {
                   LOG.warn(
-                      "Fee recipient not configured for {}. Will use a burn address for the validator registration",
+                      "Fee recipient not configured for {}. Will use a burn address for the validator registration.",
                       validator.getPublicKey());
                   return Eth1Address.ZERO;
                 });
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/SignedValidatorRegistrationFactory.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/SignedValidatorRegistrationFactory.java
@@ -42,31 +42,30 @@ public class SignedValidatorRegistrationFactory {
     this.timeProvider = timeProvider;
   }
 
-  public Optional<SafeFuture<SignedValidatorRegistration>> createSignedValidatorRegistration(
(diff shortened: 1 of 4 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/SignedValidatorRegistrationFactory.java
```diff
@@ -32,57 +32,41 @@
...
-    final UInt64 gasLimit = validatorRegistrationPropertiesProvider.getGasLimit(publicKey);
+    final Eth1Address feeRecipient =
+        proposerConfigPropertiesProvider
+            .getFeeRecipient(publicKey)
+            .orElseGet(
+                () -> {
+                  LOG.warn(
+                      "Fee recipient not configured for {}. Will use a burn address for the validator registration.",
+                      validator.getPublicKey());
+                  return Eth1Address.ZERO;
```
zilm13: My concern is previously we didn't register such validators, so they were able to produce blocks locally where fee recipient could be configured on beacon node, since this change they will be registered and it could be fee loss

zilm13 (commented):
LGTM but I have some concerns on burning fee recipient

validator/client/src/main/java/tech/pegasys/teku/validator/client/SignedValidatorRegistrationFactory.java
tbenr: I think it's ok. We are on the VC side, and we enforce the presence of a default fee recipient otherwise the client doesn't even start. `validatorRegistrationPropertiesProvider.getFeeRecipient` returns an `Optional` but should never happen in practice.

StefanBratanov: Changed it to throwing an exception in which case only that registration will be skipped with a clear error message. Think it's better.

zilm13 (approved):
LGTM

PR #7603, opened 2023-10-17
change `getAllBlobSidecarsAtSlot` API tag

tbenr:
This should be a teku specific API, similar to `GetAllBlocksAtSlot`

## Documentation

## Changelog


zilm13 (approved):
There are some errors to fix, but makes sense.
I see it's already in right package with correct path to be in `Teku` scope, just tag is wrong

PR #7604, opened 2023-10-17
Removing synchronous lock in methods returning future in Store

zilm13:
## PR Description
Finishing work on removing synchronous lock in methods returning future in Store

## Documentation

## Changelog


tbenr:
As discussed async, we may don't want to merge this since the usage of this is mainly during blockImport call, done on incoming gossip (or syncing task), so on a thread allocated from the thread pool. Adding another `asyncRunner` interaction it will just add another extra step (thread switch on the same thread pool) with no benefits cose nothing seems actually blocked by this. I hope I captured the gist of the convo we had.

zilm13:
@tbenr yeah, closing this
callers of `retrieveFinalizedCheckpointAndState` alreday runs in async threads so blocking it is not an issue

PR #7606, opened 2023-10-19
Implemented isFinalizedRootOrDescendant

lucassaldanha:
## PR Description

When checking if a node is viable as head (`ProtoArray#nodeIsViableForHead`), we were navigating backwards the nodes to check if the node was a descendant of a finalized checkpoint.

This change adds a sort of "shortcut" (ProtoArray#isFinalizedRootOrDescendant) to this check, by using the head-candidate node's `BlockCheckpoints`, checking if either of these is true:
- Does the node's finalized epoch and finalized root match the ProtoArray view of the finalized epoch and root?
- Does the node's justified epoch and justified root match the ProtoArray view of the finalized epoch and root?
- Does the node's unrealized finalized epoch and finalized root match the ProtoArray view of the finalized epoch and root?
- Does the node's unrealized justified epoch and finalized root match the ProtoArray view of finalized epoch and root?

If none of these are true, we fall back into the old traverse of the hierarchy until we find the finalized checkpoint for that head candidate.

## Fixed Issue(s)
related to #7570

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -639,13 +639,44 @@ public boolean nodeIsViableForHead(ProtoNode node) {
...
+    }
+
+    final Checkpoint nodeUnrealizedJustifiedCheckpoint = node.getUnrealizedJustifiedCheckpoint();
+    if (nodeUnrealizedJustifiedCheckpoint.getEpoch().equals(finalizedEpoch)
+        && nodeUnrealizedJustifiedCheckpoint.getRoot().equals(finalizedRoot)) {
+      return true;
+    }
+
+    final UInt64 finalizedSlot = spec.computeStartSlotAtEpoch(finalizedCheckpoint.getEpoch());
+    return hasAncestorAtSlot(node, finalizedSlot, finalizedRoot);
```
rolfyone, line 677: im not sure how we can ever get here but fail the check at 653?

rolfyone: ok i figured this out, its complicated and we can talk about why if we need to but i'll resolve.

rolfyone (approved):
LGTM

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
tbenr: I agree with the @rolfyone 's doubts that there might be no cases where we don't match the 4 conditions and we ending up returning `true` via `hasAncestorAtSlot`.

But I tried to hardcode `return false` at the end run fork_choice ref tests and fails quite a lot of them. So there are definitely cases!

PR #7611, opened 2023-10-20
Added configuration parameters for holesky testing

rolfyone:
## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/constants/NetworkConstants.java
```diff
@@ -18,5 +18,14 @@ public class NetworkConstants {
...
   public static final int DEFAULT_SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY = 128;
+
+  public static final int DEFAULT_ASYNC_P2P_MAX_THREADS = 10;
+
+  public static final int DEFAULT_ASYNC_P2P_MAX_QUEUE = 5000;
+
+  public static final int DEFAULT_ASYNC_BEACON_CHAIN_MAX_THREADS =
+      Math.max(Runtime.getRuntime().availableProcessors(), 5);
+
+  public static final int DEFAULT_ASYNC_BEACON_CHAIN_MAX_QUEUE = 5000;
```
tbenr: maybe we could have these in `Eth2NetworkConfiguration`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/constants/NetworkConstants.java
```diff
@@ -18,5 +18,14 @@ public class NetworkConstants {
   public static final int SYNC_COMMITTEE_SUBNET_COUNT = 4;
   public static final int INTERVALS_PER_SLOT = 3;
   public static final int DEFAULT_SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY = 128;
+
+  public static final int DEFAULT_ASYNC_P2P_MAX_THREADS = 10;
+
+  public static final int DEFAULT_ASYNC_P2P_MAX_QUEUE = 5000;
+
+  public static final int DEFAULT_ASYNC_BEACON_CHAIN_MAX_THREADS =
```
tbenr: I'd also put a comment here to link back the that 5 to `DEFAULT_EXECUTOR_THREADS` (executorThreads from VC).

Or even better, find a new home to `DEFAULT_EXECUTOR_THREADS` so it is visible from everywhere and we can still enforce the `max` safely.

and we could enforce this in the CLI options so that if user gets crazy and set it to 2 we can error out.

tbenr (commented):
just a couple of comments.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/constants/NetworkConstants.java
rolfyone: hrm i put those there because some others are in networkConstants but i'll have another look...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/constants/NetworkConstants.java
rolfyone: ok i think i've basically encapsulated all of this...

tbenr (approved):
LGTM.

PR #7616, opened 2023-10-20
Ignore excessive blobSidecars

### LINKED ISSUE #7610: Allow block import even if we see more blobSidecars than commitments in block
To be consistent with other clients on the agreed behaviour, we must ignore blob sidecars with an index > `block.kzgCommitments.size`.

We currently fail the DA check and we don't import the block.

tbenr:
Implements a pruning\filtering in `BlockBlobSidecarsTracker` to keep track only of blobSidecars with an index matching a commitment in the block

fixes #7610

## Documentation

## Changelog


StefanBratanov (approved):
Makes sense. LGTM.

PR #7618, opened 2023-10-20
Add block arrival time in metrics at the earliest possible point

### LINKED ISSUE #7536: Get actual block arrival time from gossip
Moving forward, it's going to be important for us to be able to determine when a block arrived, which we don't currently have.

Currently in late block import messages, we do 'say' arrival, but its when we start processing.

We need to be able to get the gossip message timestamp when processing the actual libp2p message, and start reporting that in 'late block import'.

This new time would be the 'arrival', and the current value would be when we start processing. We may also wish to report the time it was validated, but thats not a 'requirement' per say.

Once we have this data point it will unblock us from implementing the reorg late block logic in fork choice.

zilm13:
## PR Description
There are 2 options how we could get arrival block time earlier than in current `master` branch flow (https://github.com/Consensys/teku/blob/master/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java#L150-L158):

1. Less invasive, without changes in libp2p library, the way it's implemented in the current PR. We have early entrypoint for all messages, message factory, where we have full control using libp2p API https://github.com/ConsenSys/teku/blob/f31b239f7788f07d8e9c2b6e7544c5e16b01a259/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/gossip/LibP2PGossipNetworkBuilder.java#L107. It's called very early in libp2p https://github.com/libp2p/jvm-libp2p/blob/1a5f08d0a05aca80dcef8be5708aebf7b77d8186/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt#L161-L191 and it's less complicated to implement than the other option. We return `PreparedPubsubMessage` from it which is pretty easy to extend, what I did, putting time there. And here appears the first issue with such improvement: message type is more or less abstract on this stage and we cannot use specific type performance meter there. So the only way I could go is to use the same TimeProvider for time data, which we use in performance metrics. Yet, from the implementation standpoint it's not clear that we have the same time source in this measurement.<br /> 
The result of measurement of delay between previously measured arrival point, which I renamed to `pre_validate` (not a final name) goes below:
```
		Total late block arrivals for ~2 days: ~300 blocks of 15,000
		arrival to pre_validate (min, max, avg): 1ms, 101ms, 3.74ms (98% < 10ms)
```
2. Another way is to extend libp2p, which is not that simple from many points of view:
  -   Libp2p message factory interface is `(Rpc.Message) -> PubsubMessage` where the input is pubsub message type, which we shouldn't extend. As libp2p API is public and is used not only by the Teku, we should not break the API, instead we could introduce something like Experimental Message Factory, do a check in libp2p builder if it's provided and use it in Teku instead of the base factory.
  -    Between message factory and message arrival there are only minor computations. If you look at the pipeline https://github.com/libp2p/jvm-libp2p/blob/1a5f08d0a05aca80dcef8be5708aebf7b77d8186/libp2p/src/main/kotlin/io/libp2p/pubsub/AbstractRouter.kt#L116-L123 (we need only decoders from it for inbound messages) there is `LimitedProtobufVarint32FrameDecoder` which just detects which multiplexer frames belongs to one message and pushes them next. It's safe to say it adds 0ms after last frame. After it we have `ProtobufDecoder` which decodes raw bytes data to actual message, but it's a pretty optimized piece of software and protobuf decoding is fast. So it adds few milliseconds between message factory and last frame of the message. There is little value to count this time in consideration of complexity.
  Alternatively we could extend wrapper for every frame of the message with it's own timestamp and start measurement from the first frame complete arrival. This way we will measure delay based on badndwidth throughput, it will not measure anything on computation side.
  -   If we do anything with time we should have the same time source as we use in Teku, which should be a pain to implement. Alternatively we could track some delta only inside libp2p with it's own time source and provide only delta which we could safely call something like `transfer_time` because, there is only minor computation before API configured message factory, so it will actually measure message transfer time.
  -    It will still require piece of software (1) to extract this data in experimental message factory and pass in the Teku

## Fixed Issue(s)

Fixes #7536 
## Documentation

## Changelog


zilm13:
Cleaned up a bit. Ready for review

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/gossip/LibP2PGossipNetworkBuilder.java
```diff
@@ -64,19 +67,34 @@ public static LibP2PGossipNetworkBuilder create() {
...
+  private void validate() {
+    assertNotNull("metricsSystem", metricsSystem);
+    assertNotNull("gossipConfig", gossipConfig);
+    assertNotNull("networkingSpecConfig", networkingSpecConfig);
+    assertNotNull("defaultMessageFactory", defaultMessageFactory);
+    assertNotNull("gossipTopicFilter", gossipTopicFilter);
+    assertNotNull("timeProvider", timeProvider);
+  }
+
+  private void assertNotNull(String fieldName, Object fieldValue) {
```
tbenr: `final`s :-D

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/gossip/PreparedGossipMessageFactory.java
```diff
@@ -13,12 +13,18 @@
...
 /** Factory for {@link PreparedGossipMessage} instances */
 public interface PreparedGossipMessageFactory {
 
   /** Creates a {@link PreparedGossipMessage} instance */
-  PreparedGossipMessage create(String topic, Bytes payload, NetworkingSpecConfig networkingConfig);
+  PreparedGossipMessage create(
+      String topic,
+      Bytes payload,
+      NetworkingSpecConfig networkingConfig,
+      Optional<UInt64> timestamp);
```
tbenr: `arrivalTimestamp`? to be more clear as you already did in the Snappy implementation

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/gossip/LibP2PGossipNetworkBuilder.java
```diff
@@ -109,14 +127,17 @@ protected GossipRouter createGossipRouter(
...
+          final UInt64 timestamp = timeProvider.getTimeInMillis();
           String topic = msg.getTopicIDs(0);
           Bytes payload = Bytes.wrap(msg.getData().toByteArray());
 
           PreparedGossipMessage preparedMessage =
               topicHandlers
                   .getHandlerForTopic(topic)
-                  .map(handler -> handler.prepareMessage(payload))
-                  .orElse(defaultMessageFactory.create(topic, payload, networkingSpecConfig));
+                  .map(handler -> handler.prepareMessage(payload, Optional.of(timestamp)))
```
tbenr: Why the timestamp becomes optional here?.  I don't see `default` implementations on interfaces to pass empty and be backward compatible. So I don't see the point of having it optional (I may find the reason continuing the review :) )

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/BlockGossipManager.java
```diff
@@ -13,17 +13,29 @@
...
+import tech.pegasys.teku.networking.eth2.gossip.topics.topichandlers.TimedEth2TopicHandler;
 import tech.pegasys.teku.networking.p2p.gossip.GossipNetwork;
+import tech.pegasys.teku.networking.p2p.gossip.TopicChannel;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.datastructures.state.ForkInfo;
 import tech.pegasys.teku.storage.client.RecentChainData;
 
-public class BlockGossipManager extends AbstractGossipManager<SignedBeaconBlock> {
+public class BlockGossipManager implements GossipManager {
```
tbenr: what if we change the `OperationProcessor:process` to accept a Metadata or Context class in which we store stuff. And we make it optional with a default empty implementation to be backward compatible. 
This whay we can avoid Timed* duplications and `arrivalTimestamp` will be in it. We can potentially store the peerid or something else in the future.

Or, for simplicity, we could have directly the optional timestamp for now there, that can become a context when\if we will add more stuff.

tbenr (commented):
dropping some comments

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/gossip/LibP2PGossipNetworkBuilder.java
zilm13: I just didn't want to put it everywhere, while we need it only in blocks. Some memory saving. But your feedback sounds reasonable, not much profit from such optional.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/BlockGossipManager.java
zilm13: I wanted to keep OperationProcessor and Eth2TopicHandler as is, so we don't need to change 100 of classes. I will try your suggestion to compare 2 approaches

zilm13:
Pushed simple version according to the review. Ready for review.
When adding more meta fields, I would prefer `OperationAndMeta` container rather that passing `Meta`container with operation as 2 fields, because some duties may require metadata later than object arrival and may require complicated additional structures to store them (futureBlocks, futureBlobSidecars, for example, may require peerId for penalization when processed).

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/block/BlockManagerTest.java
```diff
@@ -640,7 +641,8 @@ public void onValidateAndImportBlock_shouldEarlyRejectInvalidBlocks() {
     assertValidateAndImportBlockRejectWithoutValidation(invalidBlock);
 
     // Gossip invalid block descendants, must reject with no actual validation
-    invalidBlockDescendants.forEach(this::assertValidateAndImportBlockRejectWithoutValidation);
+    invalidBlockDescendants.forEach(
```
tbenr: you can roll this back and leave the lambda

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -981,6 +981,7 @@ protected void initP2PNetwork() {
             .keyValueStore(keyValueStore)
             .requiredCheckpoint(weakSubjectivityValidator.getWSCheckpoint())
             .specProvider(spec)
+            .recordMessageArrival(true)
```
tbenr, line 1014: I don't get the benefits of setting this flag at P2PNetwork level, since it is hardcoded to be true.
If move this flag into `Eth2TopicHandler` constructor, than we can activate the recording per topic (in this case we can activate it only for blocks) and let  the messageFactory in `LibP2PGossipNetworkBuilder.createGossipRouter` dynamically chose base on topic handler if we want to get current time or not. Otherwise, for simplicity, we could always get current time and remove this flag.

tbenr (approved):
LGTM. just a nit and a consideration

PR #7620, opened 2023-10-23
Improve slow db warning

tbenr:
Consider the whole `doUpdate` method timings to consider an update as "slow".
Provide additional info when `doUpdate` is "slow".

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #7621, opened 2023-10-23
Simplify loading/freeing trusted setup

StefanBratanov:
## PR Description

- Allow reloading different trusted setup without need to free the current one. 
- Simplify tests (not reload trusted setups all the time)

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
+++ b/infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
@@ -20,13 +20,16 @@ import org.apache.tuweni.bytes.Bytes;
  * This interface specifies all the KZG functions needed for the Deneb specification and is the
  * entry-point for all KZG operations in Teku.
  */
-public interface KZG extends AutoCloseable {
+public interface KZG {
 
   KZG NOOP =
       new KZG() {
 
         @Override
         public void loadTrustedSetup(final String trustedSetupFile) throws KZGException {}
+
+        @Override
+        public void freeTrustedSetup() throws KZGException {}
 
         @Override
         public boolean verifyBlobKzgProofBatch(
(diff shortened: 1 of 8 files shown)


infrastructure/kzg/src/test/java/tech/pegasys/teku/kzg/KZGTest.java
```diff
@@ -102,11 +92,12 @@ public void testUsageWithoutLoadedTrustedSetup_shouldThrowException() {
     assertThat(exceptions)
         .allSatisfy(
             exception -> assertThat(exception).cause().hasMessage("Trusted Setup is not loaded."));
+    // load trusted setup again for other tests
+    loadTrustedSetup();
```
zilm13: I'd better add `@BeforeEach` with it, it would be more clean, so author of the new test shouldn't think about whole test suite behavior

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
```diff
@@ -20,15 +20,13 @@
  * This interface specifies all the KZG functions needed for the Deneb specification and is the
  * entry-point for all KZG operations in Teku.
  */
-public interface KZG {
+public interface KZG extends AutoCloseable {
```
zilm13: I'd add a comment that freeing could be expensive and shouldn't be used in production. Another concern is that when someone adds it in `try()` in production, the kzg will be killed in app till restart. So I think we'd better remove interface to avoid errors in production.

zilm13 (commented):
I think passing kzg as `AutoCloseable` could be dangerous for production code, so my vote is for removing interface

infrastructure/kzg/src/main/java/tech/pegasys/teku/kzg/KZG.java
StefanBratanov: Removed interface

StefanBratanov:
> I think passing kzg as `AutoCloseable` could be dangerous for production code, so my vote is for removing interface

Agree. Removed it.

zilm13 (approved):
LGTM

PR #7624, opened 2023-10-24
Capture time taken creating, signing and sending blocks on block production duty

### LINKED ISSUE #7612: Capture time taken creating, signing and sending blocks on block production duty
Now that we have started timing the total time of duties, we can start timing specific steps of the duty. In this ticket, we want to capture the time to create, sign, and send a block.

**Requirements**
- block creation time should be captured using the same operation timer we created on `ValidatorDutyMetrics`, but with the labels type=block_production, step=create.
- block signing time should be captured using the same operation timer we created on `ValidatorDutyMetrics`, but with the labels type=block_production, step=sign.
- block publishing time should be captured using the same operation timer we created on `ValidatorDutyMetrics`, but with the labels type=block_production, step=send.

courtneyeh:
fixes #7612 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/ValidatorDutyMetrics.java
```diff
@@ -38,9 +39,22 @@ public static ValidatorDutyMetrics create(final MetricsSystem metricsSystem) {
...
+    final OperationTimer timer = dutyMetric.labels(dutyType, step);
+    return timer.startTimer();
+  }
+
+  private static String getDutyType(final Duty duty) {
+    return duty.getType().getType();
+  }
+
+  public <T> SafeFuture<T> record(
+      final Supplier<SafeFuture<T>> supplier, final String dutyType, final String step) {
```
lucassaldanha: I think we could change `dutyType` from `String` to `Duty` and then use the same `getDutyType(duty)` method to get the type.

This should make this a bit more consistent and there will be no need to call `final String dutyType = getType().getType();` outside of the scope of this method.

I think I'd also rename `supplier` to something a bit more meaningful like `dutyStepFutureSupplier`.

lucassaldanha (approved):
LGTM

Issue #7617, opened 2023-10-20
Investigate regular failure of BlobSidecarsByRangeIntegrationTest


StefanBratanov:
As part of https://github.com/Consensys/teku/pull/7609 `BlobSidecarsByRangeIntegrationTest.requestBlobSidecars_shouldReturnCanonicalBlobSidecarsOnDenebMilestone()` was failing regularly on the line `assertThat(peer.getOutstandingRequests()).isEqualTo(0);`. 

That line was commented out in order for the test to pass on CI and the condition afterwards were still passing but it needs to be investigated why this assertion fails. It would be most likely related to the increase of the size of the blobs given that minimal uses the same FIELD_ELEMENTS_PER_BLOB as mainnet.

mehdi-aouadi:
@StefanBratanov This is normally fixed, right? Should we close it?

StefanBratanov:
Ah yes, thanks. Closing that now. It has been fixed as part of https://github.com/Consensys/teku/pull/7627

PR #7627, opened 2023-10-25
fix blob sidecars by range integration test

mehdi-aouadi:
## PR Description
Fix broken `BlobSidecarsByRangeIntegrationTest`

## Fixed Issue(s)

#7617 

## Documentation

## Changelog


StefanBratanov:
The default seems to be 30 seconds, so on regular basis it takes longer with the new trusted setup?

mehdi-aouadi:
> The default seems to be 30 seconds, so on regular basis it takes longer with the new trusted setup?

I just pushed a change to use the default timeout :) Let's see!

mehdi-aouadi:
@StefanBratanov the default 30s timeout looks ok

StefanBratanov (approved):
LGTM

PR #7630, opened 2023-10-25
Capture time taken creating and signing attestations in AttestationProductionDuty

### LINKED ISSUE #7613: Capture time taken creating,signing attestations on attestation production duty
Now that we have started timing the total time of duties, we can start timing specific steps of the duty. In this ticket, we want to capture the time it takes to create and sign an attestation.

**Requirements**
- attestation creation time should be captured using the same operation timer we created on `ValidatorDutyMetrics`, but with the labels type=attestation_production, step=create.
- attestation signing time should be captured using the same operation timer we created on `ValidatorDutyMetrics`, but with the labels type=attestation_production, step=sign.

courtneyeh:
fixes #7613 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/attestations/AttestationProductionDuty.java
```diff
@@ -121,14 +125,21 @@ private List<SafeFuture<ProductionResult<Attestation>>> produceAttestationsForCo
...
     unsignedAttestationFuture.propagateTo(committee.getAttestationDataFuture());
 
     return committee.getValidators().stream()
         .map(
             validator ->
-                signAttestationForValidatorInCommittee(
-                    slot, forkInfo, committeeIndex, validator, unsignedAttestationFuture))
+                validatorDutyMetrics.record(
+                    () ->
+                        signAttestationForValidatorInCommittee(
```
lucassaldanha: Adding our metric wrapper here means that we will time the "sign" step even when the attestation was not produced. We should do it inside `signAttestationForValidator`.

This is why the test `shouldFailWhenUnsignedAttestationCanNotBeCreated` is registering a call for SIGN.

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/AttestationProductionDutyTest.java
```diff
@@ -237,6 +257,11 @@ public void shouldPublishProducedAttestationsWhenSignerFailsForSomeAttestations(
         .dutyFailed(
             TYPE, SLOT, Set.of(validator1.getPublicKey().toAbbreviatedString()), signingFailure);
     verifyNoMoreInteractions(validatorLogger);
+
+    verify(validatorDutyMetrics)
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.CREATE));
+    verify(validatorDutyMetrics, times(2))
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.SIGN));
```
lucassaldanha, line 261: This is correct. In this test, we have two validators in the same committee, so we produce the same attestation for both of them. But they sign the attestation individually. That's why we have 1 create and 2 sign.

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/AttestationProductionDutyTest.java
```diff
@@ -198,6 +213,11 @@ public void shouldPublishProducedAttestationsWhenSomeUnsignedAttestationsFail()
     verify(validatorLogger)
         .dutyFailed(TYPE, SLOT, Set.of(validator1.getPublicKey().toAbbreviatedString()), failure);
     verifyNoMoreInteractions(validatorLogger);
+
+    verify(validatorDutyMetrics, times(2))
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.CREATE));
+    verify(validatorDutyMetrics, times(2))
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.SIGN));
```
lucassaldanha: This shouldn't happen twice. One of the attestations isn't being produced so we shouldn't count its signing (it never happens).

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/AttestationProductionDutyTest.java
```diff
@@ -345,6 +378,11 @@ public void shouldCreateAttestationForMultipleValidatorsInSameCommittee() {
         .dutyCompleted(
             TYPE, SLOT, 3, Set.of(attestationData.getBeaconBlockRoot()), Optional.empty());
     verifyNoMoreInteractions(validatorLogger);
+
+    verify(validatorDutyMetrics)
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.CREATE));
+    verify(validatorDutyMetrics, times(3))
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.SIGN));
```
lucassaldanha, line 382: This one is also correct. All 3 validators are in the same committee.

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/AttestationProductionDutyTest.java
```diff
@@ -406,6 +444,11 @@ public void shouldCreateAttestationForMultipleValidatorsInDifferentCommittees()
                 unsignedAttestation2.getBeaconBlockRoot()),
             Optional.empty());
     verifyNoMoreInteractions(validatorLogger);
+
+    verify(validatorDutyMetrics, times(2))
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.CREATE));
+    verify(validatorDutyMetrics, times(3))
+        .record(any(), any(AttestationProductionDuty.class), eq(Step.SIGN));
```
lucassaldanha, line 448: This is correct. 2 validators are in one committee, and the other is in a different committee. So we need to create 2 attestations and each validator will sign.

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/attestations/AttestationProductionDuty.java
```diff
@@ -144,7 +152,12 @@ private SafeFuture<ProductionResult<Attestation>> signAttestationForValidatorInC
                 maybeUnsignedAttestation
                     .map(
                         attestationData ->
-                            signAttestationForValidator(slot, forkInfo, attestationData, validator))
+                            validatorDutyMetrics.record(
+                                () ->
+                                    signAttestationForValidator(
+                                        slot, forkInfo, attestationData, validator),
+                                this,
+                                Step.SIGN))
```
lucassaldanha: I think we should do the signing inside `signAttestationForValidator` (at line 185).

The reason being: in that method, there is a `checkArgument` that checks if we are even going to try signing it. We don't want it to mark the time of the sign operation when this check fails, because it has nothing to do with signing and this could artificially lower the average signing time.

lucassaldanha: Another option is moving the check out of that method and having a validateAttestation or something.

lucassaldanha (approved):

PR #7633, opened 2023-10-26
Implement finalized checkpoint gossip rule for `blobSidecar`

### LINKED ISSUE #7632: Add new gossip rule for blob_sidecar that checks parent block descends current finalized checkpoint
https://github.com/ethereum/consensus-specs/pull/3529

tbenr:
Add the new gossip rule by extracting the method implementing the check from `BlockGossipValidator` to the helper.

`BlockGossipValidatorTest` still test end-to-end, without mocking the validator helper dependency. At some point we could simplify most of the test there (and avoid some duplications)

`BlobSidecarValidator` renamed to `BlobSidecarGossipValidator` to be consistent with the Block one.

fixes #7632 

## Documentation

## Changelog


StefanBratanov (approved):
LGTM.

PR #7636, opened 2023-10-26
Check commitment/proof when unblinding blob sidecar

jtraglia:
## PR Description

Most likely unnecessary, but I think it would be a good idea to check that the commitment and proof in the blobs bundle (from the builder) matches the blinded blob sidecar we already have. I'm sure this would have been caught somewhere further into the process, but the sooner the better in my opinion. Only really an issue if the relay doesn't do its job.

As part of this, I renamed `blindedBlobSidecarIndex` to `index` and changed it from a `UInt64` to an `int`. I did this because I thought the previous name was too long and we used the integer value anyway.

## Documentation

## Changelog


tbenr (approved):
LGTM. Thanks

Issue #7622, opened 2023-10-23
Optimize `KvStoreDatabase::doUpdate`


tbenr:
When we perform `doUpdate`, we may require to serialize several objects (several blocks, a state, blobSidecars in Deneb). Even if the Storage operation is async (forkChoice is not blocked) the entire method is executed in a single thread.

We can benefit of having serialization spanning across multiple threads so we can leverage multiple cores and speed up the Storage operation to free up the StorageChannel faster (when there us a write operation, all reads are queued).

tbenr:
As a reference, on mainnet on my local machine I get:

```
2023-10-23 16:53:12.315 WARN  - DB operation LevelDbTransaction::commit took too long: 190 ms. The alert threshold is set to: 1 ms. Additional info: N/A

2023-10-23 16:53:12.319 WARN  - DB operation KvStoreDatabase::doUpdate took too long: 1314 ms. The alert threshold is set to: 1 ms. Additional info: Finalized data updated time: 605 ms - Hot data updated time: 709 ms of which latest finalized state updated time: 113 ms
```

So the commit itself was only 190ms, but the `doUpdate` took 1.3s and I think that most of the time has been spent in serialization operations.

tbenr:
well this is actually not correct:
![image](https://github.com/Consensys/teku/assets/15999009/2407cdb6-c108-4ea9-b2a7-41859410c8ec)

serialization is not the big portion of it. We are wasting a lot of time deserializing blocks to move them from hot to finalized

tbenr:
To optimize without changing the DB schema we should avoid getting blocks with `getHotBlocks` that uses `db.get`, we should add a new `getHotBlocks` using `db.getRaw` so we can return raw bytes instead of the serialized data.
But seems like it requires non trivial changes since the `updateFinalizedData` is based on `BlockProvider` functional interface.

PR #7638, opened 2023-10-27
[spike] Avoid block hashing when moving blocks from hot db to finalized

tbenr:
This just to see how good is avoiding hashing.

We could further improve by being able to get rawData (bytes) and write Raw data as well to even avoid serialization\deserialization.

image

```
2023-10-27 15:17:12.541 WARN  - DB operation KvStoreDatabase::doUpdate took too long: 1522 ms. The alert threshold is set to: 1 ms. Additional info: Finalized data updated time: 306 ms - Hot data updated time: 1216 ms of which latest finalized state updated time: 318 ms
```

related to #7622

## Documentation

## Changelog


tbenr:
Deployed on nighly-02 but seems like I don't get the benefits I was expecting..need to double-check

tbenr:
closing in favour of #7672

PR #7639, opened 2023-10-27
Remove need of kzgNoop when initialising Spec

StefanBratanov:
## PR Description

- Remove need of `kzgNoop` across Teku and all its usages. (quite a few places)
- Remove `KZG` initialisation in `MiscHelpersDeneb` and move it to `BeaconChainController`. The rest involves basically providing the `KZG` object to other classes which need it
- Removed `BlobsBundleValidator` (it is unused)

## Fixed Issue(s)
N/A

## Documentation

## Changelog


teku/src/test/java/tech/pegasys/teku/cli/BeaconNodeCommandTest.java
```diff
@@ -426,21 +426,6 @@ public void shouldParseDenebEpochsStoreBlobs() {
...
-            beaconNodeCommand
-                .tekuConfiguration()
-                .eth2NetworkConfiguration()
-                .getSpec()
-                .forMilestone(SpecMilestone.DENEB)
-                .getConfig());
-    assertThat(specConfigDeneb.isKZGNoop()).isFalse();
-  }
-
   private Path createConfigFile() throws IOException {
```
zilm13, line 438: I'd leave `shouldHaveKzgForDeneb` and check that trusted setup is set in eth2NetworkConfiguration

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -580,11 +577,6 @@ public Builder epochsStoreBlobs(final String epochsStoreBlobs) {
       return this;
     }
 
-    public Builder kzgNoop(final boolean kzgNoop) {
-      this.kzgNoop = Optional.of(kzgNoop);
-      return this;
-    }
-
     public Builder applyNetworkDefaults(final String networkName) {
```
zilm13, line 583: Just noticed this.
Could we put "mainnet-trusted-setup.txt" and "minimal-trusted-setup.txt" in static constants also?

zilm13 (commented):
I have mixed feelings on this. 

  We had `isDataAvailable` in `MiscHelpersDeneb` because it's a spec method, it's always good to see something in a code closer to the spec when following changes and updating after a period of time. Its logic was already spread across different parts of `BlobSidecarsAvailabilityChecker`, which is a bit complicated, but you were able to spy on how it looks like not combining separated parts in `BlobSidecarsAvailabilityChecker` with whole image built in one place and looking for "Found usages" in each line one by one.

  Also I don't like passing kzg directly to `Eth2Peer` for example, previously it was passed inside MiscHelpers and was unwwrapped only in `BlobSidecarsByRangeListenerValidatingProxy` which was a bit more abstract compared to the changes.

  On the other hand almost all noKzg burden was removed, which is a great improvement!

  Regarding `isDataAvailable` I think we could restore it adding an input field `Kzg kzg` just for showcase. Are you ok with it?
  Regarding exposing kzg. Looks like I need to accept it, the World is changing ;). It's better with BLS which has static instance, but we decided not to go this way with KZG and actually there could be an issue in such approach when running tests.
  
  Overall I think we should go this way, but if we could find a way to not expose kzg where it should not belongs to, it would be better

StefanBratanov:
> I have mixed feelings on this.
> 
> We had `isDataAvailable` in `MiscHelpersDeneb` because it's a spec method, it's always good to see something in a code closer to the spec when following changes and updating after a period of time. Its logic was already spread across different parts of `BlobSidecarsAvailabilityChecker`, which is a bit complicated, but you were able to spy on how it looks like not combining separated parts in `BlobSidecarsAvailabilityChecker` with whole image built in one place and looking for "Found usages" in each line one by one.
> 
> Also I don't like passing kzg directly to `Eth2Peer` for example, previously it was passed inside MiscHelpers and was unwwrapped only in `BlobSidecarsByRangeListenerValidatingProxy` which was a bit more abstract compared to the changes.
> 
> On the other hand almost all noKzg burden was removed, which is a great improvement!
> 
> Regarding `isDataAvailable` I think we could restore it adding an input field `Kzg kzg` just for showcase. Are you ok with it? Regarding exposing kzg. Looks like I need to accept it, the World is changing ;). It's better with BLS which has static instance, but we decided not to go this way with KZG and actually there could be an issue in such approach when running tests.
> 
> Overall I think we should go this way, but if we could find a way to not expose kzg where it should not belongs to, it would be better

I removed `isDataAvailable` because it wasn't used in the code actually after the latest refactoring on the `BlobSidecarsAvailabilityChecker`. I added a comment on top of `ForkChoiceBlobSidecarsAvailabityChecker` to mention the spec URL and the method name. Not sure how to restore it since it was unused.

I don't like passing kzg as well to `Eth2Peer` to be honest, but think i prefer it than having `kzgNoop` in so many places. I couldn't think of a better solution, but think this one is better than the two and think it makes sense this object to be part of it since the P2P now does kzg verification.

Static KZG makes some sense, but makes things hard to test and can't have NOOP,  so may will make things harder long term in my opinion.

I am happy to look at alternative approaches for certain bits if there are suggestions.

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
StefanBratanov: I am following the pattern how we specify `mainnet-genesis.ssz` for loading files from classpath. Not sure we want to change this for the trusted setup.

jtraglia:
> We had isDataAvailable in MiscHelpersDeneb because it's a spec method, it's always good to see something in a code closer to the spec when following changes and updating after a period of time.

Yes, I agree with this. Alignment with the spec is good. Makes reviewing easier.

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
zilm13: I mean just
`final private static String MINIMAL_TRUSTED = "minimal-trusted-setup.txt";`
You are probably didn't get me

StefanBratanov: yeah that makes sense, done!

tbenr:
I think the positive outpace the negatives, by far. We also lose the ability to switch kzg at runtime (it is not spec dependant) but who knows if we will ever need that.

zilm13:
I think I'm ok with it. But would pass approve to @tbenr

tbenr (approved):
LGTM

Issue #7635, opened 2023-10-26
Beacon API endpoint `/eth/v1/beacon/blob_sidecars/` returns 404 on empty blobs block


marioevz:
### Description
Beacon API endpoint `/eth/v1/beacon/blob_sidecars/` returns 404 on an empty-blobs block after Deneb, where it should be returning an empty list instead.

As per the beacon API, 404 error is reserved for empty slot: https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getBlobSidecars

This example is from devnet-10:

Teku:
```bash
marioevz@hive-simulations:~/Scripts$ curl https://$DEVNET_USER:$DEVNET_PASS@bn.teku-geth-1.dencun-devnet-10.ethpandaops.io/eth/v2/beacon/blocks/24141 | jq '.data.message.body.blob_kzg_commitments'
[]
marioevz@hive-simulations:~/Scripts$ curl https://$DEVNET_USER:$DEVNET_PASS@bn.teku-geth-1.dencun-devnet-10.ethpandaops.io/eth/v1/beacon/blob_sidecars/24141
{"code":404,"message":"Not found"}
```
Lighthouse:
```bash
marioevz@hive-simulations:~/Scripts$ curl https://$DEVNET_USER:$DEVNET_PASS@bn.lighthouse-geth-1.dencun-devnet-10.ethpandaops.io/eth/v1/beacon/blob_sidecars/24141
{"data":[]}
```

Also in [hive](https://hiveinterop.ethdevops.io/) teku is failing many of the tests due to this check. The failures are intermittent because most of the blocks in the hive devnets do have blobs, but in the longer running tests, such as the builder tests, there's almost always at least one slot with no blobs.

### Steps to Reproduce (Bug)
Query a teku node for a slot where there was a block but it had empty kzg commitments (example used is in devnet-10 with recent slot 24141)

### Versions (Add all that apply)
* Software version: latest `develop`

mehdi-aouadi:
Hi @marioevz, this happens when the block is not finalised and doesn't have any blob sidecar (the API returns an empty list when the block is finalised and doesn't have any blob sidecar).
[I raised a PR to fix this](https://github.com/Consensys/teku/pull/7640)

mehdi-aouadi:
A fix has been merged

marioevz:
Confirmed to be fixed in hive, thanks!

PR #7640, opened 2023-10-27
return empty blob sidecar list when no kzg commitments

mehdi-aouadi:
## PR Description
Return an empty blob sidecars list instead of an empty Optional when the block is not finalised and doesn't contain kzg commitments
## Fixed Issue(s)

#7635 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #7641, opened 2023-10-30
Fix payloadValueResult in builderGetHeader should be set to won payload

zilm13:
## PR Description
While reviewing #7529 I made a mistake not properly understanding the spec and asking @mehdi-aouadi to always set payload value callback to the local value, while it should be set to the value of actually included payload (which could be both local and builder). Fixing it + verification of all cases in tests.

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #7646, opened 2023-11-02
Expose single blob kzg proof verification method in KZG interface

StefanBratanov:
## PR Description
Single blob kzg proof verification would be required for gossip validation as per https://github.com/ethereum/consensus-specs/pull/3531

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/kzg/KzgVerifyBlobProofTestExecutor.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/kzg/KzgVerifyBlobProofTestExecutor.java
@@ -16,17 +16,12 @@ package tech.pegasys.teku.reference.phase0.kzg;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import java.util.List;
 import org.apache.tuweni.bytes.Bytes;
 import tech.pegasys.teku.ethtests.finder.TestDefinition;
 import tech.pegasys.teku.kzg.KZG;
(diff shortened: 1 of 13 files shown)


tbenr:
can you use this in `MiscHelpersDeneb::verifyBlobSidecar` already?

tbenr:
ah but that's gone now.. was on a old branch

zilm13 (approved):
LGTM

StefanBratanov:
Actually I did bit of refactoring, but not sure about it. In my mind is cleaner, having `verifyBlobKzgProof` and `verifyBlobKzgProofBatch` both in MiscHelpers. Also created an abstract class for the validating proxy to be used for byRoot PR. Let me know @zilm13 if you like it or I am happy to revert my last two commits as well.

zilm13:
Merge it, i'm ok
abstract class will not fit in my code, but let's resolve it in another place

StefanBratanov:
> Merge it, i'm ok abstract class will not fit in my code, but let's resolve it in another place

KK. Should be ok though can just use extends `AbstractBlobSidecarsValidatingProxy` and that exposes the kzg verificiation method

zilm13:
I need to use it without instance also
I will think out how to do this after could merge your PR https://github.com/Consensys/teku/blob/979c6cfa45651ca563b4ba78bac5cc3f46626521/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/DefaultEth2Peer.java#L278-L299

zilm13:
actually i think i'll make an instance for this case too to keep logging

StefanBratanov:
> actually i think i'll make an instance for this case too to keep logging

Cool. Let me know when you finish the PR. I will have a look.

PR #7648, opened 2023-11-02
RPC blob sidecars validating refactor

StefanBratanov:
## PR Description
Just a small refactor.

- Make `BlobSidecarsByRootListenerValidatingProxy` extend `BlobSidecarsByRootValidator` instead of new object created within it.
- Some renamings
- `verifyKzg` inside `AbstractBlobSidecarsValidator` throws the exceptions (so void instead of boolean)
- Add missed fuzzing test

## Fixed Issue(s)
N/A

## Documentation

## Changelog


zilm13 (approved):
LGTM but I'd keep Set in the end
`contains` for list will be 10-100 times longer, you don't have hash index at all

StefanBratanov:
> LGTM but I'd keep Set in the end `contains` for list will be 10-100 times longer, you don't have hash index at all

Yeah. I thought about that. Agree. Will fix in the morning.

PR #7649, opened 2023-11-02
Sort endpoints and schemas in swagger UI

zilm13:
## PR Description
Noticed that it looks weird and sort both endpoints and schemas. 
We cannot have any tag order but I did my best to make it more or less with some idea.
Schemas are sorted also, looks especially better.
Sorry that code looks ugly, it's frontend.

## Documentation

## Changelog

### DIFF
--- a/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/openapi/OpenApiDocBuilder.java
+++ b/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/openapi/OpenApiDocBuilder.java
@@ -14,7 +14,17 @@
 package tech.pegasys.teku.infrastructure.restapi.openapi;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static java.util.stream.Collectors.toSet;
+import static tech.pegasys.teku.infrastructure.http.RestApiConstants.TAG_BEACON;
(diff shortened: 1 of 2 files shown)


infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/openapi/OpenApiDocBuilder.java
```diff
@@ -126,6 +143,51 @@ private void writePaths(final JsonGenerator gen) throws IOException {
...
+            TAG_VALIDATOR,
+            TAG_BUILDER,
+            TAG_REWARDS,
+            TAG_EVENTS,
+            TAG_CONFIG,
+            TAG_NODE,
+            TAG_TEKU,
+            TAG_DEBUG,
+            TAG_EXPERIMENTAL);
+    IntStream.range(0, tags.size()).forEach(i -> tagOrder.put(tags.get(i), i));
```
lucassaldanha: It would be better to have this list coming from RestApiConstants (where tags are defined). Otherwise we'll always forget to update this when we add new tags.

zilm13: Makes sense. Done.

tbenr (approved):
LGTM

PR #7653, opened 2023-11-03
only check folder exists on startup

rolfyone:
also cache the path of the slashing data file.

## Documentation

## Changelog


courtneyeh (approved):
LGTM

PR #7656, opened 2023-11-03
Fix minor issue in BlocksByRange slot check

jtraglia:
## PR Description

Really minor fix here, but BlocksByRange shouldn't allow the `startSlot + count` slot.

> Requests beacon blocks in the slot range `[start_slot, start_slot + count)`, leading up to the current head block as selected by fork choice. For example, requesting blocks starting at `start_slot=2` and `count=4` would return the blocks at slots `[2, 3, 4, 5]`.

https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beaconblocksbyrange

Also, remove old comments about end slot being 9. I think it's from when `step` was 2.

## Documentation

## Changelog


zilm13 (approved):
LGTM

zilm13:
Thank you, nice find!
I've added also minor cleanup after it #7657

PR #7659, opened 2023-11-03
Measure actual attestation production time

rolfyone:
Currently the measurement of attestation creation is actually the time taken in the channel queue. This will improve that by separately measuring the actual time to fetch the required data and just create the actual attestation.

Also added time to send attestations.

## Documentation

## Changelog


tbenr:
Seems like ok to me, not sure about `create_total` step name, but I don't have any suggestion. I leave to Lucas\Courtney for the 👍

infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/Validator/ValidatorDutyMetricsSteps.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package tech.pegasys.teku.infrastructure.metrics.Validator;
+
+public enum ValidatorDutyMetricsSteps {
+  TOTAL("total"),
+  CREATE_TOTAL("create_total"),
+  CREATE("create"),
+  SIGN("sign"),
+  SEND("send");
```
lucassaldanha: I wonder if it would be valuable to add a brief description of CREATE_TOTAL to help others understand its meaning?

lucassaldanha (approved):
LGTM

infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/Validator/ValidatorDutyMetricsSteps.java
rolfyone: in code? it might help devs i guess but does nothing to help a user... if you look at where its used its when calling validatorApiChannel... seems fairly obvious?

PR #7662, opened 2023-11-06
Added checkpoint-sync-url CLI option

### LINKED ISSUE #7664: Add support for --checkpoint-sync-url option

lucassaldanha:
This PR is introducing the `--checkpoint-sync-url` CLI option. Teku already supports checkpoint sync using `--initial-state` option, but we wanted to be more explicit on the fact that we support checkpoint sync.

The behaviour of using `checkpoint-sync-url` is similar to using `--initial-state` ([docs](https://docs.teku.consensys.net/reference/cli#initial-state)) and `--genesis-state`([docs](https://docs.teku.consensys.net/reference/cli#genesis-state)) together. First, Teku will try to load the latest finalized checkpoint from the beacon node. If it fails, Teku will then try to load the genesis from the beacon node.

Starting a sync from genesis is something we want to discourage and we are exploring the option of not doing it by default (it would require an override flag to allow this behaviour).

fixes #7664


infrastructure/http/src/test/java/tech/pegasys/teku/infrastructure/http/UrlSanitizerTest.java
```diff
@@ -69,4 +74,31 @@ void shouldNotDetectPathInStringThatIsNonUrl() {
...
+  }
+
+  @Test
+  public void AppendPathShouldFailWithEmptyUrl() {
+    assertThatThrownBy(() -> UrlSanitizer.appendPath("", "foo"))
+        .isInstanceOf(IllegalArgumentException.class);
+  }
+
+  @ParameterizedTest
+  @MethodSource("appendDataParams")
```
rolfyone, line 91: nice

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
```diff
@@ -58,6 +58,15 @@ public class Eth2NetworkOptions {
       arity = "1")
   private String genesisState;
 
+  @Option(
+      names = {"--checkpoint-sync-url"},
+      hidden = true,
+      paramLabel = "<STRING>",
+      description =
+          "The Beacon API endpoint of another synced beacon node that will be used to bootstrap this node's state.",
```
rolfyone: i'd probably go as far as saying `The Checkpointz server that will be used to bootstrap this node.`

rolfyone: https://github.com/ethpandaops/checkpointz
spelling probably important (just putting that out there :))

rolfyone (approved):
just a nit, LGTM overall

rolfyone:
oh - add changelog :)

Issue #7654, opened 2023-11-03
Move to sidecar inclusion proof


tbenr:
Issue to track, high level, the implementation status for https://github.com/ethereum/consensus-specs/pull/3531 and related spec changes

- [x] Migrate data structures
- [x] Add configuration params
- [x] Add inclusion proof creation and validation
- [x] Update gossip rules
- [x] Update RPC methods (by root \ by range)
- [x] Update block import (in-sync, forward sync, historical sync)
- [x] Update storage 
- [x] Update block production flow
- [x] Update builder block production flow https://github.com/ethereum/builder-specs/pull/90
- [x] Update block production API (remote VC) https://github.com/Consensys/teku/issues/7651
- [x] Update beacon API (get blobs)
- [x] Check that we always validate number of commitments/blobs (shouldn't exceed maxBlobs) when it comes from external source
- [x] Optimize `BlobSidecarGossipValidator` by checking `blockHeader` signature (and maybe some other fields if possible) only once for BlobSidecars from one block
- [x] Re-enable reference tests (disabled here: #7695)
- [x] Cleanup (no more traces of previous implementation)

PR #7674, opened 2023-11-07
Remove notion of `BlindedBlobsBundle`

StefanBratanov:
## PR Description

- Remove `BlindedBlobsBundle` from BuilderBid
- Pass `blobKzgCommitments` to `HeaderWithFallbackData`
- Create `BlobKzgCommitmentsSchema` to pass in the `SchemaDefinitionsDeneb` for blocks and builder bids and simplify creation from `BlobsBundle`
- Rename `randomSszKzgCommitmentsList` to `randomBlobKzgCommitments` in `DataStructureUtil`
- Removed `BlindedBlobsBundle` all around the project.
- Set `blobKzgCommitments` from the `BuilderBid` instead of the `BlindedBlobsBundle` during block production
- Changed tests accordingly
 
## Fixed Issue(s)
helps with builder block production flow for https://github.com/Consensys/teku/issues/7654

## Documentation

## Changelog


zilm13 (approved):
LGTM
But I see we have already several places where we receive but don't validate structures with kzgCommitments/blobs that they shouldn't exceed maxBlobsPerBlock. This should be addressed somewhere but it's ok if it's a separate PR

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryDenebTest.java
```diff
@@ -72,25 +71,17 @@ void shouldCreateBlockContents() {
...
-        .hasValueSatisfying(
-            blindedBlobSidecars ->
-                assertThat(blindedBlobSidecars)
-                    .hasSize(3)
-                    .map(BlindedBlobSidecar::getBlobRoot)
-                    .hasSameElementsAs(
-                        blindedBlobsBundle.getBlobRoots().stream()
-                            .map(SszBytes32::get)
-                            .collect(Collectors.toList())));
+        .hasValueSatisfying(blindedBlobSidecars -> assertThat(blindedBlobSidecars).isEmpty());
```
zilm13: I'd rather keep it broken and disabled to not forget we need to make it work than leave with incorrect check

StefanBratanov: good point

PR #7676, opened 2023-11-07
7390 block v3 refactoring

mehdi-aouadi:
## PR Description
Some block v3 related refactoring:
- Handle pre CAPELLA missing execution payload value: log warnings and set value to 0 instead of failing block production
- Move the `--Xblock-v3-enabled` option from `ValidatorClientOptions` to the `ValidatorOptions` in order to be able to use it for single process nodes

## Fixed Issue(s)

#7390 
## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #7677, opened 2023-11-07
Evict in natural order from last epoch Store caches for blocks and blobs

### LINKED ISSUE #7671: Avoid block cache drops when serving high number of RPC requests
image

RPC request can cause drops in the block cache.

zilm13:
## PR Description
These caches were designed to store the latest epoch of blocks but were found to have some weird behavior. The issue is that they were using LRU evicted map underneath, so being accessed by RPC the order was changed from the natural it was inserted and actually could hold older values while evicting the most recent blocks.

## Fixed Issue(s)

Fixes #7671 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #7683, opened 2023-11-09
Enable VC crashing on 0 keys loaded with CLI argument

### LINKED ISSUE #7647: Enable VC crashing on 0 keys loaded
Regarding PR [5145](https://github.com/Consensys/teku/pull/5145) validators are allow to run with 0 loaded keys,
Could you add an CLI argument to enable the previous behaviour of crashing / restarting when 0 keys loaded?
Thanks

Logs:
```
023-11-02 12:46:03.668 WARN  - Failed to load public keys from URL http://blabla
```

courtneyeh:
fixes #7647 

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
```diff
@@ -104,12 +104,23 @@ public class ValidatorKeysOptions {
   private int validatorExternalSignerConcurrentRequestLimit =
       ValidatorConfig.DEFAULT_VALIDATOR_EXTERNAL_SIGNER_CONCURRENT_REQUEST_LIMIT;
 
+  @CommandLine.Option(
+      names = {"--allow-no-loaded-keys-enabled"},
```
lucassaldanha: Allow and Enabhled on the same parameter makes it a bit confusing. Can it be something like: `--allow-no-loaded-keys`?

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorKeysOptions.java
```diff
@@ -122,8 +133,11 @@ public void configure(TekuConfiguration.Builder builder) {
                     convertToPath(validatorExternalSignerTruststorePasswordFile)));
   }
 
-  private List<String> parseValidatorExternalKeys() {
-    return validatorExternalSignerPublicKeys.stream().filter(StringUtils::isNotBlank).toList();
+  private List<String> parseValidatorExternalKeys(final boolean allowNoLoadedKeysEnabled) {
+    if (allowNoLoadedKeysEnabled) {
+      return validatorExternalSignerPublicKeys.stream().filter(StringUtils::isNotBlank).toList();
+    }
+    return validatorExternalSignerPublicKeys;
```
lucassaldanha: At this point, we could already check if the list is empty and throw an error when `--allow-no-loaded-keys-enabled=false` and `validatorExternalSignerPublicKeys` is empty.

PR #7685, opened 2023-11-09
Update Besu version in acceptance tests

StefanBratanov:
## PR Description

- upgrade Besu docker `STABLE` version for acceptance tests and besu libraries
- enable one very old tuweni test which has been fixed

## Fixed Issue(s)
N/A

## Documentation

## Changelog


mehdi-aouadi:
The acceptance tests are failing because Besu is failing to start because of the config option `tx-pool-limit-by-account-percentage`. The logs show the following error:
```
| ERROR | Besu | Failed to start Besu
| DEBUG | BesuNode | picocli.CommandLine$ParameterException: Could not use legacy transaction pool options with layered implementation [--tx-pool-limit-by-account-percentage]
| DEBUG | BesuNode | Could not use legacy transaction pool options with layered implementation 
```
Removing it from here seems to solve the issue (at least the doppelganger acceptance test passed):
https://github.com/StefanBratanov/teku/blob/13a8fa3367cac72efc652618cbba8127f63d12c9/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/BesuNode.java#L233

I wasn't able to push a commit directly to your branch, don't have permission :D

StefanBratanov:
> The acceptance tests are failing because Besu is failing to start because of the config option `tx-pool-limit-by-account-percentage`. The logs show the following error:
> 
> ```
> | ERROR | Besu | Failed to start Besu
> | DEBUG | BesuNode | picocli.CommandLine$ParameterException: Could not use legacy transaction pool options with layered implementation [--tx-pool-limit-by-account-percentage]
> | DEBUG | BesuNode | Could not use legacy transaction pool options with layered implementation 
> ```
> 
> Removing it from here seems to solve the issue (at least the doppelganger acceptance test passed): https://github.com/StefanBratanov/teku/blob/13a8fa3367cac72efc652618cbba8127f63d12c9/acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/BesuNode.java#L233
> 
> I wasn't able to push a commit directly to your branch, don't have permission :D

Oh thanks a lot for looking. I didn't find time today to debug it myself. Will fix in the weekend based on the commit. Thanks!

zilm13:
It will be flaky with `shanghaiTime`, that's why I've disabled it until `shanghaiBlock` is available. It was working already but I expect that it could be very flaky on CI with `shanghaiTime` while performing well locally. I've checked most recent Besu and `shanghaiBlock` option is not yet added, so I think we will add a flaky test with this PR.
As an option we can merge it but disable once it will start to show up in failed.

StefanBratanov:
> Reference

Ahh, I didn't read the `@Disabled` comment properly. My bad. Will update this PR (disable test again) and only leave the Besu version update changes.

zilm13:
Actually I'm not 100% sure it will fail often, it's my feeling, but time could drift on CI in very broad boundaries and this test could fail if it's too late or too early. Block is definitely more guaranteed. As `shanghaiBlock`was not yet added we have not much chance we could see it in the future, so we are left without CapellaAcceptanceTest, which is not good.

zilm13:
I've tested it locally again. I don't think we could not guarantee anything with time-based solution

zilm13 (approved):
LGTM

PR #7686, opened 2023-11-09
Fix API start up log

mehdi-aouadi:
## PR Description
`app.jettyServer().getServerHost()` is always returning null (in the reported issue it happened when `--validator-api-interface=0.0.0.0`) which makes the logger printing `Listening on http://localhost:5052/` instead of `Listening on https://0.0.0.0:5052/`

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #7644, opened 2023-11-01
Add broadcast validation to API handler

tbenr:
To fulfill the requirement that API have to return `200` when block has been fully imported, I had to let the `BlockImportChannel::importBlock` to return a record of both importResult future as well as broadcast validation future.

`AbstractBlockPublisher::gossipAndImportUnblindedSignedBlock` now handles the two different flows by relaying on two separated `importBlock` and `publishBlock` methods implemented by the concrete classes.

continuation of #7371
related to #7349

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [ ] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #7689, opened 2023-11-09
Broadcast validator revamped

tbenr:
Redesign the broadcast validation, with the main improvement to fail block import if broadcast validation fails (the case in which GOSSIP validation fails and block appears to be good, but we don't want that block to be published as well as avoid import to complete and becoming the head)

continuation of #7644
related to #7349

## Documentation

## Changelog


tbenr:
Tested this in interop running 2 nodes with the same key, adding an artificial delay in block publishing in one of them (so the other publish systematically first), end enabled broadcast validation.

image

I managed also to get a `FINAL_EQUIVOCATION_FAILURE` by removing the delay.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java
```diff
@@ -0,0 +1,144 @@
...
+  private final BroadcastValidationLevel broadcastValidationLevel;
+  private final SafeFuture<Boolean> consensusValidationSuccessResult;
+  private final SafeFuture<BroadcastValidationResult> broadcastValidationResult;
+
+  private BlockBroadcastValidatorImpl(
+      final BlockGossipValidator blockGossipValidator,
+      final BroadcastValidationLevel broadcastValidationLevel) {
+    this.blockGossipValidator = blockGossipValidator;
+    this.broadcastValidationLevel = broadcastValidationLevel;
+    consensusValidationSuccessResult = new SafeFuture<>();
```
zilm13: `this.`please

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    broadcastValidationResult = new SafeFuture<>();
+  }
+
+  public static BlockBroadcastValidatorImpl create(
+      final SignedBeaconBlock block,
+      final BlockGossipValidator blockGossipValidator,
+      final BroadcastValidationLevel broadcastValidationLevel) {
+    BlockBroadcastValidatorImpl blockBroadcastValidator =
```
zilm13: final please

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  @Override
+  public SafeFuture<BroadcastValidationResult> getResult() {
+    return broadcastValidationResult;
+  }
+
+  private void buildValidationPipeline(final SignedBeaconBlock block) {
+    // validateBroadcast should not be called at all but let's cover the case for safety
```
zilm13, line 82: let's add return so

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java
```diff
@@ -0,0 +1,144 @@
...
+            .thenApply(
+                gossipValidationResult -> {
+                  if (gossipValidationResult.isAccept()) {
+                    return SUCCESS;
+                  }
+                  return GOSSIP_FAILURE;
+                });
+
+    if (broadcastValidationLevel == GOSSIP) {
+      validationPipeline.propagateTo(broadcastValidationResult);
```
zilm13, line 102: Maybe I'm too paranoid but I'd also cancel or complete `consensusValidationSuccessResult` here and in NOT_REQUIRED case. Not sure though it will actually leaking

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/AbstractBlockPublisher.java
```diff
@@ -59,41 +58,63 @@ public SafeFuture<SendSignedBlockResult> sendSignedBlock(
...
+        .thenCompose(result -> calculateResult(maybeBlindedBlockContainer, result));
+  }
+
+  private SafeFuture<SendSignedBlockResult> calculateResult(
+      final SignedBlockContainer maybeBlindedBlockContainer,
+      final BlockImportAndBroadcastValidationResults blockImportAndBroadcastValidationResults) {
 
-                return SendSignedBlockResult.notImported(result.getFailureReason().name());
+    return blockImportAndBroadcastValidationResults
+        .broadcastValidationResult()
```
zilm13, line 73: We have FAILED_BROADCAST_VALIDATION now, what's purpose of this block. Could it be fired earlier? If not, I'd better organize this code more streamlined by checking only blockImportResult

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTest.java
```diff
@@ -250,7 +255,25 @@ void onBlock_consensusValidationShouldNotResolveWhenDataAvailabilityFails() {
...
+  @Test
+  void onBlock_shouldFailWhenBroadcastValidationFails() {
+    setupWithSpec(TestSpecFactory.createMinimalDeneb());
+    final SignedBlockAndState blockAndState = chainBuilder.generateBlockAtSlot(ONE);
+    storageSystem.chainUpdater().advanceCurrentSlotToAtLeast(blockAndState.getSlot());
+
+    when(blockBroadcastValidator.getResult())
+        .thenReturn(SafeFuture.completedFuture(BroadcastValidationResult.GOSSIP_FAILURE));
+
+    importBlockWithError(blockAndState, FailureReason.FAILED_BROADCAST_VALIDATION);
```
zilm13: Maybe importBlockWithExpectedError? current name always confusing me. Sounds like it sets up some when instead of asserting

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -332,7 +323,17 @@ private SafeFuture<BlockImportResult> handleBlockImport(
                     LOG.warn("Unable to import block {} due to invalid data", block.toLogString());
                     blobSidecarPool.removeAllForBlock(block.getRoot());
                     break;
-                  default:
+                  case FAILED_BROADCAST_VALIDATION:
+                    LOG.warn(
+                        "Unable to import block {} due to failed broadcast validation",
+                        block.toLogString());
+                    break;
```
zilm13, line 330: I have concerns that we could have some unresolved tails in this case. For example gossip validation failed because now parent, which means it's a future block and previously you would add it to the future blocks but with broadcast validation not. Could it be an issue (and may be other similar cases)?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java
tbenr: ahh good catch

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java
tbenr: I think it isn't necessary, an incomplete safe future won't leak. I can `cancel` for clarity though..

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
tbenr: I think the key thing here is that we are talking about a locally produced block, so timing issues shouldn't apply. 
My initial idea was mainly avoid it to be put in the invalid cache, because the block it isn't actually invalid.

An interesting situation could be:

VC calls block publishing API via 2 different BN in parallel.
the first BN publishing it can cause the other to fail the GOSSIP validation rule. So in this case we really want to not consider the block invalid otherwise because the block has been actually sent. We want the node to be able to recover.

Can't think of other cases ATM

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/AbstractBlockPublisher.java
tbenr: I was having doubts here too.
I decided to leave it here (even if we could leave the previous implementation and only deal with import result) because it can fire earlier, yes. We can have a failure here while state transition is still in progress.

What I don't like is that here we assume that block import will fail too and just return immediately here.. which might be a fair assumption..

Another option would be to "check" in the forkChoice before doing the state transition to anticipate the failure in case it is already resolved, and leave here dealing with block import result only.. I don't have a strong opinion ATM.

zilm13: if it could fire earlier, I'm good with it, may worth a comment

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
zilm13: Got you. Makes sense.

zilm13 (approved):
LGTM

Issue #7651, opened 2023-11-03
beacon-api - remove blob signing


rolfyone:
https://github.com/ethereum/beacon-APIs/pull/369

related to #7654

PR #7696, opened 2023-11-13
add kzg proofs and blobs to BlockContents and SignedblockContents

mehdi-aouadi:
## PR Description
Update BlockContents and SignedBlockContents:
- Remove `BlobSidecars` and `SignedBlobSidecars`
- Add `KzgProofs` and `Blobs`

## Fixed Issue(s)

#7651 
#7654 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/versions/deneb/SignedBlockContents.java
```diff
@@ -46,7 +52,12 @@ public SignedBeaconBlock getSignedBlock() {
   }
 
   @Override
-  public Optional<List<SignedBlobSidecarOld>> getSignedBlobSidecars() {
-    return Optional.of(getField1().asList());
+  public Optional<List<KZGProof>> getKzgProofs() {
```
StefanBratanov: I think it would be better if we have `Optional<SszList<SszKZGProof>>` and `Optional<SszList<Blob>>` similar to how we return the `blobKzgCommitments` in `BeaconBlockBodyDeneb`. It would avoid wrapping and unwrapping when using the values since most of the time we just need to loop and get the values. Same for the return types in `BlockContents`.

StefanBratanov (approved):
Looks good. Just have one comment which I think will avoid some performance overhead given the potential usage of the getters in multiple places.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/versions/deneb/SignedBlockContents.java
mehdi-aouadi: Done

PR #7702, opened 2023-11-14
remove json files usage

### LINKED ISSUE #7619: Render objects to json instead of using json files for integration tests
Some integration tests use json files for assertions. This could break if we modify the `DataStructureUtil`
Examples: the APIs integration tests.
A better way would be to render json from the generated objects and use that as input for the assertions.

mehdi-aouadi:
## PR Description
Remove json files usage in beacon API integration tests and render object to json instead

## Fixed Issue(s)

fixes #7619 

## Documentation

## Changelog


tbenr:
I'm not a super fan of this.
If we have a json serializer that renders everything to `[]` the tests will pass (and all the bugs in between).
So if the tests are supposed to be end-to-end, without trusting the serialized, this is not good.
If we think that the serializer is well-tested elsewhere I'm ok with it. WDYT @zilm13 ?

mehdi-aouadi:
> I'm not a super fan of this. If we have a json serializer that renders everything to `[]` the tests will pass (and all the bugs in between). So if the tests are supposed to be end-to-end, without trusting the serialized, this is not good. If we think that the serializer is well-tested elsewhere I'm ok with it. WDYT @zilm13 ?

Actually I think it would be better to move those json checks to the unit tests instead of the integration tests and keep the serialisation/deserialisation testing there (Which is still the case for other APIs btw)

zilm13:
@tbenr good point
In this case my suggestion is to organize tests in this manner, it will test both things. Actual field names and structures are already covered by `OpenApiIntegrationTest`.
```
--- a/data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/validator/GetNewBlockIntegrationTest.java
+++ b/data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/validator/GetNewBlockIntegrationTest.java
@@ -83,8 +83,9 @@ public class GetNewBlockIntegrationTest extends AbstractDataBackedRestAPIIntegra
         isBlindedBlock
             ? spec.atSlot(UInt64.ONE).getSchemaDefinitions().getBlindedBeaconBlockSchema()
             : spec.atSlot(UInt64.ONE).getSchemaDefinitions().getBeaconBlockSchema();
-    assertThat(node.get("data").toString())
-        .isEqualTo(JsonUtil.serialize(randomBlock, beaconBlockSchema.getJsonTypeDefinition()));
+    final BeaconBlock actualBlock =
+        JsonUtil.parse(node.get("data").toString(), beaconBlockSchema.getJsonTypeDefinition());
+    assertThat(actualBlock).isEqualTo(randomBlock);
   }
```

rolfyone:
> I'm not a super fan of this.
> If we have a json serializer that renders everything to `[]` the tests will pass (and all the bugs in between).
> So if the tests are supposed to be end-to-end, without trusting the serialized, this is not good.
> If we think that the serializer is well-tested elsewhere I'm ok with it. WDYT @zilm13 ?

`OpenApiIntegrationTest` came about because we kept breaking structure, so I'm going to be strongly against anything that check, because if we can't check the structure returned it's far worse than having to maintain these json files, as we break compatibility and don't know it...

zilm13:
btw we have `JsonProviderPropertyTest` but it doesn't cover all it could

rolfyone:
> btw we have `JsonProviderPropertyTest` but it doesn't cover all it could

The main point being we need to define our interface and keep it consistent, and OpenApiIntegrationTest answers that question. If we have 2 tests showing it, i'd suggest OpenApiIntegrationTest is the one we should be keeping given it does all the interfaces.

rolfyone:
> In this case my suggestion is to organize tests in this manner, it will test both things. Actual field names and structures are already covered by `OpenApiIntegrationTest`.

I agree this is a better suggestion in that it is basically a round-trip check.

The thing it misses is if we care about any of the data. If it's simply that we're testing round trip, then we should probably name the test appropriately (roundtrip meaning we can go object -> json -> object successfully and dont care about the content of the block)

Another option is check that, then check specific attributes we specifically care about... both of these would potentially be useful, depending on what we need.

The original json was just allowing a user a manual chance to eyeball the change and say it's sane, and use that as the basis for future tests.

An example may be that once we roundtrip, we would expect to see specific values (or lack of), or a specific type of object in the result... This makes a roundtrip test more specific and more useful, if theres a comparison that can give us a better quality test...

Issue #7670, opened 2023-11-07
[Beacon-api] add POST validators endpoints


rolfyone:
validators and validator_balances endpoints have always been a source of contention.

The new endpoints add POST functionality, and read the list from the HTTP body rather than from an ID field, which will simplify usage for many people.

ref. https://github.com/ethereum/beacon-APIs/pull/367

PR #7706, opened 2023-11-15
Add POST state validators endpoint

courtneyeh:
#7670 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
```diff
@@ -0,0 +1,148 @@
...
+
+  private static final SerializableTypeDefinition<ObjectAndMetaData<List<StateValidatorData>>>
+      RESPONSE_TYPE =
+          SerializableTypeDefinition.<ObjectAndMetaData<List<StateValidatorData>>>object()
+              .name("GetStateValidatorsResponse")
+              .withField(
+                  EXECUTION_OPTIMISTIC, BOOLEAN_TYPE, ObjectAndMetaData::isExecutionOptimistic)
+              .withField(FINALIZED, BOOLEAN_TYPE, ObjectAndMetaData::isFinalized)
+              .withField("data", listOf(STATE_VALIDATOR_DATA_TYPE), ObjectAndMetaData::getData)
+              .build();
```
rolfyone: i think the response type should be identical to the Get, so in theory we can share the object...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
```diff
@@ -0,0 +1,148 @@
...
+
+  private static final DeserializableTypeDefinition<RequestBody> REQUEST_TYPE =
+      DeserializableTypeDefinition.object(RequestBody.class)
+          .name("PostStateValidatorsRequestBody")
+          .initializer(RequestBody::new)
+          .withField(
```
rolfyone: can we make this withOptionalField? the old implementation allowed fetch with no ids, so we should replicate that...

Statuses should also be an optional field according to the beacon-apis spec...

https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/postStateValidators

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
```diff
@@ -0,0 +1,148 @@
...
+            .build());
+    this.chainDataProvider = provider;
+  }
+
+  @Override
+  public void handleRequest(RestApiRequest request) throws JsonProcessingException {
+    final RequestBody requestBody = request.getRequestBody();
+
+    final List<String> validators = requestBody.getIds();
+    final List<StatusParameter> statusParameters = requestBody.getStatuses();
```
rolfyone: after the typedef changes for optional, we'll need to flow that through here too...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/StatusParameter.java
```diff
@@ -0,0 +1,48 @@
...
+
+package tech.pegasys.teku.beaconrestapi.handlers.v1.beacon;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import tech.pegasys.teku.api.response.v1.beacon.ValidatorStatus;
+
+@SuppressWarnings("JavaCase")
```
rolfyone: we can probably fix this by having a named enum rather than doing it this way
so the field would be `pendingInitialized` but the 'name' is ("pending_initialized")
I believe we have more enums like this, definitely more that have a `getName` function...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
rolfyone: we'd want test cases showing empty working too, nowhere really to comment on that currently so will put it here :)

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public List<StatusParameter> getStatuses() {
+      return statuses;
+    }
+
+    public Optional<List<String>> getMaybeStringStatuses() {
+      return statuses.isEmpty()
+          ? Optional.empty()
+          : Optional.of(statuses.stream().map(Enum::name).collect(Collectors.toList()));
```
rolfyone: i believe we can just call `toList()` now outside of the collector, which is more efficient...

rolfyone (approved):
LGTM, just a nit.

PR #7711, opened 2023-11-15
Avoid creating tracker for known blocks

tbenr:
Avoids creating trackers that could remain for several slots in memory.
It applies especially to forward sync.  Noticed that we could import a batch a second time after it has been already imported. This happens when we are close to the head (so it is the last batch or so).

I also noticed that we can have a concurrency issue when tracing blockBlobSidecarsTrackers content, added a synchronization.

related to #7712 

## Documentation

## Changelog


mehdi-aouadi:
It's worth adding a test IMO (like the one for `onNewBlobSidecar`)

tbenr:
@mehdi-aouadi yea I got lazy. Added.

mehdi-aouadi (approved):
LGTM

PR #7716, opened 2023-11-16
teku-7679 rewards attestation api improvement

### LINKED ISSUE #7679: rewards attestation api improvement
our rewards attestations api requires the empty set to be defined if we want all validator data returned.

It'd be nice if we didnt require this, and just assumed the body is asking for all if its empty....

basically
```
		curl -s -X POST \
			-d "[]" \
     			-H 'accept: application/json' \
     			-H 'Content-Type: application/json' \
     			http://localhost:5051/eth/v1/beacon/rewards/attestations/$i  \
     			-o $i.json
```

the -d should be handled if we don't specify it.

I also think the headers should default to application/json if not set, currently the command seems very unhappy if we don't specify accept and content-type.

currently:
```
curl -s -d '[]' -X POST http://localhost:5051/eth/v1/beacon/rewards/attestations/9080 -o 9080.json
```
responds:
```
{"code":415,"message":"Request content type application/x-www-form-urlencoded is not supported. Must be one of: [application/json]"}
```
this may be a curl thing, i'm not sure.

lucassaldanha:
## PR Description
- Added support for reading an empty request body from the request.
- Updated GetAttestationRewards method to accept an empty request body and treat is as an empty list (return rewards for all validators)

## Fixed Issue(s)
fixes #7679 

## Documentation

## Changelog


courtneyeh (approved):
LGTM

PR #7723, opened 2023-11-16
add Deneb block contents signing

mehdi-aouadi:
## PR Description
Add blobs and KZG proofs to BlockContainer when signing it

## Fixed Issue(s)

#7654 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/BlockContainerSignerDeneb.java
```diff
@@ -44,15 +49,21 @@ public SafeFuture<SignedBlockContainer> sign(
     return signBlock(unsignedBlock, validator, forkInfo)
         .thenApply(
             signedBlock -> {
-              // Blinded flow
```
StefanBratanov, line 47: The flow for creating a block for Deneb is a bit unintuitive, so in case for blinded, we have to return just the `SignedBeaconBlock` as it is after signing it. For non blinded flow, we create the block contents and they should always have proofs and blobs, so I think it's better to be more harsh and use `orElseThrow()` and throw an exception.

StefanBratanov (changes requested):

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/BlockContainerSignerDeneb.java
mehdi-aouadi: Alright, I pushed some changes, is that what you meant?

StefanBratanov: Yup. It looks ok.

StefanBratanov (approved):

StefanBratanov:
LGTM. Just can't see a test that covers signing a blinded block returns only the block.

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/BlockContainerSignerDeneb.java
mehdi-aouadi: Cool, I'll add the missing tests then

tbenr (approved):
LGTM.
block production interop in single process works!
the blinded flow too with cherrypicking Mehdi's PR

StefanBratanov:
Looking at it can just add the assertions at `forDeneb_shouldCreateAndPublishBlockContents` , `forDeneb_shouldCreateAndPublishBlindedBlock` and `forDeneb_shouldUseBlockV3ToCreateAndPublishBlockContents` in BlockProductionDutyTest and that should cover it.

mehdi-aouadi:
@StefanBratanov I just pushed the missing tests. Should be good now

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/BlockContainerSignerDeneb.java
```diff
@@ -44,15 +48,37 @@ public SafeFuture<SignedBlockContainer> sign(
...
               if (signedBlock.isBlinded()) {
                 return signedBlock;
+              } else {
+                final List<KZGProof> kzpProofs =
```
StefanBratanov: just typo `kzgProofs`

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/BlockProductionDutyTest.java
```diff
@@ -218,7 +227,37 @@ public void forDeneb_shouldCreateAndPublishBlockContents() {
...
+                  .isEqualTo(kzgProofFromUnsignedBlockContents);
+            });
+
+    assertThat(signedBlockContents.getBlobs()).isPresent();
+
+    final SszList<Blob> blobsFromSignedBlockContents = signedBlockContents.getBlobs().get();
+
+    assertThat(blobsFromSignedBlockContents).isNotEmpty();
+
+    IntStream.range(0, blobsFromSignedBlockContents.size())
```
StefanBratanov: Can just compare the two collections (same for proofs) instead of looping through each

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/BlockProductionDutyTest.java
```diff
@@ -491,7 +618,37 @@ public void forDeneb_shouldUseBlockV3ToCreateAndPublishBlockContents() {
...
```
StefanBratanov: Same here, just collection comparison

StefanBratanov (approved):
LGTM with couple of nits

validator/client/src/main/java/tech/pegasys/teku/validator/client/signer/BlockContainerSignerDeneb.java
mehdi-aouadi: My bad. Fixed

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/BlockProductionDutyTest.java
mehdi-aouadi: Done

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/BlockProductionDutyTest.java
mehdi-aouadi: Done

PR #7730, opened 2023-11-17
Remove bug from getOptionalRequest method

courtneyeh:


infrastructure/restapi/src/testFixtures/java/tech/pegasys/teku/infrastructure/restapi/StubRestApiRequest.java
```diff
@@ -91,7 +91,7 @@ public <T> T getRequestBody() {
   @Override
   @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
   public <T> Optional<T> getOptionalRequestBody() throws JsonProcessingException {
-    return Optional.of((T) requestBody);
+    return Optional.ofNullable((T) requestBody);
```
rolfyone: i think if we use `getRequestBody` in here we'd remove the need for the untyped etc... but i could be wrong...

rolfyone: also, test case would be good :D

rolfyone (approved):
LGTM

PR #7732, opened 2023-11-17
Pass only the block to `getUnblindedPayload` call

StefanBratanov:
## PR Description

- Small cleanup. Pass `SignedBeaconBlock` instead of `SignedBlockContainer` to `ExecutionLayerBlockProductionManager.getUnblindedPayload` (aka builder only needs the block)
- Also made `BuilderBidValidator` be void and no `spec` passed

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidator.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidator.java
@@ -15,19 +15,19 @@ package tech.pegasys.teku.ethereum.executionlayer;
 
 import java.util.Optional;
 import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.datastructures.builder.BuilderBid;
 import tech.pegasys.teku.spec.datastructures.builder.SignedBuilderBid;
 import tech.pegasys.teku.spec.datastructures.builder.SignedValidatorRegistration;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayload;
-import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadHeader;
 import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
 
 @FunctionalInterface
 public interface BuilderBidValidator {
   BuilderBidValidator NOOP =
       (spec, signedBuilderBid, signedValidatorRegistration, state, localExecutionPayload) ->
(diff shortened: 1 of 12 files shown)


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidator.java
```diff
@@ -15,19 +15,19 @@
...
 
 @FunctionalInterface
 public interface BuilderBidValidator {
   BuilderBidValidator NOOP =
       (spec, signedBuilderBid, signedValidatorRegistration, state, localExecutionPayload) ->
-          signedBuilderBid.getMessage().getHeader();
+          signedBuilderBid.getMessage();
 
-  ExecutionPayloadHeader validateAndGetPayloadHeader(
+  BuilderBid validateAndGetBuilderBid(
```
zilm13: Maybe make it void like most other validators?

StefanBratanov: Yup. Good idea. Done. Also change it not to pass spec in method to be more precise.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidatorImpl.java
```diff
@@ -120,21 +122,19 @@ public ExecutionPayloadHeader validateAndGetPayloadHeader(
     final UInt64 proposedGasLimit = executionPayloadHeader.getGasLimit();
 
     if (parentGasLimit.equals(preferredGasLimit) && proposedGasLimit.equals(parentGasLimit)) {
-      return executionPayloadHeader;
+      return;
```
zilm13, line 126: I don't understand checks since this line. It couldn't fail. What's in the spec?
Ahh. I'm getting this. It's just an event log if none of this matches. I'd add some comment on :118

zilm13 (approved):
LGTM, just let's add some comment clarifying gas validation part which couldn't be fatal

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidatorImpl.java
StefanBratanov: Thanks. Added a more clarifying comment.

PR #7736, opened 2023-11-20
Add consistency check for the builder `BlobsBundle`

StefanBratanov:
## PR Description

- Add a consistency check for the builder `BlobsBundle`
- Remove some unused methods in `BlobSidecarManager` and `MiscHelpers`
- Add consistency check in execution `BlobsBundle` constructor

## Fixed Issue(s)
will accomplish a task in #7654 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
@@ -27,6 +27,7 @@ import tech.pegasys.teku.bls.BLSSignature;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.ssz.SszList;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.kzg.KZG;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.Blob;
 import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.BlobSidecar;
@@ -75,6 +76,7 @@ public class BlockOperationSelectorFactory {
   private final Bytes32 graffiti;
   private final ForkChoiceNotifier forkChoiceNotifier;
   private final ExecutionLayerBlockProductionManager executionLayerBlockProductionManager;
+  private final KZG kzg;
 
   public BlockOperationSelectorFactory(
       final Spec spec,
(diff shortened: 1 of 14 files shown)


zilm13 (commented):
I 'm thinking we should check 1) kzgCommitments from block match those from Bundle, 2) kzg.verifyBlobKzgProof. Not sure that it should be done here, but if any of this checks fails, block is rejected from the chain and Proposer will not get any money for it both from builder and proposer beacon chain fee. In the same time if detecting it early we can fallback to the local payload, receive proposer fee and a penny from local payload. It differs from the bad transactions in payload when proposer receives proposing fee, in case with bad BlobSidecar whole block cannot be the part of the chain. 

Also it's interesting if we get bad commitments from local execution in non-builder workflow, could we drop it and publish block without it, it should make sense too.

Or maybe we should change spec so bad BlobSidecars will not make whole block bad.

Sorry, if I understand it wrong. I'm not 100% sure on this

UPDATE: allowing to publish bad BlobSidecar and pass isDataAvailable with it is definitely not a way to go, so in case with local payload we may want to drop commitments and BlobSidecars at all (so commitment transactions will fail on execution) but continue to publish the block and I think this behavior could be the part of the spec.

StefanBratanov:
> I 'm thinking we should check 1) kzgCommitments from block match those from Bundle, 2) kzg.verifyBlobKzgProof. Not sure that it should be done here, but if any of this checks fails, block is rejected from the chain and Proposer will not get any money for it both from builder and proposer beacon chain fee. In the same time if detecting it early we can fallback to the local payload, receive proposer fee and a penny from local payload. It differs from the bad transactions in payload when proposer receives proposing fee, in case with bad BlobSidecar whole block cannot be the part of the chain.
> 
> Also it's interesting if we get bad commitments from local execution in non-builder workflow, could we drop it and publish block without it, it should make sense too.
> 
> Or maybe we should change spec so bad BlobSidecars will not make whole block bad.
> 
> Sorry, if I understand it wrong. I'm not 100% sure on this
> 
> UPDATE: allowing to publish bad BlobSidecar and pass isDataAvailable with it is definitely not a way to go, so in case with local payload we may want to drop commitments and BlobSidecars at all (so commitment transactions will fail on execution) but continue to publish the block and I think this behavior could be the part of the spec.

Thanks for the comments. I agree we should check the commitments in the block matches the hash tree root of the commitments in the bundle. If any of the checks fail, we will unfortunately fail blob sidecars production and miss the block because we have already signed it and sent it to the builder. I think adding also the kzg checks also makes sense after the blob sidecars have been created. But in my opinion, we should just fail the whole flow. In case of builder flow, it's already too late anyways and in non-blinded flow reverting and creating a new block without commitments seems unintuitive.

tbenr:
So to recap what we discussed (hopefully complete)

1- the builder's bundle check isn't a game changer because at that stage we can't fallback to local EL (we already signed). If we do the check we introduce a bit of latency but we save the network to receive bad block\blobs. I'm not a super fan of adding additional latency TBH, even if we are talking about few milliseconds. We could drop kxg verification and leave commitments match.

2- checking local provided bundle it might be worthy because it opens the door of excluding local EL from the flow and go for the builder no matter the value is. This is a weird case from trust assumptions perspective: the most trusted local EL is broken and you relay on less trusted remote builder. We could not consider this local exclusion and eventually publish a bad block locally produced (if wins the builder battle), to be fair the flow is already complicated enough.


3- CL deciding to remove commitments at runtime is not feasible and would end up publishing a bad block.

zilm13 (approved):
LGTM

PR #7739, opened 2023-11-21
Handle invalid Json bad request body in PostStateValidators

courtneyeh:
#7670 

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
@@ -15,11 +15,14 @@ package tech.pegasys.teku.beaconrestapi.handlers.v1.beacon;
 
 import static tech.pegasys.teku.beaconrestapi.BeaconRestApiTypes.PARAMETER_STATE_ID;
 import static tech.pegasys.teku.beaconrestapi.handlers.v1.beacon.StatusParameter.getApplicableValidatorStatuses;
+import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_BAD_REQUEST;
 import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_OK;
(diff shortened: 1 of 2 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
```diff
@@ -76,7 +79,20 @@ public PostStateValidators(final DataProvider dataProvider) {
...
+      requestBody = request.getOptionalRequestBody();
+    } catch (RuntimeException e) {
+      final Throwable throwable = Throwables.getRootCause(e);
+      if (throwable instanceof JsonParseException) {
+        request.respondError(SC_BAD_REQUEST, throwable.getMessage());
+      } else {
+        throw e;
+      }
+      return;
+    }
```
lucassaldanha, line 94:
```suggestion
    try {
      requestBody = request.getOptionalRequestBody();
    } catch (JsonProcessingException e) {
      final Throwable throwable = Throwables.getRootCause(e);
      request.respondError(SC_BAD_REQUEST, throwable.getMessage());
      return;
    } catch (RuntimeException e) {
      throw e;
    }
```

lucassaldanha (approved):
LGTM. Just a nit suggestion when handling the exceptions.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostStateValidators.java
courtneyeh: That would not work because the exception is a `RuntimeException`, but the root cause of the exception is a `JsonProcessingException`. So cannot just catch the `JsonProcessingException` as you have.

lucassaldanha: I'll take a look later. Maybe there is something we can do on the getOptionalRequeatBody.

But it can be done in a separate PR, no need to block this.

PR #7740, opened 2023-11-21
Add weak subjectivity period check when starting Teku with an existing database

### LINKED ISSUE #7727: Add weak subjectivity period check when starting Teku with a database that is too old
Syncing out of weak subjectivity period should also be enforced when there is an existing database that is too old. Right now it only happens when we are initializing the database.

Related to https://github.com/Consensys/teku/pull/7698

lucassaldanha:
## PR Description

Syncing out of weak subjectivity period should also be enforced when there is an existing database that is too old.

## Fixed Issue(s)
fixes #7727

## Documentation

## Changelog


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -409,17 +409,26 @@ protected SafeFuture<?> initialize() {
                     spec))
         .thenCompose(
             client -> {
+              if (isAllowSyncOutsideWeakSubjectivityPeriod()) {
+                LOG.info(
```
rolfyone: i still think this is `LOG.warn`

rolfyone: oh actually if we get to the status log, thats not a problem :)

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -502,6 +502,14 @@ public void warnFlagDeprecation(final String oldFlag, final String newFlag) {
         Color.YELLOW);
   }
 
+  public void warnIgnoringWeakSubjectivityPeriod() {
```
rolfyone, line 505: In an ideal world we'd be able to show this function is called when needed, but i had a quick look at beaconChainController and it might be more trouble than it's worth, so just tech debt for us to consider in future maybe...

rolfyone (approved):
LGTM

PR #7746, opened 2023-11-22
Don't use FailoverValidatorApiHandler if no failovers configured

StefanBratanov:
## PR Description
Essentially, if no failovers configured, just use the primary `ValidatorApiChannel` instead of using the `FailoverValidatorApiHandler`.

Also another major change in this PR is removing metrics from the `FailoverValidatorApiHandler` and moving them to the new `RemoteMetricRecordingValidatorApiChannel` which will be attached to every remote channel. This leaves the `FailoverValidatorApiHandler` cleaner and not concerned with metrics. 

The above was achieved by making `MetricRecordingValidatorApiChannel` abstract and create two implementations `InProcessMetricRecordingValidatorApiChannel` and `RemoteMetricRecordingValidatorApiChannel` (the only difference is the addition of the endpoint label)

`TODO:` Actually not sure about this, because the dashboard uses  `beacon_node_requests_total` and even if failovers are configured and for example attestations are relayed, there will be few published attestations, so have to think about that use case. 

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
+++ b/acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/RemoteValidatorAcceptanceTest.java
@@ -51,7 +51,7 @@ public class RemoteValidatorAcceptanceTest extends AcceptanceTestBase {
     beaconNode.start();
     validatorClient.start();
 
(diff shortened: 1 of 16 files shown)


StefanBratanov:
Will close this PR because of metrics used in dashboard inconstencies. Need to analyze further.

PR #7749, opened 2023-11-22
Optimize BlobSidecarGossipValidator

### LINKED ISSUE #7654: Move to sidecar inclusion proof

zilm13:
## PR Description
- Cache size calculation looks a bit ugly but I don't want to create new constant for this, open for any improvements
- At first I reused the checks moving most to dedicated method but copy-paste looks better from my side

## Fixed Issue(s)

Fixes #7654 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarGossipValidator.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarGossipValidator.java
@@ -14,6 +14,7 @@
 package tech.pegasys.teku.statetransition.validation;
 
 import static tech.pegasys.teku.infrastructure.async.SafeFuture.completedFuture;
+import static tech.pegasys.teku.spec.config.Constants.BEST_CASE_NON_FINALIZED_EPOCHS;
 import static tech.pegasys.teku.spec.config.Constants.VALID_BLOCK_SET_SIZE;
 import static tech.pegasys.teku.statetransition.validation.InternalValidationResult.ACCEPT;
 import static tech.pegasys.teku.statetransition.validation.InternalValidationResult.ignore;
@@ -66,7 +67,8 @@ public class BlobSidecarGossipValidator {
 
     final int validInfoSize = VALID_BLOCK_SET_SIZE * maybeMaxBlobsPerBlock.orElse(1);
     // It's not fatal if we miss something and we don't need finalized data
-    final int validSignedBlockHeadersSize = spec.getGenesisSpec().getSlotsPerEpoch() * 3;
+    final int validSignedBlockHeadersSize =
+        spec.getGenesisSpec().getSlotsPerEpoch() * BEST_CASE_NON_FINALIZED_EPOCHS;
 
     return new BlobSidecarGossipValidator(
         spec,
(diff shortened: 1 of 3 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarGossipValidator.java
```diff
@@ -63,14 +65,17 @@ public static BlobSidecarGossipValidator create(
     final Optional<Integer> maybeMaxBlobsPerBlock = spec.getMaxBlobsPerBlock();
 
     final int validInfoSize = VALID_BLOCK_SET_SIZE * maybeMaxBlobsPerBlock.orElse(1);
+    // It's not fatal if we miss something and we don't need finalized data
+    final int validSignedBlockHeadersSize = spec.getGenesisSpec().getSlotsPerEpoch() * 3;
```
mehdi-aouadi: I'd rather extract the 3 to a config constant

mehdi-aouadi (commented):
I'd prefer extracting the check  to avoid maintaining two code blocks 😄
I'd also add a test for the known block headers cache (make sure we check everything again if more than x epochs passed since we've verified a blob with a specific block header)...

zilm13:
@mehdi-aouadi I want to keep order the same as in the spec in long case and the checks we need in short case are not all one by one from it, I've tried to reuse it first time, but it actually looks not so good.
I will add a test, good idea

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarGossipValidator.java
zilm13: yeah, and it could have a proper name, it's not a magic number, will do it

mehdi-aouadi (approved):
LGTM

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlobSidecarGossipValidator.java
```diff
@@ -241,19 +256,63 @@ public SafeFuture<InternalValidationResult> validate(final BlobSidecar blobSidec
...
               }
 
-              if (!receivedValidBlobSidecarInfoSet.add(
-                  new SlotProposerIndexAndBlobIndex(
-                      blockHeader.getSlot(),
-                      blockHeader.getProposerIndex(),
-                      blobSidecar.getIndex()))) {
-                return ignore(
-                    "BlobSidecar is not the first valid for its slot and index. It will be dropped.");
+              synchronized (this) {
```
StefanBratanov: why synchronised when using `LimitedSet.createSynchronized` ?

zilm13: "It is imperative that the user manually synchronize on the returned map when traversing any of its collection views via Iterator, Spliterator or Stream:"
Looks like we are safe on atomic operations, removed

StefanBratanov:
LGTM to me as well

Issue #6595, opened 2022-12-13
Support late block re-orging


ajsutton:
### Description
To discourage the late publication of blocks, implement support for late block re-orging: https://github.com/ethereum/consensus-specs/pull/3034

benjaminion:
This has now [been merged](https://github.com/ethereum/consensus-specs/pull/3034), so we can get on with implementing it.

tbenr:
Don't forget to re-enable related reference tests (disabled here: #7695)

tbenr:
We merged #9817 so we can close this issue once it will be enabled by default

PR #7750, opened 2023-11-23
Added more utilities for late block reorg

rolfyone:
- is_parent_strong
 - is_head_weak
 - is_proposing_on_time
 - is_ffg_competitive

The head_weak and parent_strong functions I put into store ultimately, as it has all the context needed. It did mean a little restructuring to enable testing, but I think it currently is a sensible home for it.

partially addresses #6595

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/client/BlockTimelinessTracker.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/client/BlockTimelinessTracker.java
@@ -15,6 +15,8 @@ package tech.pegasys.teku.storage.client;
 
 import static tech.pegasys.teku.spec.constants.NetworkConstants.INTERVALS_PER_SLOT;
 
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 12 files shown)


storage/src/main/java/tech/pegasys/teku/storage/client/BlockTimelinessTracker.java
```diff
@@ -87,7 +94,34 @@ public void setBlockTimelinessFromArrivalTime(
...
+  // is_proposing_on_time from consensus-spec
+  // 'on time' is before we're half-way to the attester time. logically, if the slot is 3 segments,
+  // then splitting into 6 segments is half-way to the attestation time.
+  public boolean isProposingOnTime(final UInt64 slot) {
+    final UInt64 slotStartTimeMillis =
+        spec.getSlotStartTimeMillis(slot, genesisTimeMillisSupplier.get());
+    final UInt64 timelinessLimit = spec.getMillisPerSlot(slot).dividedBy(INTERVALS_PER_SLOT * 2);
+    final UInt64 currentTimeMillis = timeProvider.getTimeInMillis();
+    final boolean isTimely =
+        currentTimeMillis.minusMinZero(slotStartTimeMillis).isLessThan(timelinessLimit);
```
lucassaldanha, line 110: In the spec, I noticed the current time is coming from the store (`store.time`). Is the current time from the time provider equivalent or should we be using `Store.getTimeMillis()`?

The same question applies to store.genesis_time and `Store.getGenesisTime()`.

lucassaldanha (approved):
LGTM. Just added a question about using `TimeProvider` v using the time from the `Store`.

storage/src/main/java/tech/pegasys/teku/storage/client/BlockTimelinessTracker.java
rolfyone: genesisTime is memoized, we don't need to check continually...

store time is reliant on other factors, i think we want real time if we're talking about potentially doing a reorg... it'll be a much more accurate indication of how much time we have

PR #7751, opened 2023-11-23
Enable VC crashing on 0 keys loaded

### LINKED ISSUE #7647: Enable VC crashing on 0 keys loaded

courtneyeh:
fixes #7647 

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -130,6 +130,16 @@ public class ValidatorOptions {
       fallbackValue = "true")
   private boolean blockV3Enabled = ValidatorConfig.DEFAULT_BLOCK_V3_ENABLED;
 
+  @Option(
+      names = {"--allow-no-loaded-keys"},
+      paramLabel = "<BOOLEAN>",
+      description = "Enable allowing no loaded validator keys",
```
rolfyone: maybe something like `exit-when-no-validator-keys-enabled`

would at least need to finish with `enabled`

it would not be hidden...

The description would be something like "This option will terminate the process If no validator keys are found during startup"

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -54,20 +54,23 @@ public class ValidatorLoader {
...
 
   private ValidatorLoader(
       final List<ValidatorSource> validatorSources,
       final Optional<ValidatorSource> mutableLocalValidatorSource,
       final Optional<ValidatorSource> mutableExternalValidatorSource,
       final GraffitiProvider graffitiProvider,
       final Optional<DataDirLayout> maybeDataDirLayout,
-      final SlashingProtectionLogger slashingProtectionLogger) {
+      final SlashingProtectionLogger slashingProtectionLogger,
+      final boolean allowNoLoadedKeys) {
```
rolfyone: This would likely be a lot simpler at the ValidatorClientService layer rather than here...

basically I'd suggest putting a check in `ValidatorClientService.create` immediately prior to `return new ValidatorClientService` something like:
```
if (validatorLoader.getOwnedValidators().getActiveValidators().size() == 0) {
}
```

I'm not sure we're looking at only validators with a local signer, that's probably over-complicating things and reducing what it can be used for...

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -334,7 +334,8 @@ private void initialise() {
                 config.validatorClient().getValidatorConfig().getValidatorExternalSignerUrl()),
             asyncRunner,
             metricsSystem,
-            dataDirLayout);
+            dataDirLayout,
+            config.validatorClient().getValidatorConfig().allowNoLoadedKeys());
```
lucassaldanha: `config.validatorClient().getValidatorConfig()` is being used a couple of times in this method. I think extracting it into a variable can improve readability.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -223,7 +235,8 @@ public static ValidatorLoader create(
       final PublicKeyLoader publicKeyLoader,
       final AsyncRunner asyncRunner,
       final MetricsSystem metricsSystem,
-      final Optional<DataDirLayout> maybeMutableDir) {
+      final Optional<DataDirLayout> maybeMutableDir,
+      final boolean allowNoLoadedKeys) {
```
lucassaldanha: The value of `allowNoLoadedKeys` exists within `ValidatorConfig`. There is no reason to add it as an extra parameter on create(). You can just use `config.allowNoLoadedKeys()` when needed.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -130,6 +130,16 @@ public class ValidatorOptions {
       fallbackValue = "true")
   private boolean blockV3Enabled = ValidatorConfig.DEFAULT_BLOCK_V3_ENABLED;
 
+  @Option(
+      names = {"--exit-when-no-validator-keys-enabled"},
+      paramLabel = "<BOOLEAN>",
+      description = "Enable allowing no loaded validator keys",
```
rolfyone: we'll need to do some more work on this description

courtneyeh: Just fixing it now

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -270,6 +270,12 @@ public static ValidatorClientService create(
             })
         .always(() -> LOG.trace("Finished starting validator client service."));
 
+    if (validatorConfig.isExitWhenNoValidatorKeysEnabled()
+        && validatorLoader.getOwnedValidators().getActiveValidators().size() == 0) {
+      throw new IllegalStateException(
+          "No loaded validators when --exit-when-no-validator-keys-enabled option is false");
+    }
```
rolfyone: my guess is that this will spam exceptions stacktrace to the log on exit...

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -270,6 +270,12 @@ public static ValidatorClientService create(
             })
         .always(() -> LOG.trace("Finished starting validator client service."));
 
+    if (validatorConfig.isExitWhenNoValidatorKeysEnabled()
+        && validatorLoader.getOwnedValidators().getActiveValidators().size() == 0) {
+      throw new NoValidatorKeysStateException(
+          "No loaded validators when --exit-when-no-validator-keys-enabled option is false");
+    }
+
```
lucassaldanha, line 278: We need at least one test for this logic.

teku/src/main/java/tech/pegasys/teku/cli/BeaconNodeCommand.java
```diff
@@ -347,12 +348,25 @@ public Integer call() {
...
+      if (t instanceof NoValidatorKeysStateException) {
+        return 2;
+      }
     }
     return 1;
   }
 
   public void reportUnexpectedError(final Throwable t) {
-    getLogger().fatal("Teku failed to start", t);
+    if (t instanceof NoValidatorKeysStateException) {
```
lucassaldanha: Could use ExceptionUtils to check if `NoValidatorKeysStateException` is the cause (in case it gets wrapped up on something else).

lucassaldanha: Same applies to when we are doing the check and returning status = 2.

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
```diff
@@ -279,6 +283,10 @@ public boolean isBlockV3Enabled() {
     return blockV3Enabled;
   }
 
+  public boolean isExitWhenNoValidatorKeysEnabled() {
+    return exitWhenNoValidatorKeysEnabled;
+  }
```
lucassaldanha, line 288: NIT: we could add a unit test to ensure the default for `exitWhenNoValidatorKeysEnabled` is false.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -142,6 +143,21 @@ public static TekuNode create(
...
+    setUpStart();
+    container.start();
+  }
+
+  public void startWithFailure(final String expectedError) throws Exception {
+    setUpStart();
+    container.waitingFor(
+        new LogMessageWaitStrategy()
+            .withRegEx(".*Teku failed to start.*")
+            .withRegEx(".*" + expectedError + ".*")
```
lucassaldanha: The option `withRegEx()` isn't additive. The second time you use it you are effectively overriding what was set in the first one. Might as well just have the latter that uses the `expectedError`.

lucassaldanha: For reference: https://github.com/masinger/testcontainers-java/blob/fe904c40eb7459fbbd1ecf57d47a839a5f0dc32f/core/src/main/java/org/testcontainers/containers/wait/strategy/LogMessageWaitStrategy.java#L54-L57

lucassaldanha (approved):

PR #7752, opened 2023-11-23
Don't use FailoverValidatorApiHandler if no failovers configured

StefanBratanov:
## PR Description

- Essentially use the `RemoteValidatorApiHandler` instead of the `FailoverValidatorApiHandler`, when there are no failovers configured. There is a very small overhead and I assume most users don't have failovers configured by default.
- In that case, `validator_remote_beacon_nodes_requests_total` metric would not be used when there are no failovers configured. Added a line to the `CHANGELOG`. It makes sense because both metrics are replicated in case of a single BN configured and don't provide much value
- Includes some refactoring changes in `RemoteBeaconNodeApi` and `SentryBeaconNodeApi`

## Fixed Issue(s)
N/A

## Documentation

## Changelog


StefanBratanov:
Closing again until better way to reconcile `metrics` is found

PR #7753, opened 2023-11-24
Implemented get_proposer_head for late block reorg

rolfyone:
Implemented get_proposer_head function, and tested results via the reftests.

It's going to be more efficient to get the justified state once, so have refactored isHeadWeak and isParentStrong.

The ordering of checks was cheapest -> most expensive, we can avoid fetching state if any of the lighter tests fail.

Currently this function is only accessed in reftests.

partially addresses #6595

## Documentation

## Changelog


lucassaldanha (approved):
LGMT

PR #3136, opened 2020-11-03
Use ProtoArray instead of BlockTree in Store

ajsutton:
## PR Description
Follow on to #3124 to move the `ProtoArray` inside `Store`, replacing the `BlockTree` which previously held block hashes and slots.

`BlockMetadataStore` interface is introduced so `Store` can use either `ProtoArray` or `BlockTree` for block hashes which allows a `BlockTree` backed `Store` to be used to migrate the `ProtoArray` data from snapshot based to being rebuilt from the hot block data (and thus always in sync with `Store`).  Once the migration is done, the `BlockTree` in `Store` is replaced with the `ProtoArray` to free up memory.

## Fixed Issue(s)
#2832 

## Documentation

- [x] I thought about documentation and added the `documentation` label to this PR if updates are required.

PR #2425, opened 2020-07-22
[Issue 2291] Remove synchronous Store state retrieval methods

mbaxter:
## PR Description
Remove `Store.getBlockAndState()` and `Store.getBlockState()`.

## Fixed Issue(s)
Part of #2291

## Documentation

- [x] I thought about documentation and added the `documentation` label to this PR if updates are required.

PR #7756, opened 2023-11-25
Fix store deadlock

zilm13:
## PR Description
There is a possibility of blocking transaction commit by trying to acquire readLock on another thread while having writeLock, see thread dump
[deadlock.txt](https://github.com/Consensys/teku/files/13464332/deadlock.txt)
Reverts #7594 

## Documentation

## Changelog

### DIFF
--- a/ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
+++ b/ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.locks.Lock;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -34,6 +35,21 @@ public interface BlockProvider {
 
   static BlockProvider fromDynamicMap(Supplier<Map<Bytes32, SignedBeaconBlock>> mapSupplier) {
     return (roots) -> fromMap(mapSupplier.get()).getBlocks(roots);
+  }
+
+  static BlockProvider fromMapWithLock(
+      final Map<Bytes32, SignedBeaconBlock> blockMap, final Lock readLock) {
+    return (roots) -> {
+      readLock.lock();
+      try {
+        return SafeFuture.completedFuture(
+            roots.stream()
+                .filter(blockMap::containsKey)
+                .collect(Collectors.toMap(Function.identity(), blockMap::get)));
+      } finally {
+        readLock.unlock();
+      }
+    };
   }
 
   static BlockProvider fromMap(final Map<Bytes32, SignedBeaconBlock> blockMap) {
--- a/ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
+++ b/ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
@@ -18,14 +18,13 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.locks.Lock;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.tuweni.bytes.Bytes32;
-import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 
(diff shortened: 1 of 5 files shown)


zilm13:
```
    final Optional<SignedBeaconBlock> inMemoryBlock = getBlockIfAvailable(blockRoot);
    if (inMemoryBlock.isPresent()) {
      return SafeFuture.completedFuture(inMemoryBlock);
    }

    // Retrieve block
```
this block could be removed too, it's part of blockProvider
UPDATE: removed

zilm13:
Finally I think the original issue is still relevant, so I've returned lock, but dropped asyncRunner.

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
```diff
@@ -39,21 +38,18 @@ static BlockProvider fromDynamicMap(Supplier<Map<Bytes32, SignedBeaconBlock>> ma
...
-              }
-            });
+      final Map<Bytes32, SignedBeaconBlock> blockMap, final Lock readLock) {
+    return (roots) -> {
+      readLock.lock();
+      try {
+        return SafeFuture.completedFuture(
+            roots.stream()
+                .filter(blockMap::containsKey)
+                .collect(Collectors.toMap(Function.identity(), blockMap::get)));
```
rolfyone, line 48: worth re-using fromMap inside here?

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
```diff
@@ -39,21 +38,18 @@ static BlockProvider fromDynamicMap(Supplier<Map<Bytes32, SignedBeaconBlock>> ma
...
-                return roots.stream()
-                    .filter(blockMap::containsKey)
-                    .collect(Collectors.toMap(Function.identity(), blockMap::get));
-              } finally {
-                readLock.unlock();
-              }
-            });
+      final Map<Bytes32, SignedBeaconBlock> blockMap, final Lock readLock) {
+    return (roots) -> {
+      readLock.lock();
```
rolfyone, line 43: i think potentially the answer should be more like fromMapWithLock becomes a member of the store, and the store doesn't go passing its locks about - it's kind of a bit of a scary concept...

If the store instead had 
```
private BlockProvider createBlockProvider(blockMap) {
    readLock.lock();
    try {
      return BlockProvider.fromMap(blockMap);
    } finally {
      readLock.unlock();
    }
}
```

Then we wouldn't be allowing other parties play with our read locks, which would be far cleaner, and also you can see where the read lock happens, without having to go digging down paths to other objects...

Also given this is called during constructor, it might even be better off locking before that process starts, and just using fromMap, as until this is done, it'd surely be bad to get stuck waiting for a lock... But I haven't thought that through fully... The upside would be queries to the store would be waiting on the lock...

Im still not convinced it's going to end well potentially - a safe future inside the constructor, if anything goes wrong the object creation (in this case the store) will fail... we might be better off rethinking the construction and having a create method where all of this is done up front and the constructor has a simpler job...  This might be more refactoring for now than we need, but worth considering @tbenr

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
zilm13: I don't understand how it's possible?

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
zilm13: I didn't like sharing lock outside too, but the code you propose will do the lock when calling `createBlockProvider` instead of when using blockProvider. I don't see a way we could incorporate lock in BlockProvider without sharing it to BlockProvider or I don't understand something. The way we could go - drop `blocks` cache from BlockProvider at all and call `getBlockIfAvailable` along with BlockProvider whenever we need block from BlockProvider loosing the comfort we had before but not sharing the lock.

tbenr: I agree that passing locks isn't a nice thing.. and it's confusing because sometime you don't know which lock you are dealing with. But a change here seems like out of scope and I'd like to be focused on restoring the status quo if this is the issue (and seems to be at least a part of the problem)

tbenr:
What I also noticed is that this ~~recently~~ added [here](https://github.com/Consensys/teku/pull/3136):
```java
  private void calculatePrunedHotBlockRoots() {
    final BeaconBlockSummary finalizedBlock = tx.getLatestFinalized().getBlockSummary();
```

broke the assumption here:

```java
  @Override
  public AnchorPoint getLatestFinalized() {
    if (finalizedCheckpoint.isPresent()) {
      // Ideally we wouldn't join here - but seems not worth making this API async since we're
      // unlikely to call this on tx objects
      final SignedBlockAndState finalizedBlockAndState =
          retrieveBlockAndState(finalizedCheckpoint.get().getRoot()).join().orElseThrow();
```

That code\comment was added here: https://github.com/Consensys/teku/pull/2425

And the more I read about thread pools and async `CompletableFuture`s the more I get scared by those `join()`s.

tbenr (approved):
LGTM.
To move this forward, I vote for rolling the async back for now to address the "emergency". And we can address improvements without having an issue in the code.

tbenr:
I'm starting to be convinced that this is caused when the node import a block finalizing a new checkpoint but we don't have the state for it. So  the  `StoreTransaction`  containing the new checkpoint, when committing acquires the write lock and then the flow runs  `calculatePrunedHotBlockRoots` within the tx and state regeneration fires. `BlockProvider` (with lock) is then engaged that runAsync on another thread and tries to acquire the read lock before reading blocks from the store. And we and up in deadlock.

ethereum/dataproviders/src/main/java/tech/pegasys/teku/dataproviders/lookup/BlockProvider.java
rolfyone: this is redundant if we do the other comment...

zilm13:
Closed in favor of #7766

PR #7759, opened 2023-11-28
Teku fails to start when Validator API cannot read password file

### LINKED ISSUE #7708: if validator-api-bearer cannot be created or read, needs to be handled better
An example on discord was re. the validator-api-bearer existing, but being a folder rather than a file.

This could be clearer if it does occur, as the service won't be able to be accessed without that bearer token.

At a minimum I would suggest if it's not a file, or can't be read, then an error should be written to log, because that's effectively failing to start something (can't be accessed, even if it's running).

lucassaldanha:
## PR Description
Before this change, if Teku couldn't read the password file for any reason, it would start up but it wouldn't serve any Validator API requests. This can be misleading so we decided to fail-fast and don't startup if something is wrong (likely to be a configuration issue).

## Fixed Issue(s)
fixes #7708 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Breaking Changes
 - By default, Teku won't allow syncing from genesis, users should use `--checkpoint-sync-url` when starting a new node. It is possible to revert back to the previous behaviour using the flag `--ignore-weak-subjectivity-period-enabled`.
-- Teku will fail to start if the Validator API cannot read the password file used for authorization.
+- Teku will fail to start if the Validator API cannot read the password file used for authorization. Previously, if the Validator API was enabled but Teku couldn't read the password file, it would start but no request to the Validator API would be served. 
 
 ### Additions and Improvements
 - Support to new Beacon APIs `publishBlindedBlockV2` and `publishBlockV2` which introduce broadcast validation parameter. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Breaking Changes
 - By default, Teku won't allow syncing from genesis, users should use `--checkpoint-sync-url` when starting a new node. It is possible to revert back to the previous behaviour using the flag `--ignore-weak-subjectivity-period-enabled`.
+- Teku will fail to start if the Validator API cannot read the password file used for authorization.
 
 ### Additions and Improvements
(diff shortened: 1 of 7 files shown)


infrastructure/restapi/src/test/java/tech/pegasys/teku/infrastructure/restapi/AuthorizationManagerTest.java
```diff
@@ -82,31 +83,6 @@ void shouldDenyAccessIfHeaderNotSet(@TempDir final Path tempDir) throws Exceptio
...
-    when(context.method()).thenReturn(HandlerType.GET);
-    when(context.matchedPath()).thenReturn("/aPath");
-    when(context.header("Authorization")).thenReturn(null);
-    AuthorizationManager manager = new AuthorizationManager(tempDir.resolve("passwd"));
-    manager.manage(handler, context, roles);
-    verify(handler, times(0)).handle(context);
-    verify(context).status(401);
-    verify(context).json(any());
-  }
-
```
lucassaldanha, line 109: Removed these 2 tests now that we don't have a state with no bearer loaded.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Breaking Changes
 - By default, Teku won't allow syncing from genesis, users should use `--checkpoint-sync-url` when starting a new node. It is possible to revert back to the previous behaviour using the flag `--ignore-weak-subjectivity-period-enabled`.
+- Teku will fail to start if the Validator API cannot read the password file used for authorization.
```
rolfyone: Maybe add something like 'Previously, if the validator-api was enabled and the password file was not available, the validator would continue to startup without the validator-api` or something...

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/AuthorizationManager.java
```diff
@@ -34,22 +33,53 @@
...
+                    String.format("password file %s is empty", passwordFile.getAbsolutePath()));
+                LOG.error(errorMessage);
+                return new IllegalStateException(errorMessage.toString());
+              });
     } catch (IOException e) {
-      LOG.error(initFailedMessage, e);
-      bearer = Optional.empty();
+      errorMessage.append(String.format(e.getMessage(), passwordFile.getAbsolutePath()));
+      LOG.error(errorMessage);
+      throw new IllegalStateException(errorMessage.toString());
```
rolfyone: i think this is another case where we probably dont want to throw Illegal state, much for the same reason that we'd be spamming the logs with tons of data...

rolfyone (approved):
LGTM, probably worth addressing the codeql (even if to reject)

PR #7763, opened 2023-11-29
Implemented should_override_fork_choice for late block reorg

rolfyone:
Implemented should_override_fork_choice function, and tested results via the reftests.

There's a level of duplication in the 'should_override_fork_choice' and 'get_proposer_head', but I've stayed true to the spec, while fixing ordering a little to do cheaper tests first.

Ultimately, the function will exit fast once it finds a mis-match, and if it gets to the end then things pass.

Also had to adjust timeliness to use store time, and I made the Store implement the TimeProvider to fix an issue with using a system time provider when store time was needed.

Finally, while I did successfully run the reftests, I have to leave the reftests disabled until we get a release that doesnt contain backwards ticks. It's possible to remove the ticks to run the tests, but they won't run out of the box.

partially addresses #6595

## Documentation

## Changelog

### DIFF
--- a/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/ValidatorIsConenctedProviderImpl.java
+++ b/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/ValidatorIsConenctedProviderImpl.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright Consensys Software Inc., 2023
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.services.beaconchain;
+
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.statetransition.forkchoice.ForkChoiceNotifier;
+import tech.pegasys.teku.storage.client.ValidatorIsConnectedProvider;
+
+public class ValidatorIsConenctedProviderImpl implements ValidatorIsConnectedProvider {
+  private final ForkChoiceNotifier forkChoiceNotifier;
+
+  public ValidatorIsConenctedProviderImpl(ForkChoiceNotifier forkChoiceNotifier) {
+    this.forkChoiceNotifier = forkChoiceNotifier;
+  }
+
+  @Override
+  public boolean isValidatorConnected(int validatorId, UInt64 slot) {
+    return forkChoiceNotifier.validatorIsConnected(UInt64.valueOf(validatorId), slot);
+  }
+}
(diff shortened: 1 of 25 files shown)


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/ValidatorIsConenctedProviderImpl.java
lucassaldanha, line 1: Typo in class name:

`Conencted` instead of `Connected`

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
```diff
@@ -688,6 +695,93 @@ public Bytes32 getProposerHead(final Bytes32 headRoot, final UInt64 slot) {
...
+      // !shufflingStable or !ffgCompetetive or !finalizationOk, or parentSlot is not found
+      return false;
+    }
+
+    final boolean isParentSlotOk =
+        maybeParentSlot.map(uInt64 -> uInt64.increment().equals(head.getSlot())).orElse(false);
+    final boolean isProposingOnTime = isProposingOnTime(proposalSlot);
+    final boolean isCurrentTimeOk =
+        head.getSlot().equals(currentSlot)
+            || (currentSlot.increment().equals(proposalSlot) && isProposingOnTime);
```
lucassaldanha: Why are we incrementing the current slot before comparing to the proposalSlot? in:
```
currentSlot.increment().equals(proposalSlot) && isProposingOnTime
```

From the spec it looks like we are supposed to do:
```
currentSlot.equals(proposalSlot) && isProposingOnTime
```

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
```diff
@@ -688,6 +695,93 @@ public Bytes32 getProposerHead(final Bytes32 headRoot, final UInt64 slot) {
...
+        spec.atSlot(proposalSlot).getForkChoiceUtil().isShufflingStable(proposalSlot);
+
+    final boolean isFfgCompetitive =
+        store.isFfgCompetitive(headRoot, head.getParentRoot()).orElse(false);
+    final boolean isFinalizationOk =
+        spec.atSlot(proposalSlot).getForkChoiceUtil().isFinalizationOk(store, proposalSlot);
+    final Optional<UInt64> maybeParentSlot = getSlotForBlockRoot(head.getParentRoot());
+
+    if (!isShufflingStable || !isFfgCompetitive || !isFinalizationOk || maybeParentSlot.isEmpty()) {
+      // !shufflingStable or !ffgCompetetive or !finalizationOk, or parentSlot is not found
```
lucassaldanha, line 723: typo: `ffgCompetetive` should be `ffgCompetitive`

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/ValidatorIsConenctedProviderImpl.java
rolfyone: oh ick. ta

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
rolfyone: because initialization was wrong -fixed.

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
rolfyone: i hate that word so much...

lucassaldanha (approved):

PR #7765, opened 2023-11-29
Increase delay for ExternalSigner timeout integration test

StefanBratanov:
## PR Description
I feel like the delay increase to cause a timeout was too small (5 millis) which could cause the test failing like https://app.circleci.com/pipelines/github/Consensys/teku/28610/workflows/98101832-0d21-484c-b22e-f465e9114a79/jobs/210857/tests so increased the timeout to more sensible 200 millis. I was not able to reproduce the issue locally using 5 millis, but it could be my OS,

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM. Let's test the theory

PR #7766, opened 2023-11-29
Fixed store passing its locks about

rolfyone:
Noticed this issue in the deadlock PR, but there was no appetite to fix in that, so i've raised a PR to encapsulate the issue and fix.

We shouldn't be passing locks around, and I'd like this issue fixed so here's the proposed solution from #7756

## Documentation

## Changelog


zilm13:
You are not using lock while using blockProvider, you are using it only when setting it up and freeing immediately. Restore deleted test, add the same setup you made in Store, and you'll see it

rolfyone:
the test is the only usage outside the constructor. pushed a change that should make it cleaner.

zilm13:
Looks good to me, I'd just like to see test back but new BlockProvider is deep in Store, it would be ugly and complex to test it.

tbenr (approved):
LGTM

PR #7767, opened 2023-11-29
Deneb upgrade acceptance test

zilm13:
## PR Description
Did it the same way Capella acceptance test done, so it's disabled. I don't know how to move on with it. It passes locally but on CI we will get flakiness with time-based transition.

## Documentation

## Changelog


StefanBratanov:
I think we should add some assertion on blobs. Possibly change our logs to say  maybe e.g. "Published block and 3 blobs"

zilm13:
@StefanBratanov I like your suggestion, will add it

StefanBratanov (approved):
LGTM

zilm13:
Commenting on why no checks for blobs added:

We use real EL in acceptance tests, so we need some agent to generate correct Blob transactions to fall into EL pool so it will use it in payload when requested on block production duty. They should pass all EL internal verification etc.
We don't have anything like this even for basic transactions, so adding it especially for blob transaction is a big amount of work. It may worth to do it or research how to plug in stub used in Hive, but it's definitely a separate issue.

PR #7769, opened 2023-11-30
Consolidate interfaces related to BlockV3, moving from deprecated method to Optional<Boolean> blinded

tbenr:
I was adding a block production performance tracker (#7770) and I realize that there was some code duplication to apply my simple changes to.

So I decided to consolidate the interface, since probably V2 support will remain for a long time and having code duplication and `@deprecated` method for so long it doesn't make much sense to me.

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
@@ -116,14 +116,14 @@ public class ValidatorDataProvider {
       final Optional<Bytes32> graffiti,
       final boolean isBlinded) {
     checkBlockProducingParameters(slot, randao);
-    return validatorApiChannel.createUnsignedBlock(slot, randao, graffiti, isBlinded);
+    return validatorApiChannel.createUnsignedBlock(slot, randao, graffiti, Optional.of(isBlinded));
   }
 
   public SafeFuture<Optional<BlockContainerAndMetaData<BlockContainer>>> produceBlock(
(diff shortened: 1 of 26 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java
```diff
@@ -116,14 +116,14 @@ public SafeFuture<Optional<BlockContainer>> getUnsignedBeaconBlockAtSlot(
       final Optional<Bytes32> graffiti,
       final boolean isBlinded) {
     checkBlockProducingParameters(slot, randao);
-    return validatorApiChannel.createUnsignedBlock(slot, randao, graffiti, isBlinded);
+    return validatorApiChannel.createUnsignedBlock(slot, randao, graffiti, Optional.of(isBlinded));
```
zilm13, line 120: We could mark this method (`getUnsignedBeaconBlockAtSlot`) with `@Deprecated`, yeah? All it's callers should be deprecated with V2?

zilm13 (approved):
LGTM 
Why do you think it will take long time for V2 to deprecate?

tbenr:
> Why do you think it will take long time for V2 to deprecate?

I feel like as BN, before completely stop responding to V2 it would take a while: when all client VCs implements V3 and V2 will be removed from the spec? Or maybe when next fork will require a change that will be applied to V3 only, this will be the time.

zilm13:
Oh I see.
It's not even enabled by default

Issue #5808, opened 2022-06-16
Report Block Production Timings (Beacon Node)


StefanBratanov:
Addresses the first point in #5369

### Description

Currently, we track the block production time on the VC, but not on the Beacon Node. It would be useful to report on:
- Time between making the preparePayload call and the getPayload call
- Time in the slot when we ask the EL to prepare a payload
- Time in the slot when we gossip the block
- Builder block production total time if mev-boost is enabled

It could either be a log statement at the end of proposing a block or via metrics or both.

PR #7774, opened 2023-12-01
Add BlockProductionPerformance

### LINKED ISSUE #7770: Add a block production performance tracker
Following up a user reporting logs showing 3.5 seconds passed from block creation API call and first call to builder\local EL, to troubleshoot better worth adding a performance tracker specific to block production, which will work similarly to block import perf tracker.

tbenr:
Needed to create a new module for cross-module performance trackers, cose this new tracker needs to be passed along several layers.

fixes #7770
related to #5808 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/integrationTest/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandlerIntegrationTest.java
+++ b/beacon/validator/src/integrationTest/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandlerIntegrationTest.java
@@ -32,6 +32,7 @@ import tech.pegasys.teku.beacon.sync.events.SyncStateTracker;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.metrics.StubMetricsSystem;
 import tech.pegasys.teku.infrastructure.metrics.Validator.ValidatorDutyMetricUtils;
+import tech.pegasys.teku.infrastructure.time.SystemTimeProvider;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.networking.eth2.gossip.BlobSidecarGossipChannel;
 import tech.pegasys.teku.networking.eth2.gossip.BlockGossipChannel;
@@ -97,6 +98,7 @@ public class ValidatorApiHandlerIntegrationTest {
   private final DutyMetrics dutyMetrics = mock(DutyMetrics.class);
   private final ValidatorApiHandler handler =
       new ValidatorApiHandler(
+          new SystemTimeProvider(),
           chainDataProvider,
           nodeDataProvider,
           combinedChainDataClient,
@@ -117,7 +119,8 @@ public class ValidatorApiHandlerIntegrationTest {
           proposersDataManager,
           syncCommitteeMessagePool,
           syncCommitteeContributionPool,
-          syncCommitteeSubscriptionManager);
+          syncCommitteeSubscriptionManager,
+          true);
 
   @BeforeEach
   public void setup() {
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
@@ -19,6 +19,7 @@ import static tech.pegasys.teku.infrastructure.logging.ValidatorLogger.VALIDATOR
 import static tech.pegasys.teku.infrastructure.metrics.Validator.DutyType.ATTESTATION_PRODUCTION;
 import static tech.pegasys.teku.infrastructure.metrics.Validator.ValidatorDutyMetricUtils.startTimer;
 import static tech.pegasys.teku.infrastructure.metrics.Validator.ValidatorDutyMetricsSteps.CREATE;
+import static tech.pegasys.teku.infrastructure.time.TimeUtilities.secondsToMillis;
 import static tech.pegasys.teku.spec.config.SpecConfig.GENESIS_SLOT;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -45,8 +46,10 @@ import tech.pegasys.teku.api.response.v1.beacon.ValidatorStatus;
 import tech.pegasys.teku.beacon.sync.events.SyncStateProvider;
 import tech.pegasys.teku.bls.BLSPublicKey;
 import tech.pegasys.teku.bls.BLSSignature;
+import tech.pegasys.teku.ethereum.performance.trackers.BlockProductionPerformance;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.ssz.SszList;
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.networking.eth2.gossip.BlobSidecarGossipChannel;
 import tech.pegasys.teku.networking.eth2.gossip.BlockGossipChannel;
@@ -110,6 +113,7 @@ public class ValidatorApiHandler implements ValidatorApiChannel {
    */
   private static final int DUTY_EPOCH_TOLERANCE = 1;
 
+  private final TimeProvider timeProvider;
   private final ChainDataProvider chainDataProvider;
   private final NodeDataProvider nodeDataProvider;
   private final CombinedChainDataClient combinedChainDataClient;
@@ -128,10 +132,12 @@ public class ValidatorApiHandler implements ValidatorApiChannel {
   private final SyncCommitteeContributionPool syncCommitteeContributionPool;
   private final ProposersDataManager proposersDataManager;
   private final BlockPublisher blockPublisher;
+  private final boolean blockProductionPerformanceTrackingEnabled;
 
   private final AttesterDutiesGenerator attesterDutiesGenerator;
 
   public ValidatorApiHandler(
+      final TimeProvider timeProvider,
       final ChainDataProvider chainDataProvider,
       final NodeDataProvider nodeDataProvider,
       final CombinedChainDataClient combinedChainDataClient,
@@ -152,7 +158,9 @@ public class ValidatorApiHandler implements ValidatorApiChannel {
       final ProposersDataManager proposersDataManager,
       final SyncCommitteeMessagePool syncCommitteeMessagePool,
       final SyncCommitteeContributionPool syncCommitteeContributionPool,
-      final SyncCommitteeSubscriptionManager syncCommitteeSubscriptionManager) {
+      final SyncCommitteeSubscriptionManager syncCommitteeSubscriptionManager,
+      final boolean blockProductionPerformanceTrackingEnabled) {
+    this.timeProvider = timeProvider;
     this.chainDataProvider = chainDataProvider;
     this.nodeDataProvider = nodeDataProvider;
     this.combinedChainDataClient = combinedChainDataClient;
@@ -181,6 +189,7 @@ public class ValidatorApiHandler implements ValidatorApiChannel {
             performanceTracker,
             dutyMetrics);
     this.attesterDutiesGenerator = new AttesterDutiesGenerator(spec);
+    this.blockProductionPerformanceTrackingEnabled = blockProductionPerformanceTrackingEnabled;
   }
 
   @Override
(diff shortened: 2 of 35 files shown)


infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/MetricsConfig.java
```diff
@@ -40,6 +40,7 @@ public class MetricsConfig {
   public static final int DEFAULT_METRICS_PUBLICATION_INTERVAL = 60;
   public static final boolean DEFAULT_BLOCK_PERFORMANCE_ENABLED = true;
   public static final boolean DEFAULT_TICK_PERFORMANCE_ENABLED = false;
+  public static final boolean DEFAULT_BLOCK_PRODUCTION_PERFORMANCE_ENABLED = true;
```
mehdi-aouadi, line 43: IMO it would be better to default to false instead. We already have the block performance enabled by default. This could be enabled whenever we need in depth performance tracking...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
```diff
@@ -331,6 +333,9 @@ public SafeFuture<HeaderWithFallbackData> builderGetHeader(
     final SchemaDefinitions schemaDefinitions = spec.atSlot(slot).getSchemaDefinitions();
 
     return engineGetPayload(executionPayloadContext, slot)
+        .thenPeek(
```
mehdi-aouadi: Could the performance tracking be useful when stubbing?!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannel.java
```diff
@@ -137,5 +139,6 @@ SafeFuture<BuilderPayload> builderGetPayload(
   SafeFuture<HeaderWithFallbackData> builderGetHeader(
       ExecutionPayloadContext executionPayloadContext,
       BeaconState state,
-      SafeFuture<UInt256> payloadValueResult);
+      SafeFuture<UInt256> payloadValueResult,
+      Optional<BlockProductionPerformance> blockProductionPerformance);
```
mehdi-aouadi: Could add the `blockProductionPerformance` to the java doc params...

beacon/validator/src/integrationTest/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandlerIntegrationTest.java
```diff
@@ -117,7 +119,8 @@ public class ValidatorApiHandlerIntegrationTest {
           proposersDataManager,
           syncCommitteeMessagePool,
           syncCommitteeContributionPool,
-          syncCommitteeSubscriptionManager);
+          syncCommitteeSubscriptionManager,
+          true);
```
mehdi-aouadi, line 123: It could be set to false since nothing is tested for block production performance

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
```diff
@@ -97,11 +99,17 @@ public ExecutionPayloadResult initiateBlockProduction(
...
-      final boolean isBlind) {
+      final boolean isBlind,
+      final Optional<BlockProductionPerformance> blockProductionPerformance) {
     final ExecutionPayloadResult result;
     if (!isBlind) {
       final SafeFuture<GetPayloadResponse> getPayloadResponseFuture =
-          executionLayerChannel.engineGetPayload(context, blockSlotState.getSlot());
+          executionLayerChannel
+              .engineGetPayload(context, blockSlotState.getSlot())
+              .thenPeek(
```
mehdi-aouadi: Shouldn't we report `engineGetPayload` in `initiateBlockProduction` too?

ethereum/performance-trackers/src/main/java/tech/pegasys/teku/ethereum/performance/trackers/BlockProductionPerformance.java
```diff
@@ -0,0 +1,85 @@
...
+package tech.pegasys.teku.ethereum.performance.trackers;
+
+import tech.pegasys.teku.infrastructure.logging.EventLogger;
+import tech.pegasys.teku.infrastructure.time.PerformanceTracker;
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+
+public class BlockProductionPerformance {
+  public static final String COMPLETE_LABEL = "complete";
+  public static final int LATE_EVENT_MS = 100;
```
mehdi-aouadi: Wouldn't it be better to make it configurable?

mehdi-aouadi:
Maybe worth injecting the `BlockProductionPerformance` to the `ValidatorApiHandler` (instantiate it in the `BeaconChainController`) and add some tests around the `ValidatorApiHandler` to make sure the metrics are reported as expected

teku/src/main/java/tech/pegasys/teku/cli/options/MetricsOptions.java
```diff
@@ -108,6 +108,17 @@ public class MetricsOptions {
...
 
+  @Option(
+      names = {"--Xmetrics-block-production-timing-tracking-enabled"},
+      hidden = true,
+      showDefaultValue = Visibility.ALWAYS,
+      paramLabel = "<BOOLEAN>",
+      description = "Whether block production timing metrics are tracked and reported",
+      fallbackValue = "true",
+      arity = "0..1")
+  private boolean blockProductionPerformanceEnabled =
```
mehdi-aouadi, line 119: I'd rather rename it to `blockProductionPerformanceTrackingEnabled` or `blockProductionTimingEnabled` (everywhere 😁)

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -312,19 +321,44 @@ public SafeFuture<Optional<BlockContainer>> createUnsignedBlock(
...
+              if (blockProductionPerformance.isEmpty()) {
+                return;
+              }
+              maybeState.ifPresent(
+                  state ->
+                      blockProductionPerformance
+                          .get()
+                          .slotTime(secondsToMillis(spec.computeTimeAtSlot(state, slot))));
+              blockProductionPerformance.get().getStateAtSlot();
+            })
```
zilm13: If using noop pattern in tracker it may look something like this.  
```java
        .thenPeek(
            maybeState -> {
               blockProductionPerformance.slotTime(() -> {maybeState.map(state -> secondsToMillis(spec.computeTimeAtSlot(state, slot))}; 
               blockProductionPerformance.getStateAtSlot()
            })
```
And cleaner in other places. It would be good to not affect code readability much with tracking and metrics and better to put more logic in tracker than at the place of product logic. It's not a universal kind of tracker, so it's ok to put some kind of extra code in it rather than here I think

zilm13 (commented):
Just started review, will finish later.
I don't like a bit module chaos around it
`Metrics` are in `infrastructure:metrics`
`PerformanceTracker` is in `infrastructure:time`
`BlockProductionPerformance` is in `ethereum:performance-trackers` alone
From my understanding we use `ethereum` module for some core things like spec. I'd consider to move both `PerformanceTracker` and `BlockProductionPerformance` in `infrastructure:metrics` or make `infrastructure:trackers` dedicated for them if they look too different from `metrics`. Anyway `ethereum` looks not a right place by my opinion.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
rolfyone: not sure if it slides into metrics, it'd be neat if it does... worst case i think something like trackers with minimal deps could be nice...

tbenr:
> From my understanding we use ethereum module for some core things like spec. I'd consider to move both PerformanceTracker and BlockProductionPerformance in infrastructure:metrics or make infrastructure:trackers dedicated for them if they look too different from metrics. Anyway ethereum looks not a right place by my opinion.

Application level classes shouldn't be in `infrastructure`. `PerformanceTracker` is time related and general enough to go in `time`. I do agree with you that having these similar classes all around is bad. Let me try to consolidate.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
tbenr: I agree on the NOOP pattern

tbenr:
> Maybe worth injecting the BlockProductionPerformance to the ValidatorApiHandler (instantiate it in the BeaconChainController) and add some tests around the ValidatorApiHandler to make sure the metrics are reported as expected

To go for this we have to inject a factory because we create a new instance on each flow. To check all the steps we also need to test several layers... I don't feel strongly for that...

teku/src/main/java/tech/pegasys/teku/cli/options/MetricsOptions.java
```diff
@@ -108,6 +108,17 @@ public class MetricsOptions {
       arity = "0..1")
   private boolean tickPerformanceEnabled = MetricsConfig.DEFAULT_TICK_PERFORMANCE_ENABLED;
 
+  @Option(
+      names = {"--Xmetrics-block-production-timing-tracking-enabled"},
```
zilm13, line 112: Though it's hidden flag, I think we may need to add some clarity in names and description on which "metric" is logs only, and which one is real metric and could be used in Prometheus and co.
I think we could name things like this something like `--Xnotify-late-block-production-enabled`

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -312,19 +322,40 @@ public SafeFuture<Optional<BlockContainer>> createUnsignedBlock(
...
-            blockSlotState -> createBlock(slot, randaoReveal, graffiti, blinded, blockSlotState));
+            blockSlotState ->
+                createBlock(
+                    slot,
+                    randaoReveal,
+                    graffiti,
+                    blinded,
+                    blockSlotState,
+                    blockProductionPerformance))
+        .thenPeek(__ -> blockProductionPerformance.complete());
```
zilm13: Maybe it makes sense to put it in something like `alwaysRun`, so we could still recover something in timeout something cases?

ethereum/performance-trackers/src/main/java/tech/pegasys/teku/ethereum/performance/trackers/BlockProductionPerformance.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+package tech.pegasys.teku.ethereum.performance.trackers;
+
+import java.util.function.Supplier;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+
+public interface BlockProductionPerformance {
+  String COMPLETE_LABEL = "complete";
+  int LATE_EVENT_MS = 100;
```
zilm13: We need some confidence on this value.  Though block production is very rare for home stakers and it will not make logs too spammy even with low value, is it really dangerous if it's 130ms?

zilm13 (approved):
LGTM, just few nits

teku/src/main/java/tech/pegasys/teku/cli/options/MetricsOptions.java
tbenr: You're right. I'm thinking of producing some actual metrics for it so we can justify the "metrics" here

ethereum/performance-trackers/src/main/java/tech/pegasys/teku/ethereum/performance/trackers/BlockProductionPerformance.java
tbenr: From the numbers I saw on our goerli nodes, 130 will fire 99% of the times.
So if here we just want to capture a warning, this shouldn't fire that often and I'm thinking about a 250-300 value.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
tbenr: ye makes sense

mehdi-aouadi:
I just noticed I didn't submit my reviews 🤦‍♂️ Nothing really blocking anyway... Yet I just submitted them so you could have some ideas for potential future changes...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
tbenr: well.. if we do performance checks in the stub, at least this times are considered so that the following steps do not take the time spent here.

beacon/validator/src/integrationTest/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandlerIntegrationTest.java
tbenr: well the idea is to see that it doesn't affect the flow even if active.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
tbenr: ahh yes seems like you're right

Issue #7714, opened 2023-11-15
I think teku doesn't submit voluntary exits when node is not in sync?


RatanRSur:

rolfyone:
Thanks for raising, this does sound feasible, a lot of the APIs don't interact well when not in sync. We'll have a look and see if we can improve things on this front...

As an aside, it's 'safe' to submit voluntary exits more than once. Much like bls changes, there's no guarantees they make it through to a block (depends on a block producer seeing the message and retaining it) and also no penalty if it's already been seen.

You can see if it's in your local pool by querying the endpoint:

`curl http://localhost:5051/eth/v1/beacon/pool/voluntary_exits | jq`

This would list any voluntary exits that have been seen and not subsequently been listed in a block (it'll clear on a restart though).

Another 'caveat' is if you submit and then restart your node, you'll probably want to periodically resubmit the exit if it hasn't been included into a block in a reasonable (a couple of hours usually) timeframe.

lucassaldanha:
After looking around the code and doing some testing, I can confirm that Teku does not prevent a user from submitting a voluntary exit while the node is out of sync. However, there are a few scenarios in which the user might not be able to submit the exit when out of sync. Let me try to explain.

Some of the [consensus spec validation rules for voluntary exit messages](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#voluntary-exits) are "state dependent". That means that they depend on what your node sees as the current beacon state (for example: checking if you are trying to create a voluntary exit for a validator that is active).

Think of this extreme example: trying to start a new node from genesis and sending a voluntary exit for my validator. Because you are syncing from genesis, it is possible that your beacon state doesn't even know that you have a validator active. So if you try to submit a voluntary exit it is going to fail. Again, the problem isn't the fact that your node is out of sync, but that your current state results in a invalid voluntary exit.

Now for the question: how to exit a validator if I don't have a sync'd beacon node. You have two options:
1. Start a new Teku node using [checkpoint sync](https://docs.teku.consensys.io/reference/cli#checkpoint-sync-url). This will get you a sync'd node quickly, and then you can submit your exit.
2. Use a service like signed message propagation from Beaconcha.in (https://beaconcha.in/tools/broadcast) to submit an signed voluntary exit in the network w/o the need you running a beacon node.

PR #7777, opened 2023-12-03
Test for voluntary exit rejected reson on API and CLI

lucassaldanha:
Adding tests to ensure Teku returns the reason a voluntary exit message is being rejected on the API and on CLI.

Related to #7714


rolfyone (approved):
LGTM

PR #7785, opened 2023-12-05
 added command line flag to enable fork-choice late block reorgs

rolfyone:
Currently not enabled, just the flag, and defaulting to false, and a test that shows it can be toggled on.

 partially addresses #6595

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -100,6 +100,7 @@ public class ForkChoice implements ForkChoiceUpdatedResultSubscriber {
   private final Subscribers<OptimisticHeadSubscriber> optimisticSyncSubscribers =
       Subscribers.create(true);
   private final TickProcessor tickProcessor;
+  private final boolean forkChoiceLateBlockReorgEnabled;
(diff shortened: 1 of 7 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -113,6 +114,7 @@ public ForkChoice(
       final MergeTransitionBlockValidator transitionBlockValidator,
       final boolean forkChoiceUpdateHeadOnBlockImportEnabled,
       final boolean forkChoiceProposerBoostUniquenessEnabled,
+      final boolean forkChoiceLateBlockReorgEnabled,
       final MetricsSystem metricsSystem) {
```
lucassaldanha, line 118: I understand that this is out of scope for this PR (and should be done in a different PR if ever). But we could have a `ForkChoiceParams` object instead of adding flags to the constructor. Imagine how this is going to look when we have 10 different toggles for the ForkChoice? :)

lucassaldanha (approved):
LGTM

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
rolfyone: at that point it'll probably annoy me enough to refactor :)

PR #7786, opened 2023-12-05
Use POST validator statuses in VC

### LINKED ISSUE #7733: use POST validatorStatuses in VC
Now that the POST validatorStatuses is supported in the BN, we should start using it in the VC, but we probably need a feature flag or to fall back to the existing GET if we get a 404 response...

Ref. #7706

StefanBratanov:
## PR Description

- Add postValidators to `ValidatorRestApiClient`
- Change `RemoteValidatorApiHandler` to try **POST** request before doing a batched **GET** request
- Remove unused methods from `OkHttpValidatorRestApiClient` (they have been replaced by typeDef requests)

## Fixed Issue(s)
fixes #7733 

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/integration-test/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClientTest.java
+++ b/validator/remote/src/integration-test/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClientTest.java
@@ -18,7 +18,6 @@ import static java.util.Collections.emptyMap;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.junit.jupiter.api.Assertions.fail;
-import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_ACCEPTED;
 import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_BAD_REQUEST;
 import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_INTERNAL_SERVER_ERROR;
 import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_METHOD_NOT_ALLOWED;
@@ -29,7 +28,6 @@ import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_OK;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import okhttp3.HttpUrl;
@@ -43,31 +41,20 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
-import tech.pegasys.teku.api.exceptions.RemoteServiceNotAvailableException;
 import tech.pegasys.teku.api.response.v1.beacon.GetGenesisResponse;
 import tech.pegasys.teku.api.response.v1.beacon.GetStateValidatorsResponse;
 import tech.pegasys.teku.api.response.v1.beacon.PostDataFailure;
 import tech.pegasys.teku.api.response.v1.beacon.PostDataFailureResponse;
 import tech.pegasys.teku.api.response.v1.beacon.ValidatorResponse;
 import tech.pegasys.teku.api.response.v1.validator.GetAggregatedAttestationResponse;
-import tech.pegasys.teku.api.response.v1.validator.GetNewBlindedBlockResponse;
 import tech.pegasys.teku.api.response.v1.validator.GetSyncCommitteeContributionResponse;
-import tech.pegasys.teku.api.response.v2.validator.GetNewBlockResponseV2;
 import tech.pegasys.teku.api.schema.Attestation;
-import tech.pegasys.teku.api.schema.BLSSignature;
-import tech.pegasys.teku.api.schema.BeaconBlock;
 import tech.pegasys.teku.api.schema.SignedAggregateAndProof;
-import tech.pegasys.teku.api.schema.SignedBeaconBlock;
 import tech.pegasys.teku.api.schema.SignedVoluntaryExit;
 import tech.pegasys.teku.api.schema.SubnetSubscription;
 import tech.pegasys.teku.api.schema.altair.SyncCommitteeContribution;
-import tech.pegasys.teku.api.schema.bellatrix.BlindedBlockBellatrix;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.provider.JsonProvider;
-import tech.pegasys.teku.spec.Spec;
-import tech.pegasys.teku.spec.SpecMilestone;
-import tech.pegasys.teku.spec.TestSpecFactory;
-import tech.pegasys.teku.spec.util.DataStructureUtil;
 import tech.pegasys.teku.validator.api.CommitteeSubscriptionRequest;
 
 class OkHttpValidatorRestApiClientTest {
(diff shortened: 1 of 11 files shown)


zilm13 (commented):
LGTM, but I'm concerned on the spec filters definition
```
The lists of validator IDs and statuses to filter on. Either or both may be **null** to signal that no filtering on that attribute is desired.
```
empty array looks intuitively good for me but here we can get client implementations discrepancy

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
```diff
@@ -322,7 +357,7 @@ public Optional<SyncCommitteeContribution> createSyncCommitteeContribution(
             queryParams,
             EMPTY_MAP,
             createHandler(GetSyncCommitteeContributionResponse.class)
-                .withHandler(SC_NOT_FOUND, (request, response) -> Optional.empty()))
+                .withHandler(SC_NOT_FOUND, (req, res) -> Optional.empty()))
```
rolfyone: request response reads better...

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClient.java
```diff
@@ -127,8 +135,21 @@ public Optional<GetBlockHeaderResponse> getBlockHeader(final String blockId) {
...
+   * Tries <a
+   * href="https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/postStateValidators">POST
+   * Get validators from state</a> and falls back to <a
+   * href="https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getStateValidators">GET
+   * Get validators from state</a> if POST method is not allowed
+   */
   @Override
   public Optional<List<ValidatorResponse>> getValidators(final List<String> validatorIds) {
+    try {
+      return postValidators(validatorIds);
```
rolfyone: this is kind of the wrong place...
Basically getValidators has to split on id, and we don't have to with the post. It'd be better if we're determining to use post and calling for ALL keys in 1 call, rather than this approach, but it could potentially be another PR...
The other side of it is much like the SSZ concept, where if we get a post failure once, we should probably assume the endpoint doesnt support it and stop trying and failing every single time we request, and thats likely another good reason to do this either in a wrapping class, or higher up... the wrapper class could potentially encapsulate the splitting as required too...

zilm13: Great catch, on some test node with 5000 validators we could have 100 bad post calls and 100 good gets when it's not supported and 100 post calls instead of 1 when it's supported.

StefanBratanov: Got it. Working on the changes. :)

StefanBratanov:
> LGTM, but I'm concerned on the spec filters definition
> 
> ```
> The lists of validator IDs and statuses to filter on. Either or both may be **null** to signal that no filtering on that attribute is desired.
> ```
> 
> empty array looks intuitively good for me but here we can get client implementations discrepancy

good point, think setting to null is better

StefanBratanov:
> > LGTM, but I'm concerned on the spec filters definition
> > ```
> > The lists of validator IDs and statuses to filter on. Either or both may be **null** to signal that no filtering on that attribute is desired.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > empty array looks intuitively good for me but here we can get client implementations discrepancy
> 
> good point, think setting to null is better



> > LGTM, but I'm concerned on the spec filters definition
> > ```
> > The lists of validator IDs and statuses to filter on. Either or both may be **null** to signal that no filtering on that attribute is desired.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > empty array looks intuitively good for me but here we can get client implementations discrepancy
> 
> good point, think setting to null is better

Ok, actually just checked, we fail when trying for example request `{"ids":["1"],"statuses":null}` with `Array expected but got VALUE_NULL\n at [Source: REDACTED (StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION disabled); line: 5, column: 14]` so we may be against the spec. Opinions @rolfyone ?

StefanBratanov:
> > > LGTM, but I'm concerned on the spec filters definition
> > > ```
> > > The lists of validator IDs and statuses to filter on. Either or both may be **null** to signal that no filtering on that attribute is desired.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > empty array looks intuitively good for me but here we can get client implementations discrepancy
> > 
> > 
> > good point, think setting to null is better
> 
> > > LGTM, but I'm concerned on the spec filters definition
> > > ```
> > > The lists of validator IDs and statuses to filter on. Either or both may be **null** to signal that no filtering on that attribute is desired.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > empty array looks intuitively good for me but here we can get client implementations discrepancy
> > 
> > 
> > good point, think setting to null is better
> 
> Ok, actually just checked, we fail when trying for example request `{"ids":["1"],"statuses":null}` with `Array expected but got VALUE_NULL\n at [Source: REDACTED (StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION disabled); line: 5, column: 14]` so we may be against the spec. Opinions @rolfyone ? Reverted PR to be an empty list.

rolfyone:
Because Statuses is also an array, if you want to send a 'default' for statuses it isn't null. It's actually an empty array that you would send.

Having said that, because its more like query parameters, the better behaviour is to send just the fields you want to set, and let the server side default them correctly.

if you really wanted to you could do something like `{"ids":[], "statuses":[]}` to request all validator ids, but its much simpler to just send an empty body or `{}`, you'd never send the empty arrays like above...

So TL/DR, we just want to specify `{"ids":[IDS]}` 

We ideally want it without carriage returns - it sounds ugly but it can actually add a lot of size to the request when there's a lot of small numbers.  This would also be worth verifying so we're happy that the request string wont be huge for large requests. The numbers should also be strictly quoted, because many servers are treating them as strings, but teku is far less strict on this front.

validator/remote/src/integration-test/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClientTest.java
```diff
@@ -173,198 +152,46 @@ public void getValidators_WhenSuccess_ReturnsResponse() {
...
-
-    Optional<BeaconBlock> beaconBlock =
-        apiClient.createUnsignedBlock(slot, blsSignature, graffiti, false);
-
-    assertThat(beaconBlock).isPresent();
-    assertThat(beaconBlock.get()).isEqualTo(expectedBeaconBlock);
-  }
-
-  @Test
-  public void createUnsignedBlocks_shouldUseBlindedBlocksIfEnabled() {
```
rolfyone, line 239: I'm not sure why we'd be deleting tests in this PR...

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java
```diff
@@ -127,26 +128,49 @@ public SafeFuture<Map<BLSPublicKey, Integer>> getValidatorIndices(
...
+  private <T> Optional<Map<BLSPublicKey, T>> makeValidatorRequest(
+      final Collection<BLSPublicKey> publicKeys,
+      final Function<ValidatorResponse, T> valueExtractor) {
+    try {
+      return apiClient
+          .postValidators(convertPublicKeysToValidatorIds(publicKeys))
+          .map(responses -> convertToValidatorMap(responses, valueExtractor));
+    } catch (final PostStateValidatorsNotExistingException __) {
+      LOG.debug(
+          "POST method is not available for getting validators from state. Will use GET instead.");
```
rolfyone: so here, also remember that there was an exception, and never try again...

Basically each epoch we'd still attempt this post and waste time with the failure if we go this way, where as if we make our first call and that fails, then we move on to gets, the next time we can just do gets... it may just look something like
```
if (isPostWorking) {
    try {
      apiclient.postValidators(....
    } catch () {
      isPostWorking.setAtomicBoolean(false);
      LOG...
    }
}
return makeBatchedValidatorRequest...
```

StefanBratanov:
> Because Statuses is also an array, if you want to send a 'default' for statuses it isn't null. It's actually an empty array that you would send.
> 
> Having said that, because its more like query parameters, the better behaviour is to send just the fields you want to set, and let the server side default them correctly.
> 
> if you really wanted to you could do something like `{"ids":[], "statuses":[]}` to request all validator ids, but its much simpler to just send an empty body or `{}`, you'd never send the empty arrays like above...
> 
> So TL/DR, we just want to specify `{"ids":[IDS]}`
> 
> We ideally want it without carriage returns - it sounds ugly but it can actually add a lot of size to the request when there's a lot of small numbers. This would also be worth verifying so we're happy that the request string wont be huge for large requests. The numbers should also be strictly quoted, because many servers are treating them as strings, but teku is far less strict on this front.

I am not sure how other clients deal with missing fields though. I am afraid that they may fail and would prefer to send [] thus I am trying to be more cautious on this side, because the spec is not precise on that front.

**Update** Reverting on that view. The schema specifies the fields in the json as optional, so it's fine to send just **ids**

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java
StefanBratanov: Yeah that makes sense. Will implement.

validator/remote/src/integration-test/java/tech/pegasys/teku/validator/remote/apiclient/OkHttpValidatorRestApiClientTest.java
StefanBratanov: Just deleting couple of methods only used in this test in the `ValidatorRestApiClient`

rolfyone (approved):
Looks good! Thanks for your patience :)

PR #7788, opened 2023-12-05
Throttling inactivity leak log msg

### LINKED ISSUE #7329: Prevent inactivity leak info log to burst multiple messages
I noticed during some interop tests that it is possible that teku prints  several inactivity logs:

image

I restarted an interop network after keeping it down for a while. I think It processed several (empty) epochs in a row.

We should have logic to print this message within a certain interval to avoid spamming logs.

lucassaldanha:
## PR Description
- Using a Throttler to limit how often we print the inactivity leak warning. Under normal circumstances, we wouldn't need this, but when processing multiple empty slots (e.g. in devnets) it may spam messages in the log.

## Fixed Issue(s)
fixes #7329 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
@@ -125,7 +125,7 @@ public abstract class AbstractEpochProcessor implements EpochProcessor {
     processSyncCommitteeUpdates(state);
 
     if (beaconStateAccessors.isInactivityLeak(state)) {
-      loggerThrottler.invoke(currentEpoch, (log) -> log.info("Beacon chain is in activity leak"));
+      loggerThrottler.invoke(currentEpoch, (log) -> log.info("Beacon chain is in inactivity leak"));
     }
   }
 
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
@@ -26,6 +26,7 @@ import tech.pegasys.teku.infrastructure.ssz.SszMutableList;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszBitvector;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszMutableUInt64List;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszUInt64List;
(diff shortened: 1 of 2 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
```diff
@@ -122,7 +125,7 @@ protected void processEpoch(final BeaconState preState, final MutableBeaconState
     processSyncCommitteeUpdates(state);
 
     if (beaconStateAccessors.isInactivityLeak(state)) {
-      LOG.info("Beacon chain is in activity leak");
+      loggerThrottler.invoke(currentEpoch, (log) -> log.info("Beacon chain is in activity leak"));
```
rolfyone: tests or it didnt happen...

lucassaldanha: ouch forgot to add the file to the commit

rolfyone: i'd be happy with just a test showing a logThrottler in use with this mechanism.. doesnt have to be in-situe testing...

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessorTest.java
```diff
@@ -0,0 +1,110 @@
...
+  @Test
+  public void shouldThrottleInactivityLeakLogs() throws Exception {
+    final int throttlingPeriod = 1; // expect maximum of one call per epoch
+    final Logger logger = mock(Logger.class);
+    final Throttler<Logger> loggerThrottler = spyLogThrottler(logger, throttlingPeriod);
+
+    final BeaconState state = createStateInInactivityLeak();
+    final UInt64 currentEpoch = spec.getCurrentEpoch(state);
+
+    final int slotsPerEpoch = spec.getSlotsPerEpoch(state.getSlot());
```
rolfyone: i'd probably suggest moving these to member variables as they're basically intiialization.

The result would be we just have the actual test case in here, and it'll be a lot clearer.

rolfyone (approved):
one nit, overall LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
benjaminion: > Beacon chain is in activity leak

Should be "Beacon chain is in inactivity leak" (needs another "in").

PR #7792, opened 2023-12-06
Support null in JSON parsing as an empty array

zilm13:
## PR Description
Compliment to #7786 
`GetSyncCommitteeRewardsIntegrationTest`:
Test removed and logic extended because it's per spec, see https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/rewards/sync_committee.yaml#L15-L17
It's safe to test no currentToken because if we are inside some object we will throw on the next step not able to find closing, if the whole empty object is array type, it's legit to be ""
`PostAttesterDutiesIntegrationTest`:
Spec requires at least one index https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/duties/attester.yaml#L33-L39 which was not checked. Though the field is required by spec, I think it's acceptable to fail a bit afterwards

## Documentation

## Changelog

### DIFF
--- a/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableArrayTypeDefinition.java
+++ b/infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableArrayTypeDefinition.java
@@ -13,11 +13,14 @@
 
 package tech.pegasys.teku.infrastructure.json.types;
 
+import static com.fasterxml.jackson.core.JsonToken.VALUE_NULL;
+
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.exc.MismatchedInputException;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.Function;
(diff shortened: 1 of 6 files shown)


rolfyone:
I'm not sure this is syntactically correct... I think we park this, we don't need to allow the scenarios presented.

zilm13:
To be scrapped to better parts

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableArrayTypeDefinition.java
```diff
@@ -49,6 +52,9 @@ public DeserializableArrayTypeDefinition(
   @Override
   public CollectionT deserialize(final JsonParser parser) throws IOException {
     if (!parser.isExpectedStartArrayToken()) {
+      if (!parser.hasCurrentToken() || parser.currentToken().equals(VALUE_NULL)) {
+        return createFromList.apply(Collections.emptyList());
+      }
```
rolfyone, line 57: if we break out this section into a PR, we can probably merge that with any associated testing - it'd make our parser a little more tolerant, but only affect typedefs, so still wouldn't help the VC a whole lot.
Probably fair to be slightly tolerant on this specific scenario, although some interfaces (not ours) do use `null`distinct from `[]` so if that scenario ever does come along it'd be nice for a test to break so we can dig into this change...

zilm13:
scrapped  by #7801 and #7795

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableArrayTypeDefinition.java
zilm13: `parser.hasCurrentToken()` is covered more correctly with #7801
`parser.currentToken().equals(VALUE_NULL)` could be added later if someone opens an issue

PR #7793, opened 2023-12-06
Add DepositTreeSnapshot download to checkpoint sync

### LINKED ISSUE #7715: Add DepositTreeSnapshot download to checkpoint sync
### Description
Most checkpoints supports `DepositTreeSnapshot` API https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/deposit_snapshot.yaml, so when user provides `--checkpoint-sync-url,` we can try to download `DepositTreeSnapshot` from the same provider and start node with fresh snapshot.
Please, note, currently we support only ssz input so it should be updated to be able to parse json also. There is no need to temporarily save a file, `DepositSnapshotFileLoader` supports URLs.

**Reopened 15/12/2023**

Reopened as we need to implement this again, this time with a proper fallback mechanism when we can't load it from the Beacon API.

**Desired behaviour**

**Load from checkpoint-sync-url**
- Likely to be the most used option by users
- Only used when checkpoint-sync-url option is in use
- Not reliable, should fallback into bundled option (or stored db) when it fails. Should not fail if can’t load.
- Should be the first choice when available

**Load from bundled snapshot (current default behaviour)**
- The second most common option now that users are likely to start using checkpoint-sync-url
- Only available on a compatible network (network that we provide a bundle for)
- There is no reason for this to fail (it is provided as part of our JAR file). But if it fails, we should log a warning (?) and fallback into our stored snapshot in the db (even if it is an empty snapshot tree).

**Load from provided path/URL (--Xdeposit-snapshot)**
- Less common option, probably only used for testnets etc.
- Mandatory when specified
- Don’t try any other fallback options

courtneyeh:
fixes #7715 

## Documentation

## Changelog

### DIFF
--- a/services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
+++ b/services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
@@ -22,6 +22,7 @@ import java.util.Optional;
 import org.apache.commons.lang3.StringUtils;
 import tech.pegasys.teku.ethereum.execution.types.Eth1Address;
 import tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException;
+import tech.pegasys.teku.infrastructure.http.UrlSanitizer;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.networks.Eth2Network;
@@ -30,6 +31,7 @@ public class PowchainConfiguration {
   public static final int DEFAULT_ETH1_LOGS_MAX_BLOCK_RANGE = 10_000;
   public static final boolean DEFAULT_USE_MISSING_DEPOSIT_EVENT_LOGGING = false;
   public static final boolean DEFAULT_DEPOSIT_SNAPSHOT_ENABLED = true;
+  public static final String DEPOSIT_SNAPSHOT_URL_PATH = "/eth/v1/beacon/deposit_snapshot";
 
   private final Spec spec;
   private final List<String> eth1Endpoints;
@@ -176,6 +178,14 @@ public class PowchainConfiguration {
     public Builder depositSnapshotPath(final String depositSnapshotPath) {
       if (StringUtils.isNotBlank(depositSnapshotPath)) {
         this.depositSnapshotPath = Optional.of(depositSnapshotPath);
+      }
+      return this;
+    }
+
+    public Builder depositSnapshotPathFromCheckpointSyncUrl(final String checkpointSyncUrl) {
+      if (StringUtils.isNotBlank(checkpointSyncUrl)) {
+        this.depositSnapshotPath =
+            Optional.of(UrlSanitizer.appendPath(checkpointSyncUrl, DEPOSIT_SNAPSHOT_URL_PATH));
       }
       return this;
     }
--- a/services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
+++ b/services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
@@ -188,8 +188,7 @@ public class PowchainConfiguration {
             Optional.of(UrlSanitizer.appendPath(checkpointSyncUrl, DEPOSIT_SNAPSHOT_URL_PATH));
         return this;
       }
-      throw new IllegalStateException(
-          "--checkpoint-sync-url option is empty, but needed for deposit snapshot path");
+      throw new IllegalStateException("checkpointSyncUrl cannot be null/empty");
     }
 
     public Builder setDepositSnapshotPathForNetwork(final Optional<Eth2Network> eth2Network) {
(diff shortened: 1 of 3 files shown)


services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
```diff
@@ -180,6 +182,14 @@ public Builder depositSnapshotPath(final String depositSnapshotPath) {
       return this;
     }
 
+    public Builder depositSnapshotPathFromCheckpointSyncUrl(final String checkpointSyncUrl) {
+      if (StringUtils.isNotBlank(checkpointSyncUrl)) {
+        this.depositSnapshotPath =
+            Optional.of(UrlSanitizer.appendPath(checkpointSyncUrl, DEPOSIT_SNAPSHOT_URL_PATH));
+      }
```
lucassaldanha: In this case, we should throw an error. Otherwise, it will silently not use the checkpoint sync url.

teku/src/test/java/tech/pegasys/teku/cli/options/DepositOptionsTest.java
```diff
@@ -148,4 +149,80 @@ public void shouldThrowErrorIfSnapshotPathProvidedWithBundleSnapshot() {
...
+  public void shouldSetDefaultSnapshotConfig() {
+    final String[] args = {"--network=mainnet"};
+    final PowchainConfiguration config = getTekuConfigurationFromArguments(args).powchain();
+    assertThat(config.getDepositSnapshotPath())
+        .contains(
+            PowchainConfiguration.class
+                .getResource(
+                    DEFAULT_SNAPSHOT_RESOURCE_PATHS.get(
+                        Eth2Network.fromStringLenient("mainnet").get()))
+                .toExternalForm());
```
lucassaldanha: This assertion is used at least twice in the tests. It would improve readability if we moved it into a private method called `assertDepositSnapshotPathMatchesBundledDepositTreeSnapshot()` or something similar.

services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
```diff
@@ -180,6 +182,16 @@ public Builder depositSnapshotPath(final String depositSnapshotPath) {
       return this;
     }
 
+    public Builder depositSnapshotPathFromCheckpointSyncUrl(final String checkpointSyncUrl) {
+      if (StringUtils.isNotBlank(checkpointSyncUrl)) {
+        this.depositSnapshotPath =
+            Optional.of(UrlSanitizer.appendPath(checkpointSyncUrl, DEPOSIT_SNAPSHOT_URL_PATH));
+        return this;
+      }
+      throw new IllegalStateException(
```
lucassaldanha: This should be a `InvalidConfigurationException`

teku/src/test/java/tech/pegasys/teku/cli/options/DepositOptionsTest.java
```diff
@@ -148,4 +147,51 @@ public void shouldThrowErrorIfSnapshotPathProvidedWithBundleSnapshot() {
...
+      "--deposit-snapshot-enabled=false",
+      "--checkpoint-sync-url=http://checkpoint/path/"
+    };
+    final String[] depositSnapshotEnabledFalseAndSync = {
+      "--deposit-snapshot-enabled=false", "--checkpoint-sync-url=http://checkpoint/path/"
+    };
+    final String[] checkpointSyncUrlSet = {"--checkpoint-sync-url=http://checkpoint/path/"};
+
+    return Stream.of(
+        Arguments.of(emptyConfig, getDefaultBundleSnapshotPath("mainnet"), true),
```
lucassaldanha: `getDefaultBundleSnapshotPath("mainnet")` can be assigned to a local variable so we don't have to do multiple calls to `getResource()` executing this method.

services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
courtneyeh: But there should be no configuration that would allow that? Since I did checks before using the `depositSnapshotPathFromCheckpointSyncUrl` method

lucassaldanha: If this is the case, the reason for the exception is wrong. Do not mention the cli option, just the parameter name:

e.g. `checkpointSyncUrl cannot be null/empty`.

teku/src/test/java/tech/pegasys/teku/cli/options/DepositOptionsTest.java
```diff
@@ -148,4 +147,52 @@ public void shouldThrowErrorIfSnapshotPathProvidedWithBundleSnapshot() {
...
+      "--deposit-snapshot-enabled=true",
+      "--checkpoint-sync-url=http://checkpoint/path/"
+    };
+    final String[] depositSnapshotPathSet = {"--Xdeposit-snapshot=/some/path/"};
+    final String[] depositSnapshotPathSetAndCheckpointSync = {
+      "--Xdeposit-snapshot=/some/path/",
+      "--deposit-snapshot-enabled=false",
+      "--checkpoint-sync-url=http://checkpoint/path/"
+    };
+    final String[] depositSnapshotEnabledFalseAndSync = {
```
lucassaldanha: NIT: Maybe we should name this `depositSnapshotDisabledAndSync`

`EnabledFalse` kinda twisted my brain :P

lucassaldanha (approved):
LGTM

PR #7794, opened 2023-12-07
slot based duty testing

rolfyone:
Computing aggregation duties is an expensive operation.

Trying to see why every reorg we seem to recompute so many duties.

## Documentation

## Changelog


tbenr (approved):
LGTM. in case you want this to be merged

PR #7795, opened 2023-12-07
Fix PostAttesterDuties to be compliant with spec (min 1 index)

zilm13:
## PR Description
Spec requires at least one index https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/duties/attester.yaml#L33-L39 which was not checked. Check added + test.

## Documentation

## Changelog


zilm13:
I see there are 4 apis with `minItems: 1`, so I will try to implement something more generic

zilm13:
So more generic definition was made and following endpoints fixed:
https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/duties/sync.yaml#L15-L23
https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/duties/attester.yaml#L31-L39
https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/liveness.yaml#L22-L30

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableArrayTypeDefinition.java
```diff
@@ -28,13 +29,25 @@ public class DeserializableArrayTypeDefinition<ItemT, CollectionT extends Iterab
 
   private final DeserializableTypeDefinition<ItemT> itemType;
   private final Function<List<ItemT>, CollectionT> createFromList;
+  private final Optional<Integer> minItems;
```
rolfyone, line 32: we should be expanding on `DeserializableArrayTypeDefinitionTest`

rolfyone (commented):
overall looks good. Need to expand typeDef tests a little, but shouldn't be too much of a problem.

zilm13:
Added tests

rolfyone (approved):
LGTM

PR #7798, opened 2023-12-07
Support Basic Auth credentials for external signer

### LINKED ISSUE #7796: Support Basic Auth credentials in Web3signer URL
### Description
currently using Web3signer with TLS + http basic auth. This work fine with Lighthouse and Prysm when using an URL format of https://user:pass@web3signer_url but in TEKU it seems that the HTTP lib used does not support passing the basic auth credentials. 

```
2023-12-07 09:10:36.137 ERROR - External signer is currently not reachable at https://w3s:PASSWORD@web3signer.sepolia.domain.com
```

When logged into the pod I can query it fine.

Is there any way to support this easily in TEKU ? Or am I missing something ? Right now I'm passing this as en ENV var, I could try via config or args but not sure it would change anything.

### Steps to Reproduce (Bug)
Set the env vars TEKU_VALIDATORS_EXTERNAL_SIGNER_URL.

Connection is successful

**Actual behavior:** 

Unable to connect

**Frequency:** 

Constant

* Software version: 23.08
* Java version: jdk17
* OS Name & Version: Ubuntu 22.04 but running in Kubernetes via Docker image
* Docker Version: No docker but containerd
* Cloud VM, type, size: Bare metal

StefanBratanov:
## PR Description

- Modify `HttpClientExternalSignerFactory` to setup Basic authentication for the `HttpClient`
- Added integration test. Created `AbstractSecureExternalSignerIntegrationTest` which is also used by the TlsUpcheckTest
- Added check in `ValidatorConfig.validateExternalSignerURLScheme`
- I had problems running integration tests on Windows (even before changes) and defining `bcpkix-jdk18on` seems to fix it. It has to do with the mockserver usage of BouncyCastle
- Creating `validatorExternalSignerUserInfo` in ValidatorConfig to use in the `HttpClientExternalSignerFactory`

## Fixed Issue(s)
fixes #7796 

## Documentation

## Changelog

### DIFF
--- a/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
+++ b/validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
@@ -66,7 +66,7 @@ public class ValidatorConfig {
   private final List<String> validatorExternalSignerPublicKeySources;
   private final boolean validatorExternalSignerSlashingProtectionEnabled;
   private final URL validatorExternalSignerUrl;
-  private final URL unsanitizedValidatorExternalSignerUrl;
+  private final Optional<String> validatorExternalSignerUserInfo;
   private final Duration validatorExternalSignerTimeout;
   private final Path validatorExternalSignerKeystore;
   private final Path validatorExternalSignerKeystorePasswordFile;
@@ -102,7 +102,7 @@ public class ValidatorConfig {
       final List<String> validatorKeys,
       final List<String> validatorExternalSignerPublicKeySources,
       final URL validatorExternalSignerUrl,
-      final URL unsanitizedValidatorExternalSignerUrl,
+      final Optional<String> validatorExternalSignerUserInfo,
       final Duration validatorExternalSignerTimeout,
       final Path validatorExternalSignerKeystore,
       final Path validatorExternalSignerKeystorePasswordFile,
@@ -136,7 +136,7 @@ public class ValidatorConfig {
     this.validatorKeys = validatorKeys;
     this.validatorExternalSignerPublicKeySources = validatorExternalSignerPublicKeySources;
     this.validatorExternalSignerUrl = validatorExternalSignerUrl;
-    this.unsanitizedValidatorExternalSignerUrl = unsanitizedValidatorExternalSignerUrl;
+    this.validatorExternalSignerUserInfo = validatorExternalSignerUserInfo;
     this.validatorExternalSignerTimeout = validatorExternalSignerTimeout;
     this.validatorExternalSignerKeystore = validatorExternalSignerKeystore;
     this.validatorExternalSignerKeystorePasswordFile = validatorExternalSignerKeystorePasswordFile;
@@ -196,8 +196,8 @@ public class ValidatorConfig {
     return validatorExternalSignerUrl;
   }
 
-  public URL getUnsanitizedValidatorExternalSignerUrl() {
-    return unsanitizedValidatorExternalSignerUrl;
+  public Optional<String> getValidatorExternalSignerUserInfo() {
+    return validatorExternalSignerUserInfo;
   }
 
   public Duration getValidatorExternalSignerTimeout() {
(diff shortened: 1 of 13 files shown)


tbenr:
I'd also enforce that `https` schema is required if basic auth is enabled. Otherwise we could end up sending credentials on non-encrypted transport.
Apart of that, LGTM

rolfyone (approved):
Make sure we're not logging the un-sanitized url anywhere on startup, it'll need to be the url without the password, or with the user/pass redacted...
Other than that, if its limited to https as suggested it's adequate for people wanting to use basic auth...
It's not a configuration I'd recommend, but it's not such a problem that i'd block implementing the option.

tbenr:
It should already safe, but if we could extract basic auth and do url sanitization at CLI option handling level, it is even safer. And https check would be a validation step there.

I'd leave a `checkArgument` here on https+basic auth just in case we will use the client in other ways.

StefanBratanov:
Actually modified the PR quite a bit, because the `HttpClientExternalSignerFactory` is used in other places apart from the ExternalSigner, so sending the `Authorization` header only in the ExternalSigner is not enough for all use cases.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/HttpClientExternalSignerFactory.java
```diff
@@ -58,9 +68,22 @@ public HttpClient get() {
...
-    final String protocol = validatorConfig.getValidatorExternalSignerUrl().getProtocol();
-    return protocol != null && protocol.equalsIgnoreCase("https");
+  private boolean isTLSEnabled(final URL externalSignerUrl) {
+    return "https".equalsIgnoreCase(externalSignerUrl.getProtocol());
+  }
+
+  private void configureBasicAuthentication(
+      final HttpClient.Builder builder, final String userInfo) {
+    final List<String> authCredentials = Splitter.on(':').splitToList(userInfo);
+    final String username = authCredentials.get(0);
```
tbenr, line 79: This breaks if userInfo doesn't contain `:` you should check authCredentials size

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
```diff
@@ -679,11 +692,13 @@ private void validateExternalSignerURLScheme() {
         return;
       }
 
-      if (validatorExternalSignerKeystore != null || validatorExternalSignerTruststore != null) {
+      if (validatorExternalSignerKeystore != null
+          || validatorExternalSignerTruststore != null
+          || unsanitizedValidatorExternalSignerUrl.getUserInfo() != null) {
```
tbenr: Not a fan of having a unsanitized copy here... can't we simply expose the sanitized url and Optional userinfo (BasicAuth) as a separate config?

tbenr (approved):
LGTM. a couple of suggestions

validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorConfig.java
StefanBratanov: Yeah, I agree. I changed it to expose the `userInfo` as an optional.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/HttpClientExternalSignerFactory.java
StefanBratanov: I am not sure about that. Isn't it standard to use the `user:password` format?

tbenr: But you can always pass an url without ':' and seems like URL constructor doesn't complain. (I tried in test)

StefanBratanov: Yeah, get it. Added an exception for that case.

PR #7799, opened 2023-12-07
add block/selector root selector factories

### LINKED ISSUE #5870: Avoid loading full state/block for API requests for root
### Description
Requests to `/eth/v1/beacon/states/:stateId/root` or `/eth/v1/beacon/blocks/:blockId/root` should try to avoid loading the full state or block to calculate the root from. Generally the root is directly available from the database or protoarray.

mehdi-aouadi:
## PR Description
Avoid loading full block/state when requesting block/state root:
Add `BlockRootSelectorFactory` and  `StateRootSelectorFactory` which get the block/state root directly when possible and load the full block/state otherwise

## Fixed Issue(s)

fixes #5870 

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/blockrootselector/BlockRootSelectorFactory.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/blockrootselector/BlockRootSelectorFactory.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright Consensys Software Inc., 2023
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.api.blockrootselector;
+
+import static tech.pegasys.teku.spec.config.SpecConfig.GENESIS_SLOT;
+
+import java.util.Optional;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.api.AbstractSelectorFactory;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.Spec;
+import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
+import tech.pegasys.teku.spec.datastructures.metadata.ObjectAndMetaData;
+import tech.pegasys.teku.storage.client.ChainHead;
+import tech.pegasys.teku.storage.client.CombinedChainDataClient;
+
+public class BlockRootSelectorFactory extends AbstractSelectorFactory<BlockRootSelector> {
+
+  private final Spec spec;
+
+  public BlockRootSelectorFactory(final Spec spec, CombinedChainDataClient client) {
+    super(client);
+    this.spec = spec;
+  }
+
+  @Override
+  public BlockRootSelector blockRootSelector(final Bytes32 blockRoot) {
+    return () -> client.getBlockByBlockRoot(blockRoot).thenApply(this::fromBlock);
+  }
+
+  @Override
+  public BlockRootSelector headSelector() {
+    return () ->
+        client
+            .getChainHead()
+            .map(this::fromChainHead)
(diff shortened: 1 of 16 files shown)


mehdi-aouadi:
Not sure if the small optimisation is worth it since we only avoid loading the full block when querying by slot... We are still obliged to load the full block otherwise

data/provider/src/main/java/tech/pegasys/teku/api/blockrootselector/BlockRootSelectorFactory.java
```diff
@@ -0,0 +1,137 @@
...
+            .getChainHead()
+            .map(this::fromChainHead)
+            .orElse(SafeFuture.completedFuture(Optional.empty()));
+  }
+
+  @Override
+  public BlockRootSelector finalizedSelector() {
+    return () ->
+        SafeFuture.completedFuture(
+            lookupBlockRootData(client.getFinalizedBlock(), client.isChainHeadOptimistic()));
```
rolfyone: we may be able to be more tricky here...
Rather than getting the finalized block, theres a finalized Checkpoint on recentChainData, so we might be able to make a combined client fn that does something like 
```
recentChainData.getFinalizedCheckpoint().orElseThrow().getRoot()
```
will be worth investigating (im not actually suggesting using orElseThrow, but its an optional so we'd need to deal with that correctly)

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
```diff
@@ -624,6 +637,14 @@ public Optional<SignedBeaconBlock> getFinalizedBlock() {
     return getStore().getLatestFinalized().getSignedBeaconBlock();
   }
 
+  public Optional<Bytes32> getFinalizedBlockHashTreeRoot() {
```
rolfyone: can we just call this `getFinalizedBlockRoot`

data/provider/src/main/java/tech/pegasys/teku/api/blockrootselector/BlockRootSelectorFactory.java
rolfyone: it looks like you've done the fn below just not called it...

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
mehdi-aouadi: done

data/provider/src/main/java/tech/pegasys/teku/api/blockrootselector/BlockRootSelectorFactory.java
mehdi-aouadi: Yep, my bad, I've already added it. Fixed

data/provider/src/main/java/tech/pegasys/teku/api/staterootselector/StateRootSelectorFactory.java
```diff
@@ -0,0 +1,155 @@
...
+  private final Spec spec;
+
+  public StateRootSelectorFactory(final Spec spec, final CombinedChainDataClient client) {
+    super(client);
+    this.spec = spec;
+  }
+
+  @Override
+  public StateRootSelector stateRootSelector(final Bytes32 stateRoot) {
+    return () -> client.getStateByBlockRoot(stateRoot).thenApply(this::fromState);
```
rolfyone: im confused, is the input to this function a state root or a block root? we've got it getting passed to byblockRoot, which kind of suggests the param is a blockRoot...

rolfyone: I'm also not sure we're solving the actual problem in the ticket here, which suggested we could avoid loading the state ,but this still loads state to (compute) get the state root...

mehdi-aouadi: My bad for `getStateByBlockRoot`. Replaced it by `getStateByStateRoot`

mehdi-aouadi: Actually as mentioned in the ticket we don't have a slot to state root mapping for the finalized states. I added an optimisation for the `slotSelector` to try to retrieve the state root from the proto array before hitting the database. That's the only optimisation possible for state root requests as far as I can see

rolfyone:
> Not sure if the small optimisation is worth it since we only avoid loading the full block when querying by slot... We are still obliged to load the full block otherwise

I'm not sure any of this is worth while to be honest, im wondering if theres a measurable difference at all, or if we're just shuffling chairs about in reality... 

the protoarray change in particular i could see going either way...

mehdi-aouadi:
The optimisation isn't really measurable if existing at all

PR #7801, opened 2023-12-07
Optional request body in API + usage clean

zilm13:
## PR Description
Used in 
- https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/states/validator_balances.yaml#L82-L84
- https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/rewards/attestations.yaml#L16-L18
- https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/rewards/sync_committee.yaml#L15-L17

## Documentation

## Changelog

### DIFF
--- a/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
+++ b/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
@@ -47,6 +47,10 @@ public class JavalinRestApiRequest implements RestApiRequest {
   @Override
   @SuppressWarnings({"TypeParameterUnusedInFormals"})
   public <T> T getRequestBody() throws JsonProcessingException {
+    if (!metadata.isRequiredRequestBody()) {
+      throw new RuntimeException(
+          "Optional request body configured, use getOptionalRequestBody() instead");
+    }
     return metadata.getRequestBody(context.bodyInputStream(), context.headerMap());
   }
 
--- a/data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/rewards/GetSyncCommitteeRewardsIntegrationTest.java
+++ b/data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v1/rewards/GetSyncCommitteeRewardsIntegrationTest.java
@@ -62,7 +62,7 @@ public class GetSyncCommitteeRewardsIntegrationTest
   @Test
   public void handleEmptyRequestBodyList() throws IOException {
     final List<String> requestBody = List.of();
-    Response response =
+    final Response response1 =
         post(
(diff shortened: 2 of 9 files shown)


infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
```diff
@@ -47,6 +47,10 @@ public class JavalinRestApiRequest implements RestApiRequest {
   @Override
   @SuppressWarnings({"TypeParameterUnusedInFormals"})
   public <T> T getRequestBody() throws JsonProcessingException {
+    if (!metadata.isRequiredRequestBody()) {
+      throw new RuntimeException(
+          "Optional request body configured, use getOptionalRequestBody() instead");
+    }
```
rolfyone, line 53: this will end up a 500 response...

lucassaldanha (dismissed):

lucassaldanha (commented):
LGTM. Just need to look at Paul's comment re the error returned

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
lucassaldanha: Now that I thought about it, I don't know if enforcing this here makes sense. This error will happen when the application is running and the user has nothing to do with what method is being used here (the developer does).

I believe the right thing to do to enforce the usage of required/optional would be a compile-time warning. This way the developer can ensure they use the right method.

zilm13: My point here was to catch something in the first ever test you write if you are using getRequestBody without optional. There is a risk you didn't write any test, yes. I'm not sure

lucassaldanha: Fair point! We just need to make sure this error message does not make its way to the user when calling the API 👍

lucassaldanha (approved):
LGTM (again) ;)

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
zilm13: So my intention was that only developer will see it. It could go outside if someone makes 0 tests on new endpoint and we merge it. Your decision. I could remove it. I think we always make tests.

lucassaldanha: Yeah, that's what I ended up settling on... If we have a new RPC with zero tests in prod we will probably have bigger problems than the error message to worry about... :P

PR #7808, opened 2023-12-10
Adding support for loading JSON DepositTreeSnapshot

### LINKED ISSUE #7806: Deposit snapshot feature doesn't support some of the public endpoints
### Description
A follow up to #7793 
Deposit snapshot loading feature in Teku consists of 2 steps
1. Loading. If it's local file it's just read, if it's url we try to download file with header requesting `ssz` by setting type `application/octet-stream`. Support of ssz type is a part of the spec, see: https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/deposit_snapshot.yaml
2. Reading file. As we download and store it as ssz we expect to read only this type, so we read it as ssz.

Both steps are happen in `DepositSnapshotFileLoader` and few other classes it uses.

The issue is that since we start to use this feature along with `--checkpoint-sync-url` we are no longer run it only in Teku environment, most public endpoints use another CL-clients which doesn't support ssz output for `DepositTreeSnapshot`. I have checked following public endpoints:
- https://beaconstate.ethstaker.cc
- https://mainnet.checkpoint.sigp.io
- https://sync-mainnet.beaconcha.in

All 3 endpoints do not support ssz `application/octet-stream`, only json type is supported.

So to make our feature work as intended, we should add a fallback to json type. It could be done in 2 ways:
1. Simple. In [catch block](https://github.com/ConsenSys/teku/blob/cb15faa048e5330a1cda312953fd003f6369310a/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java#L68) of `DepositSnapshotFileLoader` add call to something like loadJsonFromUrl, which should look like (trying to download as json, read as json):
```
  private DepositTreeSnapshot loadFromJsonUrl(final String path) throws IOException {
    final Bytes snapshotData =
        ResourceLoader.urlOrFile()
            .loadBytes(path)
            .orElseThrow(
                () -> new FileNotFoundException(String.format("File '%s' not found", path)));

    return JsonUtil.parse(new String(snapshotData.toArray(), StandardCharsets.UTF_8), DepositTreeSnapshot.getJsonTypeDefinition());
  }
```
Pros of this approach is that we are not only try to download it as json, we will try to read it as json even if local file is provided. Also another `catch (IOException e)` block should be added on failover retry.
2. Read the answer from url and if it has error code 415 `{"code":415,"message":"unsupported content-type: application/octet-stream"}`, try failover. It could be implemented a bit cleaner than 1 but we will get json support only for urls, not for local files 

Also we may consider not exiting if deposit snapshot is not loaded, we have bundled snapshot, it's not a critical failure, but with failover to json implemented I think we are ok to exit in case of the failure, it should be very rare.

### Steps to Reproduce (Bug)
```
$ ./teku [...skipped...] --checkpoint-sync-url=https://beaconstate.ethstaker.cc

[...skipped...]
17:03:02.245 INFO  - Loading deposit tree snapshot from https://beaconstate.ethstaker.cc/eth/v1/beacon/deposit_snapshot
17:03:02.249 INFO  - Syncing     *** Target slot: 7934713, Head slot: 7934592, Remaining slots: 121, Connected peers: 0
17:03:02.253 ERROR - Execution Client request failed. Make sure the Execution Client is online and can respond to requests.
Failed to load deposit tree snapshot from https://beaconstate.ethstaker.cc/eth/v1/beacon/deposit_snapshot: File 'https://beaconstate.ethstaker.cc/eth/v1/beacon/deposit_snapshot' not found

To display full help:
teku [COMMAND] --help

> Task :teku:Teku.main() FAILED
```

lucassaldanha:
## PR Description
This change adds a fallback option of trying to deserialize the bytes for a deposit tree snapshot as json when SSZ deserialization fails.

There is an outstanding question about using the internal bundled deposit tree snapshot as a fallback when loading from file/URL fails. But this can be addressed as a separate PR as it will likely require extra changes.

## Fixed Issue(s)
fixes #7806 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7816, opened 2023-12-12
Add `id` claim to the execution engine JWT payload

### LINKED ISSUE #7567: Feature request: IDs for JWT claims
### Description

The execution API spec defines an optional `id` claim that can be set in a JWT token to convey information about the client being authenticated. This is useful for more advanced forms of authentication beyond the basic 1-CL-to-1-EL authentication that we usually do. For example, my [_Electric Eel_](https://github.com/sigp/eleel) multiplexer uses the `id` to identify the connecting client and decide which JWT secret to use. Without the ID, it has to fall back on iterating all known JWT secrets and trying to verify the token against them (which is slow).

The spec defines the `id` claim here: https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#jwt-claims

It's worth noting that the ID in the claim is _distinct_ from the key-ID concept defined by the JWT standard, which is part of the _header_, not the body. Apart from being a bit inelegant this isn't an issue.

## Steps to resolve

Add a flag to Teku like `--ee-jwt-secret-id` which takes a string to use as the JWT ID for claims made to the EL.

### Versions

* Software version: Teku v23.9.1

StefanBratanov:
## PR Description
Adds `id` to the JWT payload for the execution engine.

Tested with `ee-jwt-claim-id: foobar` and the decoded JWT payload looked like:

```json
{
  "iat": 1702385661,
  "id": "foobar"
}
```

## Fixed Issue(s)
fixes #7567 

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/auth/JwtConfig.java
+++ b/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/auth/JwtConfig.java
@@ -32,21 +32,30 @@ public class JwtConfig {
   public static final long EXPIRES_IN_SECONDS = TOLERANCE_IN_SECONDS - 2;
 
   private final SecretKey key;
+  private final Optional<String> secretId;
 
-  public JwtConfig(final SecretKey key) {
+  public JwtConfig(final SecretKey key, final Optional<String> secretId) {
     this.key = key;
+    this.secretId = secretId;
   }
 
   public SecretKey getKey() {
     return key;
   }
 
+  public Optional<String> getSecretId() {
+    return secretId;
+  }
+
   public static Optional<JwtConfig> createIfNeeded(
-      final boolean needed, final Optional<String> jwtSecretFile, final Path beaconDataDirectory) {
+      final boolean needed,
+      final Optional<String> jwtSecretFile,
+      final Optional<String> jwtSecretId,
+      final Path beaconDataDirectory) {
     if (needed) {
       final JwtSecretKeyLoader keyLoader =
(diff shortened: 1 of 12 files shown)


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/auth/JwtConfig.java
```diff
@@ -32,21 +32,30 @@ public class JwtConfig {
   public static final long EXPIRES_IN_SECONDS = TOLERANCE_IN_SECONDS - 2;
 
   private final SecretKey key;
+  private final Optional<String> secretId;
```
tbenr: I think the variable name should be `id` and not `secretId`. Or `idClaim` if you want a longer\clearer one.

tbenr: and infact I think that the param itself should be `--ee-jwt-id` (or with a claim too)

lucassaldanha: I think `id` is too generic. I like `claimId` :)

StefanBratanov: Agree, I like `claimId` as well. Thanks. Pushed change to reflect that.

tbenr (approved):
LGTM. I'd add a test in `ExecutionLayerOptionsTest` too

Issue #7663, opened 2023-11-06
Attestation duty scheduling


rolfyone:
Attestation duty scheduling has 2 aspects of interest.

One part is scheduling the attestation itself, which is a relatively small impact, but the other is determining whether a validator needs to aggregate at a given slot.

Determining if an attester is an aggregator requires a signed object to be created, and we do this asynchronously.  We also process (worst case) current AND next epoch, due to lookahead, and it all takes place in the epoch transition slot (which is already the busiest slot of the epoch).

It'd be good to spread this load a little. In the first instance, we could consider a simple solution.
in the first epoch slot, ensure that we have duties scheduled already for this epoch.
If it's not the first slot, then as long as we have duties for current epoch, ensure we have duties scheduled for the next epoch.

This relatively simple solution should distribute the load of async signing tasks in the executor queue, because we're no longer performing N x 2 signatures (worst case) in the epoch transition slot, we're only doing N signatures, and the second slot would then need to compute for next epoch.

We'd want to consider a boundary condition, where a reorg happens in the last slot of an epoch, and then we'd actually want to consider duties for the following epoch as a priority, otherwise we'd possibly not perform duties in that case.

Below a certain number of validators, this current logic is probably fine, it's just at scale - we could consider a feature toggle (supported and documented) that only is used by larger instances, as the default current logic is often more than adequate.

tbenr:
Seems related to #7465 and #7466

rolfyone:
Im closing this for now, i think the local slashing fix has resolved the worst performance bottleneck in this flow, can revisit later.

PR #7819, opened 2023-12-13
Removed synchronized section in localSlashingProtector

rolfyone:
The idea with this change is to reduce contention on the synchronized functions, which means we need to introduce more locking, but this will allow better resource usage.  Files etc are still written synchronously, but we'll now be able to check more than one signing at a time, as long as they're distinct validators.

This is behind a development flag and off by default (the synchronized function is still on by default).

Roughly related to #7663

## Documentation

## Changelog


tbenr (approved):
LGTM. Sorry I ended up pushing a commit. just a nit on testing and missing finals

PR #7825, opened 2023-12-13
Allow deferred blinded-unblinded schema selection in block body builder

### LINKED ISSUE #7835: Allow block creation to defer blinded/unblinded decision

tbenr:
With this PR `BeaconBlockBodyBuilder` doesn't need to be set for blinded or unblinded flow in advance. It is created with both schemas and behaviour changes based on the payload is provided. In `build()` If header has been provided, it will go for blinded block body. If full payload has been provided, unblinded block is created. So it won't expose `isBlinded` method any more.

The blinded-unblinded `BeaconBlock` will be chosen based on the "blindess" of the `BeaconBlockBody`

This PR do not change the fact that we currently return blinded by default. It only moves where the decision is made so that we will be dynamically choose based on the flow on a following up PR. 

fixes #7835

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
@@ -113,7 +113,7 @@ public class BlockOperationSelectorFactory {
       final Optional<Boolean> requestedBlinded,
       final BlockProductionPerformance blockProductionPerformance) {
 
-    boolean blinded = requestedBlinded.orElse(ValidatorsUtil.DEFAULT_PRODUCE_BLINDED_BLOCK);
+    final boolean blinded = requestedBlinded.orElse(ValidatorsUtil.DEFAULT_PRODUCE_BLINDED_BLOCK);
 
     return bodyBuilder -> {
       final Eth1Data eth1Data = eth1DataCache.getEth1Vote(blockSlotState);
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
@@ -53,6 +53,7 @@ import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
 import tech.pegasys.teku.spec.datastructures.type.SszKZGCommitment;
 import tech.pegasys.teku.spec.datastructures.type.SszKZGProof;
 import tech.pegasys.teku.spec.executionlayer.ExecutionLayerBlockProductionManager;
+import tech.pegasys.teku.spec.logic.common.util.ValidatorsUtil;
 import tech.pegasys.teku.spec.logic.versions.deneb.helpers.MiscHelpersDeneb;
 import tech.pegasys.teku.spec.schemas.SchemaDefinitions;
 import tech.pegasys.teku.spec.schemas.SchemaDefinitionsBellatrix;
@@ -109,7 +110,11 @@ public class BlockOperationSelectorFactory {
       final BeaconState blockSlotState,
       final BLSSignature randaoReveal,
       final Optional<Bytes32> optionalGraffiti,
+      final Optional<Boolean> requestedBlinded,
       final BlockProductionPerformance blockProductionPerformance) {
+
+    boolean blinded = requestedBlinded.orElse(ValidatorsUtil.DEFAULT_PRODUCE_BLINDED_BLOCK);
+
     return bodyBuilder -> {
       final Eth1Data eth1Data = eth1DataCache.getEth1Vote(blockSlotState);
 
@@ -171,7 +176,12 @@ public class BlockOperationSelectorFactory {
         final SchemaDefinitions schemaDefinitions =
             spec.atSlot(blockSlotState.getSlot()).getSchemaDefinitions();
         setExecutionData(
-            bodyBuilder, schemaDefinitions, parentRoot, blockSlotState, blockProductionPerformance);
+            bodyBuilder,
+            blinded,
+            schemaDefinitions,
+            parentRoot,
+            blockSlotState,
+            blockProductionPerformance);
       }
 
       blockProductionPerformance.beaconBlockPrepared();
@@ -180,6 +190,7 @@ public class BlockOperationSelectorFactory {
 
   private void setExecutionData(
       final BeaconBlockBodyBuilder bodyBuilder,
+      final boolean blinded,
       final SchemaDefinitions schemaDefinitions,
       final Bytes32 parentRoot,
       final BeaconState blockSlotState,
@@ -189,7 +200,7 @@ public class BlockOperationSelectorFactory {
 
     // Pre-Deneb: Execution Payload / Execution Payload Header
     if (!bodyBuilder.supportsKzgCommitments()) {
-      if (bodyBuilder.isBlinded()) {
+      if (blinded) {
         builderSetPayloadHeader(
             bodyBuilder,
             schemaDefinitions,
@@ -219,16 +230,17 @@ public class BlockOperationSelectorFactory {
                             new IllegalStateException(
                                 "Cannot provide kzg commitments before The Merge")),
                     blockSlotState,
-                    bodyBuilder.isBlinded(),
+                    blinded,
                     blockProductionPerformance));
-    builderSetPayloadPostMerge(bodyBuilder, executionPayloadResultFuture);
-    builderSetKzgCommitments(bodyBuilder, schemaDefinitions, executionPayloadResultFuture);
+    builderSetPayloadPostMerge(bodyBuilder, blinded, executionPayloadResultFuture);
+    builderSetKzgCommitments(bodyBuilder, blinded, schemaDefinitions, executionPayloadResultFuture);
   }
 
   private void builderSetPayloadPostMerge(
       final BeaconBlockBodyBuilder bodyBuilder,
+      final boolean blinded,
       final SafeFuture<ExecutionPayloadResult> executionPayloadResultFuture) {
-    if (bodyBuilder.isBlinded()) {
+    if (blinded) {
       bodyBuilder.executionPayloadHeader(
           executionPayloadResultFuture.thenCompose(
               executionPayloadResult ->
@@ -308,6 +320,7 @@ public class BlockOperationSelectorFactory {
 
   private void builderSetKzgCommitments(
       final BeaconBlockBodyBuilder bodyBuilder,
+      final boolean blinded,
       final SchemaDefinitions schemaDefinitions,
       final SafeFuture<ExecutionPayloadResult> executionPayloadResultFuture) {
     final SchemaDefinitionsDeneb schemaDefinitionsDeneb =
@@ -315,7 +328,7 @@ public class BlockOperationSelectorFactory {
     final SafeFuture<SszList<SszKZGCommitment>> blobKzgCommitments =
         executionPayloadResultFuture.thenCompose(
             executionPayloadResult -> {
-              if (bodyBuilder.isBlinded()) {
+              if (blinded) {
                 return getBuilderBlobKzgCommitments(executionPayloadResult);
               } else {
                 return getExecutionBlobsBundle(executionPayloadResult)
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.bellatrix;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
@@ -20,26 +21,21 @@ import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.ssz.primitive.SszBytes32;
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.BeaconBlockBody;
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.BeaconBlockBodyBuilder;
+import tech.pegasys.teku.spec.datastructures.blocks.blockbody.BeaconBlockBodySchema;
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.altair.BeaconBlockBodyBuilderAltair;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayload;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadHeader;
 import tech.pegasys.teku.spec.datastructures.type.SszSignature;
 
 public class BeaconBlockBodyBuilderBellatrix extends BeaconBlockBodyBuilderAltair {
-  private final BeaconBlockBodySchemaBellatrixImpl schema;
-  private final BlindedBeaconBlockBodySchemaBellatrixImpl blindedSchema;
   protected SafeFuture<ExecutionPayload> executionPayload;
   protected SafeFuture<ExecutionPayloadHeader> executionPayloadHeader;
-
-  public BeaconBlockBodyBuilderBellatrix() {
-    this.schema = null;
-    this.blindedSchema = null;
-  }
+  protected final BeaconBlockBodySchema<? extends BlindedBeaconBlockBodyBellatrix> blindedSchema;
 
   public BeaconBlockBodyBuilderBellatrix(
-      final BeaconBlockBodySchemaBellatrixImpl schema,
-      final BlindedBeaconBlockBodySchemaBellatrixImpl blindedSchema) {
-    this.schema = schema;
+      final BeaconBlockBodySchema<? extends BeaconBlockBodyBellatrix> schema,
+      final BeaconBlockBodySchema<? extends BlindedBeaconBlockBodyBellatrix> blindedSchema) {
+    super(schema);
     this.blindedSchema = blindedSchema;
   }
 
@@ -62,12 +58,22 @@ public class BeaconBlockBodyBuilderBellatrix extends BeaconBlockBodyBuilderAltai
   }
 
   @Override
-  protected void validateSchema() {
-    if (isBlinded()) {
-      checkNotNull(blindedSchema, "blindedSchema must be set when blinded body has been requested");
+  @SuppressWarnings("unchecked")
+  protected <T> T getAndValidateSchema(final boolean blinded, final Class<T> expectedSchemaType) {
+    final BeaconBlockBodySchema<?> schema;
+    if (blinded) {
+      schema = this.blindedSchema;
+      checkNotNull(schema, "Blinded schema must be specified");
     } else {
-      checkNotNull(schema, "schema must be set when non blinded body has been requested");
+      schema = this.schema;
+      checkNotNull(schema, "Schema must be specified");
     }
+    checkArgument(
+        expectedSchemaType == schema.getClass(),
+        String.format(
+            "Schema should be %s but was %s",
+            expectedSchemaType.getSimpleName(), schema.getClass().getSimpleName()));
+    return (T) schema;
   }
 
   @Override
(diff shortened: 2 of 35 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -109,7 +110,11 @@ public Consumer<BeaconBlockBodyBuilder> createSelector(
       final BeaconState blockSlotState,
       final BLSSignature randaoReveal,
       final Optional<Bytes32> optionalGraffiti,
+      final Optional<Boolean> requestedBlinded,
       final BlockProductionPerformance blockProductionPerformance) {
+
+    boolean blinded = requestedBlinded.orElse(ValidatorsUtil.DEFAULT_PRODUCE_BLINDED_BLOCK);
```
rolfyone: final

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/capella/BeaconBlockBodyBuilderCapella.java
```diff
@@ -30,21 +28,20 @@
...
   private SszList<SignedBlsToExecutionChange> blsToExecutionChanges;
 
-  public BeaconBlockBodyBuilderCapella schema(final BeaconBlockBodySchemaCapellaImpl schema) {
-    this.schema = schema;
-    this.blinded = Optional.of(false);
-    return this;
+  public BeaconBlockBodyBuilderCapella() {
+    this.schema = null;
+    this.blindedSchema = null;
   }
```
rolfyone: if this does ever get called it's not going to end well...

rolfyone: i guess to clarify, if we're going down this route, we really need to do some checks on build to make sure everything is as it needs to be - null of either of these will be brutal through most of the code if it gets to blockBody objects...

does look like we're doing that... maybe this is going to be ok...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/capella/BeaconBlockBodySchemaCapellaImpl.java
```diff
@@ -131,7 +131,7 @@ BlockBodyFields.EXECUTION_PAYLOAD, new ExecutionPayloadSchemaCapella(specConfig)
   @Override
   public SafeFuture<? extends BeaconBlockBody> createBlockBody(
       final Consumer<BeaconBlockBodyBuilder> builderConsumer) {
-    final BeaconBlockBodyBuilderCapella builder = new BeaconBlockBodyBuilderCapella().schema(this);
+    final BeaconBlockBodyBuilderCapella builder = new BeaconBlockBodyBuilderCapella(this, null);
```
rolfyone, line 134: i think i'd prefer the previous version and having the second param initialise as null in the builder... really concerns me passing nulls across objects, pretty much ever... if it starts as null internally and gets set when required, then that'd be a better pattern.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/phase0/BeaconBlockBodyBuilderPhase0.java
```diff
@@ -46,6 +46,14 @@ public class BeaconBlockBodyBuilderPhase0 implements BeaconBlockBodyBuilder {
   protected SszList<Deposit> deposits;
   protected SszList<SignedVoluntaryExit> voluntaryExits;
 
+  public BeaconBlockBodyBuilderPhase0() {
```
rolfyone, line 49: i think we should initialise to null and just call it if we're making it not null, and then this fn shouldn't need to exist maybe?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/capella/BeaconBlockBodySchemaCapellaImpl.java
tbenr: ye I was doing some back and forth on that. I landed here to avoid multiplication of constructors (up to 4 for each builder)

tbenr: My idea was not using the builder pattern for schema. But yeah.. I could go back on that decision...

tbenr: mmm the problem with set methods is that due to inheritance we will have all sets from supers with schema types for previous milestone.. it is bad. I think at the end I prefer the current one.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/capella/BeaconBlockBodyBuilderCapella.java
zilm13: I was thinking a lot how to deal with Builders default constructor. I'd suggest to mark it with `@Deprecated(since = "Not for use")` in each builder. I think it's the least evil to show a clear sign it should not be used. Also I think adding Exception in the end of constructor could be a way too.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
```diff
@@ -53,48 +50,36 @@ public Boolean supportsExecutionPayload() {
 
   @Override
   public BeaconBlockBodyBuilder executionPayload(SafeFuture<ExecutionPayload> executionPayload) {
```
zilm13: final please

zilm13:
@tbenr @rolfyone 
How about such directions https://github.com/zilm13/teku/commit/a73cb1e28816d3c58c5de2d4de6627e2bba6129e
Not beautiful but looks safer for me.
What I don't like: 
- blinded and non blinded schema could be relatives which makes instanceof check dangerous. Though in our current code this cannot be happened
- we cannot tighten schema type as `build` param (which could be beneficial)

tbenr:
@zilm13 interesting idea! but I don't like the runtime type check.. so from a type safety perspective my version is better but yours is better wrt forcing user passing a schema (not picking a wrong constructor).

Let me play with your version a bit...

tbenr:
@zilm13 progressed your version https://github.com/tbenr/teku/commit/0d98e33ebdcac8ac1a368afdbaf6fcebc6cdaffa

- removes builder param, moving it back to constructor.
- tightens the runtime schema check.

I think we could even avoid the `Function<>` and simply have blinded and unblinded schema generics in the constructor.
So it will be a mix of my version and your version. It make sense if we give up on having compile-time type checks and relay on runtime checks.

zilm13:
@tbenr I like the direction!

tbenr:
another step https://github.com/tbenr/teku/commit/948bb52042efdd254a622e47d3a3728a328408cc
this might be controversial but I don't like the schema resolver function. It only avoid passing null, but null can be provided by the function itself so it doesn't solve much.

The nice thing is that, pre-bellatrix, we don't even need to consider blinded schema, and types are in constructor are tighter (non perfectly, runtime check still required) with the types we need.

I know this is getting closer to the original version :)

zilm13:
I see. 

We need 3rd person for opinion
@rolfyone 
1. The old version with two constructors, nulls etc
2. No args constructor only, no schema fields in class, schema resolvers parameter for build(), schema validation in runtime https://github.com/zilm13/teku/commit/a73cb1e28816d3c58c5de2d4de6627e2bba6129e
3. Evolution of 2 but schema resolver is a single parameter for constructor https://github.com/tbenr/teku/commit/0d98e33ebdcac8ac1a368afdbaf6fcebc6cdaffa
4. Evolution of 3 but closer to 1. No default constructor, both schemas are parameters for constructor https://github.com/tbenr/teku/commit/948bb52042efdd254a622e47d3a3728a328408cc

My favorite is 3. I don't like passing nulls in constructor. But 4 has better type-check. I'll agree on anything from 2,3,4 but don't want to choose :)

tbenr:
updating PR with 4, exercising my bias as PR author :)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
```diff
@@ -13,37 +13,30 @@
...
-  public BeaconBlockBodyBuilderBellatrix blindedSchema(
-      final BlindedBeaconBlockBodySchemaBellatrixImpl schema) {
-    this.blindedSchema = schema;
-    this.blinded = Optional.of(true);
-    return this;
+  public BeaconBlockBodyBuilderBellatrix(
+      final BeaconBlockBodySchema<? extends BeaconBlockBodyBellatrix> schema,
+      final BeaconBlockBodySchema<? extends BlindedBeaconBlockBodyBellatrix> blindedSchema) {
+    super(schema);
+    this.blindedSchema = blindedSchema;
```
zilm13, line 40: May worth adding something like 
```java
    checkState(schema != null || blindedSchema != null, "At least one schema must be specified");

```
in the end

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -56,18 +60,54 @@ void equalsReturnsFalseWhenExecutionPayloadIsDifferent() {
     assertNotEquals(defaultBlockBody, testBeaconBlockBody);
   }
 
+  @Test
+  void equalsReturnsFalseWhenExecutionPayloadHeaderIsDifferent() {
+    executionPayloadHeader = dataStructureUtil.randomExecutionPayloadHeader();
+    BlindedBeaconBlockBodyBellatrix testBeaconBlockBody = safeJoin(createBlindedBlockBody());
```
zilm13: `testBlindedBlockBody` for clarity? Also you could add final

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -56,18 +60,54 @@ void equalsReturnsFalseWhenExecutionPayloadIsDifferent() {
...
+  }
+
+  @Test
+  public void builderShouldFailIfNotPassingPayload() {
+    final Exception exception =
+        assertThrows(
+            IllegalStateException.class,
+            () ->
+                createBlockBody(
+                    createContentProvider(true)
```
zilm13: Logic of this test is a bit non transparent. You create provider for blinded header, it's set underneath, next you erase both header and payload. Name of the test refers to Payload, so I expect at least false for content provider. Also as part of the logic is hidden, comment would be helpful (that we erase something).

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -56,18 +60,54 @@ void equalsReturnsFalseWhenExecutionPayloadIsDifferent() {
...
+                            b ->
+                                b.executionPayload(SafeFuture.completedFuture(executionPayload)))));
+    assertEquals(
+        exception.getMessage(),
+        "only and only one of executionPayload or executionPayloadHeader must be set");
+  }
+
   @Test
   @SuppressWarnings("unchecked")
   void builderShouldFailWhenOverridingBlindedSchemaWithANullSchema() {
```
zilm13: I'd rename this one to something like builderShouldFailWhenNonBlindedSchemaSetWithPayload or something else to highlight the purpose. We are no more overriding it

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -100,25 +141,39 @@ void builderShouldFailWhenOverridingSchemaWithANullBlindedSchema() {
...
-        .createBlockBody(contentProvider)
-        .thenApply(body -> (BeaconBlockBodyBellatrix) body);
+    final BeaconBlockBodyBuilder bodyBuilder = createBeaconBlockBodyBuilder();
+    contentProvider.accept(bodyBuilder);
+    return bodyBuilder.build().thenApply(body -> body.toVersionBellatrix().orElseThrow());
+  }
+
+  @Override
+  protected SafeFuture<BlindedBeaconBlockBodyBellatrix> createBlindedBlockBody(
+      Consumer<BeaconBlockBodyBuilder> contentProvider) {
```
zilm13: final please

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -76,22 +116,23 @@ void builderShouldFailWhenOverridingBlindedSchemaWithANullSchema() {
...
+                    .executionPayload(mock(SafeFuture.class))
+                    .syncAggregate(mock(SyncAggregate.class))
                     .build());
-    assertEquals(exception.getMessage(), "schema must be set with no blindedSchema");
+    assertEquals(exception.getMessage(), "Schema must be specified");
   }
 
   @Test
   @SuppressWarnings("unchecked")
   void builderShouldFailWhenOverridingSchemaWithANullBlindedSchema() {
```
zilm13: We need a new name for this one too. Without `overriding`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
```diff
@@ -53,58 +46,58 @@ public Boolean supportsExecutionPayload() {
...
   protected void validate() {
     super.validate();
-    if (isBlinded()) {
-      checkNotNull(executionPayloadHeader, "executionPayloadHeader must be specified");
-    } else {
-      checkNotNull(executionPayload, "executionPayload must be specified");
-    }
+    checkState(
+        executionPayload != null ^ executionPayloadHeader != null,
+        "only and only one of executionPayload or executionPayloadHeader must be set");
```
zilm13: Maybe "Only one of executionPayload and executionPayloadHeader must be set" is enough?

zilm13 (commented):
I think it worth having tests like we have for Bellatrix (in BeaconBlockBodyBellatrixTest) also running for Capella and Deneb, especially as we have runtime schema validation. It would be good if it involves some kind of extension, so when we create Electra tests you will be required to run it too.
Also some comments below

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
tbenr: yeah.. :)

tbenr: I had a chatGPT session and came up with a different version

tbenr:
> I think it worth having tests like we have for Bellatrix (in BeaconBlockBodyBellatrixTest) also running for Capella and Deneb, especially as we have runtime schema validation. It would be good if it involves some kind of extension, so when we create Electra tests you will be required to run it too.

I was thinking about that and I have mixed feelings. If we trust inheritance we can potentially check logic where they are introduced without necessarily testing them for all subclasses. It assumes that subclasses don't override and mess up with existing logic. I don't like much how current tests are organized,  and they don't help doing this systematic tests everywhere. I haven't spent much time redesigning them...
A dummy\dirty approach is to move some bellatrix tests in the abstract class, I did that but seemed to me too dirty..

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
```diff
@@ -52,59 +46,60 @@ public Boolean supportsExecutionPayload() {
...
   @Override
-  protected void validateSchema() {
-    if (isBlinded()) {
-      checkState(
-          blindedSchema != null && schema == null, "blindedSchema must be set with no schema");
+  @SuppressWarnings("unchecked")
+  protected <T> T getAndValidateSchema(final boolean blinded, final Class<T> expectedSchemaType) {
+    final BeaconBlockBodySchema<?> schema;
+    if (blinded) {
+      schema = this.blindedSchema;
```
zilm13: `this.` is a bit overkill here and 3 lines down, I think it could be safely removed

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -56,18 +60,56 @@ void equalsReturnsFalseWhenExecutionPayloadIsDifferent() {
...
+  void equalsReturnsFalseWhenExecutionPayloadHeaderIsDifferent() {
+    executionPayloadHeader = dataStructureUtil.randomExecutionPayloadHeader();
+    final BlindedBeaconBlockBodyBellatrix testBlindedBeaconBlockBody =
+        safeJoin(createBlindedBlockBody());
+
+    assertNotEquals(defaultBlindedBlockBody, testBlindedBeaconBlockBody);
+  }
+
+  @Test
+  public void builderShouldFailIfNotPassingPayload() {
```
zilm13: this one should be probably called `builderShouldFailIfNotPassingPayloadAndPayloadHeader`

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -56,18 +60,56 @@ void equalsReturnsFalseWhenExecutionPayloadIsDifferent() {
...
+                    createContentProvider(false)
+                        // let's erase all payloads, so we can test as if they were not set
+                        .andThen(b -> b.executionPayload(null).executionPayloadHeader(null))));
+    assertEquals(
+        exception.getMessage(),
+        "Exactly one of 'executionPayload' or 'executionPayloadHeader' must be set");
+  }
+
+  @Test
+  public void builderShouldFailIfNotPassingPayloadAndPayloadHeader() {
```
zilm13: while this one something like `builderShouldFailWhenPassingPayloadAndPayloadHeader`

ethereum/spec/src/test/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBellatrixTest.java
```diff
@@ -56,18 +60,56 @@ void equalsReturnsFalseWhenExecutionPayloadIsDifferent() {
...
+    final Exception exception =
+        assertThrows(
+            IllegalStateException.class,
+            () ->
+                createBlockBody(
+                    createContentProvider(true)
+                        .andThen(
+                            b ->
+                                b.executionPayload(SafeFuture.completedFuture(executionPayload)))));
```
zilm13, line 99: I'd add a comment that header is already set by `createContentProvider`

zilm13 (approved):
LGTM, just minor nits
Plus I suggest to add 2 success case tests to ` AbstractBeaconBlockBodyTest`:
```java
  @Test
  public void builderShouldCreateBlinded() {
    assumeTrue(blindedBlockBodySchema != null);
    SafeFuture<? extends BeaconBlockBody> blockBody = createBlindedBlockBody(
        createContentProvider(true));
    assertThatSafeFuture(blockBody).isCompletedWithValueMatching(
        beaconBlockBody -> beaconBlockBody.getOptionalExecutionPayload().isEmpty() && beaconBlockBody.getOptionalExecutionPayloadHeader().isPresent());
  }
```
one for blinded and another for unblinded. It will run for all versions and looks simple

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/bellatrix/BeaconBlockBodyBuilderBellatrix.java
rolfyone: omg chatgpt?? :/

PR #7826, opened 2023-12-13
7707 validator slashing protection

### LINKED ISSUE #7707: Validator slashing protection
credit to mariano.eth for the idea...

```
How fast does slashing info propagate on the beacon chain?  If you have 2 servers with the same 100 validators, and say 3 of them attest on the same slot and are slashed, could the CL client know by next slot (~12 secs) and turn itself off?
```

mehdi-aouadi:
## PR Description
Add `--shut-down-when-validator-slashed-enabled` option to enable Teku shutdown when a validator gets slashed.
When using a remote BN, this option requires the BN to support the new `attester_slashing` and `proposer_slashing` SSE events.
When running VC only, in order to ensure BN backward compatibility, the VC will subscribe to the `attester_slashing` and `proposer_slashing` SSE events only when the `--shut-down-when-validator-slashed-enabled` is enabled (in addition to the `head` events) otherwise it will only subscribe to the `head` events.

## Fixed Issue(s)

fixes #7707 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,6 @@ the [releases page](https://github.com/Consensys/teku/releases).
 - Added Deneb (aka Dencun) configuration for Chiado network for epoch 516608 (2024-01-31 18:15:40 UTC).
 - Added Deneb (aka Dencun) configuration for Holesky network for epoch 29696 (2024-02-07 11:34:24 UTC).
 - Generate key at `—p2p-private-key-file` path if specified file doesn't exist.
-- Added `--stop-vc-when-validator-slashed` option to stop the VC when a validator is slashed
+- Added `--stop-vc-when-validator-slashed-enabled` option to stop the VC when a validator is slashed
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,5 +24,6 @@ the [releases page](https://github.com/Consensys/teku/releases).
 - Added Deneb (aka Dencun) configuration for Chiado network for epoch 516608 (2024-01-31 18:15:40 UTC).
 - Added Deneb (aka Dencun) configuration for Holesky network for epoch 29696 (2024-02-07 11:34:24 UTC).
 - Generate key at `—p2p-private-key-file` path if specified file doesn't exist.
+- Added `--stop-vc-when-validator-slashed` option to stop the VC when a validator is slashed
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,6 @@ the [releases page](https://github.com/Consensys/teku/releases).
 - Added Deneb (aka Dencun) configuration for Chiado network for epoch 516608 (2024-01-31 18:15:40 UTC).
 - Added Deneb (aka Dencun) configuration for Holesky network for epoch 29696 (2024-02-07 11:34:24 UTC).
 - Generate key at `—p2p-private-key-file` path if specified file doesn't exist.
-- Added `--stop-vc-when-validator-slashed-enabled` option to stop the VC when a validator is slashed
+- Added `--shutdown-when-validator-slashed-enabled` option to shut down Teku client when a validator is slashed
 
 ### Bug Fixes
--- a/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
@@ -168,37 +168,40 @@ public class NodeDataProvider {
             });
   }
 
-  public void subscribeToReceivedBlocks(ImportedBlockListener listener) {
+  public void subscribeToReceivedBlocks(final ImportedBlockListener listener) {
     blockManager.subscribeToReceivedBlocks(listener);
   }
 
-  public void subscribeToReceivedBlobSidecar(NewBlobSidecarSubscriber listener) {
+  public void subscribeToReceivedBlobSidecar(final NewBlobSidecarSubscriber listener) {
     blockBlobSidecarsTrackersPool.subscribeNewBlobSidecar(listener);
   }
 
-  public void subscribeToAttesterSlashing(OperationAddedSubscriber<AttesterSlashing> listener) {
+  public void subscribeToAttesterSlashing(
+      final OperationAddedSubscriber<AttesterSlashing> listener) {
     attesterSlashingPool.subscribeOperationAdded(listener);
   }
 
-  public void subscribeToProposerSlashing(OperationAddedSubscriber<ProposerSlashing> listener) {
+  public void subscribeToProposerSlashing(
+      final OperationAddedSubscriber<ProposerSlashing> listener) {
     proposerSlashingPool.subscribeOperationAdded(listener);
   }
 
-  public void subscribeToValidAttestations(ProcessedAttestationListener listener) {
+  public void subscribeToValidAttestations(final ProcessedAttestationListener listener) {
     attestationManager.subscribeToAllValidAttestations(listener);
   }
 
-  public void subscribeToNewVoluntaryExits(OperationAddedSubscriber<SignedVoluntaryExit> listener) {
+  public void subscribeToNewVoluntaryExits(
+      final OperationAddedSubscriber<SignedVoluntaryExit> listener) {
     voluntaryExitPool.subscribeOperationAdded(listener);
   }
 
   public void subscribeToNewBlsToExecutionChanges(
-      OperationAddedSubscriber<SignedBlsToExecutionChange> listener) {
+      final OperationAddedSubscriber<SignedBlsToExecutionChange> listener) {
     blsToExecutionChangePool.subscribeOperationAdded(listener);
   }
 
   public void subscribeToSyncCommitteeContributions(
-      OperationAddedSubscriber<SignedContributionAndProof> listener) {
+      final OperationAddedSubscriber<SignedContributionAndProof> listener) {
     syncCommitteeContributionPool.subscribeOperationAdded(listener);
   }
 
--- a/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
@@ -168,7 +168,7 @@ public class NodeDataProvider {
             });
   }
 
-  public void subscribeToReceivedBlocks(final ImportedBlockListener listener) {
+  public void subscribeToReceivedBlocks(ImportedBlockListener listener) {
     blockManager.subscribeToReceivedBlocks(listener);
   }
 
--- a/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
@@ -172,7 +172,7 @@ public class NodeDataProvider {
     blockManager.subscribeToReceivedBlocks(listener);
   }
 
-  public void subscribeToReceivedBlobSidecar(final NewBlobSidecarSubscriber listener) {
+  public void subscribeToReceivedBlobSidecar(NewBlobSidecarSubscriber listener) {
     blockBlobSidecarsTrackersPool.subscribeNewBlobSidecar(listener);
   }
 
@@ -186,22 +186,21 @@ public class NodeDataProvider {
     proposerSlashingPool.subscribeOperationAdded(listener);
   }
 
-  public void subscribeToValidAttestations(final ProcessedAttestationListener listener) {
+  public void subscribeToValidAttestations(ProcessedAttestationListener listener) {
     attestationManager.subscribeToAllValidAttestations(listener);
   }
 
-  public void subscribeToNewVoluntaryExits(
-      final OperationAddedSubscriber<SignedVoluntaryExit> listener) {
+  public void subscribeToNewVoluntaryExits(OperationAddedSubscriber<SignedVoluntaryExit> listener) {
     voluntaryExitPool.subscribeOperationAdded(listener);
   }
 
   public void subscribeToNewBlsToExecutionChanges(
-      final OperationAddedSubscriber<SignedBlsToExecutionChange> listener) {
+      OperationAddedSubscriber<SignedBlsToExecutionChange> listener) {
     blsToExecutionChangePool.subscribeOperationAdded(listener);
   }
 
   public void subscribeToSyncCommitteeContributions(
-      final OperationAddedSubscriber<SignedContributionAndProof> listener) {
+      OperationAddedSubscriber<SignedContributionAndProof> listener) {
     syncCommitteeContributionPool.subscribeOperationAdded(listener);
   }
 
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -249,7 +249,6 @@ public class StatusLogger {
   public void validatorSlashedAlert(final Set<String> slashedValidatorPublicKeys) {
     log.fatal(
         "Validator(s) with public key(s) {} got slashed",
-        slashedValidatorPublicKeys.size(),
         String.join(", ", slashedValidatorPublicKeys));
   }
 
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -248,7 +248,7 @@ public class StatusLogger {
 
   public void validatorSlashedAlert(final Set<String> slashedValidatorPublicKeys) {
     log.fatal(
-        "Validator(s) with public key(s) {} got slashed",
+        "Validator(s) with public key(s) {} got slashed. Shutting down validator client...",
         String.join(", ", slashedValidatorPublicKeys));
   }
 
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -248,7 +248,7 @@ public class StatusLogger {
 
   public void validatorSlashedAlert(final Set<String> slashedValidatorPublicKeys) {
     log.fatal(
-        "Validator(s) with public key(s) {} got slashed. Shutting down validator client...",
+        "Validator(s) with public key(s) {} got slashed. Shutting down...",
         String.join(", ", slashedValidatorPublicKeys));
   }
 
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -246,6 +246,13 @@ public class StatusLogger {
         "No loaded validators when --exit-when-no-validator-keys-enabled option is true. Shutting down...");
   }
 
+  public void validatorSlashedAlert(final Set<String> slashedValidatorPublicKeys) {
+    log.fatal(
+        "Validator(s) with public key(s) {} got slashed",
+        slashedValidatorPublicKeys.size(),
+        String.join(", ", slashedValidatorPublicKeys));
+  }
+
   public void beginInitializingChainData() {
     log.info("Initializing storage");
   }
--- a/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
+++ b/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
@@ -180,6 +180,7 @@ import tech.pegasys.teku.storage.store.StoreConfig;
 import tech.pegasys.teku.validator.api.InteropConfig;
 import tech.pegasys.teku.validator.api.ValidatorApiChannel;
 import tech.pegasys.teku.validator.api.ValidatorPerformanceTrackingMode;
+import tech.pegasys.teku.validator.api.ValidatorTimingChannel;
 import tech.pegasys.teku.validator.coordinator.ActiveValidatorTracker;
 import tech.pegasys.teku.validator.coordinator.BlockFactory;
 import tech.pegasys.teku.validator.coordinator.BlockOperationSelectorFactory;
@@ -642,6 +643,11 @@ public class BeaconChainController extends Service implements BeaconChainControl
                 .reversed());
     blockImporter.subscribeToVerifiedBlockAttesterSlashings(attesterSlashingPool::removeAll);
     attesterSlashingPool.subscribeOperationAdded(forkChoice::onAttesterSlashing);
+    final ValidatorTimingChannel validatorTimingChannel =
+        eventChannels.getPublisher(ValidatorTimingChannel.class);
+    attesterSlashingPool.subscribeOperationAdded(
+        (operation, validationStatus, fromNetwork) ->
+            validatorTimingChannel.onAttesterSlashing(operation));
   }
 
   protected void initProposerSlashingPool() {
@@ -654,6 +660,11 @@ public class BeaconChainController extends Service implements BeaconChainControl
             beaconBlockSchemaSupplier.andThen(BeaconBlockBodySchema::getProposerSlashingsSchema),
             validator);
     blockImporter.subscribeToVerifiedBlockProposerSlashings(proposerSlashingPool::removeAll);
+    final ValidatorTimingChannel validatorTimingChannel =
+        eventChannels.getPublisher(ValidatorTimingChannel.class);
+    proposerSlashingPool.subscribeOperationAdded(
+        (operation, validationStatus, fromNetwork) ->
+            validatorTimingChannel.onProposerSlashing(operation));
   }
 
   protected void initVoluntaryExitPool() {
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
@@ -15,8 +15,8 @@ package tech.pegasys.teku.cli.options;
 
 import static tech.pegasys.teku.networks.Eth2NetworkConfiguration.DEFAULT_VALIDATOR_EXECUTOR_THREADS;
 import static tech.pegasys.teku.validator.api.ValidatorConfig.DEFAULT_DOPPELGANGER_DETECTION_ENABLED;
+import static tech.pegasys.teku.validator.api.ValidatorConfig.DEFAULT_STOP_VC_WHEN_VALIDATOR_SLASHED_ENABLED;
 import static tech.pegasys.teku.validator.api.ValidatorConfig.DEFAULT_VALIDATOR_IS_LOCAL_SLASHING_PROTECTION_SYNCHRONIZED_ENABLED;
-import static tech.pegasys.teku.validator.api.ValidatorConfig.DEFAULT_VALIDATOR_SLASHING_PROTECTION_ENABLED;
 
 import java.nio.file.Path;
 import java.util.Optional;
@@ -154,15 +154,14 @@ public class ValidatorOptions {
       DEFAULT_VALIDATOR_IS_LOCAL_SLASHING_PROTECTION_SYNCHRONIZED_ENABLED;
 
   @Option(
-      names = {"--validator-slashing-protection-enabled"},
+      names = {"--stop-vc-when-validator-slashed-enabled"},
       paramLabel = "<BOOLEAN>",
       description =
           "If an owned validator key is detected as slashed, the validator-client should terminate. In this case, the service should not be restarted",
       showDefaultValue = CommandLine.Help.Visibility.ALWAYS,
       arity = "0..1",
       fallbackValue = "true")
-  private boolean validatorSlashingProtectionEnabled =
-      DEFAULT_VALIDATOR_SLASHING_PROTECTION_ENABLED;
+  private boolean stopVcWhenValidatorSlashed = DEFAULT_STOP_VC_WHEN_VALIDATOR_SLASHED_ENABLED;
 
   public void configure(TekuConfiguration.Builder builder) {
     builder.validator(
@@ -183,7 +182,7 @@ public class ValidatorOptions {
                 .executorThreads(executorThreads)
                 .blockV3enabled(blockV3Enabled)
                 .exitWhenNoValidatorKeysEnabled(exitWhenNoValidatorKeysEnabled)
-                .validatorSlashingProtectionEnabled(validatorSlashingProtectionEnabled));
+                .stopVcWhenValidatorSlashedEnabled(stopVcWhenValidatorSlashed));
     validatorProposerOptions.configure(builder);
     validatorKeysOptions.configure(builder);
   }
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
@@ -157,7 +157,7 @@ public class ValidatorOptions {
       names = {"--stop-vc-when-validator-slashed-enabled"},
       paramLabel = "<BOOLEAN>",
       description =
-          "If an owned validator key is detected as slashed, the validator-client should terminate. In this case, the service should not be restarted",
+          "If an owned validator key is detected as slashed, the validator-client should terminate. In this case, the service should not be restarted.",
       showDefaultValue = CommandLine.Help.Visibility.ALWAYS,
       arity = "0..1",
       fallbackValue = "true")
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
@@ -157,7 +157,7 @@ public class ValidatorOptions {
       names = {"--stop-vc-when-validator-slashed-enabled"},
       paramLabel = "<BOOLEAN>",
       description =
-          "If an owned validator key is detected as slashed, the validator-client should terminate. In this case, the service should not be restarted.",
+          "If an owned validator key is detected as slashed, the validator-client or the teku node (if running the beacon-node and the validator-client in a single process) should terminate with exit code 2. In this case, the service should not be restarted.",
       showDefaultValue = CommandLine.Help.Visibility.ALWAYS,
       arity = "0..1",
       fallbackValue = "true")
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
@@ -15,8 +15,8 @@ package tech.pegasys.teku.cli.options;
 
 import static tech.pegasys.teku.networks.Eth2NetworkConfiguration.DEFAULT_VALIDATOR_EXECUTOR_THREADS;
 import static tech.pegasys.teku.validator.api.ValidatorConfig.DEFAULT_DOPPELGANGER_DETECTION_ENABLED;
-import static tech.pegasys.teku.validator.api.ValidatorConfig.DEFAULT_STOP_VC_WHEN_VALIDATOR_SLASHED_ENABLED;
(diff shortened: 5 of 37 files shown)


mehdi-aouadi:
When the VC is used with a remote BN that doesn't support the `attester_slashing` and the `proposer_slashing` events the SSE connection will fail (and retry with the failover BN if any). I'm trying to customise the retry strategy in `EventSourceBeaconChainEventAdapter::createEventSource` to exclude the `attester_slashing` and `proposer_slashing` topic from the query params when the request fails but didn't find a way to do it yet. Not sure if that's even possible

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Additions and Improvements
 - Added `--ee-jwt-claim-id` command line option to provide `id` to the execution engine JWT claims
+- Added `--stop-vc-when-validator-slashed` option to stop the VC when a validator is slashed (requires a BN that supports the `attester_slashing` and `proposer_slashing` SSE events)
```
zilm13: Maybe `stop the VC when any owned validator is slashed` to highlight and make more clear? If you up to it, please, change command line description too.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -140,6 +141,16 @@ public class ValidatorOptions {
   private boolean exitWhenNoValidatorKeysEnabled =
       ValidatorConfig.DEFAULT_EXIT_WHEN_NO_VALIDATOR_KEYS_ENABLED;
 
+  @Option(
+      names = {"--stop-vc-when-validator-slashed"},
```
rolfyone: feature flags basically end with '-enabled' - pretty much a standard...
maybe `--stop-vc-when-validator-key-slashed-enabled`

I'd probably expand on the description also...
`If a validator key is detected as slashed, the validator-client should terminate. In this case, the service should not be restarted.`

And we'd want to document how this should all be used pretty carefully.

With systemd, you can stop things from restarting on exit codes ,and i think in the same way, we'd probably want to consider terminating code 2 (rather than code 1 like normal), and then we document that using this feature, you should avoid restarts on code 2 - because if it restarts it'll likely have more slashings. Potentially restarting with doppelganger support would be effective, but definitely would be risky to restart by default...

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  public void perform(List<BLSPublicKey> pubKeys) {
+    if (!pubKeys.isEmpty()) {
+      LOG.fatal(
+          "Validator(s) with public key(s) {} got slashed. Shutting down...",
+          pubKeys.stream()
+              .map(BLSPublicKey::toAbbreviatedString)
+              .collect(Collectors.joining(", ")));
+      System.exit(1);
```
rolfyone: as suggested above, i think potentially we want this to be exiting non code 1 - probably code 2, and we can document why this is important (ability to avoid trying to restart if configured)

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -246,6 +246,12 @@ public void exitOnNoValidatorKeys() {
         "No loaded validators when --exit-when-no-validator-keys-enabled option is true. Shutting down...");
   }
 
+  public void validatorSlashedAlert(final Set<String> slashedValidatorPublicKeys) {
+    log.fatal(
```
zilm13, line 250: we usually quit on fatal, though I see it's used only for test

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorAlert.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package tech.pegasys.teku.validator.client.slashingriskactions;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import tech.pegasys.teku.bls.BLSPublicKey;
+import tech.pegasys.teku.infrastructure.logging.StatusLogger;
+
+public class SlashedValidatorAlert implements SlashingRiskDetectionAction {
```
zilm13: It takes some time to understand that alerts versions are used in tests only.  I think it could be good to highlight it somehow. Either move implementation in tests or may be add some comments

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
```diff
@@ -0,0 +1,37 @@
...
+import tech.pegasys.teku.bls.BLSPublicKey;
+
+public class SlashedValidatorShutDown implements SlashingRiskDetectionAction {
+
+  private static final Logger LOG = LogManager.getLogger();
+
+  @Override
+  public void perform(List<BLSPublicKey> pubKeys) {
+    if (!pubKeys.isEmpty()) {
+      LOG.fatal(
```
zilm13: May worth use status log here too. No messages from standard log are displayed in console log in some configurations

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorTimingActions.java
```diff
@@ -97,8 +105,28 @@ public void onAttestationAggregationDue(final UInt64 slot) {
   }
 
   @Override
-  public void onAttesterSlashing(final AttesterSlashing attesterSlashing) {}
+  public void onAttesterSlashing(final AttesterSlashing attesterSlashing) {
+    delegates.forEach(delegates -> delegates.onAttesterSlashing(attesterSlashing));
+    final Set<UInt64> slashedIndices = attesterSlashing.getIntersectingValidatorIndices();
```
zilm13: I'd put all logic in `maybeValidatorSlashedAction.ifPresent(` here and in proposer slashing to not calculate this if it's not set. Who knows how could we use this subscriptions in the future. And `validatorIndexProvider.getPublicKey` looks not cheap

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
```diff
@@ -128,7 +132,14 @@ public void onPrimaryNodeBackReady() {
   }
 
   private BackgroundEventSource createEventSource(final RemoteValidatorApiChannel beaconNodeApi) {
-    final HttpUrl eventSourceUrl = createHeadEventSourceUrl(beaconNodeApi.getEndpoint());
+    final HttpUrl eventSourceUrl =
```
zilm13, line 141: How about this? More streamlined and easier to extend
```java
    final List<EventType> eventTypes = new ArrayList<>();
    eventTypes.add(EventType.head);
    if (validatorSlashingMonitoringEnabled) {
      eventTypes.add(EventType.attester_slashing);
      eventTypes.add(EventType.proposer_slashing);
    }
    final HttpUrl eventSourceUrl =
        createEventStreamSourceUrl(
            beaconNodeApi.getEndpoint(), eventTypes.toArray(EventType[]::new));
```

zilm13 (commented):
Looks good, just nit comments

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
mehdi-aouadi: I renamed the option to `--validator-slashing-protection-enabled`, it's clearer imo.
I also updated the description and updated the exit code to 2. I'll update the documentation PR accordingly

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
mehdi-aouadi: Done

CHANGELOG.md
mehdi-aouadi: Updated it

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
mehdi-aouadi: Well I kept it a fatal at the end since I used it in the shut down action implementation...

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorAlert.java
mehdi-aouadi: Done, moved the alert implementation to the tests package (the DG alert is not used in tests only fyi)

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorTimingActions.java
mehdi-aouadi: Good point! Moved it

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
mehdi-aouadi: Done. Updated the createEventStreamSourceUrl to take a list as argument...

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
mehdi-aouadi: I tried to describe all the details in a dedicated section in the doc. I raised a draft PR for that: https://github.com/Consensys/doc.teku/pull/526

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -152,6 +153,17 @@ public class ValidatorOptions {
   private boolean isLocalSlashingProtectionSynchronizedEnabled =
       DEFAULT_VALIDATOR_IS_LOCAL_SLASHING_PROTECTION_SYNCHRONIZED_ENABLED;
 
+  @Option(
+      names = {"--validator-slashing-protection-enabled"},
+      paramLabel = "<BOOLEAN>",
+      description =
+          "If an owned validator key is detected as slashed, the validator-client should terminate. In this case, the service should not be restarted",
```
rolfyone: probably a full-stop `.` at the end...

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
rolfyone: can we not call it `validator-slashing-protection-enabled`? that suggests that without this flag we're not doing local slashing protection...

rolfyone:
im not sure where to test so general observation...
Negative testing - 
eg. a slashing occurs which is not a validator we're a custodian of - should not shut us down :)

This would be hard to replicate on a network, slashings aren't super common so we might end up not seeing it, but we'd be easily able to see we don't trigger if a slashing we don't care about arrives at a low level i assume...

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ the [releases page](https://github.com/Consensys/teku/releases).
 ### Breaking Changes
 
 ### Additions and Improvements
+- Added `--stop-vc-when-validator-slashed` option to stop the VC when a validator is slashed (requires a BN that supports the `attester_slashing` and `proposer_slashing` SSE events)
```
zilm13: need to remove this line

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ the [releases page](https://github.com/Consensys/teku/releases).
...
+- Added `--validator-slashing-protection-enabled` option to stop the VC when any owned validator is slashed (requires a BN that supports the `attester_slashing` and `proposer_slashing` SSE events)
```
zilm13: and update flag name to the latest in this line

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorTimingActions.java
```diff
@@ -97,8 +105,30 @@ public void onAttestationAggregationDue(final UInt64 slot) {
...
+                  .collect(Collectors.toList());
+          validatorSlashingAction.perform(slashedPublicKeys);
+        });
+  }
 
   @Override
-  public void onProposerSlashing(final ProposerSlashing proposerSlashing) {}
+  public void onProposerSlashing(final ProposerSlashing proposerSlashing) {
+    delegates.forEach(delegates -> delegates.onProposerSlashing(proposerSlashing));
+    final UInt64 slashedIndex = proposerSlashing.getHeader1().getMessage().getProposerIndex();
```
zilm13: Let's start with `maybeValidatorSlashedAction.ifPresent(` to avoid doing anything when it's not present

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorTimingActions.java
```diff
@@ -97,8 +105,30 @@ public void onAttestationAggregationDue(final UInt64 slot) {
   }
 
   @Override
-  public void onAttesterSlashing(final AttesterSlashing attesterSlashing) {}
+  public void onAttesterSlashing(final AttesterSlashing attesterSlashing) {
+    delegates.forEach(delegates -> delegates.onAttesterSlashing(attesterSlashing));
+    final Set<UInt64> slashedIndices = attesterSlashing.getIntersectingValidatorIndices();
```
zilm13: please put this one inside `ifPresent` block too

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
zilm13: Still no full-stop in the end

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/DoppelgangerDetectionShutDown.java
```diff
@@ -11,21 +11,24 @@
...
-            .map(BLSPublicKey::toAbbreviatedString)
-            .collect(Collectors.joining(", ")));
-    System.exit(1);
+  public void perform(final List<BLSPublicKey> pubKeys) {
+    if (!pubKeys.isEmpty()) {
+      STATUS_LOG.exitOnDoppelgangerDetected(
+          pubKeys.stream()
+              .map(BLSPublicKey::toAbbreviatedString)
+              .collect(Collectors.joining(", ")));
+      System.exit(1);
```
zilm13: we probably need exit code 2 here too

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashingRiskDetectionAction.java
```diff
@@ -11,12 +11,16 @@
...
+package tech.pegasys.teku.validator.client.slashingriskactions;
 
 import java.util.List;
 import tech.pegasys.teku.bls.BLSPublicKey;
 
-public interface DoppelgangerDetectionAction {
+public interface SlashingRiskDetectionAction {
 
-  void perform(final List<BLSPublicKey> doppelgangers);
+  void perform(final List<BLSPublicKey> pubKeys);
```
zilm13: final not needed here

zilm13 (commented):
+ Let's rename ValidatorIndexProvider -> OwnedValidatorIndexProvider for clarity
+ Please also please open issue for acceptance test or you could just post draft PR

CHANGELOG.md
mehdi-aouadi: Rebasing joy! fixed

CHANGELOG.md
mehdi-aouadi: Rebasing issue! Fixed

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorTimingActions.java
mehdi-aouadi: Oh right! Updated

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorTimingActions.java
mehdi-aouadi: Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/DoppelgangerDetectionShutDown.java
mehdi-aouadi: Yeah that would be more appropriate. Done

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashingRiskDetectionAction.java
mehdi-aouadi: Removed

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
mehdi-aouadi: Right, renamed it back to `stop-vc-when-validator-slashed-enabled`

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
mehdi-aouadi: My bad, added

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorTimingActions.java
mehdi-aouadi: Added tests too

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/DoppelgangerDetectionShutDown.java
mehdi-aouadi: Updated the documentation to indicate that too in my documentation PR: https://github.com/Consensys/doc.teku/pull/526

zilm13 (approved):
LGTM

CHANGELOG.md
```diff
@@ -16,5 +16,5 @@ the [releases page](https://github.com/Consensys/teku/releases).
 - Added a hidden flag `--Xfork-choice-updated-always-send-payload-attributes` which would cause
 payload attributes to be calculated and sent with every fcU. This could be useful for builders
 consuming the `payload_attributes` SSE events.
-
+- Added `--stop-vc-when-validator-slashed` option to stop the Validator Client when any owned validator is slashed (requires a BN that publishes the `attester_slashing` and `proposer_slashing` SSE events)
```
rolfyone: naming of cli arg inconsistent - `--stop-vc-when-validator-slashed-enabled`

data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
```diff
@@ -178,37 +178,40 @@ private SafeFuture<Optional<SubmitDataError>> addBlsToExecutionChange(
             });
   }
 
-  public void subscribeToReceivedBlocks(ImportedBlockListener listener) {
+  public void subscribeToReceivedBlocks(final ImportedBlockListener listener) {
```
rolfyone: unrelated things like this we can do as trivial PRs and not pollute PRs... keep that in mind next time? makes easier to review.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/sentry/SentryBeaconNodeApi.java
```diff
@@ -161,7 +161,8 @@ public static BeaconNodeApi create(
             validatorTimingChannel,
             serviceConfig.getMetricsSystem(),
             validatorConfig.generateEarlyAttestations(),
-            spec);
+            spec,
+            validatorConfig.isStopVcWhenValidatorSlashedEnabled());
```
rolfyone: i'd generally prefer spec being last, but thats just a nit

rolfyone:
only minor things - may be worth a demo to show it working...

CHANGELOG.md
mehdi-aouadi: My bad, must be rebase issue. Fixed

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/sentry/SentryBeaconNodeApi.java
mehdi-aouadi: Done

data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
mehdi-aouadi: Couldn't resist :) Left it for another PR...

mehdi-aouadi:
@rolfyone and @zilm13 fyi, I pushed these lasts changes: https://github.com/Consensys/teku/pull/7826/commits/d99c2fa923f5845ede483a1eb7ff37e711fe5adf in order to handle the use case when the BN/VC are running in a single process and thus not using SSE. I tested both use cases and they both work (remote BN and single process VC/BN)
I had a discussion with @tbenr and we were thinking that it might be cleaner to move the slashing actions to a separate channel (other than `ValidatorTimingChannel`, something like `ValidatorSlashingChannel`) or we could leave as is...
Let me know if you have any thought otherwise I could demo the current implementation and then we merge it

CHANGELOG.md
```diff
@@ -24,5 +24,6 @@ the [releases page](https://github.com/Consensys/teku/releases).
 - Added Deneb (aka Dencun) configuration for Chiado network for epoch 516608 (2024-01-31 18:15:40 UTC).
 - Added Deneb (aka Dencun) configuration for Holesky network for epoch 29696 (2024-02-07 11:34:24 UTC).
 - Generate key at `—p2p-private-key-file` path if specified file doesn't exist.
+- Added `--stop-vc-when-validator-slashed-enabled` option to stop the VC when a validator is slashed
```
rolfyone: If i understand this PR correctly, this is not the final part of this feature... changelog may be premature.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -152,6 +153,16 @@ public class ValidatorOptions {
   private boolean isLocalSlashingProtectionSynchronizedEnabled =
       DEFAULT_VALIDATOR_IS_LOCAL_SLASHING_PROTECTION_SYNCHRONIZED_ENABLED;
 
+  @Option(
+      names = {"--stop-vc-when-validator-slashed-enabled"},
+      paramLabel = "<BOOLEAN>",
+      description =
+          "If an owned validator key is detected as slashed, the validator-client should terminate. In this case, the service should not be restarted.",
```
rolfyone: we need something around this - either exits with code 2 should not be restarted, or something... how do they know how to 'not restart'?

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
```diff
@@ -0,0 +1,32 @@
...
+import tech.pegasys.teku.bls.BLSPublicKey;
+
+public class SlashedValidatorShutDown implements SlashingRiskDetectionAction {
+
+  @Override
+  public void perform(List<BLSPublicKey> pubKeys) {
+    if (!pubKeys.isEmpty()) {
+      STATUS_LOG.validatorSlashedAlert(
+          pubKeys.stream().map(BLSPublicKey::toAbbreviatedString).collect(Collectors.toSet()));
+      System.exit(2);
```
rolfyone: - Split out the exit from the perform, and have it on the interface as something like
 ```
default void die() {
   System.exit(2);
}
```
This allows the test case to just do a noop on die, so we could then write tests.
- once that is split out, we can actually test this function and make sure the log happens
- we can also make sure the keys are displayed in a useful way. in this instance, i would argue print the full key not abbreviated key. This will be our best point to see the error message is correct when things get refactored in future.

File: infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java [hunk 22]
```diff
@@ -246,6 +246,12 @@ public void exitOnNoValidatorKeys() {
...
+        "Validator(s) with public key(s) {} got slashed. Shutting down validator client...",
```
rolfyone: this implies its only on VC? so if we're running a full BN/VC we can't use this new logic?

rolfyone:
> Let me know if you have any thought otherwise I could demo the current implementation and then we merge it

My main thought is theres a lot of renaming in this and it makes it a lot harder than it needs to be to review. this kind of cleanup should be kept outside of substantive changes, it's just good practice and probably would have made this not look like a 'big pr' that i hate so much.
I'm not sure that the feature is 'done' given the code here, and would like to see some testing around a few things, and agree with @zilm13 that an acceptance test will be important.
A demo will make me more confident without setting it all up to see exactly what's going on, and we need to make sure we're clear about what we're asking in usage, which i briefly noted.
I think given the outstanding bits, we might want to mark the cli arg as hidden, remove the change entry, and merge this part, then unhide the entry once we're actually done and at that point we can add a change entry. We definitely want to hold off merging until after the release is done.

rolfyone (changes requested):
as outlined, lets come up with a plan for this feature.

CHANGELOG.md
mehdi-aouadi: Actually the feature is now complete. There was a missing part (BN/VC running in a single process) and it's implemented now

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
mehdi-aouadi: Done:
- Refactored the `SlashedValidatorAlert` to made it extend the `SlashedValidatorShutDown` and override the `shutDown` method to test the actual `SlashedValidatorShutDown`
- Print the hex representation of the public keys instead of the abbreviation

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
mehdi-aouadi: Updated it: Removed `validator client`
It depends if the VC is running separately (in that case only the VC is shut down) or if VC/BN are running in a single process (in that case the Teku node is shut down)

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
mehdi-aouadi: I renamed the option to `shut-down-when-validator-slashed-enabled ` to make it clear that this could shut down the VC if run separately or both VC and BN if running in a single process.
I also added a mention of the exit code.
PS: There is a documentation PR with more details about the feature https://github.com/Consensys/doc.teku/pull/526

mehdi-aouadi:
> > Let me know if you have any thought otherwise I could demo the current implementation and then we merge it
> 
> My main thought is theres a lot of renaming in this and it makes it a lot harder than it needs to be to review. this kind of cleanup should be kept outside of substantive changes, it's just good practice and probably would have made this not look like a 'big pr' that i hate so much. I'm not sure that the feature is 'done' given the code here, and would like to see some testing around a few things, and agree with @zilm13 that an acceptance test will be important. A demo will make me more confident without setting it all up to see exactly what's going on, and we need to make sure we're clear about what we're asking in usage, which i briefly noted. I think given the outstanding bits, we might want to mark the cli arg as hidden, remove the change entry, and merge this part, then unhide the entry once we're actually done and at that point we can add a change entry. We definitely want to hold off merging until after the release is done.

Few remarks here:
- The feature is now complete. We now support both separate VC and single process VC/BN
- For clarity, I renamed the option to `shut-down-when-validator-slashed` to make it clear that we shut down whatever we're running (separate VC or a VC/BN node). This is explained in the documentation PR in more details
- To make things clear, this feature works in both cases: separate VC (the VC is shut down) and VC/BN in a single process (the whole Teku node is shutdown). There is a documentation PR for more details since we can't put all of them in the option description. Please check https://github.com/Consensys/doc.teku/pull/526
- There is only one renaming/refactoring of the `SlashingRiskDetectionAction` in order to reuse what has been already done for the doppelganger detection. There are existing tests for the refactored files already and that hasn't been changed
- There is a separate PR to add an acceptance test: https://github.com/Consensys/teku/pull/7852
- There isn't a detailed spec for this feature so the implemented behaviour could be changed if needed
- I will do a demo to show how this feature works

zilm13:
I think it would be valuable to have AT tests both for separate VC and single node BN+VC as far as we found it's different cases

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -152,6 +153,16 @@ public class ValidatorOptions {
   private boolean isLocalSlashingProtectionSynchronizedEnabled =
       DEFAULT_VALIDATOR_IS_LOCAL_SLASHING_PROTECTION_SYNCHRONIZED_ENABLED;
 
+  @Option(
+      names = {"--shut-down-when-validator-slashed-enabled"},
+      paramLabel = "<BOOLEAN>",
+      description =
+          "If an owned validator key is detected as slashed, the validator-client or the teku node (if running the beacon-node and the validator-client in a single process) should terminate with exit code 2. In this case, the service should not be restarted.",
```
rolfyone, line 160: if we want to trim this, can say
```
If an owned validator key is detected as slashed, the node should terminate with exit code 2. In this case, the service should not be restarted.
```

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -152,6 +153,16 @@ public class ValidatorOptions {
   private boolean isLocalSlashingProtectionSynchronizedEnabled =
       DEFAULT_VALIDATOR_IS_LOCAL_SLASHING_PROTECTION_SYNCHRONIZED_ENABLED;
 
+  @Option(
+      names = {"--shut-down-when-validator-slashed-enabled"},
```
rolfyone, line 157: Generally we have a test showing these enabled / not enabled, sometimes checking default value...
That test is not present here. It would be beneficial to add 2 test cases, 1 showing it being set to true, the other checking default setting.

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
rolfyone: last part of that was test?

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
rolfyone: saying shutting down is what we generally say, thats fine.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -642,6 +643,11 @@ protected void initAttesterSlashingPool() {
                 .reversed());
     blockImporter.subscribeToVerifiedBlockAttesterSlashings(attesterSlashingPool::removeAll);
     attesterSlashingPool.subscribeOperationAdded(forkChoice::onAttesterSlashing);
+    final ValidatorTimingChannel validatorTimingChannel =
+        eventChannels.getPublisher(ValidatorTimingChannel.class);
+    attesterSlashingPool.subscribeOperationAdded(
+        (operation, validationStatus, fromNetwork) ->
+            validatorTimingChannel.onAttesterSlashing(operation));
```
rolfyone, line 650: we only want this if we're using this feature, otherwise we're permanently listening locally to the slow version of the message?

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -654,6 +660,11 @@ protected void initProposerSlashingPool() {
             beaconBlockSchemaSupplier.andThen(BeaconBlockBodySchema::getProposerSlashingsSchema),
             validator);
     blockImporter.subscribeToVerifiedBlockProposerSlashings(proposerSlashingPool::removeAll);
+    final ValidatorTimingChannel validatorTimingChannel =
+        eventChannels.getPublisher(ValidatorTimingChannel.class);
+    proposerSlashingPool.subscribeOperationAdded(
+        (operation, validationStatus, fromNetwork) ->
+            validatorTimingChannel.onProposerSlashing(operation));
```
rolfyone, line 667: same re. being only relevant for this feature i believe..

rolfyone:
> * The feature is now complete. We now support both separate VC and single process VC/BN

remarks / conversation welcome. imo this feature is yet to meet our definition of done, and we can continue on this path this time, but in future (im sure i said that last time) i have a very strong preference for not following this methodology. I'm not sure I can make that any clearer at this point.

rolfyone:
My suggestion:
 - hide cli arg (-X, undocumented)
 - remove changelog entry until acceptance test is ready to merge at least
 - get this tested enough to accept a merge so we can move past this huge pr, and we finish the feature when it is fully acceptance tested.
 
Alternatively use this as a starting point, break it down, as previously suggested.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
mehdi-aouadi: Addressed here [#7915](https://github.com/Consensys/teku/pull/7915/files#diff-0cb84436c57ddc4deb3a4ea450042bc95eedcdc548fbc405453511266fa93773)

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
mehdi-aouadi: Done here [#7915 ](https://github.com/Consensys/teku/pull/7915/files#diff-7158a24db1f834c9277b58ee543560dab5f6a24b1cb284854e31f6b3733dfe83)

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
mehdi-aouadi: Good catch. Fixed here [7918](https://github.com/Consensys/teku/pull/7918/files#diff-a0738cefef0ee011967c18c1fe03e9c169ab10cfc7a1e7531f31cd71505b7a39)

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
mehdi-aouadi: Same, fixed here [7918](https://github.com/Consensys/teku/pull/7918/files#diff-a0738cefef0ee011967c18c1fe03e9c169ab10cfc7a1e7531f31cd71505b7a39)

validator/client/src/main/java/tech/pegasys/teku/validator/client/slashingriskactions/SlashedValidatorShutDown.java
mehdi-aouadi: There is a test for ValidatorTimingActions where this class is tested. Check [7916](https://github.com/Consensys/teku/pull/7916/files#diff-a37ab63dcebb7b505f1e7805107f46b669deb78544a395726a33b66de4c92268)

mehdi-aouadi:
I ended up breaking this PR down to 8 different ones: 
1 #7913 
2 #7914 
3 #7915 
4 #7916 
5 #7917
6 #7918
7 #7919 
8 #7921

PR #7829, opened 2023-12-14
Fix exit when no validator keys

tbenr:
Since initialization is async, I ended up doing the same thing we do on doppleganger detection.

Actually acceptance tests were testing the bug! Interop keys were loaded and node was still exiting.

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -218,6 +218,13 @@ public static ValidatorClientService create(
             () -> validatorClientService.initializeValidators(validatorApiChannel, asyncRunner))
         .thenCompose(
             __ -> {
+              if (validatorConfig.isExitWhenNoValidatorKeysEnabled()
+                  && validatorLoader.getOwnedValidators().getActiveValidators().isEmpty()) {
+                LOG.info(
```
zilm13: Maybe error? Also this message is not displayed in console if console log option is not on. Maybe use StatusLogger?

tbenr: humm I was doing the same as `DoppelgangerDetectionShutDown`.. maybe we could migrate everithing to status logger. Not sure it is an error though...

zilm13 (approved):
LGTM

PR #7830, opened 2023-12-14
Add exit on no validator keys loaded if configured

zilm13:
## PR Description
In case of exception.
Example case:
```
2023-12-14 09:34:47.787+00:00 | validator-async-0 | WARN  | ValidatorClientService | Invalid configuration. Password file for keystore /validator-keys/node-1-keystores/nimbus-keys/0xac97a570a795d24af13ef32709d71a37c0fb90a49581e735d635721581c5553c10eb15f43531121b6026f55d603d73f7/keystore.json either doesn't exist or isn't readable. Expected to find it at /validator-keys/node-1-keystores/secrets/0xac97a570a795d24af13ef32709d71a37c0fb90a49581e735d635721581c5553c10eb15f43531121b6026f55d603d73f7/keystore.txt
2023-12-14 09:34:47.789+00:00 | main | INFO  | MetricsHttpService | Starting metrics http service on 0.0.0.0:5064
```
Shouldn't be unconditional because of things like #7422 

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
@@ -220,11 +220,7 @@ public class ValidatorClientService extends Service {
             () -> validatorClientService.initializeValidators(validatorApiChannel, asyncRunner))
         .thenCompose(
             __ -> {
-              if (validatorConfig.isExitWhenNoValidatorKeysEnabled()
-                  && validatorLoader.getOwnedValidators().getActiveValidators().isEmpty()) {
-                STATUS_LOG.exitOnNoValidatorKeys();
-                System.exit(2);
-              }
+              checkNoKeysLoaded(validatorConfig, validatorLoader);
 
               if (validatorConfig.isDoppelgangerDetectionEnabled()) {
                 validatorClientService.initializeDoppelgangerDetector(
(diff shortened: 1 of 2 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -274,13 +270,23 @@ public static ValidatorClientService create(
...
             })
         .always(() -> LOG.trace("Finished starting validator client service."));
 
     return validatorClientService;
   }
 
+  private static void checkNoKeysLoaded(
+      final ValidatorConfig validatorConfig, final ValidatorLoader validatorLoader) {
+    if (validatorConfig.isExitWhenNoValidatorKeysEnabled()
+        && validatorLoader.getOwnedValidators().getActiveValidators().isEmpty()) {
```
tbenr: As reasoned in chat, I think we should use `validatorLoader.getOwnedValidators().hasNoValidators()` here, to catch 0 loaded keys independently by their status.
In short: in my opinion VC should not exit if successfully loads keys their deposits have not been processed yet.

tbenr (commented):
Btw it looks good to me, just that change.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
rolfyone: agree, just getOwnedValidators checked here, not active status please...

rolfyone (approved):
LGTM

PR #7832, opened 2023-12-14
Rename OwnedValidators.activeValidators in favor of "enabled"

zilm13:
## PR Description
We have  `OwnedValidators.getActiveValidators()` which spreads to several other classes and it's often confused with [ValidatorStatus](https://github.com/ConsenSys/teku/blob/f31b239f7788f07d8e9c2b6e7544c5e16b01a259/data/serializer/src/main/java/tech/pegasys/teku/api/response/v1/beacon/ValidatorStatus.java) while it means only whether validator is managed by node or not. Renamed to avoid confusion in the future

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/OwnedKeyManager.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/OwnedKeyManager.java
@@ -44,41 +44,36 @@ import tech.pegasys.teku.validator.client.restapi.apis.schema.ExternalValidator;
 import tech.pegasys.teku.validator.client.restapi.apis.schema.ImportStatus;
 import tech.pegasys.teku.validator.client.restapi.apis.schema.PostKeyResult;
 
-public class EnabledKeyManager implements KeyManager {
+public class OwnedKeyManager implements KeyManager {
   private static final String EXPORT_FAILED =
       "{\"metadata\":{\"interchange_format_version\":\"5\"},\"data\":[]}";
   private static final Logger LOG = LogManager.getLogger();
   private final ValidatorLoader validatorLoader;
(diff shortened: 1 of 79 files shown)


tbenr:
To me seems like this is just `getValidators()`. `enabled` or everything else would be justified if we could get validators that are "not" enabled. But to me seems like `OwnedValidators` can just have or not have a validator.

StefanBratanov:
I think i prefer `owned` instead of `enabled` so method will be `OwnedValidators.getValidators()` and then `OwnedKeysManager` and in comments etc.

zilm13:
@StefanBratanov yeah, now looks better to me. Tried to move this way.

rolfyone (approved):
Good improvement!

validator/client/src/main/java/tech/pegasys/teku/validator/client/OwnedKeyManager.java
```diff
@@ -44,41 +44,36 @@
...
-  public ActiveKeyManager(
+  public OwnedKeyManager(
       final ValidatorLoader validatorLoader, final ValidatorTimingChannel validatorTimingChannel) {
     this.validatorLoader = validatorLoader;
     this.validatorTimingChannel = validatorTimingChannel;
   }
 
-  /**
-   * Get a listing of active validator keys
-   *
```
StefanBratanov, line 62: Just this comment needs changing I think

StefanBratanov (approved):

validator/client/src/main/java/tech/pegasys/teku/validator/client/OwnedKeyManager.java
zilm13: We agreed that just removing it is ok

PR #7842, opened 2023-12-18
Fallback to GET validators from state on 400 POST response

StefanBratanov:
## PR Description
We are handling fallback to `GET` validators from state on status codes 404 and 405 in https://github.com/Consensys/teku/pull/7786, but LH seems to be returning 400, so handling that case as well, which should cover all scenarios.

```
│ teku java.util.concurrent.CompletionException: tech.pegasys.teku.validator.remote.FailoverRequestException: Remote request (get_validator_statuses) failed on all configured Beacon Node endpoints 
│ teku ...: java.lang.IllegalArgumentException: Invalid params response from Beacon Node API (url = ...)
│ ...:5051/eth/v1/beacon/states/head/validators, status = 400, message = BAD_REQUEST: Unsupported endpoint version: v1)
```

Also add flag `--Xuse-post-validators-endpoint-enabled` until all clients implement this endpoint.
## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,6 @@ the [releases page](https://github.com/Consensys/teku/releases).
 ### Breaking Changes
 
 ### Additions and Improvements
-- Fix incompatibility between Teku validator client and beacon nodes whose `/eth/v1/beacon/states/{state_id}/validators` POST endpoint is returning status code 400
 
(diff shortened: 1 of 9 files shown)


zilm13 (approved):
LGTM

Issue #7718, opened 2023-11-16
Implement headerParam in `EndpointMetaData`


tbenr:
We don't currently support header param in our `EndpointMetadata` framework.

https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlockV2

![image](https://github.com/Consensys/teku/assets/15999009/dab4579e-9947-40bc-98ed-467b30acfc28)

PR #7844, opened 2023-12-19
Implement headerParam in `EndpointMetaData`

courtneyeh:
#7718 

## Documentation

## Changelog


infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -484,6 +505,14 @@ public EndpointMetaDataBuilder queryListParam(final ParameterMetadata<?> paramet
       return this;
     }
 
+    public EndpointMetaDataBuilder header(final ParameterMetadata<?> headerMetadata) {
```
lucassaldanha, line 508: Some Beacon API methods have optional headers. We should be able to define a header as optional in the metadata.

lucassaldanha: Regardless if the header is required or not, we probably don't want to be able to specify multiple headers with the same name. So we need to check on both.

infrastructure/restapi/src/testFixtures/java/tech/pegasys/teku/infrastructure/restapi/StubRestApiRequest.java
```diff
@@ -57,6 +57,7 @@ public class StubRestApiRequest implements RestApiRequest {
   private final Map<String, String> optionalQueryParameters = new HashMap<>();
   private final Map<String, List<String>> listQueryParameters = new HashMap<>();
   private final Map<Integer, String> contentTypeMap = new HashMap<>();
+  private final Map<String, String> contextHeaders = new HashMap<>();
```
lucassaldanha: I think a better naming would be `requestHeaders` to represent the headers we are reading from the incoming request, and `responseHeaders` to represent the headers we are adding to the response.

infrastructure/restapi/src/testFixtures/java/tech/pegasys/teku/infrastructure/restapi/StubRestApiRequest.java
```diff
@@ -235,18 +236,31 @@ public <T> List<T> getQueryParameterList(ParameterMetadata<T> parameterMetadata)
...
   }
 
   @Override
   public void startEventStream(Consumer<SseClient> clientConsumer) {
     throw new UnsupportedOperationException();
   }
 
-  public String getHeader(String name) {
-    return headers.get(name);
+  public String getContextHeader(String name) {
```
lucassaldanha: As per my other comment, I'd call this `getResponseHeaders`

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/RestApiRequest.java
```diff
@@ -54,6 +54,10 @@ public interface RestApiRequest {
 
   <T> List<T> getQueryParameterList(final ParameterMetadata<T> parameterMetadata);
 
+  <T> Optional<T> getOptionalHeader(final ParameterMetadata<T> parameterMetadata);
+
+  <T> T getHeader(final ParameterMetadata<T> parameterMetadata);
```
lucassaldanha: NIT: If we want to differentiate from the headers we add to the response, we could name these methods `getRequestHeader` and `getOptionalRequestHeader`.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -378,6 +388,24 @@ private void writeParameters(
...
+      gen.writeObjectField("name", entry.getKey());
+      if (isMandatoryField) {
+        gen.writeObjectField("required", true);
+      }
+      gen.writeObjectField("in", "header");
+      gen.writeFieldName("schema");
+      entry.getValue().serializeOpenApiTypeOrReference(gen);
+      gen.writeEndObject();
+    }
+  }
```
lucassaldanha: It looks like this method is not necessary as we could use `writeParameters(..)` passing `paramUsedIn="header"`. Maybe check if we could reuse that method.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -484,6 +514,24 @@ public EndpointMetaDataBuilder queryListParam(final ParameterMetadata<?> paramet
       return this;
     }
 
+    public EndpointMetaDataBuilder requiredHeader(final ParameterMetadata<?> headerMetadata) {
```
lucassaldanha: NIT: Following `queryParam(..)` and `queryParamRequired(..)` naming, we would call this `headerRequired(..)`.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
```diff
@@ -244,6 +246,37 @@ public <T> List<T> getQueryParameterList(final ParameterMetadata<T> parameterMet
...
+              parameterMetadata.getName(), e.getMessage()));
+    }
+  }
+
+  @Override
+  public <T> T getHeader(final ParameterMetadata<T> parameterMetadata) {
+    try {
+      return parameterMetadata
+          .getType()
+          .deserializeFromString(headerMap.get(parameterMetadata.getName()));
```
lucassaldanha, line 272: If the headerMap does not contain the key that we are looking for it will throw a NullPointerException. This is an example of a bug that would be caught with unit tests.

lucassaldanha (commented):
The PR is in the right direction but it does not have tests for the `EndpointMetadata` changes or the `JavalinRestApiRequest`.

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/JavalinRestApiRequest.java
courtneyeh: Methods like `getPathParameter` would have a similar issue. Do I need to update those to check for this too?

lucassaldanha (approved):

PR #7848, opened 2023-12-20
Fix rest api bind error message

lucassaldanha:
## PR Description

When Javalin can't bind to a port that is already in use. The Javalin `app` object will not contain a port, it will have it as -1.
Because of this, the error message that we are getting looks like:

```
TCP Port -1 is already in use. You may need to stop another process or change the HTTP port for this process.
```

To fix this, we can use Javalin's `JavalinBindException` message that already has the port. It looks like: 

```
Port already in use. Make sure no other process is using port 5000 and try again.
```

## Fixed Issue(s)
N/A

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7850, opened 2023-12-22
Fix post block endpoints returning 202 in case of unexpected errors

StefanBratanov:
## PR Description
Fixes following:

- Teku fails to broadcast the block but still returns a 202 status code to the VC, which the VC inteprets to mean "invalid but published". IMO this is a minor Teku bug, as Teku should have returned a 400 when it couldn't unblind the block: https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/ValidatorRequiredApi/publishBlindedBlockV2.

The PR makes will make unblinding (and any other unexpected error) return 500. Unblinding is done before validation, so 500 makes more sense in my opinion.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,3 +16,4 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Bug Fixes
 - Fix incompatibility between Teku validator client and Lighthouse beacon nodes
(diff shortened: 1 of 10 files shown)


zilm13 (approved):
LGTM

PR #7852, opened 2023-12-22
7707 at vc slashing

mehdi-aouadi:
## PR Description
Add acceptance test for https://github.com/Consensys/teku/pull/7826

## Fixed Issue(s)

#7707 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Additions and Improvements
 - Add `proposer_slashing`, `attester_slasing` and `payload_attibutes` (only when a loaded validator will be producing a block) events to the `/eth/v1/events` SSE stream
+- Added `--stop-vc-when-validator-slashed` option to stop the VC when a validator is slashed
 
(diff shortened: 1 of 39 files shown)


zilm13:
I've posted a fix: https://github.com/Consensys/teku/pull/7852/commits/75ffd9414e17981c6c7e8f961cb28f8291ff469e
What was the issue:
You recreate spec from the blank, while for real node you were using `altairForkEpoch = 100` (0 in original swift config). That's why in prepared slashing message you were using altair fork version, while you were running phase0 on nodes. This update fixes it, slashing is posted, validators are changing their status from `active_ongoing` to `active_slashed` but the test is not completed, I've not dig why, leaving it to you.

rolfyone:
Swift is generally hard to get running, if you use less-swift it may make life easier. also we'll want a combined BN /VC being slashed remotely and shutting down as a second test or as part of this test setup.... eg if theres 3 participants and 2 get shutdown, or similar

mehdi-aouadi:
Replaced by https://github.com/Consensys/teku/pull/7921
Will make the requested changes there

PR #7855, opened 2024-01-03
SszOptional as of EIP-6475

zilm13:
## PR Description

https://eips.ethereum.org/EIPS/eip-6475
Tests fixtures were taken from EIP (slightly modified)
Prerequisite for Verkle trees implementation

## Documentation

## Changelog


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/SszOptionalSchemaImpl.java
```diff
@@ -0,0 +1,277 @@
...
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNodeSource;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNodeSource.CompressedBranchInfo;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNodeStore;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeUtil;
+
+/**
+ * SszOptional schema according to the <a
+ * href="https://eips.ethereum.org/EIPS/eip-6475">EIP-6475</a>
+ */
+public abstract class SszOptionalSchemaImpl<ElementDataT extends SszData>
```
Nashatyrev: nit: Do we really need `abstract` class here?

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/SszOptionalSchemaImpl.java
```diff
@@ -0,0 +1,277 @@
...
+    implements SszOptionalSchema<ElementDataT, SszOptional<ElementDataT>> {
+
+  private static final byte IS_PRESENT_PREFIX = 1;
+  private static final int PREFIX_SIZE_BYTES = 1;
+
+  private static LeafNode createOptionalNode(final boolean isPresent) {
+    return isPresent ? LeafNode.create(Bytes.of(IS_PRESENT_PREFIX)) : LeafNode.ZERO_LEAVES[1];
```
Nashatyrev: nit: I believe it would be a bit clearer if define two constants 
```java
static final LeafNode EMPTY_OPTIONAL_LEAF = LeafNode.create(Bytes.of(0));
static final LeafNode PRESENT_OPTIONAL_LEAF = LeafNode.create(Bytes.of(1));
```

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/SszOptionalSchemaImpl.java
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  private final SszSchema<ElementDataT> childSchema;
+  private final TreeNode defaultTree;
+  private final Supplier<SszLengthBounds> lengthBounds =
+      Suppliers.memoize(this::calcSszLengthBounds);
+
+  public SszOptionalSchemaImpl(final SszSchema<ElementDataT> childSchema) {
+    this.childSchema = childSchema;
+    this.defaultTree = createOptionalNode(LeafNode.ZERO_LEAVES[0], false);
```
Nashatyrev: nit: `LeafNode.EMPTY_LEAF` looks clearer imo
Also I would make `default_tree` a constant as it doesn't depend on the child schema

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/SszOptionalSchemaImpl.java
```diff
@@ -0,0 +1,277 @@
...
+  @Override
+  public boolean isPresent(final TreeNode node) {
+    return getIsPresentFromOptionalNode(getOptionalNode(node));
+  }
+
+  private TreeNode getOptionalNode(final TreeNode node) {
+    return node.get(GIndexUtil.RIGHT_CHILD_G_INDEX);
+  }
+
+  private TreeNode createOptionalNode(final TreeNode valueNode, final boolean isPresent) {
```
Nashatyrev: nit: The method name is the same as static one for creating a different node. I would disambiguate them

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/SszOptionalSchema.java
```diff
@@ -0,0 +1,47 @@
...
+import tech.pegasys.teku.infrastructure.ssz.SszData;
+import tech.pegasys.teku.infrastructure.ssz.SszOptional;
+import tech.pegasys.teku.infrastructure.ssz.schema.impl.SszOptionalSchemaImpl;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNode;
+
+public interface SszOptionalSchema<
+        ElementDataT extends SszData, SszOptionalT extends SszOptional<ElementDataT>>
+    extends SszSchema<SszOptionalT> {
+
+  static <T extends SszData> SszOptionalSchemaImpl<T> create(final SszSchema<T> childSchema) {
```
Nashatyrev: nit: I would better declare `SszOptionalSchema` return type

infrastructure/ssz/src/test/java/tech/pegasys/teku/infrastructure/ssz/schema/SszSchemaTestBase.java
```diff
@@ -45,6 +45,10 @@ void getDefaultTree_shouldBeEqualToDefaultStructure(SszSchema<SszData> schema) {
   @MethodSource("testSchemaArguments")
   @ParameterizedTest
   void sszDeserialize_tooLongSszShouldFailFastWithoutReadingWholeInput(SszSchema<SszData> schema) {
+    if (schema instanceof SszOptionalSchema<?, ?>) {
+      // empty SszOptional couldn't pass this test
```
Nashatyrev, line 49: Not sure why the SszOptional is the special case here?

Nashatyrev (approved):
Looks good!
Just some minor nits and the question about skipped test

Nashatyrev:
Also thought that `SSzOptional` might potentially be derived from the `SszComposite`, but I don't think its very important

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/SszOptionalSchemaImpl.java
zilm13: Not really, updated

infrastructure/ssz/src/test/java/tech/pegasys/teku/infrastructure/ssz/schema/SszSchemaTestBase.java
zilm13: The test adds to the reader buffer some random byte sequence. It expects you to break up reading buffer before consuming any of the added data. 1 more byte read - failed. Empty serialized SszOptional is byte[0]. So we should somehow guess that we have empty SszOptional with non-zero bytes buffer provided without reading any extra of correctly serialized (0) bytes, which is impossible for this type with EIP-6475 proposed  serialization.

zilm13:
> Also thought that SSzOptional might potentially be derived from the SszComposite

Yeah, that was my first attempt to implement this, and I failed to complete it before my head explodes. Starting from the blank was more productive. Let's keep it as is.

infrastructure/ssz/src/test/java/tech/pegasys/teku/infrastructure/ssz/schema/SszSchemaTestBase.java
Nashatyrev: Oh, got it 👍

PR #7858, opened 2024-01-04
Refactored concurrent local slashing protection

rolfyone:
- broke out to its own structures
 - made implementation clearer
 - consolidated objects to use a single map and lock

 This removes the need for several concurrent maps through abstraction, makes it easier to understand the flow, and removes a lot of the complicated flows in the initial implementation.

 The configuration is also simpler, with instantiation of the object taking place in the client service, clearly separating the 2 implementations.

 Default is still to use the old functionality for now.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/LocalSlashingProtectorConcurrentAccess.java
```diff
@@ -0,0 +1,136 @@
...
+      record.ifPresent(LocalSlashingProtectionRecord::unlock);
+    }
+  }
+
+  LocalSlashingProtectionRecord acquireLock(
+      final BLSPublicKey validator, final Bytes32 genesisValidatorsRoot) {
+    final LocalSlashingProtectionRecord entry =
+        records.computeIfAbsent(validator, __ -> addRecord(validator, genesisValidatorsRoot));
+    // essential that we take a lock here
+    entry.lock();
```
zilm13: I am thinking how to improve this manual locking to be safer in usage, and may be move it out of this method? So how about calling it smth like `getRecord` and do record.lock() just after `getRecord` call, so it's in the same method where we do `finally {unlock}`. We have just 2 usages and a bit duplication should not be an issue.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/LocalSlashingProtectorConcurrentAccess.java
```diff
@@ -0,0 +1,136 @@
...
+  public Optional<ValidatorSigningRecord> getSigningRecord(final BLSPublicKey validator) {
+    final Optional<LocalSlashingProtectionRecord> record =
+        Optional.ofNullable(records.get(validator));
+    if (record.isEmpty()) {
+      // not loaded yet, just get from file if available, can create structure on use.
+      return getValidatorSigningRecordFromFile(validator);
+    }
+    try {
+      // unlike other occasions, we're not creating if absent, so only lock if present
+      record.ifPresent(LocalSlashingProtectionRecord::lock);
```
zilm13: I think these 3 lines (starting from this one) would look cleaner if we do record.get() before them.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/signatures/LocalSlashingProtectorConcurrentAccess.java
rolfyone: the whole point of the `acquireLock` method is to get the object and acquire the lock. this is the same as the previous PR.
There's now an exception which wasn't in the last PR, but i think i'd prefer to leave it here because that's what the method is designed to do (everything will do acquireLock, then finally unlock)

I'll have a little play, but at the moment this is the cleanest i've come up with imo...

rolfyone: I renamed the function, it's more consistent so we can run with that - makes the biggest impact in test cases, so thats fine.

lucassaldanha (approved):

PR #7859, opened 2024-01-04
Implement header parameter for PostBlockV2

courtneyeh:
#7718 

## Documentation

## Changelog


tbenr:
@courtneyeh @lucassaldanha following up in the conversation, a cleaner solution for the require enforcement could be to add a `validation` method in the `EndpointMetadata` and structurally call it in the framework passing it the request context. So it can check all required params and headers (`requiredQueryParams`, `requiredHeaders`) and fail accordingly.

This would be better than manually do the checks on each `handleRequest` and enforce consistency.

WDYT?

tbenr:
but the ticket was more related to the "open API" swagger UI rather than runtime checks on the required fields. So it could be a separated ticket (and for now leave this manually enforced or even not enforced at all)

courtneyeh:
Implementing the new header metadata parameter will be done in a seperate ticket.

PR #7861, opened 2024-01-05
Fix payload_attributes streaming event notification

### LINKED ISSUE #7847: Improve `payload_attributes` event frequency
### Description
One of the goals of introducing `payload_attributes` event was to help MEV builders to start constructing new payload as early as possible. Builder could be called on any slot so such systems needs new `payload_attributes` on every fcu (see [comment](https://github.com/ethereum/beacon-APIs/pull/305#issuecomment-1445011581)). Following line was added to the spec:
```
The frequency at which this event is sent may depend on beacon node configuration.
```
Our current implementations sends `payload_attributes` event only when beacon node is called to produce block, so it cannot be used for similar job. We could enhance our imlementation to send `payload_attributes` event on every fcu, so it could be used for such cases. Calculating all fields is not free, so it makes sense to introduce command line options flag which will turn this feature on, so by default we will keep sending event only in our slots when we have proposal.

StefanBratanov:
## PR Description

Issue is before we were sending the `payload_attributes` event only on an edge case

- Remove `NewBlockBuildingSubscriber` and use the already existing `ForkChoiceUpdatedResultNotification` with added `PayloadAttributes`
- Only sent `payload_attributes` if fCu has been called

## Fixed Issue(s)
related to #7847 

## Documentation

## Changelog


tbenr (approved):
LGTM, just a 🔪

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ProposersDataManager.java
```diff
@@ -237,13 +237,15 @@ private Optional<PayloadBuildingAttributes> calculatePayloadBuildingAttributes(
         Optional.ofNullable(validatorRegistrationInfoByValidatorIndex.get(proposerIndex))
             .map(RegisteredValidatorInfo::getSignedValidatorRegistration);
 
+    Eth1Address feeRecipient = getFeeRecipient(proposerInfo, blockSlot);
```
tbenr: 🔪

StefanBratanov: deserve it, fixed

PR #7864, opened 2024-01-08
Notify fCu subscribers only when fCu will be sent to EL

StefanBratanov:
## PR Description
A little optimisation after internal discussion about https://github.com/Consensys/teku/pull/7861

Basically instead of sending fCu notification to subscribers every time we check if need to send fCu `SafeFuture<Optional....>>`, we can not send notifications by changing send fCu to `Optional<SafeFuture<....>>` 

This way, notifications will be sent only when fCu will be sent to EL, which is what all subscribers check anyways `ForkChoice` , `PayloadAttributes`

Rename `StubForkChoiceNotifier` to `NoopForkChoiceNotifier` . Essentially no need to send empty notifications, which is the current behaviour of the stub.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (commented):
~~I generally like it but from the event stream this is not equivalent.~~

~~The side effects are:~~
~~1. wait the result before sending the event.~~
~~2. won't sent the event if the fcu engine call fails.~~

sorry I was wrong

tbenr (approved):
LGTM

PR #7869, opened 2024-01-10
Flattening deposit snapshot path config options

lucassaldanha:
## PR Description
This PR is part 1 of the deposit snapshots changes support for checkpoint sync url.

In this PR, we are breaking the deposit snapshot path option into 3 separate options:
- custom path set by the user
- path from our bundles distributed as part of Teku
- path for downloading the snapshot from the checkpoint sync url endpoint

At this point, we are not changing our existing behaviour:

1) default behaviour: using the bundled deposit tree snapshot by default

2) when using `--Xdeposit-snapshot`: don't use the bundled version and use the path specified on `--Xdeposit-snapshot`

## Fixed Issue(s)
related to #7715

## Documentation

## Changelog


services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
```diff
@@ -83,8 +91,16 @@ public Optional<UInt64> getDepositContractDeployBlock() {
...
 
-  public Optional<String> getDepositSnapshotPath() {
-    return depositSnapshotPath;
+  public Optional<String> getCustomDepositSnapshotPath() {
+    return customDepositSnapshotPath;
+  }
+
+  public Optional<String> getCheckpointSyncDepositSnapshotUrl() {
+    return checkpointSyncDepositSnapshotUrl;
+  }
```
lucassaldanha: This method is not being used atm. It will be relevant in my follow-up PR.

teku/src/main/java/tech/pegasys/teku/cli/options/DepositOptions.java
```diff
@@ -68,17 +69,35 @@ public class DepositOptions {
...
-                .depositSnapshotEnabled(parseDepositSnapshotEnabled()));
-  }
+    final CommandLine.ParseResult parseResult = commandSpec.commandLine().getParseResult();
+
+    final String checkpointSyncUrlCliOption = "--checkpoint-sync-url";
+    final boolean checkpointSyncUrlSet = parseResult.hasMatchedOption(checkpointSyncUrlCliOption);
 
-  private boolean parseDepositSnapshotEnabled() {
-    return depositSnapshotEnabled;
+    if (parseResult.hasMatchedOption("--Xdeposit-snapshot")) {
```
zilm13, line 77: I think the logic of lines starting from here could be implemented without using low level picocli api. We should have values set here already

teku/src/main/java/tech/pegasys/teku/cli/options/DepositOptions.java
```diff
@@ -68,17 +69,35 @@ public class DepositOptions {
...
-            b.eth1Endpoints(eth1Endpoints)
-                .eth1LogsMaxBlockRange(eth1LogsMaxBlockRange)
-                .useMissingDepositEventLogging(useMissingDepositEventLogging)
-                .depositSnapshotPath(depositSnapshotPath)
-                .depositSnapshotEnabled(parseDepositSnapshotEnabled()));
-  }
+    final CommandLine.ParseResult parseResult = commandSpec.commandLine().getParseResult();
+
+    final String checkpointSyncUrlCliOption = "--checkpoint-sync-url";
+    final boolean checkpointSyncUrlSet = parseResult.hasMatchedOption(checkpointSyncUrlCliOption);
```
zilm13: I don't like this way of using data from other options module. Is it possible to implement the same logic more casual way by checking options here and there in `TekuConfiguration.java`? As initialState could be set not only by "--checkpoint-sync-url" we could add the property checkpointSyncUrlSet as boolean to `Eth2NetworkConfiguration to clarify this case.

teku/src/main/java/tech/pegasys/teku/cli/options/DepositOptions.java
```diff
@@ -68,17 +69,35 @@ public class DepositOptions {
...
+    }
+
+    builder.powchain(
+        b -> {
+          b.eth1Endpoints(eth1Endpoints);
+          b.eth1LogsMaxBlockRange(eth1LogsMaxBlockRange);
+          b.useMissingDepositEventLogging(useMissingDepositEventLogging);
+          b.customDepositSnapshotPath(depositSnapshotPath);
+          b.depositSnapshotEnabled(depositSnapshotEnabled);
+          if (checkpointSyncUrlSet) {
```
zilm13: It could be very difficult to maintain and follow when foreign option is checked and set here.

zilm13 (commented):
I think it could be not the best way to mix two options parsing in `DepositOptions`. We'd better set all input from appropriate options to configuration. Validate and throw in `validate()` part of builder and set as many fields as we need for decision in final configuration. It will look better and easier to maintain and analyze. Also we could add DepositSnapshot inner class in PowChainConfiguration for better readability, so the options could be accessed like `tekuConfig.powchain().depositSnapshot().enabled()` or `tekuConfig.powchain().depositSnapshot().bundledPath()`

PR #7874, opened 2024-01-11
Moved initial state and genesis state config into StateBootstrapConfig

lucassaldanha:


lucassaldanha:
One test is failing because of the way I am now sanitizing the URL (and it is stripping out username:password from the URL. I need to investigate if this is something we want to support or not.

zilm13:
My understanding is that we used sanitizing only for logging or something that could be accessible by API, so potentially private data is not leaking, like here
https://github.com/ConsenSys/teku/blob/9e0635744f9fc1293305eb0014f7c4eddbfe0f38/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java#L67-L72

zilm13 (commented):
I like it but I'd remove 
```
  public Optional<String> getInitialState() {
    return stateBoostrapConfig.getInitialState();
  }

  public Optional<String> getGenesisState() {
    return stateBoostrapConfig.getGenesisState();
  }
  ```
  at all, so when needed you always access it via stateBootstrapConfig, without shortcuts. It should look more organized after this change, I think

**Comment by lucassaldanha** [comment]
> My understanding is that we used sanitizing only for logging or something that could be accessible by API

Yeah, I think I was being extra cautious but for no good reason.

File: ethereum/networks/src/main/java/tech/pegasys/teku/networks/StateBoostrapConfig.java [hunk 1]
```diff
@@ -13,32 +13,72 @@
...
+
+  public void setCustomInitialState(final String initialState) {
+    checkNotNull(initialState);
+    this.initialState = Optional.of(initialState);
+    this.isUsingCustomInitialState = true;
   }
 
-  public void setUsingCheckpointSync(final boolean usingCheckpointSync) {
-    isUsingCheckpointSync = usingCheckpointSync;
+  public void setCheckpointSyncUrl(final String checkpointSyncUrl) {
```
zilm13: I think some clarification is needed here why it's not the `isUsingCustomInitialState`

ethereum/networks/src/main/java/tech/pegasys/teku/networks/StateBoostrapConfig.java
```diff
@@ -13,32 +13,72 @@
...
+  public void setInitialState(final String initialState) {
+    checkNotNull(initialState);
+    this.initialState = Optional.of(initialState);
+    this.isUsingCustomInitialState = false;
+  }
+
+  public void setCustomInitialState(final String initialState) {
+    checkNotNull(initialState);
+    this.initialState = Optional.of(initialState);
+    this.isUsingCustomInitialState = true;
```
zilm13: it looks a bit fragile  that it could be damaged by the call order. Maybe set default to false and override it only here and don't override in `setInitialState`?

zilm13 (commented):
I'm a bit stuck now on the way I've suggested. `StateBootstrapConfig` looks like builder without result being built with it's non final values and lack of validation, which is dangerous and not common for us. As our goal is to set another value from `DepositOptions`, I think it should be builder, all current code should be moved inside this builder and we should have non builder properties for outside (or just record). So later in TekuConfguration we can run eth2NetworkConfiguration.stateBootstrapConfig.build() and it will give us StateBoostrapConfig, which shouldn't be a part of eth2NetworkConfiguration which was already built, but it could be a part of tekuConfiguration at least. What do you think?

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -290,9 +276,11 @@ public String toString() {
   public static class Builder {
     private static final String EPOCHS_STORE_BLOBS_MAX_KEYWORD = "MAX";
     private String constants;
-    private Optional<String> initialState = Optional.empty();
-    private final StateBoostrapConfig stateBoostrapConfig = new StateBoostrapConfig();
     private Optional<String> genesisState = Optional.empty();
+    private Optional<String> initialState = Optional.empty();
+    private Optional<String> checkpointSyncUrl = Optional.empty();
+    private boolean isUsingCustomInitialState = false;
+    private boolean allowSyncOutsideWeakSubjectivityPeriod = false;
```
zilm13: we probably need constant for this one

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -631,8 +632,11 @@ public Builder applyNetworkDefaults(final Eth2Network network) {
 
     private Builder reset() {
       constants = null;
-      initialState = Optional.empty();
       genesisState = Optional.empty();
+      initialState = Optional.empty();
+      checkpointSyncUrl = Optional.empty();
+      isUsingCustomInitialState = false;
+      allowSyncOutsideWeakSubjectivityPeriod = false;
```
zilm13: and use this constant here too

zilm13 (approved):
LGTM just one nit

PR #7875, opened 2024-01-12
Create the key at `—p2p-private-key-file` path if none

### LINKED ISSUE #7846: Make the `—p2p-private-key-file` a dynamic path to allow snapshots of the data volume sans the key
### Description
Im running Teku in k8s with the usual precautions of storing the chain data on a volume, and ideally I'd like to snapshot this volume so that I can restore it for future syncs. The issue however is that when I snapshot the volume at the moment I also get the `—p2p-private-key-file` which I'd prefer not to have and then then manage separately. Could the current behaviour be altered so that:
- if a file exists at the path pointed to by `—p2p-private-key-file` then use it
- if it doesn't, create the key at that path?

courtneyeh:
fixes #7846 

## Documentation

## Changelog


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfig.java
```diff
@@ -281,7 +284,23 @@ public FilePrivateKeySource(String fileName) {
...
+        if (!file.createNewFile()) {
+          return getPrivateKeyBytes();
+        }
+        final Bytes privateKeyBytes = LibP2PPrivateKeyLoader.generateNewPrivateKey(keyValueStore);
+        Files.write(file.toPath(), privateKeyBytes.toArray());
+        return privateKeyBytes;
+
+      } catch (IOException e) {
+        throw new RuntimeException(
+            "Not able to create or retrieve p2p private key file - " + fileName);
```
rolfyone: Not sure RuntimeException is the best option...

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPrivateKeyLoader.java
```diff
@@ -39,12 +39,12 @@ public LibP2PPrivateKeyLoader(
   public PrivKey get() {
     final Bytes privKeyBytes =
         privateKeySource
-            .map(PrivateKeySource::getPrivateKeyBytes)
-            .orElseGet(this::generateNewPrivateKey);
+            .map(source -> source.getOrGeneratePrivateKeyBytes(keyValueStore))
+            .orElseGet(() -> generateNewPrivateKey(keyValueStore));
```
lucassaldanha: This logic means we will never reach the `orElseGet(..)` code path, potentially losing existing functionality/behaviour from `LibP2PPrivateKeyLoader#generateNewPrivateKey(..)`.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPrivateKeyLoader.java
```diff
@@ -39,17 +39,17 @@ public LibP2PPrivateKeyLoader(
...
         privateKeySource
-            .map(PrivateKeySource::getPrivateKeyBytes)
-            .orElseGet(this::generateNewPrivateKey);
+            .map(PrivateKeySource::getOrGeneratePrivateKeyBytes)
+            .orElseGet(() -> generateAndSaveNewPrivateKey(keyValueStore));
     return KeyKt.unmarshalPrivateKey(privKeyBytes.toArrayUnsafe());
   }
 
-  private Bytes generateNewPrivateKey() {
+  public static Bytes generateAndSaveNewPrivateKey(
```
lucassaldanha: We can revert this to a private non-static method.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPrivateKeyLoader.java
```diff
@@ -58,4 +58,9 @@ private Bytes generateNewPrivateKey() {
     }
     return privateKey;
   }
+
+  public static Bytes generateNewPrivateKey() {
+    final PrivKey privKey = KeyKt.generateKeyPair(KeyType.SECP256K1).component1();
+    return Bytes.wrap(KeyKt.marshalPrivateKey(privKey));
+  }
```
lucassaldanha: I'd say either we move this into a utility class or we just use the same two lines here and in `getOrGeneratePrivateKeyBytes`. We don't want this dependency between `FilePrivateKeySource` and `LibP2PPrivKeyLoader`

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPrivateKeyLoaderTest.java
```diff
@@ -62,6 +67,33 @@ void testPrivateKeyLoaded() {
...
+    Files.writeString(file, privateKey.toHexString());
+
+    final PrivateKeySource privKeySource = new NetworkConfig.FilePrivateKeySource(file.toString());
+    final LibP2PPrivateKeyLoader loader =
+        new LibP2PPrivateKeyLoader(store, Optional.of(privKeySource));
+
+    final PrivKey expectedKey = KeyKt.unmarshalPrivateKey(privateKey.toArrayUnsafe());
+    assertThat(loader.get()).isEqualTo(expectedKey);
+  }
+
```
lucassaldanha: These tests are not in the right place... Our tests for `LibP2PPrivateKeyLoader` should not directly depend on `FilePrivateKeySource`.

In this class, we would be testing things like:
1. When `privateKeySource` is empty, we generate a new key
2. When user provides a key (privateKeySource != empty) we use the key provided by the `PrivateKeySource`.

All the tests related to actually creating a file etc should be done in the tests for `FilePrivateKeySource`

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfig.java
```diff
@@ -281,7 +283,23 @@ public FilePrivateKeySource(String fileName) {
     }
 
     @Override
-    public Bytes getPrivateKeyBytes() {
+    public Bytes getOrGeneratePrivateKeyBytes() {
```
lucassaldanha: NIT: I think we should extract `PrivateKeySource` and `FilePrivateKeySource` into their own source files.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfig.java
```diff
@@ -281,7 +283,23 @@ public FilePrivateKeySource(String fileName) {
...
+      try {
+        File file = new File(fileName);
+        if (!file.createNewFile()) {
+          return getPrivateKeyBytesFromFile();
+        }
+        final Bytes privateKeyBytes = LibP2PPrivateKeyLoader.generateNewPrivateKey();
```
lucassaldanha: This class should not depend on `LibP2PPrivateKeyLoader`

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPrivateKeyLoaderTest.java
lucassaldanha: For reference: see how on test `testPrivateKeyLoaded` we are using a simple implementation of `PrivateKeySource`.

lucassaldanha: Could even use a mock of `PrivateKeySource` to assert when it is used or not.

lucassaldanha:
We need to consider that there is no indication in the logs that we are creating a new key for the user when using `--p2p-private-key-file`. This can lead to users thinking they are using one existing key but creating a different one (e.g. if they have a typo on their config).

I think having a log message like we have with `StatusLogger.usingGeneratedP2pPrivateKey` would be nice.

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfigTest.java
```diff
@@ -57,6 +60,36 @@ void getAdvertisedIp_shouldResolveLocalhostIpWhenInterfaceIpIsAnyLocalIpv6() {
...
+            .advertisedIp(advertisedIp)
+            .networkInterface(listenIp)
+            .privateKeyFile(tempDir.toString()) // directory provided instead of file
+            .build();
+
+    assertThat(config.getPrivateKeySource()).isPresent();
+    assertThatThrownBy(() -> config.getPrivateKeySource().get().getOrGeneratePrivateKeyBytes())
+        .isInstanceOf(RuntimeException.class)
+        .hasMessageContaining("p2p private key file not found -");
+  }
```
lucassaldanha: These tests are all about testing `FilePrivateKeySource`'s failure behaviours and should be on `FilePrivateKeySourceTest`.

lucassaldanha: I believe what makes sense to test on `NetworkConfigTest`  is that we are correctly creating a `PrivateKeySource` when we use the builder with `privateKeyFile`

lucassaldanha:
We need a CHANGELOG entry for this new feature.

lucassaldanha (approved):

joshuafernandes:
Thankyou @lucassaldanha & @courtneyeh :)

PR #7892, opened 2024-01-17
Added optional/required deposit tree snapshot resource + fallback

### LINKED ISSUE #7715: Add DepositTreeSnapshot download to checkpoint sync

lucassaldanha:
## PR Description

The core idea in this PR is that `DepositSnapshotFileLoader` now has multiple resources to load (instead of a single URL). Each resource can be set as required (can't fail) or optional (can fail).

In our current implementation, when using `--Xdeposit-snapshot` to provide a custom deposit tree snapshot to Teku, we will set it as the only resource, and mark it as required. Therefore, if it fails to load, Teku will exit with an error.

For known networks like Mainnet or Testnets (networks that we have a bundled snapshot of the deposit tree), we will also set the bundled version as a required resource (there is no reason for it to fail).

Last but not least, if the user is using `--checkpoint-sync-url`, we will add an extra optional resource that takes precedence over the bundled resource. If it fails loading, we will fallback into the bundled one.

When configuring `DepositSnapshotFileLoader` through its builder, we keep the order of insertion, what reflects the order we use when evaluating resources.

## Fixed Issue(s)
fixes #7715 

## Documentation

## Changelog

### DIFF
--- a/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
+++ b/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
@@ -18,17 +18,20 @@ import static tech.pegasys.teku.infrastructure.logging.StatusLogger.STATUS_LOG;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Function;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes;
 import tech.pegasys.teku.ethereum.pow.api.DepositTreeSnapshot;
 import tech.pegasys.teku.ethereum.pow.api.schema.LoadDepositSnapshotResult;
-import tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException;
 import tech.pegasys.teku.infrastructure.http.UrlSanitizer;
 import tech.pegasys.teku.infrastructure.io.resource.ResourceLoader;
 import tech.pegasys.teku.infrastructure.json.JsonUtil;
@@ -48,38 +51,44 @@ public class DepositSnapshotFileLoader {
 
   private static final Logger LOG = LogManager.getLogger();
 
-  private final Optional<String> depositSnapshotResource;
+  private final List<String> depositSnapshotResources = new ArrayList<>();
 
-  public DepositSnapshotFileLoader(final Optional<String> depositSnapshotResource) {
-    this.depositSnapshotResource = depositSnapshotResource;
+  private final Function<String, Optional<DepositTreeSnapshot>> depositSnapshotLoaderFunction;
+
+  public DepositSnapshotFileLoader(final List<String> depositSnapshotResources) {
+    this.depositSnapshotResources.addAll(depositSnapshotResources);
+    this.depositSnapshotLoaderFunction = this::maybeLoadDepositSnapshot;
+  }
+
+  @VisibleForTesting
+  DepositSnapshotFileLoader(
+      final List<String> depositSnapshotResources,
+      final Function<String, Optional<DepositTreeSnapshot>> depositSnapshotLoaderFunction) {
+    this.depositSnapshotResources.addAll(depositSnapshotResources);
+    this.depositSnapshotLoaderFunction = depositSnapshotLoaderFunction;
   }
 
   public LoadDepositSnapshotResult loadDepositSnapshot() {
-    final Optional<DepositTreeSnapshot> depositTreeSnapshot =
-        loadDepositSnapshot(depositSnapshotResource);
+    Optional<DepositTreeSnapshot> depositTreeSnapshot = Optional.empty();
+    for (int i = 0; i < depositSnapshotResources.size(); i++) {
+      depositTreeSnapshot = depositSnapshotLoaderFunction.apply(depositSnapshotResources.get(i));
+    }
     return LoadDepositSnapshotResult.create(depositTreeSnapshot);
   }
 
-  private Optional<DepositTreeSnapshot> loadDepositSnapshot(
-      final Optional<String> depositSnapshotResource) {
-    return depositSnapshotResource.map(
-        snapshotPath -> {
-          final String sanitizedResource = UrlSanitizer.sanitizePotentialUrl(snapshotPath);
-          try {
-            STATUS_LOG.loadingDepositSnapshotResource(sanitizedResource);
-            final DepositTreeSnapshot depositSnapshot = loadFromUrl(snapshotPath);
-            STATUS_LOG.onDepositSnapshot(
-                depositSnapshot.getDepositCount(), depositSnapshot.getExecutionBlockHash());
-            return depositSnapshot;
-          } catch (IOException e) {
-            LOG.error("Failed to load deposit tree snapshot", e);
-            throw new InvalidConfigurationException(
-                "Failed to load deposit tree snapshot from "
-                    + sanitizedResource
-                    + ": "
-                    + e.getMessage());
-          }
-        });
+  private Optional<DepositTreeSnapshot> maybeLoadDepositSnapshot(
+      final String depositSnapshotResourceUrl) {
+    final String sanitizedUrl = UrlSanitizer.sanitizePotentialUrl(depositSnapshotResourceUrl);
+    try {
+      STATUS_LOG.loadingDepositSnapshotResource(sanitizedUrl);
+      final DepositTreeSnapshot depositSnapshot = loadFromUrl(depositSnapshotResourceUrl);
+      STATUS_LOG.onDepositSnapshot(
+          depositSnapshot.getDepositCount(), depositSnapshot.getExecutionBlockHash());
+      return Optional.of(depositSnapshot);
+    } catch (final IOException e) {
+      LOG.warn("Failed to load deposit tree snapshot from " + sanitizedUrl, e);
+      return Optional.empty();
+    }
   }
 
   private DepositTreeSnapshot loadFromUrl(final String path) throws IOException {
@@ -106,5 +115,10 @@ public class DepositSnapshotFileLoader {
     final JsonNode jsonNode = new ObjectMapper().readTree(json);
     return JsonUtil.parse(
         jsonNode.get("data").toString(), DepositTreeSnapshot.getJsonTypeDefinition());
+  }
+
+  @VisibleForTesting
+  public List<String> getDepositSnapshotResources() {
+    return depositSnapshotResources;
   }
 }
--- a/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
+++ b/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
@@ -22,16 +22,16 @@ import com.google.common.annotations.VisibleForTesting;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
-import java.util.function.Function;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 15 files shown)


beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
```diff
@@ -48,38 +53,48 @@ public class DepositSnapshotFileLoader {
...
-          }
-        });
+  private Optional<DepositTreeSnapshot> maybeLoadDepositSnapshot(
+      final String depositSnapshotResourceUrl) throws IOException {
+    final String sanitizedUrl = UrlSanitizer.sanitizePotentialUrl(depositSnapshotResourceUrl);
+    STATUS_LOG.loadingDepositSnapshotResource(sanitizedUrl);
+    final DepositTreeSnapshot depositSnapshot = loadFromUrl(depositSnapshotResourceUrl);
+    STATUS_LOG.onDepositSnapshot(
+        depositSnapshot.getDepositCount(), depositSnapshot.getExecutionBlockHash());
+    return Optional.of(depositSnapshot);
```
zilm13: There is no possibility to return Optional.empty() in this method. So you need to either put some exception catching here or change method name to loadDepositSnapshot, return `DepositTreeSnapshot` without Optional and remove handling of `isEmpty()` case in the calling method

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoaderTest.java
```diff
@@ -77,15 +81,69 @@ public void shouldHaveCorrectDepositsData_whenSnapshotLoaded() {
         });
   }
 
+  @Test
+  public void shouldTryAllAvailableResources() {
```
zilm13: I'd add something like `UntilFirstNonEmptyFound` to be clear

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/PowchainConfigurationTest.java
```diff
@@ -0,0 +1,72 @@
...
+    final PowchainConfiguration config =
+        PowchainConfiguration.builder()
+            .specProvider(spec)
+            .depositSnapshotEnabled(false)
+            // even though we are setting the value, because depositSnapshotEnabled = false it won't
+            // be set
+            .setDepositSnapshotPathForNetwork(Optional.of(Eth2Network.MAINNET))
+            .build();
+
+    assertThat(config.getDepositTreeSnapshotConfiguration().getBundledDepositSnapshotPath())
```
zilm13, line 69: I'd add check that bundled deposit snapshot is not enabled in result configuration

teku/src/test/java/tech/pegasys/teku/cli/options/DepositOptionsTest.java
```diff
@@ -129,23 +131,63 @@ public void shouldSetDefaultBundleSnapshotPathForSupportedNetwork(final String n
...
   @Test
   public void shouldIgnoreBundleSnapshotPathForNotSupportedNetwork() {
     final String[] args = {"--network=swift", "--deposit-snapshot-enabled"};
     final TekuConfiguration config = getTekuConfigurationFromArguments(args);
-    assertThat(config.powchain().isDepositSnapshotEnabled()).isTrue();
-    assertThat(config.powchain().getDepositSnapshotPath()).isEmpty();
+    final DepositTreeSnapshotConfiguration depositTreeSnapshotConfiguration =
+        config.powchain().getDepositTreeSnapshotConfiguration();
+    assertThat(depositTreeSnapshotConfiguration.isDepositSnapshotEnabled()).isTrue();
+    assertThat(depositTreeSnapshotConfiguration.getCustomDepositSnapshotPath()).isEmpty();
```
zilm13, line 157: bundled?

teku/src/test/java/tech/pegasys/teku/cli/options/DepositOptionsTest.java
```diff
@@ -129,23 +131,63 @@ public void shouldSetDefaultBundleSnapshotPathForSupportedNetwork(final String n
...
+    final TekuConfiguration config = getTekuConfigurationFromArguments(args);
+    final DepositTreeSnapshotConfiguration depositTreeSnapshotConfiguration =
+        config.powchain().getDepositTreeSnapshotConfiguration();
+    assertThat(depositTreeSnapshotConfiguration.isDepositSnapshotEnabled()).isTrue();
+    assertThat(depositTreeSnapshotConfiguration.getCheckpointSyncDepositSnapshotUrl())
+        .hasValue("http://checkpoint-sync.com" + DEPOSIT_SNAPSHOT_URL_PATH);
+  }
+
+  @Test
+  public void shouldDisableBundledDepositSnapshotWhenUsingCustomDepositSnapshotPath() {
```
zilm13: duplicated with `shouldSetUseDepositSnapshotToFalseWhenCustomDepositSnapshotIsProvided`

services/powchain/src/main/java/tech/pegasys/teku/services/powchain/DepositTreeSnapshotConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+  private final Optional<String> customDepositSnapshotPath;
+  private final Optional<String> checkpointSyncDepositSnapshotUrl;
+  private final Optional<String> bundledDepositSnapshotPath;
+  private final boolean depositSnapshotEnabled;
+
+  DepositTreeSnapshotConfiguration(
+      final Optional<String> customDepositSnapshotPath,
+      final Optional<String> checkpointSyncDepositSnapshotUrl,
+      final Optional<String> bundledDepositSnapshotPath,
+      final boolean depositSnapshotEnabled) {
```
zilm13: you initialize it with `isBundledSnapshotEnabled`, so it's better to continue with this name and rename getter too

zilm13 (commented):
I don't understand what happens when both checkpoint sync url and custom deposit snapshots are provided. Not sure if I missed it? Adding this case to tests would be good

teku/src/test/java/tech/pegasys/teku/cli/options/DepositOptionsTest.java
zilm13: let's check both are empty

services/powchain/src/main/java/tech/pegasys/teku/services/powchain/DepositTreeSnapshotConfiguration.java
zilm13: yeah, this thing is very confusing, if we rename it and getters it would be more clear I think

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoaderTest.java
```diff
@@ -77,15 +81,69 @@ public void shouldHaveCorrectDepositsData_whenSnapshotLoaded() {
...
+  public void shouldTryAllAvailableResourcesUntilFirstNonEmptyFound() {
+    final String validResourcePath = getResourceFilePath(SNAPSHOT_BEACON_API_RESOURCE);
+    depositSnapshotLoader =
+        new DepositSnapshotFileLoader.Builder()
+            .addOptionalResource("/foo/empty")
+            .addRequiredResource(validResourcePath)
+            .build();
+
+    final LoadDepositSnapshotResult result = depositSnapshotLoader.loadDepositSnapshot();
+    assertThat(result.getDepositTreeSnapshot()).isPresent();
```
zilm13, line 98: it would be good to compare at least few fields that it matches those from the fixture

services/powchain/src/main/java/tech/pegasys/teku/services/powchain/PowchainConfiguration.java
```diff
@@ -113,15 +110,22 @@ private Builder() {}
 
     public PowchainConfiguration build() {
       validate();
+
+      boolean isBundledSnapshotEnabled =
```
zilm13: final please

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotsBundleTest.java
```diff
@@ -57,12 +57,22 @@ public void shouldCreateCorrectDepositTreeSnapshotFromEachBundleSnapshot(
         .depositSnapshotEnabled(true)
         .setDepositSnapshotPathForNetwork(Optional.of(eth2Network));
     final PowchainConfiguration powchainConfiguration = powchainConfigBuilder.build();
-    assumeThat(powchainConfiguration.getDepositSnapshotPath())
+    assumeThat(
+            powchainConfiguration
+                .getDepositTreeSnapshotConfiguration()
+                .getCustomDepositSnapshotPath())
```
zilm13: getBundledDepositSnapshotPath

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotsBundleTest.java
```diff
@@ -57,12 +57,22 @@ public void shouldCreateCorrectDepositTreeSnapshotFromEachBundleSnapshot(
...
         .describedAs("No built-in snapshot for network %s", eth2Network)
         .isPresent();
 
     final DepositSnapshotFileLoader depositSnapshotLoader =
-        new DepositSnapshotFileLoader(powchainConfiguration.getDepositSnapshotPath());
+        new DepositSnapshotFileLoader.Builder()
+            .addRequiredResource(
+                powchainConfiguration
+                    .getDepositTreeSnapshotConfiguration()
+                    .getCustomDepositSnapshotPath()
```
zilm13: getBundledDepositSnapshotPath

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotsBundleTest.java
```diff
@@ -85,9 +95,13 @@ public void shouldValidateSnapshotViaRemoteAPI(
         .setDepositSnapshotPathForNetwork(Optional.of(eth2Network));
     final PowchainConfiguration powchainConfiguration = powchainConfigBuilder.build();
 
-    final String depositSnapshotPath = powchainConfiguration.getDepositSnapshotPath().orElseThrow();
+    final String depositSnapshotPath =
+        powchainConfiguration
+            .getDepositTreeSnapshotConfiguration()
+            .getCustomDepositSnapshotPath()
```
zilm13: getBundledDepositSnapshotPath

zilm13 (commented):
Some final fixes left, but I still don't understand the case when both `--Xdeposit-snapshot` and `checkpoint-sync-url` are provided. If we agree that we don't throw on this, we should at least verify with test that proper configuration is built (and bundled is false).

zilm13 (approved):
LGTM, just I wanted more like integration test on checkpoint sync + custom deposit snapshot, something like this in `DepositOptionsTest` (clarifying that isBundled is `false`):
```java

  @Test
  public void
      shouldDisableBundledDepositSnapshotWhenUsingCustomDepositSnapshotWithCheckpointSyncUrl() {
    final String[] args = {
      "--Xdeposit-snapshot", "/foo/bar", "--checkpoint-sync-url", "http://checkpoint-sync.com"
    };
    final TekuConfiguration config = getTekuConfigurationFromArguments(args);
    final DepositTreeSnapshotConfiguration depositTreeSnapshotConfiguration =
        config.powchain().getDepositTreeSnapshotConfiguration();
    assertThat(depositTreeSnapshotConfiguration.isBundledDepositSnapshotEnabled()).isFalse();
    assertThat(depositTreeSnapshotConfiguration.getCustomDepositSnapshotPath())
        .hasValue("/foo/bar");
  }
```
I'd appreciate if you add it too

PR #7900, opened 2024-01-19
Added docker jdk21 build configuration

### LINKED ISSUE #7899: Add JDK 21 docker image
JDK21 potentially brings performance improvements without changing code, and potentially better garbage collection logic. we should create a JDK-21 image so it can be tested out.

rolfyone:
I'm not sure if this publishes, will have to check when we get to develop maybe...

I started up a minimal node and was able to produce blocks.

fixes #7899

## Documentation

## Changelog

### DIFF
--- a/docker/jdk17/Dockerfile
+++ b/docker/jdk17/Dockerfile
@@ -46,8 +46,8 @@ ARG VCS_REF
 ARG VERSION
 LABEL org.label-schema.build-date=$BUILD_DATE \
       org.label-schema.name="Teku" \
-      org.label-schema.description="Ethereum 2.0 Beacon Chain Client" \
-      org.label-schema.url="https://consensys.net/knowledge-base/ethereum-2/teku/" \
+      org.label-schema.description="Ethereum Consensus Client" \
+      org.label-schema.url="https://docs.teku.consensys.io/" \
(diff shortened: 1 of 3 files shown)


tbenr (approved):
LGTM. @benjaminion effects are already playing out

docker/jdk17/Dockerfile
```diff
@@ -46,8 +46,8 @@ ARG VCS_REF
 ARG VERSION
 LABEL org.label-schema.build-date=$BUILD_DATE \
       org.label-schema.name="Teku" \
-      org.label-schema.description="Ethereum 2.0 Beacon Chain Client" \
-      org.label-schema.url="https://consensys.net/knowledge-base/ethereum-2/teku/" \
+      org.label-schema.description="Ethereum Consensus Client" \
+      org.label-schema.url="https://docs.teku.consensys.io/" \
```
StefanBratanov, line 50: I think this is better link marketing wise: https://consensys.io/teku

StefanBratanov (commented):
just a nit

docker/jdk17/Dockerfile
rolfyone: it had all the wrong terminology but the old link was just plain wrong... at least the doc link is maintained... its a good start and we can revisit.

PR #7873, opened 2024-01-11
Confirmation rule prerequisite fc change

tbenr:
Fixes #7802

To be checked and merged together with the PR enabling deneb fork on mainnet

PR #7903, opened 2024-01-22
Update reftests and fork choice confirmation rule update

### LINKED ISSUE #7802: fork choice filter change for deneb
https://github.com/ethereum/consensus-specs/pull/3431

rolfyone:
FC confirmation rule change did affect 2 unit tests, but I believe they're correct to change as it was the FC rule that has changed outputs.

This proposes a completed fix to #7873

Fixes #7802

## Documentation

## Changelog


tbenr:
My problem with our forkchoice unit tests is that I think they are outdated. For instance, with progressive balance PRs (ie 6141), Protoarray became "current epoch aware" but our tests never set current epoch (it remains 0 for the entire  NoVotesTest). I'll check how your change did the magic but to me there is something fundamentally broken.

(on my previous PR I was trying to fix them by moving time forward according to the test)


🤔 but looking better `actualEpoch.plus(2).isGreaterThanOrEqualTo(currentEpoch)` is always true if `currentEpoch` remains 0... ok, taking a closer look at what's happening `NoVotesTest` looks good

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -695,9 +680,11 @@ private boolean hasAncestorAtSlot(
     return node != null && requiredRoot.equals(node.getBlockRoot());
   }
 
-  private boolean doesCheckpointEpochMatch(final Checkpoint actual, final Checkpoint required) {
-    return required.getEpoch().equals(initialEpoch)
-        || actual.getEpoch().equals(required.getEpoch());
+  private boolean isVotingSourceWithinAcceptableRange(
```
tbenr, line 684: in my version I was passing epochs to this function since this actually deals only with epochs (so we get them once and lines here are shorter)

tbenr:
I pushed a change in the forkchoicestrategy test. I think the way you fixed it defeat the purpose of it, which is testing `forkChoiceData` to have optimistic head due to the right side of:
image

tbenr (approved):
LGTM.
The only nit is to pass epochs instead of checkpoints, if we want to.

We have to wait to be close to mainnet DENEB release before merging.

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
rolfyone: originally i was rationalising that change and needed checkpoint but i think its probably ok now, will look at a refactor here.

PR #7906, opened 2024-01-22
Remove EL_OFFLINE from SyncState enum

StefanBratanov:
## PR Description

- Remove `EL_OFFLINE` from `SyncState` and use the existing `ExecutionClientDataProvider` to get the information.

This will fix cases where the `SyncState` can be misinterpreted. Chain could still be in sync even if EL has timed out or errored for a single request. It's better to let `onTick()` control the `SyncState` entirely.

 For example, the following issue happened to one of our users:

- Block building was initiated using the builder flow and local EL `getPayload` was called. 
- The request timed out and `SyncState` was changed to `EL_OFFLINE` so `SyncState.isInSync()` was returning false
- Remote `VC` gets 503 (still syncing) when publishing the block and it fails.

Also `onTick()` could be triggered at the same time which checks `SyncState.isInSync()` and since it can be returning false, it can start syncing process wrongly.

We only introduced EL_OFFLINE for `/eth/v1/node/syncing` endpoint and think it doesn't belong in the syncing process.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #7909, opened 2024-01-23
Move duplicate blobSidecar log from warn to debug (developers info)

zilm13:
## PR Description
Used to detect when we try to add blobSidecar, which is already downloaded and added. Developers only info to clarify that we are not doing this too often. Therefore moved to `debug`.  

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7789, opened 2023-12-06
Initial cut of late block reorg

rolfyone:
- requires development flag to enable
 - will introduce debug logging in ForkChoice and RecentChainData so that we can observe behaviours.
 - Needed tweak some areas while I was non-validator node testing.

 Will push this to a validating node and enable debugging so that I can make broader observations as to how things are working. If anything needs tweaking we may miss some block proposals so will likely choose a smaller node rather than a bigger node.

partially addresses #6595 

## Documentation

- [X] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [X] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #7911, opened 2024-01-24
Late Block reorg refactor

rolfyone:
Removed state from isHeadWeak and isParentStrong - these values now get cached on updateHeadTransaction

Renamed newSlot to proposalSlot

Added a function to cache weights, as they're expensive to compute and we shouldn't need to compute each call.

debug logging added in recentChainData at decission points.

partially addresses #6595 - this is a subset of #7789

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/forkchoice/ForkChoiceTestExecutor.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/forkchoice/ForkChoiceTestExecutor.java
@@ -17,7 +17,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static tech.pegasys.teku.infrastructure.async.SafeFutureAssert.safeJoin;
 import static tech.pegasys.teku.infrastructure.time.TimeUtilities.secondsToMillis;
-import static tech.pegasys.teku.networks.Eth2NetworkConfiguration.DEFAULT_FORK_CHOICE_LATE_BLOCK_REORG_ENABLED;
 
 import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
@@ -137,7 +136,7 @@ public class ForkChoiceTestExecutor implements TestExecutor {
             new ForkChoiceStateProvider(eventThread, recentChainData),
             new TickProcessor(spec, recentChainData),
             transitionBlockValidator,
-            DEFAULT_FORK_CHOICE_LATE_BLOCK_REORG_ENABLED,
+            true,
             storageSystem.getMetricsSystem());
     final ExecutionLayerChannelStub executionLayer =
         new ExecutionLayerChannelStub(spec, false, Optional.empty());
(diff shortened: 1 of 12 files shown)


storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
```diff
@@ -669,6 +669,14 @@ public Bytes32 getProposerHead(final Bytes32 headRoot, final UInt64 slot) {
...
         || isProposerBoostActive
         || maybeHead.isEmpty()) {
+      LOG.debug(
+          "getProposerHead - return headRoot - isHeadLate {}, isShufflingStable {}, isFinalizationOk {}, isProposingOnTime {}, isProposerBoostActive {}, head.isEmpty {}",
+          () -> isHeadLate,
+          () -> isShufflingStable,
+          () -> isFinalizationOk,
+          () -> isProposingOnTime,
+          () -> isProposerBoostActive,
+          () -> headRoot.isEmpty());
```
tbenr, line 679: we don't need lambdas here since values are immediate.

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
```diff
@@ -685,31 +693,14 @@ public Bytes32 getProposerHead(final Bytes32 headRoot, final UInt64 slot) {
     // from the initial list, check
     // isFfgCompetitive, isSingleSlotReorg
     if (!isFfgCompetitive || !isSingleSlotReorg) {
+      LOG.debug(
+          "getProposerHead - return headRoot - isFfgCompetitive {}, isSingleSlotReorg {}",
+          () -> isFfgCompetitive,
+          () -> isSingleSlotReorg);
```
tbenr: same here

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/forkchoice/ForkChoiceTestExecutor.java
```diff
@@ -137,7 +136,7 @@ spec, new SignedBlockAndState(anchorBlock, anchorState)),
             new ForkChoiceStateProvider(eventThread, recentChainData),
             new TickProcessor(spec, recentChainData),
             transitionBlockValidator,
-            DEFAULT_FORK_CHOICE_LATE_BLOCK_REORG_ENABLED,
+            true,
```
tbenr, line 141: why now hardcoded `true`?

tbenr (commented):
lambda suggestion and a question

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/forkchoice/ForkChoiceTestExecutor.java
rolfyone: reference tests have late block reorg in them, and i evaluate that as to whether to do late block reorg ForkChoice, so if this isnt set true here, we fail late block reorg group basically.

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
rolfyone: when i dont use lamdas you tell me to use lamdas...

lucassaldanha: I believe the issue is that the Logger API does not support a mix of values AND lamdas together. So either we use all values or all lambdas.

AFAIK there is not downside in using lambdas so if it is for the sake of consistency I am ok with it.

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/forkchoice/ForkChoiceTestExecutor.java
lucassaldanha: nit: maybe a comment explaining why we have the hardcoded value?

lucassaldanha (approved):

Issue #7762, opened 2023-11-28
Update VC to use JsonTypeDef instead of annotated schemas


lucassaldanha:
Related to "Finish off removing the old beacon-api schemas" and "Consistent VC API Calls" TIP.

courtneyeh:
Requires converting all the API requests in `OkHttpValidatorRestApiClient` to use `JsonTypeDef` as seen in `OkHttpValidatorTypeDefClient`.

courtneyeh:
GET requests involve:
- Moving static type definitions into `DeserializableTypeDefinition`'s, and moving them to the `ethereum.json.types` package along with the wrapper objects.
- Creating the request extending `AbstractTypeDefRequest`, and using this instead for calls made in the `RemoteValidatorApiHandler` class.

GET requests left: GetSpec, GetBlockHeader, GetValidators, ~~GetAttestationDuties~~, ~~GetSyncCommitteeDuties~~

GetSpec is expected to be harder because the data held in `GetSpecResponse` is less simple to turn into a `DeserializableTypeDefinition` from the looks of it.

courtneyeh:
**Update:** The `getAttestationDuties` method in `OkHttpValidatorRestApiClient` Is related to the `PostAttesterDuties` endpoint. Therefore, it is a POST not a GET.

Same with the `getSyncCommitteeDuties` method. It relates to `PostSyncDuties`.

rolfyone:
partially addressed in #8452

PR #7912, opened 2024-01-24
Move statics in `GetProposerDuties` to `SharedApiTypes`

courtneyeh:
#7762 

## Documentation

## Changelog


ethereum/json-types/src/main/java/tech/pegasys/teku/ethereum/json/types/SharedApiTypes.java
```diff
@@ -55,6 +62,52 @@ public class SharedApiTypes {
...
+              ProposerDuty::getPublicKey,
+              ProposerDuty.ProposerDutyBuilder::publicKey)
+          .withField(
+              "validator_index",
+              INTEGER_TYPE,
+              ProposerDuty::getValidatorIndex,
+              ProposerDuty.ProposerDutyBuilder::validatorIndex)
+          .withField(
+              "slot", UINT64_TYPE, ProposerDuty::getSlot, ProposerDuty.ProposerDutyBuilder::slot)
+          .build();
```
rolfyone: I wonder if this will be less messy if this lives in the type, i think the SharedApiTypes class will scale very badly if we throw things into it for the entire move

ethereum/json-types/src/main/java/tech/pegasys/teku/ethereum/json/types/wrappers/ProposerDuty.java
```diff
@@ -11,14 +11,38 @@
  * specific language governing permissions and limitations under the License.
  */
 
-package tech.pegasys.teku.validator.api;
+package tech.pegasys.teku.ethereum.json.types.wrappers;
```
rolfyone: given this comes from validator endpoints, id potentially make this scale neatly...
so if we had `json/types/validator`
maybe under that also builder.... because if we split out the builder and the type lives with the builder, then thats fairly neat, and the pojo can just stay clean...

rolfyone (approved):
LGTM

PR #7918, opened 2024-01-24
VC slashing part 6 - vc slashing single process

mehdi-aouadi:
## PR Description
Validator Slashing Part 6:
Handle VC/BN single process use case

## Fixed Issue(s)

#7707 

## Documentation

## Changelog

### DIFF
--- a/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
+++ b/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
@@ -487,7 +487,6 @@ public class BeaconChainController extends Service implements BeaconChainControl
     initAttestationPool();
     initAttesterSlashingPool();
     initProposerSlashingPool();
-    initSlashingProtection();
     initVoluntaryExitPool();
     initSignedBlsToExecutionChangePool();
     initEth1DataCache();
@@ -503,6 +502,7 @@ public class BeaconChainController extends Service implements BeaconChainControl
     initAttestationTopicSubscriber();
     initActiveValidatorTracker();
     initSubnetSubscriber();
+    initSlashingEventsSubscriptions();
     initPerformanceTracker();
     initDataProvider();
     initValidatorApiHandler();
@@ -659,7 +659,7 @@ public class BeaconChainController extends Service implements BeaconChainControl
     blockImporter.subscribeToVerifiedBlockProposerSlashings(proposerSlashingPool::removeAll);
   }
 
-  protected void initSlashingProtection() {
+  protected void initSlashingEventsSubscriptions() {
     if (beaconConfig.validatorConfig().isShutdownWhenValidatorSlashedEnabled()) {
       final ValidatorTimingChannel validatorTimingChannel =
           eventChannels.getPublisher(ValidatorTimingChannel.class);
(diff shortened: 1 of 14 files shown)


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -642,6 +643,13 @@ protected void initAttesterSlashingPool() {
                 .reversed());
     blockImporter.subscribeToVerifiedBlockAttesterSlashings(attesterSlashingPool::removeAll);
     attesterSlashingPool.subscribeOperationAdded(forkChoice::onAttesterSlashing);
+    if (beaconConfig.validatorConfig().isShutdownWhenValidatorSlashedEnabled()) {
+      final ValidatorTimingChannel validatorTimingChannel =
+          eventChannels.getPublisher(ValidatorTimingChannel.class);
+      attesterSlashingPool.subscribeOperationAdded(
+          (operation, validationStatus, fromNetwork) ->
+              validatorTimingChannel.onAttesterSlashing(operation));
+    }
```
rolfyone: iirc this is where we'd need a local AT to show these (at least proposer) work... not sure how good our solution currently is for unit testing `BeaconChainController`
Arguably a bit of a gap in our infra but outside the scope of your update. I guess the problem ultimately is it would be very easy for this to disappear and us break the feature without knowing, and BN+VC is the default mode of operation.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -642,6 +643,13 @@ protected void initAttesterSlashingPool() {
                 .reversed());
     blockImporter.subscribeToVerifiedBlockAttesterSlashings(attesterSlashingPool::removeAll);
     attesterSlashingPool.subscribeOperationAdded(forkChoice::onAttesterSlashing);
+    if (beaconConfig.validatorConfig().isShutdownWhenValidatorSlashedEnabled()) {
```
zilm13: Maybe make some method like `initSlashingProtection()`, put both subscriptions in it, and add it to run list somewhere in the end after `initActiveValidatorTracker` for example, as pools are global vars anyway. 
There is already some mess here and there but I think this way it will look better.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
zilm13: It would be good to make `BeaconChainController` testable one day

rolfyone: I agree, historically we break out things that need testing into their own class but there's instance like this where it'd be neat.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
mehdi-aouadi: Makes sense. Done

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
mehdi-aouadi: I think it's better to leave this refactoring to a separate PR. The AT tests it and we could live with it for now...

rolfyone (approved):
LGTM

PR #7921, opened 2024-01-24
VC slashing Part 8 - vc slashing at

mehdi-aouadi:
## PR Description
Add validator slashing acceptance tests

## Fixed Issue(s)

#7707 

## Documentation

## Changelog


acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -285,6 +300,91 @@ private Object2BooleanMap<UInt64> getValidatorLivenessAtEpoch(
...
 
+  public String postProposerSlashing(
+      final UInt64 slot, final UInt64 index, final BLSSecretKey secretKey) throws IOException {
+
+    final Fork fork = spec.getForkSchedule().getFork(spec.computeEpochAtSlot(slot));
+    final Bytes32 genesisValidatorRoot = fetchGenesis().getGenesisValidatorsRoot();
+    final ForkInfo forkInfo = new ForkInfo(fork, genesisValidatorRoot);
+    final SigningRootUtil signingRootUtil = new SigningRootUtil(spec);
+
+    final BeaconBlockHeader beaconBlockHeader1 =
```
zilm13: It would be more readable if you extract it to something like `  private SignedBeaconBlockHeader randomSignedBeaconBlockHeader(final UInt64 slot,final UInt64 validatorIndex,
final SigningRootUtil signingRootUtil, final ForkInfo forkInfo, final BLSSecretKey secretKey) ` and use 2 times

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -285,6 +300,91 @@ private Object2BooleanMap<UInt64> getValidatorLivenessAtEpoch(
...
+      final UInt64 slashedIndex,
+      final BLSSecretKey slashedValidatorSecretKey)
+      throws IOException {
+
+    final Fork fork = spec.getForkSchedule().getFork(spec.computeEpochAtSlot(slashingSlot));
+    final Bytes32 genesisValidatorRoot = fetchGenesis().getGenesisValidatorsRoot();
+    final ForkInfo forkInfo = new ForkInfo(fork, genesisValidatorRoot);
+    final SigningRootUtil signingRootUtil = new SigningRootUtil(spec);
+
+    final AttestationData attestationData1 =
```
zilm13: Same here, please extract random indexedAttestation generation to private method

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,474 @@
...
+
+    postSlashing(
+        UInt64.valueOf(3),
+        UInt64.valueOf(slashedValidatorIndex),
+        slashedValidatorKeyPair.getSecretKey(),
+        tekuNode,
+        slashingEventType);
+
+    tekuNode.waitForLogMessageContaining(
+        String.format(slashingActionLog, slashedValidatorKeyPair.getPublicKey().toHexString()));
```
zilm13: I think there should be added ` tekuNode.waitForExit();` to this test and all others. I've added it and it doesn't work. Any ideas?

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,474 @@
...
+    final List<BLSKeyPair> blsKeyPairs =
+        new MockStartValidatorKeyPairFactory().generateKeyPairs(0, slashedValidatorIndex + 1);
+    return blsKeyPairs.get(slashedValidatorIndex);
+  }
+
+  private void postSlashing(
+      final UInt64 slashingSlot,
+      final UInt64 slashedIndex,
+      final BLSSecretKey slashedValidatorSecretKey,
+      final TekuNode tekuNode,
```
zilm13: nit: I'd put `tekuNode`  to the top of the list as it's most crucial value for the test

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,474 @@
...
+
+    secondValidatorClient.start();
+
+    firstTekuNode.waitForEpochAtOrAbove(2);
+
+    final int slashedValidatorIndex = 34;
+    final BLSKeyPair slashedValidatorKeyPair = getBlsKeyPair(slashedValidatorIndex);
+
+    postSlashing(
+        UInt64.valueOf(60),
```
zilm13: it would be good to see something like `int slotInThirdEpoch = firstTekuNode.getSpec().forMilestone(SpecMilestone.ALTAIR).getSlotsPerEpoch() * 2 + 3;` above for clarity of this magic number

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,474 @@
...
+        String.format(slashingActionLog, slashedValidatorKeyPair.getPublicKey().toHexString()));
+
+    secondBeaconNode.stop();
+    secondValidatorClient.stop();
+  }
+
+  @ParameterizedTest
+  @MethodSource("getSlashingEventTypes")
+  void
+      shouldShutDownWhenOwnedValidatorSlashed_StandAloneVC_MultiplePeers_SlashingEventsThroughGossipOnly_NoBlocks(
```
zilm13: I don't understand why it's `NoBlocks` case, both nodes will have all blocks in this case. Protection activation could be the result of either gossip or block in this case.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,474 @@
...
+    secondTekuNode.waitForLogMessageContaining(
+        String.format(slashingActionLog, slashedValidatorKeyPair.getPublicKey().toHexString()));
+
+    firstTekuNode.stop();
+  }
+
+  @ParameterizedTest
+  @MethodSource("getSlashingEventTypes")
+  void
+      shouldShutDownWhenOwnedValidatorSlashed_SingleProcess_MultiplePeers_SlashingEventsThroughGossipOnly_NoBlocks(
```
zilm13: I don't understand why it's `NoBlocks` case, both nodes will have all blocks in this case. Protection activation could be the result of either gossip or block in this case.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,474 @@
...
+import tech.pegasys.teku.bls.BLSKeyPair;
+import tech.pegasys.teku.bls.BLSSecretKey;
+import tech.pegasys.teku.infrastructure.time.SystemTimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.datastructures.interop.MockStartValidatorKeyPairFactory;
+import tech.pegasys.teku.test.acceptance.dsl.AcceptanceTestBase;
+import tech.pegasys.teku.test.acceptance.dsl.TekuNode;
+import tech.pegasys.teku.test.acceptance.dsl.TekuValidatorNode;
+
+public class ValidatorSlashingDetectionAcceptanceTest extends AcceptanceTestBase {
```
zilm13: It would be good to have a description here why we need to test so many cases (single process bn+vc/ separate vc, block slashing/gossip slashing)

zilm13 (commented):
I'm not sure we have a way to cover NoBlock case separately. But maybe I'm wrong. If not I'd just remove these cases.
Also it would be great to see in the end of each test something like
```
secondTekuNode.waitForExit();
firstTekuNode.waitForSlot();
```
to see that correcct node is shutdown and everything else not

lines like `firstTekuNode.stop();` are not needed, it should be shutdown automatically

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
mehdi-aouadi: My bad, forgot to clean up this mess 🙈

mehdi-aouadi: Done

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
mehdi-aouadi: Done

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: Done

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: Done for all

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: Done

mehdi-aouadi: Let me know if that's what you had in mind

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: But there's a param which is the validator key and it's not a simple exit but rather a specific slashing shut down log... Not sure I get what you mean here

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: This is "no blocks" because the first node is not running any validator and hence won't produce any block. The slashing post is sent to that same node which is then gossiped...

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: Same here. In the "no blocks" scenarios we're simply not running validators and assuming that node won't produce any block...

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
zilm13: But is there any difference when node is block producer or just received block via gossip for this protection?

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
zilm13: But the process exits?
I mean I'd be more happy to know that it's not only logline but Teku actually shutdown

mehdi-aouadi: Oh I see what you mean. Turned out to be a timeout issue (the existing `waitForExit` waits 30s and that wasn't enough), fixed it by adding a new `waitForShutDown`. I also added some checks to make sure the non concerned nodes are not shut down in there own...

mehdi-aouadi:
> I'm not sure we have a way to cover NoBlock case separately. But maybe I'm wrong. If not I'd just remove these cases. Also it would be great to see in the end of each test something like
> 
> ```
> secondTekuNode.waitForExit();
> firstTekuNode.waitForSlot();
> ```
> 
> to see that correcct node is shutdown and everything else not
> 
> lines like `firstTekuNode.stop();` are not needed, it should be shutdown automatically

Should be better now. Removed some non required manual shut downs and added some checks to make sure only the slashed nodes are shut down

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: If we'd like to simulate the slashing being sent within a block and not the slashing gossip topics, the node which receives the slashing should produce a block in order to include the slashing in it.
It goes like the following, assuming the slashed validator is run by Node 2
- Block scenario: Post slashing to Node 1 -> Node 1 produces a block including the slashing -> Node 1 gossip the block -> Node 2 receives the blocks and sees the slashing -> Node 2 shut down
- Proposer slashing or attester slashing gossip messages: Post slashing to Node 1 -> Node 1 sends the slashing message to the proposer or attester slashing gossip topic -> Node 2 receives the proposer or attester slashing message through gossip -> Node 2 shut down

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
```diff
@@ -285,6 +304,88 @@ private Object2BooleanMap<UInt64> getValidatorLivenessAtEpoch(
...
+    return httpClient.post(getRestApiUrl(), POST_ATTESTER_SLASHING_URL, body);
+  }
+
+  private static IndexedAttestation randomIndexedAttestation(
+      final UInt64 slot,
+      final UInt64 index,
+      final BLSSecretKey secretKey,
+      final SigningRootUtil signingRootUtil,
+      final ForkInfo forkInfo) {
+    final AttestationData attestationData1 =
```
zilm13: nit: you could remove 1's here and there

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,529 @@
...
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.datastructures.interop.MockStartValidatorKeyPairFactory;
+import tech.pegasys.teku.test.acceptance.dsl.AcceptanceTestBase;
+import tech.pegasys.teku.test.acceptance.dsl.TekuNode;
+import tech.pegasys.teku.test.acceptance.dsl.TekuValidatorNode;
+
+/**
+ * In order to cover all possible validator slashing scenarios, this acceptance test runs different
+ * combinations: <br>
+ * - Single Process: VC/BN running in a single process. In this case there is no SEE. <br>
```
zilm13: SSE

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/Node.java
```diff
@@ -115,6 +115,13 @@ public void waitForExit() {
         () -> assertThat(container.isRunning()).describedAs("Container is running").isFalse());
   }
 
+  public void waitForShutDown() {
```
zilm13: it's very unclear difference in naming, I'd rather call it the same `waitForExit` but with `seconds` parameter leaving a comment on standard one that it will wait for 30 seconds, which could be not enough for complete node shutdown

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,529 @@
...
+ * - Stand-Alone VC: VC/BN running in a separate processes and communicating through the REST APIs
+ * and SEE. In this case the slashing event is sent from the BN to the VC through SSE. <br>
```
zilm13: and SSE

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
```diff
@@ -0,0 +1,529 @@
...
+    secondBeaconNode.stop();
+    firstTekuNode.stop();
+  }
+
+  private TekuNode.Config configureNode(
+      final TekuNode.Config node, final int genesisTime, final String network) {
+    return node.withNetwork(network).withGenesisTime(genesisTime).withRealNetwork();
+  }
+
+  private static BLSKeyPair getBlsKeyPair(int slashedValidatorIndex) {
```
zilm13: final, please

zilm13 (commented):
LGTM, but let's wait for approval until APAC feedback on #7954
Also let's split it for clarity to `ValidatorSlashingDetectionAcceptanceTest` and `ValidatorSlashingDetectionNightlyAcceptanceTest`, so class description will need split too with nightly link. All tests in nightly should be disabled. What tests do you think we should leave in this one? I'd leave just first two

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNode.java
mehdi-aouadi: My bad. Done

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/Node.java
mehdi-aouadi: Done

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: Done

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: Done

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/ValidatorSlashingDetectionAcceptanceTest.java
mehdi-aouadi: Done

mehdi-aouadi:
> LGTM, but let's wait for approval until APAC feedback on #7954 Also let's split it for clarity to `ValidatorSlashingDetectionAcceptanceTest` and `ValidatorSlashingDetectionNightlyAcceptanceTest`, so class description will need split too with nightly link. All tests in nightly should be disabled. What tests do you think we should leave in this one? I'd leave just first two

IMO we could keep the two single peer tests (standalone VC and single process VC/BN: `shouldShutDownWhenOwnedValidatorSlashed_SingleProcess_SinglePeer` and `shouldShutDownWhenOwnedValidatorSlashed_StandAloneVC_SinglePeer`) which are pretty fast and broadly cover the feature. The other ones that cover more complex combinations/scenarios could be left to the nightly run if approved...

zilm13 (approved):
LGTM 
just a package name way too long even for Java
how about `slashingprotection`?
and we could also move Doppelganger test there too

mehdi-aouadi:
> LGTM just a package name way too long even for Java how about `slashingprotection`? and we could also move Doppelganger test there too

Renamed the package to `validatorslashing` and simplified the tests class names too.
I left the doppelganger where it is since it's a bit different topic...

PR #7926, opened 2024-01-25
Change comment (getting finalized checkpoint from store rather than state

### LINKED ISSUE #7905: [CONSENSUS SPEC] use finalized checkpoint from store rather than state
https://github.com/ethereum/consensus-specs/pull/3543

StefanBratanov:
## PR Description
We already get the finalizedCheckpoint from the store when constructing P2P Status message and for block and blob sidecars gossip validation, so just changed a comment.

image

image

## Fixed Issue(s)
fixes #7905 

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #7851, opened 2023-12-22
Add `builder_boost_factor` to blockV3 validator API


zilm13:
### Description
See appropriate blockV3 API section: https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/block.v3.yaml#L46-L78
The feature is similar to our [validators-builder-bid-challenge-percentage](https://github.com/Consensys/teku/pull/6878) but since the change validator will be able to set it via blockV3 call for any block production request. 
We should also decide whether blockV3 overrides BN settings (if any).

tbenr:
related to https://github.com/Consensys/teku/issues/7779

tbenr:
@zilm13 I think values coming from VC should override BN value, which I consider it as a default.
VC represents the staker, which have the ultimate right to decide how the flow should behave.

PR #7927, opened 2024-01-26
Add builder boost factor part1

tbenr:
- Adds `requestedBuilderBoostFactor` eventually coming from BlockV3
- The `requestedBuilderBoostFactor` will always override the eventual comparison factor configured on BN side.
- Uniforms our internal `builderComparisonFactor` representation to the new param (from `Optional<Integer>` to `Uint64`) so the handling is simplified

In our CLI we could support `PREFER_BUILDER` as an alias to `BUILDER_ALWAYS` and start advertising the new maintaining backward compatibility (separate PR)

related to #7851 

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
+++ b/ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
@@ -59,8 +59,9 @@ public class ExecutionBuilderModule {
   private static final Logger LOG = LogManager.getLogger();
   private static final int HUNDRED_PERCENT = 100;
 
-  private static final UInt64 VC_BUILDER_BOOST_FACTOR_PREFER_EXECUTION = UInt64.ZERO;
-  private static final UInt64 VC_BUILDER_BOOST_FACTOR_PREFER_BUILDER = UInt64.MAX_VALUE;
+  public static final UInt64 BUILDER_BOOST_FACTOR_MAX_PROFIT = UInt64.valueOf(HUNDRED_PERCENT);
+  public static final UInt64 BUILDER_BOOST_FACTOR_PREFER_EXECUTION = UInt64.ZERO;
+  public static final UInt64 BUILDER_BOOST_FACTOR_PREFER_BUILDER = UInt64.MAX_VALUE;
 
   private final Spec spec;
   private final AtomicBoolean latestBuilderAvailability;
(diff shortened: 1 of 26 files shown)


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
```diff
@@ -518,16 +543,44 @@ private void logReceivedBuilderBid(final BuilderBid builderBid) {
...
+      final UInt256 localPayloadValue,
+      final Optional<UInt64> requestedBuilderBoostFactor) {
+    final boolean isDefaultComparison;
+    final Optional<?> actualComparisonFactor;
+    final String comparisonFactorSource;
+
+    if (requestedBuilderBoostFactor.isPresent()) {
+      // If the requestedBuilderBoostFactor is set,
+      // we always use it over builderBidCompareFactor to determine whether the local payload wins
+      isDefaultComparison =
```
zilm13: Not a claim, just a suggestion: I think {max_uint64}% will look ugly in log. We could probably use some string keywords for 0 and max, like you did already with `MAX_PROFIT`

zilm13 (commented):
LGTM, but it would be great to see some more tests for it. They could be probably done in separate PR if not forgotten
Tests could include:
1. builderBoost input is delivered to `ExecutionBuilderModule.builderGetHeader()` from any source (we have very branchy createBlock flow, all looks impossible, but at least any)
2. more like `builderGetHeaderGetPayload_shouldGivePriorityToRequestedBuilderBoostFactor`, maybe unit tests for methods from ExecutionBuilderModule, even in API description it's highlighted to be careful with overflow and there could be bunch of others cases (I see by eyes that it shouldn't overflow, but who knows!)
3. Maybe considering verbosity and size of current integration tests there is a sense to do just a unit tests for   `isLocalPayloadValueWinning` and maybe `logLocalPayloadWin`

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionBuilderModule.java
tbenr: ah you're right. I'll do that

tbenr:
My plan is to move value comparison tests on a separate PR

services/executionlayer/src/main/java/tech/pegasys/teku/services/executionlayer/ExecutionLayerConfiguration.java
```diff
@@ -290,26 +292,24 @@ private void validateBuilderCircuitBreaker() {
...
       if (builderBidCompareFactor.endsWith("%")) {
         builderBidCompareFactor =
             builderBidCompareFactor.substring(0, builderBidCompareFactor.length() - 1);
       }
-      final int builderBidCompareFactorInt;
+      final UInt64 builderBidCompareFactorUint64;
       try {
-        builderBidCompareFactorInt = Integer.parseInt(builderBidCompareFactor);
-      } catch (final NumberFormatException ex) {
+        builderBidCompareFactorUint64 = UInt64.valueOf(Integer.parseInt(builderBidCompareFactor));
```
zilm13: We have Uint64.valueOf(String)

zilm13 (approved):
LGTM

PR #7929, opened 2024-01-28
implement getStateForBlockProduction for late block reorg

rolfyone:
Currently late block reorg is disabled so this defers to `getStateAtSlotExact`

This functionality diverges from `getStateAtSlotExact` with late block reorg, as sometimes the state we're wanting to use is the parent of the head. In this scenario we actually need to check the proposer head, and get the appropriate state based on that instead.

The fallback given all things being equal currently is to call getStateAtSlotExact, which is basically falling back to the old functionality.
 - if late block reorg is disabled
 - if the head root can't be read
 - if the headRoot is the same as the root we're told to use from getProposerHead.

 In late block reorg case where we're using parent, we need to call `regenerateBeaconState` to apply empty slots.

 partially addresses #6595

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
```diff
@@ -240,6 +240,30 @@ public SafeFuture<Optional<BeaconState>> getStateAtSlotExact(final UInt64 slot)
...
+    final Optional<Bytes32> headRoot = getBestBlockRoot();
+    if (headRoot.isEmpty()) {
+      return getStateAtSlotExact(slot);
+    }
+    final Bytes32 root = recentChainData.getProposerHead(headRoot.get(), slot);
+    if (root.equals(headRoot.get())) {
+      return getStateAtSlotExact(slot);
+    }
+    // otherwise we're looking for the parent slot
+    return getStateByBlockRoot(root)
```
zilm13, line 257: very nit: you could add a comment that getStateAtSlotExact is already transitioned to target slot,  while byRoot not, so we need to wind it forward. Actually the main confusion is from `get` in naming of  `getStateForBlockProduction` and  `getStateAtSlotExact`, it's not like we get just some deterministic object as you expect with get, we calculate it for target slot, so maybe worth altering these names instead

storage/src/test/java/tech/pegasys/teku/storage/client/CombinedChainDataClientTest.java
```diff
@@ -146,6 +150,79 @@ void getsEarliestAvailableBlobSidecarSlot() {
...
 
+  @Test
+  void getStateForBlockProduction_directsToStateAtSlotExact()
+      throws ExecutionException, InterruptedException {
+    final BeaconState state = dataStructureUtil.randomBeaconState(UInt64.valueOf(2));
+    final Optional<Bytes32> recentBlockRoot =
+        Optional.of(spec.getBlockRootAtSlot(state, UInt64.ONE));
+    when(recentChainData.getBlockRootInEffectBySlot(UInt64.ONE)).thenReturn(recentBlockRoot);
+    when(recentChainData.getStore()).thenReturn(store);
+    when(store.retrieveStateAtSlot(any()))
```
zilm13: nit: I would require ONE here, to be sure, we haven't requested any other slot

storage/src/test/java/tech/pegasys/teku/storage/client/CombinedChainDataClientTest.java
```diff
@@ -146,6 +150,79 @@ void getsEarliestAvailableBlobSidecarSlot() {
...
+        .thenReturn(SafeFuture.completedFuture(Optional.of(state)));
+    final SafeFuture<Optional<BeaconState>> future =
+        client.getStateForBlockProduction(UInt64.ONE, false);
+    assertThat(future.get()).contains(state);
+  }
+
+  @Test
+  void getStateForBlockProduction_whenEnabeldAndHaveNoBestBlockRoot()
```
zilm13: whenEnab_led_

storage/src/test/java/tech/pegasys/teku/storage/client/CombinedChainDataClientTest.java
```diff
@@ -146,6 +150,79 @@ void getsEarliestAvailableBlobSidecarSlot() {
...
+    when(store.retrieveStateAtSlot(any()))
+        .thenReturn(SafeFuture.completedFuture(Optional.of(state)));
+
+    final SafeFuture<Optional<BeaconState>> future =
+        client.getStateForBlockProduction(UInt64.ONE, true);
+    assertThat(future.get()).contains(state);
+  }
+
+  @Test
+  void getStateForBlockProduction_whenEnabeldAndBestBlockRootMatches()
```
zilm13: same, _Enabled_

storage/src/test/java/tech/pegasys/teku/storage/client/CombinedChainDataClientTest.java
```diff
@@ -146,6 +150,79 @@ void getsEarliestAvailableBlobSidecarSlot() {
...
+      throws ExecutionException, InterruptedException {
+    final BeaconState state = dataStructureUtil.randomBeaconState(UInt64.valueOf(2));
+    final Optional<Bytes32> recentBlockRoot =
+        Optional.of(spec.getBlockRootAtSlot(state, UInt64.ONE));
+    when(recentChainData.getStore()).thenReturn(store);
+
+    when(recentChainData.getBestBlockRoot()).thenReturn(recentBlockRoot);
+    when(recentChainData.getProposerHead(any(), any())).thenReturn(recentBlockRoot.get());
+    when(recentChainData.getBlockRootInEffectBySlot(UInt64.ONE)).thenReturn(recentBlockRoot);
```
zilm13, line 211: It's not verified here and in `getStateForBlockProduction_whenEnabeldAndHaveNoBestBlockRoot` that we actually walked this path. Adding `verify` could be beneficial

zilm13 (approved):
LGTM, just some nit comments to show I have checked it :)

storage/src/test/java/tech/pegasys/teku/storage/client/CombinedChainDataClientTest.java
rolfyone: gez i must have been all thumbs!

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
rolfyone: yeh i kind of agree, but i think the 'stateForBlockProduction' part is closer to describing it than stateAtSlotExact. That was a confusing thing to understand

zilm13: no, no, I like stateForBlockProduction, I meant something like:
getStateForBlockProduction -> calcStateForBlockProduction
getStateAtSlotExact ->  calcStateForSlotExact
hate calc anyway. So the main point is with get you expect just getting already known info from DB or smth like this.

storage/src/test/java/tech/pegasys/teku/storage/client/CombinedChainDataClientTest.java
rolfyone: it was just messy, but can...

storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
rolfyone: I'm going to leave the name for now just because its consistent but i might get inspired... :)

storage/src/test/java/tech/pegasys/teku/storage/client/CombinedChainDataClientTest.java
rolfyone: added verifies to test cases.

PR #7933, opened 2024-01-29
Reduce amount of logs when there are issues fetching deposits from EL

### LINKED ISSUE #7908: When nethermind is not able to provide deposits, reduce the amount of log traffic
Commonly, nethermind comes up 'in sync' and doesnt actually have deposits to return, and this is actually quite a significant issue.

At the very least though, we can reduce the level of logging about missing deposits, it's currently coming out many times a minute and makes the log impossible to read.  We shouldn't need this to be reported nearly as often, and we don't really need the stack trace in this circumstance.

ref. 
https://discord.com/channels/697535391594446898/697539289042649190/1199212110228488302

```
2024-01-23 04:39:47.001 INFO  - Syncing     *** Target slot: 8263397, Head slot: 8223412, Remaining slots: 39985, Connected peers: 66 
2024-01-23 04:39:47.370 FATAL - PLEASE CHECK YOUR ETH1 NODE | Encountered a problem retrieving deposit events from eth1 endpoint: Expected next deposit at index 1171691, but got 1171692
```

StefanBratanov:
## PR Description

- Reduce the logs spamming every 2 seconds in case of issues with fetching deposits to every 30 seconds. 
- Removed the stack trace because the InvalidDepositEventsException message contains all the relevant information.
- Moved hardcoding of `org.web3j:utils` version to `versions.gradle`

## Fixed Issue(s)
fixes #7908 

## Documentation

## Changelog

### DIFF
--- a/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositFetcher.java
+++ b/beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositFetcher.java
@@ -43,6 +43,9 @@ import tech.pegasys.teku.ethereum.pow.api.Eth1EventsChannel;
 import tech.pegasys.teku.ethereum.pow.api.InvalidDepositEventsException;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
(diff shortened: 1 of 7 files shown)


tbenr (approved):
LGTM

PR #7937, opened 2024-01-30
Read entries of an ERA file

rolfyone:
Currently can't unpack the framed snappy data, but this is a logical starting point which implements the reading of indexes and iterates the objects.

Included the python and a reference to the markdown that defines the file format.

This is not plugged into anything, currently just a work in progress with a static main fn. but has enough that we should save this as a starting point for the next iteration.

## Documentation

## Changelog


tbenr (approved):
LGTM semi-blindly :)

rolfyone:
> LGTM semi-blindly :)

thats fine i just dont want to work it out again :)

PR #7943, opened 2024-02-01
[Spike] New block rewards cache

### LINKED ISSUE #7942: Improve BlockV3 endpoint
Currently I think we should:
- Improve performance to provide BlockV3's `Eth-Consensus-Block-Value`
  - reuse calculations we do during state transition when produce the block
- expose the metadata to in-process VC too, instead of do additional steps only on when responding to Rest API

tbenr:
fixes: #7942 

- introduces a new cache (`StateTransitionCaches`) attached to the state that can be used to
  - store temporary intermediate data to help speeding up the state transition (atm not used)
  - store calculated data coming from state transition and block processing in general (including execution payload related stuff). I use here to cache block rewards (consensus) and payload execution value

- updated state transition functions to:
  - update block rewards in cache
  - call callback events (`onSlotProcessed` and `onBlockProcessed`)

- renamed `TransitionCaches` to `EpochTransitionCaches`

- it also makes `BlockContainerAndMetaData` be the data structure of the `ValidatorAPIChannel`. Not everything has been implemented here but the idea is that we can make the typeDef client "restore" it including all the metadata from the header variables.

## Documentation

## Changelog


tbenr:
closing this spike. It is good to go. Splitting this in multiple PRs.

PR #7945, opened 2024-02-02
Combine `ReceivedBlockListener` and `BlockImportNotifications`

StefanBratanov:
## PR Description

- Create new `ReceivedBlockEventsChannel`
- Removed `ReceivedBlockListener` and `BlockImportNotifications`
- Move sending `onBlockValidated` notifications to `BlockGossipValidator`
- Move sending all `onBlockImported` notifications to `BlockImporter`

This also fixes the current functionality of sending `block` SSE events when block is known. (already imported with `forkChoice.onBlock`)

## Fixed Issue(s)
N/A

## Documentation

## Changelog


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/block/BlockImporterTest.java
```diff
@@ -313,9 +313,19 @@ public void importBlock_parentBlockFromSameSlot() throws Exception {
   }
 
   private void assertSuccessfulResult(final BlockImportResult result) {
+    assertSuccessfulResult(result, true);
```
rolfyone, line 316: if the import result is false, wouldn't that imply notifications aren't sent?
I was thinking this would be something like `assertSuccessfulResult(result, result.isSuccessful())`

rolfyone:
I need to take a closer look before i tick it, did ask a question... but just stopped to say I really really appreciate the narrow focus of this pr. 👍 🥇

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/block/BlockImporterTest.java
StefanBratanov: My idea is that known blocks are still deemed successful, but they not necessarily need to be sent since they have already been imported and notification has been sent for them.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AttestationManager.java
```diff
@@ -197,7 +197,10 @@ private void applyFutureAttestations(final UInt64 slot) {
   }
 
   @Override
-  public void onBlockImported(final SignedBeaconBlock block) {
+  public void onBlockValidated(final SignedBeaconBlock block) {}
```
rolfyone: Nit: 
Im happy with this kind of format, but above we've used `//noop` in the function...

stylistically can probably just pick one format and stick with that....

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
```diff
@@ -197,7 +181,10 @@ public void subscribeDataUnavailable(final DataUnavailableSubscriber subscriber)
   }
 
   @Override
-  public void onBlockImported(final SignedBeaconBlock block) {
+  public void onBlockValidated(final SignedBeaconBlock block) {}
```
rolfyone: Same nit re. consistency...

rolfyone (approved):
LGTM, one nit

PR #7953, opened 2024-02-06
block-rewards-cache-part1

tbenr:
based on the spike #7943

introduces a new cache (`SlotCaches`) attached to the state that can be used to
- store calculated data coming from state transition and block processing in general (including execution payload related stuff). I use here to cache block rewards (consensus) and payload execution value
- store temporary intermediate data to help speeding up the state transition (atm not used)

related to https://github.com/Consensys/teku/issues/7942

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/AbstractBeaconState.java
```diff
@@ -31,25 +31,30 @@ public abstract class AbstractBeaconState<TMutable extends MutableBeaconState>
     extends SszContainerImpl implements BeaconState, BeaconStateCache {
 
   private final TransitionCaches transitionCaches;
+  private final SlotCaches slotCaches;
 
   protected AbstractBeaconState(final BeaconStateSchema<?, ?> schema) {
     super(schema);
     transitionCaches = TransitionCaches.createNewEmpty();
+    slotCaches = SlotCaches.createNewEmpty();
```
zilm13: nit: let's add this. to these 2 too

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/SlotCaches.java
```diff
@@ -0,0 +1,91 @@
...
+ * onSlotProcessed callback.
+ *
+ * <p>These caches are meant to store data used during the state transition and along block
+ * production flow.
+ */
+public class SlotCaches {
+  private volatile UInt64 blockProposerRewards = UInt64.ZERO;
+  private volatile UInt256 blockExecutionValue = UInt256.ZERO;
+
+  private static final SlotCaches NO_OP_INSTANCE =
```
zilm13, line 32: I'm not sure this is actually NO_OP, `set` and `increase` will still do their jobs

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/AbstractMutableBeaconState.java
```diff
@@ -51,17 +53,28 @@ protected AbstractMutableBeaconState(T backingImmutableView, boolean builder) {
...
+        builder ? TransitionCaches.createNewEmpty() : transitionCaches,
+        builder ? SlotCaches.createNewEmpty() : slotCaches);
   }
 
   protected abstract T createImmutableBeaconState(
-      TreeNode backingNode, IntCache<SszData> viewCache, TransitionCaches transitionCache);
+      TreeNode backingNode,
+      IntCache<SszData> viewCache,
+      TransitionCaches transitionCaches,
+      SlotCaches stateTransitionCaches);
```
zilm13: could you put some words on why do you call it `stateTransitionCaches`?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/BeaconStateCache.java
```diff
@@ -23,5 +24,13 @@ static TransitionCaches getTransitionCaches(BeaconState state) {
         : TransitionCaches.getNoOp();
   }
 
+  static SlotCaches getSlotCaches(BeaconState state) {
```
zilm13, line 27: but `getSlotCaches` here

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/StateCachesTest.java
```diff
@@ -0,0 +1,171 @@
...
+      case NONE -> {}
+    }
+
+    return chainBuilder.generateBlockAtSlot(3, blockOptions).getBlock();
+  }
+
+  private enum BlockRewardsSource {
+    ATTESTATION,
+    SYNC_COMMITTEE,
+    SLASHING,
```
zilm13: we could test both proposer slashing and attestation slashing separately for confidence

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/StateCachesTest.java
```diff
@@ -0,0 +1,171 @@
...
+      assertThat(BeaconStateCache.getSlotCaches(stateAtSlot3WithBlock).getBlockProposerRewards())
+          .matches(rewards -> rewards.isGreaterThan(UInt64.ZERO));
+    }
+
+    // execution value should not be affected by block processing
+    assertThat(BeaconStateCache.getSlotCaches(stateAtSlot3WithBlock).getBlockExecutionValue())
+        .isEqualByComparingTo(UInt256.valueOf(42));
+  }
+
+  private SignedBeaconBlock createWorthyBlock(final BlockRewardsSource blockRewardsSource) {
```
zilm13, line 131: it would be great to also make a test that we got sum of this if all are in block if it doesn't look too complex

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/AbstractMutableBeaconState.java
tbenr: ah that just comes from previous spike...renaming..

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/SlotCaches.java
tbenr: mm.. yeah... I used the same structure of `TransitionCaches` but at the moment there are no real caches, so NO_OP doesn't really do much. but yes, the NOOP will still behave the same way.
I actually think that `copy` should be like the base class. Removing it.

zilm13: I thought logic of noop copy was to keep singleton

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/StateCachesTest.java
tbenr: problem is that unless i calculate the exact expected reward, I cannot differentiate between one or more rewards

tbenr: maybe I can try to reuse `RewardCalculator` if I don't fall into dependency hell

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/StateCachesTest.java
tbenr: It would be nice but:
- we don't have a `ProposerSlashingGenerator`
- the both codepaths go into the same place

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/StateCachesTest.java
zilm13: > problem is that unless i calculate the exact expected reward, I cannot differentiate between one or more rewards

I meant run all separate first, then together

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/StateCachesTest.java
zilm13: > the both codepaths go into the same place

yeah I know, but it's tests, it's not bad to write them without knowledge of implementation
you never know what will be broken
paths could split in the future or whatever
if it's difficult, ok, let's skip

zilm13 (approved):
LGTM, just fix the spotless

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/StateCachesTest.java
tbenr: you're right.. let me try to add `ProposerSlashingGenerator`

PR #7964, opened 2024-02-08
Prune pending pools content and Blobsidecar trackers every epoch

zilm13:
## PR Description
Current behavior is to prune every 10 epochs, so in normal finalization we have 8 extra epochs of stuff we don't need, which is not only memory, but could cause issues with blobSidecarTrackers, because they are active items (see `BlockBlobSidecarsTrackersPoolImpl`). Plus removing little number of items more often is always more comfortable for GC.
Some notes:
- I don't want to query slots per epoch every slot, because actually SLOTS_PER_EPOCH is constant for all milestones, even if this is changed, nothing will be broken
- I think it's good to offset something from near epoch transition so moved it to the first slot (or maybe I'm too paranoid)

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
@@ -32,11 +32,13 @@ import tech.pegasys.teku.storage.api.FinalizedCheckpointChannel;
  */
 abstract class AbstractIgnoringFutureHistoricalSlot
     implements SlotEventsChannel, FinalizedCheckpointChannel {
+  public static final UInt64 PRUNE_SLOT = UInt64.ONE;
   private final Spec spec;
 
   // Define the range of slots we care about
   private final UInt64 futureSlotTolerance;
   private final UInt64 historicalSlotTolerance;
+  private final UInt64 slotsInEpoch;
 
   private volatile UInt64 currentSlot = UInt64.ZERO;
   private volatile UInt64 latestFinalizedSlot = GENESIS_SLOT;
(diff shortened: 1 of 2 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
```diff
@@ -46,12 +48,13 @@ public AbstractIgnoringFutureHistoricalSlot(
     this.spec = spec;
     this.futureSlotTolerance = futureSlotTolerance;
     this.historicalSlotTolerance = historicalSlotTolerance;
+    this.slotsInEpoch = UInt64.valueOf(spec.getGenesisSpec().getSlotsPerEpoch());
```
tbenr, line 51: not a fan of this.

tbenr: but ok :)

tbenr:
Thinking if we should also lower down historicalSlotTolerance for `BlockBlobSidecarsTrackersPoolImpl` only

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
```diff
@@ -46,12 +48,13 @@ public AbstractIgnoringFutureHistoricalSlot(
...
   }
 
   @Override
   public void onSlot(final UInt64 slot) {
     currentSlot = slot;
-    if (currentSlot.mod(historicalSlotTolerance).equals(UInt64.ZERO)) {
+    if (currentSlot.mod(slotsInEpoch).equals(PRUNE_SLOT)) {
```
tbenr, line 57: I really do think that we should change `onSlot(slot)` and make it `onSlot(slot, epoch, slotInEpoch)` so we damn calculate this things once

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
zilm13: Yeah i've noted this thing in PR description
I don't want to call spec.onSlot(slot).getSlotsPerEpoch on every slot when it's the same result all life
My code hurts me a bit too, but...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/AbstractIgnoringFutureHistoricalSlot.java
zilm13: yeah, not the first place where it could help, but it should be a separate PR

zilm13: SlotAndMeta  😁

tbenr (approved):
LGTM

PR #7970, opened 2024-02-09
Move block rewards calculation to dedicated spec util

tbenr:
The reason is because we can use the block rewards calculation for other purpose

I left non-block rewards calculation in original class.

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/RewardCalculator.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/RewardCalculator.java
@@ -13,11 +13,7 @@
 
 package tech.pegasys.teku.api;
 
-import static tech.pegasys.teku.spec.constants.IncentivizationWeights.PROPOSER_WEIGHT;
-import static tech.pegasys.teku.spec.constants.IncentivizationWeights.WEIGHT_DENOMINATOR;
-
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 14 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/RewardCalculator.java
```diff
@@ -26,76 +22,50 @@
...
+    this.blockRewardCalculatorUtil = blockRewardCalculatorUtil;
   }
 
-  ObjectAndMetaData<BlockRewardData> getBlockRewardData(
+  public ObjectAndMetaData<BlockRewardData> getBlockRewardDataAndMetaData(
       final BlockAndMetaData blockAndMetaData, final BeaconState parentState) {
     return blockAndMetaData.map(
-        __ -> getBlockRewardData(blockAndMetaData.getData().getMessage(), parentState));
+        __ ->
+            BlockRewardData.fromInternal(
```
zilm13: altair exception was wrapped in `BadRequestException` here too, so you need to just call local `getBlockRewardData` here

zilm13 (approved):
LGTM, just one nit

data/provider/src/main/java/tech/pegasys/teku/api/RewardCalculator.java
tbenr: ah ye good catch

PR #7972, opened 2024-02-12
Skip fork choice update if it has been processed

### LINKED ISSUE #7814: `/eth/v1/validator/attestation_data` taking a long time to respond
### Description

Teku takes significantly longer to respond to a request for attestation data made on `/eth/v1/validator/attestation_data` than other beacon nodes.

A recent example on mainnet showing the time taken to receive attestation data from four different beacon nodes:

screenshot_2023-12-11-234627

The calls to lighthouse, nimbus and prysm took a few milliseconds, whereas teku took 1-2 orders of magnitude longer.  Given that the attestation data should come directly from head state, this seems like a long time to have to wait for the data.

Every slot.

teku/v23.10.0/linux-x86_64/-privatebuild-openjdk64bitservervm-java-17

rolfyone:
Basically when preparing for attestations, we re-run FcU, and while we try to ensure this only happens once, we should be able to not run it at all now that FcU is run during block processing.

We should double check this on a testnet just to make sure the assertion that we can skip this extra FcU is correct. If it's an incorrect assertion we'd expect to see a significant decrease in finding the correct head in attestations.

fixes #7814

## Documentation

## Changelog


rolfyone:
compared this against several nodes, looks good, will mark ready for review.

tbenr:
have you measured any benefits? Asking because I suspect we are already skipping execution here:
https://github.com/ConsenSys/teku/blob/f31b239f7788f07d8e9c2b6e7544c5e16b01a259/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceRatchet.java#L63-L70

I'll do more checks

rolfyone:
> have you measured any benefits? Asking because I suspect we are already skipping execution here: https://github.com/ConsenSys/teku/blob/f31b239f7788f07d8e9c2b6e7544c5e16b01a259/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceRatchet.java#L63-L70
> 
> I'll do more checks

i could put a debug trace in, but im fairly sure its different... this way even via the ratchet we skip, so rather than being dependent on attestations running, it's just saying 'hey im already at current, im moving on'.

rolfyone:
the allocation did appear in transfers, so i guess that's correct?
https://starkscan.co/contract/0x01aa168324d83d19ca99a529a34927f9d52ea65f8b726e555b3f42e32ddea658#token-transfers

rolfyone:
@tbenr reminder

tbenr (approved):
LGTM

PR #7977, opened 2024-02-13
add validate blob sidecars debug command

### LINKED ISSUE #7947: Restore DebugDbCommand `validate-blob-sidecars`
Restore DebugDbCommand `validate-blob-sidecars` with BlobSidecars removed in #7085

mehdi-aouadi:
## PR Description
Add `validate-blob-sidecars` debug command which performs the following checks:
- Missing blob sidecars for finalized blocks
- Missing blob sidecars for hot blocks
- Orphaned blob sidecars (no related finalized block or hot blocks)
- Unpruned blob sidecars (outside of the DA window and still existing)

## Fixed Issue(s)

fixes #7947 

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
```diff
@@ -432,6 +435,244 @@ public int dumpBlobSidecars(
...
+            orphanBlobsSidecars++;
+            System.err.printf(
+                "ERROR: blob sidecar has no corresponding finalized or hot block %s%n",
+                blobSidecarKey);
+          }
+
+          // check Data Availability
+          if (blobSidecarKey.getSlot().isLessThan(lowerSlotBoundDataAvailability)) {
+            unprunedBlobsSidecars++;
+            System.err.printf("WARNING: non pruned blobs sidecar %s%n", blobSidecarKey);
```
zilm13, line 667: there could be a lot of them if `epochsStoreBlobs` override were used, not sure how to handle this

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
```diff
@@ -432,6 +435,244 @@ public int dumpBlobSidecars(
...
+        }
+      }
+    } catch (DatabaseStorageException ex) {
+      System.out.println("Failed to open database");
+    }
+
+    final long endTimeMillis = System.currentTimeMillis();
+    System.out.printf(
+        "Done. Performed %s block->blobs_sidecars and %s blobs_sidecars->block checks in %s ms%n",
+        blocksToSidecarsCheckCounter,
```
zilm13, line 679: I think originally it was expected that these 2 numbers match if everything is correct (we had 1 blobsSidecar per block), now `blocksToSidecarsCheckCounter` is blocks counter and `sidecarsToBlocksCheckCounter` is sidecars counter. So I think you should increment `blocksToSidecarsCheckCounter` by the number of commitments

zilm13 (commented):
I don't have a DB currently locally, so leaving Approve to smb who could check its actual job

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
```diff
@@ -432,6 +435,244 @@ public int dumpBlobSidecars(
...
+
+          // check Data Availability
+          if (blobSidecarKey.getSlot().isLessThan(lowerSlotBoundDataAvailability)) {
+            unprunedBlobsSidecars++;
+            System.err.printf("WARNING: non pruned blobs sidecar %s%n", blobSidecarKey);
+          }
+        }
+      }
+    } catch (DatabaseStorageException ex) {
+      System.out.println("Failed to open database");
```
zilm13, line 672: I'd add ex.printStacktrace() here

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
tbenr: that's a good point. An option could be to add a DA override param to the CLI to se the expected earliest slot

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
tbenr: mm.. ye, there is value in checking that numbers are the same, so we could print `block_commitments->blob_sidecars`

tbenr (commented):
Looks good. I'd just add the suggestions we had and is good to merge (tested on an interop)

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
mehdi-aouadi: Done

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
mehdi-aouadi: makes sense! added an optional `--lower-data-availability-epoch` option

teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
mehdi-aouadi: Added a `commitmentsToBlobSidecarsCheckCounter` which counts the expected blob sidecars according to the block commitments. This way the `sidecarsToBlocksCheckCounter` and the new `commitmentsToBlobSidecarsCheckCounter` should match if everything is fine. Decided to leave the block checks counter separate but could change that if you think it's better

zilm13: But the message looks correct
```
Done. Performed 163 block->blobs_sidecars and 491 blobs_sidecars->block checks in 4401 ms
```
I'd add just another counter for commitments count and add it to the message

zilm13: @mehdi-aouadi yeeah, it's what I was going to suggest. Perfect!

zilm13 (approved):
LGTM

zilm13:
has this kind of db btw
```
Done. Performed 163 block->blob_sidecars, 491 blob_sidecars->block checks and 492 block_commitments->blob_sidecars in 3095 ms
	Calculated Data Availability window (slots): 7487424 - 7618525
	Earliest BlobSidecars slot: Optional[7618305]
	Earliest finalized block slot: Optional[7618306]
	Latest finalized block slot: Optional[7618432]
	Missing finalized Blob Sidecars: 0
	Missing hot Blob Sidecars: 0
	Orphaned Blob Sidecars (corresponding block is missing): 5
	Non-pruned BlobSidecars (older than DA lower bound): 0
```
the trick is one block is both in finalized and in hot, not sure how could we simplify such finds

mehdi-aouadi:
> has this kind of db btw
> 
> ```
> Done. Performed 163 block->blob_sidecars, 491 blob_sidecars->block checks and 492 block_commitments->blob_sidecars in 3095 ms
> 	Calculated Data Availability window (slots): 7487424 - 7618525
> 	Earliest BlobSidecars slot: Optional[7618305]
> 	Earliest finalized block slot: Optional[7618306]
> 	Latest finalized block slot: Optional[7618432]
> 	Missing finalized Blob Sidecars: 0
> 	Missing hot Blob Sidecars: 0
> 	Orphaned Blob Sidecars (corresponding block is missing): 5
> 	Non-pruned BlobSidecars (older than DA lower bound): 0
> ```
> 
> the trick is one block is both in finalized and in hot, not sure how could we simplify such finds

Hmm didn't see this coming tbh. We could check if a hot block is already finalized before counting it but that's tedious

tbenr:
Or exclude hot blocks having slot <= finalized slot? (So we shouldn't check them twice)

mehdi-aouadi:
> Or exclude hot blocks having slot <= finalized slot? (So we shouldn't check them twice)

Definitely simpler 😄 Pushed a commit...

zilm13:
I think it's a task for something like blocks-validate. Because it's a thing you may want to know but it's not related to blob sidecars in anyway.

zilm13:
Anyway it should be strict isGreater without equal if merged here.
I'd prefer somthing like this, to be honest
```java

          if (hotBlockSlot.isGreaterThanOrEqualTo(lowerSlotBoundDataAvailability)) {
            // exclude finalized hot blocks (keep non finalized hot blocks only so they are not
            // counted twice)
            if (hotBlockSlot.isLessThanOrEqualTo(maybeLatestFinalizedSlot.orElse(UInt64.ZERO))) {
              System.err.printf(
                  "ERROR: Skipped hot block in slot %s, finalized already%n", hotBlockSlot);
              continue;
            }
```

PR #7980, opened 2024-02-14
Update deposit tree snapshots Feb 2024

zilm13:
## PR Description
I wonder why it's just 150k deposits in Holesky, but everything else looks good for me. So all genesis validators are not in deposit contract?
```
2024-02-14 12:15:05.020+04:00 | Test worker | INFO  | teku-status-log | A deposit snapshot was read with 150372 deposits and 0x98b484b0ac880615f02279e29625866881b154f84b5964e42c558920927e7b27 execution block hash.
```

## Documentation

## Changelog


StefanBratanov (approved):
Yeah, my impression is that only deposits after genesis are in the deposit tree snapshots

PR #7981, opened 2024-02-14
Make checkpoint sync work when finalized state is transitioned with empty slot(s)

### LINKED ISSUE #7956: Checkpoint sync is not working when finalized state is transitioned with empty slots
### Description
If finalized epoch ends with empty slot(s) most clients provide state transitioned to the end of the epoch, so it cannot be used for `AnchorPoint` initialization without extra work. To fix this we should also request finalized block header from peer and patch finalized state so it could be used for `AnchorPoint`.

Discussion on Discord: https://discord.com/channels/697535391594446898/697539289042649190/1202083995815522364

StefanBratanov:
## PR Description
Override the consistency check for an AnchorPoint  since the finalized state can be transitioned with empty slot(s)

## Fixed Issue(s)
fixes #7956 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/AnchorPoint.java
```diff
@@ -126,6 +126,18 @@ public static AnchorPoint fromInitialBlockAndState(
     return new AnchorPoint(spec, checkpoint, state, block);
   }
 
+  @Override
+  protected void verifyStateAndBlockConsistency() {
+    if (state.getSlot().isGreaterThan(blockSummary.getSlot())) {
+      // the finalized state is transitioned with empty slot(s)
+      checkArgument(
```
zilm13, line 135: I'd also add verification for state.block_roots (block slot - state slot interval) and state.state_roots (block slot)

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -858,13 +858,10 @@ private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
...
 
-    // if finalized is gone from cache we can still reconstruct that without regenerating
+    // if finalized is gone from cache we can use the finalized anchor without regenerating
     if (finalizedAnchor.getRoot().equals(blockRoot)) {
       LOG.trace("epochCache GET finalizedAnchor {}", finalizedAnchor::getSlot);
-      return SafeFuture.completedFuture(
-          Optional.of(
-              StateAndBlockSummary.create(
-                  finalizedAnchor.getBlockSummary(), finalizedAnchor.getState())));
+      return SafeFuture.completedFuture(Optional.of(finalizedAnchor));
```
zilm13, line 864: If it's transitioned anchor, it will be incorrect state for requested block_root

zilm13 (commented):
I like the way you've kept the check in `Summary` while avoiding it in some cases in `AnchorPoint` but I still have a lot of doubts on 2 major points of this change:
- Could weakening of AnchorPoint creation checks introduce a bugs when AnchorPoint is created not from checkpoint sync it has not failed, but it's a bad AnchorPoint
- How transitioned AnchorPoint could affect its consumers when we start Teku 

While from quick check I don't see any issues with 1st, 2nd could be definitely an issue in many places. I've commented one case in the `Store`, another one for example is initalization of the `Store`:
```
    // Track latest finalized block
    this.finalizedAnchor = finalizedAnchor;
    this.maybeEpochStates = maybeEpochStates;
    states.cache(finalizedAnchor.getRoot(), finalizedAnchor);
```
This way we are putting incorrect cache in states cache which is widely used. And there are a lot of other places where finalizedAnchor spreads, we should check them all and decide how to handle it. Maybe this kind of Anchor will need extra identification so we could avoid polluting some things like this state cache with it.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
StefanBratanov: The stateRoot of the anchor is still the block state root though

StefanBratanov: `AnchorPoint.getStateRoot()` still returns the `state_root` of the finalzied block (last block header), not the state

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/AnchorPoint.java
StefanBratanov: Added those checks, made it bit more complex, since don't have access to `spec` in super class.

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
zilm13: Method ` private SafeFuture<Optional<StateAndBlockSummary>> getOrRegenerateBlockAndState(
final Bytes32 blockRoot)` should return state at block with blockRoot, transitioned state is not this state. This method is used by a lot of different consumers, it could be an issue

StefanBratanov: Yeah `AnchorPoint.getStateRoot()` returns the `state_root` of the `blockSummary`, so would be the state associated with the block

StefanBratanov: In any case, `StateAndBlockSummary.create(finalizedAnchor.getBlockSummary(), finalizedAnchor.getState())` is same as just returning `finalizedAnchor`

tbenr: On startup we definitely create the Store with the anchor coming from the checkpoint URL. So it is possible that we could return it from here...

StefanBratanov:
Closing this in favour of https://github.com/Consensys/teku/pull/8221

PR #7984, opened 2024-02-15
add the teku path to the env allowing it to be called directly when using a custom entrypoint in k8s

joshuafernandes:
## PR Description
Add the teku path to the env allowing it to be called directly when using a custom entrypoint in k8s


rolfyone (approved):
LGTM

lucassaldanha (approved):

PR #7987, opened 2024-02-15
Block Rewards Cache Part 2

### LINKED ISSUE #7942: Improve BlockV3 endpoint

tbenr:
- Uses `BlockContainerAndMetaData` in `ValidatorApiChannel` for block production
- BlockFactory now uses the new cache to provide block values
- modifies the TypedefClient to get these values from header variables\json and recreate the full `BlockContainerAndMetaData`

fixes #7942 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #7992, opened 2024-02-18
Run engine capabilities monitoring only on startup/restart

StefanBratanov:
## PR Description
Change current logic of running engine capabilities monitoring every 10 epochs to only running it on EL startup/restart using the `ExecutionClientEventsChannel`.

The negative is that the warning would only be raised once on startup or restart of EL instead of every 10 epochs as right now.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


StefanBratanov:
Closing this, because regular polling is more useful.

PR #8000, opened 2024-02-20
Removing recursiion asserts from Eth2P2PNetworkOptionTest

lucassaldanha:
## PR Description

Removing some unnecessary assertions on `TekuConfig` object in favour of targeted assertions on the expected values being overwritten in the config.


rolfyone (approved):
LGTM

PR #8006, opened 2024-02-22
Fix throw when ExecutionPayloadContext is empty and move it to the top

zilm13:
## PR Description
- previous message was not correct for the case when it happens
- pre-Deneb it will think that it's preMerge even if merge already happens

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -206,6 +206,13 @@ private SafeFuture<Void> setExecutionData(
       final BeaconState blockSlotState,
       final BlockProductionPerformance blockProductionPerformance) {
 
+    if (spec.isMergeTransitionComplete(blockSlotState) && executionPayloadContext.isEmpty()) {
```
StefanBratanov, line 209: I think we don't need the `isMergeTransitionComplete()` check. It's already covered by the `bodyBuilder.supportsExecutionPayload()` check a bit up the stack.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -246,10 +253,7 @@ private SafeFuture<Void> setExecutionData(
     // Post-Deneb: Execution Payload / Execution Payload Header + KZG Commitments
     final ExecutionPayloadResult executionPayloadResult =
         executionLayerBlockProductionManager.initiateBlockAndBlobsProduction(
-            // kzg commitments are supported: we should have already merged by now, so we
-            // can safely assume we have an executionPayloadContext
-            executionPayloadContext.orElseThrow(
-                () -> new IllegalStateException("Cannot provide kzg commitments before The Merge")),
+            executionPayloadContext.orElseThrow(),
```
StefanBratanov, line 256: We can just `.get()` here for more simple. It's safe if we check only the emptiness above

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
zilm13: bellatrix supports execution payload but it could be both pre merge and post merge

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
zilm13: code style barks at me

zilm13: because it doesn't know in this part of code whether merge transition complete or not

tbenr (approved):
LGTM

PR #8009, opened 2024-02-23
Improve fcu attributes calculation stability

tbenr:
With this PR I want to solve some problems related to forkchoice notifier.

TLDR;
1. by adding `calculatePayloadAttributesSlot` which guarantees more stability in the slot for which we calculate attributes for, I solved the `A block import occurs while a proposal is in progress` scenario.
2. by removing the FcU call I removed the issues related to VC posting proposers preparation data.

I don't see drawbacks on 1.
Very minimal\absent for 2. - we currently run forkchoice notifier (via `processHead`) right before the block creation. So we will be able to propose an local 0-transaction block (or a normal one via builder)  even in the case we don't do anything during previous slot. (same thing if it is just an update to the fee recipient).

## proposer data is updated
VC clients regularly sends proposer data to "subscribe" to a BN, data also contains the fee recipients they intended to use for block production for each validator.

Before this PR we were trying to recalculate payload attributes for the next slot, with the refreshed proposer data, with the intention to cover the edge case (like VC startup) in which we received new proposer data for a proposer that is about to build a block for the next slot.

Reminder: VCs are also regularly calling BN with the whole dataset every epoch. (Teku currently does it at the beginning of the second slot).

This behaviour opens 2 problems:

### Scenario 1: we are at slot 2 and we are the proposer of that slot

the sequence will be:

`slot 1 - 4s into slot`: fcu called with attributes for slot 2: EL starts building
`slot 2 - 0s into slot`: VC calls both (the order in which teku receives matters, but assuming the following)
  - `createUnsignedBlock`
  - `prepare_beacon_proposer`

BN, while building the block, will also call EL with FcU with attributes for slot 3.

#### Scenario 1a: we are not proposer of slot 3

BN will call FcU with empty attributes. If EL is smart, it will keep the previous block building alive, waiting for getPayload call.

When teku tries to obtain the payload, internally it will have the wrong `forkChoiceUpdateData` (no context) so it will try to recover by calling FcU again with the correct payload attributes (slot 2) and obtain a payload id, then call getPayload immediately.

in the following real world case the fcu due to proposer data change happened before the block creation preparation, so we called FcU 2 times, in sequence, first `without` attributes (slot 771) and later `with` attributes (slot 770, the correct one).
image

#### Scenario 1b: we are also proposer of slot 3

BN will call FcU with attributes for slot 3. I don't know if EL are designed to start a block production (for slot 3) while previous block (slot 2) is still under construction.

In any case when teku tries to obtain the payload, internally it will have the wrong `forkChoiceUpdateData` (not viable payload Id) so it will try to recover by calling FcU again with the correct payload attributes (slot 2) and obtain a payload id, then call getPayload immediately.

### Scenario 2: we are at slot 2 and we are the proposer of slot 3

`slot 2 - 0s into slot`: VC calls 
  - `prepare_beacon_proposer`

BN will call FcU with attributes for slot 3. EL starts building the block. This call might be too early.

https://github.com/ethereum/go-ethereum/blob/5a0f468f8cb15b939bd85445d33c614a36942a8e/miner/payload_building.go#L213

if slot 2 turns out to be empty, block building on EL side is never refreshed to it could timeout (12s) before we call getPayload.

## A block import occurs while a proposal is in progress

When we import a block we end up calling the forkchoice notifier, with `proposerSlot` set to `empty`.

If we are proposer for `slot n` and we see block for `slot n-1` very late, we will reorg it out (once the feature will be ready) BUT:
- `slot n-1 - 4s into slot`: we call FcU with attributes for `slot n`. EL starts building on top of block from `n-2`
- `slot n-1 - 11.99s into slot`: block for `n-1` is imported. It won't change our head but notifier will called with `empty` `proposerSlot`, which will cause FcU to be called with attributes for `slot n+1` so with empty attributes (or wrong attributes)
- `slot n - 0s into slot`: block creation will trigger another FcU due to wrong context (non-viable or missing payloadId)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifierImpl.java
```diff
@@ -220,14 +221,58 @@ private void internalForkChoiceUpdated(
...
+            .getPayloadBuildingAttributes()
+            .map(PayloadBuildingAttributes::getProposalSlot);
+
+    if (maybeCurrentPayloadAttributesSlot.isPresent()
+        // we are still in the same slot as the last proposing slot
+        && currentSlot.get().equals(maybeCurrentPayloadAttributesSlot.get())
+        // we have not yet imported our own produced block
+        && forkChoiceState.getHeadBlockSlot().isLessThan(maybeCurrentPayloadAttributesSlot.get())) {
+
+      LOG.info("Returned same slot as current payload attributes", new Exception("Stack trace"));
```
rolfyone: this for our debug?

rolfyone: i cant really approve a log.info with `new Exception` :)

tbenr: Sure:) wasn't intended to be approved in this current form

tbenr:
@rolfyone I think `A block import occurs while a proposal is in progress` scenario has touching points with your late block reorg

StefanBratanov (approved):
LGTM

rolfyone (approved):
LGTM

PR #8018, opened 2024-02-26
Make Validator loading error logs more visible

zilm13:
## Fixed Issue(s)

Current behavior:
- errors during validator loading doesn't come to the console
- if, say, 64 validators are loaded and error happens on №32, no validators are loaded at all (with no message to console)
- application is not shutdown without flag "exit-when-no-validator-keys-enabled"

This change addresses visibility of errors during validators loading

## Documentation

## Changelog


rolfyone:
should be fully rectified with more explanation etc etc in #8036 if you want to close this?

zilm13:
Scrapped this in favor of #8044

PR #8035, opened 2024-02-29
Move fallback log and some renaming

tbenr:
Main thing is moving logging and recording of fallback when we make the decision (during block creation) and not during unblinding.

## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #8036, opened 2024-03-01
Exit with clear error messaging on lock file issues

### LINKED ISSUE #8024: When a lock file is present, teku vc hangs
When keys are unable to be loaded by the teku validator client due to the presence of a lock file it can't remove,

There is nothing displayed on the console (assuming split logging) and no duties will be performed by any of the other keys that didn't have a lock file present.

It looks like some kind of unhandled exception during key load.

rolfyone:
- When a lock file is present and cannot be removed, this enters us into a state where duties will never be loaded or performed.
 - A single lock file from a large set that cannot be removed is sufficient to show this problem

 Lock files are typically created containing the PID, and if we can determine it's safe to clean up, we will do that and not fail, but in some cases the PID is either active, or the lock file may be empty (full file system for example)

 The DSL for AT is a little messy generally, and I did have to change one function name to make it more sensible, but tried to avoid substantial refactoring.

 This is a code 2 exit, which we use to indicate that a user will have to get involved to improve the situation - eg. validator slashing detected will also exit code 2. In this case a service could not bother restarting automatically, as it won't recover without help.

 Added a number of scenarios for the BN, and a scenario to show this specific failure for VC.

 fixes #8024

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -271,14 +271,21 @@ public static ValidatorClientService create(
...
         .exceptionally(
             error -> {
-              final Optional<Throwable> maybeCause =
-                  ExceptionUtil.getCause(error, InvalidConfigurationException.class);
-              if (maybeCause.isPresent()) {
-                LOG.warn(maybeCause.get().getMessage());
-              } else {
-                LOG.error("Error was encountered during validator client service start up.", error);
-              }
-              checkNoKeysLoaded(validatorConfig, validatorLoader);
```
rolfyone, line 281: `checkNoKeysloaded` here was the core problem of the ticket leaving us parked.
In this case we need to system.exit, because some or all of the chain above has failed, and the last item in the chain is to perform duties, so we're literally parked with no duties.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuValidatorNode.java
```diff
@@ -98,14 +99,50 @@ public ValidatorKeysApi getValidatorKeysApi() {
...
             + validatorKeystores.getKeysDirectoryName()
             + ":"
             + WORKING_DIRECTORY
             + validatorKeystores.getPasswordsDirectoryName());
-    this.copyContentsToWorkingDirectory(validatorKeystores.getTarball());
+    return this;
+  }
+
+  // will mount keystores folder as a read only filesystem
+  public TekuValidatorNode withReadOnlyKeystorePath(
```
rolfyone, line 118: almost duplicated from TekuNode, we need to circle back at some point on the structure of these 2 classes

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuDepositSender.java
```diff
@@ -90,8 +90,14 @@ public ValidatorKeystores generateValidatorKeys(
   }
 
   public ValidatorKeystores generateValidatorKeys(int numberOfValidators) {
+    return generateValidatorKeys(numberOfValidators, false);
+  }
+
+  public ValidatorKeystores generateValidatorKeys(int numberOfValidators, boolean isLocked) {
```
mehdi-aouadi: Let's make them final

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -118,6 +118,10 @@ public void specificationFailure(final String description, final Throwable cause
     log.warn("Spec failed for {}: {}", description, cause, cause);
   }
 
+  public void failedToLoadValidatorKey(final String message) {
```
mehdi-aouadi, line 121: I'd rename it to `failedToLoadValidatorKeys` or `failedToLoadValidatorKeyStore` since we're talking about all the keys

mehdi-aouadi (approved):
LGTM
Just some nits.
I agree with the concept that if a user specifies a `validator-keys` option and we're unable to load the keys for any reason (fs errors, locks...) we must exit

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
mehdi-aouadi: I agree that this option should only be considered when everything (keys loading) run without exception only

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -271,14 +271,21 @@ public static ValidatorClientService create(
...
+              ExceptionUtil.getCause(error, InvalidConfigurationException.class)
+                  .ifPresentOrElse(
+                      cause -> LOG.error(cause.getMessage()),
+                      () ->
+                          LOG.error(
+                              "Error was encountered during validator client service start up.",
+                              error));
+
+              LOG.error(
```
tbenr, line 282: I'm not sure about always logging this error, no matter where\why we throw the exception. So it could be misleading if it has nothing to do with keys.

tbenr (approved):
LGMT but I have doubts with a log error.

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
rolfyone: it's literally 1 key, so for me i'm ignoring this one.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
rolfyone: I can remove it, but this is the unhandled case we used to have so we'd have a stack trace... stack traces can be handy...

rolfyone: I'll leave this as it is literally the old cause i've refactored to be cleaner..

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuDepositSender.java
rolfyone: added a bunch, the class is pretty messy

PR #8040, opened 2024-03-02
Refactored acceptance test DSL

rolfyone:
Acceptance test DSL is getting a little out of control so moved a few things around...

Reduces the duplication of code, does mean some options might be invalid under some of the paths, but hopefully overall a fair amount cleaner.

## Documentation

## Changelog


rolfyone:
might try an alternative version where this becomes a builder pattern, which might be a lot cleaner.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/AttestationGossipAcceptanceTest.java
```diff
@@ -16,27 +16,30 @@
...
-                    .withRealNetwork()
-                    .withPeers(node1)
-                    .withInteropValidators(32, 32));
+    final TekuBeaconNode node2 =
+        createTekuBeaconNode(
+            TekuNodeConfigBuilder.createBeaconNode()
+                .withGenesisTime(genesisTime.intValue())
+                .withRealNetwork()
+                .withPeers(node1)
+                .withInteropValidators(32, 32)
```
lucassaldanha, line 41: It is missing network name "minimal". The default is "swift".

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/SentryNodesAcceptanceTest.java
```diff
@@ -41,42 +41,45 @@ void sentryBeaconNodesSetup() throws Exception {
...
-      final TekuNode primaryNode, final int genesisTime) {
-    return c ->
-        c.withGenesisTime(genesisTime)
+  private TekuBeaconNode createLateJoiningNode(
+      final TekuBeaconNode primaryNode, final int genesisTime) throws IOException {
+    return createTekuBeaconNode(
+        TekuNodeConfigBuilder.createBeaconNode()
+            .withGenesisTime(genesisTime)
             .withRealNetwork()
-            .withNetwork("minimal")
```
lucassaldanha, line 78: Missing "minimal" network on new config?

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuBeaconNode.java
```diff
@@ -0,0 +1,769 @@
...
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.versions.bellatrix.BeaconStateBellatrix;
+import tech.pegasys.teku.spec.generator.BlsToExecutionChangeGenerator;
+import tech.pegasys.teku.spec.logic.versions.capella.block.BlockProcessorCapella;
+import tech.pegasys.teku.spec.schemas.SchemaDefinitionsCapella;
+import tech.pegasys.teku.spec.signatures.SigningRootUtil;
+import tech.pegasys.teku.test.acceptance.dsl.Eth2EventHandler.PackedMessage;
+import tech.pegasys.teku.test.acceptance.dsl.tools.deposits.ValidatorKeystores;
+
+public class TekuBeaconNode extends TekuNode {
```
lucassaldanha, line 102: I think calling this a Beacon Node can be a bit misleading as in most cases it will be a combined node.

lucassaldanha (approved):
I like the change to use builders. IMO it makes everything simpler. Happy to merge as is but I have two points of attention:
- In some cases, we removed the network option from the node configuration (e.g. "minimal") and now are relying on the default ("swift"). Not sure if this was intended or not.
- I think the rename to `TekuBeaconNode` can be misleading because it will act as a combined node in most cases.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/SentryNodesAcceptanceTest.java
rolfyone: it didnt need it... ran fine on swift...

rolfyone:
> * I think the rename to `TekuBeaconNode` can be misleading because it will act as a combined node in most cases.

IMO the differentiating factor is that it's running beacon node, and having 'TekuNode' was a generic term i could use to centralise, which is why this ended up BeaconNode. I'm not sure i'd call it 'CombinedNode' because all our nodes are technically unless they're VC....

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/AttestationGossipAcceptanceTest.java
rolfyone: minimal just makes the test run slower unless we have a reason to use it...

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuBeaconNode.java
rolfyone: The nice thing is we can use `Node` -> `TekuNode` -> `TekuBeaconNode`|`TekuValidatorNode`
Some things only run on either side, and it was named based on teku's default mode being a `BeaconChainController` and the actual command for running a validator is `ValidatorClient`
I think it's consistent with the use of the application...
`TekuNode` which was previously used now has startup logic etc which reduces code duplication which was getting pretty messy.

PR #8043, opened 2024-03-05
Improve block equivocation check for block publishing V2 API

### LINKED ISSUE #8039: Occasional failure to public blocks with `/eth/v2/beacon/blocks`
### Description

When publishing blocks via the `/eth/v2/beacon/blocks` endpoint on holesky I occasionally see the following error:

```
2024-03-01 19:43:03.420 WARN  - Block (and blob sidecars) publishing skipped due to broadcast validation result GOSSIP_FAILURE for slot 1118315
```

Other consensus nodes publish the block without issue.

**Frequency:**

Intermittent.  

### Versions

`teku/v24.2.0/linux-x86_64/-privatebuild-openjdk64bitservervm-java-17`

tbenr:
Based on the discussion happening in the issue.

The objective is to make gossip broadcast validation to fail only in case of equivocating blocks (slashable, proposer proposed two different blocks for the same slot) and let it pass when it is the same block root that is causing the gossip ignore.

In case of "already seen", we can say for sure that the block has been seen but we currently don't know anything about blobs. So the current decision is to publish block and blobs in that condition too, and let the libp2p layer to filter the already seen. If detected we log a debug line with the exception:
image

fixes #8039

## Documentation

## Changelog


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorTest.java
```diff
@@ -150,13 +169,15 @@ public void shouldReturnConsensusFailureImmediatelyWhenConsensusCompleteExceptio
     verifyNoMoreInteractions(blockGossipValidator);
   }
 
-  @Test
-  public void shouldReturnSuccessWhenSecondEquivocationCheckIsValidated() {
+  @ParameterizedTest
+  @EnumSource(value = EquivocationCheckResult.class)
+  public void shouldReturnFinalEquivocationFailureOnlyForEquivocatingBlocks(
```
zilm13, line 174: nit: final

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
```diff
@@ -76,14 +80,18 @@ public SafeFuture<InternalValidationResult> validate(
...
-        || !blockIsFirstBlockWithValidSignatureForSlot(block)) {
-      LOG.trace(
-          "BlockValidator: Block is either too old or is not the first block with valid signature for "
-              + "its slot. It will be dropped");
+    if (gossipValidationHelper.isSlotFinalized(block.getSlot())) {
+      LOG.trace("BlockValidator: Block is either too old. It will be dropped");
       return completedFuture(InternalValidationResult.IGNORE);
     }
 
+    final InternalValidationResult validationResult =
```
zilm13: I'd call it something like `intermediateVlidationResult` to highlight that if we get accept it's not final accept

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
```diff
@@ -76,14 +80,18 @@ public SafeFuture<InternalValidationResult> validate(
   private SafeFuture<InternalValidationResult> internalValidate(
       final SignedBeaconBlock block, final boolean isLocallyProduced) {
 
-    if (gossipValidationHelper.isSlotFinalized(block.getSlot())
-        || !blockIsFirstBlockWithValidSignatureForSlot(block)) {
-      LOG.trace(
-          "BlockValidator: Block is either too old or is not the first block with valid signature for "
-              + "its slot. It will be dropped");
+    if (gossipValidationHelper.isSlotFinalized(block.getSlot())) {
+      LOG.trace("BlockValidator: Block is either too old. It will be dropped");
```
zilm13: without `either`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
```diff
@@ -163,21 +171,57 @@ private SafeFuture<InternalValidationResult> internalValidate(
...
+    }
+
+    if (maybePreviouslySeenBlockRoot.get().equals(block.getRoot())) {
+      return BLOCK_ALREADY_SEEN_FOR_SLOT_PROPOSER;
+    }
+
+    return EQUIVOCATING_BLOCK_FOR_SLOT_PROPOSER;
+  }
+
+  EquivocationCheckResult performBlockEquivocationCheck(final SignedBeaconBlock block) {
```
zilm13, line 217: and also here I think something like `performIntermediateBlockEquivocationCheck` will highlight purpose

zilm13 (commented):
Looks good, just a few nits but I still not sure on consequences of publishing duplicate

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockGossipValidator.java
tbenr: mm... here I disagree because it is called to do the final equivocation check when building the pipeline in `buildValidationPipeline`

zilm13 (approved):
LGTM

PR #8044, opened 2024-03-05
Improve logging on validator loading errors

zilm13:
## PR Description
Currently we print only `InvalidConfigurationException` to default console log in `MultithreadedValidatorLoader.loadValidators()` and shutdown on any exception during validator startup. It could confuse user seeing exit without any visible cause (until user uses --log-destination=console, which is not default or check logs). 
- Validator loading errors in `loadValidators()`  could be not only `InvalidConfigurationException`
- Validator loading errors could happen in `addValidatorsFromSource`, for example when local keys directory is provided which doesn't exist
- Other Validator Service Loading errors should be displayed to the user in any case as we shutdown

## Documentation

## Changelog


zilm13:
3 acceptance tests fail because locally it's:
```
ERROR - Failed to load keystore, error: Keystore file /opt/teku/keys/9894149_validator.json.lock already in use.
ERROR - Unable to initialize validator client service, please manually correct errors and try again.
```
and on CI it's
```
ERROR - Failed to load keystore, error: Access Denied trying to access lock file /opt/teku/keys/ae00073_validator.json.lock
ERROR - Unable to initialize validator client service, please manually correct errors and try again.
```

Working to resolve this

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNodeConfigBuilder.java
```diff
@@ -317,6 +317,18 @@ public TekuNodeConfigBuilder withWritableKeystorePath(
     return this;
   }
 
+  public TekuNodeConfigBuilder withDefaultLogging() {
+    configMap.remove("log-destination");
+    return this;
+  }
+
+  public TekuNodeConfigBuilder withBadKeystorePath() {
+    LOG.debug("Xinterop-enabled=false");
```
rolfyone: i'd be at least adding here that we're using a bad path as part of this debug..thats a pretty important breakage if you call this fn.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNodeConfigBuilder.java
```diff
@@ -317,6 +317,18 @@ public TekuNodeConfigBuilder withWritableKeystorePath(
     return this;
   }
 
+  public TekuNodeConfigBuilder withDefaultLogging() {
+    configMap.remove("log-destination");
```
rolfyone, line 322: add debug that we're removing log destination

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -271,16 +271,15 @@ public static ValidatorClientService create(
...
         .exceptionally(
             error -> {
-              ExceptionUtil.getCause(error, InvalidConfigurationException.class)
-                  .ifPresentOrElse(
-                      cause -> LOG.error(cause.getMessage()),
-                      () ->
-                          LOG.error(
-                              "Error was encountered during validator client service start up.",
-                              error));
+              LOG.error("Error was encountered during validator client service start up.", error);
```
rolfyone: i dont really want this unconditionally to dump exception trace, its just noise if we know what the problem is

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -271,16 +271,15 @@ public static ValidatorClientService create(
...
+              if (ExceptionUtil.hasCause(error, ValidatorLoaderException.class)) {
+                STATUS_LOG.failedToLoadValidatorKey(ExceptionUtil.getRootCauseMessage(error));
+              } else {
+                STATUS_LOG.errorDuringValidatorClientInitialization(
+                    ExceptionUtil.getRootCauseMessage(error));
+              }
 
-              LOG.error(
-                  "Unable to initialize validator keys, please manually correct errors and try again.");
+              STATUS_LOG.failedToStartValidatorClient();
```
rolfyone, line 285: i'd be ok with this change so that its on the console, but im not convinced about the 274-280 change, i think it was better previously in terms of the content delivered.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
```diff
@@ -72,11 +72,15 @@ private ValidatorLoader(
...
-    slashingProtectionLogger.protectionSummary(ownedValidators.getValidators());
+    try {
+      final Map<BLSPublicKey, ValidatorProvider> validatorProviders = new HashMap<>();
+      validatorSources.forEach(source -> addValidatorsFromSource(validatorProviders, source));
+      MultithreadedValidatorLoader.loadValidators(
+          ownedValidators, validatorProviders, graffitiProvider);
+      slashingProtectionLogger.protectionSummary(ownedValidators.getValidators());
+    } catch (final Exception ex) {
+      throw new ValidatorLoaderException(ex);
+    }
```
rolfyone: this wraps with our new exception, but it doesn't really add value... we were already checking the main cases... so the only difference is now we always get a ValidatorLoaderException in this specific case, but for all other stages of loading, we don't really get more insight. I'm not sure it's that valuable unless we're trying to split out config exceptions here from other config exceptions.

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNodeConfigBuilder.java
```diff
@@ -317,6 +317,18 @@ public TekuNodeConfigBuilder withWritableKeystorePath(
...
+    return this;
+  }
+
+  public TekuNodeConfigBuilder withBadKeystorePath() {
```
lucassaldanha: I don't think it is worth polluting our DSL with such a specific usage, specially being a "negative" usage that is unlikely to see any reuse.

I'd argue for adding (more flexible way of setting up the path: 
```
withValidatorKeysPath(String path) {
  configMap.put("validator-keys", path);
}
```

And then, when creating your node, you can use:
```
final TekuBeaconNode beaconNode =
        createTekuBeaconNode(
            TekuNodeConfigBuilder.createBeaconNode()
                .withDefaultLogging()
                .withValidatorKeysPath("/abcde:/abcdef")
                .withInteropModeDisabled()
                .withInitialState(genesis)
                .build());
```

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
zilm13: I want to clearly separate keys loading and other exceptions.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ValidatorLoader.java
zilm13: I want to clearly separate keys loading and other exceptions. If you log with "keystore" wording when `scheduleValidatorsDuties` or `initializeValidatorRestApi` fails, it could be confusing. Same way when your keys directory not found, it's keystore issue

zilm13: Though `initializeValidatorRestApi` is kind of a part of a keystore too...
I didn't like full exception name with packages path when it's not `InvalidConfigurationException`

rolfyone: actually very academic. the known ones were keystore, and given the error string and stack trace at exit of unkown we'd have adequate information...

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -119,7 +119,19 @@ public void specificationFailure(final String description, final Throwable cause
   }
 
   public void failedToLoadValidatorKey(final String message) {
-    log.error("Failed to load keystore, error {}", message);
+    log.error("Failed to load keystore, error: {}", message);
+  }
+
+  public void errorDuringValidatorClientInitialization(final String message) {
+    log.error(
+        "Error was encountered during validator client service start up. error: {}. "
```
tbenr: "An error ..." ?

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -119,7 +119,19 @@ public void specificationFailure(final String description, final Throwable cause
...
+            + "Please check the logs for details.",
```
rolfyone, line 129: could probably put "\n" before Please...
```
"An error was encountered during validator client service startup.\n"+
"Error: {}\n" +
"Please check the logs for details."
```

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNodeConfigBuilder.java
```diff
@@ -317,6 +317,18 @@ public TekuNodeConfigBuilder withWritableKeystorePath(
     return this;
   }
 
+  public TekuNodeConfigBuilder withDefaultLogging() {
+    LOG.debug("Removed log-destination from config");
+    configMap.remove("log-destination");
+    return this;
+  }
+
+  public TekuNodeConfigBuilder withValidatorKeysPath(final String path) {
```
rolfyone: all of the other functions setting validator keys are named differently, and i think this needs to be consistent so people don't find this one and expect it to work...
We currently have
`withReadOnlyKeystorePath`
`withWritableKeystorePath`

I'd rename this function to 
`withInvalidKeystorePath`

If someone uses this function instead of the others, the files won't be present etc.

I'd also copy the debug mostly so 
`LOG.debug("set validator-keys folder to INVALID {}", path);`

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuNodeConfigBuilder.java
```diff
@@ -317,6 +317,18 @@ public TekuNodeConfigBuilder withWritableKeystorePath(
     return this;
   }
 
+  public TekuNodeConfigBuilder withDefaultLogging() {
+    LOG.debug("Removed log-destination from config");
+    configMap.remove("log-destination");
+    return this;
+  }
+
+  public TekuNodeConfigBuilder withInvalidKeystorePath(final String path) {
```
lucassaldanha, line 326: This method should be: `withKeystorePath(final String path)` and the user decides if they put an invalid path or not.

zilm13:
closed in favor of #8049

Issue #7898, opened 2024-01-18
[EIP-6110] Supply validator deposits on chain


lucassaldanha:
### Description
Provides validator deposits as a list of deposit operations added to the Execution Layer block

- [x] Add the necessary config
- [x] Add new container `DepositReceipt`
- [x] Extend `ExecutionPayload`  with `deposit_receipts`
- [x] Extend `ExecutionPayloadHeader` with `deposit_receipts_root`
- [x] Append BeaconState with `deposit_receipts_start_index`
- [x] Modify `ElectraStateUpgrade` to set the new field to `UNSET_DEPOSIT_RECEIPTS_START_INDEX`
- [x] Create/Modify `getPayloadV4` and `newPayloadV4` using the new ExecutionPayload
- [x] Modify `get_eth1_deposit_count` during block proposal
- [x] Modify block processing when state transitioning https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/eip6110/beacon-chain.md#block-processing
- [x] Stop `Eth1Data` polling/voting when `state.eth1_deposit_index == state.deposit_receipts_start_index`
- [x] Make `ValidatorIndexCache` re-org resistant

### References
- https://eips.ethereum.org/EIPS/eip-6110
- https://github.com/ethereum/consensus-specs/tree/dev/specs/_features/eip6110
- https://www.youtube.com/watch?v=tRTBgCN9VgY

PR #8048, opened 2024-03-06
Refactor `process_deposit` to be closer to specs

StefanBratanov:
## PR Description

- Bring the code implementation of `process_deposit` bit closer to specs: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#deposits
- Also will help EIP-6110 by making `applyDeposit` method accessible.

## Fixed Issue(s)
will help #7898 

## Documentation

## Changelog


zilm13 (approved):
LGTM
I see debug line for `DepositWithIndex` was removed but I cannot find a single way how it could arrive in process_deposit. Maybe it was somewhere in an old code.

StefanBratanov:
> LGTM I see debug line for `DepositWithIndex` was removed but I cannot find a single way how it could arrive in process_deposit. Maybe it was somewhere in an old code.

Yeah. I think it was old debug code I assume. Not important anymore in any case.

PR #8053, opened 2024-03-07
Add validation on loaded DepositTreeSnapshot

zilm13:
## PR Description

Some endpoints could provide broken deposit tree snapshot and we choose it instead of bundled one because of the lack of the validation. Let's validate snapshot right after loading.
See: https://github.com/status-im/nimbus-eth2/issues/5508

## Documentation

## Changelog


beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
```diff
@@ -68,11 +70,19 @@ public LoadDepositSnapshotResult loadDepositSnapshot() {
...
         STATUS_LOG.loadingDepositSnapshotResource(sanitizedUrl);
         final DepositTreeSnapshot depositTreeSnapshot = loadFromUrl(depositSnapshotResourceUrl);
+        // Validate
+        DepositTree.fromSnapshot(depositTreeSnapshot);
         STATUS_LOG.onDepositSnapshot(
             depositTreeSnapshot.getDepositCount(), depositTreeSnapshot.getExecutionBlockHash());
         return LoadDepositSnapshotResult.create(Optional.of(depositTreeSnapshot));
       } catch (final Exception e) {
-        LOG.warn("Failed to load deposit tree snapshot from " + sanitizedUrl, e);
+        if (ExceptionUtil.hasCause(e, IllegalArgumentException.class)) {
```
tbenr: Can we simply catch `IllegalArgumentException` and then `Exception` here, instead of using `hasCause`?

tbenr (approved):
LGTM, just a nit

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
rolfyone: not if it's wrapped...

rolfyone (approved):

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
tbenr: I don't think it is, but i might be wrong..

zilm13: I think it's just a safer approach. Write it safe rather than guess. And something could be changed there in the future.

tbenr: fair in this case, not sure if this approach holds in general (referring to exception handling, not to "write safe" statement :) )

zilm13: ok, removing, it's redundant

PR #8063, opened 2024-03-11
Remove point-at-infinity for proofs/commitments

jtraglia:
## PR Description

I noticed that `KZGProof` and `KZGCommitment` were using the wrong point-at-infinity bytes. It was 48 zero bytes and it should have been `0xc0` followed by 47 zero bytes. See the `G1_POINT_AT_INFINITY` constant here:

https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/polynomial-commitments.md#constants

I've made this mistake several times myself. It's the uncompressed bytes which are all zero; the compressed bytes include the "infinity" flag. This only appears to have been used in test code, so no worries.

* Additionally, the two classes were providing this value differently.
  * I like the static variable style, so I switched the other one to that style.
* I believe `BlobSidecar#toLogString` should print the abbreviated commitment.
  * I will revert this change if this was intentional.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
@@ -21,7 +21,7 @@ public class BlobSidecarsByRootRequestMessageSchema
     extends AbstractSszListSchema<BlobIdentifier, BlobSidecarsByRootRequestMessage> {
 
   // Size validation according to the spec (MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK) is
-  // done in the RPC handler.
+  // done in the RPC handler
   public BlobSidecarsByRootRequestMessageSchema(final SpecConfigDeneb specConfigDeneb) {
     super(
         BlobIdentifier.SSZ_SCHEMA,
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
@@ -20,8 +20,8 @@ import tech.pegasys.teku.spec.config.SpecConfigDeneb;
 public class BlobSidecarsByRootRequestMessageSchema
     extends AbstractSszListSchema<BlobIdentifier, BlobSidecarsByRootRequestMessage> {
 
-  // Size validation according to the spec (MAX_REQUEST_BLOCKS_DENEB *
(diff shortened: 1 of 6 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
```diff
@@ -20,8 +20,8 @@
 public class BlobSidecarsByRootRequestMessageSchema
     extends AbstractSszListSchema<BlobIdentifier, BlobSidecarsByRootRequestMessage> {
 
-  // Size validation according to the spec (MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK) is
-  // done in the RPC handler
+  // Size validation according to the spec (MAX_REQUEST_BLOCKS_DENEB *
```
zilm13: This message means we don't throw in constructor if it exceeds (MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK), we will have spec check later. So here we cap it by 2048 and later check that it's 6, so this comment is made for understanding why don't we cap here on 6 and where is actual validation.
So I'd revert this file change.

zilm13 (commented):
Is there a way to get a test for it?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
jtraglia: Oh, I see. I misread that. Will revert that change!

jtraglia:
> Is there a way to get a test for it?

Yes, I think so. I will work on that.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BlobSidecarsByRootRequestMessageSchema.java
zilm13: this line where actual validation happens
https://github.com/Consensys/teku/blob/94d37825e1fb4e2f14c3e964fa8eb68d5f8e4f1b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRootMessageHandler.java#L86

zilm13:
@jtraglia it would be good to have it in spec tests. We could be not alone

jtraglia:
> @jtraglia it would be good to have it in spec tests. We could be not alone

Good idea. I'll look into that. To back that up, mev-boost had the same issue:

* https://github.com/flashbots/mev-boost/pull/630

jtraglia:
I chatted with a coworker and we think it would be best to remove the infinity value from the two classes. It shouldn't be necessary in clients; it should only be needed internally in the KZG library. Would it be okay for me to remove these?

zilm13 (approved):
LGTM

zilm13:
Merged, thank you @jtraglia

PR #8068, opened 2024-03-12
Added header to builder call

### LINKED ISSUE #8061: POST to `/eth/v1/builder/blinded_blocks` missing header
### Description

When teku calls the endpoint `/eth/v1/builder/blinded_blocks` it does not supply an `Eth-Consensus-Version` header, as per the specs at https://ethereum.github.io/builder-specs/#/Builder/submitBlindedBlock

```
teku/v24.2.0/linux-x86_64/-privatebuild-openjdk64bitservervm-java-17
```

rolfyone:
When calling `/eth/v1/builder/blinded_blocks` we had missed the `Eth-Consensus-Version` header.

I changed the OkHttpRestClient from using HttpUrl as there was a warning on the interface we were using.

We now add the header. I manually lower-cased the milestone string, as the version method we use in schema isn't really what we want in this context, but at some point we probably want to build this logic into SpecMilestone rather than relying on the Schema `Version` object.

fixes #8061.

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java
+++ b/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java
@@ -69,7 +69,7 @@ public class OkHttpRestClient implements RestClient {
       final S requestBodyObject,
       final SerializableTypeDefinition<S> requestTypeDefinition) {
     return postAsyncInternal(
-        apiPath, Map.of(), requestBodyObject, requestTypeDefinition, Optional.empty());
+        apiPath, NO_HEADERS, requestBodyObject, requestTypeDefinition, Optional.empty());
   }
 
   @Override
--- a/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java
+++ b/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.ethereum.executionclient.rest;
 import static java.util.Objects.requireNonNull;
 
 import java.io.IOException;
+import java.net.URL;
 import java.util.Map;
 import java.util.Optional;
 import okhttp3.Call;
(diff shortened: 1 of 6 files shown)


lucassaldanha (approved):

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java
```diff
@@ -67,17 +68,23 @@ public <S> SafeFuture<Response<Void>> postAsync(
       final String apiPath,
       final S requestBodyObject,
       final SerializableTypeDefinition<S> requestTypeDefinition) {
-    return postAsyncInternal(apiPath, requestBodyObject, requestTypeDefinition, Optional.empty());
+    return postAsyncInternal(
+        apiPath, Map.of(), requestBodyObject, requestTypeDefinition, Optional.empty());
```
StefanBratanov: You can use the `NO_HEADERS` constant instead of `Map.of` to be more consistent

StefanBratanov (approved):
LGTM just one comment

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java
```diff
@@ -90,16 +97,17 @@ private <T> SafeFuture<Response<T>> getAsyncInternal(
...
       final Optional<DeserializableTypeDefinition<T>> responseTypeDefinitionMaybe) {
     final RequestBody requestBody = createRequestBody(requestBodyObject, requestTypeDefinition);
-    final Request request = createPostRequest(apiPath, requestBody);
+    final Request request = createPostRequest(apiPath, requestBody, headers);
     return makeAsyncRequest(request, responseTypeDefinitionMaybe);
   }
 
   private Request createGetRequest(final String apiPath, final Map<String, String> headers) {
-    final HttpUrl httpUrl = createHttpUrl(apiPath);
+    final URL httpUrl = createHttpUrl(apiPath);
```
StefanBratanov, line 110: One other small nit, I think returning `HttpUrl` is better, because internally in the library if you pass `URL` it converts it back anyways:

```kotlin
open fun url(url: URL) = url(url.toString().toHttpUrl())
```

rolfyone: httpurl was giving an error that it was using an internal function - thats why i chagned it...

StefanBratanov: Ok, fair. Weird one.

PR #8080, opened 2024-03-14
Remove redundant subtype

zilm13:
## PR Description
It's confusing and not used.

## Documentation

## Changelog


rolfyone (approved):
oh this is annotated, we'll be deleting these things anyway, but sure. it's not used just fyi

zilm13:
Yeah, I've tested it.
My confusion was in line with somebody reporting that it's not providing anything past altair (which is not correct)

PR #8086, opened 2024-03-15
Enable topic filtering on all gossip topics

rolfyone:
Added a test flag to call through to the p2p layer for topic filtering on startup. This doesn't appear to be currently done, but the flag will allow us to turn it off there's any compatibility issue going on.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM. I suppose the idea is to remove the flag after we verify it's working as expected?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java
```diff
@@ -68,7 +72,8 @@ private P2PConfig(
       final int batchVerifyMaxThreads,
       final int batchVerifyQueueCapacity,
       final int batchVerifyMaxBatchSize,
-      final boolean batchVerifyStrictThreadLimitEnabled) {
+      final boolean batchVerifyStrictThreadLimitEnabled,
+      boolean allTopicsFilterEnabled) {
```
tbenr, line 76: final :-D

PR #8096, opened 2024-03-18
let forkChoice onTick fail

tbenr:
*** NEEDS TO BE CAREFULLY ANALYZED ***

If there are possible deadlock conditions, we should avoid "joining" forever and let timeout so we can allow deadlock to resolve. We should double-check and, if we fail, we remain in a recoverable state, not in a broken-corrupted state.

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -25,6 +25,9 @@ import java.net.ConnectException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -298,7 +301,11 @@ public void onTick(
       final UInt64 currentTimeMillis, final Optional<TickProcessingPerformance> performanceRecord) {
     final UpdatableStore store = recentChainData.getStore();
     final UInt64 slotAtStartOfTick = spec.getCurrentSlot(store);
-    tickProcessor.onTick(currentTimeMillis).join();
+    try {
+      tickProcessor.onTick(currentTimeMillis).get(30, TimeUnit.SECONDS);
+    } catch (final InterruptedException | TimeoutException | ExecutionException e) {
+      throw new RuntimeException(e);
+    }
     performanceRecord.ifPresent(TickProcessingPerformance::tickProcessorComplete);
```
rolfyone, line 309: i think it's worth doing this finish in an 'always' block

also, we should probably put in a counter for failed ticks...

rolfyone:
potentially given the stacktrace we looked at, raised #8097 as an alternative

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
tbenr: yes, right

rolfyone:
i think if we do this, we'd plumb it in so we can tweak the value, and we'd start it high. I like the idea that timedTick can timeout, but being able to alter that value if its too low sometimes would be much easier if we could do it with cli arg...

rolfyone:
i think we can park this....

PR #8097, opened 2024-03-19
Added a timeout to state regeneration

rolfyone:
Added a development flag to allow us to set state regeneration timeout, with a 120 second timeout by default, and a tiny bit of sanity around the flag not being less than 1.

## Documentation

## Changelog


tbenr:
So this affects only the higher level `FinalizedStateCache` and not the lower `AsyncChainStateGenerator`.
This means that regenerations can still take longer, depends who trigger them.

for instance at startup:

```
2024-03-21 12:18:36.342 WARN  - State rebuild timeout is set outside of sensible defaults of 10 -> 300, 2 was defined. Cannot be below 1, will allow the value to exceed 300.
...
...
2024-03-21 12:18:47.351 INFO  - Storage initialization complete
2024-03-21 12:18:47.351 WARN  - Not loading specified initial state as chain data already exists.
2024-03-21 12:18:56.495 INFO  - Completed regeneration of state for block 0x273c1072c092c468c4999e4231f04785411ed91056e4dd728c831cf08f63625a at slot 8682944 by replaying 32 blocks. Took 8467ms
...
```

So this is in effect only on requests coming from API? Am I correct? I still get confused a lot by the regeneration...

rolfyone:
we only have one regeneration cache i think, but i'll check later on...

rolfyone:
it does look like the other one does need to be handled also. We can do that as a separate PR though...

storage/src/main/java/tech/pegasys/teku/storage/server/state/FinalizedStateCache.java
```diff
@@ -106,13 +136,21 @@ private BeaconState regenerateState(final UInt64 slot, final BeaconState stateFr
                       stateFromCache.getSlot().compareTo(stateFromDisk.getSlot()) >= 0)
               .orElse(stateFromDisk);
       if (preState.getSlot().equals(slot)) {
-        return preState;
+        return SafeFuture.completedFuture(preState);
+      }
+      final long regenerateSlotCount = slot.minusMinZero(stateFromDisk.getSlot()).longValue();
+      LOG.trace("Slots to regenerate state from: {}", regenerateSlotCount);
+      if (regenerateSlotCount > 10_000L) {
```
tbenr: we could have this constant statically defined up in the class.

tbenr (approved):
LGTM. Yes I checked that the other regeneration is completely separated from this.

storage/src/main/java/tech/pegasys/teku/storage/server/state/FinalizedStateCache.java
rolfyone: true, potentially at some point configurable, but initially 10k is a decent distance

PR #8098, opened 2024-03-19
Explicit Peering Agreement implementation

### LINKED ISSUE #8004: Explicit Peering Agreement implementation
### Context

Teku implements the concept of a “trusted”/”static” peer, which guarantees the ability to always remain connected to a peer ignoring defensive measures such as peer scoring, however it provides no guarantees of being part of their mesh or to send/receive full messages from them unconditionally.

### Why explicit peering agreement

[Gossipsub v1.1 introduces the concept of "Explicit Peering Agreement"](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#explicit-peering-agreements), which is an agreement that must be reached from a pair of nodes in order to *“remain connected to and unconditionally forward messages to each other outside of the vagaries of the peer scoring system and other defensive measures”.*

This feature if enabled in Teku could allow nodes to explicitly connect to some relays networks such as Fiber or Bloxroute to boost the propagation of their blocks.

### Possible implementation

Most of the work for supporting explicit peers (called "direct peers" in Teku and in jvm-libp2p) has been done in https://github.com/Consensys/teku/pull/3576. However, direct peers seem to be never assigned (I'd be happy to know I'm wrong). As such, there are only few things that needs to be done:

- set them from the command line
- allow them to be added and removed dynamically
- exclude them from peer scoring or other defensive measures

Lastly this list of nodes needs to be passed to the libp2p implementation.

If there is interest, we can open a PR that implements this functionality.

thedevbirb:
## PR Description

Implements [Gossipsub v1.1 "Explicit Peering Agreement"](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#explicit-peering-agreements) by introducing a new CLI flag `--direct-peers` ("direct" is used instead of "explicit" to follow [jvm-libp2p](https://github.com/libp2p/jvm-libp2p) convention). 
As such, documentation on https://github.com/Consensys/doc.teku should be updated as well to reflect this change (I can do it if you wish!).

### Implementation

From a discovering configuration perspective, direct peers are treated as static peers, as we want to connect with them immediately on startup and persist the connection with them without being kicked out by the reputation system. 
In order to do that, the list of static peers is extended with the direct peers provided in the CLI option.

Lastly, a function that determines whether a peer is direct or not is passed downstream to `jvm-libp2p` to enable the direct peer functionality on gossipsub.

## Fixed Issue(s)
Closes #8004.

## Documentation

## Changelog

### DIFF
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/gossip/config/GossipScoringConfig.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/gossip/config/GossipScoringConfig.java
@@ -16,7 +16,9 @@ package tech.pegasys.teku.networking.p2p.gossip.config;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Consumer;
+import tech.pegasys.teku.networking.p2p.gossip.config.GossipPeerScoringConfig.DirectPeerManager;
 
 /** Gossip scoring config. Contains peer scoring and topic scoring configs. */
 public class GossipScoringConfig {
@@ -161,6 +163,13 @@ public class GossipScoringConfig {
     public Builder opportunisticGraftThreshold(final Double opportunisticGraftThreshold) {
       checkNotNull(opportunisticGraftThreshold);
       this.opportunisticGraftThreshold = opportunisticGraftThreshold;
+      return this;
+    }
+
+    public Builder directPeerManager(final DirectPeerManager directPeerManager) {
+      checkNotNull(directPeerManager);
+
+      this.peerScoringConfigBuilder.directPeerManager(Optional.of(directPeerManager));
       return this;
     }
   }
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
@@ -334,6 +334,9 @@ public class P2POptions {
   }
 
   public void configure(final TekuConfiguration.Builder builder) {
+    // From a discovery configuration perspective, direct peers are static peers
+    p2pStaticPeers.addAll(p2pDirectPeers);
+
     builder
         .p2p(
             b ->
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
@@ -26,6 +26,7 @@ import tech.pegasys.teku.beacon.sync.SyncConfig;
 import tech.pegasys.teku.config.TekuConfiguration;
 import tech.pegasys.teku.networking.eth2.P2PConfig;
 import tech.pegasys.teku.networking.p2p.discovery.DiscoveryConfig;
+import tech.pegasys.teku.networking.p2p.libp2p.MultiaddrPeerAddress;
 import tech.pegasys.teku.networking.p2p.network.config.NetworkConfig;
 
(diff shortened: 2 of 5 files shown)


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/gossip/config/GossipScoringConfig.java
```diff
@@ -163,5 +167,26 @@ public Builder opportunisticGraftThreshold(final Double opportunisticGraftThresh
       this.opportunisticGraftThreshold = opportunisticGraftThreshold;
       return this;
     }
+
+    public Builder directPeers(final List<String> directPeers) {
```
StefanBratanov: I think is not as simple because the String will be the whole `multiaddress` but we only care about the `multihash`. I think below should suffice and also simplifies a bit with lambda:

```java
public Builder directPeers(final List<String> directPeers) {                  
  checkNotNull(directPeers);                                                  
  final List<NodeId> directPeerNodeIds =                                      
      directPeers.stream()                                                     
          .map(MultiaddrPeerAddress::fromAddress)                              
          .map(MultiaddrPeerAddress::getId)                                    
          .toList();                                                          
  final DirectPeerManager directPeerManager = directPeerNodeIds::contains;    
  peerScoringConfigBuilder.directPeerManager(Optional.of(directPeerManager)); 
  return this;                                                                
}                                                                                                                                                         
```

teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
```diff
@@ -148,6 +148,14 @@ public class P2POptions {
       arity = "0..*")
   private List<String> p2pStaticPeers = new ArrayList<>();
 
+  @Option(
+      names = {"--p2p-direct-peers"},
+      paramLabel = "<PEER_ADDRESSES>",
+      description = "Direct peers",
```
StefanBratanov: I think it's a good place to add description similar to what will be in the docs, can mention they will be treated as explicit peers, not get downscored, etc

StefanBratanov (commented):
Thanks for the PR. :) Added one comment about the `DirectPeerManager` initialization. Also will tag @Nashatyrev for his opinion.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/gossip/config/GossipScoringConfig.java
Nashatyrev: I would prefer to have it like below in the Config class:
```java
public Builder directPeers(final DirectPeerManager directPeers)
```
It would allow more flexible (dynamic peer set) and typesafe (`String` is pretty ambiguous) programmatic configuration

Nashatyrev (approved):
LGTM, just a minor suggestion 

@StefanBratanov it probably make sense to add a couple of libp2p unit tests to check if the `isDirect` is handled correctly

thedevbirb:
Hey everyone, thanks for the detailed review and feedback! I hope I've addressed all your suggestions now:
- downstream propagation of direct peers has been refactored by converting `List<String>` into `List<NodeId>` as soon as possible, and then converting it into `DirectPeerManager`
- better description of "static" and "direct" peers has been added to the respective CLI options

I also added one little test to make sure the CLI option and the configurations do the right thing indeed.
You can look at the single updates in the related commits.

Thanks!

StefanBratanov (approved):
LGTM to me as well. Feel free to raise a docs PR as well and we will approve it.

PR #8107, opened 2024-03-19
Add clients information to user's graffiti when producing a block

### LINKED ISSUE #8074: Append EL/CL information to user-defined graffiti
In addition to #7930 even if user has specified a graffiti we can attempt to append EL/CL information if the space permits it. This will be `opt-in` feature defaulted to false e.g. `append-client-versions-to-validators-graffiti`

Some implementation ideas: https://github.com/ethereum/execution-apis/pull/517#issuecomment-1920549999

Add a graffiti versioning "mode" (single CLI option):
Example:
- CL + CL_VERSION + EL + EL_VERSION (enabled by default)
- Only CL + EL
- Nothing

If not enough space, try other modes (e.g. if adding the version does not fit, add at least CL and EL).

Padding?? (need to adjust spec)

zilm13:
## PR Description
Adds `--validators-graffiti-client-append-format` to append clients information to user graffiti
Start from `ValidatorOptions` or `ClientGraffitiAppendFormat`.
- I'm not sure about ValidatorOptions. Feature is used only on BN side and being able to set on VC too is a bit confusing.
- I'm not sure where to add info, in the end or in the beginning. In the end looks better from my personal view for user, he gets configured graffiti first, but in the discussion it was proposed to add to the beginning (without much support), so I've added both options

Docs to be added.

## Fixed Issue(s)

Fixes #8074 

## Documentation

## Changelog


zilm13:
> Found illegal dependencies in project ':infrastructure:version'
  :ethereum:spec

ohh I will fix it.
just want to reuse createTekuVersion

acceptance-tests/src/testFixtures/java/tech/pegasys/teku/test/acceptance/dsl/TekuBeaconNode.java
```diff
@@ -484,6 +485,19 @@ public void waitForNonDefaultExecutionPayload() {
...
+    LOG.debug("Wait for a block satisfying certain assertions");
+
+    waitFor(
+        () -> {
+          final Optional<SignedBlock> block = fetchHeadBlock();
+          assertThat(block).isPresent();
+          assertThat(block.get()).satisfies(consumer);
+        },
+        5,
+        MINUTES);
```
rolfyone: 5 minutes is a long timeout for a default...

zilm13: copy-paste
will fix

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/GraffitiBuilderTest.java
```diff
@@ -0,0 +1,424 @@
...
+    assertThat(graffitiBuilder.joinNonEmpty(" ", "", "bb", "cc"))
+        .isEqualTo(Bytes32Parser.toBytes32("bb cc"));
+    assertThat(graffitiBuilder.joinNonEmpty(" ", "", "", "")).isEqualTo(Bytes32.ZERO);
+  }
+
+  @Test
+  public void formatClientInfo_shouldFitInDesiredLength() {
+    graffitiBuilder.onExecutionClientVersion(BESU_CLIENT_VERSION);
+
+    // 20: LH1be52536BU0f91a674
```
rolfyone, line 220: we're really bundling multiple tests here, so if it ever breaks we'll only see the first failure

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/GraffitiBuilderTest.java
```diff
@@ -0,0 +1,424 @@
...
+        .isEqualTo(expectedGraffitiBytes);
+  }
+
+  @Test
+  public void extractGraffiti_shouldReturnCorrectGraffiti() {
+    assertThat(graffitiBuilder.extractGraffiti(Optional.empty(), 0)).isEqualTo("");
+    assertThat(
+            graffitiBuilder.extractGraffiti(
+                Optional.of(Bytes32Parser.toBytes32(asciiGraffiti0)), 0))
+        .isEqualTo("");
```
rolfyone, line 131: same with most of this suite, we're putting a ton of tests into 1 test case, so breakage will mean that we only see first fail.

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/GraffitiBuilderTest.java
zilm13: ok, will split it

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/GraffitiBuilderTest.java
zilm13: ok, will split it

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -55,6 +56,24 @@ public class ValidatorOptions {
...
+  @Option(
+      names = {"--validators-graffiti-client-append-format"},
+      paramLabel = "<STRING>",
+      description =
+          """
+                          Appends CL and EL clients information separated with a space to user's graffiti on the Beacon Node.
+                          AUTO modes tries to provide complete version information if there is enough space for it.
+                          To disable use NONE. Available options:
+                            AUTO_END: (default) Clients info is appended after user's graffiti if any.
+                            AUTO_START: Clients info is added before user's graffiti.
```
rolfyone: i think the original ticket had a definite idea about where it should be, so allowing users to choose it somewhere else is probably bad....
https://hackmd.io/@wmoBhF17RAOH2NZ5bNXJVg/BJX2c9gja

rolfyone: its a long confusing thread, we may want to just verify what people are thinking on that front

zilm13: Yeah, I saw this one, mentioned it. It looks just like one of the options/opinions.
It's not a standard and not a big deal to parse it from the both sides.
User must have 100% right to put whatever he wants there

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -55,6 +56,24 @@ public class ValidatorOptions {
...
+                            NAME_END: Only clients names are appended after user's graffiti. Good when you don't want to disclose clients versions.
+                            NAME_START: Clients names are added before user's graffiti.
+                            NONE: Clients information is not added to the graffiti.
```
rolfyone: i'd basically only offer
AUTO - which applies client name and version tag to graffiti, and is default
CLIENT_NAME - allows us to apply client name only to the graffiti. This would be greatly appreciated so that client information can be more accurate kind of description
DISABLED - would be to turn this feature off

worth checking other implementation teams to see their plans if we haven't...
also i think we can actually use enum in cli args which may make this whole thing a little cleaner rather than making it string...

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
zilm13: Anyway I can modify it to have only START options (removing START) or mention in thread discussion that we want to give user an option to put it on any side. My personal view that END option looks better for user as graffiti belongs to him in the first place. But as I said I'm ok for any resolution.

StefanBratanov:
I think I would simplify it as well. From aesthetic point of view, I don't think many users (if any) will prefer `_START`. We can have only:

- AUTO
- ONLY_CLIENT_CODES (make it consistent with `ClientVersion` object
- NONE (I prefer it to disabled, since it is a type of format)

I prefer to keep it simple unless we start to get specific user requests that they want to append at start. And then basically append them to end with a space like how it is in this PR. Opinions?

zilm13:
@rolfyone ready for review

I've checked other clients:

Lighthouse - https://github.com/sigp/lighthouse/pull/5290 implementation does not merge user's graffiti with client information, either user's graffiti or client information
Lodestar - issue created https://github.com/ChainSafe/lodestar/issues/6463, non-default, optional merge in discussion
Prysm - issue created https://github.com/prysmaticlabs/prysm/issues/13558, change default graffiti, no merge from my understanding
Nimbus - no issue

Only Lodestar considering to implement merge optionally, so there will be few merged graffitis in network. It sounds for me like we could freely choose how to merge, add clients info to the beginning or to the end of the string. I suggest to add clients info to the end like: "Happy block TKBU". But your decision on this is final.

Also I'm thinking now that we are not covering this case:
BN could have VC calls with different proposers and may want not to merge anything, but is ok to put clients information in graffiti on proposers without graffiti. So `WHEN_EMPTY` and `CLIENT_NAMES_WHEN_EMPTY` may have sense.

And maybe `CLIENT_CODES` instead of `CLIENT_NAMES`? Codes is more correct.

zilm13:
Doc draft

```
\```bash
validators-graffiti-client-append-format: WHEN_EMPTY
\```

  </TabItem>
</Tabs>

Appends CL and EL clients information with a space to user's graffiti. On
a separate BN/VC setup should be set on a Beacon Node. This feature helps
developers and community to analyze client diversity and block anomalies.
  
Default value is `AUTO`.

Following options are available:
- `AUTO`: If user's graffiti is empty sets graffiti to reveal
  consensus and execution clients information using its codes and build commits 
  like "TK508459f2BUbb9ba13c", where "TK" is Teku, next its build commit, 
  execution layer client 2 characters code, "BU" for Besu in example, and it 
  ends with EL build 4 bytes commit in hexadecimal representation.
  
  If user's graffiti is set, this option will calculate space left 
  (graffiti size is 32 bytes) and if it's more than 5 characters left, appends 
  either full CL/EL version information or one of its compact forms right up to 
  4 characters (client codes only). So, if your graffiti is 
  "It's my first block!", block will be proposed with graffiti 
  "It's my first block TK50BUbb".
- `CLIENT_CODES`: Appends only CL/EL client codes like "TKBU". It's useful if
  you are not sure that version information cannot be used to exploit
  vulnerabilities or don't want to share any extra information.
- `WHEN_EMPTY`: Adds full client information in graffiti only if it's empty.
  Use it if you have different proposers on your node with some of them using 
  graffiti while others not, and you don't want to change user's graffiti if 
  it's set.
- `CLIENT_CODES_WHEN_EMPTY`: Adds client codes information in graffiti only if
  It's empty. Useful if you have different proposers on node and don't want to
  reveal client versions information.
- `DISABLED`: Client information is not appended. If user's graffiti is set it 
  goes as is in block, otherwise empty graffiti is used.
```

rolfyone:
I still think we should keep it simple...
```
AUTO
CLIENT_CODES
DISABLED
```
thats all we need, if people change graffiti they can change this too

zilm13:
@rolfyone it's all we have in this PR. I've just renamed `CLIENT_NAMES` -> `CLIENT_CODES`, codes looks better for me too. 
Ready for review and merge

rolfyone:
> @rolfyone it's all we have in this PR. I've just renamed `CLIENT_NAMES` -> `CLIENT_CODES`, codes looks better for me too.
> Ready for review and merge

previous PR has another 2 options... i'll check code, just trying to be clear about how many options we really want

zilm13:
@rolfyone I've removed it. Now just the list you posted.

rolfyone (approved):
LGTM

PR #8108, opened 2024-03-19
Set DEFAULT_BUILDER_BID_COMPARE_FACTOR to 90

fredrik0x:
## PR Description

In order to help increase censorship resistance, I propose to change the default DEFAULT_BUILDER_BID_COMPARE_FACTOR to 90. This means validators will prioritize local block building unless the bid from the external block builder is 10% or higher than what the validator would receive when building locally.

Looking at ([stats](https://censorship.pics/)), it can be seen that currently 63.7% of external builders are censoring transactions compared to 8.53% of validators who do local block building, so setting a minimum 10% as default can help increase the overall censorship resistance of the network.

It is still easy for users to opt out of this by manually setting the flag to 100, but many are likely to use the default which could help with censorship resistance for the network.

## Documentation

## Changelog


0xFredrik:
Nimbus: https://github.com/status-im/nimbus-eth2/pull/6103
Lighthouse: https://github.com/sigp/lighthouse/pull/5441
Lodestar: https://github.com/ChainSafe/lodestar/pull/6568
Teku: https://github.com/Consensys/teku/pull/8108
Prysm: https://github.com/prysmaticlabs/prysm/pull/13772

tbenr (approved):
LGTM

mcdee:
Just seen this; how does it work in practice?  https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/produceBlockV3 states about the builder boost factor "If this parameter is not provided then it should be treated as having the default value of 100", so this PR appears to clash with the API specification.

0xFredrik:
> Just seen this; how does it work in practice? https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/produceBlockV3 states about the builder boost factor "If this parameter is not provided then it should be treated as having the default value of 100", so this PR appears to clash with the API specification.

Good catch @mcdee , while "should" is more permissive than "must", I opted to make a PR ( https://github.com/ethereum/beacon-APIs/pull/438 ) that doesn't define default as any specific value.

PR #8110, opened 2024-03-20
increased attestation pool size limit

rolfyone:
32 Slots here in cache still seems appropriate given their relative value beyond that, but we should increase the cache limit to cater for storing more attestations.

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -68,7 +68,11 @@ public class AggregatingAttestationPool implements SlotEventsChannel {
    * <p>A limit of 40,000 attestations is enough for 3 slots worth at 400,000 validators which gives
    * a sane upper limit while still being above the typical 10-20k pool size seen on MainNet.
    */
-  public static final int DEFAULT_MAXIMUM_ATTESTATION_COUNT = 40_000;
+  /**
```
lucassaldanha: Is this supposed to be part of the comment block above?

Also, we probably don't want to mention the 40k limit anymore.

rolfyone: the logic was the basis for the following sttatement, i can try to clean it up more i was just being lazy 🗡️

lucassaldanha (approved):

PR #8113, opened 2024-03-20
Deposit tree snapshots should only be queried/loaded once

### LINKED ISSUE #7893: Deposit tree snapshots should only be queried/loaded once
### Description
At the moment we are loading deposit tree snapshots every single time we start Teku, to compare if it is a better version (more recent, etc) than what we have in our database.

The problem is that when using remote sources for the deposit tree snapshot (e.g. from checkpoint-sync-url) it becomes a failure point every time your node starts. We had users having issues with their checkpoint sync provider being unreliable and causing an issue when Teku starts.

The failure will be solved by #7715, using a fallback mechanism. However, we do not need to load the deposit snapshot tree on every startup.

### Done Criteria
- Upon a fresh start (clean DB), Teku will try to download the deposit tree snapshot from any configured source (custom, bundled or from checkpoint-sync).
- After the first snapshot is loaded successfully and persisted to the node's DB, Teku should not try to load a deposit tree snapshot from any other source than its database.

courtneyeh:
fixes #7893 

**Scenarios Covered**
1. If custom deposit snapshot path is present, use that (`--Xdeposit-snapshot`)
2. If database loader is not empty, use that
3. Finally, load from file loader, where we load the checkpoint-sync or embeded file (part of provided jar file) **ONLY IF deposit snapshot is enabled**

## Documentation

## Changelog


beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java
```diff
@@ -97,6 +97,7 @@ class Eth1DepositManagerTest {
           depositSnapshotFileLoader,
           depositSnapshotStorageLoader,
           false,
```
zilm13: So, default testing manager comes with takeBest == false. As we set it by following code in production
```java
      final boolean isBundledSnapshotEnabled =
          this.depositSnapshotEnabled && this.customDepositSnapshotPath.isEmpty();
```
it could be false with custom false only with depositSnapshot disabled

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java
```diff
@@ -383,22 +384,23 @@ void shouldStartWithStoredDepositsChainReorgedToBeShorter() {
   }
 
   @Test
-  void shouldStartFromSnapshotFileWhenProvided() {
+  void shouldIgnoreSnapshotFileWhenDBSavedVersionProvided() {
```
zilm13, line 388: so here we test deposit snapshot disabled case, no custom snapshot

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java
```diff
@@ -455,6 +457,7 @@ void shouldStartFromBestSnapshotWhenOptionEnabled() {
             depositSnapshotFileLoader,
             depositSnapshotStorageLoader,
             true,
+            false,
```
zilm13: And here with all default values we load both db and file and return max. You'd better clarify this by adding this line after `manager2.start()`: 
```java
    verify(depositSnapshotFileLoader, never()).loadDepositSnapshot();
```
It will fail, but our goal in this PR is to avoid loading file once db is populated

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java
```diff
@@ -407,10 +409,10 @@ void shouldStartFromSnapshotFileWhenProvided() {
     manager.start();
     notifyHeadBlock(lastBlockNumber, MIN_GENESIS_BLOCK_TIMESTAMP + 1000);
```
zilm13, line 411: I'd add here
```java
    verify(depositSnapshotFileLoader, never()).loadDepositSnapshot();
```
for clarity

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
```diff
@@ -118,31 +121,34 @@ public void start() {
...
+    return depositSnapshotStorageLoader
+        .loadDepositSnapshot()
+        .thenApply(
+            storageDepositSnapshotResult -> {
+              if (storageDepositSnapshotResult.getDepositTreeSnapshot().isPresent()) {
+                final DepositTreeSnapshot storageSnapshot =
+                    storageDepositSnapshotResult.getDepositTreeSnapshot().get();
+                STATUS_LOG.onDepositSnapshot(
+                    storageSnapshot.getDepositCount(), storageSnapshot.getExecutionBlockHash());
+                if (!takeBestDepositSnapshot) {
```
zilm13: the only case when we are inside this if is when deposit snapshot is disabled and no custom path provided. We should return DB version here, it's correct, but we should also return it when we have checkpoint-sync-url

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
```diff
@@ -118,31 +121,34 @@ public void start() {
...
+                  return storageDepositSnapshotResult;
                 }
+              }
+
+              final LoadDepositSnapshotResult fileDepositSnapshotResult =
```
zilm13, line 137: so we load here even if it's checkpoint-sync file

zilm13 (commented):
So what I think
As far as we have `customDepositSnapshotPathPresent` flag now, and we introduce weak subjectivety enforcement for full resync for users who have old dbs, we don't need any comparison now. Just remove this flag, if we have something in db - it's best to go. If we don't - try loading from file. Also remove bundled snapshot from loader when it's disabled. Something like this https://github.com/zilm13/teku/commit/3b098129355621c2c6741ce082339f32d2093333
For the tests it's good to cover:
- file loader is empty (no resources added) when deposit-snapshot is disabled
- custom on
- custom off, db snapshot present
- custom off, db snapshot not present

It will be simpler, less confusing and pretty good for users.

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
```diff
@@ -118,31 +121,34 @@ public void start() {
   }
 
   private SafeFuture<LoadDepositSnapshotResult> loadDepositSnapshot() {
-    final LoadDepositSnapshotResult fileDepositSnapshotResult =
-        depositSnapshotFileLoader.loadDepositSnapshot();
-    if (fileDepositSnapshotResult.getDepositTreeSnapshot().isPresent()
-        && !takeBestDepositSnapshot) {
-      LOG.debug("Deposit snapshot from file is provided, using it");
```
zilm13, line 125: I'd leave this line in code and add that it's custom, something like 
```java
LOG.debug("Custom deposit snapshot is provided, using it");
```

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
courtneyeh: We have already done the check above for that. So there is no way it is a custom deposit snapshot path.

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
courtneyeh: We have already done the check above for that. So there is no way it is a custom deposit snapshot path.

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
```diff
@@ -118,31 +117,29 @@ public void start() {
...
+            storageDepositSnapshotResult -> {
+              if (storageDepositSnapshotResult.getDepositTreeSnapshot().isPresent()) {
+                final DepositTreeSnapshot storageSnapshot =
+                    storageDepositSnapshotResult.getDepositTreeSnapshot().get();
+                STATUS_LOG.onDepositSnapshot(
+                    storageSnapshot.getDepositCount(), storageSnapshot.getExecutionBlockHash());
+                return storageDepositSnapshotResult;
+              }
+
+              final LoadDepositSnapshotResult fileDepositSnapshotResult =
```
lucassaldanha, line 137: It would be nice to have a STATUS_LOG.loadingDepositSnapshotFromFile(filename) when we load from file.

We could even log something like: "No deposit snapshot found on local db, loading it from file x" or something? What do you think?

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java
```diff
@@ -444,7 +479,7 @@ void shouldStartFromStorageSnapshotWhenProvided() {
   }
 
   @Test
-  void shouldStartFromBestSnapshotWhenOptionEnabled() {
+  void shouldIgnoreBestSnapshotCheckWhenDbStorageNotEmpty() {
```
lucassaldanha: Does this test make sense given we removed the check for best snapshot?

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
courtneyeh: I'm not sure if we need to say no snapshot found on local db? But can add that if you aren't satisfied with what I have done

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/DepositSnapshotFileLoader.java
```diff
@@ -71,6 +71,7 @@ public LoadDepositSnapshotResult loadDepositSnapshot() {
         final DepositTreeSnapshot depositTreeSnapshot = loadFromUrl(depositSnapshotResourceUrl);
         // Validate
         DepositTree.fromSnapshot(depositTreeSnapshot);
+        STATUS_LOG.loadingDepositSnapshotFromFile(depositSnapshotResourceUrl);
```
lucassaldanha: Sorry, I just realised we already log this on line 70:
```
STATUS_LOG.loadingDepositSnapshotResource(sanitizedUrl);
```

beacon/pow/src/main/java/tech/pegasys/teku/beacon/pow/Eth1DepositManager.java
```diff
@@ -118,31 +117,28 @@ public void start() {
...
+                  depositSnapshotFileLoader.loadDepositSnapshot();
+              LOG.debug("Deposit snapshot from file is provided, using it");
```
zilm13: Maybe "DB deposit snapshot is not present, falling back to the file provided", something like that?

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java
```diff
@@ -383,31 +386,63 @@ void shouldStartWithStoredDepositsChainReorgedToBeShorter() {
...
+        .thenReturn(LoadDepositSnapshotResult.create(Optional.of(depositTreeSnapshotFromFile)));
+    when(depositProcessingController.fetchDepositsInRange(any(), any())).thenReturn(COMPLETE);
+
+    // DepositTreeSnapshot from DB empty
+    when(depositSnapshotStorageLoader.loadDepositSnapshot())
+        .thenReturn(SafeFuture.completedFuture(LoadDepositSnapshotResult.EMPTY));
+
+    manager.start();
+    notifyHeadBlock(lastBlockNumber, MIN_GENESIS_BLOCK_TIMESTAMP + 1000);
+    verify(depositSnapshotStorageLoader, Mockito.times(1)).loadDepositSnapshot();
```
zilm13: it's by default `times(1)`, extra param could be removed in both lines (this and the next one, 443)

beacon/pow/src/test/java/tech/pegasys/teku/beacon/pow/Eth1DepositManagerTest.java
```diff
@@ -460,34 +495,26 @@ void shouldStartFromBestSnapshotWhenOptionEnabled() {
...
-        .thenReturn(
-            SafeFuture.completedFuture(
-                LoadDepositSnapshotResult.create(Optional.of(depositTreeSnapshotFromDb))));
-
-    manager2.start();
-    notifyHeadBlock(BigInteger.valueOf(3000), MIN_GENESIS_BLOCK_TIMESTAMP + 1000);
+    manager.start();
+    notifyHeadBlock(lastBlockNumber, MIN_GENESIS_BLOCK_TIMESTAMP + 1000);
+    verify(depositSnapshotStorageLoader, never()).loadDepositSnapshot();
+    verify(depositSnapshotFileLoader, Mockito.times(1)).loadDepositSnapshot();
```
zilm13: same here, times parameter could be removed

zilm13 (commented):
Just few nits, but one extra change is required to ensure we will not load any file when deposit tree snapshot is disabled: https://github.com/zilm13/teku/commit/3b098129355621c2c6741ce082339f32d2093333#diff-abde22323bdabf093cbe514f5ecefef1510ba5f722d6d55e860d3e03c120ba9cR206 
\+ test in `PowChainServiceTest` to verify resources are empty with this configuration.

zilm13 (approved):
LGTM

PR #8114, opened 2024-03-20
increased the time to keep attestations cached

rolfyone:
Also added a small amount of logging at trace level, it'll likely be noisy if we have to turn it on.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8116, opened 2024-03-20
Remove attestation worthiness checker

tbenr:
Following the update for EIP-7045, simply ignore worthless attestation is not a correct strategy anymore.

We will need to be clever in how we select attestations when there is space in block. The implementation of that will look completely different from the worthiness checker, so we can start removing it.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8117, opened 2024-03-20
Fix ssz publishBlock sent without milestone header

zilm13:
## PR Description
It was reported that VC Teku fails to propose in connection with BN Lighthouse.
This happens because Teku sends requests without required header (see https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/blocks/blocks.yaml#L37 "Use content type header"). According to the API spec it should be used both for blind block and non-blind. Fixed with the PR.

Tested also (successfully) in Kurtosis with following configuration:
```
participants:
  - el_client_type: geth
    cl_type: lighthouse
    vc_type: teku
    vc_image: consensys/teku:develop
additional_services:
  - el_forkmon
  - tx_spammer
  - dora
snooper_enabled: true
```

## Documentation

## Changelog


StefanBratanov (approved):
LGTM. Can possibly add changelog for bug fixes.

rolfyone (approved):
LGTM

rolfyone:
this looks like we're still using the v1 block interface? I guess at some point this starts calling v3 and it needs that header too...

PR #8124, opened 2024-03-21
EIP-6110 Implement `beacon-chain` and `validator` logic

StefanBratanov:
## PR Description
As per https://github.com/ethereum/consensus-specs/pull/3615

- Also removed incorrect tests in BlockProcessor...Test

Main changes are in `BlockProcessorElectra` and `DepositProvider`

## Fixed Issue(s)
solves 2 of the tasks under #7898 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/DepositReceiptsUtil.java
```diff
@@ -0,0 +1,55 @@
...
+  private final SecureRandom random = new SecureRandom();
+
+  private final Spec spec;
+
+  public DepositReceiptsUtil(final Spec spec) {
+    this.spec = spec;
+  }
+
+  public DepositReceipt createDepositReceipt(final UInt64 slot, final UInt64 index) {
+    final BLSKeyPair validatorKeyPair = BLSKeyPair.random(random);
```
zilm13: It's confusing in production files. Maybe put in `DataStructureUtil`? We have most random things there including one with domain already

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectraTest.java
```diff
@@ -0,0 +1,144 @@
...
+  public void processesDepositReceipts() {
+    final BeaconStateElectra preState =
+        BeaconStateElectra.required(
+            createBeaconState()
+                .updated(
+                    mutableState ->
+                        MutableBeaconStateElectra.required(mutableState)
+                            .setDepositReceiptsStartIndex(
+                                SpecConfigElectra.UNSET_DEPOSIT_RECEIPTS_START_INDEX)));
+    final int numberOfValidators = preState.getValidators().size();
```
zilm13: maybe something like `preElectraValidators` to highlight it's pre-transition number?

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/versions/electra/helpers/MiscHelpersElectraTest.java
```diff
@@ -0,0 +1,76 @@
...
+        BeaconStateElectra.required(preState)
+            .updated(
+                mutableState -> {
+                  mutableState.setEth1DepositIndex(UInt64.valueOf(64));
+                  MutableBeaconStateElectra.required(mutableState)
+                      .setDepositReceiptsStartIndex(
+                          SpecConfigElectra.UNSET_DEPOSIT_RECEIPTS_START_INDEX);
+                });
+
+    assertThat(miscHelpersElectra.isFormerDepositMechanismDisabled(state)).isFalse();
```
zilm13, line 74: maybe add tests to previous forks mischelpers that it's false too?

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/DepositProviderTest.java
```diff
@@ -164,18 +174,47 @@ void numberOfDepositsThatCanBeIncludedMoreThanMaxDeposits() {
...
+    // 16th deposit is using the new mechanism
+    updateStateDepositReceiptsStartIndex(16);
+
+    mockDepositsFromEth1Block(0, 10);
+    mockDepositsFromEth1Block(10, 20);
+
+    final SszList<Deposit> deposits = depositProvider.getDeposits(state, randomEth1Data);
+
+    // the pending Eth1 deposits (deposit_receipt_start_index - eth1_deposit_index)
+    assertThat(deposits).hasSize(11);
```
zilm13, line 204: 5th is already in state, next 6,7,8,9,10,   11,12,13,14,15
10 total, no?
(very concerned on any possible by 1 error in a new code, btw, hope we will have some spec reference tests)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/DepositReceiptsUtil.java
StefanBratanov: The idea behind it is that this class would also be used in the `ExecutionLayerChannelStub`, which doesn't have access to test files

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/DepositProviderTest.java
StefanBratanov: `eth1_deposit_index` is a bit confusing, but refers to the next deposit index to be processed, so 5th is not in state yet.

zilm13: Got it! Could you please add a comment about it for clarification, how do we get 11

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/util/DepositReceiptsUtil.java
zilm13: Got it. Still a lot of confusion and we loose seed we usually have. So maybe 
- refactor so we could use it from dataStructureUtil when in tests
- rename class and method to highlight Test, random etc
- add a comment that it has nothing to do with spec

Actually I think that duplicating it in DataStructureUtil and in private Stub method will lead to less confusion. And there is no issue that copy-paste could add a bugs when we need to apply some changes to this code in future. It's non-production code.

StefanBratanov: yeah actually I like it, we have similar logic for former deposits, pushed a change

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpersTest.java
```diff
@@ -231,6 +231,12 @@ public void committeeComputationShouldNotOverflow(int activeValidatorsCount, int
         });
   }
 
+  @Test
+  public void isFormerDepositReceiptMechanismDisabled_returnsFalse() {
```
zilm13: Is it possible to run it parameterized for all forks which are less than electra?

zilm13 (approved):
LGTM, just one test improvement could be good if it's not complex

ethereum/spec/src/test/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpersTest.java
StefanBratanov: I like it. It was pretty easy. Done.

PR #8125, opened 2024-03-21
Fix ssz fallback in block publish was not working on 400 error

zilm13:
## PR Description
Not 100% on this. Could it fallback some wrong case with it? It will work with Lighthouse (fallback didn't work before with it) but not sure for the other cases.

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedBlockRequest.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.validator.remote.typedef.handlers;
 
+import static tech.pegasys.teku.infrastructure.http.HttpStatusCodes.SC_BAD_REQUEST;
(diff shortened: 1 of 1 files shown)


tbenr:
This seems like a workaround on our side when the server doesn't correctly handle HTTP status code in case of unsupported content type. We could potentially do this but i'd prefer to see this fixed on LH side (if I'm correct in the anaisys).
The only problem I see is that, if we do that on V2, we could fallback in case broadcast validation fails.
In V1 we could simply retry on a incorrect block, that will be probably rejected in json too.

zilm13:
It's all good with Lighthouse.
We were getting this code without required Header. SSZ was supported.
But it's a type of error that will not be recovered with the next block (correct me if I'm wrong, so it's easy "no merge" if we could get this on one block and not on another)
However with another endpoint type we have an extra chance (but we will switch to it for all subsequent proposals).
That's my thoughts.

StefanBratanov:
My concern is that `400` is a documented error code in the spec and falling back to json for subsequent requests would be wrong in case of expected errors. I think I would rather prefer code to fail and it can highlight interoperability issues. Then clients can fix the issue rather than us trying to capture every edge case of clients potential bugs.

zilm13:
Ok, let's scrap it
I had concerns on this PR too.

PR #8129, opened 2024-03-22
Make DataUnavailablePool always accept new blocks

tbenr:
We always make room for latest block, so we guarantee to retry the latest.

The scenario to fix is:
- several block with missing blobs fill the pool. None of this blocks have been imported by any node (blobs not seen by anyone)
- a new block arrives and is discarded. But his block has been imported by others and become canonical (blobs seen by magority of the validators).
- Next proposer builds on top of that. And we don't have any retry in place.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM, is 10 a sensible default?

tbenr:
> LGTM, is 10 a sensible default?

I think yes now that we have pruning and this PR. Essentially here we are saying that we can try to resolve up to 10 competing chains. If we can't progress in any we simply fall in syncing mode.

tbenr:
but it is also true that with pruning is safe to rise it... 🤔

PR #8135, opened 2024-03-22
Add error log on non-completed trackers

tbenr:
In some flows we supply block and blobs to tracker, and it is supposed to be immediately completed.

I added an error log if it isn't the case.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8137, opened 2024-03-24
Update endpoints tags, operationId and summary according to the spec

zilm13:
## PR Description

Updated several fields for all Beacon API endpoints to reduce the gap between our implementation and the spec. All `tags`, `operationId` and `summaries` were verified to match the spec.

**Tags:**
- `getEpochSyncCommittees` - Validator required tag removed [like in the spec](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/states/sync_committees.yaml#L5-L6) and I don't see validator API is calling this.
- `registerValidator` - Validator required tag removed [like in the spec](https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/register_validator.yaml#L15-L16), I think the message is that `mev-builder` is not required to propose blocks :)

**OperationId:**
All was modified to be the same as in the spec, so the correct endpoint could be easily found when changes are needed.

**Summmary:**
Updated most by taking summaries from the spec because it's usually more descriptive, skipped only few, like blinded block where it's less descriptive in the spec.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM. Nice to have it aligned with the spec. Good job!

PR #8138, opened 2024-03-24
Add headers to Endpoint's Metadata where applicable

### LINKED ISSUE #8127: Add header and headerRequired to all EndpointMetadata's where applicable
### Description
Some endpoints have headers which is currently not included in metadata. For example, `publishBlockV1`, `publishBlockV2` and its blinded versions. It worth adding it especially in the case of `required`, so it appears in generated handler specifications and could be tested with integration tests.

zilm13:
## PR Description
Actually it's used only in block publishing APIs

## Fixed Issue(s)

fixes #8127 

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8139, opened 2024-03-25
Ability to add deposit receipts in the stub

StefanBratanov:
## PR Description
Configured by a static constant `GENERATE_DEPOSIT_RECEIPTS`. Defaulting to `false` since we don't want random deposit receipts in the stub and end up in an inactivity leak.

Also changed `engineGetPayload` to pass a state instead of slot. This could be useful in the future as well if we need more information.

## Fixed Issue(s)
related to #7898 

## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #8140, opened 2024-03-25
Express `seen_ttl` in consensus params

StefanBratanov:
## PR Description
As per https://github.com/ethereum/consensus-specs/pull/3627

I think more consistent to set with consensus params values to align with networks like `Gnosis`, etc rather than the hardcoded value we have at the moment:

```
static final Duration DEFAULT_HEARTBEAT_INTERVAL = Duration.ofMillis(700);
static final Duration DEFAULT_SEEN_TTL = DEFAULT_HEARTBEAT_INTERVAL.multipliedBy(1115);
```

## Fixed Issue(s)
N/A

## Documentation

## Changelog


Nashatyrev (approved):
LGTM

PR #8141, opened 2024-03-25
EIP-6110 Stop Eth1 data polling

StefanBratanov:
## PR Description
If we have already transitioned to the new deposit mechanism:

- Don't log missing deposit data 
- Don't do actual `Eth1Data` vote, just return the one from the state
- Don't update `Eth1Data` metrics
- Don't initialize and start `PowchainService` services
- Implement stop for `AsyncRunLoop` because `TimeBasedEth1HeadTracker` was running in loop and `stop` was not implemented

Also attached `FinalizedCheckpointChannel` to `PowchainService` to stop Eth1 polling when transitioned on the fly. I am not 100 % sure of the cleanest way to pass a state to `PowchainService` because `RecentChainData` is initialized after starting of the services, so passing it before that will be in null. That's why chose to pass a `Supplier` (along with some comments) but happy for suggestions.

**Update:**
Refactored `PowchainService` to initialize when starting. This has couple of benefits. No initialization needed when we have already finalized the transition to the new deposit mechanism. This allows thing like using the websockets web3j client after Electra instead of relying on http, so no more of this error:

```java
if (executionWeb3jClientProvider.getWeb3JClient().isWebsocketsClient()) {
   throw new InvalidConfigurationException("Eth1 endpoint fallback is not compatible with Websockets execution engine endpoint");
 }
```
It also allows to log the eth1 polling disabling only in case it has been enabled before.

## Fixed Issue(s)
will solve one of the tasks under #7898 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/DepositProviderTest.java
+++ b/beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/DepositProviderTest.java
@@ -306,6 +306,21 @@ public class DepositProviderTest {
   }
 
   @Test
+  void
+      shouldNotLogAnEventOnSlotIfFormerDepositMechanismIsDisabled_EvenIfAllDepositsRequiredForStateNotAvailable() {
+    setup(1, SpecMilestone.ELECTRA);
+    mockDepositsFromEth1Block(0, 8);
+    updateStateEth1DepositIndex(5);
+    updateStateDepositReceiptsStartIndex(5);
+    updateStateEth1DataDepositCount(10);
+    when(recentChainData.getBestState()).thenReturn(Optional.of(SafeFuture.completedFuture(state)));
+
+    depositProvider.onSlot(UInt64.ONE);
+
+    verifyNoInteractions(eventLogger);
+  }
+
+  @Test
   void shouldNotLogAnEventOnSlotWhenAllDepositsRequiredForStateAvailable() {
     setup(1);
     // To generate a valid proof we need the deposits up to state deposit count
--- a/infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/runloop/AsyncRunLoopTest.java
+++ b/infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/runloop/AsyncRunLoopTest.java
@@ -103,11 +103,11 @@ class AsyncRunLoopTest {
     verify(logic, never()).advance();
 
     advanceTime(firstAdvanceDelay);
-    verify(logic, times(1)).advance();
+    verify(logic).advance();
 
     // Shouldn't use same delay for second advance so this won't wait long enough
     advanceTime(firstAdvanceDelay);
-    verify(logic, times(1)).advance();
+    verify(logic).advance();
 
     // Advance remaining time until second advance is due and confirm it is invoked
     advanceTime(secondAdvanceDelay.minus(firstAdvanceDelay));
@@ -128,13 +128,13 @@ class AsyncRunLoopTest {
     verify(logic, never()).advance();
 
     advanceTime(advanceDelay);
-    verify(logic, times(1)).advance();
+    verify(logic).advance();
 
     runLoop.stop();
     advanceTime(advanceDelay);
 
     // advance should have been invoked only once
-    verify(logic, times(1)).advance();
+    verify(logic).advance();
   }
 
   @Test
--- a/infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/runloop/AsyncRunLoopTest.java
+++ b/infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/runloop/AsyncRunLoopTest.java
@@ -118,6 +118,26 @@ class AsyncRunLoopTest {
   }
 
   @Test
+  void shouldNotAdvanceIfStopped() {
+    final Duration advanceDelay = Duration.ofSeconds(2);
+    when(logic.advance()).thenReturn(SafeFuture.COMPLETE);
+    when(logic.getDelayUntilNextAdvance()).thenReturn(advanceDelay);
+
+    runLoop.start();
+    verify(logic).init();
+    verify(logic, never()).advance();
+
+    advanceTime(advanceDelay);
+    verify(logic, times(1)).advance();
+
+    runLoop.stop();
+    advanceTime(advanceDelay);
+
+    // advance should have been invoked only once
+    verify(logic, times(1)).advance();
+  }
+
+  @Test
   void shouldRetryAfterDelayWhenAdvanceFails() {
     final Duration advanceDelay = Duration.ofSeconds(5);
     final Throwable error = new RuntimeException("Computer says no");
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -134,8 +134,8 @@ public class StatusLogger {
         batchSize);
   }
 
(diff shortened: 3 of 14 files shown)


teku/src/main/java/tech/pegasys/teku/services/BeaconNodeServiceController.java
```diff
@@ -44,18 +48,32 @@ public BeaconNodeServiceController(
...
+    services.add(beaconChainService);
     services.add(
         new NatService(
             tekuConfig.natConfiguration(),
             tekuConfig.network().getListenPort(),
             tekuConfig.discovery().isDiscoveryEnabled()));
-    powchainService(tekuConfig, serviceConfig, maybeExecutionWeb3jClientProvider)
+    // making it a Supplier ensures that BeaconChainService has been started and RecentChainData has
+    // been initialized when `get()` is called
+    final Supplier<Optional<BeaconState>> latestFinalizedState =
```
zilm13: I'd call it `finalizedStateSupplier` here and all the way to highlight that it's dynamic, not some one-time snapshot

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -529,13 +515,6 @@ public void loadedDepositSnapshot(final long deposits, final Bytes32 executionBl
         executionBlockHash);
   }
 
-  public void warnFlagDeprecation(final String oldFlag, final String newFlag) {
```
zilm13: I guess we use this from time to time, no need to prune

infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/runloop/AsyncRunLoopTest.java
```diff
@@ -117,6 +117,26 @@ void shouldScheduleNextAdvanceWhenFirstAdvanceCompletes() {
...
+    final Duration advanceDelay = Duration.ofSeconds(2);
+    when(logic.advance()).thenReturn(SafeFuture.COMPLETE);
+    when(logic.getDelayUntilNextAdvance()).thenReturn(advanceDelay);
+
+    runLoop.start();
+    verify(logic).init();
+    verify(logic, never()).advance();
+
+    advanceTime(advanceDelay);
+    verify(logic, times(1)).advance();
```
zilm13: `times(1)` is default, not needed

infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/runloop/AsyncRunLoopTest.java
```diff
@@ -117,6 +117,26 @@ void shouldScheduleNextAdvanceWhenFirstAdvanceCompletes() {
...
+
+    runLoop.stop();
+    advanceTime(advanceDelay);
+
+    // advance should have been invoked only once
+    verify(logic, times(1)).advance();
```
zilm13: `times(1)` is default, not needed

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/PowchainServiceTest.java
```diff
@@ -39,11 +40,13 @@
 import tech.pegasys.teku.service.serviceutils.ServiceConfig;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.TestSpecFactory;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
 import tech.pegasys.teku.storage.api.Eth1DepositStorageChannel;
 
 public class PowchainServiceTest {
```
zilm13, line 47: I'd test something new here. At least that we start preElectra and don't start postElectra

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/DepositProviderTest.java
```diff
@@ -305,6 +305,21 @@ void shouldLogAnEventOnSlotWhenAllDepositsRequiredForStateNotAvailable() {
     verify(eventLogger).eth1DepositDataNotAvailable(UInt64.valueOf(9), UInt64.valueOf(10));
   }
 
+  @Test
```
zilm13, line 308: I'd test also that getDeposits() is empty since former mechanism is disabled

zilm13 (commented):
Do I understand it correctly that since EIP-6110 we don't need DepositTree and snapshots any more? 
Though I think we need to approve it I still have some concerns on transition moment, that we are not breaking something too early when we are in Electra but former mechanism is not yet disabled. Anyway, we will run testnets for this.

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/DepositProviderTest.java
StefanBratanov: This was part of a previous PR, can have a look at `noDepositsIncludedIfFormerDepositMechanismHasBeenDisabled` and `getsRemainingEth1PendingDepositsIfElectraIsEnabled`

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/PowchainServiceTest.java
StefanBratanov: Good point actually. Added a couple of relevant tests.

StefanBratanov:
> Do I understand it correctly that since EIP-6110 we don't need DepositTree and snapshots any more? Though I think we need to approve it I still have some concerns on transition moment, that we are not breaking something too early when we are in Electra but former mechanism is not yet disabled. Anyway, we will run testnets for this.

Yes, once we have finalized state where `eth1_deposit_index == deposit_receipts_start_index` we no longer need the DepositTree and snapshots since the `getDeposits` method would always be empty list from that point onwards, so no merkle proofs needed. So the idea here is to stop Eth1 polling once we finalize such a state and also don't even initialize if the anchor point on startup is such a state.

Also when syncing we don't need the deposit tree, so historical sync will still work. Essentially we only need the proofs when proposing in order to include it in the block.

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/DepositProviderTest.java
zilm13: ahhh the one I wanted comment for
All good, all done already

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/PowchainServiceTest.java
```diff
@@ -216,6 +218,54 @@ public void shouldHaveNoDepositSnapshotPathWhenNoneIsAvailable() {
...
+    final BeaconState finalizedState = mock(BeaconState.class);
+    when(spec.isFormerDepositMechanismDisabled(finalizedState)).thenReturn(true);
+
+    latestFinalizedState = () -> Optional.of(finalizedState);
+
+    final PowchainService powchainService =
+        new PowchainService(
+            serviceConfig, powConfig, Optional.of(engineWeb3jClientProvider), latestFinalizedState);
+
+    powchainService.start().join();
```
zilm13, line 253: I'd add `assertThat(powchainService.isRunning()).isTrue();` so it's clear that it's stopped only after finalized checkpoint. I know that it's clear from Service code, just the test will be more clear I think

zilm13 (approved):
LGTM, just final nit

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/PowchainServiceTest.java
StefanBratanov: Makes sense. It's more clear I agree.

PR #8155, opened 2024-03-27
Refactor `StorageUpdateChannel`

StefanBratanov:
## PR Description
`onFinalizedState` is not triggered anywhere, so having in the channel is wrong. `onStorageUpdate` can be used instead. So removing this method along with `storeFinalizedState`. Changed tests to use `storeReconstructedFinalizedState` which is an used method.

Also remove unused `removeBlobSidecars`

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8157, opened 2024-03-28
Get the rid of `DataUnavailableBlockPool`

tbenr:
TLDR;
`DataUnavailableBlockPool` was used to ensure we retry only on blocks that have been tried to be imported once and failed due to DA. We can reach the same behaviour by attaching the import to the tracker's completion future.

Benefits:
- no delay, no polling. Block import retry starts immediately.
- no additional pool, with arbitrary capacity and dedicated pruning, so we support retry as long as the tracker is still in the pool.

no really drawbacks:
- we don't control how many blocks are reimported in parallel (I honestly don't expect trackers to resolve concurrently all together, even if that the case we can handle it)
- it is a one shot retry. I think it is ok because once the tracker is completed it can't produce a `FAILED_DATA_AVAILABILITY_CHECK_NOT_AVAILABLE` import failure anymore. If the blobs are complete but invalid, block import will fail with `FAILED_DATA_AVAILABILITY_CHECK_INVALID` and we currently drop the tracker completely (if someone next proposer build on top of it, we will retrieve all via RPC from scratch)

A recap:
When we are following the chain, for maximum parallelization we start a block import as soon as we see a block. Data is supposed to complete while doing the import (note: it could be completed earlier if the block was seen after all blobs).

Since we are not tracking which are the blocks being imported, we can't blindly attach a blockImport upon tracker compilation, since we could then trigger an import while it is being imported already.
So we can do that on-demand when import fails due to data unavailability.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM. Nice improvement.

PR #8164, opened 2024-04-02
EIP7002: Implemented process_execution_layer_exit

lucassaldanha:
The "meat" of the logic for processing execution layer exits can be found on `BlockProcessorElectra#processExecutionLayerExits(..)`. This is the method exposed through the `BlockProcessor` interface, so it can be used for reference tests.

Reference to spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/eip7002/beacon-chain.md#new-process_execution_layer_exit

The method used during block processing is a helper method `AbstractBlockProcessor#processExecutionLayerExits(..)`, used as a way to "lock" the order that the operation needs to be processed. This method is overridden only on `BlockProcessorElectra`.


StefanBratanov (commented):
I can suggest a cleaner approach where we don't need to repeat everything (it's build on top of the commits PR): https://github.com/StefanBratanov/teku/commit/aa2019aed4deb59672cfc6a7486cb9fbe292ef9c

Essentially, just define `processExecutionLayerExits` in the `AbstractBlockProcessor` and override it in Electra.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
```diff
@@ -872,6 +875,18 @@ protected BlockValidationResult verifyVoluntaryExits(
...
   }
 
+  protected void processExecutionPayloadExits(
+      final MutableBeaconState state, final Optional<ExecutionPayload> executionPayload)
+      throws BlockProcessingException {
+    // No ExecutionLayer exits until Electra
+  }
+
+  @Override
+  public void processExecutionLayerExits(
```
lucassaldanha, line 888: This is exposed through the interface so we can use it for reference tests.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
```diff
@@ -440,6 +442,7 @@ protected void processOperationsNoValidation(
           processDeposits(state, body.getDeposits());
           processVoluntaryExitsNoValidation(
               state, body.getVoluntaryExits(), validatorExitContextSupplier);
+          processExecutionPayloadExits(state, body.getOptionalExecutionPayload());
```
StefanBratanov: Think we should pass the `validatorExitContextSupplier` similar to the other methods here because it's expensive operation and the Supplier is `memoized`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/AbstractBlockProcessor.java
```diff
@@ -872,6 +875,18 @@ protected BlockValidationResult verifyVoluntaryExits(
     return BlockValidationResult.SUCCESSFUL;
   }
 
+  protected void processExecutionPayloadExits(
```
StefanBratanov: I would rename to `processExecutionLayerExits` for consistency

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -119,6 +127,97 @@ protected void verifyOutstandingDepositsAreProcessed(
...
+  public void processExecutionLayerExits(
+      final MutableBeaconState state, final SszList<ExecutionLayerExit> exits) {
+    final Supplier<ValidatorExitContext> validatorExitContextSupplier =
+        beaconStateMutators.createValidatorExitContextSupplier(state);
+
+    final UInt64 currentEpoch = miscHelpers.computeEpochAtSlot(state.getSlot());
+
+    exits.forEach(
+        exit -> {
+          final OptionalInt maybeValidatorIndex =
```
StefanBratanov: We should use `validatorsUtil.getValidatorIndex(state, exit.getValidatorPublicKey())` to make use of our cache. It is expensive to loop.

StefanBratanov (commented):
Logic looks good, but added couple of comments about performance improvements

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
lucassaldanha: Ah completely forgot we had that! Thanks!

StefanBratanov (approved):
LGTM

PR #8165, opened 2024-04-02
Only cache finalized indices in (pubkey, index) cache

### LINKED ISSUE #7898: [EIP-6110] Supply validator deposits on chain

StefanBratanov:
## PR Description
This PR changes the `ValidatorIndexCache` such that only finalized validator indices are cached. This is accomplished by creating a `ValidatorIndexCacheTracker` which is subscribed to finalized checkpoint events which in turn updates the finalized index of the `ValidatorIndexCache`. This is the assumption as per https://eips.ethereum.org/EIPS/eip-6110#validator-index-invariant .

So from this PR onwards, when looking for an index for a public key, the steps are:

1. Look for an index in the cache
2. Scan within the finalized validator indices for an index and cache whilst going. If found, return it. Next iterations will start from the last cached one.
3. If not found withing the finalized validator indices, search within the non finalized ones, but do not cache.

Also on store initialization (either from initial state or disk), the cache finalized index is populated. That is required because `onNewFinalizedCheckpoint` is triggered only on finalizations after the store is initialized.

**Things to consider:**

- Performance issues when scanning for a lot of non finalized pub keys. How often does that happen? How slow it could be?
- Maybe could make the logic fork specific or do we need to? It could become more complex.

## Fixed Issue(s)
fixes #7898

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorIndexCacheTracker.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorIndexCacheTracker.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright Consensys Software Inc., 2024
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.validator.coordinator;
+
+import tech.pegasys.teku.spec.datastructures.state.Checkpoint;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconStateCache;
+import tech.pegasys.teku.storage.api.FinalizedCheckpointChannel;
+import tech.pegasys.teku.storage.client.RecentChainData;
+
+public class ValidatorIndexCacheTracker implements FinalizedCheckpointChannel {
+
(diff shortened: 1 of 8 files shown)


rolfyone:
Do we have performance tests? it might really hurt to iterate the state all the time, I'm not sure it's what we really want to do...
I think we might actually need to do something a little more complicated, somehow keeping a small non final cache so that we can look that up and not continually iterate...
Either that or start using this new cache down the line, not in the next release... for now avoiding a lot of state lookups is highly desirable.

StefanBratanov:
> Do we have performance tests? it might really hurt to iterate the state all the time, I'm not sure it's what we really want to do... I think we might actually need to do something a little more complicated, somehow keeping a small non final cache so that we can look that up and not continually iterate... Either that or start using this new cache down the line, not in the next release... for now avoiding a lot of state lookups is highly desirable.

In short, current implementation works same as before, keeps track of what is the last cached index and only iterate from there and cache on the way all the way to the last finalized index. Then it scans all the non-finalized indices, which shouldn't be a lot and it should happen infrequently that we query for those. Testing on Holesky currently.

rolfyone (approved):

PR #8166, opened 2024-04-03
Allow override of ref tests root

lucassaldanha:
This change was made to support running ref tests using custom built tests from consensus-specs.


tbenr (approved):
LGTM

Issue #8158, opened 2024-03-28
voluntary exits 'no peers on channel'


rolfyone:
When submitting a voluntary exit in deneb, 'no peers on topic' has been reported.

We need to investigate this, the Deneb spec says `The voluntary_exit topic is implicitly modified due to the lock-in use of CAPELLA_FORK_VERSION for this message signature validation for EIP-7044.`

This possibly means that the voluntary exit digest should be locked at the capella digest, so we're essentially on the wrong gossip topic to find peers for exits.

lucassaldanha:
I haven't got confirmation from the other client teams if we should fix the fork digest, but I checked a Lighthouse node in Sepolia and all topics are using an updated fork digest after Deneb, and all topics match the topics that our Teku nodes in Sepolia are subscribed to.

So either we do not need to fix the fork digest for the `voluntary_exit` topic, or both Teku and LH are doing something wrong :)

zilm13:
### Spec:


The voluntary_exit topic is implicitly modified due to the lock-in use of CAPELLA_FORK_VERSION for this message signature validation for EIP-7044. [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#topics-and-messages)

> I read it that topic is Capella fixed.

```
The specific makes one change to the state transition function:

    Modify `process_voluntary_exit` to compute the signing domain and root fixed on `CAPELLA_FORK_VERSION`.

Additionally, the `voluntary_exit` gossip conditions are implicitly modified to support this change.
```
[EIP-7044](https://eips.ethereum.org/EIPS/eip-7044), which is the part of Deneb
> This could be read in two ways: either topic is Capella-fixed or topic is changed according to the fork, but validation should not reject Capella-signed voluntary exits. 

### Our code:
[Validation lock-in to Capella](https://github.com/Consensys/teku/blob/78a929fa0bbe76eaac8d2bdd3dcea9d9125db28f/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/BeaconStateAccessorsDeneb.java#L43-L52)

[...]

PR #8168, opened 2024-04-04
Allow voluntary exit command to write to file rather than publish

rolfyone:
The advantage of this is that the message can avoid epoch validation, and also it can separate the submission of the exit from creation, so if something fails, it can have further investigation.

The format saved to json will allow it to be posted directly to the voluntary exits beacon-api POST endpoint.

Assists with #8158 but only as a potential workaround, allowing us to avoid setting up the keymanager-api to generate an exit.

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -180,6 +206,30 @@ public Integer call() {
...
+              "%s exists and is not a directory, cannot export to this path.",
+              voluntaryExitsFolder));
+      return false;
+    } else if (!voluntaryExitsFolder.exists()) {
+      voluntaryExitsFolder.mkdirs();
+    }
+    final AtomicInteger failures = new AtomicInteger();
+    getValidatorIndices(validatorsMap)
+        .forEach(
+            (k, v) -> {
```
tbenr: I'd call this publicKey and validatorIndex for readability

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -261,6 +305,43 @@ private void submitExitForValidator(final BLSPublicKey publicKey, final int vali
...
+        Optional.ofNullable(validatorsMap.get(publicKey))
+            .orElseThrow()
+            .getSigner()
+            .signVoluntaryExit(message, forkInfo)
+            .join();
+    return new tech.pegasys.teku.spec.datastructures.operations.SignedVoluntaryExit(
+        message, signature);
+  }
+
+  private boolean storeExitForValidator(BLSPublicKey blsPublicKey, Integer validatorIndex) {
```
tbenr: finals

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -261,6 +305,43 @@ private void submitExitForValidator(final BLSPublicKey publicKey, final int vali
...
+          voluntaryExitsFolder.toPath().resolve(blsPublicKey.toAbbreviatedString() + "_exit.json"),
+          prettyExitMessage(exit));
+      return true;
+    } catch (IOException e) {
+      SUB_COMMAND_LOG.error("Failed to store exit for " + blsPublicKey.toAbbreviatedString());
+      return false;
+    }
+  }
+
+  private String prettyExitMessage(
```
tbenr, line 339: to be an actual prettyPrint you should use something similar to:

https://github.com/ConsenSys/teku/blob/8c172a2eaff954a73a5f665828bf17bdf04c1b6f/teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/PrettyPrintCommand.java#L109

essentially calling `gen.useDefaultPrettyPrinter()`

tbenr (approved):
Just some nits and a prettyPrint suggestion

lucassaldanha (approved):
LGTM. Just need a follow-up docs PR.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
rolfyone: its so annoying generated interfaces miss that :/ i got most :/

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
rolfyone: fair, i was probably lazy about this a little.

PR #8169, opened 2024-04-04
Improve logging

tbenr:
Set log level to `trace` for block import failures for which user can't do much.

Add reason to Late Block Import (to avoid confusion about late blocks that actually fail at the end)

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
@@ -280,8 +280,8 @@ public class EventLogger {
       final String result) {
     final String slowBlockLog =
         String.format(
-            "Late Block Import *** Block: %s proposer %s - %s - result: %s",
-            LogFormatter.formatBlock(slot, root), proposer, timings, result);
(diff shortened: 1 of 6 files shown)


infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/EventLogger.java
```diff
@@ -273,11 +273,15 @@ public void terminalPowBlockTtdEta(final UInt256 ttd, final Duration eta, final
...
+      final Bytes32 root,
+      final UInt64 slot,
+      final UInt64 proposer,
+      final String timings,
+      final String result) {
+    final String slowBlockLog =
         String.format(
-            "Late Block Import *** Block: %s proposer %s %s",
-            LogFormatter.formatBlock(slot, root), proposer, timings);
+            "Late Block Import *** Block: %s proposer %s - %s - result: %s",
```
StefanBratanov: Just a bit of a nit but think  "Late Block Import *** Block: %s Proposer: %s Result: %s %s" is cleaner and more consistent

And swap result and timings

StefanBratanov (approved):
LGTM apart from one nit

PR #8175, opened 2024-04-05
Skip EL part of graffiti when EL info not available

zilm13:
## PR Description
We are currently filling it with something like `NA0000`, which is not beautiful, so we decided to skip it at all in this case.
Also behavior was updated to push UNKNOWN version in channel when it's not known due to failure or whatever, but do it only once per client start. So we could have normal log line with expected graffiti on start when execution client is not supporting the new API and it will be not noisy if there are issues with EL.

PS: I don't try to put more CL info in graffiti when EL version is unknown to make it simpler.

## Documentation

## Changelog


StefanBratanov:
Isn't it better if the length requirements change if EL information is not available because less bytes are needed?

StefanBratanov:
Also I think better to use null and remove UNKNOWN in some improbable case where some EL sends us same as UNKNOWN 😁

StefanBratanov (approved):
LGTM

PR #8177, opened 2024-04-07
Dump p2p on files instead of hex logs

### LINKED ISSUE #7133: Dump p2p on files instead of hex logs
We could improve the way we logs via `--Xlog-include-p2p-warnings-enabled` and implement a dedicated SSZ dumper that can write dumps in a dedicated sub dir in our data path.
We can use this dumper where we need it. the dumper should log that a dump happened and which file has been written.
This we we can reduce the log noise and be more friendly with Logz\Loki

courtneyeh:
fixes #7133 

## Documentation

## Changelog


infrastructure/logging/build.gradle
```diff
@@ -6,6 +6,8 @@ dependencies {
     implementation 'org.apache.logging.log4j:log4j-core'
     compileOnly 'com.github.oshi:oshi-core'
 
+    testImplementation testFixtures(project(':ethereum:spec'))
```
rolfyone: this is definitely an indication we're not in the right place... we're likely to be doing cyclic deps here if we're not very careful

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class P2PDumpManager {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private static final String GOSSIP_DECODING_ERROR_DIR = "gossip_decoding_error_messages";
+  private static final String GOSSIP_REJECTED_DIR = "rejected_gossip_messages";
+  private static final String INVALID_BLOCK_DIR = "invalid_blocks";
+
+  public static Optional<String> saveGossipMessageDecodingError(
+      final Path directory,
```
lucassaldanha: Why do we need this directory as part of this method? Can't we have it as part of the configuration of the `P2PDumpManager`?

Also I do not understand why these methods need to be static.

lucassaldanha: One more question re this method. Why do we need the Optional<String> return?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,106 @@
...
+      final String errorDirectory,
+      final String fileName,
+      final String identifiers,
+      final Bytes bytes) {
+    // Check directory exists
+    final Path directory = baseDirectory.resolve(errorDirectory);
+    if (directory.toFile().mkdir()) {
+      LOG.info(String.format("%s directory has been created to save %s.", errorDirectory, object));
+    }
+    final Path path = directory.resolve(fileName);
```
lucassaldanha: This is going to check if the directory exist every single time, this seems inefficient and we could only check it like 3 times in the worst scenario (1 for each subdirectory or something like that). Imagine getting thousands of invalid voluntary_exits in a topic, we would be doing thousand of directory exits checks.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,108 @@
...
+        throw new FileAlreadyExistsException(errorMessage);
+      }
+      final Path writtenPath = Files.write(path, bytes.toArray());
+      return writtenPath.toString();
+
+    } catch (IOException e) {
+      final String errorMessage =
+          String.format("Failed to save %s bytes to file. %s", object, identifiers);
+      LOG.error(errorMessage, e);
+      return "ERROR saving to file";
```
courtneyeh: Is this okay? Or should I be logging out the bytes if saving to file has failed? @lucassaldanha

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,108 @@
...
+
+  public String saveGossipRejectedMessageToFile(
+      final String topic, final String arrivalTimestamp, final Bytes decodedMessage) {
+    final String fileName = String.format("%s_%s.ssz", arrivalTimestamp, topic);
+    final String identifiers = String.format("Topic: %s", topic);
+    return saveBytesToFile(
+        "rejected gossip message", GOSSIP_REJECTED_DIR, fileName, identifiers, decodedMessage);
+  }
+
+  public String saveInvalidBlockToFile(
```
lucassaldanha: When I said the Optional<String> didn't make sense, I meant that this method should not care about returning anything (fire and forget). Because nothing that happens in this method should influence other business logic, this is a cross-cutting concern.

The same applies to the other methods.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,108 @@
...
+      final String fileName,
+      final String identifiers,
+      final Bytes bytes) {
+    final Path path = directory.resolve(errorDirectory).resolve(fileName);
+    try {
+      // Create file and save ssz
+      if (!path.toFile().createNewFile()) {
+        final String errorMessage =
+            String.format("Unable to create new file to save %s. %s", object, identifiers);
+        throw new FileAlreadyExistsException(errorMessage);
```
lucassaldanha: We don't even want to throw an exception from here without handling it. Imagine when we call this from `ForkChoice`, and then if this exception is thrown, we have no control on how this is going to impact the flow of the code.

Whatever happens, we want to make sure it gets handled within `P2PDumpManager` and does not impact the "outside world".

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
lucassaldanha: Maybe logging the filename will give an idea of what we have missed. But I don't think we want to add the whole file content to the log.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,108 @@
...
+
+  private static final String GOSSIP_DECODING_ERROR_DIR = "gossip_decoding_error_messages";
+  private static final String GOSSIP_REJECTED_DIR = "rejected_gossip_messages";
+  private static final String INVALID_BLOCK_DIR = "invalid_blocks";
+
+  private final Path directory;
+
+  public P2PDumpManager(final Path directory) {
+    this.directory = directory;
+    if (this.directory.resolve(GOSSIP_DECODING_ERROR_DIR).toFile().mkdir()) {
```
lucassaldanha: We should use `mkdirs()` so it will also handle any intermediate folders that do not exist.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,108 @@
...
+    if (this.directory.resolve(INVALID_BLOCK_DIR).toFile().mkdir()) {
+      LOG.info(
+          String.format(
+              "%s directory has been created to save invalid blocks.", INVALID_BLOCK_DIR));
+    }
+  }
+
+  public String saveGossipMessageDecodingError(
+      final String topic, final String arrivalTimestamp, final Bytes originalMessage) {
+    final String fileName = String.format("%s_%s.ssz", arrivalTimestamp, topic);
```
lucassaldanha: Suggestion: I think storing the gossip messages in subdirectories per topic can make it easier to use and find the messages we are interested in. Thoughts?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+
+public class P2PDumpManager {
```
lucassaldanha: I think that we should use a name less tied to P2P. Mostly because this feature can potentially be used for other data structures that we can dump into file that are not necessarily related to p2p (e.g. if we wanted to dump the BeaconState whenever some error happens, etc).

Maybe something a bit more generic like `DebugDataDumper` would work best?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -150,6 +155,7 @@ public ForkChoice(
       final BlobSidecarManager blobSidecarManager,
       final ForkChoiceNotifier forkChoiceNotifier,
       final MergeTransitionBlockValidator transitionBlockValidator,
```
lucassaldanha: We likely do not need to have a dumper here, as this is the method used in multiple tests. My suggestion would be to have a `NoOpDebugDataDumper` that does nothing, and can be passed to `this(..)`.

This way, you only need to pass a "real" dumper on the constructor that is used in the production code, when we initialize ForkChoice.

(another benefit is that it won't require you to change that many tests adding the mocks)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,135 @@
...
+  public void saveGossipMessageDecodingError(
+      final String topic, final String arrivalTimestamp, final Bytes originalMessage) {
+    if (!enabled) {
+      return;
+    }
+    final String fileName = String.format("%s_%s.ssz", arrivalTimestamp, topic);
+    final String identifiers = String.format("Topic: %s", topic);
+    final Path topicPath =
+        Path.of(GOSSIP_MESSAGES_DIR).resolve(DECODING_ERROR_SUB_DIR).resolve(topic);
+    checkTopicDirExists(topicPath);
```
lucassaldanha: This is still a problem that we check if this directory exits for every single invalid message in the topic (it can happen 1000s of times in a short period).

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/P2PDumpManagerTest.java
```diff
@@ -0,0 +1,142 @@
...
+import java.nio.file.Path;
+import org.apache.tuweni.bytes.Bytes;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import tech.pegasys.teku.infrastructure.time.StubTimeProvider;
+import tech.pegasys.teku.spec.TestSpecFactory;
+import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+
+class P2PDumpManagerTest {
```
lucassaldanha: We are missing a test to prove that we do not let any exception or error escape when we fail to write the file to disk.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,135 @@
...
+  public P2PDumpManager(final Path directory, final boolean enabled) {
+    this.enabled = enabled;
+    this.directory = directory;
+    if (!enabled) {
+      return;
+    }
+
+    final Path gossipMessagesPath = this.directory.resolve(GOSSIP_MESSAGES_DIR);
+    if (gossipMessagesPath.toFile().mkdirs()) {
+      LOG.info("{} directory has been created to save gossip messages.", GOSSIP_MESSAGES_DIR);
```
lucassaldanha: IMHO, these messages should be warn level because they might indicate some problems with permissions on the folder.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+    if (gossipMessagesPath.resolve(REJECTED_SUB_DIR).toFile().mkdirs()) {
+      LOG.info(
+          "{}} directory has been created to save rejected gossip messages.", REJECTED_SUB_DIR);
+    }
+    if (this.directory.resolve(INVALID_BLOCK_DIR).toFile().mkdirs()) {
+      LOG.info(
+          String.format(
+              "%s directory has been created to save invalid blocks.", INVALID_BLOCK_DIR));
+    }
```
lucassaldanha: If we fail to create these subdirectories, we are gonna keep trying to write the messages to them later. I think we should think of a circuit breaker mechanism to ensure we don't waste time trying to write to a directory that we know doesn't exit.

lucassaldanha (commented):
We need tests for the scenario when an invalid block or an invalid gossip message happens, we dump (or do not dump) the data depending on the feature being enabled, etc.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
rolfyone: id be probably making this kind of message here just debug output... as a general rule information leak isn't great anyway

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
rolfyone: all of these creates i'd be making consistent, and also i'd be using log.debug.
consistency in terms of always using log.info, we dont need to have the last one using string.format for no reason...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private void saveBytesToFile(
+      final String object,
+      final String identifiers,
+      final Path relativeFilePath,
+      final Bytes bytes) {
+    final Path path = directory.resolve(relativeFilePath);
+    try {
+      Files.write(path, bytes.toArray());
+      LOG.info("Saved {} bytes to file. {}. Location: {}", object, identifiers, relativeFilePath);
```
rolfyone: this function probably not really needed, the formatting of the identifiers to get this far purely for logging is a little bit of an anti-pattern and Im not sure i see an advantage...

rolfyone: either that or potentially the identifiers string is just redundant or something but its a bit of a curious thing to be string formatting for a potential log that may not be even needed.

teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
```diff
@@ -284,6 +284,16 @@ public class P2POptions {
       hidden = true)
   private int batchVerifyMaxBatchSize = P2PConfig.DEFAULT_BATCH_VERIFY_MAX_BATCH_SIZE;
 
+  @Option(
+      names = {"--Xp2p-dumps-to-file-enabled"},
+      paramLabel = "<BOOLEAN>",
+      showDefaultValue = Visibility.ALWAYS,
+      description = "Enables saving P2P dumps to file.",
```
rolfyone: Save objects to file that cause problems when processing, for example rejected blocks or invalid gossip.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/P2PDumpManager.java
courtneyeh: Okay I will remove the need for the identifiers because the path is quite descriptive anyway.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/DebugDataDumperTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+    final String fileName =
+        String.format(
+            "slot%s_root%s.ssz", block.getSlot(), block.getRoot().toUnprefixedHexString());
+    final Path expectedFile = tempDir.resolve("invalid_blocks").resolve(fileName);
+    checkFileNotExist(expectedFile);
+  }
+
+  @Test
+  void saveBytesToFile_shouldNotThrowExceptionWhenUnableToSave(@TempDir Path tempDir) {
```
lucassaldanha: Could you please add a test for the IOException path as well? That is the one we should be worried about as it can be caused by unexpected reasons.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,143 @@
...
+    saveBytesToFile(
+        "gossip message with decoding error", topicPath.resolve(fileName), originalMessage);
+  }
+
+  public void saveGossipRejectedMessageToFile(
+      final String topic, final String arrivalTimestamp, final Bytes decodedMessage) {
+    if (!enabled) {
+      return;
+    }
+    final String fileName = String.format("%s_%s.ssz", arrivalTimestamp, topic);
```
rolfyone: we dont require the topic in both the filename and the directory, we should clean this up.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,143 @@
...
+    saveBytesToFile("rejected gossip message", topicPath.resolve(fileName), decodedMessage);
+  }
+
+  public void saveInvalidBlockToFile(
+      final UInt64 slot, final Bytes32 blockRoot, final Bytes blockSsz) {
+    if (!enabled) {
+      return;
+    }
+    final String fileName =
+        String.format("slot%s_root%s.ssz", slot, blockRoot.toUnprefixedHexString());
```
rolfyone: i would probably drop the prefix wordings, they're going to be very long filenames as it is, and this is just noise most likely, unless we have a really good reason for them

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/topics/topichandlers/Eth2TopicHandler.java
```diff
@@ -160,6 +170,10 @@ protected ValidationResult handleMessageProcessingError(
       final PreparedGossipMessage message, final Throwable err) {
     final ValidationResult response;
     if (ExceptionUtil.hasCause(err, DecodingException.class)) {
+      debugDataDumper.saveGossipMessageDecodingError(
+          getTopic(),
+          message.getArrivalTimestamp().map(UInt64::toString).orElse("unknown"),
+          message.getOriginalMessage());
```
rolfyone: we need to think about this a little, this will be a wall of numerics that someone has to parse with potentially thousands and thousands of very large numbers.

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/topics/Eth2TopicHandlerTest.java
```diff
@@ -186,7 +209,8 @@ public void handleMessage_errorWhileProcessing_rejectedExecution() {
             asyncRunner,
             (b, __) -> {
               throw new RejectedExecutionException("No more capacity");
-            });
+            },
+            null);
```
rolfyone: these nulls are fairly sub-optimal imo... do we have a Noop or something? it looks like they could be just passing something in rather than null.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/topics/topichandlers/Eth2TopicHandler.java
courtneyeh: That is already being done just below with the P2PLogger being called, so it is not much different. If we are concerned with that, we may need to update the P2PLogger's data input too.

courtneyeh: Ah I see what you mean, fixing now.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,157 @@
...
+            object);
+      }
+    }
+  }
+
+  @VisibleForTesting
+  protected static String formatTimestamp(final Optional<UInt64> arrivalTimestamp) {
+    if (arrivalTimestamp.isEmpty()) {
+      return "unknown";
+    }
```
rolfyone: when would we have no timestamp?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,157 @@
...
+
+    final DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SS");
+    Date date = new Date(arrivalTimestamp.get().longValue());
```
rolfyone: final

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,157 @@
...
+    return df.format(date);
+  }
+
+  @VisibleForTesting
+  protected boolean isEnabled() {
+    return enabled;
+  }
```
rolfyone: package private should be fine for testing, why protected?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
rolfyone: why static also?

courtneyeh: So I don't need to create a manager to use it in the tests

courtneyeh: Can change it if needed?

courtneyeh: Because `PreparedGossipMessage.getArrivalTimestamp()` returns an optional, so need to handle both

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,157 @@
...
+    if (!enabled) {
+      return;
+    }
+    final String fileName = String.format("%s_%s.ssz", slot, blockRoot.toUnprefixedHexString());
+    saveBytesToFile("invalid block", Path.of(INVALID_BLOCK_DIR).resolve(fileName), blockSsz);
+  }
+
+  @VisibleForTesting
+  protected void saveBytesToFile(
+      final String object, final Path relativeFilePath, final Bytes bytes) {
```
rolfyone: given other uses, i'd suggest 'description' would be a far better name for the first parameter than 'object', and reduce confusion reading the code.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,157 @@
...
+    final Path path = directory.resolve(relativeFilePath);
+    try {
+      Files.write(path, bytes.toArray());
+      LOG.info("Saved {} bytes to file. Location: {}", object, relativeFilePath);
```
rolfyone: we're going to be constantly getting
```
Saved rejected gossip message to file. location: invalid_blocks/1234_badbeefbadbeef.ssz
Saved rejected gossip message to file. location: invalid_blocks/5432_aabbccddeeffaa.ssz
```
and by the thousands potentially for gossip topics. this is going to be super noisy and potentially less useful.
I do agree that its potentially the 1 place we want info rather than debug, or at the level above where we can log more informatively...

For blocks i'd log something to the effect of 
`Saved invalid block at slot 12345(badbeefbadbeef)`
we don't really need the filename because its well defined and a large component of noise.

I'd probably suggest similar for gossip, logging.
`Saved rejected gossip message on topic {} at {}` 
last field being the formatted time or something...

The reason this is important is we need to be able to refer to these log messages for source data, and we dont want them more noisy than htey need to be.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
rolfyone: in which case, we cant use 'unknown'. we need to default sensibly here. otherwise we just overwrite that file forever.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,186 @@
...
+
+  public void saveInvalidBlockToFile(
+      final UInt64 slot, final Bytes32 blockRoot, final Bytes blockSsz) {
+    if (!enabled) {
+      return;
+    }
+    final String fileName = String.format("%s_%s.ssz", slot, blockRoot.toUnprefixedHexString());
+    final String identifiers = String.format("at slot %s(%s)", slot, blockRoot);
+    saveBytesToFile(
+        "invalid block", identifiers, Path.of(INVALID_BLOCK_DIR).resolve(fileName), blockSsz);
```
rolfyone: the more i htink about it, the more i think the message previously was correct including the level, and this change we just dont have the 'full block data' at the end.
I suggest again that we fix this logging so that its just less painful than passing it to the file saving and confusing concerns. I think the previous warn of `Rejecting invalid block at slot 1410696 with root 0xd6107c0e5eeaecd1ba385e65336cbf3c043110a2a399497ebb873229c57dcdc4 because FAILED_STATE_TRANSITION.` was about perfect, we just remove the 'full block data:' and write it to file ,and then its much easier to see old v. new functionality

courtneyeh: Okay will update it. Just keeping in mind that a user can have both the `--Xlog-include-p2p-warnings-enabled` and the new `--Xp2p-dumps-to-file-enabled` flag on at the same time. In that case there would be two logs in a row of the same thing but one would not have the full block data. However, it is unlikely a user would do that and that is their choice.

rolfyone: point being then we can compare both, so having the same message or just 1 at same level simplifies seeing the data

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,208 @@
...
+      final String description, final Path relativeFilePath, final Bytes bytes) {
+    final Path path = directory.resolve(relativeFilePath);
+    try {
+      Files.write(path, bytes.toArray());
+    } catch (NoSuchFileException e) {
+      if (!path.getParent().toFile().mkdirs()) {
+        LOG.error("Failed to save {} bytes to file.", description, e);
+        return false;
+      }
+      return saveAfterCreatingTopicDirectory(description, relativeFilePath, bytes);
```
rolfyone: if we're going to build things in this way, lets be a lot more careful about this.
We're possibly throwing exceptions in an exceptionHandler, but we're also just calling a function - lets do `mkdirs` in the child function instead, and then this is a lot less susceptible to failure.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -0,0 +1,210 @@
...
+    final Date date = new Date(arrivalTimestamp.longValue());
+    return df.format(date);
+  }
+
+  @VisibleForTesting
+  String generateTimestamp(final TimeProvider timeProvider) {
+    final DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH_mm_ss.SS");
+    final Date date = new Date(timeProvider.getTimeInMillis().longValue());
+    return df.format(date);
+  }
```
rolfyone: should probably make this cleaner, i dont see why we need both.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/noop/NoOpDebugDataDumper.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Optional;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.statetransition.util.DebugDataDumper;
+
+public class NoOpDebugDataDumper extends DebugDataDumper {
+
+  public NoOpDebugDataDumper() {
+    super(Path.of("/"), true);
```
rolfyone: "/" is never going to be an ok path for noop, suggest using "."

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/DebugDataDumperTest.java
```diff
@@ -0,0 +1,232 @@
...
+  @DisabledOnOs(OS.WINDOWS) // Can't set permissions on Windows
+  void constructionOfDirectories_shouldDisableWhenFailedToCreate(@TempDir Path tempDir) {
+    assertThat(tempDir.toFile().setWritable(false)).isTrue();
+    final DebugDataDumper manager = new DebugDataDumper(tempDir, true);
+    assertThat(manager.isEnabled()).isFalse();
+  }
+
+  @Test
+  void formatOptionalTimestamp_shouldFormatTimestamp() {
+    final DebugDataDumper manager = new DebugDataDumper(Path.of("."), true);
```
rolfyone: if we're going to do this we'll need to be using Tempdir, or we'll be polluting the working directory with folders constantly... i think similar elsewhere, this will be creating a bunch of folders.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/DebugDataDumperTest.java
```diff
@@ -0,0 +1,232 @@
...
+    final String formattedTimestamp =
+        manager.formatOptionalTimestamp(Optional.of(timeProvider.getTimeInMillis()), timeProvider);
+    assertThat(formattedTimestamp)
+        .isEqualTo(formatTimestamp(timeProvider.getTimeInMillis().longValue()));
+  }
+
+  @Test
+  void formatOptionalTimestamp_shouldGenerateTimestamp() {
+    final DebugDataDumper manager = new DebugDataDumper(Path.of("."), true);
```
rolfyone: same here, this will spam folders during testing.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -318,6 +320,10 @@ public BeaconChainController(
     this.receivedBlockEventsChannelPublisher =
         eventChannels.getPublisher(ReceivedBlockEventsChannel.class);
     this.forkChoiceExecutor = new AsyncRunnerEventThread("forkchoice", asyncRunnerFactory);
+    this.debugDataDumper =
+        new DebugDataDumper(
+            serviceConfig.getDataDirLayout().getDebugDataDirectory(),
+            beaconConfig.p2pConfig().isP2pDumpsToFileEnabled());
```
rolfyone: Potentially the more 'normal' implementation would be to use the Noop implementation here, rather than have this disabled/enabled flag going through the structure. 
It'd improve the overall flow, and the small difference is that if P2pDumps is enabled, use the real one, otherwise the noop.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java
```diff
@@ -41,12 +41,9 @@ public class DebugDataDumper {
   private boolean enabled;
   private final Path directory;
 
-  public DebugDataDumper(final Path directory, final boolean enabled) {
-    this.enabled = enabled;
+  public DebugDataDumper(final Path directory) {
+    this.enabled = true;
```
rolfyone, line 47: hard coded now, lets just fix the code. it doesn't need to exist.

lucassaldanha: This is still used as part of the circuit breaker so we should leave it.

lucassaldanha (approved):

PR #8178, opened 2024-04-08
Add an option to disable syncing the deposit contract logs for non-validator nodes

### LINKED ISSUE #7240: Add an option to disable syncing the deposit contract logs for non-validator nodes
### Description
It would be useful to have a option to disable syncing the deposit contract logs. This is useful if running an execution node for a purpose other than staking. It should not be used for staking nodes as it will cause missed block proposals.
For users who doesn't run a validator it's useful because they could turn off transaction receipts and other storage consuming data on execution layer side.

StefanBratanov:
## PR Description

- Introduce `--Xdeposit-contract-logs-syncing-enabled` which is true by default
- Added a CHANGELOG, which was deleted as part of the cleanup, but will be part of the next release

## Fixed Issue(s)
fixes #7240 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Additions and Improvements
 - Increased the executor queue default maximum size to 40_000 (previously 20_000), and other queues to 10_000 (previously 5_000). If you have custom settings for these queues, check to ensure they're still required.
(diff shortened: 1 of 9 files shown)


tbenr (approved):
LGTM. I'd add CLI option test. In particular to make sure it is enabled by default.

StefanBratanov:
> LGTM. I'd add CLI option test. In particular to make sure it is enabled by default.

Good point. Added a test.

PR #8185, opened 2024-04-10
Add support for slashing interchange format tests

### LINKED ISSUE #7784: Implement slashing interchange format tests for teku
https://github.com/eth-clients/slashing-protection-interchange-tests/releases/tag/v5.3.0

We've got all of the plumbing required, I did a quick hack to see if it would be painful...

Below shows one specific test basically running (in a really nasty way)

I view this as similar to our reference tests. I'm hoping at least initially we won't need to disable groups etc like we need to in reference tests.

We should be able to download a bundle and run through the stated tests, and maybe have a manual runner in the same way we do for the eth-reference-tests.
It should probably be its own module, rather than live in the dataexchange (which is where i wrote the test below)...

The TL/DR:
 - we can use the importer to import data to a slashing folder
 - we can use `LocalSlashingProtector` to check `maySignAttestation` and `maySignBlock` to verify test conditions
 
```
public class InterchangeReferenceTest {
  final String sourceFile =
      "slashing-protection-interchange-tests-5.3.0/tests/generated/duplicate_pubkey_not_slashable.json";
  private final JsonProvider jsonProvider = new JsonProvider();
  private final ObjectMapper mapper = jsonProvider.getObjectMapper();

  @Test
  void shouldReadFile(@TempDir Path tempDir) throws IOException {
    final JsonNode jsonData = mapper.readTree(Resources.getResource(sourceFile));
    final Path slashprotectionPath = tempDir.resolve("slashprotection");
    final Bytes32 genesisValidatorsRoot =
        Bytes32.fromHexString(jsonData.get("genesis_validators_root").asText());
    final String name = jsonData.get("name").asText();
    final SlashingProtectionInterchangeFormat interchangeFormat =
        jsonProvider.jsonToObject(
            jsonData.get("steps").get(0).get("interchange").toString(),
            SlashingProtectionInterchangeFormat.class);
    Files.writeString(
        tempDir.resolve("import.yml"),
        jsonData.get("steps").get(0).get("interchange").toString(),
        StandardCharsets.UTF_8);
    final BLSPublicKey pubkey =
        BLSPublicKey.fromHexString(
            "0xa99a76ed7796f7be22d5b7e85deeb7c5677e88e511e0b337618f8c4eb61349b4bf2d153f649f7b53359fe8b94a38e44c");
    Files.createDirectories(slashprotectionPath);
    SlashingProtectionImporter importer =
        new SlashingProtectionImporter(tempDir.resolve("slashprotection"));
    importer.initialise(new File(tempDir.resolve("import.yml").toString()));
    final Map<BLSPublicKey, String> errors = importer.updateLocalRecords((__) -> {});
    assertThat(errors).isEmpty();

    LocalSlashingProtector localSlashingProtector =
        new LocalSlashingProtector(
            SyncDataAccessor.create(slashprotectionPath), slashprotectionPath);
    assertThat(
            localSlashingProtector.maySignBlock(pubkey, genesisValidatorsRoot, UInt64.valueOf(10)))
        .isCompletedWithValue(false);
    assertThat(
            localSlashingProtector.maySignBlock(pubkey, genesisValidatorsRoot, UInt64.valueOf(13)))
        .isCompletedWithValue(false);
    assertThat(
            localSlashingProtector.maySignBlock(pubkey, genesisValidatorsRoot, UInt64.valueOf(14)))
        .isCompletedWithValue(true);
    assertThat(
            localSlashingProtector.maySignAttestation(
                pubkey, genesisValidatorsRoot, UInt64.valueOf(0), UInt64.valueOf(2)))
        .isCompletedWithValue(false);
    assertThat(
            localSlashingProtector.maySignAttestation(
                pubkey, genesisValidatorsRoot, UInt64.valueOf(1), UInt64.valueOf(3)))
        .isCompletedWithValue(false);
  }
}
```

StefanBratanov:
## PR Description
image

As per https://github.com/eth-clients/slashing-protection-interchange-tests

**Things to note:**

- Our implementation is minimal, so ignoring `should_succeed_complete`
- There is a test https://github.com/eth-clients/slashing-protection-interchange-tests/blob/master/tests/generated/wrong_genesis_validators_root.json which basically asserts that import should fail if there is a GVR mismatch. Added a comment to explain why we ignore this test.
- I have ignored `containsSlashableData`. Seems we have no problems importing anyways.

## Fixed Issue(s)
fixes #7784 

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/slashing_protection_interchange/SlashingProtectionInterchangeTestExecutor.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/slashing_protection_interchange/SlashingProtectionInterchangeTestExecutor.java
@@ -106,7 +106,7 @@ public class SlashingProtectionInterchangeTestExecutor implements TestExecutor {
       final SlashingProtectionImporter importer,
       final SlashingProtectionInterchangeFormat interchange) {
     try {
-      final Path importFile = Files.createTempFile("import", ".yml");
+      final Path importFile = Files.createTempFile("import", ".json");
       TestDataUtils.writeJsonToFile(interchange, importFile);
       final Optional<String> initialiseError = importer.initialise(importFile.toFile());
       assertThat(initialiseError).isEmpty();
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/slashing_protection_interchange/SlashingProtectionInterchangeTestExecutor.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/slashing_protection_interchange/SlashingProtectionInterchangeTestExecutor.java
@@ -15,79 +15,139 @@ package tech.pegasys.teku.reference.phase0.slashing_protection_interchange;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import java.nio.charset.StandardCharsets;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.List;
 import java.util.Map;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.bls.BLSPublicKey;
 import tech.pegasys.teku.data.SlashingProtectionImporter;
+import tech.pegasys.teku.data.slashinginterchange.SlashingProtectionInterchangeFormat;
 import tech.pegasys.teku.ethtests.finder.TestDefinition;
 import tech.pegasys.teku.infrastructure.io.SyncDataAccessor;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.reference.TestDataUtils;
 import tech.pegasys.teku.reference.TestExecutor;
+import tech.pegasys.teku.reference.phase0.slashing_protection_interchange.SlashingProtectionInterchangeTestExecutor.TestData.Step;
 import tech.pegasys.teku.spec.signatures.LocalSlashingProtector;
 
 public class SlashingProtectionInterchangeTestExecutor implements TestExecutor {
 
   private static final Logger LOG = LogManager.getLogger();
 
-  // TODO: implement the logic
   @Override
   public void runTest(final TestDefinition testDefinition) throws Throwable {
-    final JsonNode testNode = TestDataUtils.loadJson(testDefinition, testDefinition.getTestName());
+    final TestData testData =
+        TestDataUtils.loadJson(testDefinition, testDefinition.getTestName(), TestData.class);
 
-    final String testName = testNode.get("name").asText();
-    final Bytes32 genesisValidatorsRoot =
-        Bytes32.fromHexString(testNode.get("genesis_validators_root").asText());
+    // our implementation does not fail importing when the GVR in the interchange mismatches the one
+    // from the spec
+    if (testData.name.startsWith("wrong_genesis_validators_root")) {
+      LOG.info("Not going to run {}", testData.name);
+      return;
+    }
 
-    LOG.info("Running {}", testName);
+    LOG.info("Running {}", testData.name);
 
-    final Path tempDir = Files.createTempDirectory(testName);
+    final Path tempDir = Files.createTempDirectory(testData.name);
+    try {
+      runTest(testData, tempDir);
+    } finally {
+      deleteDirectory(tempDir);
+    }
+  }
 
-    final Path slashProtectionPath = tempDir.resolve("slashprotection");
+  private void runTest(final TestData testData, final Path tempDir) {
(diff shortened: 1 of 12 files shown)


mehdi-aouadi:
Regarding the `should_succeed_complete`: It looks like in Teku we've implemented a minimal strategy (we only keep track of the latest signed blocks and attestations) so we should only check the `should_succeed` and that's OK (the complete strategy being a complete database containing every message signed by each validator, which is not our case: definition from the [EIP-3076](https://eips.ethereum.org/EIPS/eip-3076#rationale))

mehdi-aouadi:
Not comfortable ignoring the `containsSlashableData` in the other hand

StefanBratanov:
> Not confortable ignoring the `containsSlashableData` on the other hand

Yeah, I wasn't sure what needs to be done on that part.

StefanBratanov:
Actually think should change this to import empty slashing protection first with the top GVR. This way it will check for GVR mismatch.

eth-tests/src/main/java/tech/pegasys/teku/ethtests/finder/ReferenceTestFinder.java
```diff
@@ -47,7 +47,10 @@ public static Stream<TestDefinition> findReferenceTests() throws IOException {
   private static Stream<TestDefinition> findTestTypes(final Path specDirectory) throws IOException {
     final String spec = specDirectory.getFileName().toString();
     if (spec.equals("bls")) {
-      return new BlsRefTestFinder().findTests(TestFork.PHASE0, spec, specDirectory);
+      return new BlsRefTestFinder().findTests("", spec, specDirectory);
```
zilm13, line 50: this `findTests` doesn't use fork at all, maybe remove parameter?

eth-tests/src/main/java/tech/pegasys/teku/ethtests/finder/ReferenceTestFinder.java
```diff
@@ -47,7 +47,10 @@ public static Stream<TestDefinition> findReferenceTests() throws IOException {
...
+    }
+    if (spec.equals("slashing-protection-interchange")) {
+      return new SlashingProtectionInterchangeRefTestFinder().findTests("", spec, specDirectory);
```
zilm13, line 53: this `findTests`uses fork in creating `TestDefinition`, but inits it with bad fork actually. if `createSpec()` is called, it will throw, confusing too

zilm13:
From description:

---
If `contains_slashable_data` is true, then implementations have the option to do one of two things:
-    Import the interchange successfully, working around the slashable data by minification or some other mechanism. If the import succeeds, all checks must pass and the test should continue to the next step.
-   Reject the interchange (or partially import it), in which case the block/attestation checks and all future steps should be ignored.

---
We do the 1st option so I think nothing could be added as far as we match test result

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/slashing_protection_interchange/SlashingProtectionInterchangeTestExecutor.java
```diff
@@ -0,0 +1,158 @@
...
+
+  private static final Logger LOG = LogManager.getLogger();
+
+  @Override
+  public void runTest(final TestDefinition testDefinition) throws Throwable {
+    final TestData testData =
+        TestDataUtils.loadJson(testDefinition, testDefinition.getTestName(), TestData.class);
+
+    // our implementation fails when importing one of the keys in an interchange, which is already
+    // in our slashprotection directory with a different genesis validators root. However, the test
```
zilm13, line 51: I'd add the name of the test which covers this

eth-tests/src/main/java/tech/pegasys/teku/ethtests/finder/ReferenceTestFinder.java
StefanBratanov: It is an interface method though, not sure how best to refactor

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/slashing_protection_interchange/SlashingProtectionInterchangeTestExecutor.java
StefanBratanov: good idea

eth-tests/src/main/java/tech/pegasys/teku/ethtests/finder/ReferenceTestFinder.java
StefanBratanov: I don't see how I can change this without a bigger refactor. If we wrongly use `Spec` when we don't need to, we will fail anyways.

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/slashing_protection_interchange/SlashingProtectionInterchangeTestExecutor.java
StefanBratanov: Actually there is no test for this, will add it.

eth-tests/src/main/java/tech/pegasys/teku/ethtests/finder/ReferenceTestFinder.java
zilm13: oh sorry, haven't checked this
ok, it's a tests

zilm13 (approved):
LGTM

PR #8186, opened 2024-04-10
Electra state changes

rolfyone:
Added the required state fields for EIP-7251 (MaxEB)
Some of the initializations are zeroed but need to be coming from state functions - these will be finalized in a follow-up PR, they need to be consistent with other implementations otherwise we'll be on different state roots, but equally this is 'ok' for now.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/versions/electra/PendingPartialWithdrawal.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public SszUInt64 getIndexSchema() {
+      return (SszUInt64) getFieldSchema0();
+    }
+
+    public SszUInt64 getAmountSchema() {
+      return (SszUInt64) getFieldSchema1();
+    }
+
+    public SszUInt64 getEpochSchema() {
```
mehdi-aouadi: Same here `getWithdrawalEpoch`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/versions/electra/PendingPartialWithdrawal.java
```diff
@@ -0,0 +1,78 @@
...
+    extends Container3<PendingPartialWithdrawal, SszUInt64, SszUInt64, SszUInt64> {
+
+  public static class PendingPartialWithdrawalSchema
+      extends ContainerSchema3<PendingPartialWithdrawal, SszUInt64, SszUInt64, SszUInt64> {
+    public PendingPartialWithdrawalSchema() {
+      super(
+          "PendingPartialWithdrawal",
+          namedSchema("index", SszPrimitiveSchemas.UINT64_SCHEMA),
+          namedSchema("amount", SszPrimitiveSchemas.UINT64_SCHEMA),
+          namedSchema("epoch", SszPrimitiveSchemas.UINT64_SCHEMA));
```
mehdi-aouadi: This should be `withdrawable_epoch` instead [according to the spec](https://github.com/ethereum/consensus-specs/blob/88f5dde0f973151e9df84ba6b15d1c67287df7ee/specs/_features/eip7251/beacon-chain.md?plain=1#L182)

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/schema/PendingPartialWithdrawal.json
```diff
@@ -0,0 +1,25 @@
+{
+  "title" : "PendingPartialWithdrawal",
+  "type" : "object",
+  "required" : [ "index", "amount", "epoch" ],
```
mehdi-aouadi: Same here should be `withdrawable_epoch`

data/serializer/src/main/java/tech/pegasys/teku/api/schema/electra/PendingPartialWithdrawal.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class PendingPartialWithdrawal {
+  @JsonProperty("index")
+  public final int index;
+
+  @JsonProperty("amount")
+  public final UInt64 amount;
+
+  @JsonProperty("withdrawable_epoch")
+  public final UInt64 epoch;
```
mehdi-aouadi: Could rename the variable accordingly: `withdrawble_epoch`

mehdi-aouadi (approved):
LGTM
Just a small variable renaming needed (`withdrawable_epoch` instead of `epoch`)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/versions/electra/PendingPartialWithdrawal.java
rolfyone: i'll have to check the spec, it looked like it was called 'epoch' but maybe its been updated again... thanks!

rolfyone: it was always called that i think i just missed a rename in copying things, so appreciate the catch, one less thing to fix when implementing :)

PR #8188, opened 2024-04-11
Disable requirement for EIP-7251 configuration in Electra config

zilm13:
## PR Description
Adding defaults to EIP-7251 constants because they are not yet part of the official Electra config, they have [separate EIP-7251 fork](https://github.com/ethereum/consensus-specs/blob/dev/configs/mainnet.yaml#L56-L58), so some other tools (Kurtosis) are not expecting that we require these constants for Electra config.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8189, opened 2024-04-11
Don't encode ids array comma separators in getStateValidator request

zilm13:
## PR Description
- fixed validator states were queried with GET ids list encoded, but shouldn't, check old rationale for this in #7757
- added test for this
- clarified that we are not required to call any other BN API with array in query parameter

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8192, opened 2024-04-12
Fix sync committee when syncing

### LINKED ISSUE #8005: We should not create and publish sync committee if node is syncing
I stopped EL, node switched to syncing mode. We correctly stopped producing attestations but we continued contributing to sync committee.
Is it also related to #7862 ?
```
{"@timestamp":"2024-02-20T14:45:40,001","level":"WARN","thread":"validator-async-6","class":"teku-validator-log","message":"Validator   *** Skipped producing attestation while node is syncing  Count: 320, Slot: 7661028","throwable":""}
{"@timestamp":"2024-02-20T14:45:40,007","level":"INFO","thread":"validator-async-0","class":"teku-validator-log","message":"Validator   *** Published sync_signature     Count: 7, Slot: 7661028, Root: 630c707447131405438f1e418ed82dbfd6b5a933e88c394282e27b3dbe87c198","throwable":""}
{"@timestamp":"2024-02-20T14:45:41,870","level":"ERROR","thread":"pool-4-thread-346","class":"TimeBasedEth1HeadTracker","message":"Failed to update eth1 chain head - Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:43,870","level":"ERROR","thread":"pool-4-thread-346","class":"TimeBasedEth1HeadTracker","message":"Failed to update eth1 chain head - Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:44,000","level":"WARN","thread":"ValidatorTimingChannel-0","class":"teku-validator-log","message":"Validator   *** Skipped producing aggregate while node is syncing  Count: 15, Slot: 7661028","throwable":""}
{"@timestamp":"2024-02-20T14:45:45,871","level":"ERROR","thread":"pool-4-thread-346","class":"TimeBasedEth1HeadTracker","message":"Failed to update eth1 chain head - Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:47,872","level":"ERROR","thread":"pool-4-thread-346","class":"TimeBasedEth1HeadTracker","message":"Failed to update eth1 chain head - Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:48,023","level":"INFO","thread":"TimeTick-1","class":"teku-event-log","message":"Syncing     *** Target slot: 7661029, Head slot: 7660932, Remaining slots: 97, Connected peers: 64","throwable":""}
{"@timestamp":"2024-02-20T14:45:48,024","level":"ERROR","thread":"beaconchain-async-35","class":"ForkChoice","message":"Failed to update fork choice. tech.pegasys.teku.ethereum.executionclient.response.InvalidRemoteResponseException: Invalid remote response from the execution client: Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:49,206","level":"WARN","thread":"sync-async-3","class":"BatchSync","message":"Unable to import blocks because execution client is offline.","throwable":""}
{"@timestamp":"2024-02-20T14:45:49,873","level":"ERROR","thread":"pool-4-thread-346","class":"TimeBasedEth1HeadTracker","message":"Failed to update eth1 chain head - Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:51,873","level":"ERROR","thread":"pool-4-thread-346","class":"TimeBasedEth1HeadTracker","message":"Failed to update eth1 chain head - Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:52,000","level":"WARN","thread":"validator-async-2","class":"teku-validator-log","message":"Validator   *** Skipped producing attestation while node is syncing  Count: 327, Slot: 7661029","throwable":""}
{"@timestamp":"2024-02-20T14:45:52,004","level":"INFO","thread":"validator-async-3","class":"teku-validator-log","message":"Validator   *** Published sync_signature     Count: 7, Slot: 7661029, Root: 630c707447131405438f1e418ed82dbfd6b5a933e88c394282e27b3dbe87c198","throwable":""}
{"@timestamp":"2024-02-20T14:45:53,874","level":"ERROR","thread":"pool-4-thread-346","class":"TimeBasedEth1HeadTracker","message":"Failed to update eth1 chain head - Failed to connect to /127.0.0.1:8551","throwable":""}
{"@timestamp":"2024-02-20T14:45:54,301","level":"WARN","thread":"sync-async-3","class":"BatchSync","message":"Unable to import blocks because execution client is offline.","throwable":""}
{"@timestamp":"2024-02-20T14:45:55,523","level":"WARN","thread":"powchain-async-1617","class":"teku-status-log","message":"Eth1 service down or still syncing for 120s, retrying","throwable":""}
```

tbenr:
fixes #8005

The aggregation fix is non-controversial while the sync message is not that clean, but it is the easiest implementation I came up with so far.

Essentially we will stop producing sync committee message if the head we have is older than 32 slot, which means:
1- we are connected to a BN which is falling out of sync.
2- the chain is not production a block for more than entire epoch, we have bigger problems.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
@@ -149,9 +149,7 @@ public class ExecutionLayerChannelStub implements ExecutionLayerChannel {
 
   @Override
   public SafeFuture<Optional<PowBlock>> eth1GetPowBlock(final Bytes32 blockHash) {
-    if (!online) {
-      return SafeFuture.failedFuture(new RuntimeException("stub is offline"));
-    }
+    offlineCheck();
 
     if (!transitionEmulationEnabled) {
       requestedPowBlocks.add(blockHash);
@@ -178,9 +176,7 @@ public class ExecutionLayerChannelStub implements ExecutionLayerChannel {
 
   @Override
   public SafeFuture<PowBlock> eth1GetPowChainHead() {
-    if (!online) {
-      return SafeFuture.failedFuture(new RuntimeException("stub is offline"));
-    }
+    offlineCheck();
 
     if (!transitionEmulationEnabled) {
       return SafeFuture.failedFuture(
@@ -205,9 +201,7 @@ public class ExecutionLayerChannelStub implements ExecutionLayerChannel {
   public SafeFuture<ForkChoiceUpdatedResult> engineForkChoiceUpdated(
       final ForkChoiceState forkChoiceState,
       final Optional<PayloadBuildingAttributes> payloadBuildingAttributes) {
-    if (!online) {
-      return SafeFuture.failedFuture(new RuntimeException("stub is offline"));
-    }
+    offlineCheck();
 
     if (!bellatrixActivationDetected) {
       LOG.info(
@@ -243,9 +237,7 @@ public class ExecutionLayerChannelStub implements ExecutionLayerChannel {
   @Override
   public SafeFuture<GetPayloadResponse> engineGetPayload(
       final ExecutionPayloadContext executionPayloadContext, final BeaconState state) {
-    if (!online) {
-      return SafeFuture.failedFuture(new RuntimeException("stub is offline"));
-    }
+    offlineCheck();
 
     if (!bellatrixActivationDetected) {
       LOG.info(
@@ -330,9 +322,7 @@ public class ExecutionLayerChannelStub implements ExecutionLayerChannel {
 
   @Override
   public SafeFuture<PayloadStatus> engineNewPayload(final NewPayloadRequest newPayloadRequest) {
-    if (!online) {
-      return SafeFuture.failedFuture(new RuntimeException("stub is offline"));
-    }
+    offlineCheck();
 
     final ExecutionPayload executionPayload = newPayloadRequest.getExecutionPayload();
     final PayloadStatus returnedStatus =
@@ -349,9 +339,7 @@ public class ExecutionLayerChannelStub implements ExecutionLayerChannel {
 
   @Override
   public SafeFuture<List<ClientVersion>> engineGetClientVersion(final ClientVersion clientVersion) {
-    if (!online) {
-      return SafeFuture.failedFuture(new RuntimeException("stub is offline"));
-    }
+    offlineCheck();
 
     return SafeFuture.completedFuture(List.of(STUB_CLIENT_VERSION));
   }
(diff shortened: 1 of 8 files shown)


StefanBratanov (approved):
LGTM

tbenr:
@rolfyone I'm not a fan of the solution for the sync committee. The aggregate is fine.

dumping some messages wrote in chat:

> essentially VC tracks the head, creates the message, and publish to the BN
> 
> so this is the only duty that is not created by the BN
> so the creation can not be directly associated to a syncing state
> to do a similar thing on VC we should have a ChainHeadTracker that becomes aware of the syncing state of the BN we are currently connected to
> 
> we could subscribe to sync event stream
> but i think there are a lot of edge cases in case of failovers

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
```diff
@@ -392,6 +425,10 @@ public SafeFuture<HeaderWithFallbackData> builderGetHeader(
   public SafeFuture<BuilderPayload> builderGetPayload(
       final SignedBeaconBlock signedBeaconBlock,
       final Function<UInt64, Optional<ExecutionPayloadResult>> getCachedPayloadResultFunction) {
+    if (!online) {
+      return SafeFuture.failedFuture(new RuntimeException("stub is offline"));
```
StefanBratanov: Can probably extract this in a method cause too many repetitions

rolfyone:
> so this is the only duty that is not created by the BN
> so the creation can not be directly associated to a syncing state
> to do a similar thing on VC we should have a ChainHeadTracker that becomes aware of the syncing state of the BN we are currently connected to

ultimately its 1 persistent duty for sync committees which is different to other dutes.. that's why the VC is doing more... I think it's practical overall, but it's definitely different to all the others...

tbenr:
> ultimately its 1 persistent duty for sync committees which is different to other dutes.. that's why the VC is doing more... I think it's practical overall, but it's definitely different to all the others...

I think It is fine yes. We could be wrong only if the chain is not producing blocks for an entire epoch. In that case i don't think there would be enough block space to include them all.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
tbenr: mm not super sure how we can effectively do that

tbenr: done

Issue #8180, opened 2024-04-08
Degraded attestation performance when Teku VC has a secondary BN defined


rolfyone:
From Nash on discord:

Scenario A. Teku VC has a primary (teku) and secondary (lighthouse) BNs. The secondary BN reports libp2p issues and is later taken down, the whole time we are observing abnormal attestation issues and extremely poor sync performance. As soon as we remove the secondary BN from the config the performance normalizes! While experiencing the issue Teku-VC is complaining with
Apr 04 09:52:41 ursa-Xset teku[2401221]: 09:52:41.974 ERROR - Validator   *** Error while connecting to beacon node event stream com.launch
darkly.eventsource.StreamIOException: java.net.ConnectException: Failed to connect to /127.0.0.1:1752 (See log file for full stack trace)
Apr 04 09:52:44 ursa-Xset teku[2401221]: 09:52:44.000 WARN  - Validator   *** There are no beacon nodes from the configured ones that are ready to be used as an event stream failover

Scenario B. Teku VC has a primary (teku) and secondary (lighthouse) BNs. We observe minor sporadic attestation issues, but no obvious logs on Teku-VC side. We bring down secondary-BN, no logs or complaints from Teku-VC but attestation issues continue. Once we remove secondary-BN from VC config perfomance is normalized.
We have disabled dual-homing of our Teku-VC until further understanding of the situation. There's several obvious issues here:
overall perf degradation when using 2BNs.
decreased perf when failover (!) BN is down.

-----
will ask to confirm lighthouse / teku versions and we'll need further investigation.

astrostakers:
Astro-Stakers team here, additional comment:
When experiencing the issue we swapped Lighthouse for Nimbus (suspecting Teku <> LH incompatibility issues) but the issues continued with Nimbus, we suspect there's a general degradation issue when Teku has 2 BNs. 

Versions:
- Ligthouse Gul'karna
- Teku 24.3.1

VC config:
```
[Unit]
Description=Teku VC Holesky
Wants=network-online.target
After=network-online.target

[Service]
Type=simple
User=validator
Group=validator
Restart=on-failure
RestartSec=3
KillSignal=SIGINT
TimeoutStopSec=900
ExecStart=/usr/local/bin/teku/bin/teku validator-client \
  --network=holesky \
  --data-path=/var/lib/teku_validator \
  --validator-keys=/var/lib/teku_validator/validator_keys:/var/lib/teku_validator/validator_keys \
[...]

PR #8193, opened 2024-04-12
Trigger primary BN readiness check when failover has issues

StefanBratanov:
## PR Description
Currently we wait `onAttestationAggregationDue` to trigger the readiness check for beacon nodes, which in turn trigger event streaming failing over/returning to primary. However, if there is a failover issue and there are no other failovers available, in some edge cases VC has to wait around a slot time before the `onPrimaryNodeBackReady` callback is called and the failover to primary can happen.

The edge case is  `onAttestationAggregationDue` happens. Both primary and failover are in a bad state. We are still connected to a failover, so we try switching but nothing is available. In the meantime, primary has recovered. However, we still have to wait until the next `onAttestationAggregationDue` before reconnecting.

## Fixed Issue(s)
related to #8180 

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
@@ -77,6 +77,10 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
         .iterator();
   }
 
+  public SafeFuture<Void> performPrimaryReadinessCheck() {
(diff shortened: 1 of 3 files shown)


mehdi-aouadi (approved):
LGTM
I'd suggest adding a test for this edge case though.
Overall I'm thinking if we could de-correlate the readiness check from the `onAttestationAggregationDue` and make it run autonomously, async and time based or something like that...

StefanBratanov:
> LGTM I'd suggest adding a test for this edge case though. Overall I'm thinking if we could de-correlate the readiness check from the `onAttestationAggregationDue` and make it run autonomously, async and time based or something like that...

Good idea. Added a test.

PR #8195, opened 2024-04-12
Fix PendingPartialWithdrawal deserialization mistake

jtraglia:
## PR Description

This PR adds property tests for the three new "pending*" structures. While doing this, I noticed that `createFromBackingNode` was returning `null` which meant `sszDeserialize` would return `null` too. The property tests for this structure were failing because of it.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM. Thanks a lot for these!

rolfyone:
Thanks, you're ahead of me!

PR #8196, opened 2024-04-12
Invert validatorActiveLongEnough check to be more clear

jtraglia:
## PR Description

There was a hint of confusion when reading this check. The variable name (`validatorActiveLongEnough`) doesn't quite match what it represents. I was reading it as: "if the validator is active long enough, reject this" which is wrong.

https://github.com/Consensys/teku/blob/c862af719a68b93d3394d957196486fe5dfced42/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java#L186-L192

Also add a few `final` specifiers which I know we like.

## Documentation

## Changelog


courtneyeh (approved):
LGTM

PR #8205, opened 2024-04-15
Fix capella signed voluntary exit was not published on Deneb

zilm13:
## PR Description

Not sure it's the case of #8158 report because it lacks of the information, but it's a valid case and we shouldn't fail on it. Fixed now.
Not sure about Changelog.

So, user could create and sign `VoluntaryExit` in Capella and send it later in any other fork and it's a valid `SignedVoluntaryExit` which we should publish.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #8209, opened 2024-04-16
Fix flakiness in BlockProposalAcceptanceTest

zilm13:
## PR Description
With execution layer stub (but could happen in a real life too probably) the subscription was made after execution client version was pushed in channel. So it was lost. Fixed now.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8212, opened 2024-04-17
Add final to method input params - beacon module

gfukushima:
## PR Description
This is PR is part of a series where the aim is to add missing final to enable the MethodInputParametersMustBeFinal error prone validation.

## Documentation

## Changelog


zilm13 (approved):
LGTM
So it will be not required on lines like this: https://github.com/Consensys/teku/blob/fc62d3bdeb96a4ed1832b71d665f3e76e86fd10e/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/AttestationPerformance.java#L66 
Only for method inputs?

lucassaldanha:
> Only for method inputs?

That's a good point. I believe we are only applying `MethodInputParametersMustBeFinal`. @gfukushima could you check if we have a rule for member variables and local variables that can be final as well?

gfukushima:
The EP check for finals in variables is a different one (FieldCaBeFinal) and we should probably enable it as well. I'd like to avoid mixing those for now but I do plan to submit a PR to enable that as well

PR #8215, opened 2024-04-17
Refactor block production flow

### LINKED ISSUE #8052: Don't prepare blinded content from local execution data, during blockV3 flow
When creating block via BlockV3, the API will respond with full content when the flow falls back to local.
So in that situation there is no point in preparing blinded content

StefanBratanov:
## PR Description

- Rename `HeaderWithFallbackData` to `BuilderBidOrFallbackData` (essentially storing BuilderBid - either local or real builder)
- Also removed the block value future in the `ExecutionBuilderModule` since we can use the builder bid value (regardless if we fallback to local or use real builder, it will be set now)
- Changed `ExecutionLayerBlockProductionManager` to have only one method. Essentially, there is no benefit, the only difference is that we pass Optional for a future of blobs, but we can instead rely on the response from builder or local where there it would resolve to empty anyways.
- Refactored `BlockOperationSelectorFactory` qui te a bit given that there is one method in `ExecutionLayerBlockProductionManager`.
- Removed using `setUnblindedContentIfBuilderFallbacks` when setting kzg commitments in the block. Can just use the builder then (since it's mimicked). Content in the block is the same regardless if blinded or not.
- Adapted tests.
- Refactored `ExecutionPayloadResult` initialisation

## Fixed Issue(s)
fixes #8052 

## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
```diff
@@ -77,59 +77,10 @@ public ExecutionPayloadResult initiateBlockProduction(
       final BlockProductionPerformance blockProductionPerformance) {
     final ExecutionPayloadResult result;
     if (!isBlind) {
```
zilm13: I'd reverse if, will be easier to read

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerManagerImplTest.java
```diff
@@ -486,41 +431,18 @@ public void builderGetHeaderGetPayload_shouldReturnHeaderAndPayloadViaEngineOnBu
...
-    final HeaderWithFallbackData expectedResult =
-        HeaderWithFallbackData.create(
-            expectedHeader,
-            Optional.of(expectedBlobKzgCommitments),
+    final BuilderBidWithFallbackData expectedResult =
+        BuilderBidWithFallbackData.create(
+            createBuilderBid(getPayloadResponse),
             new FallbackData(
                 getPayloadResponse.getExecutionPayload(),
                 getPayloadResponse.getBlobsBundle(),
```
zilm13: I'd add a check above that blobsBundle is not empty

zilm13 (commented):
Great refactoring but I think we shouldn't be `BuilderBid` centric and mimicry it in `BuilderBidWithFallbackData`. It's a bit confusing to found local data in `BuilderBid` with stub on builder's `pubkey` So I'd change to either have `HeaderAndMeta` inside this object with all fields from `BuilderBid` except `pubkey` or just expanding these fields into `BuilderBidWithFallbackData`. Also its name would be better without `BuilderBid`.

StefanBratanov:
> Great refactoring but I think we shouldn't be `BuilderBid` centric and mimicry it in `BuilderBidWithFallbackData`. It's a bit confusing to found local data in `BuilderBid` with stub on builder's `pubkey` So I'd change to either have `HeaderAndMeta` inside this object with all fields from `BuilderBid` except `pubkey` or just expanding these fields into `BuilderBidWithFallbackData`. Also its name would be better without `BuilderBid`.

I have been thinking about that. It will bring bit of inconsistency, because on the same note, we also have `BuilderPayload` which could be either local fallback or real builder. In that way having `BuilderBid` for builder or local fallback makes sense, but I can see what you mean. What is your opinion @tbenr and @mehdi-aouadi  ?

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -214,200 +214,125 @@ private SafeFuture<Void> setExecutionData(
     }
 
     // if requestedBlinded has been specified, we strictly follow it otherwise, we should run
```
tbenr: `..., we strictly follow it, otherwise we should run ...`

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -214,200 +214,125 @@ private SafeFuture<Void> setExecutionData(
...
-                            .setBlockExecutionValue(value)))
-        .orElse(SafeFuture.COMPLETE);
+        .thenAccept(
+            blockExecutionValue ->
+                BeaconStateCache.getSlotCaches(blockSlotState)
+                    .setBlockExecutionValue(blockExecutionValue));
   }
 
-  private SafeFuture<Void> builderSetPayloadPostMerge(
+  private SafeFuture<Void> setPayloadPostMerge(
```
tbenr: I would throw away "PostMerge"

tbenr (commented):
added a couple of comments.

WRT BuilderBid discussion. I feel it would be more readable if we can get the rid of this mimicking..

For instance, I'm thinking about what it we expand the `BuilderBidWithFallbackData` class to me more clever, and actually store a BuilderBid in case of builder flow, and the actual local data as separate data fields.

we could have the 2 different `create` be used for the two flows.

The class than would be responsible to expose consistent data externally, independently from the flow that has been executed.

The class would be named something like `PayloadBuildingResult` or whatever more general name

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -213,201 +216,154 @@ private SafeFuture<Void> setExecutionData(
...
+    final Function<BuilderBidOrFallbackData, Boolean> setUnblindedContentIfBuilderFallbacks =
+        builderBidOrFallbackData ->
+            builderBidOrFallbackData.getFallbackData().isPresent() && requestedBlinded.isEmpty();
+
     return SafeFuture.allOf(
         cacheExecutionPayloadValue(executionPayloadResult, blockSlotState),
-        builderSetPayloadPostMerge(
-            bodyBuilder, setUnblindedContentIfBuilderFallbacks, executionPayloadResult),
-        builderSetKzgCommitments(
+        // Execution Payload / Execution Payload Header
```
tbenr, line 255: I think we don't need these comments here anymore

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
```diff
@@ -76,61 +73,12 @@ public ExecutionPayloadResult initiateBlockProduction(
...
-          new ExecutionPayloadResult(
-              context,
-              Optional.of(executionPayloadFuture),
-              Optional.empty(),
-              Optional.empty(),
-              Optional.of(executionPayloadValueFuture));
-    } else {
-      result =
-          builderGetHeader(
+          processBlindedFlow(
```
tbenr, line 78: `execute(Non)BlindedFlow`?

tbenr (approved):
LGTM

Issue #7690, opened 2023-11-10
[keymanager-api] add graffiti management 


rolfyone:
https://github.com/ethereum/keymanager-APIs/pull/63

tbenr:
This is also related to https://github.com/Consensys/teku/issues/3471 in which I suggested to add graffiti in the proposer-config file

PR #8216, opened 2024-04-17
Create graffiti management

courtneyeh:
#7690 

## Documentation

## Changelog


validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,74 @@
...
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.function.Supplier;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import tech.pegasys.teku.bls.BLSPublicKey;
+import tech.pegasys.teku.service.serviceutils.layout.DataDirLayout;
+
+public class GraffitiManager {
+  static final String GRAFFITI_MANAGEMENT_DIR = "graffiti-management";
```
lucassaldanha: NIT: we can probably go for `graffiti`

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public GraffitiManager(final DataDirLayout dataDirLayout) {
+    this.graffitiPath = createManagementDirectory(dataDirLayout);
+  }
+
+  public Optional<String> setGraffiti(final BLSPublicKey publicKey, final String graffiti) {
+    return updateGraffiti(publicKey, () -> Bytes32Parser.toBytes32(graffiti).toArray());
+  }
+
+  public Optional<String> deleteGraffiti(final BLSPublicKey publicKey) {
```
lucassaldanha: Why return `Optional<String>` on delete?

validator/api/src/main/java/tech/pegasys/teku/validator/api/UpdatableGraffitiProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.nio.file.Path;
+import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.bls.BLSPublicKey;
+import tech.pegasys.teku.service.serviceutils.layout.DataDirLayout;
+
+public class UpdatableGraffitiProvider implements GraffitiProvider {
```
lucassaldanha, line 20: I think this implementation is not correct. After deleting a graffiti through the API we should fallback into returning the value from the default graffiti provider.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,74 @@
...
+      return Optional.empty();
+    }
+    return Optional.of(graffitiDirectory);
+  }
+
+  private Optional<String> updateGraffiti(
+      final BLSPublicKey publicKey, final Supplier<byte[]> graffiti) {
+    if (graffitiPath.isEmpty()) {
+      return Optional.of("graffiti-management directory does not exist to handle update.");
+    }
```
lucassaldanha: I am a bit confused by this flow. If during startup we can't create the graffiti subdirectory we are probably in a bad state as the user running Teku should have full access to those folders.

If we just ignore the fact that the directory was not created, how will the user know that the method is not working?

Should we even start up the node successfully if we can't create/access the graffiti directory?

validator/api/src/main/java/tech/pegasys/teku/validator/api/UpdatableGraffitiProvider.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final Logger LOG = LogManager.getLogger();
+
+  private final Path graffitiPath;
+  private final GraffitiProvider defaultProvider;
+
+  public UpdatableGraffitiProvider(
+      final DataDirLayout dataDirLayout,
```
lucassaldanha: One thing I do not like about this approach is the fact that two objects now need to know details about where the graffiti is being saved and the format of the file, etc.

I think centralizing this knowledge on the `GraffitiManager` is more cohesive.

Have you considered this approach?

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -153,8 +157,16 @@ public static ValidatorClientService create(
         new GenesisDataProvider(asyncRunner, validatorApiChannel);
     final ForkProvider forkProvider = new ForkProvider(config.getSpec(), genesisDataProvider);
 
-    final ValidatorLoader validatorLoader = createValidatorLoader(services, config, asyncRunner);
     final ValidatorRestApiConfig validatorApiConfig = config.getValidatorRestApiConfig();
+    final GraffitiManager graffitiManager = new GraffitiManager(services.getDataDirLayout());
```
lucassaldanha: We should make this initialization conditional to the KeyManager API being enabled.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,97 @@
...
+      return Optional.of(errorMessage);
+    }
+    return Optional.empty();
+  }
+
+  public Optional<Bytes32> getGraffitiFromStorage(final BLSPublicKey publicKey) {
+    final Path filePath = graffitiPath.resolve(resolveFileName(publicKey));
+    try {
+      return Optional.of(GraffitiParser.loadFromFile(filePath)).filter(this::graffitiNotEmpty);
+    } catch (GraffitiLoaderException | IllegalArgumentException e) {
```
lucassaldanha: I believe we need to differentiate an error reading the graffiti from an error in the input parameter (e.g. invalid public key). How are we going to fulfill the API without separating those two cases?

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,97 @@
...
+    } catch (IOException e) {
+      final String errorMessage =
+          String.format("Unable to update graffiti for validator %s", publicKey);
+      LOG.error(errorMessage, e);
+      return Optional.of(errorMessage);
+    }
+    return Optional.empty();
+  }
+
+  public Optional<Bytes32> getGraffitiFromStorage(final BLSPublicKey publicKey) {
```
lucassaldanha: NIT: we could call this function `getGraffiti`

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,97 @@
...
+    if (!graffitiDirectory.toFile().exists() && !graffitiDirectory.toFile().mkdirs()) {
+      throw new IllegalStateException(
+          "Unable to create " + GRAFFITI_DIR + " directory for graffiti management.");
+    }
+    return graffitiDirectory;
+  }
+
+  private Optional<String> updateGraffiti(final BLSPublicKey publicKey) {
+    return updateGraffiti(publicKey, "");
+  }
```
lucassaldanha: Using this approach for deleting the graffiti is a bad idea. Two problems I can think of:
1. We can end up with a bunch of empty files that do not add any value;
2. It prevents us from being able to set their graffiti as an empty string;

IMHO we should delete the file when the user deletes their graffiti. It solves both problems.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return Optional.empty();
+  }
+
+  public Optional<Bytes32> getGraffitiFromStorage(final BLSPublicKey publicKey) {
+    final Path filePath = graffitiPath.resolve(resolveFileName(publicKey));
+    try {
+      return Optional.of(GraffitiParser.loadFromFile(filePath)).filter(this::graffitiNotEmpty);
+    } catch (GraffitiLoaderException | IllegalArgumentException e) {
+      LOG.error("Unable to read graffiti from storage.", e);
```
lucassaldanha: We also need to treat differently when the file does not exist. This isn't an error, it just means that the user has not created a graffiti for that validator.

validator/api/src/test/java/tech/pegasys/teku/validator/api/GraffitiManagerTest.java
```diff
@@ -0,0 +1,220 @@
...
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.io.TempDir;
+import tech.pegasys.techu.service.serviceutils.layout.SimpleDataDirLayout;
+import tech.pegasys.teku.bls.BLSPublicKey;
+import tech.pegasys.teku.service.serviceutils.layout.DataDirLayout;
+import tech.pegasys.teku.spec.TestSpecFactory;
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+
+class GraffitiManagerTest {
```
lucassaldanha, line 38: We are missing a test case for when we read the graffiti and the file for that validator does not exist.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
courtneyeh: The public key will be checked using the key manager in the API `handleRequest(...)` methods. This class just manages saving / getting graffiti. A key would not make it to this manager without being valid.

courtneyeh: In this method the graffiti is being retrieved and the public key is the one that is input into the Validator when it is created. There is no scenario I can see where the public key would not be the same as one inside a validator (i.e. always valid). This method is used in the new graffiti provider, which is initialised with the public key that is also inputted into the constructor of the validator at the same time.

CHANGELOG.md
```diff
@@ -19,5 +19,6 @@ the [releases page](https://github.com/Consensys/teku/releases).
 - Updated the bootnodes for Chiado and Gnosis networks.
 - Added hidden option `--Xp2p-dumps-to-file-enabled` to enable saving p2p dumps to file.
 - Appends consensus layer (CL) and execution layer (EL) clients' information to the validator graffiti. Check [documentation](https://docs.teku.consensys.io/development/reference/cli#validators-graffiti-client-append-format) for available configuration options.
+- Added GET `/eth/v1/validator/{pubkey}/graffiti` validator API.
```
lucassaldanha: We should updated the message to say that we have implemented all Graffiti key management API methods.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  public GraffitiManager(final Path directory) {
+    this.directory = directory;
+    if (!directory.toFile().exists() && !directory.toFile().mkdirs()) {
+      throw new IllegalStateException("Unable to create directory for graffiti management.");
+    }
+  }
+
+  public void setGraffiti(final BLSPublicKey publicKey, final String graffiti) throws IOException {
```
lucassaldanha: The point of throwing an exception at the business logic layer is to be able to encapsulate different errors related to the business logic implemented into an exception class associated with your logic. In this example, you would have an exception for GraffitiManagement (similar to how we have `SetFeeRecipientException` for example).

We probably do not need one exception for set and another for delete, but something like `GraffitiManagementException`.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
lucassaldanha: Here is the scenario I am thinking:
1. User uses POST method to create a graffiti for validator x;
2. For some reason, file gets corrupted;
3. User calls GET to retrieve graffiti for validator x;
4. User will get the system-default graffiti (even though the POST request was successful).

From the user point-of-view, the POST is broken because it feels like it did not save his graffiti.

What I would expect is to get a 500 response in the API in this case (from the spec: 	
Internal server error. The server encountered an unexpected error indicative of a serious fault in the system, or a bug).

How do you think we should handle this?

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
lucassaldanha: The same applies to the other methods.

validator/api/src/main/java/tech/pegasys/teku/validator/api/UpdatableGraffitiProvider.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public UpdatableGraffitiProvider(
+      final Supplier<Optional<Bytes32>> storageProvider, final GraffitiProvider defaultProvider) {
+    this.storageProvider = storageProvider;
+    this.defaultProvider = defaultProvider;
+  }
+
+  @Override
+  public Optional<Bytes32> get() {
+    return storageProvider.get().or(defaultProvider::get);
```
lucassaldanha, line 47: I think it would be safer to encapsulate the call to the `storageProvider` function into a try-catch block to avoid any surprises. Or do something similar to what is done in the `FileBackedGraffitiProvider` (that has a try-catch Exception block to handle ANY errors). No matter how we do it, we need to ensure this won't fail during block production.

The logic should be: no matter what explodes in the storageProvider, we want to ignore it and delegate to the defaultProvider.

The following test should pass:
```
  @Test
  public void shouldDelegateToDefaultProviderWhenStorageProviderFails() {
    final Supplier<Optional<Bytes32>> fn = mock(Supplier.class);
    when(fn.get()).thenThrow(new RuntimeException("Error"));

    provider =
        new UpdatableGraffitiProvider(fn, () -> Optional.of(defaultGraffiti));
    assertThat(provider.get()).hasValue(defaultGraffiti);
  }
```

validator/client/src/test/java/tech/pegasys/teku/validator/client/restapi/apis/DeleteGraffitiTest.java
```diff
@@ -0,0 +1,118 @@
...
+  void shouldSuccessfullyDeleteGraffiti() throws IOException {
+    final Validator validator = new Validator(publicKey, NO_OP_SIGNER, Optional::empty);
+    when(keyManager.getValidatorByPublicKey(any())).thenReturn(Optional.of(validator));
+    graffitiManager.deleteGraffiti(any());
+
+    handler.handleRequest(request);
+
+    assertThat(request.getResponseCode()).isEqualTo(SC_NO_CONTENT);
+    assertThat(request.getResponseBody()).isNull();
+  }
```
lucassaldanha: I don't understand what is being tested here. Why are we calling the `graffitiManager` directly instead of checking that the `DeleteGraffiti` is correctly calling the `graffitiManager`?

validator/client/src/test/java/tech/pegasys/teku/validator/client/restapi/apis/SetGraffitiTest.java
```diff
@@ -0,0 +1,143 @@
...
+    request.setRequestBody(graffiti);
+
+    final Validator validator = new Validator(publicKey, NO_OP_SIGNER, Optional::empty);
+    when(keyManager.getValidatorByPublicKey(any())).thenReturn(Optional.of(validator));
+
+    handler.handleRequest(request);
+
+    assertThat(request.getResponseCode()).isEqualTo(SC_NO_CONTENT);
+    assertThat(request.getResponseBody()).isNull();
+  }
```
lucassaldanha: Similar to the test on deleteGraffiti. We need to check that the POST API method is calling the graffitiManager.setGraffiti(..) method. Otherwise, we are not testing anything.

CHANGELOG.md
```diff
@@ -19,5 +19,6 @@ the [releases page](https://github.com/Consensys/teku/releases).
 - Updated the bootnodes for Chiado and Gnosis networks.
 - Added hidden option `--Xp2p-dumps-to-file-enabled` to enable saving p2p dumps to file.
 - Appends consensus layer (CL) and execution layer (EL) clients' information to the validator graffiti. Check [documentation](https://docs.teku.consensys.io/development/reference/cli#validators-graffiti-client-append-format) for available configuration options.
+- Added GET, POST and DELETE `/eth/v1/validator/{pubkey}/graffiti` validator APIs.
```
lucassaldanha: I suggest updating to something like:
```
- Added support for [Graffiti management](https://ethereum.github.io/keymanager-APIs/?urls.primaryName=dev#/Graffiti) in the Key Manager API
```

validator/api/src/main/java/tech/pegasys/teku/validator/api/UpdatableGraffitiProvider.java
```diff
@@ -0,0 +1,47 @@
...
+  public Optional<Bytes32> get() {
+    return getFromStorage().or(defaultProvider::get);
+  }
+
+  private Optional<Bytes32> getFromStorage() {
+    try {
+      return storageProvider.get();
+    } catch (Throwable e) {
+      return Optional.empty();
+    }
```
lucassaldanha: Catching Throwable and swallowing it without handling it properly is a bad idea. This includes all JVM Errors (e.g. OutOfMemory, StackOverflow, etc.). So we will be preventing the program from handling errors properly.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,96 @@
...
+  public Optional<Bytes32> getGraffiti(final BLSPublicKey publicKey)
+      throws GraffitiManagementException {
+    final Path filePath = directory.resolve(resolveFileName(publicKey));
+    if (!filePath.toFile().exists()) {
+      return Optional.empty();
+    }
+
+    try {
+      return Optional.of(GraffitiParser.loadFromFile(filePath));
+    } catch (GraffitiLoaderException | IllegalArgumentException e) {
```
lucassaldanha, line 87: NIT: missing `final`

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public void deleteGraffiti(final BLSPublicKey publicKey) throws GraffitiManagementException {
+    final Path file = directory.resolve(resolveFileName(publicKey));
+    if (!file.toFile().exists()) {
+      return;
+    }
+
+    try {
+      Files.delete(file);
+    } catch (IOException e) {
```
lucassaldanha, line 72: NIT: missing final

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,96 @@
...
+      throw new IllegalArgumentException(
+          String.format(
+              "'%s' converts to %s bytes. Input must be 32 bytes or less.",
+              strippedGraffiti, graffitiSize));
+    }
+
+    try {
+      final Path file = directory.resolve(resolveFileName(publicKey));
+      Files.writeString(file, strippedGraffiti);
+    } catch (IOException e) {
```
lucassaldanha, line 57: NIT: missing final

validator/api/src/main/java/tech/pegasys/teku/validator/api/UpdatableGraffitiProvider.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override
+  public Optional<Bytes32> get() {
+    return getFromStorage().or(defaultProvider::get);
+  }
+
+  private Optional<Bytes32> getFromStorage() {
+    try {
+      return storageProvider.get();
+    } catch (Throwable e) {
```
lucassaldanha: NIT: missing final

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
```diff
@@ -0,0 +1,96 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.bls.BLSPublicKey;
+import tech.pegasys.teku.service.serviceutils.layout.DataDirLayout;
+
+public class GraffitiManager {
```
lucassaldanha, line 27: One thing that we need to consider is the fact that we can have multiple requests trying to modify the same resource (the graffiti file) at the same time. We need synchronize access to the file resource.

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -153,8 +157,23 @@ public static ValidatorClientService create(
...
     final ForkProvider forkProvider = new ForkProvider(config.getSpec(), genesisDataProvider);
 
-    final ValidatorLoader validatorLoader = createValidatorLoader(services, config, asyncRunner);
     final ValidatorRestApiConfig validatorApiConfig = config.getValidatorRestApiConfig();
+    final Optional<GraffitiManager> graffitiManager =
+        Optional.ofNullable(
+            validatorApiConfig.isRestApiEnabled()
+                ? new GraffitiManager(services.getDataDirLayout())
+                : null);
+    final ExceptionThrowingFunction<BLSPublicKey, Optional<Bytes32>> updatableGraffitiProvider =
```
lucassaldanha: I believe this would be simpler if `GrafitiManagerException was a RuntimeException`. Then you can deal with it only in `UpdatableGraffitiProvider`.

validator/api/src/main/java/tech/pegasys/teku/validator/api/UpdatableGraffitiProvider.java
```diff
@@ -0,0 +1,47 @@
...
+      return Optional.empty();
+    }
+  }
+
+  public Optional<Bytes32> getWithThrowable() throws Throwable {
```
lucassaldanha: NIT: I am not sure I like this name. Maybe something like `getUnsafe()` (similar to other methods like `toArrayUnsafe()`).

In any case, it would be a good idea to add a javadoc explaining it.

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
courtneyeh: N/A

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
courtneyeh: N/A

validator/api/src/main/java/tech/pegasys/teku/validator/api/GraffitiManager.java
courtneyeh: N/A

validator/api/src/main/java/tech/pegasys/teku/validator/api/UpdatableGraffitiProvider.java
courtneyeh: N/A

lucassaldanha (approved):
LGTM

PR #8219, opened 2024-04-18
Error prone `final` - ethereum module

gfukushima:
## PR Description
This is PR is part of a series where the aim is to add missing final to enable the MethodInputParametersMustBeFinal error prone validation.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/Predicates.java
```diff
@@ -42,29 +42,31 @@ public Predicates(final SpecConfig specConfig) {
...
+      final UInt64 activationEpoch, final UInt64 exitEpoch, final UInt64 epoch) {
     return activationEpoch.compareTo(epoch) <= 0 && epoch.compareTo(exitEpoch) < 0;
   }
 
   public boolean isValidMerkleBranch(
       final Bytes32 leaf,
       final SszBytes32Vector branch,
       final int depth,
-      int index,
+      final int index,
```
tbenr: Is this actually a legit case in which we don't want it to be `final`?
since this is a primitive type I don't see why we shouldn't allow it.

tbenr (approved):
Mind-bending. LGTM but I'd like to discuss one point.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/Predicates.java
tbenr: if the error-prone plugin will be designed to enforce it, maybe we can have a custom annotation signalling we actually want it to be mutable (may only for primitive types)

rolfyone: yeh does look like we need index to mutate through this fn... pretty unusual!

rolfyone: looking at where its used though, i dont htink we actually need to persist the change up, so this one might be fine...

gfukushima: For this case (and I think I found another somewhere else so far 2 out of thousands) there is mutation needed within the method. 
Although we could let these be exceptions, people who aren't familiar with the code and its workflow might have no context of why and how this is okay (in this case) to be used.
A defensive mutable copy to mutate with is probably a safer approach, IMO.

gfukushima: Worse case scenario there's always the good old `@SuppressWarnings` but I'd like to avoid using those as much as we can.

tbenr: thinking about it again, I see the safety\readability argument to make a local copy of it, which signal that the value we end up having there will be different. For my preference i'd call it `localIndex` or `indexTmp`, or call the param `startingIndex` and leave the local as `index`.

gfukushima: Variables renamed

Issue #7955, opened 2024-02-06
Finalized state from `debug/state` API is incorrect on epochs with empty slots in the end


zilm13:
### Description
We return `finalizedAnchor` in `debug/state` API, though if this state is associated with block which is not the last in finalized epoch, finalized state should be transitioned through empty slots to the end of the epoch. It sounds logical for this kind of `id` and `state` API endpoint and is default behavior of the other clients.

lucassaldanha:
Internal discussion: https://consensyssoftware.atlassian.net/wiki/spaces/PEG/pages/399517614166/Starting+from+transitioned+AnchorPoint+discussion

gfukushima:
I have tried to replicate this with Kurtosis and we're reporting the same the same state as lighthouse.

Managed to get an empty slot on the last slot of the epoch.
 
Logs from Lighthouse:
```
2024-10-21 14:41:14 Oct 21 04:41:14.000 INFO Synced                                  slot: 351, block:    …  empty, epoch: 10, finalized_epoch: 8, finalized_root: 0xdcca…9ac0, exec_hash: 0xa4dd…3578 (verified), peers: 1, service: slot_notifier
2024-10-21 14:41:14 Oct 21 04:41:14.000 INFO Ready for Electra                       info: ensure the execution endpoint is updated to the latest Electra/Prague release, service: slot_notifier
2024-10-21 14:41:20 Oct 21 04:41:20.001 INFO Prepared beacon proposer                parent_root: 0x511ff680ea5cf3d0fd292adc146c7b834cd231d0858f1c0adc3713b0b946f489, validator: 42, prepare_slot: 353, service: beacon
2024-10-21 14:41:20 Oct 21 04:41:20.003 ERRO No valid eth1_data votes, `votes_to_consider` empty, outcome: casting `state.eth1_data` as eth1 vote, genesis_time: 1729481456, earliest_block_timestamp: 1729481436, lowest_block_number: 0, service: deposit_contract_rpc
2024-10-21 14:41:20 Oct 21 04:41:20.016 INFO Signed block published to network via HTTP API, publish_delay_ms: 0, slot: 352
2024-10-21 14:41:20 Oct 21 04:41:20.025 INFO Valid block from HTTP API               slot: 352, proposer_index: 43, root: 0x8d9b…bacf, block_delay: 15.60722ms
[...]

PR #8221, opened 2024-04-18
Starting from transitioned initial anchor with tests

zilm13:
## PR Description
Based on #7981 
Work in progress.

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/tekuv1/beacon/GetStateByBlockRoot.java
```diff
@@ -69,6 +69,7 @@ public void handleRequest(RestApiRequest request) throws JsonProcessingException
     request.header(Header.CACHE_CONTROL, CACHE_NONE);
 
     final String blockId = request.getPathParameter(PARAMETER_BLOCK_ID);
+    // FIXME: not sure. Are we ok to return here transitioned state on keyword finalized?
```
gfukushima, line 72: my understand is that this is ok since this endpoint should mimic the same behaviour as the `/eth/v2/debug/beacon/states/{state_id}` but allowing to search for block. We're using the same blockroot for the transitioned state so it looks logical to me that we returned the transitioned state

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
```diff
@@ -375,6 +376,7 @@ private Optional<ReorgContext> computeReorgContext(
     return optionalReorgContext;
   }
 
+  // TODO: Test the consequences of having transitioned chainHead
```
gfukushima, line 379: why would we transition the chainHead? I thought we were only going to transition finalized states/block

rolfyone:
#7955 solution will resolve this.

PR #8230, opened 2024-04-19
Consolidate `SszKZGProof/KZGProof` and `SszKZGCommitment/KZGCommitment`

StefanBratanov:
## PR Description
I think there is overhead with having both `SszKZGProof` and `KZGProof` and also `SszKZGCommitment` and `KZGCommitment`. My thinking is that we can do the same we do for `Blob` type, just expose `Bytes` and `Bytes48` in `KZG` interface and then directly wrap it in the SSZ object. Regardless if we use local or builder at some point, we have to make it into a SSZ object, so can just skip the middle step. Also makes it consistent with the `Blob` type.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


StefanBratanov:
Will close this because of all the tests that are breaking, number of changed files and not an obvious performance improvement

PR #8235, opened 2024-04-22
Enable ErrorProne `FieldCanBeFinal`

gfukushima:
## PR Description
This PR enables the error prone check `FieldCanBeFinal` and apply some of the patches found by the error prone plugin.

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/PrettyPrintCommand.java
```diff
@@ -59,24 +59,24 @@ public class PrettyPrintCommand implements Callable<Integer> {
...
 
   @Parameters(
       index = "0",
       paramLabel = "<MILESTONE>",
       description = "Milestone to use when decoding objects.",
       arity = "1")
-  @SuppressWarnings("FieldMayBeFinal")
-  private SpecMilestone milestone = SpecMilestone.ALTAIR;
+  @SuppressWarnings("FieldCanBeFinal")
+  private final SpecMilestone milestone = SpecMilestone.ALTAIR;
```
tbenr: you should remove `final` here?

gfukushima: Think the patch was applied here because of the outdated annotation.
I probably forgot to remove this when I fixed the annotation. Thanks for catching this!

teku/src/main/java/tech/pegasys/teku/cli/subcommand/internal/validator/tools/ConsoleAdapter.java
```diff
@@ -16,10 +16,18 @@
 import java.io.Console;
 
 public class ConsoleAdapter {
-  private Console console = System.console();
+  private final Console console = System.console();
 
+  @SuppressWarnings("SystemConsoleNull") // https://errorprone.info/bugpattern/SystemConsoleNull
```
gfukushima, line 21: This came up after adding `final` to the console variable above.
It has nothing to do with the `FieldCanBeFinal` validation itself but it triggered another EP check and mentioned in the comment.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/CommitteeAssignment.java
gfukushima, line 1: I might push a follow up PR to make this class a record.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/internal/validator/tools/ConsoleAdapter.java
lucassaldanha: Although this change makes sense given the imminent JDK 22 update, I would like to know that we have tested this using JDK 22 to be sure. Otherwise, we may think that everything is fine but won't find out until we upgrade our testing to JDK 22.

lucassaldanha (approved):
LGTM. Just a comment re `Console.class.getMethod("isTerminal").invoke(console);` testing.

teku/src/main/java/tech/pegasys/teku/cli/subcommand/internal/validator/tools/ConsoleAdapter.java
gfukushima: tested and feature still works as expected

Issue #8149, opened 2024-03-26
process_operations


rolfyone:
- check process_operations
 - check Deposits https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.0/specs/electra/validator.md#deposits
 - voluntary exits https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.0/specs/electra/validator.md#voluntary-exits

rolfyone:
deposits change is in #8250

PR #8236, opened 2024-04-22
EIP7251 voluntary exit validation

rolfyone:
- also had to change voluntary exit random function, which affected a bunch of tests.
  - Reformatted json files to multiline while I was going, but the unit test ones are specifically doing text comparisons, so i had to leave them the same format.

  partially addresses #8149

## Documentation

## Changelog


rolfyone:
im going to re-open this without the dataStructureUtil change

PR #8237, opened 2024-04-22
EIP7251 - voluntary exits operation validation

rolfyone:
Extended the voluntary exits validation for electra, where voluntary exits can't be processed if there are pending withdrawals in the queue.

Structured this validation to be individually tested but didn't refactor the rest. also looks like a lot of the specLogic classes have heavily duplicated code but this is not really new, so I left it.

partially addresses #8149

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

PR #8247, opened 2024-04-24
Fix SszCollectionSchema.createFromElements()

Nashatyrev:
## PR Description

Fix the case when for a custom SSZ Vector or List implementations `SszCollectionSchema.createFromElements()` returns wrong `SszVector/SszList` implementation instead of a custom class. 
The same issue appeared for `SszByteVector.fromBytes()` 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #8249, opened 2024-04-25
DepositWithIndex: replace inheritance with composition

Nashatyrev:
## PR Description

When fixing #8247 got some unexpected `DepositProviderTest` failures. 
It turned out that POJO `DepositWithIndex` is inherited from SSZ `Deposit`, added to an `SszList<Deposit>` and casted back to `DepositWithIndex` in the test. 

This PR 
- Makes an explicit composition in `DepositWithIndex` instead of inheritance 
- Makes explicit methods to get either `SszList<Deposit>` SSZ structure  or plain `List<DepositWithIndex>` from `DepositProvider`

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
@@ -109,7 +109,7 @@ public class DepositProvider
               }
 
               depositNavigableMap.put(deposit.getIndex(), deposit);
-              depositMerkleTree.pushLeaf(deposit.getData().hashTreeRoot());
+              depositMerkleTree.pushLeaf(deposit.getDeposit().getData().hashTreeRoot());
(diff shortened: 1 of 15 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/DepositProvider.java
```diff
@@ -302,16 +307,17 @@ private SszList<Deposit> getDepositsWithProof(
         .stream()
         .map(
             deposit -> {
```
zilm13: could be `depositWithIndex` for clarity like in previous change

zilm13 (approved):
LGTM, just one small nit, but I'd also remove all `extends Deposit` from the code as it's now doesn't have any sense and could confuse.

Nashatyrev:
> remove all `extends Deposit` from the code

Good point 👍 Done

PR #8250, opened 2024-04-25
7251 - Added deposit processing

rolfyone:
- needed to update genesis handling to process deposits straight away, so that we have active validators and can start.

 moved `switchToCompoundingValidator` because we need it elsewhere so stateMutators makes more sense

 The genesis generator may not be completely correct, i've consumed all balances for any that had min activation balance, but im not sure that this will always be correct. It doesnt look like this is in spec anywhere, I'll raise it with the maxEB chan.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/genesis/GenesisGenerator.java
```diff
@@ -103,6 +109,36 @@ public void updateCandidateState(
...
+      final SpecConfig specConfig,
+      final SpecVersion genesisSpec,
+      final int validatorIndex) {
+    final BeaconStateMutatorsElectra mutatorsElectra =
+        new BeaconStateMutatorsElectra(
+            specConfig, genesisSpec.miscHelpers(), genesisSpec.beaconStateAccessors());
+    final List<PendingBalanceDeposit> pendingBalanceDeposits =
+        stateElectra.getPendingBalanceDeposits().stream().toList();
+
+    final UInt64 depositsAmount =
```
gfukushima: we only retrieve one deposit here, right? could this be named `depositAmount` then?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -307,4 +314,65 @@ public void processDepositReceipts(
...
+      final MutableBeaconState state,
+      final BLSPublicKey pubkey,
+      final Bytes32 withdrawalCredentials,
+      final UInt64 amount) {
+    final Validator validator = getValidatorFromDeposit(pubkey, withdrawalCredentials);
+    LOG.debug("Adding new validator with index {} to state", state.getValidators().size());
+    state.getValidators().append(validator);
+    final int validatorIndex = validatorsUtil.getValidatorIndex(state, pubkey).orElseThrow();
+
+    state.getBalances().appendElement(UInt64.ZERO);
```
gfukushima: it might make no difference here because we're creating a `stateElectra` on top of the mutable `state`, but shouldn't we be using the `stateElectra` to append the balance?

gfukushima (commented):
partially posting this so we can move faster with the review, still finishing the review on this PR

gfukushima (approved):
lgtm, asked some question to clarify some of the doubts I had when I was reviewing this.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/genesis/GenesisGenerator.java
rolfyone: specifically pending balance amount for the specified validator index, maybe more than 1 deposit contributing but i think depositAmount would be easier to read, will fix.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
rolfyone: the balance is just on the base class, but yep can do either... maybe its easier to see what's going on if we're only mutating the electra state, i'll update.

PR #8258, opened 2024-04-29
Don't create folders in current directory when dumping p2p data is disabled

### LINKED ISSUE #8256: Unable to create gossip_messages directory to save gossip messages. Disabling saving debug data to file.
### Description
During startup, Teku tries to create the following folders in the current directory instead of using the configured `data-base-path`.

```
./gossip_messages
./gossip_messages/decoding_error
./gossip_messages/rejected
./invalid_blocks
```

This is probably due to the recent introduction of the [DebugDataDumper ](https://github.com/Consensys/teku/blob/730358fc0ca258b3d16fd9790e5eaf60adb05471/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/DebugDataDumper.java#L51).

### Steps to Reproduce (Bug)
1. [download](https://artifacts.consensys.net/public/teku/raw/names/teku.zip/versions/24.4.0/teku-24.4.0.zip) and unzip teku 24.4.0
2. navigate to any folder you have read only access (i.e., `/home`)
3. launch teku via absolute or relative path
4. wait a few seconds
5. stop teku (CTRL+C)
6. check the log for the error message `ERROR | DebugDataDumper | Unable to create gossip_messages directory to save gossip messages. Disabling saving debug data to file.`
7. navigate to a temp folder you have write permissions
8. launch teku via absolute or relative path
9. wait a few seconds
10. stop teku (CTRL+C)
11. verify debug folders are created inside the current folder

**Expected behavior:**

Teku complies with the `data-base-path` in the configuration file.

**Actual behavior:**

Teku is trying to use the current folder for debugging purposes.

**Frequency:**

Always

### Versions
* Software version: `teku/v24.4.0/linux-x86_64/-privatebuild-openjdk64bitservervm-java-17`
* Java version: `OpenJDK Runtime Environment (build 17.0.10+7-Ubuntu-122.04.1)`
* OS Name & Version: `Ubuntu 22.04.4 LTS`

StefanBratanov:
## PR Description
The `NoOpDataDumper` had the following constructor:

```java
public NoOpDebugDataDumper() {
  super(Path.of("."));
}
```

It tries to create the p2p dumping directories in the current directory, when the p2p dumping is disabled (this is the default). There is no reason to do that. It can cause error in the logs if not possible or it will create empty directories, which will never have any files in them.

Changed the `DebugDataDumper` to be an interface and added `NOOP` implementation. Most of the changes are renamings.

## Fixed Issue(s)
fixes #8256 

## Documentation

## Changelog


tbenr (approved):
LGTM.
But, not sure about adding "P2P" as a prefix, because:
1. we could use the dumper for other purpose, more detached from P2P
2. `saveInvalidBlockToFile` could be already called within sync logic or even when importing a block via beacon API call (not related to P2P)

but the option itself is named `--Xp2p`.. so ok it's up to you :)

StefanBratanov:
> LGTM. But, not sure about adding "P2P" as a prefix, because:
> 
> 1. we could use the dumper for other purpose, more detached from P2P
> 2. `saveInvalidBlockToFile` could be already called within sync logic or even when importing a block via beacon API call (not related to P2P)
> 
> but the option itself is named `--Xp2p`.. so ok it's up to you :)

I just did it for consistency. We have for example `P2PLogger` which also have `onInvalidBlock` which is called in same situations.

rolfyone:
I probably wouldnt have renamed it, but im not so worried that we'll raise another PR...

Issue #8148, opened 2024-03-26
Withdrawals Processing


rolfyone:
Update withdrawals computation.

rolfyone:
Part of block processing
 - `process_withdrawals`
 - `process_execution_payload` 
 
Make sure Execution_payload is updated
 - `process_execution_payload`

PR #8261, opened 2024-04-29
7251 - Update withdrawals operation for maxEB

rolfyone:
- refactored current withdrawals into a class
 - the class structure allows us to return the count of partial withdrawals, which will let us do electra `process_withdrawals`
 - added electra handling into that class
 - added tests, had to fix a BlockStateUtilTest because it didn't account for effective balance limitations.

 partially addresses #8148, specifically get_expected_withdrawals is complete.

## Documentation

## Changelog


lucassaldanha (approved):
LGTM.

It would be good to include a fully withdrawable validator in one of the test cases in ExpectedWithdrawalsTest to ensure we never break that path unknowingly.

PR #8263, opened 2024-04-30
Updated electra fork upgrade logic

lucassaldanha:
Reference: https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/fork.md?plain=1#L164-L180


rolfyone (approved):
LGTM

PR #8268, opened 2024-04-30
Use isGreaterThanOrEqualTo when checking committee index

jtraglia:
## PR Description

Noticed what I believe to be two minor accidents.

* A duplicate statement which had no effect.
* An incorrect `committeeIndex` check.

For the second one, since it's an index, the max value is `committeeCountPerSlot - 1`, right?

## Documentation

## Changelog


tbenr:
omg thanks @jtraglia

tbenr (approved):

PR #8271, opened 2024-04-30
Use correct end index for subList calls

jtraglia:
## PR Description

Noticed another little mistake. The `toIndex` value for these `List.subList` calls are wrong. The end is exclusive.

> Returns a view of the portion of this list between the specified fromIndex, inclusive, and toIndex, exclusive.

https://docs.oracle.com/javase/8/docs/api/java/util/List.html#subList-int-int-

image

I'm aware all of this code is a work in progress 😄 I know there will be problems.

Also, I replaced the two loops with more java-esque for-each patterns. This was just a nit.

## Documentation

## Changelog


zilm13 (approved):
LGTM
Thank you, nice catch
`Returns a view of the portion of this list between the specified fromIndex, inclusive, and toIndex, exclusive. `

rolfyone:
I noticed it in a new spot yesterday during testing.... thanks for cleaning this up!

PR #8273, opened 2024-04-30
Default to currentEpoch in findEarliestExitEpoch

jtraglia:
## PR Description

Noticed another little problem. In `findEarliestExitEpoch`, if none of the validators have an exit epoch set, `lastExitEpoch` should be the current epoch (plus one), not zero (plus one). Right?

```python
exit_epochs = [v.exit_epoch for v in pre.validators if v.exit_epoch != FAR_FUTURE_EPOCH]
if not exit_epochs:
    exit_epochs = [get_current_epoch(pre)]
earliest_exit_epoch = max(exit_epochs) + 1
```

Primarily:

* Update `findEarliestExitEpoch` to be a little closer to the spec.
  * Replace for loop with stream, map, ...
  * Use `FAR_FUTURE_EPOCH` instead of `UInt64::MAX_VALUE`.
  * Replace `isLessThan` with `!equals`, like the spec.
 
Also:

* Rename `exits` to `withdrawalRequests`. I missed this when renaming variables.
* Use lambda operation in `getPendingBalanceToWithdraw` (just a nit).

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/forktransition/ElectraStateUpgrade.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/forktransition/ElectraStateUpgrade.java
@@ -119,7 +119,7 @@ public class ElectraStateUpgrade implements StateUpgrade<BeaconStateDeneb> {
               state.setDepositBalanceToConsume(UInt64.ZERO);
               state.setExitBalanceToConsume(
                   beaconStateAccessors.getActivationExitChurnLimit(state));
-              state.setEarliestExitEpoch(findEarliestExitEpoch(state));
+              state.setEarliestExitEpoch(findEarliestExitEpoch(state, epoch));
               state.setConsolidationBalanceToConsume(
                   beaconStateAccessors.getConsolidationChurnLimit(state));
               state.setEarliestConsolidationEpoch(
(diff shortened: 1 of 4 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/forktransition/ElectraStateUpgrade.java
```diff
@@ -153,11 +152,10 @@ public BeaconStateElectra upgrade(final BeaconState preState) {
...
-    final SszList<Validator> validators = state.getValidators();
     UInt64 lastExitEpoch = UInt64.ZERO;
-    for (int i = 0; i < validators.size(); i++) {
-      final UInt64 exitEpoch = validators.get(i).getExitEpoch();
-      if (exitEpoch.isLessThan(UInt64.MAX_VALUE)) {
+    for (final Validator validator : state.getValidators()) {
+      final UInt64 exitEpoch = validator.getExitEpoch();
+      if (!exitEpoch.equals(FAR_FUTURE_EPOCH)) {
         lastExitEpoch = lastExitEpoch.max(exitEpoch);
       }
```
jtraglia, line 160: Actually, I think I just noticed a problem in `findEarliestExitEpoch`. If none of the validators have an exit epoch, `lastExitEpoch` should be the current epoch, not zero.

rolfyone (approved):
i had a go at the test - happy now - ta

PR #8276, opened 2024-05-01
Make graffiti append able to add codes to 28 bytes graffiti

### LINKED ISSUE #8275: Append graffiti should be able to add clients info when only 4 bytes are available
### Description
With #8107 and #8175 we have added a feature to append CL and EL clients info to the graffiti when producing the block. Current minimal requirements are 5 bytes: `<SPACE>` + `<CL CODE>` + `<EL CODE>`. It would be good to keep this option available but also add 4 bytes option without a space, so users could utilize 28 bytes for their own graffiti while still be able to share clients info.

zilm13:
## PR Description
Drops space in this case.
Not very elegant but it works. Not sure how to make it more beautiful, if you have any ideas, I'm open!

## Fixed Issue(s)

Fixes #8275 
## Documentation

## Changelog


zilm13:
Docs PR https://github.com/Consensys/doc.teku/pull/572

StefanBratanov (approved):
LGTM

PR #8277, opened 2024-05-01
Refactor `GraffitiBuilder`

StefanBratanov:
## PR Description
Didn't change the `GraffitiBuilder` tests to not modify the current behaviour, but thought the code could be refactored a bit.

I don't see value in `ClientVersion.UNKNOWN`, can just rely on volatile `Optional`. This also simplifies `ExecutionClientVersionProvider` while keeping the same functionality. Also don't see value in `safeConsensusCode` since we set it explictily.

Added `void onExecutionClientVersionNotAvailable()` to `ExecutionClientVersionChannel` in order to print graffiti in cases where `engine_getClientVersion` is not available.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/test/java/tech/pegasys/teku/ethereum/executionclient/ExecutionClientVersionProviderTest.java
+++ b/ethereum/executionclient/src/test/java/tech/pegasys/teku/ethereum/executionclient/ExecutionClientVersionProviderTest.java
@@ -19,7 +19,6 @@ import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 
 import java.util.List;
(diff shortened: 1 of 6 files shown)


ethereum/executionclient/src/test/java/tech/pegasys/teku/ethereum/executionclient/ExecutionClientVersionProviderTest.java
```diff
@@ -44,32 +48,39 @@ public void setUp() {
...
 
-    new ExecutionClientVersionProvider(
-        executionLayerChannel, publishChannel, ClientVersion.UNKNOWN);
-    verify(publishChannel).onExecutionClientVersion(ClientVersion.UNKNOWN);
+    final ExecutionClientVersionProvider executionClientVersionProvider =
+        new ExecutionClientVersionProvider(
+            executionLayerChannel, publishChannel, consensusClientVersion);
+
+    executionClientVersionProvider.onAvailabilityUpdated(false);
+    executionClientVersionProvider.onAvailabilityUpdated(true);
```
zilm13: these two lines are not needed, we check it on class creation

zilm13 (approved):
Great refactoring, looks more clear and undestandable now.
Just one small nit.

ethereum/executionclient/src/test/java/tech/pegasys/teku/ethereum/executionclient/ExecutionClientVersionProviderTest.java
StefanBratanov: > these two lines are not needed, we check it on class creation

You are right, the "in middle" verification that no notification has been sent is in another test.

PR #8278, opened 2024-05-01
GetBlobSidecars API update

### LINKED ISSUE #8191: [beacon-api] add more context to getBlobSidecars
ref. https://github.com/ethereum/beacon-APIs/pull/441

Note these are ultimately become required fields, but are initially just optional in the spec to reduce breakage potential.

we don't want to fail if the fields are not present.

gfukushima:
## PR Description
This PR adds the metadata fields to the result of the endpoint `/eth/v1/beacon/blob_sidecars/{block_id}` (GetBlobSidecars) updated here: https://github.com/ethereum/beacon-APIs/pull/441

## Fixed Issue(s)

Fixes #8191 

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetBlobSidecars.java
```diff
@@ -86,21 +90,26 @@ public void handleRequest(final RestApiRequest request) throws JsonProcessingExc
...
       final SchemaDefinitionCache schemaCache) {
     final DeserializableTypeDefinition<BlobSidecar> blobSidecarType =
         SchemaDefinitionsDeneb.required(schemaCache.getSchemaDefinition(SpecMilestone.DENEB))
             .getBlobSidecarSchema()
             .getJsonTypeDefinition();
-    return SerializableTypeDefinition.<List<BlobSidecar>>object()
+    return SerializableTypeDefinition.<BlobSidecarsAndMetaData>object()
         .name("GetBlobSidecarsResponse")
-        .withField("data", listOf(blobSidecarType), Function.identity())
+        .withOptionalField("version", MILESTONE_TYPE, (b) -> Optional.of(b.getMilestone()))
```
StefanBratanov, line 101: I don't think we need the `OptionalField` here for version, execution_optimistic and finalized. Can make it similar to `GetBlock`. The SerializableTypeDefinition is only used for serializing our fields which we know that are always present in our implementation.

data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
```diff
@@ -96,7 +151,11 @@ public BlobSidecarSelector slotSelectorForAll(final UInt64 slot) {
             .getAllBlobSidecars(slot, indices)
             .thenApply(
                 blobSidecars ->
-                    blobSidecars.isEmpty() ? Optional.empty() : Optional.of(blobSidecars));
+                    blobSidecars.isEmpty()
+                        ? Optional.empty()
+                        : addMetaData(
+                            //TODO we don't care about metadata here...need to find a way to get around the lack of a block root?
```
StefanBratanov: Yeah, we use slotSelectorForAll only in our internal `/teku/v1/beacon/blob_sidecars/{slot}` which doesn't have any of the metadata fields, so I would just add a comment why we don't care about the block root and just default it to ZERO as you do.

StefanBratanov (commented):
LGTM overall with couple of comments.

StefanBratanov:
I think also worth adding to `CHANGELOG` we provide those fields now in the `eth/v1/beacon/blob_sidecars/{block_id}` endpoint

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/GetBlobSidecars.java
lucassaldanha: The OpenAPI spec has defined them as optional. So I believe we need to use `withOptionalField()` to ensure we are compliant, don't we?

StefanBratanov: Yes, if we want to deserialize as well. Otherwise, not needed, since we always add them, but not fuzzed about it. Probably defining them as optional makes it more clear.

lucassaldanha: I was thinking from the perspective of the open api spec that we generate. If we want to match the "official" spec, we would need the optional field so we would generate the open api schema spec as optional as well.

StefanBratanov: Yeah, that makes sense.

lucassaldanha:
The changes look good for me. The only thing missing is some tests around the metadata logic in `BlobSidecarSelectorFactory` (could probably re-use some of the test cases on `BlobSidecarSelectorFactoryTest`).

gfukushima:
Changelog and tests added

data/provider/src/test/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactoryTest.java
```diff
@@ -66,9 +67,9 @@ public void headSelector_shouldGetHeadBlobSidecars()
     when(client.getBlobSidecars(blockAndState.getSlotAndBlockRoot(), indices))
         .thenReturn(SafeFuture.completedFuture(blobSidecars));
 
-    final Optional<List<BlobSidecar>> result =
+    final Optional<BlobSidecarsAndMetaData> result =
         blobSidecarSelectorFactory.headSelector().getBlobSidecars(indices).get();
-    assertThat(result).hasValue(blobSidecars);
+    assertThat(Optional.of(result.get().getData())).hasValue(blobSidecars);
```
lucassaldanha: NIT: I don't understand why we need to wrap it into an Optional for this assertion. Can't we do it like:
```
assertThat(result.get().getData()).isEqualTo(blobSidecars);
```

lucassaldanha: This applies to other assertions in this class as well but I won't add a comment on each one :)

data/provider/src/test/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactoryTest.java
```diff
@@ -220,4 +235,55 @@ public void shouldLookForBlobSidecarsOnlyAfterDeneb(
...
+    when(client.getBlobSidecars(block.getSlot(), indices))
+        .thenReturn(SafeFuture.completedFuture(blobSidecars));
+    when(client.isChainHeadOptimistic()).thenReturn(false);
+
+    final Optional<BlobSidecarsAndMetaData> result =
+        blobSidecarSelectorFactory.slotSelector(block.getSlot()).getBlobSidecars(indices).get();
+
+    assertThat(result).isNotEmpty();
+    final BlobSidecarsAndMetaData blobSidecarsAndMetaData = result.get();
+    assertThat(blobSidecarsAndMetaData.isFinalized()).isEqualTo(true);
```
lucassaldanha: NIT: you can use `.isTrue()` instead of `.isEqualTo(true)`.

data/provider/src/test/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactoryTest.java
```diff
@@ -220,4 +235,55 @@ public void shouldLookForBlobSidecarsOnlyAfterDeneb(
...
+    when(client.isChainHeadOptimistic()).thenReturn(false);
+    when(client.getBlobSidecars(block.getSlotAndBlockRoot(), indices))
+        .thenReturn(SafeFuture.completedFuture(blobSidecars));
+
+    final Optional<BlobSidecarsAndMetaData> result =
+        blobSidecarSelectorFactory.slotSelector(block.getSlot()).getBlobSidecars(indices).get();
+
+    assertThat(result).isNotEmpty();
+    final BlobSidecarsAndMetaData blobSidecarsAndMetaData = result.get();
+    assertThat(blobSidecarsAndMetaData.isFinalized()).isEqualTo(false);
```
lucassaldanha: NIT: can use `.isFalse()` :)

lucassaldanha (approved):
LGTM. Just added a few NITs.

PR #8281, opened 2024-05-02
EIP-7251: Implementing process_consolidation block processing function

### LINKED ISSUE #8150: process_operations process_consolidation
https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.0/specs/electra/validator.md#consolidations

lucassaldanha:
## PR Description
Spec reference: https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#new-process_consolidation

Summary:
- Implemented `process_consolidation` function on `BlockProcessorCapella`
- Created method `assertCondition(..)` to make it easier to write assertions as their python spec `assert` counterpart.
- Updated reference-tests OperationTestExecution to include consolidation operations
- Verified `process_consolidation` operations with reference tests for minimal (mainnet tests seem to have an [issue](https://github.com/ethereum/consensus-specs/issues/3739))

## Fixed Issue(s)
fixes #8150

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/BlockProcessor.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/BlockProcessor.java
@@ -53,8 +53,7 @@ import tech.pegasys.teku.spec.logic.versions.bellatrix.block.OptimisticExecution
 
 public interface BlockProcessor {
 
-  Optional<OperationInvalidReason> validateAttestation(
-      final BeaconState state, final AttestationData data);
+  Optional<OperationInvalidReason> validateAttestation(BeaconState state, AttestationData data);
 
   BeaconState processAndValidateBlock(
       SignedBeaconBlock signedBlock,
@@ -124,10 +123,10 @@ public interface BlockProcessor {
       throws BlockProcessingException;
 
   void processDepositWithoutCheckingMerkleProof(
-      final MutableBeaconState state,
-      final Deposit deposit,
-      final Optional<Object2IntMap<BLSPublicKey>> maybePubkeyToIndexMap,
-      final boolean signatureAlreadyVerified);
+      MutableBeaconState state,
+      Deposit deposit,
+      Optional<Object2IntMap<BLSPublicKey>> maybePubkeyToIndexMap,
+      boolean signatureAlreadyVerified);
 
   void processVoluntaryExits(
       MutableBeaconState state,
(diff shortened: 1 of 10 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -335,6 +344,105 @@ public void processDepositReceipts(
...
+          sourceValidator.getExitEpoch().equals(FAR_FUTURE_EPOCH),
+          "Source validator in consolidation is exiting");
+      assertCondition(
+          targetValidator.getExitEpoch().equals(FAR_FUTURE_EPOCH),
+          "Target validator in consolidation is exiting");
+
+      // Consolidations must specify an epoch when they become valid; they are not valid before then
+      assertCondition(
+          currentEpoch.isGreaterThanOrEqualTo(consolidation.getEpoch()),
+          "Consolidation must have activation epoch in the future");
```
rolfyone: This is more like
"Consolidation must have an activation epoch no greater than current epoch"

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -335,6 +344,105 @@ public void processDepositReceipts(
...
+          "Source validator in consolidation must have execution withdrawal credentials");
+      assertCondition(
+          predicatesElectra.hasExecutionWithdrawalCredential(targetValidator),
+          "Target validator in consolidation must have execution withdrawal credentials");
+      // Verify the same withdrawal address
+      assertCondition(
+          sourceValidator
+              .getWithdrawalCredentials()
+              .equals(targetValidator.getWithdrawalCredentials()),
+          "Source and Target validators in consolidation must have the same withdrawal credentials");
```
rolfyone: this is more complicated, comparing the eth address portion of it only - it may be worth having a helper that just extracts that (we might have one...)
the reason is target may be already 0x02 so in that case this check would fail

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/BlockProcessor.java
```diff
@@ -177,6 +179,10 @@ void processExecutionLayerWithdrawalRequests(
       final Supplier<ValidatorExitContext> validatorExitContextSupplier)
       throws BlockProcessingException;
 
+  void processConsolidations(
+      MutableBeaconState state, final SszList<SignedConsolidation> consolidations)
```
rolfyone: we dont usually go final on interfaces, but either way consistency probably good.... both final or both not...

rolfyone (approved):
LGTM

PR #8284, opened 2024-05-02
Fix `toWithdraw` calculation in `processExecutionLayerWithdrawalRequests`

jtraglia:
## PR Description

Reviewed `processExecutionLayerWithdrawalRequests` and noticed a mistake that I made. My bad.

The calculation for `toWithdraw` used `min` when it should have used `minus`. See the spec:

```python
to_withdraw = min(
    state.balances[index] - MIN_ACTIVATION_BALANCE - pending_balance_to_withdraw,
    amount
)
```

Also, the `throws BlockProcessingException` is unnecessary.

## Documentation

## Changelog


lucassaldanha (approved):

lucassaldanha:
Nice catch! No need to apologize, it happens more often than you'd think :D

lucassaldanha:
Nice catch! No need to apologize, it happens more often than you'd think :D

PR #8289, opened 2024-05-05
Retry loading of external validators if failed

### LINKED ISSUE #7586: Retry loading of external validators if failed
### Description
When Teku fails to load external validator keys it prints the only message and gives up
```
{"@timestamp":"2023-10-09T01:32:28,197","level":"WARN","thread":"validator-async-0","class":"ValidatorClientService","message":"Failed to load public keys from URL http://some_url","throwable":""}
```
As a result service continue to run without noticeable issues while validator duties are not performed. Also `local_validator_count` gauge is not created which fails Grafana to display validator stats.

Retrying of loading external validator keys with error message repeated could solve the issue. Plus metrics gauge could be registered in advance, so Grafana dashboard is rendered correctly.

courtneyeh:
fixes #7586 

## Documentation

## Changelog


validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalUrlKeyReader.java
```diff
@@ -0,0 +1,86 @@
...
+    final String[] keys =
+        readUrl()
+            .exceptionallyCompose(
+                ex -> {
+                  if (ex instanceof MalformedURLException) {
+                    return SafeFuture.failedFuture(
+                        new InvalidConfigurationException(
+                            "Failed to load public keys from invalid URL: " + url, ex));
+                  }
+                  return retry();
```
lucassaldanha: This structure is going to cause it to run twice in sequence (one because of `readUrl()` and another because of `retry()`). We could just rely on `asyncRunner.runWithRetry()` to run the calls.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalUrlKeyReader.java
```diff
@@ -0,0 +1,86 @@
...
+  @VisibleForTesting
+  SafeFuture<String[]> retry() {
+    return asyncRunner
+        .runWithRetry(
+            () -> {
+              STATUS_LOG.failedToLoadPublicKeysFromUrl(url);
+              return readUrl();
+            },
+            FIXED_DELAY,
+            59)
```
lucassaldanha: NIT: can we move this into a constant?

teku/src/main/java/tech/pegasys/teku/cli/subcommand/VoluntaryExitCommand.java
```diff
@@ -423,7 +423,9 @@ private void initialise() {
             new RejectingSlashingProtector(),
             slashingProtectionLogger,
             new PublicKeyLoader(
-                externalSignerHttpClientFactory, validatorConfig.getValidatorExternalSignerUrl()),
+                externalSignerHttpClientFactory,
+                validatorConfig.getValidatorExternalSignerUrl(),
+                asyncRunner),
```
lucassaldanha: It does not make sense to have the retry behavior on the VoluntaryExitCommand. This is an interactive command and the user won't sit there waiting for this to eventually run... It should fail early in this case.

lucassaldanha: Also, when we fail to load the keys for VoluntaryExitCommand, the command just sits there and it looks like it is frozen (even though the retries are happening in the background, there is no feedback for the user). There is no output to the console.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalUrlKeyReader.java
```diff
@@ -0,0 +1,86 @@
...
+    this.mapper = mapper;
+    this.asyncRunner = asyncRunner;
+  }
+
+  public Stream<BLSPublicKey> readKeys() {
+    final String[] keys =
+        readUrl()
+            .exceptionallyCompose(
+                ex -> {
+                  if (ex instanceof MalformedURLException) {
```
lucassaldanha: We can validate this when we are instantiating `ExternalUrlKeyReader`. It would simplify a lot this block of handling if the URL is a valid URL, etc.

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ExternalUrlKeyReaderTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.jupiter.api.Test;
+import tech.pegasys.teku.bls.BLSPublicKey;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.async.StubAsyncRunner;
+import tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException;
+import tech.pegasys.teku.infrastructure.time.StubTimeProvider;
+import tech.pegasys.teku.spec.TestSpecFactory;
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+
+class ExternalUrlKeyReaderTest {
```
lucassaldanha, line 45: None of these tests check that when `readUrl()` fails, we are triggering the `retry()` logic. This an important flow to test.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalUrlKeyReader.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private final URL url;
+  private final ObjectMapper mapper;
+  private final Optional<AsyncRunner> asyncRunner;
+
+  @VisibleForTesting
+  public ExternalUrlKeyReader(
+      final String urlString, final ObjectMapper mapper, final AsyncRunner asyncRunner) {
+    this(urlString, mapper, Optional.of(asyncRunner));
+  }
```
lucassaldanha, line 45: This constructor does not seem necessary. Just use `Optional.of(asyncRunner)` wherever this is being called.

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/ExternalUrlKeyReader.java
```diff
@@ -0,0 +1,90 @@
...
+                ex ->
+                    SafeFuture.failedFuture(
+                        new InvalidConfigurationException(
+                            "Failed to load public keys from URL: " + url, ex)))
+            .join();
+    return Arrays.stream(keys).map(key -> BLSPublicKey.fromSSZBytes(Bytes.fromHexString(key)));
+  }
+
+  @VisibleForTesting
+  SafeFuture<String[]> getKeysWithRetry(final AsyncRunner asyncRunner) {
```
lucassaldanha, line 69: NIT: I feel like `getKeysWithRetry` and `readUrl` names should be similar, with the difference that one has the Retry word on it. Suggestion:
```
readKeysFromUrl(..)
readKeysFromUrlWithRetry(..)
```

validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/PublicKeyLoaderTest.java
```diff
@@ -121,16 +123,18 @@ void shouldHandleEmptyResponseFromUrl() throws IOException {
...
-        .isInstanceOf(InvalidConfigurationException.class)
-        .hasRootCause(exception);
+  void shouldThrowExceptionWhenUrlFailsToLoad() {
+    final InvalidConfigurationException cause = new InvalidConfigurationException("Unknown host");
+    final CompletionException exception = new CompletionException(cause);
+    final ExternalUrlKeyReader reader = mock(ExternalUrlKeyReader.class);
+    when(reader.readKeys()).thenThrow(exception);
+
+    assertThatThrownBy(() -> loader.getPublicKeys(List.of(urlSource), (key) -> reader))
+        .isEqualTo(exception);
```
lucassaldanha, line 133: I am confused by this test. It seems that all we want to test is that any exceptions bubble up to the caller of `PublicKeyLoader.getPublicKeys(..)`. If this is the case, we probably do not care about the exception type or cause. Might just use something simpler like:
```
final RuntimeException exception = new RuntimeException("Foo");
when(reader.readKeys()).thenThrow(exception);
```

Later we can test that the same exception is propagated up.

We could also rename the test to `shouldThrowExceptionWhenExternalUrlKeyReaderFailsUnexpectedly` to make it clear that this is a case where we do not handle the error.

rolfyone:
parking for now.

Issue #7965, opened 2024-02-08
[EIP-7549] Move committee index outside attestation


lucassaldanha:
### Description
Move the committee index field outside of the signed Attestation message to allow aggregation of equal consensus votes.

### References
- [EIP-7549](https://eips.ethereum.org/EIPS/eip-7549)
- [Consensus Spec PR](https://github.com/ethereum/consensus-specs/pull/3559)
- [Updated Consensus Spec](https://github.com/ethereum/consensus-specs/tree/dev/specs/_features/eip7549)

### Tasks
- [x] #8089 Refactor get_attesting_indices
- [x] #8026 Add/update attestations containers 
- [x] #8123 Add/update index attestations containers 
- [x] #8031 Update the attestation processing
- [x] #8156 Flattern aggregation bits
- [x] #8103 Update preset values
- [x] #8130 Update beacon block body

Once the previous done, we can start merging the changes in master while progressing on the following:

- [x] #8101 Update attestation production
- [x] #8027 Update attestation aggregation logic
~~#8270 Replace `committee_bits` with `committee_indices`~~
- [x] #8099 Update the aggregation gossip topic
- [x] #8100 Update the attestation gossip topic
- [x] #8340 Move `committee_bits` property to the end
- [x] #8029 Add the `GetAttestations` v2 APIs (attestations pool)
- [x] #8424 Add the `PostAttestation` v2 APIs (attestations pool)
- [x] #8264 Add the `GetBlockAttestations` v2 APIs (block's attestations)
- [x] #8122 Add the `GetAttesterSlashings` v2 API (get attester slashing)
- [x] #8425 Add the `PostAttesterSlashing` v2 API (post attester slashing)
- [x] #8420 Add the `GetAggregateAttestation` v2 API
- [x] #8439 Add the `PostAggregateAttestation` v2 API

### Tech debt:
- [x] ELECTRA milestone over `AggregateAttestationValidatorTest` and implement `ElectraAttestationValidatorTest`

### Possible evolutions
- [x] #8554 Separate type for on-chain attestation aggregates
- [x] #8555 Separate type for unaggregated network attestations

PR #8300, opened 2024-05-08
7549 - VC aggregation

tbenr:
- changes `ValidatorApiChannel::createAggregate` by adding an optional `committeeIndex` to allow BN to know about committee to aggregate for. (Implemented only for in-process VC)
- changes aggregation iteration (`MatchingDataAttestationGroup`) to apply the committee filtering if required.
- enables several test to run over ELECTRA too

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/ValidatorLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/ValidatorLogger.java
@@ -170,7 +170,7 @@ public class ValidatorLogger {
     return blockRoots.stream().map(LogFormatter::formatHashRoot).collect(Collectors.joining(", "));
   }
 
-  public void aggregationSkipped(final UInt64 slot, final int committeeIndex) {
+  public void aggregationSkipped(final UInt64 slot, final UInt64 committeeIndex) {
     log.warn(
         ColorConsolePrinter.print(
             PREFIX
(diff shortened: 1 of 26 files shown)


infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/ValidatorLogger.java
```diff
@@ -170,7 +170,7 @@ private String formatBlockRoots(final Set<Bytes32> blockRoots) {
     return blockRoots.stream().map(LogFormatter::formatHashRoot).collect(Collectors.joining(", "));
   }
 
-  public void aggregationSkipped(final UInt64 slot, final int committeeIndex) {
+  public void aggregationSkipped(final UInt64 slot, final UInt64 committeeIndex) {
```
rolfyone, line 173: we only see this used in test code in this PR which is curious... if its only testing code we should probably move it, and Im less concerned with the interface change, otherwise i think committeeIndex would have been better to stay as int :)

rolfyone (approved):
LGTM, one nit about something only used in testing code

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/ValidatorLogger.java
tbenr: I had a back and forth among int<->UInt64. I landed there just to maintain similarity to `validatorIndex`.
And i also saw some UInt64 related to committteeIndex in some API params.

tbenr: why you say it is only used in testing? I see [this](https://github.com/ConsenSys/teku/blob/ac5cc170654a039c32b824edfb2c2f4d5a49ffa8/validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/attestations/AggregationDuty.java#L149) :)

rolfyone: weird that interface changes and calling point doesnt...

PR #8302, opened 2024-05-08
provide ssz bitvector OR function

tbenr:
provides bitvecor OR, useful for 7549

## Documentation

## Changelog

### DIFF
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/BitvectorImpl.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/BitvectorImpl.java
@@ -77,7 +77,7 @@ class BitvectorImpl {
   }
 
   public BitvectorImpl or(final BitvectorImpl other) {
-    if (other.getSize() > getSize()) {
+    if (other.getSize() != getSize()) {
       throw new IllegalArgumentException(
           "Argument bitfield size is greater: " + other.getSize() + " > " + getSize());
     }
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/BitvectorImpl.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/BitvectorImpl.java
@@ -81,7 +81,7 @@ class BitvectorImpl {
       throw new IllegalArgumentException(
           "Argument bitfield size is greater: " + other.getSize() + " > " + getSize());
     }
-    BitSet newData = (BitSet) this.data.clone();
+    final BitSet newData = (BitSet) this.data.clone();
     newData.or(other.data);
     return new BitvectorImpl(newData, size);
   }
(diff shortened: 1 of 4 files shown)


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/BitvectorImpl.java
```diff
@@ -76,6 +76,16 @@ public List<Integer> getSetBitIndices() {
     return data.stream().boxed().toList();
   }
 
+  public BitvectorImpl or(final BitvectorImpl other) {
+    if (other.getSize() > getSize()) {
+      throw new IllegalArgumentException(
+          "Argument bitfield size is greater: " + other.getSize() + " > " + getSize());
+    }
+    BitSet newData = (BitSet) this.data.clone();
```
zilm13: final please

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/SszBitvectorImpl.java
```diff
@@ -89,6 +89,11 @@ public SszBitvector withBit(final int i) {
     return new SszBitvectorImpl(getSchema(), value.withBit(i));
   }
 
+  @Override
+  public SszBitvector or(SszBitvector other) {
```
zilm13: final please

zilm13 (approved):
LGTM just nit finals

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/collections/impl/BitvectorImpl.java
```diff
@@ -76,6 +76,16 @@ public List<Integer> getSetBitIndices() {
     return data.stream().boxed().toList();
   }
 
+  public BitvectorImpl or(final BitvectorImpl other) {
+    if (other.getSize() > getSize()) {
```
rolfyone: i'd think we want to make sure the sizes match exactly if we're doing a bitwise or, otherwise we'll be needing to decide if the stragglers are left or right etc... `other.size() != getSize`

tbenr: mm yes but this mimics the `BitlistImpl` or which does the same. We may say that this is a vector so it could be more strict but I'm more on the side to have a similar behavior.

rolfyone: its probably bad behaviour to mimic... you can mis-match and do bitwise or partially.... future @tbenr problem to find i guess...

zilm13: Is this feature really used somewhere? (I mean or with smaller list/vector)
If yes we could split it to several safer steps, but it could affect performance

tbenr: k lets be more stringent and maybe open in the future.
@zilm13 we won't (at least for 7549)

rolfyone: i really do think we're better off making this one right rather than copying a bad idea - especially on a bitwise or... at least then we know why its a greater than if it causes a break...

rolfyone (approved):
other than the discussion about sizings, looks good. won't block it but i do feel like we should check it fully on a bitwise op.

PR #8303, opened 2024-05-08
7549 aggregating pool filter phase0 att in electra

tbenr:
We don't want to select phase0 attestations when electra enables.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidatableAttestation.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/attestation/ValidatableAttestation.java
@@ -154,7 +154,7 @@ public class ValidatableAttestation {
     saveCommitteeShufflingSeed(state);
     // The committees size is only required when the committee_bits field is present in the
     // Attestation
-    if (attestation.getCommitteeBits().isPresent()) {
+    if (attestation.requiresCommitteeBits()) {
       saveCommitteesSize(state);
     }
   }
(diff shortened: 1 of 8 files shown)


rolfyone:
so this is a step forward for sure, now can create blocks in a single node configuration...

in a multi-node minimal setup, aggregates are getting rejected over the wire with messages like:
```
09:53:46.354 ERROR - Validator   *** Produced invalid aggregate for slot 47: Rejecting aggregate because attester is not in committee. Should have been one of [82, 26, 7, 88]
09:53:46.354 ERROR - Validator   *** Produced invalid aggregate for slot 47: Rejecting aggregate because attester is not in committee. Should have been one of [82, 26, 7, 88]
09:53:46.355 ERROR - Validator   *** Produced invalid aggregate for slot 47: Rejecting aggregate because attester is not in committee. Should have been one of [82, 26, 7, 88]
09:53:46.356 INFO  - Validator   *** Published aggregate          Count: 1, Slot: 47, Root: 133cd269c309cc4aa48f380081ccc43b006b48be26acac14d33bd661ba1b9846
```
not sure if its my config...

Definitely better than previous where blocks were not being produced even in a single node, but I figured i'd give a bigger statement of progress in case it helps narrow things down...

rolfyone:
ok so 
 - 2 nodes on minimal-electra kind of setup is ok up to and including 63 validators, at 64 the aggregator issues start, it might be a committee calculation issue...
 - single minimal-electra node with 64 keys does actually show the issue - i had previously used less validators so didn't see it.

tbenr:
Ye the aggregation during block production is not done yet (currently it is still doing phase0 aggregation which can produce broken blocks)

A PR for this is planned :)

rolfyone (approved):

PR #8304, opened 2024-05-09
Provide committeesSize to MatchingDataAttestationGroup

StefanBratanov:
## PR Description

- `committeesSize` is required for EIP-7549 aggregations when adding attestations coming from an imported block to track the attestations which are already onchain, so adding a way to retrieve it using the state
- Also there could be cases when `committeesSize` is not present in a `ValidatableAttestation`, so having a fallback to the state ensures we don't fail.

## Fixed Issue(s)
related to #7965 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
@@ -135,7 +135,7 @@ public class AggregatingAttestationPool implements SlotEventsChannel {
    *     href="https://eips.ethereum.org/EIPS/eip-7549">EIP-7549</a>
    */
   private MatchingDataAttestationGroup getOrCreateAttestationGroup(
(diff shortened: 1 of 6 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -113,14 +131,36 @@ public synchronized void add(final ValidatableAttestation attestation) {
...
+  private Supplier<Int2IntMap> getCommitteesSizeSupplierUsingTheState(
+      final AttestationData attestationData) {
+    return () -> {
+      final SlotAndBlockRoot slotAndBlockRoot =
+          new SlotAndBlockRoot(attestationData.getSlot(), attestationData.getBeaconBlockRoot());
+      final BeaconState state =
+          recentChainData
+              .retrieveStateAtSlot(slotAndBlockRoot)
+              // the attestation pool flow is entirely synchronous so joining here
+              .join()
```
rolfyone, line 159: eep here be dragons...

tbenr:
merged via https://github.com/Consensys/teku/pull/8305

PR #8312, opened 2024-05-14
add null validations when deserializing DepositRequestV1

StefanBratanov:
## PR Description
Better error messaging if there are `DepositRequestV1` missing fields. Avoids confusing errors like:

```
java.lang.NullPointerException: Cannot invoke "org.apache.tuweni.bytes.Bytes.size()" because "bytes" is null
``` 

## Fixed Issue(s)
N/A

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #8317, opened 2024-05-16
Fix state retrieval for calculating the committees size during aggregating

StefanBratanov:
## PR Description
Use the target root when retrieving the relevant state. Querying state by the attestation slot and block root is not correct.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8323, opened 2024-05-17
Do not add Electra attestations if committees size is not retrievable

StefanBratanov:
## PR Description
Filter out Electra attestations for which we can't retrieve the committees size from the relevant state

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.statetransition.attestation;
 
+import com.google.common.base.Suppliers;
 import it.unimi.dsi.fastutil.ints.Int2IntMap;
 import java.util.Collection;
 import java.util.HashMap;
@@ -113,12 +114,14 @@ public class AggregatingAttestationPool implements SlotEventsChannel {
             .getCommitteesSize()
             .<Supplier<Int2IntMap>>map(committeesSize -> () -> committeesSize)
             .orElseGet(() -> getCommitteesSizeSupplierUsingTheState(attestationData));
-    final boolean add =
-        getOrCreateAttestationGroup(attestationData, committeesSizeSupplier).add(attestation);
-    if (add) {
-      updateSize(1);
+    if (attestationIsApplicable(attestation.getAttestation(), committeesSizeSupplier)) {
+      final boolean add =
+          getOrCreateAttestationGroup(attestationData, committeesSizeSupplier).add(attestation);
+      if (add) {
+        updateSize(1);
+      }
     }
-    // Always keep the latest slot attestations so we don't discard everything
+    // Always keep the latest slot attestations, so we don't discard everything
     int currentSize = getSize();
     while (dataHashBySlot.size() > 1 && currentSize > maximumAttestationCount) {
       LOG.trace("Attestation cache at {} exceeds {}, ", currentSize, maximumAttestationCount);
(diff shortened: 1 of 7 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -128,39 +128,49 @@ public synchronized void add(final ValidatableAttestation attestation) {
...
-              .getBlockStateIfAvailable(targetRoot)
-              .orElseThrow(
-                  () ->
-                      new IllegalStateException(
-                          "No state available for attestation with target root " + targetRoot));
-      return spec.getBeaconCommitteesSize(state, attestationData.getSlot());
-    };
+    final Bytes32 targetRoot = attestationData.getTarget().getRoot();
+    LOG.debug(
+        "Committees size was not readily available for attestation with target root {}. Will attempt to retrieve it using the relevant state.",
```
rolfyone: i'd probably change this to just be something short like
`Committee size was not found via target root {} for attestation at slot {}`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -128,39 +126,48 @@ public synchronized void add(final ValidatableAttestation attestation) {
     }
   }
 
+  private Optional<Int2IntMap> getCommitteesSize(final Attestation attestation) {
+    if (attestation.requiresCommitteeBits()) {
+      final AttestationData attestationData = attestation.getData();
+      LOG.debug(
+          "Committees size was not found for target root {} for attestation at slot {}. Will attempt to retrieve it using the relevant state.",
```
rolfyone: i think id still delete 'will attempt...'

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -107,18 +106,17 @@ public AggregatingAttestationPool(
...
-      updateSize(1);
-    }
-    // Always keep the latest slot attestations so we don't discard everything
+    final Optional<Int2IntMap> committeesSize =
+        attestation.getCommitteesSize().or(() -> getCommitteesSize(attestation.getAttestation()));
+    getOrCreateAttestationGroup(attestation.getAttestation(), committeesSize)
+        .ifPresent(
+            attestationGroup -> {
+              final boolean add = attestationGroup.add(attestation);
+              if (add) {
```
rolfyone: mm i get what you're doing, but i wonder if either 'added' or 'success' or just calling the add in the if fn would be cleaner... `add` is almost right but may be cleaner in another way to read.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroup.java
```diff
@@ -56,14 +55,14 @@ public class MatchingDataAttestationGroup implements Iterable<ValidatableAttesta
   private final Spec spec;
   private Optional<Bytes32> committeeShufflingSeed = Optional.empty();
   private final AttestationData attestationData;
-  private final Supplier<Int2IntMap> commiteesSizeSupplier;
+  private final Optional<Int2IntMap> committeesSize;
```
rolfyone, line 58: i think `maybeCommitteeSize` would be cleaner

rolfyone (approved):

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
StefanBratanov: I deleted the whole log and moved it to `getOrCreateAttestationGroup`. It is now:
```
LOG.debug("Committees size was not available for attestation at slot {} and block root {}. Will NOT add this attestation to the pool.",attestationData.getTarget().getRoot(), attestationData.getBeaconBlockRoot());
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroup.java
StefanBratanov: Left it as it is because we use that name in other places (where it is still `Optional`)

PR #8326, opened 2024-05-22
Refactor flaky test 

gfukushima:
## PR Description
This PR aims top fix one of our most flaky test by adding a small logic to clean up the temp directory right after loading the validators. This is a small tweak to avoid the flakness caused in the windows run.

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ValidatorLoaderTest.java
+++ b/validator/client/src/test/java/tech/pegasys/teku/validator/client/loader/ValidatorLoaderTest.java
@@ -809,7 +809,7 @@ class ValidatorLoaderTest {
         stream.sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
       }
     }
-    OwnedValidators validators = validatorLoader.getOwnedValidators();
+    final OwnedValidators validators = validatorLoader.getOwnedValidators();
     assertThat(validators.getValidatorCount()).isEqualTo(0);
     checkGraffitiProviderTypes(validators.getValidators(), FileBackedGraffitiProvider.class);
   }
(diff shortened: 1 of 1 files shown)


zilm13:
I'm not sure it could help, `@TempDir` injects different random paths in every method, is not reused and is cleaned by default when the test is finished

zilm13:
I was trying to reproduce it without any luck, because I'm not on Windows, probably.
So from my understanding, what's happening in this test (I could be wrong, of course): 
- `MultithreadedValidatorLoader` got 2 tasks for keys running in executor service
- Bad key task throws and is catch by test, test ends
- Good key task is still running, holding file resource, `executorService.shutdownNow()` is not waiting for file release
- JUnit cleanup is started and as file resource is still busy, it fails

There is a thread here regarding this issue https://github.com/junit-team/junit5/issues/2811 

So from my understanding we will start failing in `finally` block since this change, but I could be wrong and chaos alteration will solve the issue.
What could we do (from my understanding, any option):
- run the test only with one, bad key
- use executorService.awaitTermination() with some timeout
- apply bad hack from JUnit thread

We could probably merge this PR as is and try one of my suggestion if this change is not helping

zilm13 (approved):

gfukushima:
Thanks for the feedback @zilm13, I did find the same thread when I was investigating the possible cause of this flaky test, figured I could add this to the test to justify why we're doing all this workaround for that specifically case. I tried to reproduce as well and like you had no success.

The intention of cleaning at the finally block is to try and get the directory clear as soon as possible so when the `@TempDir` logic executes the directory is cleaned and we avoid the exception.

I think from the approaches you suggested 1 and 2 are valid and I'd probably try them one at a time in case this approach doesn't solve the issue.

PR #8328, opened 2024-05-22
Fix `StartupLogConfigTest` test to work on other environments

Nashatyrev:
## PR Description

Fix `StartupLogConfigTest` test to work on other environments

The test might fail: 
- With a different default locale with a different decimal separator
- With a different -Xmx JVM option

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8332, opened 2024-05-23
Enhanced IPv6 support (libp2p)

### LINKED ISSUE #8069: Enhanced IPv6 support
### Description
Currently Teku will serve on ipv6 only when ipv4 is not enabled. But there is a common usage case to run both IPv4 and IPv6. Also we miss any configuration for it: setting different ports, advertising ipv6 etc. Check, for example, lighthouse configuration:
https://lighthouse-book.sigmaprime.io/advanced_networking.html#ipv6-support

#### Remaining Tasks:
- [x] return Optional<List<String>> in `DiscV5Service.getDiscoveryAddress`
- [x] When we are searching for peers and streaming known peers, we only use `nodeRecord.getTcpAddress()` to connect to them. We should also see how we can use `nodeRecord.getTcp6Address`.
- [x] Port forward the upd IPv6 port

StefanBratanov:
## PR Description
Add new options (hidden for now until discovery is adapted for dual stack as well):

- `--Xp2p-port-ipv6`
- `--Xp2p-advertised-port-ipv6`

Also enhanced `--p2p-interface` and `--p2p-advertised-ip` to accept multiple entries to support dual-stack (IPv4 and IPv6), but for the moment didn't add changes to description until we unhide the ipv6 feature entirely.

**Testing (dual stack)**
```
13:49:11.601 INFO  - Listening for connections on: /ip4/127.0.0.1/tcp/9000/p2p/16Uiu2HAm8vhQ3odDV2YkNFH5ZE2HmCEFX1Q83nSX3fLRroinQdtS,/ip6/2a01:4b00:875c:9500:d55c:71df:3af7:9f1f/tcp/9090/p2p/16Uiu2HAm8vhQ3odDV2YkNFH5ZE2HmCEFX1Q83nSX3fLRroinQdtS
```

Response from `eth/v1/node/identity`
```json
{
  "data": {
    "peer_id": "16Uiu2HAm8vhQ3odDV2YkNFH5ZE2HmCEFX1Q83nSX3fLRroinQdtS",
    "enr": "enr:-LK4QL4QeqrRSLtOvXSW4n1IstncRls_CPhhrBzPiOG4yAXlWtXq2GqAFXk2kC_Cxne7QH7F-mkjxvcnhI6zt8ySdHUEh2F0dG5ldHOIBgAAAAAAAACEZXRoMpBqlaGpBAAAAP__________gmlkgnY0gmlwhH8AAAGJc2VjcDI1NmsxoQLIl2CHsAAMhZN_Jmh_Tr3dM9huh4cYFN0_ItPq9XBwv4N0Y3CCIyiDdWRwgiMo",
    "p2p_addresses": [
      "/ip4/127.0.0.1/tcp/9000/p2p/16Uiu2HAm8vhQ3odDV2YkNFH5ZE2HmCEFX1Q83nSX3fLRroinQdtS",
      "/ip6/2a01:4b00:875c:9500:d55c:71df:3af7:9f1f/tcp/9090/p2p/16Uiu2HAm8vhQ3odDV2YkNFH5ZE2HmCEFX1Q83nSX3fLRroinQdtS"
    ],
    "discovery_addresses": [
      "/ip4/127.0.0.1/udp/9000/p2p/16Uiu2HAm8vhQ3odDV2YkNFH5ZE2HmCEFX1Q83nSX3fLRroinQdtS"
    ],
    "metadata": {
      "seq_number": "1",
      "attnets": "0x0600000000000000",
      "syncnets": "0x00"
    }
  }
}
```

## Fixed Issue(s)
related to #8069 

## Documentation

## Changelog

### DIFF
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.networking.p2p.libp2p;
 import static tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork.REMOTE_OPEN_STREAMS_RATE_LIMIT;
 import static tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork.REMOTE_PARALLEL_OPEN_STREAMS_COUNT_LIMIT;
 
+import com.google.common.base.Preconditions;
 import identify.pb.IdentifyOuterClass;
 import io.libp2p.core.Host;
 import io.libp2p.core.PeerId;
@@ -108,13 +109,16 @@ public class LibP2PNetworkBuilder {
 
     final List<Multiaddr> advertisedAddresses;
     final List<String> advertisedIps = config.getAdvertisedIps();
+    Preconditions.checkState(
+        advertisedIps.size() == 1 || advertisedIps.size() == 2,
+        "The configured advertised IPs must be either 1 or 2");
     if (advertisedIps.size() == 1) {
       advertisedAddresses =
           Collections.singletonList(
               MultiaddrUtil.fromInetSocketAddress(
                   new InetSocketAddress(advertisedIps.get(0), config.getAdvertisedPort()), nodeId));
     } else {
-      // IPv4 and IPv6
+      // IPv4 and IPv6 (dual-stack)
       advertisedAddresses =
           advertisedIps.stream()
               .map(
@@ -177,13 +181,16 @@ public class LibP2PNetworkBuilder {
   protected Host createHost(final PrivKey privKey, final List<Multiaddr> advertisedAddresses) {
     final String[] listenAddrs;
     final List<String> networkInterfaces = config.getNetworkInterfaces();
+    Preconditions.checkState(
+        networkInterfaces.size() == 1 || networkInterfaces.size() == 2,
+        "The configured network interfaces must be either 1 or 2");
     if (networkInterfaces.size() == 1) {
       final Multiaddr addr =
           MultiaddrUtil.fromInetSocketAddress(
               new InetSocketAddress(networkInterfaces.get(0), config.getListenPort()));
       listenAddrs = new String[] {addr.toString()};
     } else {
-      // IPv4 and IPv6
+      // IPv4 and IPv6 (dual-stack)
       listenAddrs =
           networkInterfaces.stream()
               .map(
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
@@ -37,6 +37,7 @@ import java.time.Duration;
 import java.util.List;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
+import tech.pegasys.teku.infrastructure.io.IPVersionResolver;
 import tech.pegasys.teku.infrastructure.time.TimeProvider;
 import tech.pegasys.teku.infrastructure.version.VersionProvider;
 import tech.pegasys.teku.networking.p2p.gossip.PreparedGossipMessageFactory;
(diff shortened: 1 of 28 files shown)


tbenr:
have you tried single stack ipV6 ? Discovery won't work but we should be able to run over outbound connections only.

StefanBratanov:
> have you tried single stack ipV6 ? Discovery won't work but we should be able to run over outbound connections only.

Yes. Discovery still works though. It doesn't support dual-stack but works with single stack IPv6.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
```diff
@@ -101,21 +103,48 @@ public P2PNetwork<Peer> build() {
...
-    NodeId nodeId = new LibP2PNodeId(host.getPeerId());
-    Multiaddr advertisedAddr =
-        MultiaddrUtil.fromInetSocketAddress(
-            new InetSocketAddress(config.getAdvertisedIp(), config.getAdvertisedPort()), nodeId);
+    final PrivKey privKey = privateKeyProvider.get();
+    final NodeId nodeId = new LibP2PNodeId(PeerId.fromPubKey(privKey.publicKey()));
+
+    final List<Multiaddr> advertisedAddresses;
+    final List<String> advertisedIps = config.getAdvertisedIps();
+    if (advertisedIps.size() == 1) {
```
tbenr, line 115: looks a bit weird this distinction between single and multi. So looking at the code we actually support only one or two, ips\ports. So maybe a checkState on size to be <= 2. Moreover seems like we do not guarantee alignment  ip,port: if IPv6 comes first, we may land on addresses=ipv6,ipv4 and ports=v4port,v6port

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfig.java
```diff
@@ -153,24 +177,29 @@ private String getSiteLocalOrUniqueLocalAddress(final boolean useIPV6)
...
+          if (IPVersionResolver.resolve(inetAddress) != ipVersion) {
+            // incompatible IP version
+            continue;
           }
-          // IPv6 (include only site local addresses or ULAs)
-          if (useIPV6
-              && inetAddress instanceof Inet6Address
-              && (inetAddress.isSiteLocalAddress() || isUniqueLocalAddress(inetAddress))) {
-            return inetAddress.getHostAddress();
+          if (ipVersion == IPVersion.IP_V4) {
```
tbenr: switch ? :)

tbenr (dismissed):
One thing that does not convince me is that we are leaking the feature via `--p2p-advertised-ips` and `--p2p-interfaces` aliases but then we hide the v6 options via `--X`. Maybe you're right but i'd like to discuss it

tbenr:
Sorry I didn't mean to "request changes", I pressed the wrong button on intelliJ :)

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
StefanBratanov: Agree about check state, will add it. The alignment is as follows:

If only one IP address is specified, the p2p-port would be used. There is no concept of v4 only port. v6port is used only when there are two IPs v4 and v6. I follow the lighthouse pattern: https://lighthouse-book.sigmaprime.io/advanced_networking.html#ipv6-support

tbenr: I see.. what if we want to run in v6 only?

StefanBratanov: Then it uses the p2p-port

tbenr: ok i see, sorry for the back and forth. But the end state, when we unhide the new `--X` port params, we might end up in a slightly confusing position. I feel like separating address and port param isn't a good thing after all. Maybe we can have a convo tomorrow on it.

StefanBratanov: Kk, let's have a talk for it tomorrow

StefanBratanov: Added `checkState` to be more clear

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfigTest.java
```diff
@@ -122,4 +129,37 @@ void checkDirectPeersConfigCreatedCorrectly() {
...
+    assertThatThrownBy(
+            () ->
+                NetworkConfig.builder()
+                    .advertisedIps(
+                        Optional.of(List.of("::", "2001:db8:3333:4444:5555:6666:7777:8888")))
+                    .build())
+        .isInstanceOf(InvalidConfigurationException.class)
+        .hasMessage(
+            "Expected an IPv4 and an IPv6 address for --p2p-advertised-ip but only [IP_V6] was set");
+  }
```
tbenr, line 179: is it tested the dual stack happy case too?

tbenr (approved):
LGTM. with a doubt

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfigTest.java
StefanBratanov: Good point, will add a test for it tomorrow

PR #8338, opened 2024-05-27
Change `skip_randao_verification` param to string

### LINKED ISSUE #8223: [beacon api] check if we need to fix anything in our definitions
https://github.com/ethereum/beacon-APIs/pull/444

I haven't read this closely, but if someone can take a look... its probably worth testing the query parameter.

gfukushima:
## PR Description
This PR changes the `skip_randao_verification` to comply with the changes in https://github.com/ethereum/beacon-APIs/pull/444 to a string flag like. The impact for us here is pretty much minimal since we ignore that param.
  
## Fixed Issue(s)

Fixes #8223 

## Documentation

## Changelog


infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/StringBasedPrimitiveTypeDefinition.java
```diff
@@ -105,6 +113,12 @@ public void serializeOpenApiType(final JsonGenerator gen) throws IOException {
     if (format.isPresent()) {
       gen.writeStringField("format", format.get());
     }
+    if (minLength.isPresent()) {
+      gen.writeNumberField("minLength", minLength.get());
+    }
+    if (maxLength.isPresent()) {
+      gen.writeNumberField("maxLength", maxLength.get());
+    }
```
rolfyone, line 121: what should be the effect here in the swagger definitions?

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
```diff
@@ -473,10 +489,24 @@ public EndpointMetaDataBuilder queryParam(final ParameterMetadata<?> parameterMe
       return this;
     }
 
+    public EndpointMetaDataBuilder queryParamAllowsEmpty(
+        final ParameterMetadata<?> parameterMetadata) {
```
tbenr, line 490: intelliJ suggesting to move to a dedicated method this snippet:

```
final String param = parameterMetadata.getName();
if (queryParams.containsKey(param)
|| queryParamsAllowEmpty.containsKey(param)
|| requiredQueryParams.containsKey(param)
|| queryListParams.containsKey(param)) {
throw new IllegalStateException("Query parameters already contains " + param);
}
```

makes sense to me

tbenr (commented):
Since we are not using it, and we never added this param to the previous API versions, I'd be tempted to solve this by removing the definition completely.

We added in V3 by just following the new API spec without really thinking about previous version.

The only side-effect is us printing a swagger which is not 100% compliant but I don't think it's a big deal in this case.

WDYT?
cc @rolfyone 

maybe we could add a note in the descriptions saying that Teku doesn't verify the field

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/StringBasedPrimitiveTypeDefinition.java
gfukushima: If you define this fields in a string they will be used to limit the input allowed in the swagger interface.

gfukushima:
tbh, I'm about not adding unnecessary code but since this is pretty much done and the changes required weren't too big I think there might be some value in keeping it in case any other param go through the same change.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/StringBasedPrimitiveTypeDefinition.java
rolfyone: yep just our swagger output is hand-done so if it was first use it would have needed enhancing...

rolfyone:
> We added in V3 by just following the new API spec without really thinking about previous version.
> 
> The only side-effect is us printing a swagger which is not 100% compliant but I don't think it's a big deal in this case.

really depends on the beacon-api spec.
If we define in beacon-api it's needed i'd be happy with listing it in the type but noting that we don't pay attention to the field...

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/paths/_eth_v3_validator_blocks_{slot}.json
```diff
@@ -49,6 +42,16 @@
         "example" : "1",
         "format" : "uint64"
       }
+    },{
+      "name" : "skip_randao_verification",
+      "allowEmptyValue" : true,
+      "in" : "query",
+      "schema" : {
+        "type" : "string",
+        "description" : "Skip verification of the `randao_reveal` value (currently ignored by Teku implementation). If this flag is set then the\n `randao_reveal` must be set to the point at infinity (`0xc0..00`). This query parameter\n  is a flag and does not take a value",
```
rolfyone: there's a must here - do we care?
i'm wondering if we check that at all, im guessing not. i'd just truncate before 'if this flag is set'
`Skip verification of the randao_reveal value. Ignored in the Teku implementation.`

infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/endpoints/EndpointMetadata.java
rolfyone: oh i thought this was used somewhere - technically a list could be expressed as 
`validator_id=[1,2,3,4]` or `validator_id=1&validator_id=2&validator_id=3&validator_id=4` if you really want to make your life difficult...
eg. https://medium.com/@AADota/spring-passing-list-and-array-of-values-as-url-parameters-1ed9bbdf0cb2

anyway - TL/DR, lets not do this.

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/paths/_eth_v3_validator_blocks_{slot}.json
gfukushima: I think we don't care based on this comment: https://github.com/ethereum/beacon-APIs/pull/222#issuecomment-1188455636. Cehcked the other impl blockV2 and blinded_block didn't care about this param.
Changed the description as you suggested.

rolfyone (approved):
LGTM i think... im not sure where discussions landed but i think we could live with this in its current form.

tbenr:
I don't oppose to merge it.

Just for the records: the reason why I'm not convinced is because in my opinion our swagger should reflect our functionalities. I think there is no point being compliant with swagger-doc and ignoring the underlying behaviour.
Advertising the param in swagger and saying "hey bare in mind we ignore it", is more confusing than not showing the param at all (which obviously means we don't support the param and we ignore it)

But this is just an opinion and it isn't super strong, so go for it!

gfukushima:
Thanks everyone for the input! 
To your point @tbenr, I'm slightly inclined towards having in swagger and making it clear to users that we ignore that parameter rather than not displaying it and having users confused and asking why we don't have that param specified in our API.

PR #8341, opened 2024-05-28
Fix BlocksByRange RPC requests when count is 0

StefanBratanov:
## PR Description
We had an error processing RPC blocks by range request when count is set to 0
```
Unhandled error while processing request /eth2/beacon_chain/req/beacon_blocks_by_range/2/ssz_snappy","throwable":" java.lang.ArithmeticException: uint64 underflow
 tech.pegasys.teku.infrastructure.unsigned.UInt64.minus(UInt64.java:203)
tech.pegasys.teku.infrastructure.unsigned.UInt64.minus(UInt64.java:157)
tech.pegasys.teku.spec.datastructures.networking.libp2p.rpc.BeaconBlocksByRangeRequestMessage.getMaxSlot(BeaconBlocksByRangeRequestMessage.java:70)
```

We technically should be able to handle such requests even if they are deemed useless. We have requests limiting so we are not prone to DOS attacks. 

Had to also fix `BlobSidecarsByRange` since `maxSlot` was not calculated correctly for `count == 0`

Added tests for both blocks and blob sidecars to ensure they both work.

Did some other small refactors along the way.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BeaconBlocksByRangeRequestMessage.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BeaconBlocksByRangeRequestMessage.java
@@ -67,7 +67,7 @@ public final class BeaconBlocksByRangeRequestMessage
   }
 
   public UInt64 getMaxSlot() {
-    return getStartSlot().plus(getCount().minus(1).times(getStep()));
+    return getStartSlot().plus(getCount().minusMinZero(1).times(getStep()));
   }
 
   @Override
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
@@ -244,12 +244,14 @@ public class BeaconBlocksByRangeMessageHandler
   }
 
   private class RequestState {
+
     private final UInt64 headSlot;
     private final ResponseCallback<SignedBeaconBlock> callback;
     private final UInt64 step;
     private final NavigableMap<UInt64, Bytes32> knownBlockRoots;
     private UInt64 currentSlot;
     private UInt64 remainingBlocks;
+
     private final AtomicInteger sentBlocks = new AtomicInteger(0);
 
     RequestState(
@@ -261,9 +263,7 @@ public class BeaconBlocksByRangeMessageHandler
         final ResponseCallback<SignedBeaconBlock> callback) {
       this.currentSlot = startSlot;
       this.knownBlockRoots = knownBlockRoots;
-      // Minus 1 to account for sending the block at startSlot.
-      // We only decrement this when moving to the next slot but we're already at the first slot
-      this.remainingBlocks = count.minusMinZero(ONE);
+      this.remainingBlocks = count;
       this.step = step;
       this.headSlot = headSlot;
       this.callback = callback;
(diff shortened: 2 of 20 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
```diff
@@ -183,7 +187,7 @@ private SafeFuture<RequestState> sendMatchingBlocks(
                       .map(MinimalBeaconBlockSummary::getSlot)
                       .orElse(ZERO);
               final NavigableMap<UInt64, Bytes32> hotRoots;
-              if (combinedChainDataClient.isFinalized(endSlot)) {
+              if (combinedChainDataClient.isFinalized(message.getMaxSlot())) {
```
gfukushima, line 185: I'm not quite sure this will end up in the same slot number as the previous approach used in line 168. Was this change intended?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BeaconBlocksByRangeRequestMessage.java
```diff
@@ -67,7 +67,7 @@ public UInt64 getStep() {
   }
 
   public UInt64 getMaxSlot() {
-    return getStartSlot().plus(getCount().minus(1).times(getStep()));
+    return getStartSlot().plus(getCount().minusMinZero(1).times(getStep()));
```
gfukushima, line 70: Alright, so after treading the spec I think this won't actually have any unexpected effects because step is always 1 so it should be fine. Maybe worth the change in the order so we don't trip into a issue if `step` for some reason gets re-used for another purpose.

StefanBratanov: Actually `final UInt64 endSlot = startSlot.plus(step.times(count)).minus(ONE);` won't be correct when step > 1, thus I replaced it with `message.getMaxSlot()` which is more accurate.

StefanBratanov: There is a test in `BeaconBlocksByRangeRequestMessageTest` which covers this method

StefanBratanov:
@gfukushima after some discussion with Enrico and Mehdi, we decided to change the logic to be able to handle such requests, just letting you know

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/BeaconBlocksByRangeRequestMessage.java
gfukushima: Should we use the same logic for `BlobSidecarsByRangeRequestMessage.getMaxSlot()` then?

StefanBratanov: There is no `step` there, so we are safe

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
mehdi-aouadi: The `getMaxSlot` gives a different result but is correct. IMU the `endSlot` calculation wasn't correct and returned `maxSlot + 1`. The first blocks (from the requested count) is always at the `startSlot` so we should first subtract 1 from the `count` to calculate the `endSlot`/`maxSlot`

storage/src/main/java/tech/pegasys/teku/storage/server/CombinedStorageChannelSplitter.java
```diff
@@ -235,7 +235,7 @@ public SafeFuture<List<SlotAndBlockRootAndBlobIndex>> getAllBlobSidecarKeys(fina
 
   @Override
   public SafeFuture<List<SlotAndBlockRootAndBlobIndex>> getBlobSidecarKeys(
-      UInt64 startSlot, UInt64 endSlot, UInt64 limit) {
+      UInt64 startSlot, UInt64 endSlot, long limit) {
```
mehdi-aouadi: Could add the missing finals here

mehdi-aouadi (approved):
LGTM

PR #8346, opened 2024-05-29
Improve electra attestation aggregation

tbenr:
- avoid recalculating starting positions when can be avoided (improved performance in calculation too)
- improved lookup to last bit set in committee bits
- avoid triggering ssz hashing on committee bits in `isSuperSetOf`
- rewrote `or`, now is ~4x faster and consumes far less memory
  - added SszBitlist `wrapBitSet` constructor to operate directly with a bitSet
- improved tests

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8348, opened 2024-05-30
fixed flaky test

rolfyone:
Looking at other solutions for this flaky test, it became apparent that the 'problem' is actually our use of system time, and we have a solution for that, but it's a bit more plumbing.

 - built `TimeProvider` into the epoch processor
 - created a static `SystemTimeProvider`
 - fixed test.

 I only built out a copy constructor for Capella epoch processing, which is what we're using in the test.
 It does mean we need to define the time provider in epoch processing, but it's arguably a cleaner solution than using system time natively.

 The magic sauce is using the `StubTimeProvider` in the test, but that took some refactoring.

## Documentation

## Changelog


gfukushima (approved):

mehdi-aouadi (approved):
LGTM

PR #8349, opened 2024-05-31
Enable `MethodInputParametersMustBeFinal`

gfukushima:
## PR Description
This PR add some missing `final` and enables the Error Prone `MethodInputParametersMustBeFinal` 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/datastructures/util/MerkleTree.java
+++ b/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/datastructures/util/MerkleTree.java
@@ -124,15 +124,15 @@ public class MerkleTree {
     if (depth == 0) {
       return zeroHashes.get(0);
     }
-    int mutableDepth = depth - 1;
-    Bytes32 deeperRoot = calcViewBoundaryRoot(mutableDepth, viewLimit);
+    final int deeperDepth = depth - 1;
+    final Bytes32 deeperRoot = calcViewBoundaryRoot(deeperDepth, viewLimit);
     // Check if given the viewLimit at the leaf layer, is root in left or right subtree
-    if ((viewLimit & (1 << mutableDepth)) != 0) {
+    if ((viewLimit & (1 << deeperDepth)) != 0) {
       // For the right subtree
-      return Hash.sha256(tree.get(mutableDepth).get((viewLimit >> mutableDepth) - 1), deeperRoot);
+      return Hash.sha256(tree.get(deeperDepth).get((viewLimit >> deeperDepth) - 1), deeperRoot);
     } else {
       // For the left subtree
-      return Hash.sha256(deeperRoot, zeroHashes.get(mutableDepth));
+      return Hash.sha256(deeperRoot, zeroHashes.get(deeperDepth));
     }
   }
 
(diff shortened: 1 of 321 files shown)


ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/datastructures/util/MerkleTree.java
```diff
@@ -113,25 +114,25 @@ public List<Bytes32> getProof(int itemIndex) {
...
   }
 
-  private Bytes32 calcViewBoundaryRoot(int depth, int viewLimit) {
+  private Bytes32 calcViewBoundaryRoot(final int depth, final int viewLimit) {
     if (depth == 0) {
       return zeroHashes.get(0);
     }
-    depth -= 1;
-    Bytes32 deeperRoot = calcViewBoundaryRoot(depth, viewLimit);
+    int mutableDepth = depth - 1;
```
rolfyone: wait so this is just off by 1? looks like this is not altered after decrementing.

infrastructure/ssz/src/test/java/tech/pegasys/teku/infrastructure/ssz/SszListHintsTest.java
```diff
@@ -93,18 +97,19 @@ <TElement extends SszData> void assertListElements(
...
 
     listVariants1.forEach(
         listVariant1 ->
-            listVariants2.forEach(
-                listVariant2 -> assertListEquals(type, listVariant1, listVariant2)));
+            listVariants2.forEach(listVariant2 -> assertListEquals(listVariant1, listVariant2)));
   }
 
   <TElement extends SszData> void assertListEquals(
-      SszListSchema<TElement, ?> type, SszList<TElement> list1, SszList<TElement> list2) {
```
rolfyone, line 107: assuming type wasn't referenced

rolfyone (approved):
minor bits...

infrastructure/ssz/src/test/java/tech/pegasys/teku/infrastructure/ssz/SszListHintsTest.java
gfukushima: it wasn't

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/datastructures/util/MerkleTree.java
gfukushima: you're right this is just single decrement operation done on top of the input param `depth`. New variable can actually be `final` as well

rolfyone: probably worth just looking where this is called too and see whether the depth passed in is just wrong lol

gfukushima: Mutation of this index is actually part of the recursive approach used in the this method, so I think we should keep it.

rolfyone: i do still think final, and potentially deeperRoot could be final also

so mutableDepth => final internalDepth
and deeperRoot -> final... at least on the code here...

Issue #4126, opened 2021-07-01
Validate locally created blocks prior to gossiping


ajsutton:
### Description
Currently Teku gossips blocks submitted to the publish block API immediately without checking validity.  We should only publish blocks we've validated.  As part of this we will need to ensure we publish blocks that are saved for later because the slot hasn't started yet (potentially because of clock skew with the validator client) which we don't appear to do currently.

lucassaldanha:
We need to check if this change affects the Builder flow.

rolfyone:
curious what we need here given block v3 etc @tbenr

PR #8352, opened 2024-06-04
Use v2 of the block publishing endpoints in VC

StefanBratanov:
## PR Description
Few changes:

- VC uses `v2` of the block publishing endpoints
- BN does not run only equivocation validation if the block has been created locally
- Changed acceptance test to run with a newer version (Deneb mainnet one) which supports `v2`
- Added `EQUIVOCATION` to the `BroadcastValidationLevel` enum to cater for only doing equivocations

## Fixed Issue(s)
related to https://github.com/Consensys/teku/issues/4126

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -636,7 +646,12 @@ public SafeFuture<SendSignedBlockResult> sendSignedBlock(
             maybeBlindedBlockContainer.getSlot());
     return blockPublisher
         .sendSignedBlock(
-            maybeBlindedBlockContainer, broadcastValidationLevel, blockPublishingPerformance)
+            maybeBlindedBlockContainer,
+            // no validation required for locally created blocks
+            isLocallyCreatedBlock(maybeBlindedBlockContainer)
+                ? BroadcastValidationLevel.NOT_REQUIRED
```
mehdi-aouadi: I'm a bit worried here since we override the request param which could be confusing if for some reason the caller would like to perform the check but ends up not having it. Maybe we should at least log something here?

mehdi-aouadi (approved):
LGTM
Just a nit about overriding the request parameter

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
tbenr: We could add a debug log, but I don't think is really important.

What instead is important is that, thinking more about it, I don't think this is what we want @StefanBratanov .

If the block is produced locally we want just to skip GOSSIP and CONSENSUS but i think that the last EQUIVOCATION step needs to be done if it has been requested (because it doesn't depend on our local behaviour).

tbenr: So I feel like `isLocallyCreatedBlock` should affect `BlockBroadcastValidatorImpl::buildValidationPipeline`.

Happy to discuss if it doesn't convince you.

tbenr: A simpler approach is to apply the override only if the default GOSSIP has been requested, but it wont boost the other validation levels.

StefanBratanov: I agree, it makes sense. I went for the simpler approach and added a CLI option for enabling this.

tbenr (approved):
LGTM

PR #8354, opened 2024-06-05
reverted validatorIndexCache

rolfyone:
Reverted the changes to validatorIndexCache, we'll have to take a look again for full electra support.

Although the intent was to only index finalized validators, the result was to constantly be doing StateTransitions somehow to get indices, and the outcome was a serious slow down in the important state transitions that happen constantly, so as a starting point, this will put us back to releasable code that will be ok for Deneb.

Before reverting, it seemed to process 1 state transition every 20 seconds or so, and even just rebuilding a handful of states was taking 10 minutes or longer, which is not really going to work in production.

We'll also need to build out testing for this once we re-implement. Archive storage is already slow due to epoch processing, we can't afford to take this kind of increased performance hit.

I'm guessing the tracker may still be part of an eventual solution, haven't removed at this point but i've cleaned it out functionally... i probably need to clean that before any merge...

`BlockProcessorElectra` was changed to not use the cache to find validator index, as it's a simple operation, and it was causing problems.

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -522,7 +522,17 @@ protected void addValidatorToRegistry(
...
     state.getValidators().append(validator);
-    final int validatorIndex = validatorsUtil.getValidatorIndex(state, pubkey).orElseThrow();
+    int validatorIndex = -1;
+    for (int i = state.getValidators().size() - 1; i >= 0; i--) {
+      if (state.getValidators().get(i).getPublicKey().equals(pubkey)) {
+        validatorIndex = i;
+        break;
+      }
+    }
+    if (validatorIndex < 0) {
```
gfukushima, line 532:
```suggestion
    if (validatorIndex <= 0) {
```
I think this catches a very edge case scenario where pubKey wasn't in the validators set?

gfukushima (approved):
lgtm...with minor suggestion

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
rolfyone: zero is valid, this is referring back to the initialization of -1... in that case, we didnt find a validator so we'd throw `IllegalStateException`

rolfyone: basically a validator index is (theoretically) uint64, and -1 is strictly out of bounds

gfukushima: Nevermind my comment, I had another scenario in mind when I was suggesting this.

rolfyone: discussed offline, agreement so will resolve this.

PR #8355, opened 2024-06-05
shut down VC when no keys

### LINKED ISSUE #8290: Fail when no keys provided to the VC
When the validator client is not provided with any key or any way to get keys, we should fail and shut down. The validator client only exists to service keys. It should not hang.

```
teku vc
```

mehdi-aouadi:
## PR Description
Shut down immediately when running a standalone validator (`teku vc` subcommand) without providing any key sources:
- No local keys path
- No external signer
- Not enabling the key-manager API

## Fixed Issue(s)

fixes #8290 

## Documentation

## Changelog


rolfyone:
so i guess just for context, to test... startup a minimal node, then create validator config (`min-vc.yaml`):
```
network: "auto"
data-path: min-vc
log-destination: "both"
```

then run 
```
$TEKU vc --config-file min-vc.yaml
```
expect: fail

run
```
$TEKU vc --config-file min-vc.yaml --validator-api-enabled --Xvalidator-api-ssl-enabled=false
```
expect: starts, keymanager is enabled

run
```
$TEKU vc --config-file min-vc.yaml --validators-external-signer-url http://localhost:19000
```
expect: starts, remote signer enabled
even without web3signer active

run
```
$TEKU vc --config-file min-vc.yaml --validator-api-enabled --Xvalidator-api-ssl-enabled=false \
 --validators-external-signer-url http://localhost:19000
```
expect: starts

all of this was fine.

teku/src/integration-test/java/tech/pegasys/teku/cli/subcommand/ValidatorClientCommandTest.java
```diff
@@ -84,22 +84,31 @@ public void setup(final ClientAndServer server) {
         getMockBeaconServerEndpoint(failoverMockBeaconServer);
 
     argsNetworkOptDefault =
-        new String[] {"vc", "--beacon-node-api-endpoint", mockBeaconServerEndpoint};
+        new String[] {
+          "vc", "--validator-keys=keys:pass", "--beacon-node-api-endpoint", mockBeaconServerEndpoint
```
rolfyone, line 88: are we implying here that we need to pass `--validator-keys` now?

rolfyone: my concern is this implies that we used to have a default location or something, and that will no longer work potentially?

mehdi-aouadi: Yes, the `vc` subcommand now requires at least one key source as you summarised it in your first comment (file path, external signer url or enabling the key manager API) and since it's a fail fast logic, it must be added even to correctly parse the command. I chose the `--validator-keys` for the convenience... It could have been `--validator-api-enabled` or `--validators-external-signer-url=http://localhost`.
This doesn't mean we used to have a default location or anything, the tests are simply parsing the command and we weren't requiring any validator keys source at that level

rolfyone: the implied default is gone then? it used to be that one of the local keys was a default... im not sure we really want to force a breaking change...
actually im not sure - you're just saying that there was no way to load keys, so it's a fail... how did we get through parsing previously?

mehdi-aouadi: AFAIK there isn't a default key source if we don't specify any.
IMU we were parsing it successfully even without any key source, which is not good and that's what we're fixing...
So it could be a breaking change for someone running a separate VC without keys (no BN, only VC without any keys, which doesn't make sense).
Running a BN node without keys is still possible in the other hand of course. If you just want to support/follow the network...

rolfyone (approved):

PR #8356, opened 2024-06-05
Fix equivocation checks when publishing blocks via the API

StefanBratanov:
## PR Description
When `CONSENSUS_EQUIVOCATION` is used via the publish API, initially when performing the gossip validation we don't want to mark the block as received, because we will reject blocks coming from gossip and then the final `EQUIVOCATION` check wouldn't be consistent. Instead, we can mark it as received in the last equivocation check unless other block has arrived in gossip during that time.

When our custom only `EQUIVOCATION` is used (default for locally created block and `broadcast_validation: gossip`), we mark the block as received when performing the check in order to reject other blocks coming from gossip afterwards.

In other cases `GOSSIP`, `CONSENSUS`, we mark the block as received as part of the gossip validation. This way we would reject other blocks coming from gossip for this slot and proposer.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8358, opened 2024-06-06
Do gossip validation in `v1` publishing endpoint + refactor

### LINKED ISSUE #4126: Validate locally created blocks prior to gossiping

StefanBratanov:
## PR Description
`v1` of the block publishing endpoint hasn't been aligned with the spec. The spec suggests that gossip validation should be done either before or after broadcasting, but we don't do any validation. Since we have `v2` implemented, it's easy to adapt `GOSSIP` validation and don't change other parts. For the blinded `v1` variant validation is not required, so using `NOT_REQUIRED` validation.

Also did some other small refactors added some defaults in `BlockImportChannel` to simplify calls and changed API descriptions.

## Fixed Issue(s)
fixes #4126 

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #8361, opened 2024-06-07
Metrics upcoming breaking change

gfukushima:
## PR Description
This PR adds an entry to the changelog announcing the upcoming breaking changes in some metrics names due to the upgrade of the dependency `io.prometheus:simpleclient`.
## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Upcoming Breaking Changes
 
+- Next release will introduce a naming breaking change to some of the metrics due to a library upgrade. Revision 12 of the `Teku - Detailed` dashboard has been released and it is backward compatible. We recommend all users to upgrade their dashboard as soon as possible. Documentation with all metrics that have been renamed will be provided.
+
 ## Current Releases
 
 For information on changes in released versions of Teku, see
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## Upcoming Breaking Changes
 
-- Next release will introduce a naming breaking change to some of the metrics due to a library upgrade. Revision 12 of the `Teku - Detailed` dashboard has been released and it is backward compatible. We recommend all users to upgrade their dashboard as soon as possible. Documentation with all metrics that have been renamed will be provided.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Upcoming Breaking Changes
 
+- Next release will introduce a naming breaking change to some of the metrics due to a library upgrade.
```
lucassaldanha: NIT: I think we can break this a bit to make is clearer. Here is the layout I'm thinking:

Also, it would be good to have the version of the dashboard they should be using (and possibly the link to it?).

`Next release will introduce a breaking change to Teku's metrics. This is due to some metrics changing names after a library upgrade. A few existing Grafana dashboard panels won't work with the new metric names. We recommend all users to upgrade their dashboard to version <???> as soon as possible. 
Documentation with all metrics that have been renamed will be provided.`

lucassaldanha: > Documentation with all metrics that have been renamed will be provided.

Can we have the doc page up before we publish this? This way we can directly link to it.

gfukushima: I think the docs being published before the changes getting merged could be a little confusing, specially because we don't know in which release they will be. IMO it only makes sense to publish the docs after having the metrics changes merged into master.

lucassaldanha: Any docs change will go into the "development" version tag of the docs. Any user looking at the stable version of the document (or any other specific version) won't see the new content. This is the same with any other feature that has not been released in a stable version, we update the docs, and, after the release, the docs get "promoted" into the stable version.

gfukushima: Pushed a little change to the entry based on your suggestion. I think it makes sense to be specific on which dashboard has changed because this is the only dashboard affected.

gfukushima: This is a resume of what I think it could lead into some confusion. 
If we add the docs now, they will be part of the next release supposedly 24.6.0. Breaking changes will only be merged after 24.6.0 which means they will be part of a possible 24.7.0 release.
If you look at the docs in 24.6.0 they will be telling you to rename the metrics but in reality 24.6.0 don't have the rename.
Also we can add the docs link to the breaking changes entry in 24.7.0.

lucassaldanha (approved):

Issue #8357, opened 2024-06-06
Startup error: No enum constant tech.pegasys.teku.storage.server.StateStorageMode


lucassaldanha:
### Description

From Discord:

> I just had a power failure, and my there is some problem getting up and running again. I get this message "main | FATAL | BeaconNodeCommand | No enum constant tech.pegasys.teku.storage.server.StateStorageMode."

This isn't the first time I have seen this issue happening. For some reason, the file `<data_folder>/beacon/data-storage-mode.txt` gets corrupted or something and causes this issue.

Need to investigate under what circumstances this happens and how we can better handle it.

[Detailed description of the problem and the impact it has]

### Steps to Reproduce (Bug)
Modify an existing `data-storage-mode.txt` file in the beacon data directory with an invalid value, or empty. Then start teku.

**Expected behavior:**
Service should be able to start with no issues. Or should warn the user when something is wrong (and how to fix it).

1. If user set a storage option, we ALWAYS stick to that option. That option must override whatever is on `data-storage-mode.txt`. We also log a message to let the user know.
2. If user does not set a storage option, and we have a valid `data-storage-mode.txt` file, we use the value on the file to configure the database;
3. If user does not set the storage option, and we have a invalid `data-storage-mode.txt` file, we should warn the user and halt execution. In this case, we don't have a reliable way of deciding what mode makes sense. So we are going to stop and print an error message letting the user know that something is wrong, and that they should restart the node using the option `--data-storage-mode` to override the existing cached config. In the end, `data-storage-mode.txt` will be updated with the mode selected by the user using the option `--data-storage-mode`.

**Actual behavior:**
Node does not start and gives the error: `main | FATAL | BeaconNodeCommand | No enum constant tech.pegasys.teku.storage.server.StateStorageMode.`

**Frequency:**
Once the conditions to trigger this problem appear, they happen consistently, until some manual intervention happens (e.g. deleting the `data-storage-mode.txt` file so it gets recreated.

### Versions (Add all that apply)
TODO
* Software version: [`teku --version` or look for log lines starting with `Teku version:`]
* Java version: [`java -version`]
* OS Name & Version: [`cat /etc/*release`]
* Docker Version: [`docker version`]
* Cloud VM, type, size: [Amazon Web Services I3-large]

haavardf:
... 14 more
[...]

PR #8363, opened 2024-06-10
Using atomic file writer to update data-storage-mode.txt

lucassaldanha:
This is related to #8357 and should prevent instances when we end up with the `data-storage-mode.txt` file empty (and invalid).


rolfyone (approved):
LGTM, possibly worth a changelog entry

PR #8364, opened 2024-06-10
Prefer non optimistic BN failover

### LINKED ISSUE #6847: BNs failover mechanism improvement
In some cases, the request keep being sent to an optimistic node even though there could be a non optimistic and synced one (See point 1). That could lead to failed attestations production first before falling back to a non optimistic node if present. In the next call, the same behaviour will be observed (even though we could already have concluded from the first attempt that there is a "better" alternative).
In this case there is not enough logs to investigate this behaviour too (See point 2). The errors are reported as metrics without printing any logs.

The following improvements could be implemented:

1. Prefer non optimistic nodes: The readiness check could prioritise non optimistic nodes over optimistic ones. The order of preference would be: non optimistic and in sync -> optimistic and in sync -> not in sync
2. Add some logs when the requests fail: When the requests fail, the exceptions are captured but never logged. They are reported as metrics only.

mehdi-aouadi:
## PR Description
Prefer non optimistic over optimistic BN failovers:
- Differentiate optimistic and non optimistic failovers (not only ready/not ready)
- Added a weight logic to the readiness status
- Order the failovers based on their weight `READY > READY_OPTIMISTIC > NOT_READY`:
READY : In sync and non optimistic
READY_OPTIMISTIC : In sync and optimistic 
NOT_READY: Not in sync

## Fixed Issue(s)

fixes #6847 
## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
@@ -73,8 +73,8 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
     return readinessStatus.isReady();
   }
 
-  public ReadinessStatus getReadinessStatus(final RemoteValidatorApiChannel beaconNodeApi) {
-    return readinessStatusCache.getOrDefault(beaconNodeApi, ReadinessStatus.READY);
+  public int getReadinessStatus(final RemoteValidatorApiChannel beaconNodeApi) {
+    return readinessStatusCache.getOrDefault(beaconNodeApi, ReadinessStatus.READY).weight;
   }
 
   public Iterator<? extends RemoteValidatorApiChannel> getFailoversInOrderOfReadiness() {
@@ -161,7 +161,7 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
                 LOG.debug(
                     "{} is ready to accept requests: {}", beaconNodeApiEndpoint, syncingStatus);
                 return syncingStatus.getIsOptimistic().orElse(false)
-                    ? ReadinessStatus.OPTIMISTIC
+                    ? ReadinessStatus.READY_OPTIMISTIC
                     : ReadinessStatus.READY;
               }
               LOG.debug(
@@ -180,7 +180,7 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
         .thenAccept(
             readinessStatus -> {
               readinessStatusCache.put(beaconNodeApi, readinessStatus);
-              if (readinessStatus.weight > 0) {
+              if (readinessStatus.isReady()) {
                 processReadyResult(isPrimaryNode);
               } else {
                 processNotReadyResult(beaconNodeApi, isPrimaryNode);
@@ -212,17 +212,13 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
 
   public enum ReadinessStatus {
     NOT_READY(0),
-    OPTIMISTIC(1),
+    READY_OPTIMISTIC(1),
     READY(2);
 
     private final int weight;
 
     ReadinessStatus(final int weight) {
       this.weight = weight;
-    }
-
-    int getWeight() {
-      return this.weight;
     }
 
     boolean isReady() {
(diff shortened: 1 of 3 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -201,4 +209,24 @@ private void processNotReadyResult(
       beaconNodeReadinessChannel.onFailoverNodeNotReady(beaconNodeApi);
     }
   }
+
+  public enum ReadinessStatus {
+    NOT_READY(0),
+    OPTIMISTIC(1),
```
StefanBratanov: I would rename to `READY_OPTIMISTIC` for more clarity

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -167,12 +175,12 @@ private SafeFuture<Void> performReadinessCheck(
...
-              return false;
+              return ReadinessStatus.NOT_READY;
             })
         .thenAccept(
-            isReady -> {
-              readinessStatusCache.put(beaconNodeApi, isReady);
-              if (isReady) {
+            readinessStatus -> {
+              readinessStatusCache.put(beaconNodeApi, readinessStatus);
+              if (readinessStatus.weight > 0) {
```
StefanBratanov: Can use the `isReady()` method I guess

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -201,4 +209,24 @@ private void processNotReadyResult(
...
+    OPTIMISTIC(1),
+    READY(2);
+
+    private final int weight;
+
+    ReadinessStatus(final int weight) {
+      this.weight = weight;
+    }
+
+    int getWeight() {
```
StefanBratanov: No need of getWeight actually since can directly access `.weight` on inner classes

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
```diff
@@ -188,7 +189,7 @@ private boolean findReadyFailoverAndSwitch() {
         failoverBeaconNodeApis.stream()
             .filter(beaconNodeReadinessManager::isReady)
             .filter(failover -> !currentEventStreamHasSameEndpoint(failover))
-            .findFirst();
+            .max(Comparator.comparing(beaconNodeReadinessManager::getReadinessStatus));
```
StefanBratanov: We are comparing the ordinal of the ReadinessStatus enum but shouldn't we consider the `weight`? It's technically the same thing, so can either use one or the other in my opinion. If we use the ordinal, the `isReady()` check can become simply `!= NOT_READY` and get rid of the weight

StefanBratanov (approved):
Couple of comments, otherwise LGTM

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
mehdi-aouadi: Good point, however I'd rather use the `weight` to make the concept clear and avoid any confusion when maintaining it in the future... Not a fan of using enum ordinals since they're made for sophisticated data structures

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
mehdi-aouadi: Done

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
mehdi-aouadi: Removed it

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
StefanBratanov: Yeah I agree. The change looks good now. I would just rename `getReadinessStatus` to `getReadinessStatusWeight`

mehdi-aouadi: Done

PR #8365, opened 2024-06-11
Further cleanup to deneb validator index cache

rolfyone:
- was a problem with noop cache (admittedly non production)
 - was still filtering non-final indices, and that could cause problems with the way updateLast was working.
 - disabled the non final index testcase

 Now we should be
  - caching all indices from state, as long as we're not noop
  - cleanup interface that was allowing cache access (used in tests but only for size)

  We shouldn't release the last version, i think it'll cause issues in some scenarios

  This is only a step towards the final version, but this one should be safer to release than the previous version.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/ValidatorIndexCache.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/ValidatorIndexCache.java
@@ -25,16 +25,16 @@ import tech.pegasys.teku.spec.datastructures.state.Validator;
 import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
 
 public class ValidatorIndexCache {
-  protected final Cache<BLSPublicKey, Integer> validatorIndices;
-  protected final AtomicInteger lastCachedIndex;
+  private final Cache<BLSPublicKey, Integer> validatorIndices;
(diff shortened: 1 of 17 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/common/ValidatorIndexCache.java
```diff
@@ -62,31 +59,18 @@ public Optional<Integer> getValidatorIndex(
...
   }
 
-  private Optional<Integer> findIndexFromState(
-      final SszList<Validator> validatorList, final BLSPublicKey publicKey) {
-    for (int i = Math.max(lastCachedIndex.get() + 1, 0); i < validatorList.size(); i++) {
-      BLSPublicKey pubKey = validatorList.get(i).getPublicKey();
-      // from electra, only finalized validator indices are stable,
-      // so we look to store only finalized indices not already stored.
-      validatorIndices.invalidateWithNewValue(pubKey, i);
-      updateLastIndex(i);
```
rolfyone, line 72: this is basically a bug... 
basically the net effect here is that the second lookup would fail because it'll scan after that index and miss cache if you're in a NOOP scenario (test added)
Normally its not a noop, and would work, so not awful in deneb currently. fixing now and adding a test stops the problem though.

lucassaldanha (approved):
LGTM

PR #8369, opened 2024-06-12
Added implicit PRUNE mode warning & Handling invalid data-storage-mode file

### LINKED ISSUE #8357: Startup error: No enum constant tech.pegasys.teku.storage.server.StateStorageMode

lucassaldanha:
## PR Description
- Handling error when data-storage-mode.txt file is invalid/empty/corrupt, including a manual step the user can execute to fix it.
- Added deprecation warning for implicit PRUNE data storage mode.

**WARNING**
In the next release, Teku will stop supporting implicit PRUNE database storage mode configuration. If you want to keep using PRUNE mode, you must update your config file or CLI options with `--data-storage-mode=PRUNE`

## Fixed Issue(s)
fixes #8357

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 We recommend all users of the `Teku - Detailed` dashboard to upgrade to version [Revision 12](https://grafana.com/api/dashboards/16737/revisions/12/download)  
 as soon as possible. Documentation with all metrics that have been renamed will be provided.
 - Next release will require Java 21. The current release is compatible, please consider upgrading before the next release.
+- Next release will stop supporting implicit PRUNE database storage mode configuration. If you want to keep using PRUNE mode, you must update your config file or CLI options with --data-storage-mode=PRUNE.
 
 ## Current Releases
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 We recommend all users of the `Teku - Detailed` dashboard to upgrade to version [Revision 12](https://grafana.com/api/dashboards/16737/revisions/12/download)  
 as soon as possible. Documentation with all metrics that have been renamed will be provided.
 - Next release will require Java 21. The current release is compatible, please consider upgrading before the next release.
-- Next release will stop supporting implicit PRUNE database storage mode configuration. If you want to keep using PRUNE mode, you must update your config file or CLI options with --data-storage-mode=PRUNE.
+- From the next release, you will need to explicitly set `--data-storage-mode=(prune|archive)` unless you're using minimal data-storage-mode (which is the default behaviour).
 
 ## Current Releases
 
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -532,6 +532,15 @@ public class StatusLogger {
             Color.YELLOW));
   }
 
+  public void warnUsageOfImplicitPruneDataStorageMode() {
+    log.warn(
+        print(
+            "Prune mode being used as default without a explicit --data-storage-mode option. This will NOT be "
+                + "supported in future Teku versions. Please add --data-storage-mode=prune to your CLI arguments"
+                + " or config file.",
+            Color.YELLOW));
+  }
+
   private void logWithColorIfLevelGreaterThanInfo(
       final Level level, final String msg, final ColorConsolePrinter.Color color) {
     final boolean useColor = level.compareTo(Level.INFO) < 0;
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -537,7 +537,7 @@ public class StatusLogger {
         print(
             "Prune mode being used as default without a explicit --data-storage-mode option. This will NOT be "
                 + "supported in future Teku versions. Please add --data-storage-mode=prune to your CLI arguments"
-                + " or config file.",
+                + " or config file if you want to keep using PRUNE.",
             Color.YELLOW));
   }
 
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
@@ -18,9 +18,6 @@ import static tech.pegasys.teku.storage.server.StateStorageMode.NOT_SET;
 import static tech.pegasys.teku.storage.server.StateStorageMode.PRUNE;
 import static tech.pegasys.teku.storage.server.VersionedDatabaseFactory.STORAGE_MODE_PATH;
 
-import java.io.IOException;
-import java.io.UncheckedIOException;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.Optional;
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Next release will introduce a breaking change to Teku's metrics. This is due to some metrics changing names after a library upgrade.
 We recommend all users of the `Teku - Detailed` dashboard to upgrade to version [Revision 12](https://grafana.com/api/dashboards/16737/revisions/12/download)  
 as soon as possible. Documentation with all metrics that have been renamed will be provided.
+- Next release will stop supporting implicit PRUNE database storage mode configuration. If you want to keep using PRUNE mode, you must update your config file or CLI options with --data-storage-mode=PRUNE.
```
zilm13: It's difficult to understand. Maybe focus on default behavior changing from PRUNE to MINIMAL unless explicitely configured something else ?

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -532,6 +532,15 @@ public void warnIgnoringWeakSubjectivityPeriod() {
             Color.YELLOW));
   }
 
+  public void warnUsageOfImplicitPruneDataStorageMode() {
+    log.warn(
+        print(
+            "Prune mode being used as default without a explicit --data-storage-mode option. This will NOT be "
+                + "supported in future Teku versions. Please add --data-storage-mode=prune to your CLI arguments"
+                + " or config file.",
```
zilm13: .. if you want to keep it"?

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
```diff
@@ -267,39 +264,36 @@ public StorageConfiguration build() {
...
+
         this.dataStorageMode =
             determineStorageDefault(
-                dataDirLayout.getBeaconDataDirectory().toFile().exists(),
-                getStorageModeFromPersistedDatabase(dataDirLayout),
-                dataStorageMode);
+                beaconDataDirectory.toFile().exists(), storageModeFromStoredFile, dataStorageMode);
       } else {
         if (dataStorageMode.equals(NOT_SET)) {
           dataStorageMode = PRUNE;
```
zilm13, line 287: I don't understand this, it defaults to PRUNE

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
```diff
@@ -315,6 +309,20 @@ static StateStorageMode determineStorageDefault(
...
+      final StateStorageMode stateStorageMode = maybeHistoricStorageMode.get();
+      if (stateStorageMode == PRUNE) {
+        STATUS_LOG.warnUsageOfImplicitPruneDataStorageMode();
+      }
+      return stateStorageMode;
+    }
+
+    if (isExistingStore) {
+      STATUS_LOG.warnUsageOfImplicitPruneDataStorageMode();
+      return PRUNE;
```
zilm13, line 323: so we wll change this behavior in another release, right?

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
```diff
@@ -88,20 +88,25 @@ public VersionedDatabaseFactory(
...
+        LOG.info(
+            "Storage mode that was persisted differs from the command specified option; file: {}, CLI: {}",
+            maybeStorageModeFromFile.get(),
+            maybeConfiguredStorageMode.get());
       }
-    } catch (IOException e) {
-      LOG.error("Failed to read storage mode file", e);
     }
 
     return maybeConfiguredStorageMode.orElse(StateStorageMode.DEFAULT_MODE);
```
zilm13: method name is `getStateStorageModeFromConfigOrDisk`, but we read only for config actually and warn if persisted doesn't match. Maybe rename method?

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
```diff
@@ -88,20 +88,25 @@ public VersionedDatabaseFactory(
...
+    } catch (final DatabaseStorageException e) {
+      if (maybeConfiguredStorageMode.isEmpty()) {
+        throw e;
+      }
+    }
+
+    if (maybeStorageModeFromFile.isPresent() && maybeConfiguredStorageMode.isPresent()) {
+      if (!maybeStorageModeFromFile.get().equals(maybeConfiguredStorageMode.get())) {
+        LOG.info(
+            "Storage mode that was persisted differs from the command specified option; file: {}, CLI: {}",
```
zilm13: It would be good to see here which one will be actually used. And maybe it should be `warn`?

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
zilm13: If next one answer is "YES", this one is clear. Do we need to print warn here too?

CHANGELOG.md
rolfyone: i think this is specifically focussed on what we need users to do, im not sure we can make it easier to understand without saying
```
From the next release, you will need to explicitly set `--data-storage-mode=(prune|archive)` unless you're using minimal data-storage-mode (which is default behaviour)
```

zilm13: I like your version more. Some users don't even know which mode they are using and what is the difference

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
lucassaldanha: To be honest, I think I can clear this up... I don't think there is a code path where `dataConfig` will ever be null. So this else should never happen in production.

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
lucassaldanha: Correct: https://github.com/Consensys/teku/issues/8371

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
lucassaldanha: We have another log line that says what data storage mode is being used. Not sure if there is value in adding it here as well.

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
lucassaldanha: Now that I am thinking about it, we can probably get rid of this method... There is no point in warning about this anymore as we are planning to get rid of this behaviour soon... Thoughts?

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
zilm13: Right, this one is called on init, another one later will log actual

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
zilm13: Yeah, we print actually used, we don't use file info here, so it makes sense (to clean it)

rolfyone (approved):
LGTM check tests pls

PR #8376, opened 2024-06-13
Dump blob sidecars which failed data availability

### LINKED ISSUE #8259: Make `P2PDebugDataDumper` more generic
There is value in making `P2PDebugDataDumper` more generic. Currently, name is slightly misleading, since for example `saveInvalidBlockToFile` can occur when syncing or when called via the beacon-API, so it is not strictly `P2P`. We could rename `--Xp2p-dumps-to-file-enabled`  to sth like `--Xdebug-data-dumping-enabled` and rename `P2PDebugDataDumper` to `DebugDataDumper` and use it in several places. Some options are;

- dump blob sidecars when `is_data_available` fails
- dump blobs bundle coming from builder if it fails the consistency checks

StefanBratanov:
## PR Description
Dump blob sidecars which failed the data availability check (invalid)
## Fixed Issue(s)
fixes #8259 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ the [releases page](https://github.com/Consensys/teku/releases).
 
 ### Breaking Changes
 
+- Renamed `--Xp2p-dumps-to-file-enabled` hidden CLI option to `--Xdebug-data-dumping-enabled`
+
 ### Additions and Improvements
 
 - Added metadata fields to `/eth/v1/beacon/blob_sidecars/{block_id}` Beacon API response as per https://github.com/ethereum/beacon-APIs/pull/441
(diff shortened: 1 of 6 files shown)


tbenr:
I perfectly get the INVALID dump, but dumping the available but incomplete ones seems... weird

the INVALID is more "one or more of the dumped ones is invalid".


the invalid will be kept by the gossip validation first (in most cases), so I guess we will dump it as a raw message

StefanBratanov:
> I perfectly get the INVALID dump, but dumping the available but incomplete ones seems... weird
> 
> the INVALID is more "one or more of the dumped ones is invalid".
> 
> the invalid will be kept by the gossip validation first (in most cases), so I guess we will dump it as a raw message

Thanks, actually you are right `NOT_AVAILABLE` is empty sidecars, so no reason. I changed the code a bit where basically we only dump the sidecars on `INVALID` but also dump the kzg commitments from the block. Got rid of the failure reason and throwable since only used in logging which I don't think it's necessary looking at the code.

tbenr (approved):
LGTM

PR #8379, opened 2024-06-14
Add pruner for finalized states

### LINKED ISSUE #8331: Allow pruning "archive" node
In some scenarios, users could be interested in running archive nodes (with finalized states) up to a certain time in the past. This could enable having a low value in `--data-storage-archive-frequency`, so the node can be very responsive within the interested time interval.

To achieve this we could add a `--data-storage-archive-retained-epochs` (or similar). Setting value to `1024`, teku will prune finalized states older than 1024 epochs compared to the current epoch.

A node running this way won't be a strict archive node, kind of a "time limited archive node".

Note:
We should not allow state reconstruction when running in this mode.

gfukushima:
## PR Description
This PR adds a state pruner that can prune finalized states keeping only the finalized states for the amount of epochs specified using the experimental flag `--Xdata-storage-archive-finalized-states-retained`. 
Available in archive mode exclusively.

## Fixed Issue(s)

Fixes #8331 
## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDaoAdapter.java
```diff
@@ -553,6 +558,11 @@ public void addFinalizedState(final Bytes32 blockRoot, final BeaconState state)
       finalizedUpdater.addFinalizedState(blockRoot, state);
     }
 
+    @Override
+    public void deleteFinalizedState(final UInt64 slot) {
+      finalizedUpdater.deleteFinalizedState(slot);
+    }
+
```
rolfyone, line 565: i do wonder if hte only interface we want is the deleteFinalizedState, and the rest is hidden and done for sanity

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -61,6 +66,7 @@ public StorageService(
     this.config = storageConfiguration;
     this.depositSnapshotStorageEnabled = depositSnapshotStorageEnabled;
     this.blobSidecarsStorageCountersEnabled = blobSidecarsStorageCountersEnabled;
+    LOG.info("Storage service created");
```
rolfyone: probably leave this as debug

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -111,6 +117,25 @@ protected SafeFuture<?> doStart() {
                             pruningTimingsLabelledGauge,
                             pruningActiveLabelledGauge));
               }
+              LOG.info(
+                  "Data storage mode: {}, Retained epochs {}",
+                  config.getDataStorageMode(),
+                  config.getRetainedEpochs());
```
rolfyone: we probably want to be mindful of logs here... maybe either put this to debug, or only log this when we're in archive mode...
also probalby want to think about this service WRT tree storage mode..

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
```diff
@@ -38,6 +38,9 @@ public class StorageConfiguration {
   public static final Duration DEFAULT_BLOCK_PRUNING_INTERVAL = Duration.ofMinutes(15);
   public static final int DEFAULT_BLOCK_PRUNING_LIMIT = 5000;
   public static final Duration DEFAULT_BLOBS_PRUNING_INTERVAL = Duration.ofMinutes(1);
+  public static final Duration DEFAULT_STATE_PRUNING_INTERVAL = Duration.ofMinutes(10);
+  public static final long DEFAULT_STORAGE_RETAINED_EPOCHS = -1;
+  public static final int DEFAULT_STATE_PRUNING_LIMIT = 3;
```
rolfyone: i think if this runs during epoch transition it may get interesting, more of a general observation about all of these types of tasks i guess... we probably want to build out a design at some point that gives us more confidence about what parts of an epoch these tasks might be allowed...

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
gfukushima: yes, these are some leftovers I missed from a test. I'm doing a cleanup on some of those atm. the one that survive the clean up in the end they will all be at debug level

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDaoAdapter.java
gfukushima: I'm following similar approach used for addFinalizedState, imo it isn't too complicated to track down what changes are made when we store/delete a finalized state in the db atm.

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
gfukushima: tbh, this has been the most difficult part so far, to find/define a good set up, I thought about something a little more dynamic. E.g having the interval being defined by epochs_retained and epoch_durantion

gfukushima:
There is a question which is, should we allow users to config `--Xdata-storage-archive-finalized-states-retained` to a value lower than `--data-storage-archive-frequency`? This would mean that you might have a finalized state for a short period of time in the DB.

Example:
 --data-storage-archive-frequency = 2048 (default = 64 epochs) 
--Xdata-storage-archive-finalized-states-retained = 32 (1 epoch)

You might save a finalized state for slot number 10, and by the time slot 42 reaches finalization state for slot 10 will be pruned. Next finalized state to be stored in the db will be on slot 2058 only.

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
tbenr: I agree with Paul that we could have a "no flight time window" which could be epoch transition +/- 1 or 2 slots (ie from 31 to 2). The question will be how to handle tasks falling in that window:
- skip: (pro: easy, cons: no execution guarantee for a long time)
- queue them to the end of the time window (slot 2): (pro: execution guarantees, cons: multiple tasks can potentially run at the same time)

 Just dumped some thoughts.

tbenr:
> Next finalized state to be stored in the db will be on slot 2058 only.

I guess you intended 2048 here otherwise I'm missing something in the reasoning :)

In general I think it is an edge case, and definitely a weird configuration to have. My mild feeling would be to enforce retain to be > than frequency

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
tbenr: Another option is to link these prunings explicitly to certain slots, so the frequency will be an epoch for all of them, and we will have an implicit control over the epoch transition overlap (not guaranteed but reasonable).

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -432,6 +432,86 @@ private void deleteFinalizedBlocks(final List<Pair<UInt64, Bytes32>> blocksToPru
...
+                          slot,
+                          dao.getFinalizedBlockAtSlot(slot).map(SignedBeaconBlock::getStateRoot)))
+              .toList();
+    }
+    if (slotsToPruneStateFor.isEmpty()) {
+      LOG.debug("No finalized state to prune up to {} slot", lastSlotToPrune);
+      return lastSlotToPrune;
+    }
+
+    final UInt64 lastPrunedBlockSlot =
```
tbenr: should we remove "block" from var name here?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -432,6 +432,86 @@ private void deleteFinalizedBlocks(final List<Pair<UInt64, Bytes32>> blocksToPru
...
+      final UInt64 earliestFinalizedStateSlot,
+      final UInt64 lastSlotToPrune,
+      final long pruneLimit) {
+    final List<Pair<UInt64, Optional<Bytes32>>> slotsToPruneStateFor;
+
+    try (final Stream<UInt64> stream =
+        dao.streamFinalizedStateSlots(earliestFinalizedStateSlot, lastSlotToPrune)) {
+      slotsToPruneStateFor =
+          // should probably create a cache (populated when adding finalized states)
+          // to query from rather than hit the db
```
tbenr: I'd make it an issue and remove the comments. Seems to me like there might be some debate on if\how we should do it.

tbenr (approved):
LGTM. some comments and still a LOG level change requested by Paul

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
gfukushima: done

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
gfukushima: done

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
gfukushima: changed this to be a more informative log but kept it at INFO level since it is only printed at startup, WDYT?

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
gfukushima: removed

gfukushima:
Adding some numbers collected from instances before merging this PR:

DB Size of different setups:
Holesky-01 (minimal)
DB - 72G	teku/beacon/db

Holesky-archive 
DB - 212G	teku/beacon/db

Holesky-archive-64epochs
DB - 110G	teku/beacon/db

PR #8380, opened 2024-06-14
Added electra validator index caching

rolfyone:
Added performance test

Altered some tests as we don't really want any of this functionality in deneb. It's gated currently on a state identifying as an electra state. The new performance test goes through unstable lookups only, so its good to compare against the ValidatorIndexCache which is the pre-electra implementation (and finalized generally)

This may end up being not required, depending on some outstanding consensus spec changes. It should be noted cache misses in the non final range with 400k validators is a lot slower than it would be if the non final range wasn't the entire validator set.

## Documentation

## Changelog


rolfyone:
will no longer be required.

PR #8387, opened 2024-06-19
change defaults for all-subnets mode

rolfyone:
When we're running in all-subnets mode, there's a number of considerations

 - we basically always recommend increasing queue sizes
 - we could make our lives easier by reducing our peers, given we're on all subnets.

 Added configurations to override only if they're not default values, so this would mean user-configuration will still be preferred ouver either set of our configuration.

 An alternative to this would be always allow 40k max for all queues, which is a smaller change overall, but we'd probably want to still consider peer count defaults for all-subnets. I believe that we should need less peers if we're listening on all subnets.

 Will probably need to test this, may be able to look at our 'correct head' metric to see if all-subnet subscriptions with reduced peers is more effective than current nodes.

## Documentation

## Changelog


tbenr:
I'm not convinced that reducing number of peers lowers down the number of messages we process (and lowers pressure on our queues). I believe it will reduce pressure at libp2p level only (which should have a relative low benefit).  We could have a discussion offline (or prove it in practice).

At the end I agree with both changes :)

tbenr (approved):
LGTM. We can tune params later on if needed

PR #8402, opened 2024-06-25
Provide Electra body builder test

tbenr:
The only required change is due to the fact that in electra we lower down attester shlashings block from 2 to 1.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8405, opened 2024-06-25
Improve failover robustness

### LINKED ISSUE #8180: Degraded attestation performance when Teku VC has a secondary BN defined

zilm13:
## PR Description
Covered following cases:
- `switchToFailoverEventStreamIfAvailable()` is not always fired when currently used api fails, it could be fired by connection error fallback for example and switch endpoint (not considering primary, when it could be the best) when not needed
- there could be a case when `BeaconNodeReadinessManager` is switched to primary as ready, sent event once and will not repeat it, but by some reason this event slipped in `EventSourceBeaconChainEventAdapter` (say callback `switchToFailoverEventStreamIfAvailable()` fired just after it). Repeated events guarantee that we will switch to primary in this case sooner or later and price is nothing.

Wanted to add some tests, but it doesn't look it could be tested easily. Maybe you could provide some suggestion?

## Fixed Issue(s)

Fixes #8180 (not 100% on this)

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
@@ -199,7 +199,7 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
       return;
     }
     // Filtering of duplicates if needed happens on receiver's side
-    beaconNodeReadinessChannel.onPrimaryNodeBackReady();
+    beaconNodeReadinessChannel.onPrimaryNodeReady();
     if (latestPrimaryNodeReadiness.compareAndSet(false, true)) {
       validatorLogger.primaryBeaconNodeIsBackAndReady();
     }
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
@@ -14,6 +14,7 @@
 package tech.pegasys.teku.validator.remote;
 
 import com.google.common.collect.Maps;
+import com.google.common.collect.Streams;
 import java.time.Duration;
 import java.util.Comparator;
 import java.util.Iterator;
@@ -71,6 +72,10 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
     final ReadinessStatus readinessStatus =
         readinessStatusCache.getOrDefault(beaconNodeApi, ReadinessStatus.READY);
     return readinessStatus.isReady();
+  }
+
+  public ReadinessStatus getReadinessStatus(final RemoteValidatorApiChannel beaconNodeApi) {
+    return readinessStatusCache.getOrDefault(beaconNodeApi, ReadinessStatus.READY);
   }
 
   public int getReadinessStatusWeight(final RemoteValidatorApiChannel beaconNodeApi) {
@@ -143,7 +148,8 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
     final SafeFuture<Void> primaryReadinessCheck = performPrimaryReadinessCheck();
     final Stream<SafeFuture<?>> failoverReadinessChecks =
         failoverBeaconNodeApis.stream().map(this::performFailoverReadinessCheck);
-    return SafeFuture.allOf(primaryReadinessCheck, SafeFuture.allOf(failoverReadinessChecks));
+    return SafeFuture.allOf(
+        Streams.concat(Stream.of(primaryReadinessCheck), failoverReadinessChecks));
   }
 
   private SafeFuture<Void> performFailoverReadinessCheck(final RemoteValidatorApiChannel failover) {
@@ -192,9 +198,10 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
     if (!isPrimaryNode) {
       return;
     }
+    // Filtering of duplicates if needed happens on receiver's side
+    beaconNodeReadinessChannel.onPrimaryNodeBackReady();
     if (latestPrimaryNodeReadiness.compareAndSet(false, true)) {
       validatorLogger.primaryBeaconNodeIsBackAndReady();
-      beaconNodeReadinessChannel.onPrimaryNodeBackReady();
     }
   }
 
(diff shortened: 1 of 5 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -192,9 +198,10 @@ private void processReadyResult(final boolean isPrimaryNode) {
     if (!isPrimaryNode) {
       return;
     }
+    // Filtering of duplicates if needed happens on receiver's side
+    beaconNodeReadinessChannel.onPrimaryNodeBackReady();
```
mehdi-aouadi: I don't understand the purpose of this change. At this point the primary node is not yet confirmed **back** ready (could be already marked as ready) so why should we notify the subscribers before checking that its readiness status changed?

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
```diff
@@ -181,6 +182,12 @@ private synchronized boolean switchToFailoverEventStreamIfAvailable() {
     if (failoverBeaconNodeApis.isEmpty()) {
       return false;
     }
+    // No need to change anything if current node is READY
+    if (beaconNodeReadinessManager
+        .getReadinessStatus(currentBeaconNodeUsedForEventStreaming)
+        .equals(READY)) {
```
mehdi-aouadi, line 188: IMU the idea here is to check if the current BN is ready before switching to another one, right?
If that's the case, shouldn't we rather check `isReady()` instead of `equals(READY)`?
What if we refactor the `findReadyFailoverAndSwitch` instead and add the `currentBeaconNodeUsedForEventStreaming` to the stream and let it select the best available BN?

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
zilm13: The point is that we could have inconsistency between Manager and subscribers. We send readiness only once and if subscriber failed to process it correctly, it will never receive another readiness keeping running on failover

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapter.java
zilm13: > shouldn't we rather check isReady()

It will remove an option to switch from optimistic_ready to ready

> What if we refactor the findReadyFailoverAndSwitch instead 

It will not guarantee order on each call. What I've tried  to do - to keep the current node if switch is not needed. This method could be called even in case Primary, ready, active, Failover - callback on disconnect and it will change primary to failover in this case in current master.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
mehdi-aouadi: Ok I see your point. Wouldn't it then be better to rename `onPrimaryNodeBackReady` to simply `onPrimaryNodeReady` since there is no _back ready_ notion anymore?

zilm13: Nice suggestion! Updated

mehdi-aouadi:
Regarding the tests, It would be great to add a test that covers the `EventSourceBeaconChainEventAdapter` not switching to another BN when the current one is `READY`. We normally can mock the `beaconNodeReadinessManager.getReadinessStatus(currentBeaconNodeUsedForEventStreaming)` call to make it return `READY` and then check that the `findReadyFailoverAndSwitch` hasn't been called (check that the `currentBeaconNodeUsedForEventStreaming` is still the same for example)

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapterTest.java
```diff
@@ -97,6 +100,45 @@ public void performsPrimaryReadinessCheckWhenFailoverNotReadyAndNoOtherFailovers
     verify(beaconNodeReadinessManager).performPrimaryReadinessCheck();
   }
 
+  @Test
+  public void dontJumpBetweenFailoversWhenFailoverIsReady() {
```
mehdi-aouadi: Maybe rename it to `doNotSwitchToFailoverWhenCurrentBeaconNodeIsReady`

mehdi-aouadi (approved):
LGTM
Just a small nit

validator/remote/src/test/java/tech/pegasys/teku/validator/remote/eventsource/EventSourceBeaconChainEventAdapterTest.java
zilm13: Nice name, updated

PR #8410, opened 2024-06-26
Enhanced IPv6 support (discovery)

StefanBratanov:
## PR Description

- Add `--Xp2p-udp-port-ipv6` and `--Xp2p-advertised-udp-port-ipv6`
- Upgrade discovery to `24.6.0`
- return `Optional<List<String>>` in `DiscV5Service.getDiscoveryAddress`

This PR doesn't break any current functionality and everything else is hidden.

## Fixed Issue(s)
related to #8069 

## Documentation

## Changelog

### DIFF
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/DiscV5Service.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/DiscV5Service.java
@@ -15,6 +15,8 @@ package tech.pegasys.teku.networking.p2p.discovery.discv5;
 
 import static java.util.Collections.emptyList;
 
+import com.google.common.base.Preconditions;
+import java.net.InetSocketAddress;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
@@ -37,6 +39,8 @@ import org.hyperledger.besu.plugin.services.MetricsSystem;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.Cancellable;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.io.IPVersionResolver;
+import tech.pegasys.teku.infrastructure.io.IPVersionResolver.IPVersion;
 import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
 import tech.pegasys.teku.networking.p2p.discovery.DiscoveryConfig;
 import tech.pegasys.teku.networking.p2p.discovery.DiscoveryPeer;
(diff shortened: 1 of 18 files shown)


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfig.java
```diff
@@ -362,9 +362,9 @@ private void validateAddresses(final List<String> addresses, final String cliOpt
       }
     }
 
-    private void validatePort(final int port, final String cliOption) {
+    private void validatePort(final int port, final String parameter) {
```
zilm13: Not sure on this change, we still have cliOptions in `validateAddresses`

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/NodeRecordConverter.java
```diff
@@ -36,8 +36,10 @@ public class NodeRecordConverter {
 
   public Optional<DiscoveryPeer> convertToDiscoveryPeer(
       final NodeRecord nodeRecord, final SchemaDefinitions schemaDefinitions) {
+    // TODO: https://github.com/Consensys/teku/issues/8069
```
zilm13, line 39: Not clear why it's needed here
We get tcp6 in `or` here, isn't it not enough? We need list here too?

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/DiscV5Service.java
```diff
@@ -121,15 +166,25 @@ public DiscV5Service(
         () -> discoverySystem.getBucketStats().getTotalLiveNodeCount());
   }
 
-  private NewAddressHandler maybeUpdateNodeRecord(
-      final boolean userExplicitlySetAdvertisedIpOrPort, final int advertisedTcpPort) {
-    if (userExplicitlySetAdvertisedIpOrPort) {
+  private NewAddressHandler maybeUpdateNodeRecord(final NetworkConfig p2pConfig) {
+    if (p2pConfig.hasUserExplicitlySetAdvertisedIps()) {
```
zilm13, line 170: I don't understand this, shouldn't it be inversed?

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/NodeRecordConverter.java
StefanBratanov: to be fair, probably `or` is enough. It's just that in this case, if we are dual-stack, we would always try to connect to IPv4 when ideally we would prefer IPv6

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfig.java
StefanBratanov: Yeah, I like cliOption better to be honest. Changed.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/NodeRecordConverter.java
StefanBratanov: Will tackle in future PR

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/DiscV5Service.java
StefanBratanov: Same behaviour as before. If user explicitly sets advertised IPs, they are added in the ENR. Otherwise, NewAddressHandler is communicated between nodes who don't have explicit advertised ips, in that case, we want to have a handler configured in discovery

zilm13 (approved):
LGTM

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/DiscV5Service.java
zilm13: Needed to look into discovery to understand. So it fires when user address is changed and when advertised is configured it means that we should prefer advertised over anything changed, clear now

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/NodeRecordConverter.java
zilm13: So my message was in case if someone other will take it in future, I want it to be clear for anyone else, what do we want there ideally

Issue #7356, opened 2023-07-12
VC failover switch back to primary node when target peer count is met


StefanBratanov:
Inspired from a discussion on Discord: https://discord.com/channels/697535391594446898/697539289042649190/1128622402822213692

After a primary BN is recovering from network outage/restart, it would be good to be able to configure VC to switch back to primary only when certain amount of peers is met. This will allow for a smoother failover. A standard beacon api endpoint could be used to get the peer count: https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Node/getPeerCount

Tasks:
- [x] #8413 Add getPeerCount to the validator api channel
- [x] #8414 Add the getPeerCount remote validator api channel handler
- [x] #8415 Use the new `getPeerCount` in the BN fallback logic

PR #8413, opened 2024-06-29
PART 1 - GetPeerCount validator api channel

mehdi-aouadi:
## PR Description
Add `getPeerCount` to the validator API channel: This new method will be used by the BN fallback mechanism in order to select new failovers based on a minimum connected peers count.
The remote request is not implemented yet at this stage. It will be added in a separate PR. 
Part of #7356

## Documentation

## Changelog

### DIFF
--- a/ethereum/json-types/src/main/java/tech/pegasys/teku/ethereum/json/types/node/PeerCountBuilder.java
+++ b/ethereum/json-types/src/main/java/tech/pegasys/teku/ethereum/json/types/node/PeerCountBuilder.java
@@ -48,9 +48,9 @@ public class PeerCountBuilder {
           .withField("data", PEER_COUNT_DATA_TYPE, Function.identity(), PeerCountBuilder::peerCount)
           .build();
 
-  private UInt64 disconnected;
+  private UInt64 disconnected = UInt64.ZERO;
   private UInt64 connecting = UInt64.ZERO;
-  private UInt64 connected;
+  private UInt64 connected = UInt64.ZERO;
   private UInt64 disconnecting = UInt64.ZERO;
 
   public PeerCountBuilder disconnected(final UInt64 disconnected) {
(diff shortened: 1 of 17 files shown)


ethereum/json-types/src/main/java/tech/pegasys/teku/ethereum/json/types/node/PeerCountBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+  public static final DeserializableTypeDefinition<PeerCount> PEER_COUNT_TYPE =
+      DeserializableTypeDefinition.object(PeerCount.class, PeerCountBuilder.class)
+          .name("GetPeerCountResponse")
+          .initializer(PeerCountBuilder::new)
+          .finisher(PeerCountBuilder::build)
+          .withField("data", PEER_COUNT_DATA_TYPE, Function.identity(), PeerCountBuilder::peerCount)
+          .build();
+
+  private UInt64 disconnected;
+  private UInt64 connecting = UInt64.ZERO;
```
zilm13, line 52: Why connecting and disconnecting defaults to ZERO? all are required by spec.

zilm13 (approved):
LGTM

ethereum/json-types/src/main/java/tech/pegasys/teku/ethereum/json/types/node/PeerCountBuilder.java
zilm13: Ahh I see, we are not setting it in builder. I'd just set ZERO to all by default to avoid confusion. You cannot know all usage when creating isolated class.

mehdi-aouadi: Actually we used to set `connecting` and `disconnecting` to 0 because we don't have that information in the `LibP2PPeer`... But I agree that it's better to avoid confusion, updated it to initialise all the values to 0

PR #8416, opened 2024-07-01
Port forward IPv6 port

StefanBratanov:
## PR Description
Port forwarding the ipv6 listen port (similar to the default ipv4)

## Fixed Issue(s)
will resolve one of the tasks under #8069 

## Documentation

## Changelog


zilm13 (approved):
It would be great to refactor ports in configs one day, so we don't need to check it in this hacky way (interfaces == 2)

StefanBratanov:
> It would be great to refactor ports in configs one day, so we don't need to check it in this hacky way (interfaces == 2)

I think will do this as part of another PR to avoid too many changes

PR #8417, opened 2024-07-03
Fixing remaining reference tests for devnet-1

### LINKED ISSUE #8409: Fix remaining Electra reference tests for Devnet 1

lucassaldanha:
## PR Description

Fixes function `get_expected_withdrawals` for Electra.

There were three issues:
1. Processing of full withdrawals, when checking if the validator had execution credentials, we were using `predicates.hasEth1WithdrawalCredential(..)`. However, in Electra the validator can also have `0x02` (compounding) credentials. The simple solution was to move `predicates.hasExecutionWithdrawalCredential(..)` into `Predicates`, with a default implementation that only looks at `0x01` credentials, but with a override in `PredicatesElectra` that checks for `0x01` OR `0x02` creds. 
2. When calculating the withdrawal amount, we need to make sure that when we are in Electra we use the correct max effective balance depending on the type of withdrawal credentials of the validator. The solution was to use `predicates.getValidatorMaxEffectiveBalance(..)`.
3. The function `is_partially_withdrawable_validator` has been updated to use `get_validator_max_effective_balance` instead of `MAX_EFFECTIVE_BALANCE`. 

References: 
- https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-get_expected_withdrawals
- https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-is_partially_withdrawable_validator

## Fixed Issue(s)
fixes #8409

## Documentation

## Changelog

### DIFF
--- a/eth-tests/src/main/java/tech/pegasys/teku/ethtests/finder/ReferenceTestFinder.java
+++ b/eth-tests/src/main/java/tech/pegasys/teku/ethtests/finder/ReferenceTestFinder.java
@@ -31,7 +31,12 @@ public class ReferenceTestFinder {
       Path.of("src", "referenceTest", "resources", "consensus-spec-tests", "tests");
   private static final List<String> SUPPORTED_FORKS =
       List.of(
-          TestFork.PHASE0, TestFork.ALTAIR, TestFork.BELLATRIX, TestFork.CAPELLA, TestFork.DENEB);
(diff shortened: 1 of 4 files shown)


gfukushima (approved):
lgtm

PR #8419, opened 2024-07-03
Integrate fork choice compliance test suite

mkalinin:
## PR Description

Adds support of the fork choice compliance test suite, the source of which can be found in the following spec PR:
* https://github.com/ethereum/consensus-specs/pull/3831

### Running tests

To run tests one need to download one of the test corpus (differentiated by a number of tests):
* [tst_tiny.tar](https://drive.google.com/file/d/1dWbkJY27fhOVX8i4aUuZ7VBkH3P_Vr1i/view), 135 tests
* [tst_small.tar](https://drive.google.com/file/d/1EAIeTL5F3zelXK5pBkSLuawJjuLWQx3r/view), 1472 tests
* [tst_standard.tar](https://drive.google.com/file/d/1_56JObwYWHARgm5QYmE4vrkcLEru854G/view), 13240 tests

Extract with command:
```bash
tar xvf tst_standard.tar
```

And use `ManualReferenceTestRunner` with `-Dteku.ref-test-module.override-root=~/Downloads/tst_standard/` to specify the directory with the test vectors.

It is required to expand the `consensus-spec-tests` suites to build the module where `ManualReferenceTestRunner` is located so running the new tests this way looks more like a temporary solution and potentially a separate test runner is required.

### Implementation detail

The new test suite is based on the fork choice test suite and extends the fork choice test format with only one check, namely `viable_for_head_roots_and_weights`, so this PR leverages on the existing `ForkChoiceTestExecutor` implementation and extend it to adopt the suite. However, existing `ForkChoiceTestExecutor` implementation and the suite design details required a more extensive changes to the codebase, the following list might not be exhaustive:
* Read `blsSetting` from `meta.yaml` and disable BLS for block and attestation processing if that is required by the setting, likewise, do fake pubkey aggregation computation in the sync committee updates during epoch transition. For performance reasons the new suite generates tests with fake BLS that should run with BLS disabled. The support is added via provisioning `SpecConfig` with the BLS setting, this is done to avoid a ton of boilerplate that would be required in the next sync committee update case.
* Add an option to discard deferred attestations and turn this option on in the fork choice test runner. The new suite shuffles messages and expects that if the attestation comes earlier it is deemed invalid and ins’t applied in the future. We were debating internally on how to properly design this particular part, and come to the conclusion that for simplicity we don’t expect anything is deferred. But this part of the suite can be revisited after feedback from client developers gets collected.
* Attestations can be invalid hence `valid` flag support for attestations is added.
* Support `viable_for_head_roots_and_weights` check — this entailed implementation of `getViableChainHeads` method.

### Known issues

A few tests (around 10) from the `tst_standard` corpus fail due to the following reasons:
* Teku does not process attestations voting for a non-viable head, results into an `UNKNOWN_BLOCK` attestation status
* Teku does not process attestation if it attests to a block that is older than two epochs with the following message in the exception: `Committee information must be derived from a state no older than the previous epoch`

The above looks like fair optimisation measure protecting from spamming nodes with the old attestations that have almost no value.

A couple of tests from the `tst_standard` corpus fail due to the following spec issue:
* If attestation from the current epoch is included in the block which parent is two epochs old then due to different shuffling seed aggregation bits of these attestations would be resolved differently depending on the state that is used. Teku caches validator indices from the block processing and uses them when attestations are applied to the fork choice which is a fair optimisation. While the spec resolves those indices based on the state to which attestations were attesting to. Anyway, when spec processes such an attestation it uses wrong indices for activity and rewards mechanism.

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/forkchoice/ForkChoiceTestExecutor.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/forkchoice/ForkChoiceTestExecutor.java
@@ -17,15 +17,21 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static tech.pegasys.teku.infrastructure.async.SafeFutureAssert.safeJoin;
 import static tech.pegasys.teku.infrastructure.time.TimeUtilities.secondsToMillis;
+import static tech.pegasys.teku.reference.BlsSetting.IGNORED;
+import static tech.pegasys.teku.reference.TestDataUtils.loadYaml;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
 import java.nio.ByteOrder;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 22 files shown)


lucassaldanha:
Closing some stale PRs. Feel free to re-open if this still makes sense.

Issue #8029, opened 2024-02-28
Add the attestations pool v2 API


mehdi-aouadi:
Add new `GetAttestations` V2 API to handle electra attestations
[Spec PR link](https://github.com/ethereum/beacon-APIs/pull/448)

PR #8431, opened 2024-07-09
PART 1 - GET/Attestation Pool API - Handle Electra attestations in the aggregating pool

mehdi-aouadi:
## PR Description
In preparation for the GET attestation from the aggregating attestation pool API.
Handle the Electra attestations when querying attestations from the aggregation pool

## Fixed Issue(s)

#8029 
## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -288,9 +295,35 @@ public synchronized List<Attestation> getAttestations(
...
+
+  private Predicate<Map.Entry<UInt64, Set<Bytes>>> getSlotFilter(final Optional<UInt64> maybeSlot) {
+    if (maybeSlot.isPresent()) {
+      return (entry) -> entry.getKey().equals(maybeSlot.get());
+    } else {
+      return (entry) ->
+          recentChainData
+              .getCurrentSlot()
+              .map(currentSlot -> entry.getKey().equals(currentSlot))
+              .orElse(true);
```
rolfyone: this is actually a non trivial change, the only time getCurrentSlot will return empty is if recentChainData isn't initialised...
i'd like to know why we're thinking about this change i think. there's no test justification or anything.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPoolTest.java
```diff
@@ -471,18 +472,56 @@ public void getAttestationsForBlock_shouldNotAddAttestationsFromWrongFork() {
...
+            mockRecentChainData,
+            new NoOpMetricsSystem(),
+            DEFAULT_MAXIMUM_ATTESTATION_COUNT);
+    // Adding a phase0 attestation to the aggregation pool
+    final Spec phase0Spec = TestSpecFactory.createMinimalPhase0();
+    when(mockedSpec.atSlot(any())).thenReturn(phase0Spec.getGenesisSpec());
+    final DataStructureUtil phase0DataStructureUtil = new DataStructureUtil(phase0Spec);
+    final AttestationData phase0AttestationData =
+        phase0DataStructureUtil.randomAttestationData(SLOT.minus(4));
+    addAttestationFromValidators(aggregatingPool, phase0AttestationData, phase0Spec, 4, 5);
```
rolfyone, line 503: this is interesting - beacon-api kind of implied in a new issue that the better behaviour woudl be to convert any non electra attestations to electra and return them too...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
mehdi-aouadi: Ok so I put back the old slot filtering logic and left the current slot usage to determine if we must filter Electra attestations. To summarise, here is the implemented logic:
- If the slot is provided, we filter the attestation based on it otherwise we retrieve all attestations
AND
- If the slot is provided, we use it to determine the milestone otherwise we use the current slot and finally we fallback to genesis: This let us know if we reached Electra and hence should filter Electra attestations only

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPoolTest.java
mehdi-aouadi: I left a comment in the issue.
IMO that won't be inline with the decision of dropping pre Electra attestations once we reach the fork. The first Electra block will contain Electra attestations only and the GET request should do the same.
Plus I'm not sure adding complexity to handle and edge case is worth it, I'm wondering what are the real world use cases and if it's worth the pain

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -269,16 +269,24 @@ public synchronized SszList<Attestation> getAttestationsForBlock(
...
                 .orElse(true);
 
+    final SchemaDefinitions schemaDefinitions = getSchemaDefinitions(maybeSlot);
+
+    final SszListSchema<Attestation, ?> attestationsSchema =
+        schemaDefinitions.getBeaconBlockBodySchema().getAttestationsSchema();
+
+    final boolean requestRequiresAttestationWithCommitteeBits =
+        schemaDefinitions.getAttestationSchema().requiresCommitteeBits();
+
```
rolfyone: this seems like a lot of work to know we're in electra... 
We're getting schemadefinitions, possibly creating them every slot, all to just check we're in electra milestone or beyond, i think we might want to think about what we're doing here in a synchronized function that will be used a bunch.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPoolTest.java
mehdi-aouadi: Well, we agreed to keep filtering Electra attestations only. The issue is closed

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
mehdi-aouadi: right, made the schema definitions retrieval synchronised and did the same for the already existing attestation for blocks

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -288,9 +296,23 @@ public synchronized List<Attestation> getAttestations(
         .filter(filterForCommitteeIndex)
         .flatMap(MatchingDataAttestationGroup::stream)
         .map(ValidatableAttestation::getAttestation)
+        .filter(
+            attestation ->
+                attestation.requiresCommitteeBits() == requestRequiresAttestationWithCommitteeBits)
+        .limit(attestationsSchema.getMaxLength())
```
tbenr: why this limit? I could see the reason but seems like it comes from nowhere

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -288,9 +296,23 @@ public synchronized List<Attestation> getAttestations(
...
 
+  private synchronized SchemaDefinitions getSchemaDefinitions(final Optional<UInt64> maybeSlot) {
+    if (maybeSlot.isPresent()) {
+      return spec.atSlot(maybeSlot.get()).getSchemaDefinitions();
+    } else if (recentChainData.getCurrentSlot().isPresent()) {
+      return spec.atSlot(recentChainData.getCurrentSlot().get()).getSchemaDefinitions();
+    } else {
+      return spec.getGenesisSchemaDefinitions();
+    }
+  }
```
tbenr: I think here you can have a nice functional version: like 
slot = maybeSlot or recentChainData.getCurrentSlot() or GENESIS_SLOT
spec.atSlot(slot).getSchemaDefinitions()

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
tbenr: It should be called once per slot from every connected validator client (when they are aggregating). maybe not that big deal? need to think a bit more.
A shortcut could be that if `maybeCommitteeIndex` is present we know we are in ELECTRA (which speeds up the aggregation flow post electra)

rolfyone: I think at the very least we should probably be using the schema definition cache and avoiding potentially creating it all continually...

rolfyone: ok i think we're ok here actually, its created as the specVersion is created. this should be ok.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
rolfyone: also this is only used inside synchronized functions already so we dont really need 'synchronized' on this fn.
i'd suggest:
```
  private SchemaDefinitions getSchemaDefinitions(final Optional<UInt64> maybeSlot) {
    final UInt64 slot = maybeSlot.orElse(recentChainData.getCurrentSlot().orElse(UInt64.ZERO));
    return spec.atSlot(slot).getSchemaDefinitions();
  }
```

rolfyone: actually given that above, i'd delete the function, put the call in `getAttestationsForBlock` back, then make getAttestations do its own slot computation
```
    final UInt64 slot = maybeSlot.orElse(recentChainData.getCurrentSlot().orElse(UInt64.ZERO));

    final SchemaDefinitions schemaDefinitions = spec.atSlot(slot).getSchemaDefinitions();
```
as its far more clear.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
mehdi-aouadi: My bad, removed it

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
mehdi-aouadi: thanks for the commit, looks cleaner now

tbenr (approved):
LGTM

PR #8437, opened 2024-07-10
PART 2 - GET/Attestation Pool API - Retrieve attestations and metadata

mehdi-aouadi:
## PR Description
Rebased on top of #8431
Retrieve attestations and metadata for the GET attestations from aggregation pool API

## Fixed Issue(s)

#8029 
## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
@@ -15,7 +15,6 @@ package tech.pegasys.teku.api;
 
 import static tech.pegasys.teku.statetransition.validatorcache.ActiveValidatorCache.TRACKED_EPOCHS;
 
-import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
```diff
@@ -89,13 +94,30 @@ public NodeDataProvider(
...
+      final Optional<UInt64> maybeSlot, final Optional<UInt64> maybeCommitteeIndex) {
+    return lookupMetaData(
+        attestationPool.getAttestations(maybeSlot, maybeCommitteeIndex), maybeSlot);
+  }
+
+  private ObjectAndMetaData<List<Attestation>> lookupMetaData(
+      final List<Attestation> attestations, final Optional<UInt64> maybeSlot) {
+    Preconditions.checkArgument(
+        !attestations.isEmpty() || maybeSlot.isPresent(),
+        "Unable to determine spec version. No slot parameter provided and no attestations found");
```
tbenr: humm.. so the API will fail if we have no attestation and no slot as parameter. I wonder if we can default to PHASE0 in that case.
I don't like to fail in that case just because we can't decide the milestone, which is actually unused because the array will be empty.

mehdi-aouadi: Good catch, ended up using the recent chain data: try getting the current slot otherwise fallback to 0. The same logic we implemented when retrieving the attestations from the pool.
Added unit tests for those new use cases too.

tbenr (approved):
LGTM

PR #8438, opened 2024-07-10
PART 3 - GET/Attestation Pool API - Add API interface

mehdi-aouadi:
## PR Description
Add the new `GetAttestationV2` `/eth/v2/beacon/pool/attestations`

## Fixed Issue(s)

#8029 

## Documentation

## Changelog


data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/schema/getPoolAttestationsV2Response.json
```diff
@@ -0,0 +1,31 @@
...
+          "$ref" : "#/components/schemas/AttestationPhase0"
+        }, {
+          "$ref" : "#/components/schemas/AttestationPhase0"
+        }, {
+          "$ref" : "#/components/schemas/AttestationPhase0"
+        }, {
+          "$ref" : "#/components/schemas/AttestationPhase0"
+        }, {
+          "$ref" : "#/components/schemas/AttestationElectra"
+        } ]
```
lucassaldanha: I am currently working on a way to fix this issue where we get the duplicate schemas in our openapi def.

lucassaldanha: Actually @rolfyone fixed it here https://github.com/Consensys/teku/pull/8440

mehdi-aouadi: Fixed, thanks

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttestationsV2Test.java
```diff
@@ -0,0 +1,120 @@
...
+    List<Attestation> attestations;
+    if (specMilestone.isGreaterThanOrEqualTo(ELECTRA)) {
+      attestations =
+          List.of(
+              dataStructureUtil.randomElectraAttestation(),
+              dataStructureUtil.randomElectraAttestation());
+    } else {
+      attestations =
+          List.of(dataStructureUtil.randomAttestation(), dataStructureUtil.randomAttestation());
+    }
```
rolfyone: wait, so we cant use `dataStrucutreUtil.randomAttestation` after electra? thats weird....

lucassaldanha: I understand that you could if you were initializing datastructureUtil until with an Electra spec, because `randomAttestation()` uses `spec.getGenesisSchemaDefinitions()`.

But in the case you have datastructureUtil using a pre-Electra spec, you might need a way to say that you want Phase0 or Electra attestations.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttestationsV2Test.java
```diff
@@ -0,0 +1,120 @@
...
+import tech.pegasys.teku.infrastructure.json.JsonTestUtil;
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.TestSpecContext;
+import tech.pegasys.teku.spec.TestSpecInvocationContextProvider;
+import tech.pegasys.teku.spec.datastructures.metadata.ObjectAndMetaData;
+import tech.pegasys.teku.spec.datastructures.operations.Attestation;
+import tech.pegasys.teku.spec.schemas.SchemaDefinitionCache;
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+
+@TestSpecContext(allMilestones = true)
```
rolfyone: given this is a unit test to show something specifically works, it creates a ton of noise and maintenance to run this for all milestones. i'd suggest something like phase0 and electra, or deneb and electra, and allow us to maintain a lot less json. Or alternatively maybe we need to think more about why we're comparing json, as really we care more about the structure, and the level of data here is not really the point of the test.

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
```diff
@@ -813,6 +813,16 @@ public Attestation randomAttestation() {
...
+  public Attestation randomElectraAttestation() {
+    return getElectraSchemaDefinitions(randomSlot())
+        .getAttestationSchema()
+        .create(
+            randomBitlist(),
+            randomAttestationData(),
+            randomSignature(),
+            this::randomCommitteeBitvector);
+  }
+
```
rolfyone: i commented above, but would prefer this did not exist.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttestationsV2Test.java
rolfyone: create dataStructureUtil with electra spec and use that, but we can't follow this model and have our templating continue to work happily, please don't introduce this, it wont work.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttestationsV2Test.java
mehdi-aouadi: right, limited it to Phase0 and Electra since there are only two really different json structures

ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
mehdi-aouadi: Removed it since the DataStructure util uses the spec it's been initialised with so there is no need for this in the end

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttestationsV2Test.java
mehdi-aouadi: Right, the DataStructureUtil uses the spec it's been initialised with and in this case it will use the correct one. So no need for this distinction in the end. Removed it

data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
```diff
@@ -89,13 +94,30 @@ public NodeDataProvider(
...
+      final Optional<UInt64> maybeSlot, final Optional<UInt64> maybeCommitteeIndex) {
+    return lookupMetaData(
+        attestationPool.getAttestations(maybeSlot, maybeCommitteeIndex), maybeSlot);
+  }
+
+  private ObjectAndMetaData<List<Attestation>> lookupMetaData(
+      final List<Attestation> attestations, final Optional<UInt64> maybeSlot) {
+    Preconditions.checkArgument(
+        !attestations.isEmpty() || maybeSlot.isPresent(),
+        "Unable to determine spec version. No slot parameter provided and no attestations found");
```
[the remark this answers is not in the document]
rolfyone: oh thats kinda a neat tidbit @lucassaldanha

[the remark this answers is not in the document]
mehdi-aouadi: it's outdated now since I removed this check but that was a good catch

tbenr (commented):
as discussed in call,  the current commiteeIndex filter requires an update. We can leverage the committee filtering we already have for the aggregation flow, but requires to be updated to cover the phase0 because it has been design for electra only.

mehdi-aouadi:
> as discussed in call, the current commiteeIndex filter requires an update. We can leverage the committee filtering we already have for the aggregation flow, but requires to be updated to cover the phase0 because it has been design for electra only.

Done, made the filter applied differently based on the milestone. Pre electra we apply the filter to the index at the attestation data level and post Electra we rather apply it to the committee bits

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/AggregatingAttestationPool.java
```diff
@@ -273,32 +275,50 @@ public synchronized List<Attestation> getAttestations(
...
+    // applied to the index at the attestation data level)
+    // Post Electra the index is always set to 0 at the attestation data level (the filter
+    // is rather applied to the committee bits)
+    final Predicate<MatchingDataAttestationGroup> filterForCommitteeIndex =
+        (group) ->
+            requestRequiresAttestationWithCommitteeBits
+                || maybeCommitteeIndex
+                    .map(index -> group.getAttestationData().getIndex().equals(index))
+                    .orElse(true);
+
```
tbenr: mmm.. why we need this filtering here? Can't we just filter for both pre and post electra cases in `matchingDataAttestationGroup.stream`? Maybe I'm missing something.

mehdi-aouadi: We need this filtering to handle the pre Electra attestation since the filtering must done on the attestation data index in that case.
Unless you mean that we could refactor the `maybeFilterOnCommitteeIndex` in `MatchingDataAttestationGroup` (which currently filters on the committee bits) and make it filter on the attestation data index if the committee bits is empty?

mehdi-aouadi: I pushed a commit where I made the maybeFilterOnCommitteeIndex handle the pre Electra attestations and hence simplifies the logic. Let me know if that's what you had in mind

mehdi-aouadi: I refactored the filtering as discussed

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroup.java
```diff
@@ -149,18 +149,11 @@ public Stream<ValidatableAttestation> stream(final Optional<UInt64> committeeInd
 
   public Stream<ValidatableAttestation> stream(
       final Optional<UInt64> committeeIndex, final boolean requiresCommitteeBits) {
-    // Group Attestation type mismatch
-    if (requiresCommitteeBits != includedValidators.requiresCommitteeBits()) {
+    if (noMatchingAttestations(committeeIndex, requiresCommitteeBits)) {
       return Stream.empty();
+    } else {
+      return StreamSupport.stream(spliterator(committeeIndex), false);
```
tbenr: You don't need `else` ;-)

mehdi-aouadi: Done

tbenr (approved):
LGTM

PR #8441, opened 2024-07-11
Add flag to make state pruning limit configurable

gfukushima:
## PR Description
Follow up fix to make the state pruning limit configurable as I forgot to add this flag to the CLI params in the previous PR.
Also introducing a hard limit of 100 for now just to ensure users don't misconfigure this limit. Still performing some test on what might be a ideal limit for the pruner. 

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
```diff
@@ -289,7 +289,7 @@ public Builder statePruningInterval(final Duration statePruningInterval) {
     }
 
     public Builder statePruningLimit(final int statePruningLimit) {
-      if (statePruningLimit < 0) {
```
lucassaldanha, line 292: NIT: I'd add a static variable named `MAX_STATE_PRUNING_LIMIT` with a comment about the value 100 being an arbitrary choice. We recently had users questioning some of our choices for default values and limits, and I think making it explicit that it was arbitrary can help us remember in the future.

teku/src/main/java/tech/pegasys/teku/cli/options/BeaconNodeDataOptions.java
```diff
@@ -76,6 +76,16 @@ public class BeaconNodeDataOptions extends ValidatorClientDataOptions {
   private long statePruningIntervalSeconds =
```
lucassaldanha, line 76: I know this isn't part of this PR but in our recent call you mentioned the limit being in minutes, not seconds. Just checking that this is actually what we want.

gfukushima: Yeah, I thought it was defined in minutes when we had the conversation but it's actually following the convention used for the other pruners which is seconds. Could easily change that if we think it makes sense but it would be a little confusing for users to have the state pruner using a different unit than the other pruners

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
gfukushima: I'm still testing this PR to get a better sense of what is a more sensible limit.

teku/src/main/java/tech/pegasys/teku/cli/options/BeaconNodeDataOptions.java
rolfyone: seconds is ok, but hten we'd want a sanity check. i certainly wouldnt want this happening more often than every slot (ish) for example - i'd potentially set a min of 30 seconds and a max of a day or something...

gfukushima: added sanity check for this

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
gfukushima: done and tests have confirmed that 100 limit is safe (doesn't show any signs of degradation) in the current instance we use for mainnet archive

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
```diff
@@ -284,12 +287,17 @@ public Builder statePruningInterval(final Duration statePruningInterval) {
       if (statePruningInterval.isNegative() || statePruningInterval.isZero()) {
         throw new InvalidConfigurationException("Block pruning interval must be positive");
       }
+      if (statePruningInterval.toSeconds() < 30L
+          || statePruningInterval.toSeconds() > Duration.ofDays(1).toSeconds()) {
+        throw new InvalidConfigurationException(
+            "Block pruning interval must be a value between 30 seconds and 1 day");
+      }
```
lucassaldanha, line 294: Are we missing unit tests for `StorageConfiguration` for the desired checks?

gfukushima: Added some unit test just to make sure that we test those limits and sanity checks in the tests as well

lucassaldanha (approved):

PR #8444, opened 2024-07-15
Stable Containers - EIP-7495

### LINKED ISSUE #8428: Implement general Stable Container support - EIP-7495
https://github.com/ethereum/consensus-specs/pull/3789

tbenr:
fixes #8428 

Introduces new interfaces:
- `SszStableContainer`
- `SszStableContainerSchema`
- `SszProfile`
- `SszProfileSchema`

which are based on new `SszStableContainerBase` and `SszStableContainerBaseSchema` interfaces representing the common base  functionalities.
Corresponding `Abstract` and `Impl` classes implement the logic.

### Key points
- these containers have a `maxFieldCount` (maximum future field count): the backing tree will always have that size despite the "actual" field count.
- fields can be required, optional or disallowed (disallowed fields are the ones that are not specified as required nor as optional)
- `StableContainers` have all fields as optional, `Profile` can have a mix of required and optional fields.
- To represent which fields are active (required+optionals) an `SszBitvector` is used (sized as max future field count). It will be used in serialization as well as merkelization.
- due to optionals and (and potentially disabled) fields, these containers are sparse, which means that `size() = last_field_index + 1` does not hold anymore.
-  There is a difference between `maxLength` and `maxChunks`: the first is the possible maximum number of fields (required+optionals) the latter is related to the backing tree, so it is calculated based on `maxFieldCount`.
- `getOptional` has been introduced to support optionality.  Similarly we have `createFromOptionalFieldValues` on the schema side to handle optionality.
- then getting fields values, the method can now throw `NoSuchElementException` along with existing `IndexOutOfBoundsException`. The general logic is:
  - `NoSuchElementException` is thrown whenever a field was allowed but (optional) but was not actually specified.
  - `IndexOutOfBoundsException` is thrown whenever a field was not allowed (nor optional nor required), which also covers the case the index is truly off.
  This allows to have nice implementation of the `getOptional` which simply delegate to `get` catching `NoSuchElementException` and returning `Optional.empty`.
- `merkelization` works the same for both `StableContainers` and `Profiles`.
- `serialization` are different. This polymorphism is achieved via abstract methods `AbstractSszStableContainerBaseSchema`.

### High Level Usage:
to create a `StableContainerSchema` you need:
- `definedChildrenNamedSchemas`: a list of `NamedSchema` representing the defined schemas.
- `maxFieldCount`: theoretical future maximum field count.

to create a `ProfileSchema` you need:
- `stableContainerSchema`: a `SszStableContainer` to create the profile schema against
- `requiredFieldIndices`: a set of indices to indicate which fields are mandatory
- `optionalFieldIndices`: a set of indices to indicate which fields are optional

`AbstractSszStableContainerBaseSchemaTest` showcases how to use them.

### What is missing
- Mutability is only implemented for Profiles with no optional fields. Optionality in mutation requires some additional works.
  - a new interface with optional-enabled `set` methods.
  - handling the expansion\contraction of the number of fields (and good handling of the "sparse" index set)
- benchmarking?
- allow to specify optional and required fields by name when creating Profiles?

## Documentation

## Changelog


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszStableContainerBaseSchema.java
```diff
@@ -0,0 +1,774 @@
...
+  }
+
+  /**
+   * Delegates deserialization to deriving classes: Profile and StableContainer have different
+   * serialization\deserialization rules
+   */
+  abstract SszBitvector sszDeserializeActiveFieldsTree(final SszReader reader);
+
+  @Override
+  public TreeNode sszDeserializeTree(final SszReader reader) {
```
Nashatyrev, line 434: nit: It looks like `sszSerialize/DeserializeTree()` have a lot of common logic with `AbstractSszContainerSchema`. Maybe we could extract this logic to some util class?

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/impl/AbstractSszMutableComposite.java
```diff
@@ -91,7 +92,15 @@ public void set(final int index, final SszChildT value) {
 
     childrenChanges.put(index, createChangeRecordByValue(immutableValue));
 
-    sizeCache = index >= sizeCache ? index + 1 : sizeCache;
+    if (!(backingImmutableData instanceof SszStableContainerBase)) {
```
Nashatyrev: Could we extract some protected method like `calcNewSize()` and override it in `SszMutableStableContainerBaseImpl` ?
That `instanceof` check on descendant class looks a bit hacky

Nashatyrev (approved):
LGTM, just a couple of minor notes. 
Many design question we discussed and agreed in our chat before.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/impl/AbstractSszMutableComposite.java
tbenr: Nice idea! Implemented.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/impl/AbstractSszStableContainerBaseSchema.java
tbenr: I did some refactor, checkout https://github.com/Consensys/teku/pull/8444/commits/5b13a1e9708d694af8c829287cdef56d525ba788

tbenr: I decided to cut it at that point even I could have continued along that direction it even further.

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/SszStableContainerBase.java
```diff
@@ -0,0 +1,39 @@
...
+    } catch (final NoSuchElementException __) {
+      return Optional.empty();
+    }
+  }
+
+  @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
+  // container is heterogeneous by its nature so making unsafe cast here
+  // is more convenient and is not less safe
+  default <C extends SszData> Optional<C> getAnyOptional(final int index) {
+    return (Optional<C>) getOptional(index);
```
rolfyone, line 37: could check instanceof here before cast....

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/impl/AbstractSszMutableComposite.java
```diff
@@ -239,6 +239,15 @@ public SszMutableComposite<SszChildT> createWritableCopy() {
    */
   protected abstract void checkIndex(int index, boolean set);
 
+  /**
+   * Determines the new size given the field has been set or updated
+   *
+   * @param index of the field set or updated
+   */
+  protected int calcNewSize(final int index) {
```
rolfyone, line 247: should probably just call this `calculateNewSize` rather than abbreviate...

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/impl/SszProfileImpl.java
```diff
@@ -0,0 +1,63 @@
...
+          memberValues[i].getSchema().equals(type.getChildSchema(i)),
+          "Wrong child schema at index %s. Expected: %s, was %s",
+          i,
+          type.getChildSchema(i),
+          memberValues[i].getSchema());
+    }
+  }
+
+  private static IntCache<SszData> createCache(final SszData... memberValues) {
+    ArrayIntCache<SszData> cache = new ArrayIntCache<>(memberValues.length);
```
rolfyone: final...

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/SszStableContainerBaseSchema.java
```diff
@@ -0,0 +1,76 @@
...
+   */
+  int getMaxFieldCount();
+
+  List<NamedSchema<?>> getChildrenNamedSchemas();
+
+  SszBitvectorSchema<SszBitvector> getActiveFieldsSchema();
+
+  TreeNode createTreeFromOptionalFieldValues(List<Optional<? extends SszData>> fieldValues);
+
+  default C createFromOptionalFieldValues(final List<Optional<? extends SszData>> fieldValues) {
```
rolfyone, line 56: maybe not finals in interfaces - i know some people feel strongly about finals in interfaces...

tbenr: In default methods we have actual code in it, so we should have finals (and I think our errorprone checks that too).
I agree that when we only declare method (no code) we can leave finals away

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/impl/AbstractSszMutableComposite.java
tbenr: problem is that in the ssz module seems like "calc" is the normal:
image

I wanted to maintain the style here. WDYT?

rolfyone: ok i guess but i do hate abbreviations...

rolfyone (approved):
LGTM

lucassaldanha:
Code has been pushed to a branch `stable-containers` in this repo. Will close the PR for now.

PR #8450, opened 2024-07-17
Use IPv6 address of discovery nodes when applicable

StefanBratanov:
## PR Description
If local discovery node supports IPv6, use the IPv6 address of the discovery node (if it is available). The fallback will always be IPv4 (this could be the case in dual-stack mode where the discovery node does not advertise IPv6 address). 

## Fixed Issue(s)
fixes last of the tasks in #8069 

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #8424, opened 2024-07-05
Add the new PostAttestation v2 API


mehdi-aouadi:
Add new `PostAttestation` V2 API to handle electra attestations
[Spec PR link](https://github.com/ethereum/beacon-APIs/pull/448)

PR #8456, opened 2024-07-19
8424 attestations handler

mehdi-aouadi:
## PR Description
Add `PostAttestationsRequest` for the new post attestations V2 API (remote validator api handler).
The new `sendSignedAttesationsV2` isn't used in production code yet. A hidden option will be added in a separate PR in order to activate it progressively (same way for block v3).
The `sendSignedAttestations` is deprecated and should be progressively replaced by the new `sendSignedAttestationsV2`.

## Fixed Issue(s)

#8424 

## Documentation

## Changelog


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedAttestationsRequest.java
```diff
@@ -13,29 +13,48 @@
...
 
 public class SendSignedAttestationsRequest extends AbstractTypeDefRequest {
+
   public SendSignedAttestationsRequest(
       final HttpUrl baseEndpoint, final OkHttpClient okHttpClient) {
     super(baseEndpoint, okHttpClient);
   }
 
-  public List<SubmitDataError> submit(final List<Attestation> attestations) {
+  public List<SubmitDataError> submit(final List<Attestation> attestations, final Spec spec) {
```
rolfyone: i'm lost on prs they all look so similar, but in teh same way i think maybe spec in the constructor is probably better....

mehdi-aouadi: It's the same pattern but different APIs that's why they look the same :)
Anyway I refactored the request to pass the spec in the constructor. Left the attestations to the submit method though because we have to check if the list is empty and get the first element to get the slot and the milestone otherwise...

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedAttestationsRequest.java
```diff
@@ -13,29 +13,51 @@
...
   public List<SubmitDataError> submit(final List<Attestation> attestations) {
     if (attestations.isEmpty()) {
       return Collections.emptyList();
     }
+    // Use attestation v2 api post Electra only. This logic can be removed once we reach the Electra
+    // milestone
+    final SpecMilestone specMilestone =
+        spec.atSlot(attestations.getFirst().getData().getSlot()).getMilestone();
+    if (specMilestone.isGreaterThanOrEqualTo(SpecMilestone.ELECTRA)) {
+      return submitPostElectra(attestations, specMilestone);
```
tbenr, line 55: here I remember the discussion we had about versioning data structures. My position was that we need to send always ELECTRA if data structure remains unchanged in next forks.

tbenr (commented):
generally ok, need to clarify one thing

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSignedAttestationsRequest.java
tbenr: Ok we clarified this

tbenr (approved):
LGTM

Issue #8449, opened 2024-07-17
Investigate compaction on archived pruned nodes


gfukushima:
At the moment stage we're not seeing the expected impact in the database (significant reduction in size) when the finalized state pruner is enabled.

Looking at the logs in the leveldb LOG file we see compaction is being done but still the db directory has not presented any significant change.

gfukushima:
This lack of compaction was caused by the misconfiguration of the node, which was using `data-storage-archive-frequency=1` and had it accidentally changed but the dbversion is still leveldb tree which doesn't have a implementation for the pruner at this stage.

PR #8459, opened 2024-07-22
Tide up state pruner validation

gfukushima:
## PR Description
Removes the value "0" from the possible range of retained slots. Adds validation to prevent state pruner to be created when the database version is on leveldb tree. 
We already have a validation at the storage configuration level that checks for the frequency (which is ultimately how leveldb tree gets set) but if we change the frequency and leveldb tree is still used. We don't want to support that at the moment. 
  
## Fixed Issue(s)

Part of #8449

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
```diff
@@ -275,7 +275,7 @@ public Builder blobsPruningLimit(final int blobsPruningLimit) {
     }
 
     public Builder retainedSlots(final long retainedSlots) {
-      if (retainedSlots < -1) {
```
lucassaldanha, line 278: Are we missing a test case for this scenario? How come we changed the validation and no tests had to be updated?

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -117,24 +118,28 @@ protected SafeFuture<?> doStart() {
...
-                            storagePrunerAsyncRunner,
-                            config.getStatePruningInterval(),
-                            config.getRetainedSlots(),
-                            config.getStatePruningLimit(),
-                            "state",
-                            pruningTimingsLabelledGauge,
-                            pruningActiveLabelledGauge));
+                  && config.getRetainedSlots() > 0) {
+                if (config.getDataStorageCreateDbVersion() == DatabaseVersion.LEVELDB_TREE) {
+                  LOG.warn("State pruning is not supported with leveldb_tree database.");
```
rolfyone: it's basically invalid configuration exception. i think we just barf here, and make the user fix the configuration... I think potentially just throwing an InvalidConfigurationException, which iirc would terminate the client ultimately...
My reasoning is basically that this will never work...
Come to think of it the historical batch builder and this pruner are explicitly non compatible too, not sure if we've handled that...

gfukushima: It makes sense to bail at this stage, agreed. Changed it to do so.

storage/src/main/java/tech/pegasys/teku/storage/server/StorageConfiguration.java
gfukushima: Added a test to ensure a negative value here is considered invalid

rolfyone (approved):

PR #8461, opened 2024-07-22
removed requirement for block headers in voluntary exit

rolfyone:
Instead of using the block header we can compute the current slot from genesis data.

## Documentation

## Changelog


tbenr (approved):
LGTM but `InvalidConfigurationException` message needs to be updated

PR #8464, opened 2024-07-23
Add GetAggregateAttestationV2 API

### LINKED ISSUE #8420: Add the new GetAggregateAttestation V2 API
Add new `GetAggregateAttestation` v2 API to handle Electra attestations
[Spec PR](https://github.com/ethereum/beacon-APIs/pull/447)

mehdi-aouadi:
## PR Description
Add `GetAggregateAttestationV2` API

## Fixed Issue(s)

fixes #8420

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/validator/GetAggregateAttestationV2.java
```diff
@@ -0,0 +1,144 @@
...
+                          request.header(
+                              HEADER_CONSENSUS_VERSION,
+                              Version.fromMilestone(attestationAndMetaData.getMilestone()).name());
+                          return AsyncApiResponse.respondOk(attestationAndMetaData);
+                        })
+                    .orElseGet(AsyncApiResponse::respondNotFound)));
+  }
+
+  @SuppressWarnings("unchecked")
+  private static SerializableTypeDefinition<ObjectAndMetaData<Attestation>> getResponseType(
```
lucassaldanha, line 104: This seems to be the exact same response as `GetBlockAttestationsV2`. We should probably create a single definition and re-use in both places?

lucassaldanha: I have a feeling we will have this `oneOf(electra, phase0)` logic in other places as well...

mehdi-aouadi: Well it's not really the same response since one is responding with a single attestation and the other with a list.
I refactored the code to reuse the `Attestation` type definition but it would be ugly to generalise the `getResponseType` to make it handle lists and single elements at the same time.
I put the `Attestation` type definition in `BeaconRestApiTypes` which seemed to be the best place but not sure about that...

lucassaldanha: Ah my bad, I looked into it and for some reason thought it was a list as well!

lucassaldanha: Not sure if `BeaconRestApiTypes` is the right place for it, but nothing that we can't move later if we don't like it I guess.

lucassaldanha (approved):

PR #8466, opened 2024-07-24
Add GetAttesterSlashingV2

gfukushima:
## PR Description
This PR introduces the skeleton of the new method GetAttesterSlashingV2. Note though that as of now this differ minimally from the V1 as Teku currently doesn't not make a distinction of Electra.IndexedAttestation and Phase0.IndexedAttestation since the only difference is the size limit of the `attesting_indices` array. 
This implementation does not make us 100% complaint with the spec but should result in the right json result after all.

## Fixed Issue(s)

Part of 
 
## Documentation

## Changelog


data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttesterSlashingV2Test.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Test
+  void metadata_shouldHandle400() throws JsonProcessingException {
+    verifyMetadataErrorResponse(handler, SC_BAD_REQUEST);
+  }
+
+  @Test
+  void metadata_shouldHandle500() throws JsonProcessingException {
+    verifyMetadataErrorResponse(handler, SC_INTERNAL_SERVER_ERROR);
+  }
```
rolfyone: should be able to test handling 200 even without implementation data :)

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/schema/GetPoolAttesterSlashingsV2Response.json
```diff
@@ -0,0 +1,29 @@
...
+      ]
+    },
+    "data": {
+      "type": "object",
+      "oneOf": [
+        {
+          "type": "array",
+          "items": {
+            "$ref": "#/components/schemas/AttesterSlashing"
+          }
```
rolfyone, line 24: see if we can get attester slashing for electra to work here too...

also no mention of the `Eth-Consensus-Version` header in the response...

gfukushima: Eth-Consensus-Version will be added once the implementation is working. It's prob going to come down to pulling the milestone from one of the AttesterSlashing data.

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttesterSlashingV2Test.java
gfukushima: added a 200 for phase 0 for now. Will add electra in the next PR

data/beaconrestapi/src/integration-test/resources/tech/pegasys/teku/beaconrestapi/beacon/schema/GetPoolAttesterSlashingsV2Response.json
rolfyone: its visible in metadata, we should just add it to the definition so it comes up...

rolfyone: actually is it automatic... i cant find it in any gets... hrm...

rolfyone (approved):

PR #8448, opened 2024-07-16
Add PostAttestationV2 API

mehdi-aouadi:
## PR Description
- Add PostAttestationV2 API
- Handle arrays with different item types deserialisation
- Add a header based schema selector

## Fixed Issue(s)

#8424 

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #8478, opened 2024-07-26
add Electra attester slashing schema

mehdi-aouadi:
## PR Description
Separate the `AttesterSlashing` and `IndexedAttestation` schemas based on the milestone: `Phase0` and `Electra`

## Documentation

## Changelog


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostAttesterSlashing.java
```diff
@@ -26,29 +26,31 @@
...
-  public PostAttesterSlashing(final DataProvider dataProvider, final Spec spec) {
-    this(dataProvider.getNodeDataProvider(), spec);
+  public PostAttesterSlashing(
+      final DataProvider dataProvider, final SchemaDefinitionCache schemaDefinitionCache) {
+    this(dataProvider.getNodeDataProvider(), schemaDefinitionCache);
   }
 
-  public PostAttesterSlashing(final NodeDataProvider provider, final Spec spec) {
+  public PostAttesterSlashing(
+      final NodeDataProvider provider, final SchemaDefinitionCache schemaDefinitionCache) {
```
gfukushima: Do we see value in refactoring to remove the `spec` object and use the `SchemaDefinitionCache` in what should be a soon deprecated API? If so I can add this to my post PR.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttesterSlashingsV2.java
```diff
@@ -68,20 +76,49 @@ public void handleRequest(final RestApiRequest request) throws JsonProcessingExc
...
         .withField("version", MILESTONE_TYPE, ObjectAndMetaData::getMilestone)
-        .withField(
-            "data",
-            listOf(attesterSlashingSchema.getJsonTypeDefinition()),
-            ObjectAndMetaData::getData)
+        .withField("data", oneOfTypeDefinition, ObjectAndMetaData::getData)
         .build();
   }
+
+  private static Predicate<List<AttesterSlashing>> electraAttesterSlashingsPredicate(
```
gfukushima, line 105: think the predicates can re-use the `SchemaDefinitionCache.milestoneAtSlot` and get rid of the `spec.atSlot`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttesterSlashing.java
```diff
@@ -31,9 +31,10 @@ public static class AttesterSlashingSchema
       extends ContainerSchema2<AttesterSlashing, IndexedAttestation, IndexedAttestation> {
 
     public AttesterSlashingSchema(
-        final IndexedAttestation.IndexedAttestationSchema indexedAttestationSchema) {
+        final IndexedAttestation.IndexedAttestationSchema indexedAttestationSchema,
+        final boolean isElectra) {
       super(
-          "AttesterSlashing",
+          isElectra ? "AttesterSlashingElectra" : "AttesterSlashingPhase0",
```
gfukushima, line 37: This is a much simpler solution for a problem that drove me crazy. I went to different paths to try and get this but came to the conclusion that my implementation added some large refactor to get this done in a more sustainable way. imo this is not a real issue since we don't distinguish the schemas at the API level due to the lack of support for the attribute `maxItems` atm.

gfukushima (commented):
this solve an issue that I had postponed for a little later. And it does with a lot more simplicity than my approach was going for. Although this is not a very extensible approach (for future forks) I don't think we going changing this attestation related classes so soon...

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostAttesterSlashing.java
mehdi-aouadi: Actually we don't need the `Spec` anymore since we know this API supports Pre Electra attester slashing only so we can't directly get the schema from the schema definition cache specifying the `PHASE0` mile stone. So yes, this could also be applied to the `PostAttesterSlashings` API as well, check [this PostAttestationV2 API PR](https://github.com/Consensys/teku/pull/8448) for reference

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v2/beacon/GetAttesterSlashingsV2.java
mehdi-aouadi: Good catch, refactored it to use the schema definition cache instead

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/AttesterSlashing.java
mehdi-aouadi: The problem is that without this change which allows us to differentiate the two (Phase0 and Electra) schemas we can't be aligned with the [API spec](https://github.com/ethereum/beacon-APIs/blob/860584e71109cc9dc34ab6b67c646ae9a818c07e/apis/beacon/pool/attester_slashings.v2.yaml#L25)
Both [`IndexedAttestation`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#indexedattestation) and the [`AttesterSlashing`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#attesterslashing) containers have been marked as modified in the Electra consensus spec update and that should be reflected in our code (the same way as the [`Attestation`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#attesterslashing) change). But since in this case, the change is the list size only, we decided to go with applying the container name and list size changes at runtime, which might look a bit hacky but changing it "properly" would be very expensive (changing the `IndexedAttestation` container implies a huge change cascade up down the `BeaconBlockBody`) and it's not worth the effort.
This is basically impacting an API which is not even public. IMO it's OK to keep it way for now

mehdi-aouadi: Plus, as you mentioned, we're not even supporting the `maxItem` constraints at the API level... So unless we get that done, this would be sufficient

mehdi-aouadi: BTW, I removed the size limit test since it's not really testing the API but rather the backend... We might add such test if we add the `maxItem` constraint

rolfyone (approved):
i do see this as a step down the road, but definitely in the right direction i think...
LGTM overall.

PR #8483, opened 2024-07-30
add get aggregate attestation request

mehdi-aouadi:
## PR Description
Add the `GetAggregateAttestation` Request to use the new `GetAggregateAttestationV2` API

## Fixed Issue(s)

#8420 

## Documentation

## Changelog


validator/api/src/main/java/tech/pegasys/teku/validator/api/ValidatorApiChannel.java
```diff
@@ -234,6 +240,9 @@ SafeFuture<Optional<BlockContainerAndMetaData>> createUnsignedBlock(
   SafeFuture<Optional<Attestation>> createAggregate(
       UInt64 slot, Bytes32 attestationHashTreeRoot, Optional<UInt64> committeeIndex);
 
+  SafeFuture<Optional<Attestation>> createAggregateV2(
+      UInt64 slot, Bytes32 attestationHashTreeRoot, UInt64 committeeIndex);
+
```
tbenr: I see your point introducing a V2 here, but maybe is better to revert the original `createAggregate` removing the `Optional` and add a `createAggregate` with the `committeeIndex`. So we can avoid polluting the channel with API versioning (there are none here) and still deprecate the old one. WDYT?

mehdi-aouadi: And what about the remote API call which in the `createAggregateV2` uses the new `GetAggregateAttestationV2` endpoint? That was the main purpose of this versioning. I thought we could completely remove the old `createAggregate` logic once the V2 endpoint is tested and confirmed (we could still add a fallback to the V1 endpoint if we don't want to completely remove it)...

tbenr: mmm k but this means that here at channel level we could leave with the Optional until we drop v1 support, when Optional will be removed. let's have a chat later

mehdi-aouadi: As discussed, I kept the validator channel interface as is and put the api versioning logic in the remote api handler. Let me know what you think about it now

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/GetAggregateAttestationRequest.java
```diff
@@ -0,0 +1,126 @@
...
+                GetAggregateAttestationResponse::setData)
+            .build();
+
+    this.responseHandler = new ResponseHandler<>(getAggregateAttestationTypeDef);
+  }
+
+  public Optional<Attestation> createAggregate(
+      final Bytes32 attestationHashTreeRoot, final UInt64 committeeIndex) {
+    final Map<String, String> queryParams = new HashMap<>();
+    queryParams.put(SLOT_QUERY_DESCRIPTION, slot.toString());
```
tbenr: `SLOT` not `SLOT_QUERY_DESCRIPTION`!

This should have been caught by a test...

validator/remote/src/integration-test/resources/responses/get_aggregate_attestation_responses/getAggregateAttestationELECTRA.json
```diff
@@ -0,0 +1,21 @@
+{
+  "version": "phase0",
```
tbenr: why `phase0` here?

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/GetAggregateAttestationRequest.java
mehdi-aouadi: Fixed and added a test!

validator/remote/src/integration-test/resources/responses/get_aggregate_attestation_responses/getAggregateAttestationELECTRA.json
mehdi-aouadi: Well this required some refactoring: Basically the milestone was ignored from the response (since the old api doesn't need it...).
I did the required changes which are basically returning an `ObjectAndMetaData<Attestation>` instead of `Attestation`. The additional meta data (basically the header) is simply ignored when not needed (the v1 api for example).
The test now checks the milestone

mehdi-aouadi: actually the milestone was ignored from the response, did some refactoring to check it too

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/CreateAggregateAttestationRequest.java
```diff
@@ -37,22 +37,20 @@
...
 
   public CreateAggregateAttestationRequest(
       final HttpUrl baseEndpoint,
       final OkHttpClient okHttpClient,
-      final SchemaDefinitionCache schemaDefinitionCache,
-      final Spec spec) {
+      final SchemaDefinitionCache schemaDefinitionCache) {
     super(baseEndpoint, okHttpClient);
     this.schemaDefinitionCache = schemaDefinitionCache;
-    this.spec = spec;
```
rolfyone: may as well pass spec in here like we were if we need it for milestone below... arguably could just pass spec milestone or just store specMilestone in the object, but its not super pretty passing spec into submit instead of constructor...

I do wonder if following this single request per object model ends up in all of our fields in constructor and just calling a submit with no parameters...
so if we've passed in spec and slot so we can store milestone, then the only real fields not in constructor would be attestationHashTreeRoot and committeeIndex...

mehdi-aouadi: In the end I liked putting everything in the constructor and then calling the submit with no parameters. Makes sense in this use case, will try doing the same for the other requests (in the other PRs). What do you think of it now?

rolfyone: yeh i think thats cleaner. just a shame we can't have a standard interface defined but i think if w'ere going 1 object per request this is pretty reasonable.

rolfyone (approved):
LGTM

PR #8485, opened 2024-07-31
move send signed contribution and proof to typedef client.

rolfyone:
Partially addresses #7762

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendContributionAndProofsRequest.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendContributionAndProofsRequest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright Consensys Software Inc., 2024
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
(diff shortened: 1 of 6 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendContributionAndProofsRequest.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  public void submit(final Collection<SignedContributionAndProof> signedContributionAndProofs) {
+    final List<SignedContributionAndProof> requestData =
+        new ArrayList<>(signedContributionAndProofs);
+    postJson(
+        SEND_CONTRIBUTION_AND_PROOF,
+        Collections.emptyMap(),
+        requestData,
+        listOf(requestData.getFirst().getSchema().getJsonTypeDefinition()),
```
tbenr, line 49: what if the collection is empty? getFirst will throw

tbenr: image

it is a bit annoying to resolve with the current design.

rolfyone: yep this is the same as previous, i had considered that and gone through a couple of iterations including a warning, but decided to leave the behaviour... i can check and see if we can ever call that with 0, but it has never seemed to cause us a problem so i was favouring old behaviour... will dig a tiny bit into old one, not sure we'd ever send signed with 0 (my initial code just returned)

lucassaldanha (approved):
LGTM

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendContributionAndProofsRequest.java
rolfyone: hopefully that warning doesnt spam - it shouldnt - it should be a bug if we get that far.

PR #8492, opened 2024-08-02
Updated createAggregateAttestation to use typedef client

rolfyone:
The concept of `? extends Attestation` got a little messy... wondering if we need to make it closer to the way blocks are, where Attestation is a class, but also not sure how much 'better' this would look without spiking it.

partially addresses #7762

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

Issue #7779, opened 2023-12-04
Update the block production workflow to use the block V3 API by default


mehdi-aouadi:
### Description
Currently the block production have two APIs: block v2 and block v3. In the current block V3 flow, the `blinded` param (similar to the block V2 API) is hard coded and set to true. The following changes are required to make the block V3 API the default:
- [x] #7889
  - [x] #7835
  - [x] Remove the hard coded blinded param in block V3
- [x] Enabled block V3 as the default
- [x] Test compatibility among CLs
- For next release:
- [x] Remove the deprecated block V2 methods - done via (#8508 and #8504)
- [x] Remove the `--Xblock-v3-enabled` option which is used to use block V3 apis (#8511)

These are the last steps of https://github.com/Consensys/teku/issues/7390

gfukushima:
Closing this as last PR was merged today.

PR #8493, opened 2024-08-02
Promote GetNewBlockV3 and enable by default

gfukushima:
## PR Description
This PR unhides the GetNewBlockV3 and enables it by default.

## Fixed Issue(s)

Part of #7779 

## Documentation

## Changelog


tbenr (commented):
Don't know if we really want to expose the `--block-v3-enabled`.. if we do that we theoretically add it in docs and it will be a problem when we deprecate the option completely (cose it will be a breaking change). I might be wrong, just posing the questions :)

gfukushima:
Make sense to me, I'll keep it as experimental them.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -139,10 +139,9 @@ public class ValidatorOptions {
   private int executorThreads = DEFAULT_VALIDATOR_EXECUTOR_THREADS;
 
   @Option(
-      names = {"--Xblock-v3-enabled"},
+      names = {"--block-v3-enabled"},
```
rolfyone: i think leaving it as a dev option thats just enabled by default was what i was thinking, i am hoping that this flag disappears once we prove we don't require it...

CHANGELOG.md
```diff
@@ -23,5 +23,6 @@
 - When `p2p-subscribe-all-subnets-enabled`, `p2p-peer-lower-bound` now defaults to 60 (previously 64), and `p2p-peer-upper-bound` now defaults to 80 (previously 100).
 - When `p2p-subscribe-all-subnets-enabled`,  (`Xnetwork-async-p2p-max-queue`, `Xnetwork-async-beaconchain-max-queue`, `Xp2p-batch-verify-signatures-queue-capacity`)  now default to 40_000 (previously 10_000)
 - When `p2p-subscribe-all-subnets-enabled`,  `Xvalidator-executor-max-queue-size`  now defaults to 60_000 (previously 40_000).
+- Enable [GetNewBlockV3](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/ValidatorRequiredApi/produceBlockV3) to default.
```
lucassaldanha: "by default"?

lucassaldanha (approved):

PR #8503, opened 2024-08-11
Cleaned up interfaces for remote request

rolfyone:
Also added some tests for minimal typedef functions which were missing. I'd like to ultimately move everything to this model where we're testing per class, but it's a larger cleanup, so here's a good sanity check point.

It's also interesting that we're not consistent with our error handling even in the first few that i've tested.

partially addresses #7762

## Documentation

## Changelog


validator/remote/src/integration-test/java/tech/pegasys/teku/validator/remote/typedef/handlers/GetProposerDutiesRequestTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.TestTemplate;
+import tech.pegasys.teku.api.exceptions.RemoteServiceNotAvailableException;
+import tech.pegasys.teku.ethereum.json.types.validator.ProposerDuties;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.TestSpecContext;
+import tech.pegasys.teku.spec.networks.Eth2Network;
+import tech.pegasys.teku.validator.remote.typedef.AbstractTypeDefRequestTestBase;
+
+@TestSpecContext(allMilestones = true, network = Eth2Network.MINIMAL)
```
gfukushima, line 34: probably not related to the work done in this PR but why are we testing this with allMilestone=true while a lot of the other tests only do with PHASE0? I'm guessing most of these don't require being tested against multiple milestones but I was wondering if we want to narrow down specific milestones or just test it with allMilestones more often

gfukushima (approved):
lgtm, left a question only

validator/remote/src/integration-test/java/tech/pegasys/teku/validator/remote/typedef/handlers/GetProposerDutiesRequestTest.java
rolfyone: in a lot the data doesnt change - still a thought in flight but we probably dont need to run all milestones for every endpoint... especially on the request side was my theory

PR #8509, opened 2024-08-12
moved SendSyncCommitteeRequest to typedefs

rolfyone:
This meant handling our failure list responses also, which will be needed for a couple of other interfaces also, so i've put the handling into a spot we can get to.

partially addresses #7762

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSyncCommitteeRequest.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSyncCommitteeRequest.java
@@ -18,7 +18,6 @@ import static tech.pegasys.teku.validator.remote.typedef.FailureListResponse.get
 
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
 import tech.pegasys.teku.infrastructure.json.types.DeserializableTypeDefinition;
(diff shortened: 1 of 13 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSyncCommitteeRequest.java
```diff
@@ -0,0 +1,51 @@
...
+public class SendSyncCommitteeRequest extends AbstractTypeDefRequest {
+  public SendSyncCommitteeRequest(final HttpUrl baseEndpoint, final OkHttpClient okHttpClient) {
+    super(baseEndpoint, okHttpClient);
+  }
+
+  public List<SubmitDataError> submit(final List<SyncCommitteeMessage> syncCommitteeMessages) {
+    if (syncCommitteeMessages.isEmpty()) {
+      return Collections.emptyList();
+    }
+    final Map<String, String> foo = Collections.emptyMap();
```
lucassaldanha: NIT: Did you mean to have this name for this variable? Maybe something like `emptyParams` makes a bit more sense?

lucassaldanha (approved):
LGTM. Just a comment about a variable name

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/SendSyncCommitteeRequest.java
rolfyone: oh thats a booboo - will fix that. I shouldnt need it I think.

PR #8511, opened 2024-08-13
Remove --xblock-v3-enabled and useBlindedBlock variable

### LINKED ISSUE #7779: Update the block production workflow to use the block V3 API by default

gfukushima:
## PR Description
This PR removes the experimental flag that allows users to enable or disable block production using the produceBlockV3 API. V3 is now the default behaviour and V2 deprecated implementations have been removed.

## Fixed Issue(s)

Fixes #7779 
## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -224,11 +221,9 @@ private SafeFuture<Void> setExecutionData(
     // if requestedBlinded has been specified, we strictly follow it, otherwise we should run
```
StefanBratanov: can remove this comment

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -287,8 +281,7 @@ private SafeFuture<Void> setPayloadOrPayloadHeader(
             builderBidOrFallbackData -> {
               // we should try to return unblinded content only if no explicit "blindness" has been
               // requested and builder flow fallbacks to local
-              if (requestedBlinded.isEmpty()
-                  && builderBidOrFallbackData.getFallbackData().isPresent()) {
+              if (builderBidOrFallbackData.getFallbackData().isPresent()) {
```
StefanBratanov, line 276: This whole if statement should be simplified. After this change, we have only one condition, if fallback data is present, we return unblinded executionPayload, otherwise we have the builderBid and return the blinded header.

Essentially, we can skip this, since it will never happen:

```java
final ExecutionPayload executionPayload =builderBidOrFallbackData
    .getFallbackDataRequired()
    .getExecutionPayload();
 return schemaDefinitions
     .getExecutionPayloadHeaderSchema()
     .createFromExecutionPayload(executionPayload);
```

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/OkHttpValidatorTypeDefClient.java
```diff
@@ -169,11 +150,8 @@ public Optional<BlockContainerAndMetaData> createUnsignedBlock(
     try {
       return produceBlockRequest.submit(randaoReveal, graffiti, requestedBuilderBoostFactor);
     } catch (final BlockProductionV3FailedException ex) {
```
StefanBratanov: I think we can get rid of this `BlockProductionV3FailedException`. Was used for fallback specifically.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
gfukushima: well that's interesting, I understand the logic behind this clean up but wouldn't that be somehow a weird enforcement we would do? Every CL must have a EL in order to be fully functional in the network as a proposer, no? If so the first condition will always be true and even if you configure a builderBid we will default to use the fallback always isn't that so?

StefanBratanov: So, the way it works is that either the builder bid is present or the fallback is present. The two won't be present at the same time.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -291,16 +288,7 @@ private SafeFuture<Void> setPayloadOrPayloadHeader(
                         // from the builder bid
```
StefanBratanov, line 283: We can create similar method to `getFallbackDataRequired()` like `getBuilderBidRequired()` to make it bit cleaner

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -218,21 +218,20 @@ private SafeFuture<Void> setExecutionData(
...
-                executionPayloadContext
-                    .map(ExecutionPayloadContext::isValidatorRegistrationPresent)
-                    .orElse(false);
+        executionPayloadContext
+            .map(ExecutionPayloadContext::isValidatorRegistrationPresent)
+            .orElse(false);
 
     // pre-Merge Execution Payload / Execution Payload Header
     if (executionPayloadContext.isEmpty()) {
-      if (shouldTryBuilderFlow) {
```
StefanBratanov, line 235: Hmm why did we remove the if statement?

gfukushima: if `executionPayloadContext` is empty then `shouldTryBuilderFlow` will always be false

StefanBratanov: Oh yes, got it. In this case, we can just leave the `.executionPayload`

StefanBratanov: And move shouldTryBuilderFlow initialization afterwards

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/OkHttpValidatorTypeDefClient.java
gfukushima: I think we could keep it just maintain the pattern that has been used for others catch handlers.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
gfukushima: leaving just the `.executionPayload` break one of the bellatrix unit tests (`shouldIncludeExecutionPayloadHeaderOfDefaultPayload`). which in my opinion should probably be fine to be removed but I'd like a second opinion here.

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/OkHttpValidatorTypeDefClient.java
gfukushima: removed the try and catch but left the BlockProductionV3FailedException class so follow the same behaviour other api have to report the runtime exception.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
StefanBratanov: Yeah, we can just remove that test, it is no longer applicable. We will always use the unblinded flow in that case.

StefanBratanov (approved):
LGTM with one comment

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
StefanBratanov: I think leaving only .executionPayload() will fix the acceptance test failure

StefanBratanov:
I think we should remove this in `ProduceBlockRequest`

```           
.withHandler(SC_NOT_FOUND,
                (__, ___) -> {
                  throw new BlockProductionV3FailedException();
                });
```

So that endpoint call from VC correctly returns 404. We don't have any fallback, so better to return correct response code to VC.

PR #8517, opened 2024-08-20
add post aggregate and proofs V2 API

### LINKED ISSUE #8439: [beacon api] [electra] Add publishAggregateAndProofsV2 to support EIP-7549
https://github.com/ethereum/beacon-APIs/pull/456

mehdi-aouadi:
## PR Description
Add PostAggregateAndProofsV2 API:
- Refactor the `SignedAgrgegateAndProofs` and `AggregateAndProofs` schemas to consider pre/post Electra
- Use `SchemaDefinitionCache` instead of `SchemaDefinition` in the old V1
- Deprecate the V1 API

## Fixed Issue(s)

fixes #8439 

## Documentation

## Changelog


data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+            JsonUtil.serialize(aggregateAndProofs, aggregateAndProofsListTypeDef),
+            Collections.emptyMap(),
+            Optional.of(specMilestone.name().toLowerCase(Locale.ROOT)));
+
+    assertThat(response.code()).isEqualTo(SC_OK);
+    assertThat(response.body().string()).isEmpty();
+  }
+
+  @TestTemplate
+  void shouldReturnErros() throws Exception {
```
rolfyone: typo

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+    dataStructureUtil = specContext.getDataStructureUtil();
+    aggregateAndProofsListTypeDef =
+        SerializableTypeDefinition.listOf(
+            spec.getGenesisSchemaDefinitions()
+                .getSignedAggregateAndProofSchema()
+                .getJsonTypeDefinition());
+  }
+
+  @TestTemplate
+  void shouldPostAggregateAndProofs_NoErrors() throws Exception {
```
rolfyone: generally i think we try to do no capital directly after underscore? we should probably have a rule for it really...

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+    final Response response =
+        post(
+            PostAggregateAndProofsV2.ROUTE,
+            JsonUtil.serialize(aggregateAndProofs, aggregateAndProofsListTypeDef));
+
+    assertThat(response.code()).isEqualTo(SC_BAD_REQUEST);
+
+    final JsonNode resultAsJsonNode = JsonTestUtil.parseAsJsonNode(response.body().string());
+    assertThat(resultAsJsonNode.get("message").asText())
+        .isEqualTo("(Eth-Consensus-Version) header value was unexpected");
```
rolfyone: if this header is missing and we require it, i'd expect a more clear message, this is very ambiguous... i'd expect a message along the lines of 'missing required header value for Eth-Consensus-Version'...

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+            PostAggregateAndProofsV2.ROUTE,
+            JsonUtil.serialize(aggregateAndProofs, aggregateAndProofsListTypeDef),
+            Collections.emptyMap(),
+            Optional.of("NonExistingMileStone"));
+
+    assertThat(response.code()).isEqualTo(SC_BAD_REQUEST);
+
+    final JsonNode resultAsJsonNode = JsonTestUtil.parseAsJsonNode(response.body().string());
+    assertThat(resultAsJsonNode.get("message").asText())
+        .isEqualTo("(Eth-Consensus-Version) header value was unexpected");
```
rolfyone: similar here - 'unexpected' is not great. can we have a message like 'the value supplied for Eth-Consensus-Version was invalid: NoExistingMilestone' - then we'd have context...

For me it comes back to a person on discord raising a question with 'Header value was unexpected' - can i help that person?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/SignedAggregateAndProof.java
```diff
@@ -27,9 +27,10 @@ public class SignedAggregateAndProof
   public static class SignedAggregateAndProofSchema
       extends ContainerSchema2<SignedAggregateAndProof, AggregateAndProof, SszSignature> {
 
-    public SignedAggregateAndProofSchema(final AggregateAndProofSchema aggregateAndProofSchema) {
+    public SignedAggregateAndProofSchema(
+        final AggregateAndProofSchema aggregateAndProofSchema, final boolean isElectra) {
       super(
-          "SignedAggregateAndProof",
+          isElectra ? "SignedAggregateAndProofElectra" : "SignedAggregateAndProofPhase0",
```
rolfyone: i do wonder if we can get this information from the schema passed in rather than having a boolean flag which won't really scale out if we have to end up changing this in a future fork...

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
mehdi-aouadi: My bad, fixed

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
mehdi-aouadi: Removed the _ anyway

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
mehdi-aouadi: Made it clearer when missing: `Missing required header value for (Eth-Consensus-Version)`

data/beaconrestapi/src/integration-test/java/tech/pegasys/teku/beaconrestapi/v2/validator/PostAggregateAndProofsV2IntegrationTest.java
mehdi-aouadi: Made it clearer when we're having an unexpected value: `Invalid value for (Eth-Consensus-Version) header: NoExistingMilestone `

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/operations/SignedAggregateAndProof.java
mehdi-aouadi: Well, removed that boolean and used the attestation schema to determine that instead, it looks cleaner but yet all this schema versioning will need a refactoring in the future...

rolfyone: yeh but by doing something like your new version, we no longer need 'isElectra', 'isF*', 'isG*' etc etc, so i think its closer to what we want...

mehdi-aouadi: I guess you're talking about the other usages of the `isElectra` "hack", the problem is that in those other cases, we don't have access to the fork info from the available schemas. Take for example `IndexedAttestationSchema`, there's no way to know at which fork we're at without passing an additional info in the constructor. It played out well for the `AggregateAndProofSchema` and the `SignedAggregateAndProofSchema` because both have access to the attestation schema which can be used to determine if we reached Electra. And even that is not very clean IMO.
Some thoughts here:
- The really proper way to solve this is to extends the Electra affected schemas the same way we did for the `AttestationPhase0` and `AttestationElectra` but that means a huge refactoring and clearly must be done in a separate PR. We've already discussed this solution but thought we could do it later because the refactoring will impact the beacon block body schema...
- The idea is definitely not having `isF`, `isG`... that definitely doesn't scale and is almost an anti pattern. It's just a "hacky" intermediate solution

rolfyone (approved):

PR #8522, opened 2024-08-21
use header based selector with slot fallback

### LINKED ISSUE #7720: Support eth-consensus-version header as parameter
We could support eth-consensus-version on API  consuming posted objects.

We already have a ssz deserializer considering the header or falling back to slot from ssz. We could have a similar selector for json that use the header or fallback to slot.

mehdi-aouadi:
## PR Description
Use the `Eth-Consensus-Version` header when available to determine the request body type definition and fallback to the slot selector when the header is not required and not present:
- Some APIs have the `Eth-Consensus-Version` required: In that case we don't fallback to the slot selector since the request should fail when the consensus version header isn't provided or a bad value is used, we use the `headerBasedSelector` without fallback to the slot. This is the case for the V2 APIs: [/eth/v2/beacon/pool/attestations](https://github.com/ethereum/beacon-APIs/blob/5f7dd7ba88de99e0b3fa718aef6fdf8ce61ad5ea/apis/beacon/pool/attestations.v2.yaml#L67),  [/eth/v2/beacon/pool/attester_slashings](https://github.com/ethereum/beacon-APIs/blob/5f7dd7ba88de99e0b3fa718aef6fdf8ce61ad5ea/apis/beacon/pool/attester_slashings.v2.yaml#L42), [/eth/v2/beacon/blinded_blocks](https://github.com/ethereum/beacon-APIs/blob/5f7dd7ba88de99e0b3fa718aef6fdf8ce61ad5ea/apis/beacon/blocks/blinded_blocks.v2.yaml#L42) and [/eth/v2/beacon/blocks](https://github.com/ethereum/beacon-APIs/blob/5f7dd7ba88de99e0b3fa718aef6fdf8ce61ad5ea/apis/beacon/blocks/blocks.v2.yaml#L41)
- The other APIs have the `Eth-Consensus-Version` optional: In taht case, we try to get the type definition using the header otherwise we fall back to the slot based selector if the header is not present, we use the `headerBasedSelectorWithSlotFallback`. This is the case for the V1 APIs: [/eth/v1/beacon/blinded_blocks](https://github.com/ethereum/beacon-APIs/blob/5f7dd7ba88de99e0b3fa718aef6fdf8ce61ad5ea/apis/beacon/blocks/blinded_blocks.yaml#L20) and [/eth/v1/beacon/blocks](https://github.com/ethereum/beacon-APIs/blob/5f7dd7ba88de99e0b3fa718aef6fdf8ce61ad5ea/apis/beacon/blocks/blocks.yaml#L19)

PS: This only covers the json content type use case. For the ssz use case, we always fall back to the slot selector even when the header is required but not provided which is not ideal IMO. I thought we should maybe be more strict about that and make the request fail when the header is required but not provided for the ssz content type

## Fixed Issue(s)

fixes #7720 

## Documentation

## Changelog

### DIFF
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
@@ -118,7 +118,9 @@ public class MilestoneDependentTypesUtil {
       final Function<SchemaDefinitions, SszSchema<? extends T>> getSchema) {
     final Optional<UInt64> slot =
         // SignedBeaconBlock
-        getSlot(json, "message", "slot");
+        getSlot(json, "message", "slot")
+            // SignedBlockContents
+            .or(() -> getSlot(json, "signed_block", "message", "slot"));
     final SpecMilestone milestone =
         schemaDefinitionCache.milestoneAtSlot(
             slot.orElseThrow(() -> new BadRequestException("Could not locate slot in JSON data")));
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
@@ -79,21 +79,16 @@ public class MilestoneDependentTypesUtil {
     return builder.build();
   }
 
-  public static <T extends SszData> DeserializableTypeDefinition<? extends T> slotBasedSelector(
-      final String json,
-      final SchemaDefinitionCache schemaDefinitionCache,
-      final Function<SchemaDefinitions, SszSchema<? extends T>> getSchema) {
-    final Optional<UInt64> slot =
-        // SignedBeaconBlock
-        getSlot(json, "message", "slot")
-            // SignedBlockContents
-            .or(() -> getSlot(json, "signed_block", "message", "slot"));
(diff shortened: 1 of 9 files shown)


data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
```diff
@@ -79,38 +79,56 @@ SerializableOneOfTypeDefinition<T> getMultipleSchemaDefinitionFromMilestone(
...
 
+  private static <T extends SszData> DeserializableTypeDefinition<? extends T> slotBasedSelector(
+      final String json,
+      final SchemaDefinitionCache schemaDefinitionCache,
+      final Function<SchemaDefinitions, SszSchema<? extends T>> getSchema) {
+    final Optional<UInt64> slot =
+        // SignedBeaconBlock
+        getSlot(json, "message", "slot")
+            // SignedBlockContents
+            .or(() -> getSlot(json, "signed_block", "message", "slot"));
```
rolfyone, line 123: is block contents still a thing? I thought that went away?

rolfyone (approved):
LGTM just a question about block contents container

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/MilestoneDependentTypesUtil.java
mehdi-aouadi: Yep, we added them since Deneb to handle block and blobs. thought we could remove this use case but turned out necessary when we send Deneb block contents to the post block V1 apis, they might retrieve the slot in order to get the correct schema

PR #8543, opened 2024-08-29
Add native support for ephemery testnet and config files

### LINKED ISSUE #8537: Add Native Support for Ephemery Testnet and Config Files
### Description
There is currently no native support for [Ephemery](https://ephemery.dev/) using the --network=ephemery flag.
We also need to include the Ephemery config to connect to the Ephemery Testnet easily.

For context Ephemery Testnet was initiated by @taxmeifyoucan. Ephemery testnet is a network that rolls back to the genesis after a set period of time. It is focused on short term and heavy testing usecases. This is also to avoid problems like insufficient testnet funds, inactive validators, state bloat, and similar issues faced by long-running testnets.

We want to be able to connect to Ephemery like other supported networks like this
```bash
teku \
    --network=ephemery                             \
    --ee-endpoint=http://localhost:8551          \
    --ee-jwt-secret-file=<path to jwtsecret.hex> \
    --metrics-enabled=true                       \
    --rest-api-enabled=true                      \
    --checkpoint-sync-url=<checkpoint sync URL>
```

Currently to connect to Ephmery network it has to be done manually.

```bash
teku \
--genesis-state=ephemery/besu.json  \
--ee-endpoint=http://localhost:8551/  \
--ee-jwt-secret-file=<path to jwtsecret.hex> \
--metrics-enabled=true \
--rest-api-enabled=true  \
--p2p-discovery-bootnodes=$BOOTNODE_ENR_LIST   \
--checkpoint-sync-url=<checkpoint sync URL>

```
Whenever, we need to start an Ephemery Testnet

As part of my contribution for my Ethereum Protocol Fellowship (EPF) project, I would like this issue to be assigned to me.

gconnect:
## PR Description
- Add Ephemery Testnet config files

## Fixed Issue(s)

fixes #8537

## Changelog


gconnect:
@rolfyone please help review when you have the time. Thanks.

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/ephemery.yaml
gconnect, line 1: @taxmeifyoucan and @pk910 please help review the content of the ephemery.yaml if the values are correct. And possibly point me for where I can get the other values in $. Thank you.

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -905,6 +907,21 @@ private Builder applyHoleskyNetworkDefaults() {
...
     }
 
+    private Builder applyEphemeryNetworkDefaults() {
+      return applyTestnetDefaults()
+              .constants(EPHEMERY.configName())
+              .startupTimeoutSeconds(120)
+              .trustedSetupFromClasspath(MAINNET_TRUSTED_SETUP_FILENAME)
+              .eth1DepositContractDeployBlock(0)
+              .defaultInitialStateFromUrl(
+                      "https://checkpoint-sync.ephemery.ethpandaops.io/eth/v2/debug/beacon/states/finalized")
```
gconnect: @taxmeifyoucan please help me confirm if this URL is correct

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotsBundleTest.java
```diff
@@ -152,6 +152,7 @@ public static Stream<Arguments> getSupportedNetworks() {
         Arguments.of(Eth2Network.GNOSIS, "https://rpc.ankr.com/gnosis"),
         Arguments.of(Eth2Network.MAINNET, "https://mainnet.infura.io/v3/%INFURA_KEY%"),
         Arguments.of(Eth2Network.SEPOLIA, "https://sepolia.infura.io/v3/%INFURA_KEY%"),
-        Arguments.of(Eth2Network.HOLESKY, "https://holesky.infura.io/v3/%INFURA_KEY%"));
+        Arguments.of(Eth2Network.HOLESKY, "https://holesky.infura.io/v3/%INFURA_KEY%"),
+        Arguments.of(Eth2Network.EPHEMERY, "https://otter.bordel.wtf/erigon"));
```
gconnect, line 156: @taxmeifyoucan this is the rpc url I was talking about. Although it's for testing purpose. I just added this https://otter.bordel.wtf/erigon let me know if there is any one I can replace it with

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -905,6 +907,21 @@ private Builder applyHoleskyNetworkDefaults() {
...
+          .startupTimeoutSeconds(120)
+          .trustedSetupFromClasspath(MAINNET_TRUSTED_SETUP_FILENAME)
+          .eth1DepositContractDeployBlock(0)
+          .defaultInitialStateFromUrl(
+              "https://checkpoint-sync.ephemery.ethpandaops.io/eth/v2/debug/beacon/states/finalized")
+          .customGenesisState(
+              "https://github.com/ephemery-testnet/ephemery-genesis/releases/latest/download/genesis.ssz")
+          .discoveryBootnodes(
+              // TEKU bootnode
+              "enr:-Iq4QNMYHuJGbnXyBj6FPS2UkOQ-hnxT-mIdNMMr7evR9UYtLemaluorL6J10RoUG1V4iTPTEbl3huijSNs5_ssBWFiGAYhBNHOzgmlkgnY0gmlwhIlKy_CJc2VjcDI1NmsxoQNULnJBzD8Sakd9EufSXhM4rQTIkhKBBTmWVJUtLCp8KoN1ZHCCIyk");
```
rolfyone: we won't have a teku bootnode i don't expect, but this would be where we'd list any ephemery network discovery nodes that we can use during startup

gconnect: > we won't have a teku bootnode i don't expect, but this would be where we'd list any ephemery network discovery nodes that we can use during startup

Yes. This is the one Ephemery is using on Teku. I will remove the comment to avoid the confusion. That was how it was labelled on the Ephemery repo

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/ephemery.yaml
pk910: Heya @gconnect
I've taken a first look into your commit :)

The `genesis.ssz` you've included in the commit is valid for the current iteration 136 only. 
It's already invalid in a few hours as iteration 137 starts today at 19:00 UTC ;)
You'll need to load the latest `genesis.ssz` via one of these ways:
1. From the [latest release](https://github.com/ephemery-testnet/ephemery-genesis/releases) in the genesis repository
2. Or from https://ephemery.dev/latest/metadata/genesis.ssz  ([ephemery.dev/latest/*](https://ephemery.dev/latest/) is a mirror for the latest release)

Also, you gonna need to replace all the variables within the `config.yaml` with actual values.
These are mostly static,  but there's also 3 that need to be set dynamically for each iteration: `MIN_GENESIS_TIME`, `DEPOSIT_CHAIN_ID`, `DEPOSIT_NETWORK_ID`

```
# ephemery configuration
FIRST_GENESIS = 1393527600
FIRST_CHAINID = 39438000
GENESIS_INTERVAL = 2419200   # 4 weeks

# calculate iteration number based on current time
ITERATION_NUMBER = (now() - FIRST_GENESIS) // GENESIS_INTERVAL

# config values
MIN_GENESIS_TIME = FIRST_GENESIS + (GENESIS_INTERVAL * ITERATION_NUMBER)
DEPOSIT_CHAIN_ID = FIRST_CHAINID + ITERATION_NUMBER
DEPOSIT_NETWORK_ID = FIRST_CHAINID + ITERATION_NUMBER
```

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
pk910: That should work 👍

services/powchain/src/test/java/tech/pegasys/teku/services/powchain/DepositSnapshotsBundleTest.java
taxmeifyoucan: I see, this works! I am maintaining the endpoint and it can stay as it is

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/ephemery.yaml
taxmeifyoucan: @gconnect Since some of the genesis values are dynamic and updated with each iteration, yaml and ssz need to updated. We recently discussed it, there are multiple ways to obtain latest `genesis.ssz`. 

Here is some discussion from lodestar implementation on how it can be handled https://github.com/ChainSafe/lodestar/issues/6064
And notes from previous fellow implementing it in lighthouse: https://hackmd.io/@teri-b/S1D6Np_Q6

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
rolfyone: oh ok - that's fine... just for clarity we generally put 'teku bootnode' when its run by us specifically so that we can track it, so maybe just remove the comment in that case, and we'll look at it as a black box...

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/ephemery.yaml
gconnect: > Heya @gconnect I've taken a first look into your commit :)
> 
> The `genesis.ssz` you've included in the commit is valid for the current iteration 136 only. It's already invalid in a few hours as iteration 137 starts today at 19:00 UTC ;) You'll need to load the latest `genesis.ssz` via one of these ways:
> 
> 1. From the [latest release](https://github.com/ephemery-testnet/ephemery-genesis/releases) in the genesis repository
> 2. Or from https://ephemery.dev/latest/metadata/genesis.ssz  ([ephemery.dev/latest/*](https://ephemery.dev/latest/) is a mirror for the latest release)
> 
> Also, you gonna need to replace all the variables within the `config.yaml` with actual values. These are mostly static, but there's also 3 that need to be set dynamically for each iteration: `MIN_GENESIS_TIME`, `DEPOSIT_CHAIN_ID`, `DEPOSIT_NETWORK_ID`
> 
> ```
> # ephemery configuration
> FIRST_GENESIS = 1393527600
> FIRST_CHAINID = 39438000
> GENESIS_INTERVAL = 2419200   # 4 weeks
> 
> # calculate iteration number based on current time
> ITERATION_NUMBER = (now() - FIRST_GENESIS) // GENESIS_INTERVAL
> 
> # config values
> MIN_GENESIS_TIME = FIRST_GENESIS + (GENESIS_INTERVAL * ITERATION_NUMBER)
> DEPOSIT_CHAIN_ID = FIRST_CHAINID + ITERATION_NUMBER
> DEPOSIT_NETWORK_ID = FIRST_CHAINID + ITERATION_NUMBER
> ```

@pk910 and @taxmeifyoucan  Thanks for this!

I had an extensive meeting with @rolfyone and this is the route we will be taking for implementation on Teku.   
I have removed the ssz and added a check in the config to pull the update from checkpointSync with the url. 

For the yaml file I have write a custom Ephemery class to handle the update of the yaml file for those values that needs to change.

teku/src/main/java/tech/pegasys/teku/cli/BeaconNodeCommand.java
```diff
@@ -401,6 +401,11 @@ public StartAction getStartAction() {
   protected TekuConfiguration tekuConfiguration() {
     try {
       TekuConfiguration.Builder builder = TekuConfiguration.builder();
+
+      //      String netwxafork =  eth2NetworkOptions.getNetwork();
+      //      if(network.equals(Eth2Network.EPHEMERY)){
+      //        EphemeryNetwork.updateConfig();
+      //      }
```
tbenr: Can we clean this up if not needed?

gconnect: Removed thanks for checking.

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -910,6 +908,21 @@ private Builder applyHoleskyNetworkDefaults() {
               "enr:-LS4QG0uV4qvcpJ-HFDJRGBmnlD3TJo7yc4jwK8iP7iKaTlfQ5kZvIDspLMJhk7j9KapuL9yyHaZmwTEZqr10k9XumyCEcmHYXR0bmV0c4gAAAAABgAAAIRldGgykGm32XQEAXAAAAEAAAAAAACCaWSCdjSCaXCErK4j-YlzZWNwMjU2azGhAgfWRBEJlb7gAhXIB5ePmjj2b8io0UpEenq1Kl9cxStJg3RjcIIjKIN1ZHCCIyg");
     }
 
+    private Builder applyEphemeryNetworkDefaults() {
+      return applyTestnetDefaults()
+          .constants(EPHEMERY.configName())
+          .startupTimeoutSeconds(120)
+          .trustedSetupFromClasspath(MAINNET_TRUSTED_SETUP_FILENAME)
```
rolfyone, line 923: i think we can remove the trusted setup file here now...

ethereum/networks/src/main/java/tech/pegasys/teku/networks/EphemeryNetwork.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class EphemeryNetwork {
+  private static final long GENESIS_CHAINID = 39438135;
+  private static final long GENESIS_TIMESTAMP = 1720119600;
+  private static final int PERIOD = 28;
+  private static final long PERIOD_IN_SECONDS = (PERIOD * 24 * 60 * 60);
+  public static final long PERIODS_SINCE_GENESIS =
+      ChronoUnit.DAYS.between(Instant.ofEpochSecond(GENESIS_TIMESTAMP), Instant.now()) / PERIOD;
+
+  public static void updateConfig(final SpecConfigBuilder builder) {
```
rolfyone, line 39: One thing that may be handy is if we expand this a tiny bit, and pass in TimeProvider... that would allow us to write better tests because we could show it handling specific inputs and wind forward periods of time...

If we can adjust the time we can poke at the exact second of change etc which allows for some interesting testing of boundaries if thats helpful... Basically in tests you can use StubTimeProvider and set the time to whatever you like, and generally speaking use SystemTimeProvider in the rest of the system...

At its simplest, you'd potentially have an updateConfig that takes a timeProvider, and have a default interface like this that calls through with a SystemTimeProvider
so you'd have 
```
public static void updateConfig(final SpecConfigBuilder builder) {
  return updateConfig(builder, new SystemTimeProvider());
}
static void updateConfig(final SpecConfigBuilder builder, final TimeProvider timeProvider);
```
then potentially we'd make `periodsSinceGenesis` a package private method that can use a timeProvider to compute the result (its only used in your test so package private should be fine)

In this way we could even write tests just for that periodsSinceGenesis fn and playing with the StubTimeProvider...

ethereum/networks/src/test/java/tech/pegasys/teku/networks/EphemeryNetworkTest.java
```diff
@@ -0,0 +1,150 @@
...
+  private long expectedChainId;
+  private final SpecConfigReader reader = new SpecConfigReader();
+
+  @BeforeEach
+  void setUp() {
+    long periodSinceGenesis = EphemeryNetwork.getPeriodsSinceGenesis(timeProvider);
+    expectedMinGenesisTime = GENESIS_TIMESTAMP + (periodSinceGenesis * PERIOD_IN_SECONDS);
+    expectedChainId = GENESIS_CHAINID + periodSinceGenesis;
+    builder = mock(SpecConfigBuilder.class);
+  }
```
rolfyone: so yep this is one use... (add final to periods as nit)

The reason we do this though, is because then i can test the first period since genesis, or even pre the first genesis time, and that's the real power of making the timeProvider change, because then we can test with predictableOutput rather than using SystemTimeProvider...

So for example if you wanted 1000 periods since genesis, you could compute a time that gives you that, and then using a StubTimeProvider set that value, and then you're always running at that point in time...

Or before the first genesis...

As a first step having time provider there is adequate, and we can expand these tests separately, i might raise it as a good first issue for someone maybe?

gconnect: Yes, you are right. I have added final and more test cases.

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -399,6 +400,7 @@ public Eth2NetworkConfiguration build() {
             SpecFactory.create(
                 constants,
                 builder -> {
+                  EphemeryNetwork.updateConfig(builder);
```
rolfyone: this will be applying these defaults to all the networks, so that'll explain why the AT is breaking... Just as an observation, I haven't proven it, but this will apply to all networks here.

gconnect: Making update now, thanks

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
```diff
@@ -423,11 +428,6 @@ public Eth2NetworkConfiguration build() {
                         if (maybeEpochsStoreBlobs.isPresent()) {
                           denebBuilder.epochsStoreBlobs(maybeEpochsStoreBlobs);
                         }
-                        if (trustedSetup.isEmpty()) {
-                          LOG.warn(
-                              "Setting a default for trusted setup as nothing was set explicitly");
-                          trustedSetupFromClasspath(MAINNET_TRUSTED_SETUP_FILENAME);
-                        }
```
rolfyone: I think we shouldn't be altering this part in our change. can we restore this please?

ethereum/networks/src/test/java/tech/pegasys/teku/networks/EphemeryNetworkTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+  @Test
+  void shouldNotUpdateConfigBeforeGenesis() {
+    final Spec spec =
+        getSpec(phase0Builder -> phase0Builder.minGenesisTime(UInt64.valueOf(MIN_GENESIS_TIME)));
+
+    final long preGenesisTime = MIN_GENESIS_TIME - ONE_PERIOD;
+    final StubTimeProvider stubTimeProvider =
+        StubTimeProvider.withTimeInMillis(preGenesisTime * 1000);
```
rolfyone: nit: could use `withTimeInSeconds` if you don't want to do the multiplication everywhere...

ethereum/networks/src/main/java/tech/pegasys/teku/networks/Eth2NetworkConfiguration.java
gconnect: I am not sure what is been altered here.

gconnect: I have added it

ethereum/networks/src/test/java/tech/pegasys/teku/networks/EphemeryNetworkTest.java
gconnect: Noted thanks.

rolfyone (approved):
LGTM - thanks for the persistence!

PR #8546, opened 2024-08-29
add --Xattestations-v2-apis-enabled option

mehdi-aouadi:
## PR Description
Add a hidden `--Xattestations-v2-apis-enabled` option to enable attestations v2 APIs even before Electra:
- `/eth/v2/beacon/pool/attestations`
- `/eth/v2/validator/aggregate_and_proofs`
- `/eth/v2/validator/aggregate_attestation`

## Documentation

## Changelog


tbenr:
maybe we could switch this default true when we will be close to Electra release, so that we can be sure that most clients have implemented this, while maintaining the ability to switch it off just in case.

tbenr (approved):
LGTM

PR #8548, opened 2024-08-29
handle missing committee index

mehdi-aouadi:
## PR Description
Handle the use case when using the `eth/v2/validator/aggregate_attestation` API and `committee_index` parameter is missing. 

## Documentation

## Changelog


tbenr (approved):
LGTM. after a spotless run :)

PR #8553, opened 2024-08-30
attestation duties generation optimization

tbenr:
(all the noise outside `AttesterDutiesGenerator` is due to refactoring `CommitteeAssignment` into a `record`)

Close to 3 order of magnitude better. Too good to be true?

A 2x boost was removing Integer boxing. But then I tried `ArrayMap` instead of `OpenHashMap` and boom.

old:
```
Benchmark                                               (querySize)  (validatorsCount)   Mode  Cnt     Score   Error  Units
AttesterDutiesGeneraterBenchmark.computeAttesterDuties       500000             800000  thrpt   10  1225.670 ± 6.127  ops/s
```

new:
```
Benchmark                                               (querySize)  (validatorsCount)   Mode  Cnt       Score      Error  Units
AttesterDutiesGeneraterBenchmark.computeAttesterDuties       500000             800000  thrpt   10  974975.378 ± 7047.143  ops/s
```

newer (avoid second pass on duties):
```
Benchmark                                               (querySize)  (validatorsCount)   Mode  Cnt        Score      Error  Units
AttesterDutiesGeneraterBenchmark.computeAttesterDuties       500000             800000  thrpt   10  1072974.753 ± 5894.667  ops/s
```


rolfyone (approved):
LGTM

PR #8558, opened 2024-09-03
Added deneb remote signer AT

rolfyone:
Had to leave this test disabled, because it needs a change not currently in web3signer develop. Once it's in develop, we can start adding it in.

Also changed so that trusted-setup always defaults to mainnet if it's not set, which will remove the need in the first place for the web3signer change, but to be able to test this is the current easiest path i could see.

## Documentation

## Changelog


acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/DenebRemoteSignerAcceptanceTest.java
```diff
@@ -0,0 +1,134 @@
...
+  private static final String NETWORK_NAME = "swift";
+  public static final Eth1Address WITHDRAWAL_ADDRESS =
+      Eth1Address.fromHexString("0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee");
+  private static final URL JWT_FILE = Resources.getResource("auth/ee-jwt-secret.hex");
+
+  @Test
+  void denebWithRemoteSigner() throws Exception {
+    final UInt64 currentTime = new SystemTimeProvider().getTimeInSeconds();
+    final int genesisTime =
+        currentTime.intValue() + 30; // genesis in 10 seconds to give node time to start
```
lucassaldanha, line 45: NIT: The comment needs to be updated to 30 otherwise it will be lying :)

lucassaldanha (approved):

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/DenebRemoteSignerAcceptanceTest.java
rolfyone: oops yes i missed this comment, will fix

PR #8566, opened 2024-09-05
Completed removal of JsonProvider from IntegrationTests

### LINKED ISSUE #8526: Replace uses of JsonProvider in `AbstractDataBackedRestApiIntegrationTest`
It's possible that ObjectMapper is adequate, but we'd have to investigate.

Basically the intent is not to reference objects in `schema/*`
So if we're reading specific values from JSON, an option is to just use ObjectMapper. If we need complex objects this is more complicated, because the JsonProvider has custom deserializers, and we may need to consider using TypeDefinitions, but we need to be really careful about using `TypeDefinitions` on both implementation and testing sides, as it'll hide problems...

At the end of this task, The expectation is that `AbstractDataBackedRestApiIntegrationTest` and everything extending it no longer uses `JsonProvider`.

rolfyone:
AbstractDataBackedRestAPIIntegrationTest no longer requires JsonProvider.

fixes #8526

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #8573, opened 2024-09-05
From bytes and from hex string

### LINKED ISSUE #8506: create and test `fromHexString` for `SszBitVector`
As a developer using the `SszBitVector` class, 
it would be convenient to have `fromHexString` and `fromBytes` defined,
So that I can create an `SszBitVector` from a hex or bytes payload.

Basically, `fromBytes` would do the heavy lifting, `fromHexString` would convert to bytes then call `fromBytes`.

I'd suggest following how `ofBits` is done basically.

The implementation will likely be in `SszBitVectorImpl` where `ofBits` is defined currently. and would be on the `SszBitVectorSchema`

Testing will be in SszBitVectorTest.

gconnect:
## PR Description
- [x] Create and test fromBytes
- [x] Create and test fromHexString

## Fixed Issue(s)
fixes #8506


rolfyone (approved):
LGTM, thanks for your patience on this one!

PR #8574, opened 2024-09-05
Removed error message that we no longer need

### LINKED ISSUE #8526: Replace uses of JsonProvider in `AbstractDataBackedRestApiIntegrationTest`

### LINKED ISSUE #8562: Erroneous warning in logs
### Description

teku is generating warnings when blocks are requested via the v3 endpoint that appear to be erroneous.

### Steps to Reproduce (Bug)

Request a block via the v3 endpoint https://ethereum.github.io/beacon-APIs/#/Validator/produceBlockV3

Check the logs, e.g.:

```
2024-09-04 11:45:13.001 INFO  - Creating unsigned block for slot 2462326
2024-09-04 11:45:13.850 WARN  - Builder endpoint is available but a non-blinded block has been requested
```

It appears that this warning will always be generated for the v3 endpoint as by the specification of this call teku will always fetch both blinded and non-blinded payloads to compare values.  The warning should not be generated for a call to the v3 endpoint.

```
teku --version
teku/v24.8.0/linux-x86_64/-ubuntu-openjdk64bitservervm-java-21
```

rolfyone:
This used to be needed for v2 but now that we're completely using v3 there's no real need for the message to exist...

fixes #8562 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,3 +14,4 @@
 - Implemented [PostAggregateAndProofsV2](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/publishAggregateAndProofsV2) (adding support for Electra)
 
 ### Bug Fixes
+ - removed a warning from logs about non blinded blocks being requested (#8562)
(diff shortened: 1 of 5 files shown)


lucassaldanha (approved):
LGTM

gfukushima (approved):

PR #8577, opened 2024-09-06
Alter host allow logic in rest api

### LINKED ISSUE #8567: Unable to Access Teku REST API from Local Network Client
### Description
I am attempting to use the Teku REST API from a machine within my local network but different from the one running the Teku client. By default, the REST API access is restricted, so I have been trying to configure the Teku settings to allow access from within the local network.

### Steps to Reproduce (Bug)
Setup details:

- --rest-api-enabled=true

- ufw status for port 5051 is allowed on the Teku machine

- I can successfully ping from the client machine to the Teku machine and back

- --rest-api-host-allowlist=* (This works, but it exposes the API globally, which I want to avoid)

**Attempted configuration (preferred solution):**
```bash
teku --rest-api-enabled=true \
     --rest-api-interface=0.0.0.0 \
     --rest-api-port=5051 \
     --rest-api-host-allowlist=192.168.1.*,localhost,127.0.0.1
```
**Additional attempts (all unsuccessful):**
- `--rest-api-host-allowlist=192.168.1.0/24,localhost,127.0.0.1` 
- `--rest-api-host-allowlist=192.168.1.111,localhost,127.0.0.1` (allowing only the specific client machine)
-  `--rest-api-host-allowlist=my_host,localhost,127.0.0.1`  (where `my_host` is the result of `hostname` on the client machine)

**Expected behavior:**
Get a result when running the following (or similar) on the client machine:
```bash
curl http://192.168.1.112:5051/eth/v1/beacon/states/head/validators?status=pending_queued
```

**Actual behavior:**
These configuration run without any errors, but from the client machine, I receive the following error when trying to access the API:
```json
{
    "title": "Host not authorized",
    "status": 403,
    "type": "https://javalin.io/documentation#forbiddenresponse",
    "details": {}
}
```

### Versions
* Software version: teku `24.8.0`
* Java version:
```
openjdk 21.0.4 2024-07-16
OpenJDK Runtime Environment (build 21.0.4+7-Ubuntu-1ubuntu224.04)
OpenJDK 64-Bit Server VM (build 21.0.4+7-Ubuntu-1ubuntu224.04, mixed mode, sharing)
```
* OS Name & Version: `Ubuntu 24.04 LTS`

rolfyone:
it appears that at least in currently documented behaviour that we were filtering on server address not client address, which made it hard to reason about.

fixes #8567

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,3 +14,4 @@
 - Implemented [PostAggregateAndProofsV2](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/publishAggregateAndProofsV2) (adding support for Electra)
 
 ### Bug Fixes
+ - Updated allow-hosts logic to fix an issue with filtering (#8567).
--- a/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/HostAllowlistHandler.java
+++ b/infrastructure/restapi/src/main/java/tech/pegasys/teku/infrastructure/restapi/HostAllowlistHandler.java
@@ -33,7 +33,7 @@ public class HostAllowlistHandler implements Handler {
 
   @Override
   public void handle(final Context ctx) throws Exception {
-    String header = ctx.host();
+    String header = ctx.ip();
     if (!isHostAuthorized(hostAllowlist, header)) {
       LOG.debug("Host not authorized " + header);
       throw new ForbiddenResponse("Host not authorized");


benjaminion:
See my [comment](https://github.com/Consensys/teku/issues/8567#issuecomment-2333592652). Please don't break this 🙃

rolfyone:
closing because I was incorrect about the intent - thanks @benjaminion !

PR #8591, opened 2024-09-12
use enumMap where possible on SpecMilestone as key

tbenr:
We should use `EnumMap` (and `EnumSet`) when we deal with enums as key, since it is generally faster than regular hashmaps.
Moreover they preserve ordering so we don't need to use `TreeMap` when we require it.

I left the `ConcurrentHashMap` version unchanged.


StefanBratanov (approved):
Nice, LGTM

PR #8600, opened 2024-09-16
Moving requests out of ExecutionPayload

### LINKED ISSUE #8593: Move requests out of execution_payload into beacon_block.body
Update our code according to: https://github.com/ethereum/consensus-specs/pull/3875

lucassaldanha:
## PR Description
- Removes `withdrawal_requests`, `deposit_requests` and `consolidation_requests` from Execution Payload (execution payload is exactly like Deneb, removed all Electra classes for it);
-  Added `execution_requests` into Beacon Block Body (Electra)
- Kept ExecutionPayloadV4 classes around as we will soon add some Electra-specific logic into them.
- Updating reference tests to v1.5.0-alpha.6
- Temporarily disabling slashing tests (will be re-enabled as part of https://github.com/Consensys/teku/pull/8612)

Also added some TODOs that are going to be addressed on:
- https://github.com/Consensys/teku/issues/8624
- https://github.com/Consensys/teku/issues/8620

**Reference spec PR:** https://github.com/ethereum/consensus-specs/pull/3875/files

## Fixed Issue(s)
fixes #8593 

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/schema/electra/BeaconBlockBodyElectra.java
```diff
@@ -30,17 +30,20 @@
...
 import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.electra.BeaconBlockBodySchemaElectra;
+import tech.pegasys.teku.spec.datastructures.execution.versions.electra.ExecutionRequestsBuilderElectra;
 import tech.pegasys.teku.spec.datastructures.type.SszKZGCommitment;
+import tech.pegasys.teku.spec.schemas.SchemaDefinitionsElectra;
 
 public class BeaconBlockBodyElectra extends BeaconBlockBodyAltair {
 
   @JsonProperty("execution_payload")
-  public final ExecutionPayloadElectra executionPayload;
+  public final ExecutionPayloadDeneb executionPayload;
```
gfukushima, line 45: is it worth a comment here explaining that we use `ExecutionPayloadDeneb` because no changes have been introduced in Electra? It's is clear now but it might be confusing in the future.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/electra/BlindedBeaconBlockBodyElectraImpl.java
```diff
@@ -88,7 +91,8 @@ class BlindedBeaconBlockBodyElectraImpl
         syncAggregate,
         executionPayloadHeader,
         blsToExecutionChanges,
-        blobKzgCommitments);
+        blobKzgCommitments,
+        executionRequests);
```
gfukushima, line 95: is this missing its correspondent get method?

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/common/operations/DefaultOperationProcessor.java
```diff
@@ -145,43 +144,26 @@ public void processWithdrawals(
...
       throws BlockProcessingException {
-    final SszList<ConsolidationRequest> consolidationRequests =
-        BeaconBlockBodySchemaElectra.required(beaconBlockBodySchema)
-            .getExecutionPayloadSchema()
-            .getConsolidationRequestsSchemaRequired()
-            .of(consolidationRequest);
-
     spec.getBlockProcessor(state.getSlot())
-        .processConsolidationRequests(state, consolidationRequests);
+        .processConsolidationRequests(state, List.of(consolidationRequest));
```
rolfyone: it seems like we've hard coded the configuration of needing 1 consolidation request half way through, this could be fun when we need to now handle 2 consolidations, i wonder why we bothered going half way with a list?
If we're hard coding 1 anyway, why do we bother taking a list into the spec function?

lucassaldanha: All EL request operation tests in the reference-tests have two parameters: a pre state and a single operation (ssz encoded) as input. 

With the new `ExecutionRequests` container, the alternative would be to update the reference-tests to use that containers ssz as input (instead of a single EL request). This approach is similar to how some operation tests have the `ExecutionPayload` container ssz as an input. However the current reference-tests are still using the individual operations.

TLDR: reference-tests need to change first, before we can change anything here.

lucassaldanha: > If we're hard coding 1 anyway, why do we bother taking a list into the spec function?

On the spec, each block can have a list of each EL request type, so the spec function does require a list. It is only on the reference-tests that the input is a single operation.

lucassaldanha: I'll follow up with the ref-tests to see if we need to change anything there.

rolfyone: the list does make sense it just seems weird to have a single up to a point hten a list, so it'd almost be a better idea to have the list all the way through imo so that when we decide 1 consolidation wasnt enough it's less crazy to fix
but either or, just looked weird.

data/serializer/src/main/java/tech/pegasys/teku/api/schema/electra/BeaconBlockBodyElectra.java
rolfyone: its really common, i'd prefer no comment as its just noise.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/blockbody/versions/electra/BlindedBeaconBlockBodyElectraImpl.java
lucassaldanha: Maybe, but I did the bare minimum on blinded block. This is likely going to be addressed on https://github.com/Consensys/teku/issues/8624

eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/common/operations/DefaultOperationProcessor.java
lucassaldanha: Yeah, did the change it is better :)

rolfyone (approved):
LGTM

PR #8603, opened 2024-09-16
Fix wei/gwei/eth conversions

StefanBratanov:
## PR Description

- `gweiToEth` factor should be 9, not 18.
- `BlockProductionDuty` should use weiToEth since the values are wei for `execution_payload_value` and `consensus_block_value` coming from the API

Technically, it was working before since was using gweiToEth method, which was wrongly using 18, which is the wei to eth factor, but fixed it so the methods becomes more usable.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/Converter.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/Converter.java
@@ -14,22 +14,30 @@
 package tech.pegasys.teku.infrastructure.logging;
 
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.math.RoundingMode;
 import org.apache.tuweni.units.bigints.UInt256;
 import org.web3j.utils.Convert;
+import org.web3j.utils.Convert.Unit;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 
 public class Converter {
 
-  static BigDecimal gweiToEthFactor = BigDecimal.TEN.pow(18);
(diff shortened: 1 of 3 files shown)


infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/Converter.java
```diff
@@ -14,22 +14,30 @@
...
 
   public static String weiToEth(final UInt256 wei) {
     final BigDecimal result = Convert.fromWei(wei.toDecimalString(), Convert.Unit.ETHER);
     return result.setScale(6, RoundingMode.HALF_UP).toString();
   }
 
-  public static String gweiToEth(final UInt256 gwei) {
-    return new BigDecimal(gwei.toBigInteger())
+  public static String gweiToEth(final UInt64 gwei) {
+    return new BigDecimal(gwei.bigIntegerValue())
```
zilm13: nit: we could use `fromWei(toWei)` here or `weiToEth(toWei)` to exclude such errors in future

zilm13 (approved):
LGTM
So the output was correct, but it was correct just because of error in formula, right?

StefanBratanov:
> LGTM So the output was correct, but it was correct just because of error in formula, right?

Yes

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/Converter.java
StefanBratanov: I like that, implemented the second approach

PR #8610, opened 2024-09-17
Add `floodPublish` CLI flag

### LINKED ISSUE #8559: Make `floodPublish` under CLI flag control
After some discussions with LH guys, seems like turning off floodPublish may help especially in case of low upload bandwidth. 

The reason is that it seems to improve dissemination latency if the full message is received earlier by one of the peer of the mesh (rather than start sending the message in parallel to all peers of the mesh). This could help homestakers on low bandwidth, when publishing block and blobs (locally built)

the parameter could be named `--p2p-flood-publish-enabled`, default `true`

context:
`.floodPublish(true)` in `addGossipParamsMiscValues`

KatyaRyazantseva:
## PR Description
Add `--p2p-flood-publish-enabled` CLI flag to enable/disable [flood publishing](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#flood-publishing).
## Fixed Issue(s)
fixes #8559

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java
```diff
@@ -284,6 +288,11 @@ public Builder peerRequestLimit(final Integer peerRequestLimit) {
       return this;
     }
 
+    public Builder isFloodPublishEnabled(final Boolean floodPublishEnabled) {
+      isFloodPublishEnabled = floodPublishEnabled;
```
Nashatyrev: I would add explicit `this.` to be in line with the rest of code

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java
```diff
@@ -42,6 +43,7 @@ public class P2PConfig {
   public static final int DEFAULT_BATCH_VERIFY_QUEUE_CAPACITY = 15_000;
   public static final int DEFAULT_BATCH_VERIFY_MAX_BATCH_SIZE = 250;
   public static final boolean DEFAULT_BATCH_VERIFY_STRICT_THREAD_LIMIT_ENABLED = false;
+  public static final boolean DEFAULT_FLOOD_PUBLISH_ENABLED = true;
```
Nashatyrev: Doesn't it duplicate the same constant in `GossipConfig` ?

Nashatyrev (approved):
LGTM, but I would let someone from Teku team approve this PR as well  
Just some nits from my side

teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
```diff
@@ -366,6 +367,18 @@ The network interface(s) on which the node listens for P2P communication.
...
   private boolean yamuxEnabled = NetworkConfig.DEFAULT_YAMUX_ENABLED;
 
+  // More about flood publishing
+  // https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#flood-publishing
+  @Option(
+      names = {"--p2p-flood-publish-enabled"},
+      paramLabel = "<BOOLEAN>",
+      showDefaultValue = Visibility.ALWAYS,
+      description = "Enables gossip 'floodPublish' feature",
+      hidden = true,
```
tbenr: unit tests detected this is declared hidden even if it is not a `--X` config. You should remove this.

tbenr (commented):
Thanks! looks nice!
I'd fix the `hidden` property and add a test in `P2POptionsTest`. There we check that the parameter actually affects the configuration.

KatyaRyazantseva:
@tbenr ready for the test review

tbenr (approved):
LGTM! Thanks!

PR #8592, opened 2024-09-12
Schema definitions redesign

tbenr:
this is the clean commit without SszTypeGenerator https://github.com/Consensys/teku/pull/8592/commits/fabf2b2d8630685ad15aef9b7a06d0c925cd7a48

this is the messy commit with SszTypeGenerator on Attestations https://github.com/Consensys/teku/pull/8592/commits/d4dbc4cf5431d52e85d77e45827f6fde74c1dba6

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [ ] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #8614, opened 2024-09-18
Introduces `SchemaRegistry` and related classes

tbenr:
This PR introduces
- `SchemaTypes` that will hold the static `schemaId`s for all datastructres types (they won't be milestone dependant, we will have only one even if the schema will change along the different forks)
- `SchemaRegistry` class and `SchemaCache` responsible for implementing the registry with its internal cross-milestone cache
-  `SchemaProvider` interface and a base abstract class which implements the rules for generating a schema (the actual schema instantiation for which milestone(s)) for a given `schemaId`.
- `SchemaRegistryBuilder` will be responsible to build the schema registry for each `milestone` via `build` method.

***Note*** the design of the registry is based on the assumption that all registries will be built at startup (during `SpecVersion` creations) and registries will be immediately "primed". This guarantees that cache writes (`put`s) will happen only during that phase, which is a single-threaded (non concurrent). This is why the cache is not dealing with any concurrency\synchronization. After initialization the cache will be "readonly".

related to #8592 
the full direction can be seen here: https://github.com/Consensys/teku/pull/8592

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/AbstractSchemaProvider.java
```diff
@@ -0,0 +1,82 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.util.EnumMap;
+import java.util.Map;
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.config.SpecConfig;
+import tech.pegasys.teku.spec.schemas.SchemaTypes.SchemaId;
+
+public abstract class AbstractSchemaProvider<T> implements SchemaProvider<T> {
```
lucassaldanha: NIT: we could move this class into a `[..].schemas.providers` package and makes package-private? This way we end up with all new provider implementations in the same package.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaRegistry.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public SpecConfig getSpecConfig() {
+    return specConfig;
+  }
+
+  public void primeRegistry() {
+    for (final SchemaId<?> schemaClass : providers.keySet()) {
+      get(schemaClass);
+    }
+  }
```
lucassaldanha: Q: Do we want to include some gate logic so it is guaranteed we can't register new providers after the registry is primed? In the current implementation, there isn't anything preventing the user from doing so.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaCache.java
```diff
@@ -0,0 +1,45 @@
...
+
+package tech.pegasys.teku.spec.schemas;
+
+import java.util.EnumMap;
+import java.util.HashMap;
+import java.util.Map;
+import tech.pegasys.teku.spec.SpecMilestone;
+import tech.pegasys.teku.spec.schemas.SchemaTypes.SchemaId;
+
+public interface SchemaCache {
```
lucassaldanha: This can be package private.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaRegistry.java
```diff
@@ -0,0 +1,88 @@
...
+        return schema;
+      }
+    }
+    schema = provider.getSchema(this);
+    cache.put(effectiveMilestone, schemaClass, schema);
+    if (effectiveMilestone != milestone) {
+      cache.put(milestone, schemaClass, schema);
+    }
+    return schema;
+  }
```
lucassaldanha: We are relying on the fact that we are calling `SchemaRegistryBuilder` from a "concurrency free" part of the code. I understand this is by design but maybe we can remove that assumption (so it does not bite us in the future).

Once the registry is primed, we should always hit the cache. However, it isn't thread-safe before priming. 

What about separating the logic on `get()` to only read the cache. If we attempt to call `get()` w/o priming the registry first, we fail. We can move the logic that populates the cache to `primeRegistry()`, and make it synchronized so we know it is thread-safe no matter where it is being called from.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaCache.java
```diff
@@ -0,0 +1,45 @@
...
+  static SchemaCache createDefault() {
+    return new SchemaCache() {
+      private final Map<SpecMilestone, Map<SchemaId<?>, Object>> cache =
+          new EnumMap<>(SpecMilestone.class);
+
+      @SuppressWarnings("unchecked")
+      @Override
+      public <T> T get(final SpecMilestone milestone, final SchemaId<T> schemaId) {
+        return (T) cache.computeIfAbsent(milestone, __ -> new HashMap<>()).get(schemaId);
```
lucassaldanha: I am a bit confused by this part. If we are always priming the registry, and the registry is using `SchemaCache.put()` to update the cache, why do we need to have the `cache.computeIfAbsent` logic here? Can't we just use `get()`?

I belive if we use `computeIfAbsent` here, we will risk updating the underlying EnumMap concurrently, unless I am missing something?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaTypes.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.common.base.Converter;
+import com.google.common.base.MoreObjects;
+import java.util.Locale;
+import tech.pegasys.teku.infrastructure.ssz.collections.SszBitvector;
+import tech.pegasys.teku.infrastructure.ssz.schema.collections.SszBitvectorSchema;
+import tech.pegasys.teku.spec.SpecMilestone;
+
+public class SchemaTypes {
+  // PHASE0
+  public static final SchemaId<SszBitvectorSchema<SszBitvector>> ATTNETS_ENR_FIELD_SCHEMA =
```
lucassaldanha: Q: Did you leave this here as an example or did you forget to use it? It looks like this SchemaId isn't being used anywhere yet.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaRegistry.java
```diff
@@ -0,0 +1,88 @@
...
+import tech.pegasys.teku.spec.config.SpecConfig;
+import tech.pegasys.teku.spec.schemas.SchemaTypes.SchemaId;
+
+public class SchemaRegistry {
+  private final Map<SchemaId<?>, SchemaProvider<?>> providers = new HashMap<>();
+  private final SpecMilestone milestone;
+  private final SchemaCache cache;
+  private final SpecConfig specConfig;
+
+  public SchemaRegistry(
```
lucassaldanha: We are relying on the builder to instantiate the registry to the outside world so we can change this to package-private.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaRegistry.java
tbenr: Nice idea!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaTypes.java
tbenr: Yes it was intentional

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaRegistry.java
tbenr: Yes I agree! Generally, a better separation between priming and lookup reflects the usage we expect.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaCache.java
tbenr: you are right..traces of me using AI here...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/SchemaRegistry.java
tbenr: Relooking at this, the separation can't be done.
The way this works is that, when the schema is being created, it can lookup in the registry itself so it may trigger a creation of a schema (dependency). This help us not dealing with ordering (it works as soon as there are no dependency loops).

The real point here is if we want or not a dependency loop detection. To me it can only happen if as a developer you pick the wrong schemaId (and it is compatible with the type you need in that moment). Let me think

tbenr:
- I added a loop dependency check (just in case) because it won't affect performance after everything is primed and could catch some dev problems instead of causing a bad stack overflow.

- I added some general checks in the registry but no concurrency handling, relying on the building phase to be synchronized.

- from visibility perspective, I decided to move everything under the same package `registry` because it makes all visibility clear: the only public things are `Builder.build` and `Registry.get`, all the rest is package private. The drawback is that we will have all providers together with the main classes, but I think it pays off.

- Added some additional safety checks in the Builder.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/AbstractSchemaProvider.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  private void checkIfAlreadyMapped(final SpecMilestone milestone) {
+    if (milestoneToEffectiveMilestone.containsKey(milestone)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Milestone %s is already mapped to %s",
+              milestone, milestoneToEffectiveMilestone.get(milestone)));
+    }
+  }
```
mehdi-aouadi: I'd suggest separating the mapping overlapping check and make the add milestone mapping fails fast. I was confused when trying to understand the logic mainly because in some cases, we were adding the mapping and then failing.
What do you think if we use a `TreeMap` to hold a sorted `milestoneToEffectiveMilestone` and use the `floorEntry` and `ceilingEntry` to check the overlapping as follows:

```suggestion
  private final NavigableMap<SpecMilestone, SpecMilestone> milestoneToEffectiveMilestone =
      new TreeMap<>();
  private final SchemaId<T> schemaId;

  protected AbstractSchemaProvider(final SchemaId<T> schemaId) {
    this.schemaId = schemaId;
  }

  protected void addMilestoneMapping(
      final SpecMilestone milestone, final SpecMilestone untilMilestone) {
    checkArgument(
        untilMilestone.isGreaterThan(milestone),
        "%s must be earlier than %s",
        milestone,
        untilMilestone);

    checkOverlappingVersionMappings(milestone, untilMilestone);

    SpecMilestone currentMilestone = untilMilestone;
    while (!currentMilestone.equals(milestone)) {
      milestoneToEffectiveMilestone.put(currentMilestone, milestone);
      currentMilestone = currentMilestone.getPreviousMilestone();
    }
  }

  private void checkOverlappingVersionMappings(final SpecMilestone milestone, final SpecMilestone untilMilestone) {
    final Map.Entry<SpecMilestone, SpecMilestone> floorEntry =
        milestoneToEffectiveMilestone.floorEntry(untilMilestone);
    if (floorEntry != null && floorEntry.getValue().isGreaterThanOrEqualTo(milestone)) {
      throw new IllegalArgumentException(
          String.format(
              "Milestone %s is already mapped to %s",
              floorEntry.getKey(), getEffectiveMilestone(floorEntry.getValue())));
    }
    final Map.Entry<SpecMilestone, SpecMilestone> ceilingEntry =
        milestoneToEffectiveMilestone.ceilingEntry(milestone);
    if (ceilingEntry != null && ceilingEntry.getKey().isLessThanOrEqualTo(untilMilestone)) {
      throw new IllegalArgumentException(
          String.format(
              "Milestone %s is already mapped to %s",
              ceilingEntry.getKey(), getEffectiveMilestone(ceilingEntry.getValue())));
    }
  }
```

lucassaldanha (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/AbstractSchemaProvider.java
tbenr: I see what you mean, and your version could be indeed more readable. I'll check it out.

PR #8615, opened 2024-09-18
Add schema registry in schema definitions

tbenr:
Adds the registry building in SpecVersion and provides registry in schemaDefinitions

It shares `schemaRegistryBuilder` among milestones creations (when possible) so that the global cache is shared and non-changing schemas will maintain the same instance cross-milestones

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #8626, opened 2024-09-21
[Draft] Add database reset on Ephemery Network

### LINKED ISSUE #8589: Add automatic reset for  Ephemery Network
### Description

- Manual delete of db

Always

gconnect:
## PR Description
- Add deposit chainId to the generated network.yml file
- Reset the db

## Todo
- Still need to ensure that the above only happens on Ephemery

Doing this tends not to work as expected
```
  final Eth2NetworkConfiguration networkConfig =
        Eth2NetworkConfiguration.builder(Eth2Network.EPHEMERY).build();

    if (networkConfig.getEth2Network().equals(Optional.of(Eth2Network.EPHEMERY))) {
      ....
    }
```
Still figuring how to handle the check.

## Fixed Issue(s)
 "fixes #8589" 

## Documentation

## Changelog


gconnect:
@rolfyone please take a look when you have the time.

rolfyone:
Thanks for raising, will review shortly. One initial comment - 
We'll only want 'partially addresses' in the 'fixes, because this would otherwise close the issue...

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
...
     final String forkVersionString = forkVersion.toHexString().toLowerCase(Locale.ROOT);
     final String depositContractString = depositContract.toHexString().toLowerCase(Locale.ROOT);
     final ObjectMapper objectMapper =
         new ObjectMapper(new YAMLFactory().disable(WRITE_DOC_START_MARKER));
+
     if (source.exists()) {
       final DatabaseNetwork databaseNetwork =
           objectMapper.readerFor(DatabaseNetwork.class).readValue(source);
 
+      // Fork version validation
```
rolfyone, line 79: we can remove these comments, they don't really add to the readability of the code... (basically all of the comments in this file at least)

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
...
+                  depositChainId.get(),
+                  String.valueOf(databaseNetwork.depositChainId)));
+        }
+      }
       return databaseNetwork;
     } else {
-      DatabaseNetwork databaseNetwork =
-          new DatabaseNetwork(forkVersionString, depositContractString);
+      // Handle Ephemery or other networks
+      String depositChainIdString = depositChainId.orElse(null);
```
rolfyone, line 109: final

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
...
+      DatabaseNetwork databaseNetwork;
+      System.out.println(
+          "network " + NETWORK_CONFIG.getEth2Network().equals(Optional.of(Eth2Network.EPHEMERY)));
```
rolfyone, line 112: we don't want `System.out` in our code, we'd want to be using loggers... Is this just a test message?

gconnect:
> Thanks for raising, will review shortly. One initial comment - We'll only want 'partially addresses' in the 'fixes, because this would otherwise close the issue...

Not sure what you meant here.

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
gconnect: Alright, I will remove it.

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
...
+      if (NETWORK_CONFIG.getEth2Network().equals(Optional.of(Eth2Network.EPHEMERY))) {
+        databaseNetwork =
+            new DatabaseNetwork(forkVersionString, depositContractString, depositChainIdString);
+      } else {
+        databaseNetwork = new DatabaseNetwork(forkVersionString, depositContractString, null);
+      }
```
rolfyone, line 118: We can just always write deposit_chain_id to the file which makes this a lot more readable. The conditional check I'd be more worried about would be ensuring we don't crash if it's not there on startup once we're checking that.

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -95,11 +140,12 @@ public boolean equals(final Object o) {
     }
     final DatabaseNetwork that = (DatabaseNetwork) o;
     return Objects.equals(forkVersion, that.forkVersion)
-        && Objects.equals(depositContract, that.depositContract);
+        && Objects.equals(depositContract, that.depositContract)
+        && Objects.equals(depositChainId, that.depositChainId); // Compare deposit_chain_id
```
rolfyone, line 144: remove comment

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
...
+  // Single constructor that handles both cases
   @JsonCreator
   DatabaseNetwork(
       @JsonProperty("fork_version") final String forkVersion,
-      @JsonProperty("deposit_contract") final String depositContract) {
+      @JsonProperty("deposit_contract") final String depositContract,
+      @JsonProperty("deposit_chain_id") final String depositChainId) {
     this.forkVersion = forkVersion;
     this.depositContract = depositContract;
+    this.depositChainId = depositChainId; // Can be null for non-Ephemery networks
```
rolfyone, line 60: remove comment - will also be null for any old databases that ARE ephemery that you open... we can't assume all ephemery will have this data.
Also though it's less important, people just have to manually reset their db (it'll literally be a dev - you or me basically)

I do also think that we could have a constructor with just the old interface and default the new field to null, which would simplify testing and make it clearer...

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
...
   }
 
+  // Initialization logic
   public static DatabaseNetwork init(
-      final File source, final Bytes4 forkVersion, final Eth1Address depositContract)
+      final File source,
+      final Bytes4 forkVersion,
+      final Eth1Address depositContract,
+      final Optional<String> depositChainId)
```
rolfyone, line 68: make this a long coming in, not optional

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
...
+        if (depositChainId.isPresent()
+            && databaseNetwork.depositChainId != null
+            && !databaseNetwork.depositChainId.equals(depositChainId.get())) {
+          throw DatabaseStorageException.unrecoverable(
+              formatMessage(
+                  "deposit chain ID",
+                  depositChainId.get(),
+                  String.valueOf(databaseNetwork.depositChainId)));
+        }
+      }
```
rolfyone, line 105: defer this check to another PR please. it'll make testing a lot easier to see.

storage/src/test/java/tech/pegasys/teku/storage/server/network/DatabaseNetworkTest.java
```diff
@@ -37,11 +37,12 @@ public void shouldCreateNetworkFile(@TempDir final File tempDir) throws IOExcept
     assertThat(networkFile).doesNotExist();
     final Bytes4 fork = dataStructureUtil.randomFork().getCurrentVersion();
     final Eth1Address eth1Address = dataStructureUtil.randomEth1Address();
-    assertThat(DatabaseNetwork.init(networkFile, fork, eth1Address))
+    assertThat(DatabaseNetwork.init(networkFile, fork, eth1Address, null))
```
rolfyone, line 40: i'd be surprised that any of these tests pass? we're pushing a null into an optional field which rarely ends well...
My suggestion above was to make the init not take an optional, so this will likely be a better answer after that...

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
```diff
@@ -372,4 +393,48 @@ private void saveStorageMode(final StateStorageMode storageMode) {
...
+                p -> {
+                  try {
+                    Files.delete(p);
+                  } catch (IOException e) {
+                    throw new RuntimeException("Failed to delete file: " + p, e);
+                  }
+                });
+      }
+    }
+  }
```
rolfyone, line 439: We'll defer this to another PR please, lets get the new field in and not breaking anything.

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedDatabaseFactory.java
```diff
@@ -233,7 +242,10 @@ private Database createLevelDbV1Database() {
       final V5DatabaseMetadata metaData =
           V5DatabaseMetadata.init(getMetadataFile(), V5DatabaseMetadata.v5Defaults());
       DatabaseNetwork.init(
-          getNetworkFile(), spec.getGenesisSpecConfig().getGenesisForkVersion(), eth1Address);
+          getNetworkFile(),
+          spec.getGenesisSpecConfig().getGenesisForkVersion(),
+          eth1Address,
+          Optional.of(String.valueOf(spec.getGenesisSpecConfig().getDepositChainId())));
```
rolfyone, line 248: i feel like at this level we just want the depositChainId as a long going in...

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,39 +42,81 @@ public class DatabaseNetwork {
   @VisibleForTesting
   final String depositContract;
 
+  @JsonProperty("deposit_chain_id")
+  @VisibleForTesting
+  final String depositChainId; // Can be null for non-Ephemery networks
+
+  static final Eth2NetworkConfiguration NETWORK_CONFIG =
+      Eth2NetworkConfiguration.builder(Eth2Network.EPHEMERY).build();
```
rolfyone, line 50: Ideally I think we avoid something like this... at the very least we just want to add our deposit chain id in this pr.

rolfyone (commented):
We could potentially treat this as a 'tracer' PR and just create a new branch with what we want in the first one... Alternatively we need to remove all of the functionality other than the network yaml change
Basically, I need a PR that just changes the network yaml and doesn't break anything when reading the old configuration file that doesn't have the chain id in it... This is really critical, because thats basically ensuring its not a breaking change.

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
gconnect: > we don't want `System.out` in our code, we'd want to be using loggers... Is this just a test message?

Yes it a test message, forgot to remove it before pushing

gconnect:
> We could potentially treat this as a 'tracer' PR and just create a new branch with what we want in the first one... Alternatively we need to remove all of the functionality other than the network yaml change Basically, I need a PR that just changes the network yaml and doesn't break anything when reading the old configuration file that doesn't have the chain id in it... This is really critical, because thats basically ensuring its not a breaking change.

What changes will I apply to the network.yaml file if I am not adding the chainId?

gconnect:
opened a new PR here https://github.com/Consensys/teku/pull/8631

PR #8631, opened 2024-09-23
Update network file

### LINKED ISSUE #8589: Add automatic reset for  Ephemery Network

gconnect:
## PR Description
- Add optional chainId and constructor
- Wrote test to write and readfile with and without chainId

This PR is in partial fulfilment of the below issue.

## Fixed Issue(s)
fixes #8589

## Documentation

## Changelog


gconnect:
@rolfyone please review when you have the time

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,12 +40,23 @@ public class DatabaseNetwork {
   @VisibleForTesting
   final String depositContract;
 
+  @JsonInclude(JsonInclude.Include.NON_NULL)
```
rolfyone: i believe this goes on the class, then on the other 2 fields you can put `required = true` in their JsonProperty...

We could easily write a test case that shows the 2 first fields are required by opening an empty file or just putting in one of the fields and seeing the failure condition...

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,12 +40,23 @@ public class DatabaseNetwork {
...
       @JsonProperty("fork_version") final String forkVersion,
-      @JsonProperty("deposit_contract") final String depositContract) {
+      @JsonProperty("deposit_contract") final String depositContract,
+      @JsonProperty("deposit_chainId") final Long depositChainId) {
     this.forkVersion = forkVersion;
     this.depositContract = depositContract;
+    this.depositChainId = depositChainId;
+  }
+
+  DatabaseNetwork(final String forkVersion, final String depositContract) {
```
rolfyone, line 59: can we annotate this one `@VisibleForTesting`

storage/src/test/java/tech/pegasys/teku/storage/server/network/DatabaseNetworkTest.java
```diff
@@ -82,4 +86,45 @@ public void shouldNotThrowIfForkAndContractMatch(@TempDir final File tempDir) th
...
+
+    final Bytes4 fork = dataStructureUtil.randomFork().getCurrentVersion();
+    final Eth1Address eth1Address = dataStructureUtil.randomEth1Address();
+    final Long depositId = dataStructureUtil.randomLong();
+    DatabaseNetwork databaseNetwork =
+        new DatabaseNetwork(fork.toHexString(), eth1Address.toHexString(), depositId);
+
+    ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
+    objectMapper.writerFor(DatabaseNetwork.class).writeValue(networkFile, databaseNetwork);
+    DatabaseNetwork readDatabaseNetwork =
```
rolfyone: can add some finals

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -39,12 +40,24 @@ public class DatabaseNetwork {
...
+  @JsonProperty("deposit_chainId")
+  @VisibleForTesting
+  final Long depositChainId;
+
   @JsonCreator
   DatabaseNetwork(
       @JsonProperty("fork_version") final String forkVersion,
-      @JsonProperty("deposit_contract") final String depositContract) {
+      @JsonProperty("deposit_contract") final String depositContract,
+      @JsonProperty("deposit_chainId") final Long depositChainId) {
```
rolfyone: the name here is a mix of 2 worlds, i think deposit_chain_id would be more consistent with other fields.

storage/src/test/java/tech/pegasys/teku/storage/server/network/DatabaseNetworkTest.java
```diff
@@ -82,4 +86,45 @@ public void shouldNotThrowIfForkAndContractMatch(@TempDir final File tempDir) th
...
+      throws IOException {
+    final File networkFile = new File(tempDir, "network.yml");
+
+    final Bytes4 fork = dataStructureUtil.randomFork().getCurrentVersion();
+    final Eth1Address eth1Address = dataStructureUtil.randomEth1Address();
+    final Long depositId = dataStructureUtil.randomLong();
+    final DatabaseNetwork databaseNetwork =
+        new DatabaseNetwork(fork.toHexString(), eth1Address.toHexString(), depositId);
+
+    final ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
```
rolfyone: can be shared between tests, create with same option as the static function in real implementation

storage/src/test/java/tech/pegasys/teku/storage/server/network/DatabaseNetworkTest.java
```diff
@@ -82,4 +86,45 @@ public void shouldNotThrowIfForkAndContractMatch(@TempDir final File tempDir) th
...
+
+    assertEquals(fork.toHexString(), readDatabaseNetwork.forkVersion);
+    assertEquals(eth1Address.toHexString(), readDatabaseNetwork.depositContract);
+    assertEquals(depositId, readDatabaseNetwork.depositChainId);
+  }
+
+  @Test
+  void shouldWriteAndReadDatabaseNetworkWithoutChainId(@TempDir final File tempDir)
+      throws IOException {
+    final File networkFile = new File(tempDir, "network.yml");
```
rolfyone: just use the string in VersionedDatabaseFactory for filename...

storage/src/test/java/tech/pegasys/teku/storage/server/network/DatabaseNetworkTest.java
```diff
@@ -82,4 +86,45 @@ public void shouldNotThrowIfForkAndContractMatch(@TempDir final File tempDir) th
...
+
+    final Bytes4 fork = dataStructureUtil.randomFork().getCurrentVersion();
+    final Eth1Address eth1Address = dataStructureUtil.randomEth1Address();
+    final DatabaseNetwork databaseNetwork =
+        new DatabaseNetwork(fork.toHexString(), eth1Address.toHexString(), null);
+
+    final ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
+    objectMapper.writerFor(DatabaseNetwork.class).writeValue(networkFile, databaseNetwork);
```
rolfyone, line 132: because we're writing here with the mapper, we'll want to read the file by hand and check deposit_chain_id is not present in the file... otherwise we're not really testing what we think we are here

storage/src/test/java/tech/pegasys/teku/storage/server/network/DatabaseNetworkTest.java
gconnect: `  @VisibleForTesting static final String NETWORK_FILENAME = "network.yml";
`
This is not public. I am unable to access it in the test.

storage/src/test/java/tech/pegasys/teku/storage/server/network/DatabaseNetworkTest.java
rolfyone: just this one looks unresolved @gconnect  can we read the file nad make sure its not got that field?

gconnect: Done

rolfyone (approved):
LGTM

PR #8639, opened 2024-09-24
Add deposit chainId to network file on start

### LINKED ISSUE #8589: Add automatic reset for  Ephemery Network

gconnect:
## PR Description
Add deposit chainId to network config file on start

## Fixed Issue(s)
fixes #8589 

## Documentation

## Changelog


gconnect:
@rolfyone please review when you have the time.

rolfyone (approved):
LGTM - thanks.

PR #8640, opened 2024-09-24
Implement a DatabaseArchiveWriter and add to pruneOldestBlobSidecars

### LINKED ISSUE #8513: Allow the option to store blobs to filesystem as they're being removed from database
Blobs are currently stored for the mandatory time in database and then deleted.

Rather than simply deleting them, it'd be helpful to be able to write the blobs to filesystem.

I could imagine keeping this relatively simple and just specifying something like `--data-storage-store-old-blobs-path=/data` and we'd then start writing all blobs to the filesystem before they get archived.

Theres a tool already in use https://github.com/base-org/blob-archiver that basically dumps the blobs api output per block, we could do similar re. file naming etc potentially if its not too crazy.. Remembering there'll be 7200 slots per day so we probably want output in some kind of structure.

Work is split into following tasks:
- [x] #8640 adds the concept of a DatabaseArchiveWriter functional interface. This provides the interface to have BlobSidecars written to an seperate archive when pruned.
- [x] #8674 Create a DataArchive interface that has a file system based implementation (e.g. FileSystemDataArchive) to allow writing to BlobSidecars out.
- [x] #8674 Create the configuration option to a DatabaseArchiveWriter implementation to write to the FileSystemDataArchive.
- [x] Add read functionality to the DataArchive interface and allow the ability to return the selected BlobSidecar by slot.
- [ ] Update the Beacon-api /blob_sidecars/{block_id} to allow searching both the database and DataArchive if available.

david-ry4n:
## PR Description
Initial change to add the concept of DatabaseArchiveWriter to pruneOldestBlobSidecars method. The initial implementation is a NoOp that returns true. See #8513 for additional tasks. With this change it sets up the structure. Future changes will implement a DatabaseArchive interface and required configuration options to configure where/how the archive is written.

## Fixed Issue(s)

partially addresses #8513 

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlobSidecarPruner.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlobSidecarPruner.java
@@ -32,7 +32,10 @@ import tech.pegasys.teku.service.serviceutils.Service;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.config.SpecConfig;
 import tech.pegasys.teku.spec.config.SpecConfigDeneb;
+import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.BlobSidecar;
 import tech.pegasys.teku.storage.server.Database;
+import tech.pegasys.teku.storage.server.DatabaseArchiveNoopWriter;
+import tech.pegasys.teku.storage.server.DatabaseArchiveWriter;
 import tech.pegasys.teku.storage.server.ShuttingDownException;
 
 public class BlobSidecarPruner extends Service {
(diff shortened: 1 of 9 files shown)


tbenr:
welcome to `MethodInputParametersMustBeFinal` world 😄

storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlobSidecarPruner.java
```diff
@@ -81,6 +85,9 @@ public BlobSidecarPruner(
     this.pruningActiveLabelledGauge = pruningActiveLabelledGauge;
     this.storeNonCanonicalBlobSidecars = storeNonCanonicalBlobSidecars;
 
+    // To be updated with other implementations. e.g. filesystem or s3
```
tbenr, line 88: I'd use `// TODO: ` prefix so it will be captured nicely by intellij, or simply remove since we have a (set of) issues on the topic.

tbenr (approved):
LGTM. just a nit

storage/src/main/java/tech/pegasys/teku/storage/server/pruner/BlobSidecarPruner.java
david-ry4n: I'll fix that up in the next PR. Thanks for feedback.

PR #8642, opened 2024-09-24
Check network on start and reset db

gconnect:
## PR Description
- Check if the network is Ephemery and due for an update then reset the db.

## Fixed Issue(s)
fixes #[8589](https://github.com/Consensys/teku/issues/8589)

## Documentation

## Changelog


gconnect:
@rolfyone please review when you have the time. Thanks!

storage/src/main/java/tech/pegasys/teku/storage/server/network/DatabaseNetwork.java
```diff
@@ -84,6 +87,15 @@ public static DatabaseNetwork init(
             formatMessage(
                 "deposit contract", depositContractString, databaseNetwork.depositContract));
       }
+      if (!depositContractString.equals(EPHEMERY_DEPOSIT_CONTRACT_ADDRESS)
+          && databaseNetwork.depositChainId != null
+          && !databaseNetwork.depositChainId.equals(depositChainId)) {
```
rolfyone: so ideally check not null then check everything else...

gconnect: fixed

rolfyone:
I'm not sure that's really at the right point currently, because we're leaking that deposit check in multiple places, and theres public functions that can delete folders etc, so i think we really need to think about how it all fits together... this feels very confused...

Are you able to have a look at how we might do this neatly? otherwise we can find a little time to get some design thought out so its relatively clean...

gconnect:
> I'm not sure that's really at the right point currently, because we're leaking that deposit check in multiple places, and theres public functions that can delete folders etc, so i think we really need to think about how it all fits together... this feels very confused...
> 
> Are you able to have a look at how we might do this neatly? otherwise we can find a little time to get some design thought out so its relatively clean...

Thanks for reviewing. I will dive deeper and see how I can make the code better.

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -80,14 +84,24 @@ protected SafeFuture<?> doStart() {
                       1,
                       DEFAULT_MAX_QUEUE_SIZE,
                       Thread.NORM_PRIORITY - 1);
-              final VersionedDatabaseFactory dbFactory =
```
rolfyone: this can still be final, we're still only instantiating the factory once...

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -80,14 +84,24 @@ protected SafeFuture<?> doStart() {
...
+                database.migrate();
+              } catch (EphemeryException e) {
+                try {
+                  resetDatabaseDirectories(serviceConfig);
+                  database = dbFactory.createDatabase();
+                  database.migrate();
+                } catch (Exception ex) {
+                  throw new RuntimeException("Failed to reset and recreate the database.", ex);
+                }
+              }
```
rolfyone: yep kind of. we'll just clean this up a little bit, but its superficial...
```
              try {
                database = dbFactory.createDatabase();
              } catch (EphemeryException e) {
                database = resetDatabaseAndCreate(dbfactory);
             }
            database.migrate();

......
 private Database resetDatabaseAndCreate(dbfactory) {
   try {
       return dbFactory.createDatabase();
  } catch (final Exception ex) {
    throw new InvalidConfigurationException("The existing ephemery database was old, and was unable to reset it.", ex);
 }
        
}
```

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -222,4 +236,30 @@ protected SafeFuture<?> doStop() {
...
   }
+
+  /** This method is called only on Ephemery network when reset is due. */
+  void resetDatabaseDirectories(final ServiceConfig serviceConfig) throws IOException {
+    final Path beaconDataDir = serviceConfig.getDataDirLayout().getBeaconDataDirectory();
+    final Path slashProtectionDir =
+        serviceConfig.getDataDirLayout().getValidatorDataDirectory().resolve("slashprotection");
+    deleteDirectoryRecursively(beaconDataDir);
+    deleteDirectoryRecursively(slashProtectionDir);
+  }
```
rolfyone: now we probably want to setup some testing around this a little...

rolfyone: Even if we just did a test around startup, or directly on this... also depending on how that washes out we'd probably put VisibleForTesting

rolfyone (commented):
We're getting there for sure! Hope the latest comments help!

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
rolfyone: side note - another PR currently is removing that 'migrate' function.. so it'l disappear

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
gconnect: > now we probably want to setup some testing around this a little...

Not sure which of the files the test should go in.

gconnect:
@rolfyone I have added some test please review and let me know what you think.

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -58,6 +59,7 @@ public class StorageService extends Service implements StorageServiceFacade {
   private final boolean depositSnapshotStorageEnabled;
   private final boolean blobSidecarsStorageCountersEnabled;
   private static final Logger LOG = LogManager.getLogger();
+  private final EphemeryDatabaseReset ephemeryDatabaseReset;
```
rolfyone: this can just be a local variable where we use it

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -85,10 +88,12 @@ protected SafeFuture<?> doStart() {
...
                       serviceConfig.getDataDirLayout().getBeaconDataDirectory(),
                       config);
-              database = dbFactory.createDatabase();
-
-              database.migrate();
 
+              try {
+                database = dbFactory.createDatabase();
+              } catch (EphemeryException e) {
+                database = ephemeryDatabaseReset.resetDatabaseAndCreate(serviceConfig, dbFactory);
```
rolfyone, line 95: silently doing things like this can make it hard to reason about when doing support tickets. 
I'd suggest 
 - logging a message here as a warning - something like `"Ephemery network deposit contract has updated, resetting the stored database and slashing protection data."` or something...
 - moving the variable definition inside this block so we can see its only in this one place that its used.

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/EphemeryDatabaseReset.java
```diff
@@ -0,0 +1,63 @@
...
+import tech.pegasys.teku.storage.server.Database;
+import tech.pegasys.teku.storage.server.VersionedDatabaseFactory;
+
+public class EphemeryDatabaseReset {
+
+  /** This method is called only on Ephemery network when reset is due. */
+  Database resetDatabaseAndCreate(
+      final ServiceConfig serviceConfig, final VersionedDatabaseFactory dbFactory) {
+    try {
+      final Path beaconDataDir = serviceConfig.getDataDirLayout().getBeaconDataDirectory();
```
rolfyone, line 30: we need to be careful here, we're actually deleting more than we want.
beacon has more than just the db folder in it...
All I want deleted is `db` and `network.yml` - the rest can stay, and some is troublesome to lose sometimes like `kvstore` folder.
We should verify this in a non mock test also, or expand mock tests to show that we're only calling this on a few specific paths.

services/chainstorage/src/test/java/tech/pegasys/teku/services/chainstorage/EphemeryDatabaseResetTest.java
```diff
@@ -55,30 +59,44 @@ void setUp() throws IOException {
...
+    Files.createDirectory(kvStoreDir);
+    final Path dbVersion = beaconDataDir.resolve("db.version");
+    Files.createFile(dbVersion);
+
     when(dbFactory.createDatabase()).thenReturn(database);
+
     final Database result = ephemeryDatabaseReset.resetDatabaseAndCreate(serviceConfig, dbFactory);
-    verify(ephemeryDatabaseReset).deleteDirectoryRecursively(beaconDataDir);
+    verify(ephemeryDatabaseReset).deleteDirectoryRecursively(dbDataDir);
+    verify(ephemeryDatabaseReset).deleteFile(networkFile.toPath());
```
rolfyone: this is pretty close now i think, we probably just also want to `verifyNoMoreInteractions` after what we expect... in basically every test if we're defining explicitly what we want.

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/StorageService.java
```diff
@@ -92,7 +90,10 @@ protected SafeFuture<?> doStart() {
               try {
                 database = dbFactory.createDatabase();
               } catch (EphemeryException e) {
+                final EphemeryDatabaseReset ephemeryDatabaseReset = new EphemeryDatabaseReset();
                 database = ephemeryDatabaseReset.resetDatabaseAndCreate(serviceConfig, dbFactory);
+                LOG.warn(
+                    "Ephemery network deposit contract id has updated, resetting the stored database and slashing protection data.");
```
rolfyone, line 94: these sorts of things we want to print prior to the logic in case this is what kills the system, then we still have the context...

services/chainstorage/src/main/java/tech/pegasys/teku/services/chainstorage/EphemeryDatabaseReset.java
```diff
@@ -38,8 +37,13 @@ Database resetDatabaseAndCreate(
       } else {
         slashProtectionDir = validatorDataDir.resolve("slashprotection");
       }
+
+      System.out.println("beaconDir" + beaconDataDir);
+      System.out.println("networkFile" + networkFile);
+      System.out.println("dbdir" + dbDataDir);
```
rolfyone: we can't ever merge System printlns...

gconnect: Apologies, forgot to remove it before pushing.

rolfyone (approved):
Thanks for your patience, looks better now. will move towards merging.

PR #8653, opened 2024-09-27
Fix blockImportCompleted

tbenr:
When we revisited the block publishing WRT `BroadcastValidationLevel` we started returning a `Future<BlockImportAndBroadcastValidationResults>` from blockImport, so we started logging `blockImportCompleted` at the wrong moment.

## Documentation

## Changelog


rolfyone (approved):
LGTM

lucassaldanha:
We need a changelog entry about this fix. I'll add one before merging this PR.

Issue #8622, opened 2024-09-19
Add support for `engine_getBlobs`


tbenr:
We can add support for this new api call and integrate it in our flow in `BlockBlobSidecarsTrackersPoolImpl`.

Looking at https://github.com/ethereum/execution-apis/pull/559 looks like some ELs have already implemented it.

Very high level, if EL supports the api, we can ask to EL missing blobs as soon as the block arrives and put them in the corresponding tracker. This can complete (or improve) the tracker blobs and avoid (or reduce) the number of RPC lookups we do to retrieve them (`BlockBlobSidecarsTrackersPoolImpl::fetchMissingContent`) when attestation due approaches.

PR #8662, opened 2024-09-30
add engine get blobs

mehdi-aouadi:
## PR Description
Implement [`engine_getBlobsV1`](https://github.com/ethereum/execution-apis/pull/559)

## Fixed Issue(s)
#8622 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/executionlayer/ExecutionLayerChannelStub.java
```diff
@@ -348,6 +351,22 @@ public SafeFuture<List<ClientVersion>> engineGetClientVersion(final ClientVersio
...
+            slot,
+            blobsToGenerate.orElseGet(
+                () -> random.nextInt(spec.getMaxBlobsPerBlock().orElseThrow() + 1)));
+    final List<KZGCommitment> commitments = blobsUtil.blobsToKzgCommitments(blobs);
+    final List<KZGProof> proofs = blobsUtil.computeKzgProofs(blobs, commitments);
+    return SafeFuture.completedFuture(
+        IntStream.range(0, blobs.size())
+            .mapToObj(index -> new BlobAndProof(blobs.get(index), proofs.get(index)))
+            .toList());
+  }
```
tbenr: we need to be careful returning random stuff here because we could make our production code go in an inconsistent state and stop working with the stub.
Maybe it is not a problem but currently my gut says we should think about it.

rolfyone: i'd be ok with the stub just not returning data? otherwise we can't use stub on mainnet any more for testing or playing about...
potentially i guess we might end up being able to turn that on/off at some point or something...

mehdi-aouadi: I'm thinking we could either return an empty list and consider the `engineGetBlobs` unavailable when using the stub EL or simulate the blobs mempool by caching the generated blobs when we call `engineGetPayload` for the block generation and serve those (a LRU cache for example). Using a cache might make the blobs always available from the mempool though unless adjusting the cache capacity to make them unavailable  sometimes to simulate real world situations

tbenr: this could become complicated and not very useful by default, better no data for sure

tbenr: > an empty list

a list of nulls :-)

tbenr: To avoid nulls we could migrate to `List<Optional<BlobAndProof>>`

mehdi-aouadi: Done, made the blobs optional and made the stub return a list of empty optionals

tbenr (approved):
LGTM

PR #8663, opened 2024-09-30
Adopt `engineGetBlobs`

### LINKED ISSUE #8622: Add support for `engine_getBlobs`

tbenr:
This PR implements:

- query to local el via `engine_GetBlobsV1` when blobs are missing. If blobs are still missing after the query, RPC request will be triggered against remote peers asking for remaining blobs
- when a blob is successfully recovered via local el, than it will be published over p2p. This is the reason why the `blob_sidecar` is fully reconstructed when interacting with local EL.

fixes #8622

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Updated LUKSO configuration with Deneb fork scheduled for epoch 123075 (November 20, 2024, 16:20:00 UTC)
 - Support for `IDONTWANT` libp2p protocol messages
 - `/eth/v1/node/peers` endpoint now populates `enr` field of the peer whenever is possible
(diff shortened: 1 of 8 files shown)


mehdi-aouadi (approved):
LGTM

Issue #8620, opened 2024-09-19
[Engine API] Extract execution requests from payload


lucassaldanha:
References: 
- https://github.com/ethereum/execution-apis/pull/591
- https://github.com/ethereum/consensus-specs/pull/3950

Tasks:
- [x] Update `EngineNewPayloadV4` to pass new `executionRequestHash` parameter
- [x] Implement decoding/hashing logic for execution requests
- [x] Update block validation with EL to include the hash of execution requests on `NewPayloadRequest`
- [x] Update `EngineGetPayloadV4` to receive new parameter `executionRequests`
- [x] Update block creation to consume execution requests (cleanup + TODOs)
- [x] Remove `ExecutionPayloadV4`
- [x] Update NewPayloadV4 to send full requests data (not only the hash)

PR #8666, opened 2024-10-01
Added executionRequestsHash to EngineNewPayloadV4

lucassaldanha:
## PR Description
- Replaced `ExecutionPayloadV4` for `ExecutionPayloadV3` in `EngineNewPayloadV4` (eventually `ExecutionPayloadV4` will be deleted)
- Added new parameter `executionRequestHash` to `EngineNewPayloadV4`
- Updated `ExecutionClientHandler.#engineNewPayload` method with a new parameter `slot`. This is used to resolve the milestone later, replacing the old logic that was getting the milestone from the `executionPayload`
- Updated `NewPayloadRequest` object with a new optional field `executionRequestsHash`. Later this field will be populated by the hash of the requests in the beacn block body.

This PR is a preparation for the work where we need to calculate the hash of the execution requests in the beacon block body, and pass it on `engine_newPayloadV4`. In this PR, I am only updating the helper objects and the client call for `engine_newPayloadV4`.

## Fixed Issue(s)
part of #8620

## Documentation

## Changelog


StefanBratanov (approved):
LGTM. I would just remove `.getMilestone` now from `ExecutionPayload` to avoid it getting misused.

lucassaldanha:
> LGTM. I would just remove `.getMilestone` now from `ExecutionPayload` to avoid it getting misused.

Will do on the follow up PR 👍

PR #8675, opened 2024-10-03
Mark local-el retrieved blobs as seen WRT gossip equivocation cache

tbenr:
updates the equivocation data used in the gossip rule when we publish the `blob_sidecar` reconstructed via local el blob lookup.
Moreover it avoids republishing  when the recovered blob isn't from current slot
https://github.com/ethereum/consensus-specs/pull/3864

### DIFF
--- a/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/BlockBlobSidecarsTrackersPoolImplTest.java
+++ b/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/BlockBlobSidecarsTrackersPoolImplTest.java
@@ -208,7 +208,7 @@ public class BlockBlobSidecarsTrackersPoolImplTest {
   }
 
   @Test
-  public void onNewBlobSidecar_shouldMarkAsSeenAndPublishWhenOriginIsLocalEL() {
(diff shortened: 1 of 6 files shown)


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/BlockBlobSidecarsTrackersPoolImplTest.java
```diff
@@ -220,16 +225,58 @@ public void onNewBlobSidecar_shouldPublishWhenOriginIsLocalEL() {
...
 
     assertBlobSidecarsCount(3);
     assertBlobSidecarsTrackersCount(3);
 
+    verify(blobSidecarGossipValidator).markForEquivocation(blobSidecar1);
     verify(blobSidecarPublisher, times(1)).accept(blobSidecar1);
   }
 
+  @Test
+  public void onNewBlobSidecar_shouldPublishEvenWhenOriginIsLocalELButEquivocating() {
```
mehdi-aouadi: I'd rename the test to `onNewBlobSidecar_shouldPublishWhenOriginIsLocalELAndEquivocating`

mehdi-aouadi (approved):
LGTM
Just a nit

PR #8677, opened 2024-10-04
add rpc method return type checks in unit tests

### LINKED ISSUE #8672: Enforce EL RPC method types in the RPC method resolver
Currently the json RPC methods extending the abstract `AbstractEngineJsonRpcMethod` used by the `ExecutionClientHandler` might have different type parameter from the one supplied to the `EngineJsonRpcMethodsResolver::getMethod` which could lead to class cast exceptions.
Example:

```java
DummyRpcMethodClass extends AbstractEngineJsonRpcMethod<ReturnTypeClass>
```

```java
ExecutionClientHandlerImpl {
...
@Override
  public SafeFuture<BadReturnTypeClass> dummyRpcMethod(
    return engineMethodsResolver
        .getMethod(
            EngineApiMethod.ENGINE_DUMMY_RPC_METHOD,
            () -> spec.atSlot(slot).getMilestone(),
            BadReturnTypeClass.class)
        .execute(paramsBuilder.build());
  }
...
}
```

In the example above, the `dummyRpcMethod` shoudl have `ReturnTypeClass` instead of `BadReturnTypeClass`.
We could make the `EngineJsonRpcMethodsResolver` detect the RPC method return type dynamically.
Possibles approaches could be to define the return type in the method itself or make use of reflection to dynamically determine it...

mehdi-aouadi:
## PR Description
Add execution engine method return types checks.
This ensures that the return type is the expected one and not only that response future completes.

## Fixed Issue(s)

fixes #8672 

## Documentation

## Changelog


tbenr (approved):
LGTM

Issue #8680, opened 2024-10-07
Investigate failures with EL withdrawal requests sanity blocks ref tests


lucassaldanha:
The following tests are failing on reference tests `v1.5.0-alpha.7`"
- ~~sanity/blocks - basic_el_withdrawal_request~~
- ~~sanity/blocks - cl_exit_and_el_withdrawal_requests_in_same_block~~
- sanity/blocks - basic_btec_and_el_withdrawal_request_in_same_block

They are being ignored on `SanityBlocksTestExecutor`

jtraglia:
Just to double check, are you sure that `basic_btec_and_el_withdrawal_request_in_same_block` was failing?

That would be unexpected. We merged a fix for the other two though.

lucassaldanha:
> Just to double check, are you sure that `basic_btec_and_el_withdrawal_request_in_same_block` was failing?

It was, but I haven't looked at the reason yet. It could be a problem with our code :)

Issue #8618, opened 2024-09-19
[EIP-6110] Queue deposit requests and apply them during epoch processing


lucassaldanha:
Reference: https://github.com/ethereum/consensus-specs/pull/3818

Tasks:
- [x] Created PendingDeposit container
- [x] Updated presets
- [x] Epoch Processing: remove `process_pending_balance_deposits` / Add `apply_pending_deposit ` and `process_pending_deposit `
- [x] Block Processing: `process_deposit` and related changes (`get_validator_from_deposit`, `add_validator_to_registry`)
- [x] Remove `PendingPartialDeposit` containers

Issue #8617, opened 2024-09-19
[EIP-7251] Switch to compounding when consolidating with source==target


lucassaldanha:
### Description
Reference: https://github.com/ethereum/consensus-specs/pull/3918

Need to re-enable consolidation tests that were disabled:
- `OperationsTestExecutor`
- `EpochProcessingTestExecutor`

PR #8681, opened 2024-10-07
EIP-6110: Queue deposit requests and apply them during epoch processing

lucassaldanha:
## PR Description
- Updated block processing and epoch processing with changes from https://github.com/ethereum/consensus-specs/pull/3818
- Updated reference tests to v1.5.0-alpha.7
  - Had to ignore a few tests:
    - ` epoch_processing/process_consolidations` and `operations/consolidation_request` (will be re-enabled as part of https://github.com/Consensys/teku/issues/8617)
    - 3 sanity/blocks tests realted to EL withdrawals that are being investivated, and other CL teams are also having issues with (https://github.com/Consensys/teku/issues/8680)

Pending:
- Remove duplication on deposit related code between Epoch Processing and Block Processing. Better done as a separate PR to avoid adding many changes on this PR.

## Fixed Issue(s)
part of #8618 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/statetransition/epoch/EpochProcessorElectra.java
```diff
@@ -186,73 +201,176 @@ public void processEffectiveBalanceUpdates(
...
-        stateElectra.getPendingBalanceDeposits();
-    for (final PendingBalanceDeposit deposit : pendingBalanceDeposits) {
-      final Validator validator = state.getValidators().get(deposit.getIndex());
+      // Check if number of processed deposits has not reached the limit, otherwise, stop processing
+      if (nextDepositIndex
+          >= SpecConfigElectra.required(specConfig).getMaxPendingDepositsPerEpoch()) {
+        break;
+      }
+
+      // TODO-lucas What is the impact of doing this? Should we use the transition cache?
```
StefanBratanov: Hmm, yeah this for sure needs to change. Think we should use `ValidatorsUtil.getValidatorIndex()` (internally uses the cache) and then just use .get(index) on the validators in the state.

StefanBratanov (approved):
LGTM, logic wise with one comment on necessary optimization

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/statetransition/epoch/EpochProcessorElectra.java
lucassaldanha: Done!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
```diff
@@ -122,7 +122,7 @@ protected void processEpoch(final BeaconState preState, final MutableBeaconState
     processRegistryUpdates(state, validatorStatuses.getStatuses());
     processSlashings(state, validatorStatuses);
     processEth1DataReset(state);
-    processPendingBalanceDeposits(state);
+    processPendingDeposits(state);
```
tbenr, line 135: I thought about the mutating `ValidatorStatuses` and seems an additional complication to me.
At current stage I'd just approach the "problem" with verbose comments.
We could be noisy here saying that here `processPendingDeposits` may add validators in the state and `ValidatorStatuses` won't contain these new added vals.

tbenr (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
lucassaldanha: Added a big as warning in AbstractEpochProcessor :)

PR #8692, opened 2024-10-09
Migrate Attestation and AttnetsENR schemas to registry

tbenr:
Migrates `AttestationSchema` and `AttnetsENRSchema` to schema registry

## Documentation

## Changelog


zilm13 (approved):
LGTM
For future:
- maybe it worth to add something like `MilestoneSchemaProvider.java`
- as there will be hundred of providers, it's worth to organize them in packages

tbenr:
> as there will be hundred of providers, it's worth to organize them in packages

it was initially like that but than we got problems on making class\methods visibility well constrained within the package and expose only what is needed. Java has its limits here...

PR #8697, opened 2024-10-09
Check that ExecutionRequests is not null in block body

jtraglia:
## PR Description

This PR does the following:

* Add a `checkNotNull(message.getExecutionRequests(), ...)` call.
* Replace manual execution requests parsing with call to `new ExecutionRequests()`.
* Fix inconsistencies in messages (only for Electra).
* Add punctuation to a random return doc.

## Documentation

## Changelog


lucassaldanha (approved):

lucassaldanha:
Thanks @jtraglia :)

PR #8701, opened 2024-10-10
Reduce wait period to avoid flakiness

### LINKED ISSUE #8665: [Flaky] MultiPeersStandAloneVcAcceptanceTest
The test MultiPeersStandAloneVcAcceptanceTest has been presenting some flakiness for a little while on Circle CI. Needs investigation.

gfukushima:
## PR Description
This PR reduces the wait to post the Slashing event and start the second node from the second to the first epoch.
It also reduces parallelism for the ATs since last run shows some of these are peeking near 100% of CPU usage.

## Fixed Issue(s)

Fixes #8665

## Documentation

## Changelog


acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/validatorslashing/MultiPeersStandAloneVcBlocksAcceptanceTest.java
```diff
@@ -54,7 +54,7 @@ public class MultiPeersStandAloneVcBlocksAcceptanceTest
 
     firstTekuNode.start();
 
-    firstTekuNode.waitForEpochAtOrAbove(2);
+    firstTekuNode.waitForEpochAtOrAbove(1);
```
rolfyone: if we're changing this anyway, i'd change this back to 2, and i'd change the genesis wait time to 60 seconds to let both nodes be online close to genesis...

rolfyone: but i'd also be starting node 2 prior to waiting for epoch 2 i think...

gfukushima: This wouldn't work because we would have to wait until epoch 2 to start the second node in the current logic. This would only drag the test length to a longer period of waiting. What we could do is possibly change the slashing to be in a block for the epoch 4 or 5.

rolfyone: the BN will come up slow, so if the second BN starts before genesis and BOTH BN wait for the 2 epochs, then the second VC starts, by then the world order is a lot less ugly i think you'll find.

gfukushima: I still don't think we should increase the lenght of genesis from 10 to 60 and make this longer than it needs to be, test usually succeeds in 50 secs. So making it longer for no reason doesn't sound a better solution. I'm ok with starting beacon2 earlier.

acceptance-tests/src/acceptance-test/java/tech/pegasys/teku/test/acceptance/validatorslashing/MultiPeersStandAloneVcBlocksAcceptanceTest.java
```diff
@@ -39,43 +38,25 @@ public class MultiPeersStandAloneVcBlocksAcceptanceTest
...
-        slashingEventType);
-
     final TekuBeaconNode secondBeaconNode =
         createTekuBeaconNode(
             TekuNodeConfigBuilder.createBeaconNode()
                 .withGenesisTime(genesisTime)
                 .withNetwork(network)
                 .withRealNetwork()
-                .withRealNetwork()
-                .withAltairEpoch(altairEpoch)
```
tbenr, line 78: is removal of `withAltairEpoch` intentional?

gfukushima: think so, although this was added by Paul's commit, we never actually reached the epoch where Altair was set.

rolfyone: yes it was pointless to have altair at epoch 100 when we're finishing in a few epochs...

rolfyone (approved):
LGTM

PR #8702, opened 2024-10-10
Refactor ExecutionPayloadResult

StefanBratanov:
## PR Description
It is a bit inconsistent to have the `GetPayloadResponse` fields separate for the local flow but store the whole `GetPayloadResponse` in `FallbackData` for the builder flow. 

The refactor will also help to make it easier when new fields are added to `GetPayloadResponse`.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


gfukushima (approved):
lgtm, I might have missed something though. I don't see where we're doing this:

> It is a bit inconsistent to include the GetPayloadResponse in FallbackData

StefanBratanov:
> lgtm, I might have missed something though. I don't see where we're doing this:
> 
> > It is a bit inconsistent to include the GetPayloadResponse in FallbackData

The description wasn't worded properly. Updated it now. Basically, when using the builder flow and we fallback to local we store the whole `GetPayloadResponse` and that wasn't the case for the local flow. We just extracted the needed fields and set them as variables.

PR #8709, opened 2024-10-10
Fix isFullExitRequest return

jtraglia:
## PR Description

When reviewing `processWithdrawalRequests`, I noticed two differences from the spec.

* [`if is_full_exit_request: ... return`](https://github.com/ethereum/consensus-specs/blob/6bbe3aed1748a5bd3988264136b523e4ecd56740/specs/electra/beacon-chain.md?plain=1#L1476-L1480)
  * In Teku, this return statement was in the wrong spot. It would only return if we initiated the validator exit.
  * I believe this could have led to a more severe issue; maybe a consensus issue.

image

---------------

* [`if len(state.pending_partial_withdrawals) == PENDING_PARTIAL_WITHDRAWALS_LIMIT ...`](https://github.com/ethereum/consensus-specs/blob/6bbe3aed1748a5bd3988264136b523e4ecd56740/specs/electra/beacon-chain.md?plain=1#L1446)
  * In Teku, this was `>=`. Effectively the same thing & only changing to be consistent with the spec.

image

## Documentation

## Changelog


rolfyone (approved):
LGTM Thanks

PR #8715, opened 2024-10-11
Add support for nightly reference tests

jtraglia:
## PR Description

This is still a work in progress.

* Create a GitHub access token at https://github.com/settings/tokens with the following perms:
  * `repo`, `workflow`, `admin:repo_hook` (not sure these are actually necessary, but it's what I did)
  * Unfortunately, we need a token to download artifacts programmatically.
  * I would recommend using a burner GitHub account for this just in-case it gets burned.
* Export this as an environment variable under `GITHUB_TOKEN`.
  * I just added a `export GITHUB_TOKEN="ghp_..."` line to my `~/.zshrc` file.
* Run `NIGHTLY=true ./gradlew expandRefTests`

Next steps:

Add a scheduled CI test that runs at 6am UTC. This is 4 hours after the reference tests are generated. Normally, it takes about 3.5 hours for the test generators to finish.

Notes:

It would have been slightly easier to use `gh` (GitHub's CLI tool) in a CI test, but I think doing it with Gradle is the better approach for Teku. Easier to maintain and works better with the dev flow.

I tried to convert the `NIGHTLY=true` environment variable to a `--nightly` flag, but it was too difficult & not worth it the headache. I defined my own task with an option flag but the problem was calling the task from other tasks.

## Documentation

## Changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,17 +193,17 @@ jobs:
     executor: large_executor
     steps:
       - prepare
-      - run:
+      - run: &assemble
           name: Assemble
           command: |
             ./gradlew --no-daemon --parallel clean compileJava compileTestJava compileJmhJava compileIntegrationTestJava compileAcceptanceTestJava compilePropertyTestJava assemble
-      - run:
+      - run: &prep_artifacts
           name: Prep Artifacts
           command: |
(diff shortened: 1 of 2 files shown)


jtraglia:
Going to break this up into multiple PRs.

PR #8722, opened 2024-10-14
Save earliest slot block in a DB variable

### LINKED ISSUE #8658: [RocksDB] Slow getEarliestAvailableBlockSlot
While investigating issue #8447 I've found that the nodes running with RocksDB present a significant worse performance on `getEarliestAvailableBlockSlot()`

Logs for comparison purpose:

Holesky nodes:

LevelDB node:  ./teku --config-file=/etc/teku/teku.yml debug-tools db get-earliest-available-block-slot
Time to retrieve earliest available block slot in the database 183ms
1578560

RocksDB node:./teku --config-file=/etc/teku/teku.yml debug-tools db get-earliest-available-block-slot
Time to retrieve earliest available block slot in the database 25398ms
1591424

gfukushima:
## PR Description
During the investigation on the mainnet sync issue it was found that this query can take a descent amount of time (it can vary from a few ms to 2 digits seconds). This information is used when a peer send a BeaconBlocksByRangeRequestMessage ultimately causing a disconnection from the peer's side if this query times out. 
This data fetch could be optimised by storing the earliest block slot in a variable in the database.  

## Fixed Issue(s)

Fixes #8658 

## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -426,6 +448,8 @@ private void deleteFinalizedBlocks(final List<Pair<UInt64, Bytes32>> blocksToPru
       try (final FinalizedUpdater updater = finalizedUpdater()) {
         blocksToPrune.forEach(
             pair -> updater.deleteFinalizedBlock(pair.getLeft(), pair.getRight()));
+        // update earliest finalized block slot
+        updater.setEarliestBlockSlot(blocksToPrune.getLast().getLeft().plus(ONE));
```
rolfyone: so the fun bit here would be if we're in a scenario where we delete to N, but N+1 had no block... maybe N+2, N+3... depends on network quality... but i think this will then leave us in an inconsistent state which would be bad.

gfukushima: you're right, but we're only updating the slot number, wouldn't that always exists even if the block is empty?

rolfyone: so if we use this slot for anything, it'll be wrong if the block is missing... whats the use-case? usually when we're looking for earliest block we're going to start returning blocks in that range...

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -428,6 +452,16 @@ private void deleteFinalizedBlocks(final List<Pair<UInt64, Bytes32>> blocksToPru
...
+      // retrieve earliest finalized slot left after pruning and set it in the DB
+      final Optional<UInt64> earliestBlockSlot =
+          dao.getEarliestFinalizedBlock().map(SignedBeaconBlock::getSlot);
+
+      if (earliestBlockSlot.isPresent()) {
+        try (final FinalizedUpdater updater = finalizedUpdater()) {
+          updater.setEarliestBlockSlot(earliestBlockSlot.get());
+          updater.commit();
+        }
+      }
```
rolfyone: Actually this kind of screams database inconsistency, which was my concern with storing this in a database. We might need to think about this a little more. If we're storing this variable in the database at all, it needs to be maintained inside the same tx importing earlier blocks, and in the same tx pruning... otherwise we're not going to be able to trust the value at any point in time...

I think inside the context here, we just need to check before we commit the deletion... the earliest block fn will be inside the tx context at that point...

gfukushima: Agree that we should be updating that inside the same tx to ensure the integrity of the value we're storing but changes don't get effectively made in the DB until we commit them. I could move this fetch/set to inside the first   try(updater) statement but that to me feels like holding the tx from being committed unnecessarily. 

On the other hand I don't see how we would update this concurrently. My understanding is that we are always effectively executing one or the other(downloading historic blocks or pruning them), never both at the same time. Is that correct?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
gfukushima: We look use the earliest slot when we receive requests from peers to provide blocks by range, just to check if the request is within a range we have. 
https://github.com/Consensys/teku/blob/015f7bd46d07b34d5f285f5b0ba8ce63110c4ca4/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java#L169-L173

It doesn't necessarily mean we're using the earliest block at all.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
gfukushima: I have done some changes to address the issue you spotted and some others I found during the tests, thanks for the early review/feedback

rolfyone (approved):
LGTM now

PR #8725, opened 2024-10-14
8525 beacon states cleanup

### LINKED ISSUE #8525: clean up old BeaconStates.
We should probably provide at least a debug tool to cleanup the states table.

Currently if you switch from archive -> prune mode, we'll not load the archive tables so they're not accessible, but the data remains. Generally speaking, its likely that people are switching from archive mode to cleanup space, so it'd be nice to be able to at the very least clean this as a debug tool...

An alternative here would be to verify the table is empty when we start in prune or minimal mode and slowly clean it up if its not empty (oldest first), but this would be more complicated.... Maybe we should discuss the best option here...

mehdi-aouadi:
## PR Description
Clean up the finalized states after switching from ARCHIVE to MINIMAL or PRUNE mode.

## Fixed Issue(s)

fixes #8525 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,5 +9,6 @@
 ### Breaking Changes
 
 ### Additions and Improvements
+- Clean up old beacon states when switching from ARCHIVE to PRUNE or MINIMAL data storage mode
 
 ### Bug Fixes
--- a/services/chainstorage/build.gradle
+++ b/services/chainstorage/build.gradle
@@ -12,4 +12,7 @@ dependencies {
   implementation project(':infrastructure:events')
 
   implementation 'org.hyperledger.besu:plugin-api'
+
+  testImplementation testFixtures(project(':infrastructure:async'))
+  testImplementation testFixtures(project(':ethereum:execution-types'))
  }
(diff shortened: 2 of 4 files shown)


rolfyone (approved):
LGTM. It'll recover slowly this way but im not sure if we want to configure it to clean up more aggressively.... we'd easily be able to delete 1 state per minute for example, which would be x5 faster that this.

mehdi-aouadi:
> LGTM. It'll recover slowly this way but im not sure if we want to configure it to clean up more aggressively.... we'd easily be able to delete 1 state per minute for example, which would be x5 faster that this.

We have the `--Xdata-storage-state-pruning-limit` and the `--Xdata-storage-state-pruning-interval` if we want to make it more agressive...

rolfyone:
> We have the `--Xdata-storage-state-pruning-limit` and the `--Xdata-storage-state-pruning-interval` if we want to make it more agressive...

sure, but would you expect to need to set either of those options when you're just changing to minimal or prune mode ?

mehdi-aouadi:
> > We have the `--Xdata-storage-state-pruning-limit` and the `--Xdata-storage-state-pruning-interval` if we want to make it more agressive...
> 
> sure, but would you expect to need to set either of those options when you're just changing to minimal or prune mode ?

Definitely not obvious. We can configure the pruner to delete 1 state per minute (if none of the `--Xdata-storage-state-pruning-limit` and `--Xdata-storage-state-pruning-interval` is set) when switching the storage mode as you suggested then, wdyt?
Here's a PR https://github.com/Consensys/teku/pull/8734

Issue #8588, opened 2024-09-12
Redesign SchemaDefinitions classes


tbenr:

tbenr:
migration to schema registry is done

PR #8726, opened 2024-10-14
Migrate all schemas in AbstractSchemaDefinitions

tbenr:
To make provider type more more explicit, I added specific `constantProviderBuilder` and `variableProviderBuilder`

related to #8588 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/SchemaTypes.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/SchemaTypes.java
@@ -23,8 +23,10 @@ import tech.pegasys.teku.infrastructure.ssz.schema.collections.SszBitvectorSchem
 import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadHeader;
 import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadHeaderSchema;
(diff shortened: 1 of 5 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/SchemaTypes.java
```diff
@@ -33,6 +35,13 @@ public class SchemaTypes {
   // PHASE0
   public static final SchemaId<SszBitvectorSchema<SszBitvector>> ATTNETS_ENR_FIELD_SCHEMA =
       create("ATTNETS_ENR_FIELD_SCHEMA");
+  public static final SchemaId<SszBitvectorSchema<SszBitvector>> SYNCNETS_ENR_FIELD_SCHEMA =
+      create("SYNCNETS_ENR_FIELD_SCHEMA");
```
rolfyone, line 39: are we going to reference by strings elsewhere? Wonder if we should use const's or an enum or something so that we're not typing the string a bunch if we are

tbenr: Ths string are not referenced elsewhere, they are just used internally (they are exposed  in error logs to be able to identify where the error is in the schema definitions)
It is not possible to move to an enum (at least i haven't found any good alternative) since we need to capture the type here.

I attempted this trick to move the string as an enum but it doesn't not worth it IMO.
image


Btw strings are enforced to be equal to the const via unit test.

zilm13 (approved):
LGTM

PR #8728, opened 2024-10-15
Decouple block and blobs publishing\import

### LINKED ISSUE #8682: Publish block before constructing blob sidecars
In our current implementation, we wait for creating blob sidecars (inclusion proofs etc) before publishing both the block and the blob sidecars at the same time. It would make more sense to publish the block immediately before waiting for the creation of the blob sidecars. This way, nodes can take advantage of `engine_getBlobsV1` more efficiently.

Inspired by https://x.com/jimmygchen/status/1843441689622708245

tbenr:
fixes #8682

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/AbstractBlockPublisher.java
```diff
@@ -62,47 +68,54 @@ public SafeFuture<SendSignedBlockResult> sendSignedBlock(
...
     }
 
     // when broadcast validation is enabled, we need to wait for the validation to complete before
     // publishing the block (and blob sidecars)
 
     final SafeFuture<BlockImportAndBroadcastValidationResults>
         blockImportAndBroadcastValidationResults =
-            importBlockAndBlobSidecars(
-                block, blobSidecars, broadcastValidationLevel, blockPublishingPerformance);
+            importBlock(block, broadcastValidationLevel, blockPublishingPerformance);
```
StefanBratanov, line 112: I feel it requires more refactoring. Isn't block import always going to fail, because blob sidecars are not imported.

tbenr: I changed the way I push data in the tracker. Now i put them separated and the import should wait until the tracker completes.
This is what I have in mind, need to test if it actually behaves this way:)

tbenr: this is actually working.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/AbstractBlockPublisher.java
```diff
@@ -62,47 +68,54 @@ public SafeFuture<SendSignedBlockResult> sendSignedBlock(
...
 
     blockImportAndBroadcastValidationResults
         .thenCompose(BlockImportAndBroadcastValidationResults::broadcastValidationResult)
         .thenAccept(
             broadcastValidationResult -> {
               if (broadcastValidationResult == BroadcastValidationResult.SUCCESS) {
-                publishBlockAndBlobSidecars(block, blobSidecars, blockPublishingPerformance);
+                publishBlock(block, blockPublishingPerformance)
+                    .always(
```
StefanBratanov: Actually, why do we wait for block to finish publishing again and not just send both across async?

tbenr: that's a strategy that has been discussed in several places. The reason are essentially help solo stakers with low up bandwidth to reduce latency in block propagation. With the adoption of `engine_getBlobs` blobs can be retrieved once the block is known, thus it is good to give priority to the block over the blobs.

I'll put this under a feature flag.

StefanBratanov (approved):
LGTM

PR #8730, opened 2024-10-15
Fix missing Gossip subnet subscription on startup

### LINKED ISSUE #166: Not subscribing to Gossip subnets occasionally
Sometime a client for some reasons doesn't subscribe to subnets (including columns, attestations, etc).
image

Nashatyrev:
## PR Description

`GossipForkManager` has two different sources of `isHeadOptimistic`flag:
- `RecentChainData.isChainHeadOptimistic()` when starting up
- `onOptimisticHeadChanged()` to update the flag

There are possible cases (e.g. https://github.com/Nashatyrev/teku/issues/166) when the value differs in those two sources. Thus subnet subscriptions (which are to be activated when `isHeadOptimistic` becomes `false`) may not be activated.

This PR initializes the flag with `RecentChainData.isChainHeadOptimistic()` value and then relies on `onOptimisticHeadChanged()` notifications only

## Fixed Issue(s)

Fix https://github.com/Nashatyrev/teku/issues/166

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8736, opened 2024-10-16
Migrate attester slashing and indexed attestation to schema registry

tbenr:
Rolls back to non-versioned schema and type (nice simplification) for `AttesterSlashing` and `IndexedAttestation`
Starts using the registry inside block schema to lookup these schemas instead of recreating them.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #8750, opened 2024-10-18
Info log when starting state regeneration

gfukushima:
## PR Description
Change the initial log to info level since teku might hang for a little while regenerating states with no information on what's happening.

## Documentation

## Changelog


rolfyone:
I'd prefer we don't do this, it should never happen and if we want to see it we can turn on debug mode...

PR #8758, opened 2024-10-21
made post block header work with available information

rolfyone:
In the case of JSON specifically, we do have the slot number, and we don't actually need the header with a slot present.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,4 +12,5 @@
 - Clean up old beacon states when switching from ARCHIVE to PRUNE or MINIMAL data storage mode
 
 ### Bug Fixes
- - Fixed a block production issue for Validator Client (24.10.0 to 24.10.2 teku VC), where required headers were not provided for JSON payloads. Default SSZ block production was unaffected. The required header is in the beacon-api spec but was not updated in all places for the VC.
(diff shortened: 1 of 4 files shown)


mehdi-aouadi:
Question here: following this reasoning, should all the APIs that get the slot within the json don’t be strict with the `Eth-Consensus-Version` header and fallback to the slot?
IMO since the header is required, the request should fail when it’s missing. Plus, [according to the spec](https://github.com/ethereum/beacon-APIs/blob/2e7c4aabcf60f5ae6d414866c268aff350a70e81/beacon-node-oapi.yaml#L466), we should use the `Eth-Consensus-Version` header to deserailise/serialise the data and not the slot.
If we go in this direction, other APIs relying merely on the header should maybe fallback to the slot when available too

rolfyone:
> Question here: following this reasoning, should all the APIs that get the slot within the json don’t be strict with the `Eth-Consensus-Version` header and fallback to the slot?

potentially, but more importantly blocks are the one we've actually made a compatibility issue for. I think given there's been no issues with the others, that we just make this change here and not everywhere, because otherwise we're really just increasing the opportunities for new issues.

mehdi-aouadi (approved):
LGTM

PR #8766, opened 2024-10-22
Change the builderResultCache of ExecutionLayerBlockProductionManagerImpl

### LINKED ISSUE #8625: [SSV] Fix cache behaviour when requesting blocks in parallel
There are scenarios where multiple operators may be connected to the same Beacon node,each of which requests the Beacon block proposal, in case they become the leader and must propose it on SSV. We have some cases in code, which cache data by slot for block proposal (with the usual assumption only one block proposal is requested per slot) so in case of two responses (one builder response and one fallback), there could be issues. On first look, the only two cases are in `ExecutionLayerBlockProductionManagerImpl` and in `FailoverValidatorApiHandler`.
 
More background on Discord and an issue occuring: https://discord.com/channels/697535391594446898/697539289042649190/1286291762167545958

david-ry4n:
## PR Description
Change the builderResultCache of ExecutionLayerBlockProductionManagerImpl keys from slot to SlotAndBlockRoot. This allows the cache to return proposals based on slot and block root, which is required if there's multiple operators requesting different beacon block proposals. Using slotAndBlockRoot to allow the onSlot function to continue to clear key/value pairs from prior slots.

fixes #8625 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryDenebTest.java
+++ b/beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryDenebTest.java
@@ -156,7 +156,7 @@ public class BlockFactoryDenebTest extends AbstractBlockFactoryTest {
     final SignedBlockContainer block = blockAndBlobSidecars.block();
     final List<BlobSidecar> blobSidecars = blockAndBlobSidecars.blobSidecars();
 
-    verify(executionLayer).getCachedUnblindedPayload(block.getSlot());
+    verify(executionLayer).getCachedUnblindedPayload(block.getSignedBlock().getSlotAndBlockRoot());
 
     final SszList<SszKZGCommitment> expectedCommitments =
         block.getSignedBlock().getMessage().getBody().getOptionalBlobKzgCommitments().orElseThrow();
--- a/beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryDenebTest.java
+++ b/beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryDenebTest.java
@@ -156,7 +156,7 @@ public class BlockFactoryDenebTest extends AbstractBlockFactoryTest {
     final SignedBlockContainer block = blockAndBlobSidecars.block();
     final List<BlobSidecar> blobSidecars = blockAndBlobSidecars.blobSidecars();
 
-    verify(executionLayer).getCachedUnblindedPayload(block.getSignedBlock().getSlotAndBlockRoot());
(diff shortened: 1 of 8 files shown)


beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryDenebTest.java
```diff
@@ -156,7 +156,7 @@ void shouldCreateValidBlobSidecarsForBlindedBlock() {
     final SignedBlockContainer block = blockAndBlobSidecars.block();
     final List<BlobSidecar> blobSidecars = blockAndBlobSidecars.blobSidecars();
 
-    verify(executionLayer).getCachedUnblindedPayload(block.getSlot());
+    verify(executionLayer).getCachedUnblindedPayload(block.getSignedBlock().getSlotAndBlockRoot());
```
StefanBratanov: We can just do `block.getSlotAndBlockRoot()`. It is the same thing under the hood.

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
```diff
@@ -55,9 +57,9 @@ public void onSlot(final UInt64 slot) {
     executionResultCache
         .headMap(slot.minusMinZero(EXECUTION_RESULT_CACHE_RETENTION_SLOTS), false)
         .clear();
-    builderResultCache
-        .headMap(slot.minusMinZero(BUILDER_RESULT_CACHE_RETENTION_SLOTS), false)
-        .clear();
+    SlotAndBlockRoot toSlotAndBlockRoot =
```
StefanBratanov: I feel more comfortable looping through the cache and filtering the ones who match the condition with `removeIf`. Otherwise, we rely `compare` to be implemented correctly in the `SlotAndBlockRoot` object.

StefanBratanov (approved):
LGTM with couple of nits. There are few other changes which I observed that needs to be changed. Can discuss tomorrow. For the moment, can change `fixes` to `related to`, since there would be need of subsequent PRs.

beacon/validator/src/test/java/tech/pegasys/teku/validator/coordinator/BlockFactoryDenebTest.java
david-ry4n: Turns out SignedBlockContainer didn't have getSlotAndblockRoot, so I added it.

StefanBratanov (approved):
LGTM

PR #8773, opened 2024-10-23
Produce block only once for a slot

### LINKED ISSUE #8625: [SSV] Fix cache behaviour when requesting blocks in parallel

StefanBratanov:
## PR Description
Essentially, return the same block if the createBlock endpoint is called twice. This means we no longer require the changes in #8766 

## Fixed Issue(s)
fixes #8625

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
@@ -122,7 +122,10 @@ public class ValidatorApiHandler implements ValidatorApiChannel {
    */
   private static final int DUTY_EPOCH_TOLERANCE = 1;
 
-  private final Map<UInt64, BlockContainerAndMetaData> createdBlocksBySlotCache =
+  private final Map<UInt64, SafeFuture<Optional<BlockContainerAndMetaData>>>
+      blockProductionFuturesBySlotCache = LimitedMap.createSynchronizedLRU(2);
+
+  private final Map<UInt64, Bytes32> createdBlockRootsBySlotCache =
       LimitedMap.createSynchronizedLRU(2);
 
   private final BlockProductionAndPublishingPerformanceFactory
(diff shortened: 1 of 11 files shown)


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/ExecutionLayerBlockProductionManagerImpl.java
```diff
@@ -72,35 +67,42 @@ public ExecutionPayloadResult initiateBlockProduction(
...
+        blockSlotState.getSlot(),
+        __ -> {
+          if (attemptBuilderFlow) {
+            return executeBuilderFlow(
+                context, blockSlotState, requestedBuilderBoostFactor, blockProductionPerformance);
+          } else {
+            return executeLocalFlow(context, blockSlotState, blockProductionPerformance);
+          }
+        });
+  }
```
tbenr: this is not guaranteed to be computed once. I think we need to synchronize.

StefanBratanov: Modified PR, so no longer applicable

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -122,6 +122,9 @@ public class ValidatorApiHandler implements ValidatorApiChannel {
    */
   private static final int DUTY_EPOCH_TOLERANCE = 1;
 
+  private final Map<UInt64, SafeFuture<Optional<BlockContainerAndMetaData>>>
+      blockProductionFuturesBySlotCache = LimitedMap.createSynchronizedLRU(2);
+
   private final Map<UInt64, Bytes32> createdBlockRootsBySlotCache =
       LimitedMap.createSynchronizedLRU(2);
```
tbenr: I think we can get the rid of `createdBlockRootsBySlotCache` and use only the new `blockProductionFuturesBySlotCache`.

StefanBratanov: Yup, makes sense. Done.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -866,11 +883,13 @@ private List<ProposerDuty> getProposalSlotsForEpoch(final BeaconState state, fin
...
-    final Bytes32 locallyCreatedBlockRoot =
-        createdBlockRootsBySlotCache.get(blockContainer.getSlot());
-    return Objects.equals(blockRoot, locallyCreatedBlockRoot);
+  private boolean isLocallyCreatedBlock(final UInt64 slot) {
+    final SafeFuture<Optional<BlockContainerAndMetaData>> localBlockProduction =
+        localBlockProductionBySlotCache.get(slot);
+    if (localBlockProduction == null) {
+      return false;
+    }
+    return localBlockProduction.isCompletedNormally();
```
tbenr: I'd keep the blockRoot check here.

like 

```java
if (localBlockProduction == null) {
  return false;
}

if(! localBlockProduction.isCompletedNormally()) {
return false;
}
return localBlockProduction.join() .... .equals(.....));
```

tbenr (approved):
LGTM

PR #8775, opened 2024-10-23
Add BlobSidecar inclusion proof verification in req/resp

zilm13:
## PR Description
Check should be there according to the spec https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md?plain=1#L343

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8776, opened 2024-10-23
Tiny `getTopic` optimization

tbenr:
Reduces `getTopic` calculations by caching topic name where isn't supposed to change.

## Documentation

## Changelog


mehdi-aouadi:
As discussed, I'd rather initialise the topic in the constructor once and for all since all `gossipEncoding`, `forkDigest` and the `topicName` are all final and will never change.
That would break the `Eth2TopicHandlerTest` but it would be better to refactor the test instead. Otherwise we will end up with a memoizer not really memoizing but simply for not breaking the test...

tbenr:
> That would break the Eth2TopicHandlerTest but it would be better to refactor the test instead. Otherwise we will end up with a memoizer not really memoizing but simply for not breaking the test...

Just change it without breaking the test, avoiding using getters :)

mehdi-aouadi (approved):
LGTM

PR #8782, opened 2024-10-24
Transition finalized state until the first slot after the finalized epoch

### LINKED ISSUE #7955: Finalized state from `debug/state` API is incorrect on epochs with empty slots in the end

gfukushima:
## PR Description
This PR aims to return the finalized state applying processing empty slot to make the transtion of the epoch fixing the case when we return a different state when the first slot of the new epoch is an empty block.
Missing some unit tests but tests with kurtosis have shown good results.

## Fixed Issue(s)

Fixes #7955

## Documentation

## Changelog

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
@@ -16,6 +16,8 @@ package tech.pegasys.teku.api.stateselector;
 import static tech.pegasys.teku.spec.config.SpecConfig.GENESIS_SLOT;
 
 import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.api.AbstractSelectorFactory;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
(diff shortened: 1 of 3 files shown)


data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
```diff
@@ -79,20 +85,45 @@ public StateSelector genesisSelector() {
...
-                            finalized.getState(),
-                            // The finalized checkpoint may change because of optimistically
-                            // imported blocks at the head and if the head isn't optimistic, the
-                            // finalized block can't be optimistic.
-                            client.isChainHeadOptimistic(),
-                            true,
-                            true)));
+                    finalized -> {
+                      final UInt64 slot = finalized.getSlot();
+                      final Spec spec = finalized.getSpec();
```
zilm13, line 96: we have spec already at `StateSelectorFactory`

data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
```diff
@@ -79,20 +85,45 @@ public StateSelector genesisSelector() {
...
+                      final BeaconState finalizedState;
+                      //if this isn't the first slot after the finalized epoch than
+                      // we'll compute empty slots until we get there
+                      if (!slot.dividedBy(spec.getSlotsPerEpoch(slot)).isZero()) {
```
zilm13, line 100: I'd avoid checks like this especially we consider changes in slots time or maybe slots per epoch in the future. Use spec helpers like `computeStartSlotAtEpoch` and compare actual and expected slot

zilm13 (commented):
I think it's a hacky way to go. The that I'd prefer is to have transitioned anchor in Store and just return it here as is. Because If we start with such anchor, it's what we returns if someone queries our node via beacon API

data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
gfukushima: how did I miss that?

data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
gfukushima: yeah, that is much better, thanks for the heads up.

gfukushima:
Closing in favour of #8985

PR #8785, opened 2024-10-24
Add get*Schema functions to PendingConsolidation

jtraglia:
## PR Description

While these are not used, this class is the only one without these functions.

## Documentation

## Changelog


tbenr (commented):
Not sure here. In principle we add things when they are useful, not because they look nice or symmetric to others.

jtraglia:
That's fair. I'll close this.

PR #8789, opened 2024-10-25
Refactor BeaconStateSchema.getUniqueFields() methods

jtraglia:
## PR Description

Instead of starting with Altair fields and re-adding everything since then, I think it would be better to inherit the fields from the previous upgrade then, if necessary, replace updated fields.

## Documentation

## Changelog


tbenr:
> Instead of starting with Altair fields and re-adding everything since then, I think it would be better to inherit the fields from the previous upgrade then, if necessary, replace updated fields.

Yes I agree with you. there has been a discussion in the past over this topic, I should go back and find my proposal.
But essentially what we need is a way of overriding but also deprecating fields coming from the prior fork.
I'll look into your approach probably on monday :)

tbenr:
And thanks for your contribution, as usual!!

jtraglia:
> There has been a discussion in the past over this topic, I should go back and find my proposal.

Sounds good! If it's similar enough to your proposal, feel free to make whatever changes you want to this PR. Otherwise, I'm okay with closing this one and going with a different approach.

tbenr (approved):
LGTM.
I found my old proposal back when we were defining Deneb. It is essentially very similar to this one.

PR #8820, opened 2024-11-05
run `Forkchoice::onBlock` async in block importer

tbenr:
When importing a block, we end up having the state transition running synchronously due to how `Forkchoice::onBlock` is written. In this way we run it using an async runner, which then enables a true async block import which have downstream effect on block publishing:

block publishing now starts just after the EQUIVOCATION check, without synchronously waiting the state transition
```
2024-11-05 18:41:13.224	
Slow Block Publishing *** Slot: 2910506 start 887ms, block_publishing_initiated +3ms, blob_sidecars_publishing_initiated +21ms, blob_sidecars_imported +7ms, block_import_completed +283ms, complete +1ms total: 315ms
```

for reference this is the old flow:
```
2024-11-05 18:42:37.393	
Slow Block Publishing *** Slot: 2910513 start 853ms, blob_sidecars_imported +26ms, block_publishing_initiated +291ms, blob_sidecars_publishing_initiated +0ms, block_import_completed +44ms, complete +0ms total: 361ms
```

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

PR #8829, opened 2024-11-08
Simplify schema provider definition

tbenr:
Removes the need to specify `constant` or `variable` schema provider when defining them via builder.
With this PR the non-changing instance across multiple forks is automatically guaranteed via an equality check.

This approach provide a nice benefit: you don't need to care about if the schema is changing or not. If the shema is an "high order" schema that looks up in the registry to pick the schemas for the fields, you don't know if those fields are changing across forks. This way we can be sure that we will always keep the new one whenever the new schema is semantically different from the previous one.

There are some edge cases in which the equality check may not do what we want (maybe?) so I provided a way to disable it.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/BaseSchemaProvider.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/BaseSchemaProvider.java
@@ -29,15 +29,15 @@ class BaseSchemaProvider<T> implements SchemaProvider<T> {
   private final TreeMap<SpecMilestone, SchemaProviderCreator<T>> milestoneToSchemaCreator =
       new TreeMap<>();
   private final SchemaId<T> schemaId;
-  private final boolean disableSchemaEqualityCheck;
+  private final boolean alwaysCreateNewSchema;
 
   private BaseSchemaProvider(
       final SchemaId<T> schemaId,
       final List<SchemaProviderCreator<T>> schemaProviderCreators,
       final SpecMilestone untilMilestone,
-      final boolean disableSchemaEqualityCheck) {
+      final boolean alwaysCreateNewSchema) {
     this.schemaId = schemaId;
-    this.disableSchemaEqualityCheck = disableSchemaEqualityCheck;
+    this.alwaysCreateNewSchema = alwaysCreateNewSchema;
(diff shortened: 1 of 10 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/BaseSchemaProvider.java
```diff
@@ -90,100 +96,104 @@ public Set<SpecMilestone> getSupportedMilestones() {
...
-   * Creates a builder for a constant schema provider.<br>
-   * This can be used when schema remains the same across multiple milestones. <br>
+   * Creates a builder for a schema provider.<br>
    * Example usage:
    *
    * <pre>{@code
-   * constantProviderBuilder(EXAMPLE_SCHEMA)
-   *    .withCreator(ALTAIR, (registry, config) -> new ExampleSchemaAltair())
-   *    .withCreator(ELECTRA, (registry, config) -> new ExampleSchemaElectra())
+   * variableProviderBuilder(EXAMPLE_SCHEMA)
```
zilm13: just `providerBuilder` here and there

ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/BaseSchemaProvider.java
```diff
@@ -90,100 +96,104 @@ public Set<SpecMilestone> getSupportedMilestones() {
...
+   *    .withCreator(BELLATRIX, (registry, config) -> ExampleSchema1.create(registry, config))
+   *    .withCreator(ELECTRA, (registry, config) -> ExampleSchema2.create(registry, config))
    *    .build();
    *
    * }</pre>
    *
-   * this will create a schema provider that will generate: <br>
-   * - a new instance of ExampleSchema1 for each milestone from ALTAIR to ELECTRA <br>
-   * - a new instance of ExampleSchema2 for each milestone from ELECTRA to last known milestone <br>
+   * - disabling the schema equality check by calling {@link Builder#disableSchemaEqualityCheck()}
```
zilm13: it's not clear for me what disableSchemaEqualityCheck does. Maybe we renaming/changing the doc towards outcome of its usage rather than logic of builder change could make sense?

zilm13 (approved):
LGTM just few nits

PR #8833, opened 2024-11-16
Migrate beacon state

tbenr:
We don't need to do the `LastExecutionPayloadHeader` field update, it is transparently handled by the registry.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8834, opened 2024-11-18
Provide schema name to creator

tbenr:
Let provider prepare the schema name and pass it to the creator. This avoids error in copy\pasting the schemaId when we want to use the generated schemaName.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

Issue #8803, opened 2024-10-31
p2p: Deprecate TTFB, RESP_TIMEOUT, introduce rate limiting


tbenr:
https://github.com/ethereum/consensus-specs/pull/3767

this retro-compatible change and is not strictly tight to Electra fork

Haven't implemented yet:

- [ ] The responder MAY penalise peers that concurrently open more than `MAX_CONCURRENT_REQUESTS` streams for the same request type, for the protocol IDs defined in this specification.
- [ ] The responder MAY rate-limit chunks by withholding each chunk until capacity is available.

rolfyone:
@StefanBratanov @tbenr can this be closed?

StefanBratanov:
@rolfyone I left it open because of the MAYs that we haven't implemented

PR #8839, opened 2024-11-20
Deprecate TTFB, RESP_TIMEOUT, introduce MAX_CONCURRENT_REQUESTS

StefanBratanov:
## PR Description
Main changes are in `Eth2IncomingRequestHandler`, `Eth2OutgoingRequestHandler` and `RpcHandler`

Reintroduced the timeouts as constants to handle slow servers. Furthermore, they could be should probably be configured per different methods. (as per the consensus-specs PR describes)

## Fixed Issue(s)
related to #8803 

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/core/Eth2OutgoingRequestHandler.java
```diff
@@ -135,16 +110,10 @@ public void processData(final NodeId nodeId, final RpcStream rpcStream, final By
       for (TResponse maybeResponse : maybeResponses) {
         getResponseProcessor(rpcStream).processResponse(maybeResponse);
       }
-      if (chunksReceived < maximumResponseChunks) {
-        if (!maybeResponses.isEmpty()) {
-          ensureNextResponseArrivesInTime(rpcStream, chunksReceived, currentChunkCount);
-        }
-      } else {
+      if (chunksReceived == maximumResponseChunks) {
```
tbenr: 🤔 ... `>=`  here? I don't know if it can happen but...

tbenr: we could also merge those conditions?

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/rpc/RpcHandler.java
```diff
@@ -59,14 +62,20 @@ public class RpcHandler<
...
 
   public RpcHandler(
       final AsyncRunner asyncRunner,
-      final RpcMethod<TOutgoingHandler, TRequest, TRespHandler> rpcMethod) {
+      final RpcMethod<TOutgoingHandler, TRequest, TRespHandler> rpcMethod,
+      final int maxConcurrentRequests,
+      final Duration concurrentRequestsQueueTimeout) {
     this.asyncRunner = asyncRunner;
     this.rpcMethod = rpcMethod;
+    concurrentRequestsSemaphore = new Semaphore(maxConcurrentRequests);
```
tbenr: 🤔 ... this is a blocking approach, with a tight timeout... should we have an actual async queue here?

StefanBratanov: Yes, I realised that, actually I am working on a different idea right now. Should be done soon.

StefanBratanov: Changed the approach to use `ThrottlingTaskQueue` which makes much more sense

Nashatyrev:
Shouldn't we remove the obsolete constants from `NetworkingSpecConfig`? 
```java
  // in seconds
  int getTtfbTimeout();

  // in seconds
  int getRespTimeout();
```

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/rpc/RpcHandler.java
```diff
@@ -56,17 +57,21 @@ public class RpcHandler<
...
+  private final ThrottlingTaskQueue concurrentRequestsQueue;
 
   public RpcHandler(
       final AsyncRunner asyncRunner,
-      final RpcMethod<TOutgoingHandler, TRequest, TRespHandler> rpcMethod) {
+      final RpcMethod<TOutgoingHandler, TRequest, TRespHandler> rpcMethod,
+      final int maxConcurrentRequests) {
     this.asyncRunner = asyncRunner;
     this.rpcMethod = rpcMethod;
+    concurrentRequestsQueue = ThrottlingTaskQueue.create(maxConcurrentRequests);
```
Nashatyrev, line 74: As far as I understand here we are limiting ourselves for max concurrent outgoing requests? Do we have the code which enforces inbound limit?

StefanBratanov: Do you mean `The responder MAY penalise peers that concurrently open more than `MAX_CONCURRENT_REQUESTS` streams for the same request type, for the protocol IDs defined in this specification.` ? I was thinking it could be an incremental change and do it in a further PR since it is MAY instead of SHOULD/MUST.

Nashatyrev: Oh, right! Fair point!

StefanBratanov:
> Shouldn't we remove the obsolete constants from `NetworkingSpecConfig`?
> 
> ```java
>   // in seconds
>   int getTtfbTimeout();
> 
>   // in seconds
>   int getRespTimeout();
> ```

I tried doing this and there were some test failures regarding unrecognised constants. Think should look more into it.

Nashatyrev (approved):
LGTM

PR #8840, opened 2024-11-20
Simplify blob sidecar availability checker

### LINKED ISSUE #8740: Simplify blobSidecar availability checker
Assuming the following is true:
validating `kzg commitments` and `inclusion proof` guarantees that blobs and commitments are valid and that the commitments are part of the block.

Thus, if we always do those checks before sending the blobSidecar in the pool (`BlockBlobSidecarsTrackersPoolImpl`), then we can remove all the complex, two-step, checks we do on in `ForkChoiceBlobSidecarsAvailabilityChecker` and the only remaining logic is the Tracker signalling when the blobs are completed.

sources are:
- Syncing code
- RPC requests
- inbound GOSSIP
- local EL

If we trust local EL, and we apply the validation over all the sources, than it means this is feasable.

If we want to be paranoid, we could have a `BlobSidecar` to be an `interface`, the actual `BlobSidecar` will be `BlobSidecarImpl` and then we can have a `ValidatedBlobSidecar` type. The validation process will accept `BlobSidecarImpl` and produce `ValidatedBlobSidecar`. Then in the `BlockBlobSidecarsTrackersPoolImpl` we can only allow validated blobs to go in.

NOTE:
The availability checker has being designed when `BlobSidecar` wasn't having the inclusion proof. So we were forced to "postpone" the validation to the moment in which the block become available.

tbenr:
blobs get from EL are marked as valid (trust assumption on EL)

fixes #8740

## Documentation

## Changelog


eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
```diff
@@ -63,7 +63,7 @@ public class ManualReferenceTestRunner extends Eth2ReferenceTestCase {
    *
    * <p>May be overridden by the ENV_DISPLAY_NAME environment variable.
    */
-  private static final String DISPLAY_NAME = "";
+  private static final String DISPLAY_NAME = "invalid_incorrect_proof";
```
mehdi-aouadi: I guess this should be removed

tbenr: yes i have that rolled back locally

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/BlobSidecar.java
```diff
@@ -39,6 +39,9 @@ public class BlobSidecar
         SignedBeaconBlockHeader,
         SszBytes32Vector> {
 
+  private volatile boolean kzgAndInclusionProofValidated = false;
+  private volatile boolean signatureValidated = false;
```
mehdi-aouadi, line 44: I find these validation steps a bit confusing and I'm wondering if we can't simply mark the blob sidecar as valid or not (merge them) or split the `kzgAndInclusionProofValidated` into two validation flags: `kzgValidated` and `inclusionProofValidated`. An `and` in a boolean variable name is red flag to me. Or do we really need this design for some reason I'm missing?!

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -53,333 +33,93 @@
...
-          && !kzgCommitmentsInBlock.isEmpty()
-          && isBlockOutsideDataAvailabilityWindow()) {
-        // there are no available blobs so far, but we are outside the availability window. We can
-        // skip additional checks
-        return Optional.of(BlobSidecarsAndValidationResult.NOT_REQUIRED);
+      if (!blobSidecar.isSignatureValidated()
+          && !blobSidecar
+              .getSignedBeaconBlockHeader()
+              .hashTreeRoot()
+              .equals(block.hashTreeRoot())) {
```
mehdi-aouadi: Isn't this the same check done twice?!

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/BlobSidecar.java
tbenr: the reason I put kzg and inclusion proof together and signatureValidation as a separate boolean is because we can perform kzg and inclusion proof validation just having the blob itself (you don't need anything else). For the signature validation you need a prevalidated signed block OR the state to get the pubkey from and do the signature check.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
tbenr: same check as which one?

mehdi-aouadi: I meant same as `blobSidecar.isSignatureValidated()`

tbenr: ok but where am I doing this the second time?
I'm not sure this always checked. For instance the RPC lookups check blobs via BlobSidecarsByRootValidator.validate, which can only do kzg and inclusion proof.
So if we end up processing blobs coming from RPC source, we are not checking the block header in the blobSidecar (correct me if i'm wrong :) )

tbenr: oh. I actually forgot to do the marking in `BlobSidecarsByRangeListenerValidatingProxy`

tbenr: mmm maybe I can move those marking deeper in the stack and have them separated as you suggested

tbenr:
since i forgot to do this marking i think the way I did it is a bit error prone. I'll change it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
mehdi-aouadi, line 104: > So if we end up processing blobs coming from RPC source, we are not checking the block header in the blobSidecar (correct me if i'm wrong :) )

Ok I see what you mean. One should keep in mind the whole flow to understand that...

mehdi-aouadi: > So if we end up processing blobs coming from RPC source, we are not checking the block header in the blobSidecar (correct me if i'm wrong :) )

Ok I see what you mean. One should keep in mind the whole flow to understand that...

zilm13:
I will join Mehdi feedback:
- if some validation was not performed I expect that we want to run it now
- if it means that it was actually performed but failed it's confusing
- I'm not sure on extending base `BlobSidecar` class, maybe better make some wrapper? It would be cleaner.
- I'm happy with split kzg+inclusion / signature

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
```diff
@@ -402,20 +400,30 @@ private void validateBlobSidecars(final SignedBeaconBlock block) {
...
+        blobSidecars.stream()
+            .allMatch(
+                blobSidecar -> {
+                  final boolean signedHeaderMatchesBlock =
+                      blobSidecar
+                          .getSignedBeaconBlockHeader()
+                          .hashTreeRoot()
+                          .equals(block.hashTreeRoot());
+
+                  if (signedHeaderMatchesBlock) {
```
zilm13: it's a bit dirty, "side-effect", I'd just mark it in separate cycle after throw. Or even better mark it first and then boolean = allmatch(isSignatureValidated).

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
```diff
@@ -402,20 +400,30 @@ private void validateBlobSidecars(final SignedBeaconBlock block) {
     final List<BlobSidecar> blobSidecars =
```
zilm13, line 402: Maybe `validateBlobSidecarsBlockSignature` as it's the only thing we are checking here? The reason why I'm so meticulous here is because we don't even check we have all required blobSidecars here (I guess it's done in other place)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/BlobSidecar.java
```diff
@@ -139,6 +143,30 @@ public String toLogString() {
...
+  public boolean isKzgCommitmentInclusionProofValidated() {
+    return kzgCommitmentInclusionProofValidated;
+  }
+
+  public boolean isSignatureValidated() {
+    return signatureValidated;
+  }
+
+  public void markKzgAsValidated() {
+    kzgValidated = true;
```
zilm13, line 159: Still some uncertainty in this approach. Original container is immutable, this structure is mutable for these fields. Plus we add fields which are not used in equals, changes purpose etc. So (I'm not sure) I'd think on 
1) wrapper/keep as is 
2) immutable/mutable

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
```diff
@@ -84,10 +81,20 @@ public MiscHelpersDeneb(
...
+        kzg.verifyBlobKzgProof(
+            blobSidecar.getBlob().getBytes(),
+            blobSidecar.getKZGCommitment(),
+            blobSidecar.getKZGProof());
+
+    if (result) {
+      blobSidecar.markKzgAsValidated();
+    }
+
+    return result;
```
zilm13, line 100: For this and further check implementations. If check was failed we have a chance to run it twice (if called again). So may worth separate checkIsPerformed, checkIsPassed flags.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -104,282 +72,64 @@ public ForkChoiceBlobSidecarsAvailabilityChecker(
...
-          blobSidecarsToValidate.size(),
-          kzgCommitmentsInBlock.size());
-
-      if (blobSidecarsToValidate.isEmpty()
-          && !kzgCommitmentsInBlock.isEmpty()
-          && isBlockOutsideDataAvailabilityWindow()) {
-        // there are no available blobs so far, but we are outside the availability window. We can
-        // skip additional checks
-        return Optional.of(BlobSidecarsAndValidationResult.NOT_REQUIRED);
+    for (final BlobSidecar blobSidecar : blobSidecars) {
```
zilm13: Could we put this check in some helper and reuse here and in `HistoricalBatchFetcher.java`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
```diff
@@ -104,282 +72,64 @@ public ForkChoiceBlobSidecarsAvailabilityChecker(
...
-
-    return BlobSidecarsAndValidationResult.NOT_AVAILABLE;
-  }
-
-  private BlobSidecarsAndValidationResult validateBatch(final List<BlobSidecar> blobSidecars) {
-    final BeaconBlock block = blockBlobSidecarsTracker.getBlock().orElseThrow().getMessage();
-    final SlotAndBlockRoot slotAndBlockRoot = blockBlobSidecarsTracker.getSlotAndBlockRoot();
-
-    final MiscHelpers miscHelpers = spec.atSlot(slotAndBlockRoot.getSlot()).miscHelpers();
+  private BlobSidecarsAndValidationResult validateCompletedBlobSidecars() {
```
zilm13, line 97: where do we check that we have all blobSidecars for the number of commitments, I miss it?

zilm13 (commented):
Have not checked tests, yet (starting)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/historical/HistoricalBatchFetcher.java
tbenr: I think you're right here. We rely on availability checker to verify the completeness. Need to think about if we want to continue exposing the check from the manager or do it here if it is the only place we need it.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/BlobSidecar.java
tbenr: In my idea, due to the fact that the container is immutable, than we can safely have these additional flags.
So it is intentional that this fields are not part of the equality and the object must behave as these flags are not here at all. We can discuss it

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
tbenr: This is true but I just wanted to keep it simple.
An option would be to have the flags as Optional<Boolean>.

But at the end I don't think it is really relevant because we discard the invalid blobs immediately (ie via gossip\rpc) so I don't think that we ever risk to validate an invalid one twice.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
tbenr: yeh I was thinking that I've been too aggressive by removing the completeness check (which is also needed for historical sync). It's a cheap check and we should not rely on blobs tracker which rely only on blob indices to determine completeness. In general it is better that this class takes all it's responsibilities.

tbenr: well.. checked again and in theory what `BlockBlobSidecarsTracker` is doing should be good enough. 🤔

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/deneb/helpers/MiscHelpersDeneb.java
zilm13: Didn't think about discard, looks reasonable, but I'd add some info somewhere for clarity. I don't know, javadoc to isKzgValidated maybe.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java
zilm13: ok, I see, so it will fail by timeout. I'd add some comment for transparency.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blobs/versions/deneb/BlobSidecar.java
zilm13: Ok, just to check you are confident on it. Will not oppose.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlobSidecarManagerImpl.java
```diff
@@ -140,15 +165,35 @@ public BlobSidecarsAndValidationResult createAvailabilityCheckerAndValidateImmed
...
+
+    forkChoiceBlobSidecarsAvailabilityChecker.initiateDataAvailabilityCheck();
+
+    final SafeFuture<BlobSidecarsAndValidationResult> availabilityCheckResult =
+        forkChoiceBlobSidecarsAvailabilityChecker.getAvailabilityCheckResult();
+
+    if (availabilityCheckResult.isDone()) {
+      return availabilityCheckResult.join();
+    } else {
+      throw new IllegalStateException("Availability check should be done immediately");
```
zilm13: Maybe change the message, it's a bit confusing

zilm13 (approved):
Great work!
LGTM
Worth run on some test nodes before merging into master.

tbenr:
> Worth run on some test nodes before merging into master.

I run an historical sync test (mainnet) , I'll run more tests (mainnet and local)

tbenr:
I actually forgot this:
image

discovered during tests :)

tbenr:
@zilm13 take a look to latest commit. Testing session was fine, historical sync, forward symc, chain follow.

zilm13:
@tbenr loooks good, the only nit, it maybe better to mark it as validated in `if`, by to your decision.

PR #8848, opened 2024-11-26
[Mekong] Fix electra new validator effective balance logic

lucassaldanha:
In the event of multiple deposits to the same validator being processed on a single epoch transition, we get into a scenario where a validator's effective balance has to be updated as part of the epoch processing. However, due to how our validator caches works, we were missing those updates. We are temporarily recalculating the cache mid-epoch transition, but it isn't a great solution, we have this ticket for the correct fix: https://github.com/Consensys/teku/issues/8849

Changes:
- Merged both methods `processEffectiveBalanceUpdates` from AbstractEpochProcessor` and `EpochProcessorElectra`
- Updated `validatorStatuses` mid-epoch processing (not great for performance but good enough for fixing Mekong)

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
@@ -131,7 +131,7 @@ public abstract class AbstractEpochProcessor implements EpochProcessor {
     processEth1DataReset(state);
     processPendingDeposits(state);
     // This is only being used in Mekong temporarily
-    validatorStatuses = validatorStatusFactory.createValidatorStatuses(preState);
+    validatorStatuses = validatorStatusFactory.createValidatorStatuses(state);
     processPendingConsolidations(state);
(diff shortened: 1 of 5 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
```diff
@@ -460,32 +460,41 @@ public void processEffectiveBalanceUpdates(
...
-        BeaconStateCache.getTransitionCaches(state)
-            .getProgressiveTotalBalances()
-            .onEffectiveBalanceChange(status, newEffectiveBalance);
+      final Validator validator = validators.get(index);
+      final UInt64 effectiveBalanceLimit = getEffectiveBalanceLimitForValidator(validator);
+
+      // In Electra, it is possible that the validator set is updated within epoch processing. To
+      // avoid recalculating the ValidatorStatuses cache, we are manually updating the effective
+      // balance of any new
+      // validators that are not in the cache.
```
rolfyone: nit.. collapse that last line...

rolfyone (approved):
LGTM just some nits

PR #8851, opened 2024-11-26
Performance tracker

### LINKED ISSUE #8847: Performance tracker memory leak
There might be a memory leak in the attestation performance tracker

This happens systematically at every epoch transition, righ after sync committee pref log
image

mehdi-aouadi:
## PR Description
Update the performance tracker to support the new Electra attestation aggregation logic.
This prevents a memory leak triggered in Electra.

## Fixed Issue(s)

Fixes #8847 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
@@ -317,7 +317,10 @@ public class DefaultPerformanceTracker implements PerformanceTracker {
         slotToBitlists.merge(
             entry.getKey(),
             AttestationBitsAggregator.of(attestation, committeesSize),
-            AttestationBitsAggregator::or);
+            (firstBitsAggregator, secondBitsAggregator) -> {
+              firstBitsAggregator.or(secondBitsAggregator);
+              return firstBitsAggregator;
+            });
       }
     }
 
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.validator.coordinator.performance;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.annotations.VisibleForTesting;
+import it.unimi.dsi.fastutil.ints.Int2IntMap;
 import it.unimi.dsi.fastutil.ints.IntArrayList;
 import it.unimi.dsi.fastutil.ints.IntList;
 import it.unimi.dsi.fastutil.ints.IntSet;
@@ -42,7 +43,6 @@ import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.logging.StatusLogger;
 import tech.pegasys.teku.infrastructure.metrics.SettableGauge;
-import tech.pegasys.teku.infrastructure.ssz.collections.SszBitlist;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.datastructures.blocks.BeaconBlock;
@@ -51,6 +51,7 @@ import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
 import tech.pegasys.teku.spec.datastructures.operations.Attestation;
 import tech.pegasys.teku.spec.datastructures.operations.versions.altair.SyncCommitteeMessage;
 import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.statetransition.attestation.utils.AttestationBitsAggregator;
 import tech.pegasys.teku.storage.client.CombinedChainDataClient;
 import tech.pegasys.teku.validator.api.ValidatorPerformanceTrackingMode;
 import tech.pegasys.teku.validator.coordinator.ActiveValidatorTracker;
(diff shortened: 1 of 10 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
```diff
@@ -300,16 +301,23 @@ private AttestationPerformance calculateAttestationPerformance(
...
-    final Map<Bytes32, NavigableMap<UInt64, SszBitlist>> slotAndBitlistsByAttestationDataHash =
-        new HashMap<>();
+    final Map<Bytes32, NavigableMap<UInt64, AttestationBitsAggregator>>
+        slotAndBitlistsByAttestationDataHash = new HashMap<>();
     for (Map.Entry<UInt64, List<Attestation>> entry : attestationsIncludedOnChain.entrySet()) {
+      final Optional<Int2IntMap> committeesSize =
+          Optional.of(spec.getBeaconCommitteesSize(state, entry.getKey()));
       for (Attestation attestation : entry.getValue()) {
         final Bytes32 attestationDataHash = attestation.getData().hashTreeRoot();
-        final NavigableMap<UInt64, SszBitlist> slotToBitlists =
```
rolfyone, line 308: we could potentially use JMH on this...

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
```diff
@@ -314,9 +317,9 @@ private AttestationPerformance calculateAttestationPerformance(
         slotToBitlists.merge(
             entry.getKey(),
             AttestationBitsAggregator.of(attestation, committeesSize),
-            (first, second) -> {
-              first.or(second);
-              return first;
+            (firstBitsAggregator, secondBitAggregator) -> {
+              firstBitsAggregator.or(secondBitAggregator);
```
tbenr: I think here it make sense to let `or` return `this`... so we can chain operations and also have a compact code here..

tbenr: but we can leave it as a separate PR

tbenr: thinking about this a second time, I let the method return void on purpose to be clear that the object mutates.
I think it is safer this way, so ignore me.

mehdi-aouadi: The change is tiny and makes it more readable/concise, I added it :)

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
mehdi-aouadi: Doing that, will share the results

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
mehdi-aouadi: Just saw your last comment. Well, ok. Reverted it

mehdi-aouadi:
I run some benchmarks to compare the old and the new attestations performance algorithm and the results are encouraging.

**Old Version**

Phase0 Attestations
```
Benchmark                                                    (attestationsCount)   Mode  Cnt      Score     Error  Units
PerformanceTrackerBenchmark.calculateAttestationPerformance                  100  thrpt   10  27119.554 ± 597.657  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance                 1000  thrpt   10   2558.557 ±  99.012  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance                10000  thrpt   10    161.646 ±   1.570  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance               100000  thrpt   10     10.397 ±   0.026  ops/s
```
Doesn't work for the electra attestations, so no Electra attestations benchmarking
____________

**New Version**

Phase0 Attestations
```
Benchmark                                                    (attestationsCount)   Mode  Cnt      Score     Error  Units
PerformanceTrackerBenchmark.calculateAttestationPerformance                  100  thrpt   10  26828.725 ± 410.393  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance                 1000  thrpt   10   2638.513 ±  34.554  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance                10000  thrpt   10    160.236 ±   2.944  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance               100000  thrpt   10     11.682 ±   0.054  ops/s
```

Electra Attestations
```
Benchmark                                                    (attestationsCount)   Mode  Cnt      Score     Error  Units
PerformanceTrackerBenchmark.calculateAttestationPerformance                  100  thrpt   10  13805.354 ± 120.632  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance                 1000  thrpt   10   1733.083 ±   8.922  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance                10000  thrpt   10     98.000 ±   0.780  ops/s
PerformanceTrackerBenchmark.calculateAttestationPerformance               100000  thrpt   10      9.142 ±   0.030  ops/s
```
____________

Context:
- The operations refer to the `calculateAttestationPerformance` algorithm
- The `attestationsCount` is the number of attestations being processed. They are all attesting from the same slot and hence all aggregated (which is extreme)

Observations:
- The new algorithm performance is almost the same for the phase0 attestations
- The Electra attestations require almost twice the processing time but that's normal since the aggregation bit data structure is more complex. If we take into consideration that there are less attestations in Electra (more aggregation) this should be fine

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
mehdi-aouadi: Done: https://github.com/Consensys/teku/pull/8851#issuecomment-2506724226

rolfyone (commented):
LGTM

rolfyone (approved):

rolfyone:
we'll need to cherrypick this to the mekong branch too just remember...

PR #8853, opened 2024-11-27
alter ActiveP2pNetwork concept of close to in sync

rolfyone:
- `isInSync` is often set when there's no other peers to sync to so its not a great measure

 - moved the isCloseToInSync into recentChaindata

## Documentation

## Changelog

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/events/SyncStateTracker.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/events/SyncStateTracker.java
@@ -136,7 +136,6 @@ public class SyncStateTracker extends Service
       currentState = SyncState.SYNCING;
     } else if (startingUp) {
       currentState = SyncState.START_UP;
-
     } else {
       currentState = SyncState.IN_SYNC;
     }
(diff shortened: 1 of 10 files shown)


rolfyone:
i've done a few syncs on different networks, and its behaving as I expect, ready to review.

tbenr:
There is a side effect with this, which can be replicated this way:

1. run a single interop node and let it progress the chain for a while
2. stop the node and wait until isCloseToInSync will return false
3. start the node again
4. the node `SyncStateTracker` service start will timeout and call `onStartupTimeout`
5. we are not in a state where the single node start building the chain again but never starts the gossip again (so it will never communicate with an eventual node joining the network later)

previous behaviour
image

new behaviour:
image
image

so if a new node connects you end up in this weird situation where we don't publish messages to the other peer even if it is connected and just synced to the head.
image
image

tbenr:
I think we need `ActiveEth2P2PNetwork` to subscribe like `network.subscribeConnect(peer -> onPeerConnected());` and make sure we start gossip when we are in sync and we got peers connected `onPeerConnected`

tbenr:
@rolfyone  pushed a commit to cover the condition i was mentioning.

tbenr (approved):
maybe we could provide `isCloseToInSync` on the `onSyncStateChanged` method, instead of `isInSync`. The consumer doesn't get much value by just receiving that event and most likely need to call recenChainData to get the info. Doing that we could avoid multiple calls (for each subscriber) to recentChainData (and also avoid subscriber to require recenChainData dep injection)

PR #8855, opened 2024-11-27
Track produced block when builder doesn't reveal the payload

### LINKED ISSUE #8841: Wrong block performance when builder doesn't reveal the block but publishes it
when we publish and relay API timeouts we and up in this situation

```
2024-11-21 09:03:17.874 | Slow Block Publishing *** Slot: 3022816 start 2639ms, builder_get_payload +3002ms, complete +0ms total: 3002ms 
2024-11-21 09:03:18.877 | Validator   *** Published sync_signature     Count: 5, Slot: 3022816, Root: 54721197c06f2a585f35be1167cd4c5e0d82d024ced02bd52b60049df219c254 |  
2024-11-21 09:03:18.877 | Validator   *** Failed to produce block  Slot: 3022816 Validator: 902afe8

```

`Block performance: epoch 94463, expected 1, produced 0, included 0 (0%)`

but the block was actually produced but published by the relay

StefanBratanov:
## PR Description
Move the tracking of the produced block to the ValidatorApiHandler where we produce the unsigned block. I find this approach easier and more consistent than passing the `PerformanceTracker` all the way down to `ExecutionBuilderModule`

## Fixed Issue(s)
fixes #8841 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/AbstractBlockPublisher.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/AbstractBlockPublisher.java
@@ -73,6 +73,10 @@ public abstract class AbstractBlockPublisher implements BlockPublisher {
       final SignedBlockContainer blockContainer,
       final BroadcastValidationLevel broadcastValidationLevel,
       final BlockPublishingPerformance blockPublishingPerformance) {
+    // always track the block as produced even in case of publishing failures (e.g.
+    // relay API timeouts during unblinding), because we later do comparison against the chain data
+    // anyway
+    performanceTracker.saveProducedBlock(blockContainer.getSignedBlock().getSlotAndBlockRoot());
     return blockFactory
         .unblindSignedBlockIfBlinded(blockContainer.getSignedBlock(), blockPublishingPerformance)
         .thenCompose(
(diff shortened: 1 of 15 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/AbstractBlockPublisher.java
```diff
@@ -75,7 +77,18 @@ public SafeFuture<SendSignedBlockResult> sendSignedBlock(
...
+        .whenComplete(
+            (block, ex) -> {
+              if (ex != null) {
+                // in case of relay API timeouts when unblinding, we still want to assume the block
+                // as produced, since the relay could have published it
+                if (ExceptionUtil.hasCause(ex, TimeoutException.class)) {
+                  performanceTracker.saveProducedBlock(blockContainer.getSignedBlock());
+                }
+              } else {
+                performanceTracker.saveProducedBlock(block);
```
tbenr: do we really need to discriminate the exception here? wandering if we can save the block as produced

StefanBratanov: It doesn't hurt since we filter in the tracker compared to the chain we have anyways and exceptions shouldn't really happen

tbenr: but you could have a bunch of errors on API side, not just a timeout, and you never know if on the other side it got the info to publish.
Moreover, I think that we know we attempted to publish a produced block, so it make sense to me to blindly track it at the very beginning of the flow...

StefanBratanov: Committed a change to always save the block as produced

tbenr (approved):
LGTM.

Here we change the definition of "produced" from "produced, signed and about to publish" to "block produced but not yet signed nor published".
To me makes sense.

PR #8868, opened 2024-12-02
Enable publish blobs after block by default

tbenr:
I've been monitoring on holesky nodes with this enabled and seems not affect much on our high-upload-bandwidth nodes.
It should be worthy for low upload-bandwidth nodes, like home stakers building blocks locally.

Wait| average wait time
-- | --
disabled | 18.04 ms
enabled | 23.48 ms

over last 2000 blocks produced on holesky

it doesn't go lower than a certain delay because we anyway spend time in blobs preparation while block publishing is already started. So we additionally wait (when enabled) only when the "time-to-first-peer-published-block" is higher.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Additions and Improvements
 - improve block publishing performance, especially relevant with locally produced blocks
+- delay blobs publishing until the block is published to at least 1 peer, especially relevant with locally produced blocks with low upload bandwidth connections. Can be disabled via `--Xp2p-gossip-blobs-after-block-enabled=false`
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


mehdi-aouadi:
So is this to prioritise block publishing for low bandwidth nodes (when block and blobs publishing are done simultaneously, blobs could delay the block publishing by consuming bandwidth)?

tbenr:
yes, this is essentially to wait initiating blobs publishing until we successfully published the block to at least one peer

mehdi-aouadi (approved):
LGTM
Makes sense since it doesn't badly affect high bandwidth node and helps the low bandwidth ones

PR #8873, opened 2024-12-02
Added SingleAttestationEvent

rolfyone:
- if the attestation arriving is a singleAttestation, will fire a single_attestation event, otherwise will trigger an attestation event.

- at the moment we only see attestations, but added unit test and can re-verify once we have single attestations being produced...

 - can essentially listen to both by : `curl -H "Accept:text/event-stream" -N http://localhost:5051/eth/v1/events\?topics\=attestation,single_attestation`

## Documentation

## Changelog


tbenr (commented):
As mentioned during stand-up, I suspect some tests need to be updated to have something next to `tech/pegasys/teku/beaconrestapi/handlers/v1/events/schema/AttestationElectra.json` and the other events to check the schema.

rolfyone:
> As mentioned during stand-up, I suspect some tests need to be updated to have something next to `tech/pegasys/teku/beaconrestapi/handlers/v1/events/schema/AttestationElectra.json` and the other events to check the schema.

we haven't ever maintained that list, its basically just wrong but i'll try to make it less random.

lucassaldanha (approved):
LGTM. For this user-facing change, it might be worth adding a changelog entry.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/BeaconRestApiTypes.java
```diff
@@ -208,8 +211,11 @@ public class BeaconRestApiTypes {
           TOPICS,
           CoreTypes.string(
               "Event types to subscribe to."
-                  + " Available values include: [`head`, `finalized_checkpoint`, `chain_reorg`, `block`, "
-                  + "`attestation`, `voluntary_exit`, `contribution_and_proof`, `blob_sidecar`]\n\n",
+                  + " Supported event types: ["
+                  + EnumSet.allOf(EventType.class).stream()
+                      .map(val -> "`" + val.toString() + "`")
+                      .sorted()
+                      .collect(Collectors.joining(", ")),
```
tbenr, line 218: haven't you missed the closing `]` here?

rolfyone: yes :/ will put it in

PR #8874, opened 2024-12-03
Logic to handle new validators during epoch processing

### LINKED ISSUE #8849: Handle validator set update mid-epoch processing (Electra+)
After Electra, it is possible that the validator set is updated within epoch processing (process_pending_deposits). This means that the validator set in the state can get out of sync with our validatorStatuses cache. This is not a problem for the current epoch processing, but it can cause undesired side effects in the future.

Temporary fix: for Mekong, we are recalculating the validatorStatuses after process_pending_deposits

Correct fix: TBD

lucassaldanha:
## PR Description
This PR updates epoch processing logic to consider that post-Electra, the validator set can be updated in the mid-epoch processing (this did not happen before). The gist of the change is that we might need to update our `ValidatorStatuses` object, which acts as a cache of the validators in the validator set. We don't ever want to run this logic before Electra, and even in Electra, we only want to do it if we detect the validator set has changed.

## Fixed Issue(s)
fixes #8849 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
@@ -15,6 +15,8 @@ package tech.pegasys.teku.spec.logic.common.statetransition.epoch;
 
 import static tech.pegasys.teku.infrastructure.unsigned.UInt64.ZERO;
 
+import com.google.common.annotations.VisibleForTesting;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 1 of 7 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
```diff
@@ -148,6 +148,40 @@ protected void processEpoch(final BeaconState preState, final MutableBeaconState
     }
   }
 
+  @VisibleForTesting
+  public ValidatorStatuses recreateValidatorStatusIfNewValidatorsAreFound(
+      final BeaconState state,
+      final ValidatorStatuses validatorStatuses,
+      final UInt64 currentEpoch) {
+    final int preValidatorCount = validatorStatuses.getValidatorCount();
```
rolfyone: i'd probably call this `cachedValidatorCount`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/statetransition/epoch/AbstractEpochProcessor.java
```diff
@@ -148,6 +148,40 @@ protected void processEpoch(final BeaconState preState, final MutableBeaconState
...
+    final int postValidatorCount = state.getValidators().size();
```
rolfyone: `stateValidatorCount`

rolfyone (approved):
LGTM just nits.

PR #8880, opened 2024-12-03
Fix ExecutionRequestsDataCodec handling of missing request data

lucassaldanha:
## Fixed Issue(s)
With recent changes in the execution requests, it is not valid to send an empty list for
a request type. So we should not accept an element with only a single byte as the
request type. 

Thanks @jtraglia for finding this issue! :)

## Documentation

## Changelog


rolfyone (approved):

jtraglia (approved):
LGTM 👍

PR #8889, opened 2024-12-05
Fix `is_syncing` API reporting when head is synced optimistically

### LINKED ISSUE #8888: Teku reports `"is_syncing": true` when `sync_distance ==0`
### Description

When teku is in optimistic sync mode and 0 distance away from the head, it still reports `is_syncing: true`. 
Every other clients reports as false. Thus I assume this is a teku bug.  API call: `/eth/v1/node/syncing`

```json
teku-nimbusel-1 | CHANGED | rc=0 >>
{
  "data": {
    "head_slot": "214947",
    "sync_distance": "0",
    "is_syncing": true,
    "is_optimistic": true,
    "el_offline": false
  }
}
nimbus-nimbusel-1 | CHANGED | rc=0 >>
{
  "data": {
    "head_slot": "214947",
    "sync_distance": "0",
    "is_syncing": false,
    "is_optimistic": true,
    "el_offline": false
  }
}
grandine-nimbusel-1 | CHANGED | rc=0 >>
{
  "data": {
    "head_slot": "214947",
    "sync_distance": "0",
    "is_syncing": false,
    "is_optimistic": true,
    "el_offline": false
  }
}
lighthouse-nimbusel-1 | CHANGED | rc=0 >>
{
  "data": {
    "is_syncing": false,
    "is_optimistic": true,
    "el_offline": false,
    "head_slot": "214947",
    "sync_distance": "0"
  }
}
lodestar-nimbusel-1 | CHANGED | rc=0 >>
{
  "data": {
    "head_slot": "214947",
    "sync_distance": "0",
    "is_syncing": false,
    "is_optimistic": true,
    "el_offline": false
  }
}
prysm-nimbusel-1 | CHANGED | rc=0 >>
{
  "data": {
    "head_slot": "214947",
    "sync_distance": "0",
    "is_syncing": false,
    "is_optimistic": true,
    "el_offline": true
  }
}
```
### Steps to Reproduce (Bug)
Setup a node with a broken EL, and let teku sync optimistically. 

I would expect teku to also report `is_syncing: false` when the follow distance is 0. 

Teku reports `is_syncing: true` when follow distance is 0.

always

`ethpandaops/teku:mekong-297263b`

StefanBratanov:
## PR Description
When we optimistically track head and sync distance is 0, technically we should report `is_syncing: false` even though we are waiting for the EL.

## Fixed Issue(s)
fixes #8888 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,3 +15,4 @@
 
 ### Bug Fixes
 - Added a startup script for unix systems to ensure that when jemalloc is installed the script sets the LD_PRELOAD environment variable to the use the jemalloc library
+- Set `is_syncing` to `false` instead of `true` for the `/eth/v1/node/syncing` API endpoint when the head is optimistic and the sync distance is 0
--- a/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetSyncing.java
+++ b/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/node/GetSyncing.java
@@ -97,12 +97,11 @@ public class GetSyncing extends RestApiEndpoint {
         final ExecutionClientDataProvider executionClientDataProvider) {
       final SyncingStatus status = syncProvider.getSyncingStatus();
       final SyncState syncState = syncProvider.getCurrentSyncState();
-      this.isSyncing = !syncState.isInSync();
+      this.slotsBehind = calculateSlotsBehind(status, syncState);
+      this.isSyncing = !slotsBehind.isZero();
       this.elOffline = Optional.of(!executionClientDataProvider.isExecutionClientAvailable());
       this.isOptimistic = Optional.of(syncState.isOptimistic());
       this.currentSlot = status.getCurrentSlot();
-      // do this last, after isSyncing is calculated
-      this.slotsBehind = calculateSlotsBehind(status);
     }
 
     SyncStatusData(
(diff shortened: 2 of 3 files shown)


mehdi-aouadi (commented):
I'm a bit bothered by the logic difference between the API and the `SyncStateTracker`: the latter doesn't take the slots behind into consideration at all to mark the node `IN_SYNC`. This definitely fixes the issue we have though.
I'm wondering if it wouldn't be better to first check the `SyncStateTracker` to see if the node is `IN_SYNC` otherwise consider checking the slot behind...
Not a strong opinion here, just making sure we cover all the cases...

StefanBratanov:
> I'm a bit bothered by the logic difference between the API and the `SyncStateTracker`: the latter doesn't take the slots behind into consideration at all to mark the node `IN_SYNC`. This definitely fixes the issue we have though. I'm wondering if it wouldn't be better to first check the `SyncStateTracker` to see if the node is `IN_SYNC` otherwise consider checking the slot behind... Not a strong opinion here, just making sure we cover all the cases...

I had a look at the `SyncStateTracker` and we have this notion that if we are `AWAITING_EL` (tracking optimistic head with sync_distance = 0) then we are still syncing and isInSync() returns false. This cascades to a few places in the code, specifically around the fact that when we are tracking optimistic head, we shouldn't perform any duties and also in `onTick()`. I think I would raise an issue instead explaining this and leave this change relatively simple only impacting the API code.

StefanBratanov:
Raised https://github.com/Consensys/teku/issues/8890 to track usages of `isInSync`

mehdi-aouadi (approved):
Let's fix the API and we'll review the SyncStateTracker in the issue you raised. Thanks!

PR #8896, opened 2024-12-06
Fix blob's gossip subscriptions

tbenr:
We are incorrectly calculating relevant blobs topics for upcoming forks.

When new specConfig has been introduced, an assumption has been broken in `Eth2GossipTopicFilter`, which was passing the current milestone to `getAllTopics` even when calculating topics for the future forks.
It was fine because `getAllTopics` is only getting the config and from the spec, which in previous specConfig design was always the "latest defined milestone config".
With recent changes, it is no more the case, so we need to pick config from the correct milestone.

Unit test was broken because it was passing the wrong milestone (the next fork milestone, instead of the current one)

I fixed the bug and the tests.

Moreover the Electra blobs bump would have break `getAllTopics` in blob topics calculation. Thanks to new dynamic spec config, that works transparently now.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #8901, opened 2024-12-09
add missing electra attestation processing committee check

### LINKED ISSUE #8892: Fix reference test invalid_nonset_bits_for_one_committee (alpha.9)
We need to understand why the test is broken and fix it. Also, we need to remove it from the ignore list.

mehdi-aouadi:
## PR Description
Add missing committee attester check when processing Electra attestations:
https://github.com/ethereum/consensus-specs/blob/31cd9cbec2d93e5a6df84dcee9ca9e8aa77803ee/specs/electra/beacon-chain.md?plain=1#L1330

## Fixed Issue(s)

fixes #8892 

## Documentation

## Changelog


tbenr:
humm looks like we do things slightly differently there.. for instance where is `assert len(attestation.aggregation_bits) == committee_offset`?

mehdi-aouadi:
> humm looks like we do things slightly differently there.. for instance where is `assert len(attestation.aggregation_bits) == committee_offset`?

That's done: 
```
if (participantsCount != aggregationBits.size()) {
      return Optional.of(AttestationInvalidReason.PARTICIPANTS_COUNT_MISMATCH);
    }
```
https://github.com/ConsenSys/teku/blob/8f291ea277a3ace4e1aee6e9c388cc6e325efedc/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java#L748-L750

As discussed, I'll push some other changes to make it more aligned with the spec.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -732,6 +733,16 @@ private Optional<OperationInvalidReason> checkCommittees(
...
+      final int committeeOffset = participantsCount;
+      final long committeeAttesterCount =
+          IntStream.range(0, committee.size())
+              .filter(
+                  committeeParticipantIndex ->
+                      aggregationBits.isSet(committeeOffset + committeeParticipantIndex))
+              .count();
+      if (committeeAttesterCount == 0) {
+        return Optional.of(AttestationInvalidReason.PARTICIPANTS_COUNT_MISMATCH);
+      }
```
tbenr: I think this could be implemented more efficiently as the following:

```java
      final boolean aggregationBitsHasAtLeastOneParticipantFromTheCommittee =
              IntStream.range(0, committee.size())
                      .anyMatch(
                              committeeParticipantIndex ->
                                      aggregationBits.isSet(committeeOffset + committeeParticipantIndex));

      if (!aggregationBitsHasAtLeastOneParticipantFromTheCommittee) {
        return Optional.of(AttestationInvalidReason.PARTICIPANTS_COUNT_MISMATCH);
      }
```

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -725,16 +726,25 @@ private Optional<OperationInvalidReason> checkCommittees(
...
+    int committeeOffset = 0;
     for (final UInt64 committeeIndex : committeeIndices) {
       if (committeeIndex.isGreaterThanOrEqualTo(committeeCountPerSlot)) {
         return Optional.of(AttestationInvalidReason.COMMITTEE_INDEX_TOO_HIGH);
       }
       final IntList committee =
           beaconStateAccessorsElectra.getBeaconCommittee(state, slot, committeeIndex);
-      participantsCount += committee.size();
+      final int currentCommitteeOffset = committeeOffset;
+      final boolean hasAtLeastOneParticipant =
```
tbenr: Looking here again, maybe is better to resemble the `committee_attesters` variable from the spec and call this `committeeHasAtLeastOneAttester`

tbenr (approved):
LGTM. Just a var name suggestion

PR #8902, opened 2024-12-09
Fix `SingleAttestation` in `PerformanceTracker` bug

tbenr:
When we produce an attestation we save it in `saveProducedAttestation`.
If it is a singleAttestation it will break the `PerformanceTracker` because it is designed to work for non-single attestations.

This pr makes sure we save converted attestations


mehdi-aouadi (approved):
LGTM

PR #8907, opened 2024-12-10
Allow pending attestations to trigger recent block fetcher

tbenr:
## Documentation

## Changelog


lucassaldanha:
This is interesting. Totally something I expected Teku was already doing, but it looks like it hasn't done it in a long time... Was it deliberate??

The changes make sense. My only concern is if we could end up fetching blocks too aggressively. Either way, we are better doing some throughout sync testing to see.

tbenr:
@lucassaldanha 
I did some syncing test and all seems fine. since we don't do any RPC if we are not in sync, then all the attempts when we are syncing close to the head are skipped anyway. I think we can go for it.

lucassaldanha (approved):

PR #8909, opened 2024-12-11
Implemented expectedGasLimit from spec

rolfyone:
Made a more complete test suite on the FN to show boundaries.

## Documentation

## Changelog


ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidatorImpl.java
```diff
@@ -122,20 +122,21 @@ public void validateBuilderBid(
...
+      eventLogger.builderBidNotHonouringGasLimit(
+          parentGasLimit, proposedGasLimit, preferredGasLimit);
     }
+  }
 
-    if (preferredGasLimit.isLessThan(parentGasLimit)
-        && proposedGasLimit.isLessThan(parentGasLimit)) {
-      return;
+  static UInt64 expectedGasLimit(final UInt64 parentGasLimit, final UInt64 targetGasLimit) {
+    final UInt64 maxGasLimitDifference = parentGasLimit.dividedBy(1024L).minusMinZero(1);
```
jtraglia, line 133: Why is the `.minusMinZero(1)` here?

Ah. It's because it asserts that it's strictly less than or greater than the value.

image

For example, given that

* `parent_gas_limit == 1024_000`
* `parent_gas_limit // 1024 == 1000`

`gas_limit` must be less than `1024_000 + 1000` (1025_000), so the max value is 1024_999.

rolfyone: i'll try to find the original spec entry, i was referencing a PR from prysm, which did the -1...

ethereum/executionlayer/src/test/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidatorTest.java
```diff
@@ -184,27 +200,44 @@ void shouldNotLogEventIfGasLimitStaysTheSame() throws BuilderBidValidationExcept
...
+        // down inside bounds
+        Arguments.of(1024_000L, 1023_500L, 1023_500L),
+        // down outside bounds - results in a partial move
+        Arguments.of(36_000_000L, 35_000_000L, 35_964_845L),
+        Arguments.of(1024_000L, 1020_000L, 1023_001L),
+        // increase outside bounds - results in a partial move
+        Arguments.of(1024_000L, 1025_000L, 1024_999L),
+        Arguments.of(30_000_000L, 36_000_000L, 30_029_295L),
+        // increase inside bounds
+        Arguments.of(1024_000L, 1024_500L, 1024_500L));
```
rolfyone, line 240: @jtraglia so the args are parent, target, expected output...

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidatorImpl.java
rolfyone: @jtraglia The test below does demonstrate the bounds of this function also... i think you've explained exactly why the minus 1...

rolfyone: i think given the thread and the test match, i'll mark this resolved but feel free to ping me if this is an incorrect assertion

tbenr (approved):
LGTM

ethereum/executionlayer/src/main/java/tech/pegasys/teku/ethereum/executionlayer/BuilderBidValidatorImpl.java
nflaig: > i'll try to find the original spec entry, i was referencing a PR from prysm, which did the -1...

so I ended up here while looking for this in the spec, only to notice it actually is not part of any spec but only documented in EIP-1559...

@jtraglia should we add this to the bellatrix builder spec?

PR #8864, opened 2024-11-29
Reduce blobs lookup min wait time to 0

tbenr:
Now that we have decent getBlobs support, we can lookup blobs faster when the block shows up close to 4s boundary

related to https://github.com/Consensys/teku/issues/8796

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [ ] I thought about adding a changelog entry, and added one if I deemed necessary.

Issue #8796, opened 2024-10-29
Remove delay when looking up blobs via local EL


tbenr:
Once there is enough adoption of `engine_getBlobs` we can remove the delay and do the local lookup of the missing blobs immediately once we receive the block.
RPC delay should remain unchanged.

samr874:
Any update about this issue?? who is taking over @tbenr ?

tbenr:
some informations:
1. we should not try to lookup the block via RPC too fast when we first see a blob sidecar. The process when we receive a blob first could be
  - store it
  - wait (maybe with the same "elastic" logic but with different params)
  - fetch block RPC if not received yet.
2. I think we can remove the attempt to recover blobs when we still don't have a block. Currently we try to download all max blobs (6). In Electra it will be even 9. I think we could be less aggressive now that we have local EL option. 
3. the process when we receive a block could be
  - immediately lookup for blobs via local EL
  - (not sure we should add a delay here, currently we don't)
  - broadcast reconstructed blobs
  - (delay here?)
  - lookup still missing blobs via RPC

PR #8911, opened 2024-12-11
Revert "Reduce blobs lookup min wait time to 0 (#8864)" and "Deprecate TTFB, RESP_TIMEOUT, ... (#8839)"

tbenr:
This reverts commit 809ea98c4413c2f678be9a97fff316f2374d7b2f.
reverts also d633566c17a9602c5ef7351394e65458479b43d5 because we don't want to release that.

too many RPC requests due to this change (block and blobs)
(last column is the amount of requests in a 12h timeframe. canary is old, holesky-01 is new)
image

I'll put a PR to improve the logic.

related to #8796

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8916, opened 2024-12-12
Use 16-bit random value in validator filter

jtraglia:
## PR Description

This PR implements the changes defined here:

* https://github.com/ethereum/consensus-specs/pull/4039

Notably, this PR will:

* Override `computeProposerIndex` and `getNextSyncCommitteeIndices`.
* In `computeProposerIndex`, rename `hash` to `randomBytes` to match the spec.
* Add cached `randomBytes` to `getNextSyncCommitteeIndices`.
* Define new `MAX_RANDOM_VALUE` equal to `2**16-1`.

See the original implementations here:

https://github.com/Consensys/teku/blob/7de9aa456a0bf96c773beb541b113a60c2b56a6b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/helpers/MiscHelpers.java#L112-L139

https://github.com/Consensys/teku/blob/7de9aa456a0bf96c773beb541b113a60c2b56a6b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/altair/helpers/BeaconStateAccessorsAltair.java#L104-L132

## Documentation

## Changelog


tbenr:
in this PR we should re-enable2 fizzing tests disabled here: https://github.com/Consensys/teku/pull/8923

jtraglia:
Hmm I'm not sure why these are still failing. The only difference is the `body_root`.

image

rolfyone:
> Hmm I'm not sure why these are still failing. The only difference is the `body_root`.

I've got a body root diff in mine too...

tbenr:
this is weird. I'll spend some more time on it tomorrow, if in the meantime you don't solve the mystery

tbenr:
lol, my bad. Seems like I copied the wrong spec files. I also fixed a uint test that seems to be affected by the change

jtraglia:
Haha no worries. Thank you for fixing this. Comparing `prettySerialize` outputs is a good idea too!

tbenr (approved):
LGTM

PR #8917, opened 2024-12-13
added test case for comparing mainnet config output

rolfyone:
## Documentation

- [X] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [X] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #8919, opened 2024-12-13
Fix config provider to fix spec config api

tbenr:
This brings back all missing values from the spec config api

related to #8917

## Documentation

## Changelog


StefanBratanov (approved):
I guess also could add a line in the CHANGELOG

rolfyone (approved):

PR #8927, opened 2024-12-16
No delay when fetching blobs with known block and no attempt to recover blobs for unknown block

### LINKED ISSUE #8796: Remove delay when looking up blobs via local EL

StefanBratanov:
## PR Description

- Separate between RPC block fetching and RPC blobs fetching

Changed the flow to be as such:

- If we first receive a blob sidecar: Only trigger block fetch after a delay (same as before)
- If we first receive block: No delay and trigger EL lookup. RPC lookup is attempted after a delay (same as before)
- On arrival of block, if tracker exists: We trigger the blobs lookup from EL again and RPC lookup (if tracker is not complete)

## Fixed Issue(s)
fixes #8796 

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
@@ -112,7 +112,7 @@ public class BlockBlobSidecarsTracker {
 
   public Stream<BlobIdentifier> getMissingBlobSidecars() {
     final Optional<Integer> blockCommitmentsCount = getBlockKzgCommitmentsCount();
-    checkState(blockCommitmentsCount.isPresent(), "Block must me known to call this method");
+    checkState(blockCommitmentsCount.isPresent(), "Block must be known to call this method");
 
     return UInt64.range(UInt64.ZERO, UInt64.valueOf(blockCommitmentsCount.get()))
         .filter(blobIndex -> !blobSidecars.containsKey(blobIndex))
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
@@ -40,13 +40,15 @@ import tech.pegasys.teku.statetransition.block.BlockImportChannel;
 
 public class BlockBlobSidecarsTracker {
   private static final Logger LOG = LogManager.getLogger();
+
   private static final UInt64 CREATION_TIMING_IDX = UInt64.MAX_VALUE;
   private static final UInt64 BLOCK_ARRIVAL_TIMING_IDX = CREATION_TIMING_IDX.decrement();
-  private static final UInt64 RPC_FETCH_TIMING_IDX = BLOCK_ARRIVAL_TIMING_IDX.decrement();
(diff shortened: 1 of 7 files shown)


tbenr:
nice! I'll play around with this tomorrow

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -315,22 +308,31 @@ private void printDebugTimings(final Map<UInt64, Long> debugTimings) {
...
     } else {
-      timingsReport.append("Local EL fetch wasn't required - ");
+      timingsReport.append("Local EL blobs fetch wasn't required - ");
+    }
+
+    if (debugTimings.containsKey(RPC_BLOCK_FETCH_TIMING_IDX)) {
+      timingsReport
+          .append("RPC block fetch delay ")
+          .append(debugTimings.get(RPC_BLOCK_FETCH_TIMING_IDX) - creationTime)
+          .append("ms");
```
tbenr:
```
          .append("ms - ");
    } else {
      timingsReport.append("RPC block fetch wasn't required - ");
```

tbenr (commented):
As discussed I'd add an delay before requesting blobs via RPC when local el doesn't have the missing blob(s)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -315,22 +308,31 @@ private void printDebugTimings(final Map<UInt64, Long> debugTimings) {
...
+    if (debugTimings.containsKey(RPC_BLOCK_FETCH_TIMING_IDX)) {
       timingsReport
-          .append("Local EL fetch delay ")
-          .append(debugTimings.get(LOCAL_EL_FETCH_TIMING_IDX) - creationTime)
+          .append("RPC block fetch delay ")
+          .append(debugTimings.get(RPC_BLOCK_FETCH_TIMING_IDX) - creationTime)
           .append("ms - ");
     } else {
-      timingsReport.append("Local EL fetch wasn't required - ");
+      timingsReport.append("RPC block fetch wasn't required");
```
tbenr: final ` - ` here :-)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/blobs/BlockBlobSidecarsTracker.java
```diff
@@ -111,29 +111,11 @@ public Optional<BlobSidecar> getBlobSidecar(final UInt64 index) {
...
-
-    // We may return maxBlobsPerBlock because we don't know the block
-    return UInt64.range(UInt64.ZERO, maxBlobsPerBlock)
-        .filter(blobIndex -> !blobSidecars.containsKey(blobIndex))
-        .map(blobIndex -> new BlobIdentifier(slotAndBlockRoot.getBlockRoot(), blobIndex));
-  }
-
-  public Stream<BlobIdentifier> getUnusedBlobSidecarsForBlock() {
     final Optional<Integer> blockCommitmentsCount = getBlockKzgCommitmentsCount();
     checkState(blockCommitmentsCount.isPresent(), "Block must me known to call this method");
```
mehdi-aouadi: Since we're at it: must **be** know

mehdi-aouadi (approved):
LGTM
I'd suggest running a local interop or deploy it to a Holesky node and see how it goes

PR #8929, opened 2024-12-17
Fix single attestation publication

tbenr:
When publishing we need to make sure we select the unconverted attestation otherwise we end up publishing the converted `AttestationElectra` instead of `SingleAttestation` type.

## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #8941, opened 2024-12-19
optimize indexed attestation validation

tbenr:
I Noticed an optimization when we validate IndexedAttestations

- we can preallocate the array to the size we expect
- we can do the sorting check while retrieving the public keys

test run on a very recent mainnet state with an >500 aggregated indexed attestation

407MB vs 1.17GB of Memory allocations

30% faster

```
Benchmark                                                          Mode  Cnt   Score   Error   Units
IndexedAttestationValidationBenchmark.validateIndexedAttestation  thrpt   25  89.007 ± 1.200  ops/ms
```

```
Benchmark                                                          Mode  Cnt    Score   Error   Units
IndexedAttestationValidationBenchmark.validateIndexedAttestation  thrpt   25  116.165 ± 5.730  ops/ms
```

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #8952, opened 2024-12-21
Batch call for retrieving blobs from RPC

### LINKED ISSUE #8928: Batch call for blobs RPC requests
When we follow the chain, we fallback to the RPC for retrieving missing blobs if EL retrieval doesn't work. However, we call `requestBlobSidecarByRoot` for each blob identifier, so we end up making _n_ amount of requests, rather than one bulk request using `requestBlobSidecarsByRoot`.

StefanBratanov:
## PR Description
When we need to retrieve blobs from RPC when following the head, we do one request per a sidecar. This PR changes this functionality to do one batch call. It will save bandwith (when used) and could help with rate limiting. Ideally, this should be merged after https://github.com/Consensys/teku/pull/8927

## Fixed Issue(s)
fixes #8928 

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/fetch/FetchBlobSidecarsTask.java
```diff
@@ -13,55 +13,63 @@
...
 import tech.pegasys.teku.beacon.sync.fetch.FetchResult.Status;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.networking.eth2.peers.Eth2Peer;
 import tech.pegasys.teku.networking.p2p.network.P2PNetwork;
+import tech.pegasys.teku.networking.p2p.rpc.RpcResponseListener;
 import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.BlobSidecar;
 import tech.pegasys.teku.spec.datastructures.networking.libp2p.rpc.BlobIdentifier;
 
-public class FetchBlobSidecarTask extends AbstractFetchTask<BlobIdentifier, BlobSidecar> {
+public class FetchBlobSidecarsTask extends AbstractFetchTask<Bytes32, List<BlobSidecar>> {
```
tbenr: I have doubts here.
In principle `K` type should uniquely identify a task, so I think we should have a `record(blockroot,  List<BlobIdentifier>)`.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/gossip/blobs/RecentBlobSidecarsFetchService.java
```diff
@@ -87,34 +86,62 @@ public void subscribeBlobSidecarFetched(final BlobSidecarSubscriber subscriber)
...
+            .toList();
+    if (requiredBlobIdentifiers.isEmpty()) {
+      // We already have all required blob sidecars
       return;
     }
-    final FetchBlobSidecarTask task = createTask(blobIdentifier);
-    if (allTasks.putIfAbsent(blobIdentifier, task) != null) {
+    final FetchBlobSidecarsTask task =
+        fetchTaskFactory.createFetchBlobSidecarsTask(blockRoot, requiredBlobIdentifiers);
+    if (allTasks.putIfAbsent(blockRoot, task) != null) {
```
tbenr: so here we can have problems if we change our mind and we initiate a request with a different `List<blobIdentifier>` for the same block root.

tbenr: we could argue that the number of blobs identifier can only decrease with time, so we don't care if we take the initial request and cancel the new one, but:
- introduces an assumption that needs to be handled here, and it is risky (if we don't check we may cancel the task containing what we really need)
- anyway this approach breaks the design of the abstract class IMO

so if we move to the theoretical correct `key` we could and up requesting the same blob(s) in two different tasks, but at least we know that we will have them all executed despite the order of arrival

tbenr: so at the end:

option 1:
- change the key and allow multiple request for same blockRoot (potential duplication of blobs download)
- cancel method still based on blockRoot (looping on allTasks)

option 2:
- leave the key as blockRoot, adding some checks when adding a task for a blockRoot already present (logging an error if this is not supposed to happen, or some logic that compares the identifier list, cancel one and leave the other)
- cancel method remains as is

StefanBratanov:
Had an internal discussion and decided to potentially revisit this idea in the future

PR #8953, opened 2024-12-23
fix: Fix comparison operator for numeric values

famouswizard:
## PR Description

I noticed that the script was using `==` to compare two numeric values in the conditional statement, which is typically used for string comparisons in bash. For numeric comparisons, it's more appropriate to use `-eq`. I've updated the conditional to use `-eq` instead of `==`, ensuring that the comparison works as expected for numbers.

Here’s the change:
- Replaced `==` with `-eq` in the comparison:
  ```bash
  if [ "$file_dircount" -eq "$testfile_dircount" ]; then
  ```

This update ensures that the comparison is treated correctly as a numerical one, preventing potential issues in certain scenarios.

## Documentation

## Changelog

### DIFF
--- a/scripts/testcheck.sh
+++ b/scripts/testcheck.sh
@@ -59,7 +59,7 @@ for (( i=0; i<${#TEST_TYPE[@]}; i++ )); do
           # directories. This helps identify obviously incorrect paths;
           # the test file is probably in the wrong directory.
           #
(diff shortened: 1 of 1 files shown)


rolfyone:
closing due to contribution guidelines, the script is working as is.

PR #8955, opened 2024-12-23
Add spec tags for Electra functions

jtraglia:
## PR Description

This PR adds comments with specification functions to Electra functions. This uses [`eth-spec-tags`](https://github.com/jtraglia/eth-spec-tags) to do this. What I did was locate each function, add a tag, and then run a script. This is marked as a draft until I (1) finish adding spec tags and (2) add a check (either a non-required CI check or a scheduled action) to ensure things are up-to-date.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExpectedWithdrawals.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExpectedWithdrawals.java
@@ -289,6 +289,37 @@ public class ExpectedWithdrawals {
         genericState, schemaDefinitionsCapella, beaconStateMutators, specConfigCapella);
   }
 
+  /*
+   * <spec function="process_withdrawals" fork="electra">
+   * def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
+   *     # [Modified in Electra:EIP7251]
+   *     expected_withdrawals, processed_partial_withdrawals_count = get_expected_withdrawals(state)
+   *
+   *     assert payload.withdrawals == expected_withdrawals
+   *
+   *     for withdrawal in expected_withdrawals:
+   *         decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
+   *
+   *     # Update pending partial withdrawals [New in Electra:EIP7251]
+   *     state.pending_partial_withdrawals = state.pending_partial_withdrawals[processed_partial_withdrawals_count:]
+   *
+   *     # Update the next withdrawal index if this block contained withdrawals
+   *     if len(expected_withdrawals) != 0:
+   *         latest_withdrawal = expected_withdrawals[-1]
+   *         state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
+   *
+   *     # Update the next validator index to start the next withdrawal sweep
+   *     if len(expected_withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
+   *         # Next sweep starts after the latest withdrawal's validator index
+   *         next_validator_index = ValidatorIndex((expected_withdrawals[-1].validator_index + 1) % len(state.validators))
+   *         state.next_withdrawal_validator_index = next_validator_index
+   *     else:
+   *         # Advance sweep by the max length of the sweep if there was not a full set of withdrawals
+   *         next_index = state.next_withdrawal_validator_index + MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+   *         next_validator_index = ValidatorIndex(next_index % len(state.validators))
+   *         state.next_withdrawal_validator_index = next_validator_index
+   * </spec>
+   */
   void processWithdrawalsUnchecked(
       final MutableBeaconState genericState,
       final SchemaDefinitionsCapella schemaDefinitionsCapella,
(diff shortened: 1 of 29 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExpectedWithdrawals.java
```diff
@@ -289,6 +289,37 @@ public void processWithdrawals(
         genericState, schemaDefinitionsCapella, beaconStateMutators, specConfigCapella);
   }
 
+  /*
+   * <spec function="process_withdrawals" fork="electra">
+   * def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
+   *     # [Modified in Electra:EIP7251]
+   *     expected_withdrawals, processed_partial_withdrawals_count = get_expected_withdrawals(state)
+   *
```
rolfyone, line 297: do we want the whole function? i can see why the function name is useful and potentially a URL, but maybe the whole spec fn is too much? would be interested in other opinions here for sure.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/electra/BeaconStateElectra.java
```diff
@@ -35,6 +35,62 @@
 import tech.pegasys.teku.spec.datastructures.state.versions.electra.PendingDeposit;
 import tech.pegasys.teku.spec.datastructures.state.versions.electra.PendingPartialWithdrawal;
 
+/*
+ * <spec ssz_object="BeaconState" fork="electra">
+ * class BeaconState(Container):
+ *     # Versioning
+ *     genesis_time: uint64
+ *     genesis_validators_root: Root
```
rolfyone: state for example is a good case of it being almost impossible to reconcile if its inline in the code imo...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -240,7 +279,70 @@ public void processDepositRequests(
     }
   }
 
-  /** Implements process_withdrawal_request from consensus-specs (EIP-7002 & EIP-7251). */
+  /*
+   * <spec function="process_withdrawal_request" fork="electra">
+   * def process_withdrawal_request(
+   *     state: BeaconState,
+   *     withdrawal_request: WithdrawalRequest
+   * ) -> None:
```
rolfyone, line 285: this is another good example of it just being too much maybe...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExpectedWithdrawals.java
jtraglia: Yeah it certainly can be a lot, but that's what makes it valuable.

* If the spec changes, the CI action will notify devs. This isn't an option without the whole thing.
* It makes it easier for auditors to compare implementation to specification.
* When implementing new spec, having the whole spec there will make implementation easier.
* The spec tags themselves makes it easier to location important things.

How do you feel about a "diff" option, something like this:

```diff
<spec function="process_withdrawals" fork="electra" style="diff">
--- capella
+++ electra
@@ -1,10 +1,15 @@
 def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
-    expected_withdrawals = get_expected_withdrawals(state)
+    # [Modified in Electra:EIP7251]
+    expected_withdrawals, processed_partial_withdrawals_count = get_expected_withdrawals(state)
+
     assert payload.withdrawals == expected_withdrawals

     for withdrawal in expected_withdrawals:
         decrease_balance(state, withdrawal.validator_index, withdrawal.amount)

+    # Update pending partial withdrawals [New in Electra:EIP7251]
+    state.pending_partial_withdrawals = state.pending_partial_withdrawals[processed_partial_withdrawals_count:]
+
     # Update the next withdrawal index if this block contained withdrawals
     if len(expected_withdrawals) != 0:
         latest_withdrawal = expected_withdrawals[-1]
</spec>
```

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
jtraglia: This is a tough one. It's just a really large function in the specs.

StefanBratanov:
I have mixed feelings about it. From one side, I like it, from another I think it is too much meta and cumbersome to maintain. I don't see much value in annotating the containers. I like the benefit of annotating each spec function (we do have something similar with comments, but not everywhere and also inconsistent). However, instead of the whole spec code, can't we use commit hash and get the CI to pick the code dynamically?

jtraglia:
> I think it is too much meta and cumbersome to maintain.

Just to be clear, these are generated with a script. It will automatically pull in the spec.

> can't we use commit hash and get the CI to pick the code dynamically?

This is sort of what this script is doing actually! But with a tag instead of a link; I believe the former is more readable.

```
<spec function="foo" fork="electra"></spec>
```

We could have the CI do this automatically, but how would client devs know when the spec has changed & the implementation needs to be updated? One of my goals in this is make the situation very clear. One of the reasons I made this was because Prysm manually copied spec into their comments like this, but would then update the implementation and not update the comment, which made reviewing their code really confusing.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExpectedWithdrawals.java
jtraglia: I added a `diff` style option which might help here.

Please see: https://github.com/jtraglia/eth-spec-tags/blob/main/README.md#style

jtraglia:
I've decided to close this. I have some ideas on how to deal with large spec items, but it will require some work. I'll ping you guys on discord if/when that's ready & we can re-evaluate it then.

PR #8961, opened 2024-12-31
Make `engine_getBlobsV1` not fork specific

### LINKED ISSUE #8669: Make the engine get blobs method required
The current implementation of the `engine_getBlobsV1` is marked as non critical and doesn't trigger an EL status update (in `DefaultExecutionWeb3jClientProvider::buildClient`) since it hasn't been implemented by all the ELs yet.
This should be removed from the non critical methods once it's implemented widely.
The required change:
Remove `engine_getBlobsV1` from `DefaultExecutionWeb3jClientProvider::buildClient`

A new notion of optional execution RPC methods has also been added to handle the use case when an EL doesn't support the new method. This should be removed once all ELs have implemented it.
The required change:
Remove `isOptional` from `EngineJsonRpcMethod`
Remove `MilestoneBasedEngineJsonRpcMethodsResolver::getOptionalCapabilities`
Remove `optionalCapabilitiesSupplier` from `EngineCapabilitiesMonitor`

StefanBratanov:
## PR Description
I think it is cleaner if the `EngineJsonRpcMethodsResolver` only deals with consensus criticial Engine API methods. Since `engine_getBlobsV1` usage is optional (similar to `engine_getClientVersionV1`), it makes sense to implement it in a similar way. It is unlikely the `BlobSchema` will change with upcoming forks and if it does, we can tackle it then. In the meantime, we can avoid some added complexity.

## Fixed Issue(s)
Fixes #8669 

## Documentation

## Changelog


tbenr:
related to https://github.com/Consensys/teku/issues/8669
this may close that

tbenr (approved):
LGTM. Pinging @mehdi-aouadi for an extra check since he worked on this last time.

mehdi-aouadi (approved):
LGTM

PR #8962, opened 2024-12-31
Reduce response chunk arrival timeout

StefanBratanov:
## PR Description
As part of https://github.com/Consensys/teku/pull/8839, reintroduced the timeouts as constants, but made the response chunk arrival timeout to 30s instead of 10s. Raising this PR to align this timeout. Potentially, the timeouts could be configured per method when needed so. (maybe with introduction of PeerDAS etc)

## Fixed Issue(s)
related to #8803

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #8980, opened 2025-01-09
Exit Teku if any service fails to start

zilm13:
## PR Description
**Current behavior:** Exit if any of these services cannot bind port
**New behavior:** Exit if any of this services fails to start:
- Attestation Manager
- P2P Manager (gossip)
- Block Manager
- Sync Service
- Merge Block Monitor

So my arguments are: Teku is not really alive if any of these services fails to start. If we keep it running, we keep zombie, which confuses users and complicates automation. We'd better quit in this case

PS: it will require some test changes, will update after discussion consensus

## Documentation

## Changelog

### DIFF
--- a/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
+++ b/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
@@ -65,6 +65,7 @@ import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.async.eventthread.AsyncRunnerEventThread;
 import tech.pegasys.teku.infrastructure.collections.LimitedMap;
 import tech.pegasys.teku.infrastructure.events.EventChannels;
+import tech.pegasys.teku.infrastructure.exceptions.ExceptionUtil;
 import tech.pegasys.teku.infrastructure.exceptions.InvalidConfigurationException;
 import tech.pegasys.teku.infrastructure.io.PortAvailability;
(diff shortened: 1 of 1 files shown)


lucassaldanha:
startServices() is called as part of the chain:
```
SafeFuture<?> doStart() -> SafeFuture<?> initialize() -> void startServices()
```
I am just wondering if a hard exit at that point could prevent other services etc of a clean shutdown. Not sure though, need to dig a bit.

Nashatyrev:
We started this discussion with @zilm13 at the point where the `tech.pegasys.teku.config.TekuConfigurationTest` failed after this change. 
It failed because services were still starting while shutdown was initiated. 
The shutdown was initiated that early because `doStart()` future has been completed but `startServices()` was not yet executed/completed.
This led me to the idea that `doStart()` future should depend on `startServices()` async task. Which from my perspective makes sense cause `Service.start()` future completion implies that all necessary 'sub-services' are up and running at this point. 

As for calling `System.exit()` inside a service startup procedure it rather looks like a workaround which is a candidate for addressing now. 

I would suggest for `startServices()` to return a future, bind it to `doStart()` future and invoke `System.exit()` on some very upper level (e.g. in `Teku` or `BeaconNode`) when the `Service.start()` future completes with some specific (or not) exception.

rolfyone:
I think for now we'll close this one. I've done some investigation and this is in the realm of redesign, and I agree with Anton/Lucas about the flows.

IIUC the zombies should be closed correctly when the parent process (teku) dies, and so we can live with the current startup messiness unless / until we refactor BeaconChainController...

PR #8981, opened 2025-01-10
BeaconChainController.start() future should complete when all services started

Nashatyrev:
## PR Description

PR addresses the issue described in comment: https://github.com/Consensys/teku/pull/8980#issuecomment-2581913923

1. `BeaconChainController.start()` returned `Future` should complete when all 'sub-services' are up and running
2. Move abrupt `System.exit()` to the upper level from the `BeaconChainController` to have more control on startup errors (for example in unit tests)  

## Documentation

## Changelog


Nashatyrev:
Oh, it looks like that doesn't work with pre-genesis start. Services are expected to start after genesis only. 
Then these changes are not valid :(

PR #8982, opened 2025-01-10
Changes in forward sync and rate limit constants

### LINKED ISSUE #8871: Revisit forward sync constants
`DEFAULT_FORWARD_SYNC_BATCH_SIZE` = 50
considering that this number has been set pre-merge when beacon blocks were small and state transition was fast, maybe we should lower this down to `25`. moreover we now have blobs, so the batch implies a 50*6 (that will be 8 or 9 for ELECTRA) blobs in the batch. 

`DEFAULT_FORWARD_SYNC_MAX_BLOCKS_PER_MINUTE` = 500
this is even more important because it implies a blob limit 500*6 (that will be 8 or 9 for ELECTRA), which are a lot of objects per minute, which might be the reason why we got rate limited in the NFT devnet.
I think the best approach would be to add a specific  `PER_MINUTE` constant for blobs. So we don't need to lower down the block value too much. The rational is to avoid impacting sync performance when syncing outside the data availability window (18gg), where we don't try do download blobs at all.

StefanBratanov:
## PR Description
- Changed `DEFAULT_FORWARD_SYNC_BATCH_SIZE` to 25 instead of 50 to cater for the time for validating blobs, state transition, etc
- Removed `-Xp2p-sync-rate-limit` in favour of `-Xp2p-sync-blocks-rate-limit` and `-Xp2p-sync-blob-sidecars-rate-limit`
- Removed `--Xpeer-rate-limit` in favour of `--Xpeer-blocks-rate-limit` and `--Xpeer-blob-sidecars-rate-limit`
- Set `DEFAULT_FORWARD_SYNC_MAX_BLOB_SIDECARS_PER_MINUTE` to 2000 (outgoing rate limit). The previous `500 * maxBlobsPerBlock` seems too large.
- Set `DEFAULT_PEER_BLOB_SIDECARS_RATE_LIMIT` to 2000 (incoming rate limit). The previous `500 * maxBlobsPerBlock` seems too large.
- Increased `DEFAULT_PEER_REQUEST_LIMIT` to 100 from 50, because it hasn't been changed after Deneb. It makes sense to double to account for blobs requests when syncing.

2000 Blobs (TBD) is ~250 MB per minute per peer maximum allowed

Keep in mind that `SyncSourceFactory` uses the formula:
```
maxBlobSidecarsPerMinute - (batchSize * maxBlobsPerBlock) - 1)
```

So essentially `2000` turns into `1849` (pre-Electra) and `1774` (after Electra is scheduled)

## Fixed Issue(s)
fixes #8871 

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
 ## Unreleased Changes
 
 ### Breaking Changes
-`--Xvalidators-builder-registration-default-gas-limit` is removed in favour of `--validators-builder-registration-default-gas-limit`
+- `--Xvalidators-builder-registration-default-gas-limit` is removed in favour of `--validators-builder-registration-default-gas-limit`
```
rolfyone: maybe we should list `-Xp2p-sync-rate-limit` as no longer available just in case anyone uses it. could potentially reference this PR removing it.... not sure anyone is using it but it'll be confusing if they are...

StefanBratanov: Yeah, good point, added it to CHANGELOG. Not sure how diligent we should be with `--X` option since users should be aware they could be changed at any point, but I suppose it is good to inform the users anyways.

tbenr (approved):
LGTM

PR #8984, opened 2025-01-11
Fix `ThrottlingSyncSource`

StefanBratanov:
## PR Description
In `ThrottlingSyncSource`, in `requestBlobSidecarsByRange`, we treat count as the number of blob sidecars but it is the number of slots. The number of blobs should be approximately `count * maxBlobsPerBlock`. This means that we don't limit nearly as much as we should when we forward sync. Also, we don't make use of `adjustObjectsRequest` which can account for missed slots or less blobs than the max. It makes sense to align this with how we handle incoming RPC requests and to avoid unnecessarily limiting outgoing requests in case of lots of missed slots in a row.

Also removed `ResponseStream` which only job is to provide `getResponseChunkCount()` which is not used anywhere, so got rid of it to avoid unnecessary delegation.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
```diff
@@ -44,15 +44,20 @@ public SyncSourceFactory(
   public SyncSource getOrCreateSyncSource(final Eth2Peer peer, final Spec spec) {
     // Limit request rate to just a little under what we'd accept
     final int maxBlocksPerMinute = this.maxBlocksPerMinute - batchSize - 1;
+    final Optional<Integer> maybeMaxBlobsPerBlock = spec.getMaxBlobsPerBlockForHighestMilestone();
```
rolfyone, line 47: so if electra goes to 9 and hasn't been activated we'd still use electra's max blobs?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -31,23 +33,29 @@
 
 public class ThrottlingSyncSource implements SyncSource {
   private static final Logger LOG = LogManager.getLogger();
+
+  // 1 minute
```
rolfyone: i'd probably prefer the const was 'time_out_seconds' than a floating comment like this...

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -61,33 +69,64 @@ public SafeFuture<Void> requestBlocksByRange(
...
-      return delegate.requestBlobSidecarsByRange(startSlot, count, listener);
-    } else {
-      LOG.debug(
-          "Rate limiting request for {} blob sidecars. Retry in {} seconds",
-          count,
-          PEER_REQUEST_DELAY.toSeconds());
-      return asyncRunner.runAfterDelay(
-          () -> requestBlobSidecarsByRange(startSlot, count, listener), PEER_REQUEST_DELAY);
-    }
+    long blobSidecarsCount =
```
rolfyone: final

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -61,33 +69,64 @@ public SafeFuture<Void> requestBlocksByRange(
...
+        maybeMaxBlobsPerBlock
+            .map(maxBlobsPerBlock -> maxBlobsPerBlock * count.longValue())
+            .orElse(0L);
+    return blobSidecarsRateTracker
+        .approveObjectsRequest(blobSidecarsCount)
+        .map(
+            requestApproval -> {
+              LOG.debug("Sending request for ~ {} blob sidecars", blobSidecarsCount);
```
rolfyone: remove the `~` probably reads better

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -61,33 +69,64 @@ public SafeFuture<Void> requestBlocksByRange(
...
+                      () ->
+                          // adjust for slots with empty blocks and slots with blobs <
+                          // maxBlobsPerBlock
+                          blobSidecarsRateTracker.adjustObjectsRequest(
+                              requestApproval, listenerWithCount.count.get()));
+            })
+        .orElseGet(
+            () -> {
+              LOG.debug(
+                  "Rate limiting request for ~ {} blob sidecars. Retry in {} seconds",
```
rolfyone: remove `~`

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSourceTest.java
```diff
@@ -101,32 +142,42 @@ void shouldDelayRequestIfBlockLimitAlreadyExceeded() {
...
 
   @Test
   void shouldDelayRequestIfBlobSidecarsLimitAlreadyExceeded() {
-    final UInt64 count = UInt64.valueOf(MAX_BLOB_SIDECARS_PER_MINUTE);
+    // 17 * 6 = 102 > 100
+    final UInt64 count = UInt64.valueOf(17);
     ignoreFuture(source.requestBlobSidecarsByRange(UInt64.ZERO, count, blobSidecarsListener));
     ignoreFuture(
-        source.requestBlobSidecarsByRange(UInt64.valueOf(100), count, blobSidecarsListener));
+        source.requestBlobSidecarsByRange(UInt64.valueOf(17), count, blobSidecarsListener));
```
rolfyone: shouldnt we just re-use count instead of `uint64.valueOf(17)`?

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSourceTest.java
```diff
@@ -75,23 +106,33 @@ void shouldRequestBlocksImmediatelyIfRateLimitNotExceeded() {
...
 
   @Test
   void shouldRequestBlobSidecarsImmediatelyIfRateLimitNotExceeded() {
-    final UInt64 count = UInt64.valueOf(MAX_BLOB_SIDECARS_PER_MINUTE - 1);
+    // 100 / 6 = 16, 16 * 6 = 92 < 100
+    final UInt64 count = UInt64.valueOf(16);
     ignoreFuture(source.requestBlobSidecarsByRange(UInt64.ZERO, count, blobSidecarsListener));
     ignoreFuture(
-        source.requestBlobSidecarsByRange(UInt64.valueOf(100), count, blobSidecarsListener));
+        source.requestBlobSidecarsByRange(UInt64.valueOf(16), count, blobSidecarsListener));
```
rolfyone: same here, i think we should be using count if we've computed it anyway... given this shows it changing so substantially we'll need to be super careful that we've fixed everywhere that uses `requestBlobSidecarsByRange`

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
StefanBratanov: Changed ~ to approximately since it is not exact number

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSourceTest.java
StefanBratanov: yeah, good point, `count` is more cleaner

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
StefanBratanov: Yeah, this one I have to talk about Enrico what is the best strategy. We pre-initialize the RateTracker on startup, so it's not dynamic per milestone.

tbenr: ye this is a little bit of a pain. At the end in this cases I think it is an acceptable side-effect but we could be rate-limited more during the transition period. We could go this simplified way and rise an issue describing the issue, so we can later think about more correct ways

tbenr: ..... we could constantly track the current maxBlobsPerBlock and if changes we could reinitialize RateTracker . Maybe this approach is simple enough to be adopted (elsewhere too).

another one could be to have these type of spec-related objects somewhere in the spec and could be looked up like `spec.at(slot).getSpecRelatedObjects().getBlobSidecarRateTracker()`

with the "new specConfig" behaviour we could even expose this config-related objects from the config itself

tbenr (approved):
LGTM nice catch, we could merge as is and eventually address the Tracker initialization problem for a separate PR.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/SyncSourceFactory.java
StefanBratanov: I think there is not that much benefit, keep in mind we `adjust` the number after getting `x` number of blobs and we are the default blobs per minute will still be the same for future upgrades, so all in all not much reason to complicate the code for little benefit in my opinion

StefanBratanov: In the DefaultEth2Peer, the RateTracker just use the defaults. Only in `SyncSourceFactory`, we modify the number slightly to make sure we don't rate limit ourselves.

PR #8985, opened 2025-01-13
change StateSelector logic

### LINKED ISSUE #7955: Finalized state from `debug/state` API is incorrect on epochs with empty slots in the end

rolfyone:
fixes #7955

alternative solution to #8782 

## Documentation

## Changelog


rolfyone:
also moved some private functions to the bottom (breaks my own rule of not refactoring). The new functionality was a finalized selector similar to teh justified flow in CombinedChainDataClient, and a `getBestFinalizedState` which essentially determines the finalized slot to refer to in the same way `on_block` does.

data/provider/src/main/java/tech/pegasys/teku/api/stateselector/StateSelectorFactory.java
```diff
@@ -80,19 +80,13 @@ public StateSelector genesisSelector() {
...
-        SafeFuture.completedFuture(
-            client
-                .getLatestFinalized()
-                .map(
-                    finalized ->
-                        addMetaData(
-                            finalized.getState(),
-                            // The finalized checkpoint may change because of optimistically
-                            // imported blocks at the head and if the head isn't optimistic, the
-                            // finalized block can't be optimistic.
```
rolfyone, line 92: there was nothing testing this.

i went with the on_block implementation of the finalized state where we base it off of the checkpoint.

lucassaldanha (approved):
LGTM

PR #8991, opened 2025-01-14
Added validator_next_attestation_slot metric

### LINKED ISSUE #8795: Create metric for next attestation duty slot
### Context
Rocket Pool Grafana dashboard has a panel to show your validator's next attestation duty. This can help users to choose a good time to update their node or do any other maintenance tasks.

It probably does not help much if you are running heaps of validators, but it can be helpful for solo stakers running a hand full of validators.

### Description
We can probably use a labelled gauge. Using the validator id/pubkey as a label. We should also consider if we want to have similar metrics for block publishing and sync commitees.

### References
In Lighthouse the metric is called `vc_attestation_duty_slot`.
In Prysm the metric is called `validator_next_attestation_slot`.

lucassaldanha:
## PR Description

Created a new metric `validator_next_attestation_slot` gauge which value corresponds to the next earliest slot any validator running in the node is expected to publish an attestation.

This is particularly helpful for home stakers running a small number of validators, helping them to find a good gap for being offline without missing attestation rewards. 

## Fixed Issue(s)
fixes #8795

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/SlotBasedScheduledDuties.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/SlotBasedScheduledDuties.java
@@ -13,7 +13,9 @@
 
 package tech.pegasys.teku.validator.client.duties;
 
+import java.util.Map.Entry;
 import java.util.NavigableMap;
+import java.util.Optional;
 import java.util.TreeMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


gfukushima (approved):
lgtm

validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/SlotBasedScheduledDuties.java
```diff
@@ -43,6 +46,14 @@ public SlotBasedScheduledDuties(
     this.dutyFunction = dutyFunction;
   }
 
+  public Optional<UInt64> getNextAttestationProductionDutyScheduledSlot() {
+    return productionDuties.entrySet().stream()
+        .filter(e -> e.getValue().getType() == DutyType.ATTESTATION_PRODUCTION)
+        .map(Entry::getKey)
+        .sorted()
+        .findFirst();
+  }
```
rolfyone: just so we have a brief note, we talked about ways to make this less expensive, we should do that before we merge..

rolfyone (approved):

PR #8994, opened 2025-01-14
Introduce SSZ support in Builder API

### LINKED ISSUE #8746: Builder API: Add SSZ support to Builder API
Reference: https://github.com/ethereum/builder-specs/pull/104

tbenr:
Expands the Builder Rest client to handle ssz in `getHeader` and `postPayload` methods.
The general idea is that in `getHeader` we signal via `Accept` header that we support `ssz` and `json` (with `ssz` preference).
If we receive the header in `ssz`, we remember this and on the upcoming `postPayload` we will post the signed blinded block in `ssz` format.

possible additional things:
- (optional) use the new `isUnsupportedMediaTypeError` in `Response` to try to fallback to json in postPayload (maybe not really important)
- (optional) try to post validator registration in `ssz` too. At immediately fallback to `JSON` (and remain in `JSON`) if the first call fails for whatever reason (independently by 415, which could be not implemented on server side). If the first call succeed we could keep sending `ssz` (not falling back on generic error). This assumes that the relay doesn't change in the mean time from one supporting `ssz` to another not supporting `ssz`. But this would be out of spec RN.

## Fixed Issue(s)

fixes #8746

## Documentation

## Changelog


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/schema/Response.java
```diff
@@ -13,51 +13,45 @@
...
-    this.payload = payload;
-    this.errorMessage = errorMessage;
+  public static <T> Response<T> fromPayloadReceivedAsSsz(final T payload) {
+    return new Response<>(payload, null, true, false);
   }
 
-  public Response(final T payload) {
-    this.payload = payload;
-    this.errorMessage = null;
+  public static <T> Response<T> fromPayloadReceivedAsJson(final T payload) {
```
rolfyone, line 34: maybe better just as `fromJsonPayload`

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/schema/Response.java
```diff
@@ -13,51 +13,45 @@
...
-  private final T payload;
-  private final String errorMessage;
+  public static <T> Response<T> fromNullPayload() {
+    return new Response<>(null, null, false, false);
+  }
 
-  public Response(final T payload, final String errorMessage) {
-    this.payload = payload;
-    this.errorMessage = errorMessage;
+  public static <T> Response<T> fromPayloadReceivedAsSsz(final T payload) {
```
rolfyone, line 30: maybe `fromSszPayload`

ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
```diff
@@ -177,7 +177,7 @@ void registerValidators_success() {
         .satisfies(
             response -> {
               assertThat(response.isSuccess()).isTrue();
-              assertThat(response.getPayload()).isNull();
```
rolfyone, line 180: i think i prefer getter naming...

lucassaldanha: I don't think there is value in customizing getter methods for Record classes. By default they will have the field name so in this case I'd keep it as `payload()`.

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java
```diff
@@ -143,63 +144,36 @@ private URL createHttpUrl(final String apiPath) {
...
+    call.enqueue(responseCallback);
+    return responseHandler.getFutureResponse();
   }
 
-  private String getErrorMessageForFailedResponse(final okhttp3.Response response)
-      throws IOException {
-    try (final ResponseBody responseBody = response.body()) {
-      if (bodyIsEmpty(responseBody)) {
-        return response.code() + ": " + response.message();
+  private Callback createResponeCallback(
```
lucassaldanha: `s/createResponeCallback/createResponseCallback`

lucassaldanha:
> to verify that we try to post payload in ssz when only when we receive the header in ssz

we also need to test that we correctly "revert" back to JSON if the getHeader response changes to JSON (e.g. builder software gets downgraded or some config changes).

ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
tbenr: yep, that's the record style

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/schema/Response.java
tbenr: but technically the payload has been already decoded at this stage. It indicates how you received the payload rather than the format of the payload passed here. I have mixed feelings

StefanBratanov (approved):
LGTM

PR #9000, opened 2025-01-15
feat: add block signature verification in EraFile

crStiv:
Implements block signature verification that was previously marked as TODO:

- Added block signature verification in verifyBlocksWithReferenceState method
- Validates signatures using proposer's public key and signing root
- Throws IllegalArgumentException for invalid signatures or missing public keys

This change enhances security by ensuring cryptographic verification of block signatures.


rolfyone:
This is just POC that we have no intention of expanding until we use it in a feature.

Closing, as fails spotless. If you're looking for good first issues, i'd suggest either asking on discord for things we need, or looking at issues tagged 'good first issue'

PR #9010, opened 2025-01-17
Fix ValidatorIndexProviderTest for Apple M4

zilm13:
## PR Description

I assume that order is not guaranteed here and it was working because of good luck (read: java machine implementations). Alternatively we could add sorting in provider if it makes a sense.

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM
Was passing fine with corretto-22 but fails with corretto-23

PR #9011, opened 2025-01-17
Fix builder http leak

tbenr:
fixes:

```
2025-01-17 12:18:32 WARNING: A connection to http://mev-boost-2-teku-geth:18550/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
```

## Documentation

## Changelog


StefanBratanov (approved):
As another case, I have a feeling in `handleResponseError`, we still need to close the body when `response.code == SC_UNSUPPORTED_MEDIA_TYPE`

PR #9017, opened 2025-01-21
Added block production restrictions for electra

rolfyone:
voluntary exits should not be included in a block if
 - they have a withdrawal pending (thats invalid)

The pending withdrawal was pointed out in devnet5, so this resolves that.

## Documentation

## Changelog


rolfyone:
i was convinced that deposit queue is not a problem so removed that check.

tbenr (approved):
LGTM

PR #9019, opened 2025-01-21
Handle overflows when validating BlobSidecarsByRange requests

StefanBratanov:
## PR Description
Fixes edge cases where `maxBlobsPerBlock * message.getCount().longValue()` could overflow and be negative or when `message.getCount.longValue()` is negative.

## Fixed Issue(s)
fixes https://github.com/Consensys/teku-internal/issues/176

## Documentation

## Changelog


tbenr (commented):
wander if we could use `.intValue()` (which has overflow checks) and consider `requestedCount` as `int`

tbenr (approved):
After offline discussions LGTM.
We could address `longValue()` safety on a dedicated PR if we want to

PR #9020, opened 2025-01-21
Handle edge case when validating BlobSidecarsByRange requests

StefanBratanov:
## PR Description
Addition to https://github.com/Consensys/teku/pull/9019

Handles edge case, where a large number can overflow and the requested calculation can become a positive number, e.g.

`( (Long.MAX_VALUE / 3) + 100 ) * 6 = 596`

## Fixed Issue(s)
related to https://github.com/Consensys/teku-internal/issues/176

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
@@ -128,7 +128,7 @@ public class BlobSidecarsByRangeMessageHandler
         startSlot);
 
     final SpecConfigDeneb specConfig = SpecConfigDeneb.required(spec.atSlot(endSlot).getConfig());
-    final long requestedCount = calculateRequestedCount(message, specConfig.getMaxBlobsPerBlock());
+    final int requestedCount = calculateRequestedCount(message, specConfig.getMaxBlobsPerBlock());
 
     final Optional<RequestApproval> blobSidecarsRequestApproval =
         peer.approveBlobSidecarsRequest(callback, requestedCount);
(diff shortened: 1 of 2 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
```diff
@@ -91,10 +91,9 @@ public Optional<RpcException> validateRequest(
     final int maxRequestBlobSidecars = specConfig.getMaxRequestBlobSidecars();
     final int maxBlobsPerBlock = specConfig.getMaxBlobsPerBlock();
 
-    final long requestedCount = calculateRequestedCount(request, maxBlobsPerBlock);
+    final int requestedCount = calculateRequestedCount(request, maxBlobsPerBlock);
```
lucassaldanha: NIT: I think we don't need another method just to handle the exception and return -1. It kinda hides that we are using -1 as a "flag" for ArithimeticException.

Here is a suggestion (and we can remove the `calculateRequestedCount` method.

```suggestion
    int requestedCount;
    try {
      requestedCount = request.getCount().times(maxBlobsPerBlock).intValue();
    } catch (final ArithmeticException e) {
      requestedCount = -1;
    }
```

lucassaldanha (approved):
LGTM. Just a nit comment.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BlobSidecarsByRangeMessageHandler.java
StefanBratanov: Implemented this. I like it better. I didn't remove the method since it is used in `onIncomingMessage` but it will never fail there, since the message would have been rejected anyways.

jtraglia:
This was a good find. Maybe we could an errorprone check ([to this repo](https://github.com/Consensys/errorprone-checks)) for this pattern.

```
(int|long) * UInt64.(int|long)Value()
```

StefanBratanov:
> This was a good find. Maybe we could an errorprone check ([to this repo](https://github.com/Consensys/errorprone-checks)) for this pattern.
> 
> ```
> (int|long) * UInt64.(int|long)Value()
> ```

I can see this being useful! Thank you for the idea. And for transparency, this was found as part of the Immunefi Attackathon.

PR #9025, opened 2025-01-22
Cache SlashingProtectedSigner instances for performance optimization

VolodymyrBg:
## PR Description
Implements a caching mechanism for `SlashingProtectedSigner` instances in `SlashingProtectedValidatorSource` to ensure that only one instance is created per public key. This optimization:
- Prevents creation of redundant signer instances
- Guarantees consistent signer usage per key
- Improves performance by caching signers in a thread-safe manner using `ConcurrentHashMap`

## Fixed Issue(s)
Fixes the TODO comment in `SlashingProtectedValidatorSource.java` that suggested caching signer instances.

## Documentation
- [x] No documentation updates required as this is an internal performance optimization that doesn't affect the public API or user behavior.

## Changelog
- [] Added changelog entry:

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/SlashingProtectedValidatorSource.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/SlashingProtectedValidatorSource.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.validator.client.loader;
 import java.net.URL;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


rolfyone:
Thanks for the submission.

I think we'd need a lot of careful planning and Im not sure I completely agree with the TODO we're replacing here. I may actually remove the todo for now.

Between this consideration, it failing spotless, and having no tests or anything, I'll close this PR for now, but appreciate attention being brought. Please feel free to talk in teku-contribs on discord if you're looking for ways to contribute to the project.

PR #9026, opened 2025-01-22
Write latest finalized state in chunks

### LINKED ISSUE #9018: Support serializing state in chunks when saving it on DB
When we write latest finalized state via `setLatestFinalizedState`, we serialize the entire state in memory. The state is growing and we should start trying to avoid it.

```
java.util.concurrent.CompletionException: java.lang.OutOfMemoryError: Java heap space
    at java.base/java.util.concurrent.CompletableFuture.reportJoin(CompletableFuture.java:413)
    at java.base/java.util.concurrent.CompletableFuture.join(CompletableFuture.java:2118)
    at tech.pegasys.teku.storage.server.RetryingStorageUpdateChannel.retry(RetryingStorageUpdateChannel.java:133)
    at tech.pegasys.teku.storage.server.RetryingStorageUpdateChannel.onStorageUpdate(RetryingStorageUpdateChannel.java:86)
    at tech.pegasys.teku.storage.server.CombinedStorageChannelSplitter.onStorageUpdate(CombinedStorageChannelSplitter.java:68)
    at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
    at java.base/java.lang.reflect.Method.invoke(Method.java:580)
    at tech.pegasys.teku.infrastructure.events.DirectEventDeliverer.executeMethod(DirectEventDeliverer.java:74)
    at tech.pegasys.teku.infrastructure.events.DirectEventDeliverer.deliverToWithResponse(DirectEventDeliverer.java:67)
    at tech.pegasys.teku.infrastructure.events.AsyncEventDeliverer.lambda$deliverToWithResponse$1(AsyncEventDeliverer.java:80)
    at tech.pegasys.teku.infrastructure.events.AsyncEventDeliverer$QueueReader.deliverNextEvent(AsyncEventDeliverer.java:125)
    at tech.pegasys.teku.infrastructure.events.AsyncEventDeliverer$QueueReader.run(AsyncEventDeliverer.java:116)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
    at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.OutOfMemoryError: Java heap space
    at tech.pegasys.teku.infrastructure.ssz.sos.SszByteArrayWriter.<init>(SszByteArrayWriter.java:23)
    at tech.pegasys.teku.infrastructure.ssz.schema.SszType.sszSerializeTree(SszType.java:64)
    at tech.pegasys.teku.infrastructure.ssz.SszData.sszSerialize(SszData.java:58)
    at tech.pegasys.teku.storage.server.kvstore.serialization.BeaconStateSerializer.serialize(BeaconStateSerializer.java:36)
    at tech.pegasys.teku.storage.server.kvstore.serialization.BeaconStateSerializer.serialize(BeaconStateSerializer.java:21)
    at tech.pegasys.teku.storage.server.leveldb.LevelDbTransaction.put(LevelDbTransaction.java:50)
    at tech.pegasys.teku.storage.server.kvstore.dataaccess.CombinedKvStoreDao$V4CombinedUpdater.setLatestFinalizedState(CombinedKvStoreDao.java:576)
    at tech.pegasys.teku.storage.server.kvstore.KvStoreDatabase$$Lambda/0x00007c47949fe8b0.accept(Unknown Source)
    at java.base/java.util.Optional.ifPresent(Optional.java:178)
    at tech.pegasys.teku.storage.server.kvstore.KvStoreDatabase.doUpdate(KvStoreDatabase.java:1127)
    at tech.pegasys.teku.storage.server.kvstore.KvStoreDatabase.update(KvStoreDatabase.java:646)
```

current mainnet mem reuirement is 246MB

Image

The underlining DB system (levelDB at least) doesn't support streaming data into DB.
An approach would be to implement a `KvStoreVariable` that supports having variable stored in chunks, in a similar way to how we deal with "tables" in level db. This way we can have at least the state stored in chunks and not pretending to have a 246MiB contiguous memory, which it should not trigger big heap compactions

An example: https://github.com/andris9/level-stream-access/blob/master/lib/level-stream-access.js

So we could have chunks of (ie 25 MBs) and leverage the `int sszSerialize(final SszWriter writer)` interface and implementing a chunking `KvStoreSerializer` and `BeaconStateSerializer`.

tbenr:
- serializes beaconState into chunks via `SszByteArrayChunksWriter` (currently hardcoded to 8MiB chunk sizes)
- on DB side introduced a `KvStoreUnchunkedVariable` which stores chunks in a similarly to what we do with columns:
  - the base key `(id)->byte` stores a byte with the number of chunks (max 255 chunks)
  - chunks are stored as`(id,chunkIndex) -> chunk bytes`

There is an automatic backward compatibility behavior: if the base `(id)->byte` returns something bigger than a single byte, than it is considered as a big chunk (essentially the previous storage mode)

TODO:
- verify that this actually improves memory allocations (no more big 256mb single byte array)
- implement an iterator when reading chunks
- trimming: if the variable becomes smaller, previous additional chunks remains on DB (we could not care for BeaconState, it always grows)
- rocksDB implementation
- tests

fixes https://github.com/Consensys/teku/issues/9018

## Documentation

## Changelog

### DIFF
--- a/docker/jdk21/Dockerfile
+++ b/docker/jdk21/Dockerfile
@@ -1,7 +1,7 @@
 FROM eclipse-temurin:21 as jre-build
 
 # Create a custom Java runtime
-RUN JAVA_TOOL_OPTIONS="-Djdk.lang.Process.launchMechanism=vfork" $JAVA_HOME/bin/jlink \
+RUN JAVA_TOOL_OPTIONS="-Djdk.lang.Process.launchMechanism=vfork -XX:UseSVE=0" $JAVA_HOME/bin/jlink \
          --add-modules ALL-MODULE-PATH \
          --strip-debug \
          --no-man-pages \
--- a/docker/jdk21/Dockerfile
+++ b/docker/jdk21/Dockerfile
@@ -1,7 +1,7 @@
 FROM eclipse-temurin:21 as jre-build
 
 # Create a custom Java runtime
-RUN JAVA_TOOL_OPTIONS="-Djdk.lang.Process.launchMechanism=vfork -XX:UseSVE=0" $JAVA_HOME/bin/jlink \
+RUN JAVA_TOOL_OPTIONS="-Djdk.lang.Process.launchMechanism=vfork" $JAVA_HOME/bin/jlink \
          --add-modules ALL-MODULE-PATH \
          --strip-debug \
          --no-man-pages \
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/sos/SszByteArrayChunksWriter.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/sos/SszByteArrayChunksWriter.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright Consensys Software Inc., 2022
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.infrastructure.ssz.sos;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class SszByteArrayChunksWriter implements SszWriter {
+  private final byte[][] chunks;
+  private final int maxChunkSize;
+  private final int maxChunks;
+
+  private int lastChunkIndex;
+  private int size = 0;
+
+  public SszByteArrayChunksWriter(final int maxSize, final int maxChunkSize) {
+    this.maxChunks = (maxSize + maxChunkSize - 1) / maxChunkSize;
+    this.chunks = new byte[maxChunks][];
+    this.maxChunkSize = maxChunkSize;
+  }
+
+  @Override
+  public void write(final byte[] bytes, final int offset, final int length) {
+    final int chunk = this.size / maxChunkSize;
+    final int chunkOffset = this.size % maxChunkSize;
+    final byte[] chunkData = getChunk(chunk);
+
+    if (chunkOffset + length > maxChunkSize) {
+      final int lengthToFillCurrentChunk = maxChunkSize - chunkOffset;
+      System.arraycopy(bytes, offset, chunkData, chunkOffset, lengthToFillCurrentChunk);
+      this.size += lengthToFillCurrentChunk;
+
+      write(bytes, offset + lengthToFillCurrentChunk, length - lengthToFillCurrentChunk);
+    } else {
+      System.arraycopy(bytes, offset, chunkData, chunkOffset, length);
+      this.size += length;
+    }
+  }
+
+  private byte[] getChunk(final int chunk) {
+    if (chunk >= maxChunks) {
+      throw new IndexOutOfBoundsException(
+          "Chunk index out of bounds: " + chunk + ", max chunks: " + maxChunks);
+    }
+    byte[] chunkData = chunks[chunk];
+    if (chunkData == null) {
+      chunkData = new byte[maxChunkSize];
+      chunks[chunk] = chunkData;
+      lastChunkIndex = chunk;
+    }
+    return chunkData;
+  }
+
+  public List<byte[]> getChunks() {
+    int lastChunkSize = size % maxChunkSize;
+    if (lastChunkSize != 0) {
+      byte[] lastChunk = chunks[lastChunkIndex];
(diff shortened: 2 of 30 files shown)


Nashatyrev:
Looks like a great idea! 

However: 
- When you write you still keep all chunks in memory when serializing 
- Not sure how would you implement (non-blocking) reading cause chunk boundaries would be at random places? 

From my understanding the ideal approach would be to make `SszWriter/Reader` asynchronous so they could flush/read chunks on the fly. But I'm not sure how much overhead it would introduce as we would need to return and handle `SafeFuture` on every tiny `read/write` operation.

tbenr:
@Nashatyrev
I hear you on those points. I did in that way due to the fact that our current interface (at least towards LevelDB) requires all data to be there in memory for a given transaction. It boils down to usage of `org.iq80.leveldb.WriteBatch`.
If we breaks down the write in multiple transaction it would work but we loose atomicity.
I'll doublecheck again and will look into rocksDB if it has a way of stream data into transactions.

Nashatyrev:
If you only care about atomicity of a single state record then it could be the option to stream all chunks besides the first one and write the first chunk just at the very end. So you would either have committed  the first chunk and _all_ others or you would have no first chunk which would be treated as absence of the state. 
Does it make any sense?

tbenr:
well, it is not only atomicity in writing the state. We update that state within a bigger transaction, which contains several updates.

https://github.com/Consensys/teku/blob/2550254549fe6146512b9bdd7999ab60f0e770b4/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java#L1110

Nashatyrev:
Could it be generalized to such a statement: if there are `KvStoreChunkedVariable`s update in a batch then all of them should be written first (without first chunks), and then only first chunks of them are included into the final batch. 
How do you think would it work?

storage/src/main/java/tech/pegasys/teku/storage/server/leveldb/CustomJniDBFactory.java
```diff
@@ -0,0 +1,151 @@
...
+              @Override
+              public void log(final String message) {
+                userLogger.log(message);
+              }
+            };
+        this.options.infoLog(this.logger);
+      }
+    }
+
+    public void close() {
```
rolfyone: `private`

storage/src/main/java/tech/pegasys/teku/storage/server/leveldb/CustomJniDBFactory.java
```diff
@@ -0,0 +1,151 @@
...
+    NativeLogger logger;
+    NativeOptions options;
+
+    private OptionsResourceHolder() {
+      this.cache = null;
+      this.comparator = null;
+      this.logger = null;
+    }
+
+    public void init(final Options value) {
```
rolfyone: `private`

tbenr:
> Could it be generalized to such a statement: if there are KvStoreChunkedVariables update in a batch then all of them should be written first (without first chunks), and then only first chunks of them are included into the final batch.
How do you think would it work?

this would work from the atomicity standpoint, but it wont be transactional. Once we start writing there wont be any rollback. I.e. If the client shuts down in the middle of the sequence we loose the previous state too and client wont be able to recover.

An approach could be to have two version of a given variable (lets say A and B) and have an additional meta value tracking which is the "current" (updating this info only after all chunks have been written) and we will "ping-pong" between A and B on each write.

Doable (maybe complicated?), but if we will end up moving from levelDB and focus on RocksDB again as a primary DB, it could be a waste of time.

Nashatyrev:
Oh, right! I forgot it's the DB _variable_. 
I'm sure there is a suitable solution, but yeah it could make things complicated

tbenr:
The other problem is that we currently assume that the data in db (hot and finalized) is consistent as we update it in a big transaction (updating tables and variables all together).
If we split in multiple transactions we could break this consistency and thus break those assumptions.

Nashatyrev:
I would suggest the following approach for chunked variable transactions: 
1. You write and commit data chunks with keys `hash_root(new_data) + chunk_idx`
2. Execute a batch (transaction) which:
    - deletes old chunks referred by `hash_root(old_data)` 
    - sets the chunked variable reference to `hash_root(new_data)`
    - applies other transactional changes 

Looks pretty atomic and transactional to me 🤔 

The only drawback would be unmanaged data left in the DB if the transaction (steps 1+2) is aborted in the middle. The same could happen for concurrent transactions.  But that could be neglected due to rare occasions imo

tbenr:
yeah i like that model and it could be extended to values in columns too.
I'd like to dig into rocksDB. If it isn't possible to stream data into it as well, this could be a nice overall improvement.

tbenr:
we could even think about writing those chunks directly as files on FS and write the full path to them as values (instead of the actual data). This would turn the db to be just an index.

We could easily abstract this as `KvStoreSerializer`

This would even lead us to offload blobs on FS, pretty transparently

tbenr:
closing this for now. It remains a valid potential idea that could be revived in the future.

PR #9027, opened 2025-01-23
Add block signature verification in EraFile

VolodymyrBg:
# Add block signature verification in EraFile

## PR Description
Implements block signature verification in the `EraFile` class during block verification process. This adds an important security check to ensure that all blocks in the era file have valid signatures.

The change:
- Adds signature verification using the spec's `verifyBlockSignature` method
- Throws an exception with a clear error message if an invalid signature is detected
- Maintains the existing verification flow while adding this additional security check

## Fixed Issue(s)
Addresses TODO comment in `EraFile.java` that indicated missing signature verification.

## Documentation
- [x] No documentation updates required as this is an internal security enhancement that doesn't affect the public API.

## Changelog
- [ ] Added changelog entry:


rolfyone:
Thanks for the submission.
This is just POC code, we wont' be altering it until we use ERA fully.

Issue #3356, opened 2020-12-03
Support forks


rolfyone:
### Description
Refactor teku so that spec logic is versionable.  See prototype for direction: https://github.com/ConsenSys/teku/pull/3546

Tasks:
- [x] Investigate how to version datastructures (#3504)
- [ ] Migrate from static constants to instance-based constants (#3394)
- [ ] Migrate container definitions and datastructures into `spec` (#3564)
- [ ] Migrate all spec-related logic into `spec` package (#3596)
- [ ] Address any pending todos (search: `TODO(#3356)`)

rolfyone:
https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/HkTzLKOov

PR #9028, opened 2025-01-23
cleanup stale todo's

### LINKED ISSUE #7532: [libp2p] Don't use custom maxBufferedConnectionWrites for yamux
When `jvm-libp2p` release a new version with improved yamux backpressure, we should stop using the custom 150 MiB `maxBufferedConnectionWrites` and rely on the default value instead which should improve memory utilization overall.

rolfyone:
Removed the SlashingProtectedValidatorSource todo as its had several PRs from bots, and it's actually grossly over-simplifying.

Re-worded to remove TODO in EraFile, as its just a POC, and we won't be accepting any tiny changes there, the note was for when its rewritten.

Removed the TODO from SszCollectionSchema but left the content, as it does seem valid.

remove stale reference to a PR #3356
Fixes #7532

## Documentation

## Changelog

### DIFF
--- a/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/SszCollectionSchema.java
+++ b/infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/SszCollectionSchema.java
@@ -49,7 +49,6 @@ public interface SszCollectionSchema<
   }
 
   default TreeNode createTreeFromElements(final List<? extends SszElementT> elements) {
-    // TODO: probably suboptimal method implementation:
     // This is a generic implementation which works for both Vector and List but it potentially
     // could do better if construct the tree directly in List/Vector subclasses
     checkArgument(elements.size() <= getMaxLength(), "Too many elements for this collection type");
--- a/data/dataexchange/src/main/java/tech/pegasys/teku/data/eraFileFormat/EraFile.java
+++ b/data/dataexchange/src/main/java/tech/pegasys/teku/data/eraFileFormat/EraFile.java
@@ -214,7 +214,7 @@ public class EraFile {
             block.getParentRoot().equals(previousArchiveLastBlock.getRoot()),
             "First block in archive does not match last block of previous archive.");
       }
-      // TODO should verify signature
+      // when fully implemented, we would check signature also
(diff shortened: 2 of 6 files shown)


infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/SszCollectionSchema.java
```diff
@@ -49,7 +49,6 @@ default SszCollectionT createFromElements(final List<? extends SszElementT> elem
   }
 
   default TreeNode createTreeFromElements(final List<? extends SszElementT> elements) {
-    // TODO: probably suboptimal method implementation:
```
gfukushima, line 52: I'd leave the comment of this being prob a suboptimal impl

gfukushima (approved):

infrastructure/ssz/src/main/java/tech/pegasys/teku/infrastructure/ssz/schema/SszCollectionSchema.java
rolfyone: we can just re-visit that, we have lots of sub-optimal implementations of things IMO

StefanBratanov: Problem is how we will revisit it if the comment is gone. Possibly raise an issue?

rolfyone: Literally 80% of our code is in the same spot.... We typically don't write optimised code because the spec isn't optimised, and I dont think commenting every non-optimised section of code is going to leave readable code.

rolfyone: also please note the entire comment of the reasoning is there if you look at the next lines. its just the 'probably suboptimal' that is gone.
added more noise with https://github.com/Consensys/teku/issues/9035 to also track onthe board but i think where i left it was more than adequate. am marking this thread resolved.

PR #9029, opened 2025-01-24
Clean up blobsSidecar pruner workflow

gfukushima:
## PR Description
This is a quickly cleanup to make the creation of the updater only when necessary. Arguably we could create and commit this within the for loop to avoid holding the updater for too long.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #9030, opened 2025-01-24
debug command to see what values are in variables

rolfyone:
Probably incomplete, but the state variable was unset, and I was looking at what other values were in a broken database.

Probably worth keeping for next time.

## Documentation

## Changelog


gfukushima (approved):
lgtm

PR #9031, opened 2025-01-24
Reuse earliest blob slot

gfukushima:
## PR Description
Based on top of ##9029
This uses similar approach of the earliest block available. Reuse the variable value when present otherwise keep the current behaviour of creating a stream from 0.

This reduces the time that it takes for the blob pruner to run from 60 secs to ~1-2 secs on rocksdb.
`{"@timestamp":"2025-01-28T08:52:00,212","level":"DEBUG","thread":"storagePrunerAsyncRunner-async-1","class":"BlobSidecarPruner","message":"Pruning blobs up to slot 3381555, limit 12","throwable":""}
{"@timestamp":"2025-01-28T08:52:01,736","level":"INFO","thread":"validator-async-1","class":"teku-validator-log","message":"Validator   *** Published attestation        Count: 1, Slot: 3512660, Root: f70a2b57992861ea63f576d7d0cc5bf891d2d2c475cac54ae42e61d0b8ca71d5","throwable":""}
{"@timestamp":"2025-01-28T08:52:01,991","level":"DEBUG","thread":"storagePrunerAsyncRunner-async-1","class":"KvStoreDatabase","message":"Pruned 12 BlobSidecars","throwable":""}
{"@timestamp":"2025-01-28T08:52:01,991","level":"DEBUG","thread":"storagePrunerAsyncRunner-async-1","class":"BlobSidecarPruner","message":"Blobs pruning finished in 1779 ms. Limit reached: true","throwable":""}`
## Documentation

## Changelog


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -406,24 +406,37 @@ public void deleteHotBlocks(final Set<Bytes32> blockRootsToDelete) {
...
-        dao.getEarliestFinalizedBlock().map(SignedBeaconBlock::getSlot);
+    final Optional<UInt64> earliestFinalizedBlockSlot = dao.getEarliestFinalizedBlockSlot();
     LOG.debug(
         "Earliest block slot stored is {}",
-        () -> earliestBlockSlot.isEmpty() ? "EMPTY" : earliestBlockSlot.get().toString());
-    if (earliestBlockSlot.isEmpty()) {
+        () ->
+            earliestFinalizedBlockSlot.isEmpty()
+                ? "EMPTY"
+                : earliestFinalizedBlockSlot.get().toString());
```
tbenr: `earliestFinalizedBlockSlot.map(UInt64::toString).orElse("EMPTY")` :)

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -436,7 +449,7 @@ private UInt64 pruneToBlock(
     }
 
     try (final Stream<SignedBeaconBlock> stream =
-        dao.streamFinalizedBlocks(UInt64.ZERO, checkpointInitialSlot)) {
+        dao.streamFinalizedBlocks(earliestFinalizedBlockSlot, checkpointInitialSlot)) {
```
tbenr, line 449: these two rounds of `streamFinalizedBlocks` are a bit of a mystery to me.
why we need the second? and if we really need it why we cant simply stream keys (we just need slots here and the slot in the key is always block.getSlot()).

tbenr (commented):
generally LGTM I just want to check on a pre-existing thing that has been recently introduced (maybe a quick chat on it?)

teku/src/main/java/tech/pegasys/teku/services/ServiceController.java
```diff
@@ -37,7 +40,17 @@ protected SafeFuture<?> doStart() {
 
   @Override
   protected SafeFuture<?> doStop() {
-    return SafeFuture.allOf(services.stream().map(Service::stop).toArray(SafeFuture[]::new));
+    // Stop services in reverse order
+    final List<Service> reversedServices = new ArrayList<>(services);
+    Collections.reverse(reversedServices);
```
tbenr: is this on anther PR?

gfukushima: it was, I've removed from this one

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
gfukushima: I haven't looked into this optimization tbh, sound fair but I think I ran into issues when I first implemented this with 1 stream only

tbenr (approved):
LGTM. let's address my comment on another PR

PR #9033, opened 2025-01-24
Add custom iterator for LevelDB

tbenr:
This PR:

- will improve all API using `stream`, because checking `isValidKey()` and then calling `next()` will cause copy both key and value in JVM memory twice! With this change only the `key` will be copied twice. This will also avoid loading the `value` then we reached the `key` which is not valid (the next column)
  - a particularly heavy case is loading an `hotState` in memory when streaming via `getStateRootsBeforeSlot` (which has nothing to do with states), which happens whenever we finalize a new epoch.
- will avoid coping values in JVM memory completely when streaming keys
  - in practice it allows us to avoid coping blobs in JVM memory during blobs pruning. Running `DatabaseTest.verifyBlobsLifecycle()` I counted 76 calls to `NativeIterator::value`. With this PR they go down to 53.

Allocation of hot state (247MB allocation is an entire mainnet state) during `getStateRootsBeforeSlot`:
image

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/leveldb/CustomDBIterator.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/leveldb/CustomDBIterator.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright Consensys Software Inc., 2025
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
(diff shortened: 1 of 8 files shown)


mehdi-aouadi (approved):
LGTM
Just wondering if we could wrap the `JniDBIterator` in the `CustomJniDB` to reuse the already defined methods and only implement the new ones. Just an option I had in mind, no strong opinion here...

tbenr:
> Just wondering if we could wrap the JniDBIterator

mmm problem is that `JniDBIterator` constructor is not `public`.
Did you found an alternative way?

mehdi-aouadi:
> > Just wondering if we could wrap the JniDBIterator
> 
> mmm problem is that `JniDBIterator` constructor is not `public`. Did you found an alternative way?

Oh right! We can keep it this way then

PR #9043, opened 2025-01-27
update empty storage initialization log

mehdi-aouadi:
## PR Description
Update the storage initialisation log to make it clear when the storage is found empty.
In the current implementation we cannot differentiate an empty from a non empty storage initialisation.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #9049, opened 2025-01-28
testnet config updates

rolfyone:
Found some holesky fields missing also and raised a PR on holesky testnet config.

addresses https://github.com/ethereum/pm/issues/1265

## Documentation

## Changelog


ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/holesky.yaml
```diff
@@ -74,12 +74,6 @@ MAX_PER_EPOCH_ACTIVATION_CHURN_LIMIT: 8
 # ---------------------------------------------------------------
 # 40%
 PROPOSER_SCORE_BOOST: 40
-# 20%
-REORG_HEAD_WEIGHT_THRESHOLD: 20
-# 160%
-REORG_PARENT_WEIGHT_THRESHOLD: 160
-# `2` epochs
-REORG_MAX_EPOCHS_SINCE_FINALIZATION: 2
```
tbenr: why you removed those?

tbenr: i see that those are missing on https://github.com/eth-clients/ but they are present in consensus-spec mainnet config.

rolfyone: i removed them so they're the same as the holesky file, because they're not in the current version or the version i proposed... so i'd prefer they're the same in our code - unless we actually need them? are they more missing required config?

rolfyone: In my ideal world this file is exactly the same as whats in eth-clients/holesky basically...

rolfyone: I could add them back to the change i've put into holesky repo, it does look like they should be there...

lucassaldanha (approved):
LGTM. Do we need to wait for https://github.com/eth-clients/holesky/pull/119/files before merging this?

ethereum/spec/src/main/resources/tech/pegasys/teku/spec/config/configs/holesky.yaml
tbenr: Ye, I think the consensus-spec repo wins, and eth-clients repo should be compliant with it

tbenr:
makes sense to wait

rolfyone:
either/or. We can wait in case the dates change, thats fine.

tbenr:
They are indeed changing

PR #9054, opened 2025-01-28
Stop all pruners in storage service `doStop`

tbenr:
We missed to handle storage stop correctly when we added blobs and state pruners

## Documentation

## Changelog


rolfyone (approved):
LGTM

gfukushima (approved):
lgtm, but letting you know that I've tried to this and didn't make any difference for the sigsegv we're getting on shutdown

PR #9057, opened 2025-01-29
Improve CommonAncestor logic for sync

### LINKED ISSUE #8872: Ideas about forward sync improvements
When we select a “target chain” to sync to, we consider our current “hot blocks” but only when they exceed by OPTIMISTIC_HISTORY_LENGTH (3000) the current `firstNonFinalSlot`. Thus in long period on no-finality we always redownload 3000 slots of block\blobs if the node goes offline for more than `isCloseToInSync`. This happens because, if we are not `isCloseToInSync` we don’t even start gossip and we don’t pretend to be able catch-up via blockByRoot lookups.
We may consider lowering that down  `OPTIMISTIC_HISTORY_LENGTH` to reduce the overhead but the algorithm that selects the slot from which start redownloading is not that safe (see below).
3000 has been set pre-merge, when blocks where small and state transition was fast (small state too). 

I think the algorithm of CommonAncestor::getCommonAncestor can be improved. The current implementation only reason about slot and doesn’t try to to check that the selected ancestor is REALLY common, by checking block roots: the remote block root at the selected slot must match the local block root. By Improving this check we could also significantly lower the “redownload overhead” . We could have an algorithm that, based on local heads, searches for remote common ancestors going backwards. So we can have higher security and lower overhead.
NOTE: the search of the exact common ancestor can be done on a single specific slot but rather inside a range (like 10?) because our head(s) can be on a fork compared to the remote canonical chain, so we need to check in the range of <ourCurrentBestHeadSlot-10,ourCurrentBestHeadSlot> and, if no match is found, retrying backwards by jumping like 100 slots back until we find a correspondence.

tbenr:
This is the algorithm I had in mind.
This should drastically reduce the amount of block\blobs redownload under non-finality conditions.

It will:
- download 16 blocks per attempt
- start from a "common" head slot
- (on each subsequent attempt) jump back of 100 slots with power of 2 exponentiation. So it will be jump back of 100 slots at first attempt, then 200 (300 from head slot), then 400 (700 from selected head slot), and so on for max 6 times (up to 6300 slots from head slot)
- select the finalized slot if it is  reached along the way

fixes #8872

## Documentation

## Changelog


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/CommonAncestor.java
```diff
@@ -24,63 +26,107 @@
...
-  static final UInt64 OPTIMISTIC_HISTORY_LENGTH = UInt64.valueOf(3000);
-  // prysm allows a maximum range of 1000 blocks (endSlot - startSlot) due to database limitations
-  static final UInt64 BLOCK_COUNT = UInt64.valueOf(100);
+  private final RecentChainData recentChainData;
+  private final int maxAttempts;
 
   public CommonAncestor(final RecentChainData recentChainData) {
     this.recentChainData = recentChainData;
+    this.maxAttempts = DEFAULT_MAX_ATTEMPTS;
+  }
```
rolfyone: should probably call the other constructor with default_max_attempts here

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/CommonAncestorTest.java
```diff
@@ -89,27 +156,40 @@ void shouldSearchCommonAncestorWithSufficientRemoteData()
...
-        requestFuture,
+        requestFutureAttempt,
         blockResponseListenerArgumentCaptor.getValue(),
         syncStartSlot,
-        CommonAncestor.BLOCK_COUNT);
+        BLOCK_COUNT_PER_ATTEMPT);
 
-    requestFuture.complete(null);
-    // last received slot is the best slot
-    assertThat(futureSlot.get()).isEqualTo(syncStartSlot.plus(CommonAncestor.BLOCK_COUNT).minus(1));
```
rolfyone, line 113: seems important to check where we're at?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/CommonAncestor.java
tbenr: fair :)

tbenr: well, on a second pass, I don't want to leak the config outside the class.. exposing only the max attempts seems inconsistent, and I should expose all of them, and I'd avoid that :)

tbenr: oh.. I misread your comment

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/singlepeer/CommonAncestorTest.java
tbenr: mmm. I haven't fund where you think I missed the check.. additional hints?

rolfyone: i just mean we've removed this validation? im not sure why we would remove it...

tbenr: No it isn't removed. It's just the diff looking weird.

rolfyone (approved):
LGTM

PR #9060, opened 2025-01-30
update process_registry_updates

rolfyone:
- looking at 4104 in consensus-specs, its relatively simple to change our processRegistryUpdates to do this.

## Documentation

## Changelog


tbenr (approved):
LGTM. but you back-ported the change to phase0 but spec are still on plain `if`s

rolfyone:
i had also optimised electra before the spec PR...

PR #9065, opened 2025-01-30
Add support to builder validator registration using ssz

### LINKED ISSUE #9003: Add SSZ suport to `validatorRegistration` builder endpoint
https://github.com/ethereum/builder-specs/pull/110

lucassaldanha:
## PR Description
Implement support for SSZ requests when registering validators via the Builder endpoint. Logic based on this spec https://github.com/ethereum/builder-specs/pull/110.

Since we don't have a way to know if the relay supports SSZ, our best bet is to try SSZ and if it fails fallback into JSON. However, we do not want to keep trying SSZ for every request after a failure so we are adding a backoff logic (24h).

To support testing, I created a `RestBuilderClientOptions` record. The reasoning was to bypass the way that we were relying directly on static configs, which made testing harder.

## Fixed Issue(s)
fixes #9003

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
+++ b/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
@@ -282,9 +282,10 @@ class RestBuilderClientTest {
         .succeedsWithin(WAIT_FOR_CALL_COMPLETION)
         .satisfies(response -> assertThat(response.isSuccess()).isTrue());
 
-    assertThat(mockWebServer.getRequestCount()).isEqualTo(2);
     verifyPostRequestSsz("/eth/v1/builder/validators", signedValidatorRegistrations);
     verifyPostRequestJson("/eth/v1/builder/validators", signedValidatorRegistrationsRequest);
+
+    assertThat(mockWebServer.getRequestCount()).isEqualTo(2);
   }
 
   @TestTemplate
@@ -299,9 +300,13 @@ class RestBuilderClientTest {
     final SszList<SignedValidatorRegistration> signedValidatorRegistrations =
         createSignedValidatorRegistrations();
 
+    // First call, will try SSZ and fallback into JSON
     assertThat(restBuilderClient.registerValidators(SLOT, signedValidatorRegistrations))
         .succeedsWithin(WAIT_FOR_CALL_COMPLETION)
         .satisfies(response -> assertThat(response.isSuccess()).isTrue());
+
+    verifyPostRequestSsz("/eth/v1/builder/validators", signedValidatorRegistrations);
+    verifyPostRequestJson("/eth/v1/builder/validators", signedValidatorRegistrationsRequest);
 
     // After backoff period, will try SSZ again
     timeProvider.advanceTimeBy(Duration.ofDays(1));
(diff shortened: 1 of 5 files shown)


ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClient.java
```diff
@@ -203,6 +250,7 @@ private SchemaDefinitionsBellatrix getSchemaDefinitionsBellatrix(final SpecVersi
             () ->
                 new IllegalArgumentException(
                     specVersion.getMilestone()
-                        + " is not a supported milestone for the builder rest api. Milestones >= Bellatrix are supported."));
+                        + " is not a supported milestone for the builder rest api. Milestones >= Bellatrix are "
+                        + "supported."));
```
rolfyone: could be unchanged really...

ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
```diff
@@ -261,10 +268,57 @@ void registerValidators_failures() {
...
+    // After backoff period, will try SSZ again
+    timeProvider.advanceTimeBy(Duration.ofDays(1));
+    assertThat(restBuilderClient.registerValidators(SLOT, signedValidatorRegistrations))
+        .succeedsWithin(WAIT_FOR_CALL_COMPLETION)
+        .satisfies(response -> assertThat(response.isSuccess()).isTrue());
+
+    assertThat(mockWebServer.getRequestCount()).isEqualTo(3);
+    verifyPostRequestSsz("/eth/v1/builder/validators", signedValidatorRegistrations);
+    verifyPostRequestJson("/eth/v1/builder/validators", signedValidatorRegistrationsRequest);
+    verifyPostRequestSsz("/eth/v1/builder/validators", signedValidatorRegistrations);
```
rolfyone: nit: we should probably make a tiny change here to show that the first set are verified and the second set is just the ssz call... we basically show that with the asserts above but it'd be clearer if this last check is just seeing 1 request, and just ssz.

rolfyone (approved):
LGTM just nits.

ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
lucassaldanha: I guess I can move the early assertions closer to where we exercise the request to make them a bit more "sequential".

ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClient.java
```diff
@@ -98,12 +114,43 @@ public SafeFuture<Response<Void>> registerValidators(
...
+                    "Failed to register validator using SSZ. Will retry using JSON (error: {})",
+                    response.errorMessage());
+
+                nextSszRegisterValidatorsTryMillis =
+                    timeProvider.getTimeInMillis().plus(REGISTER_VALIDATOR_SSZ_BACKOFF_TIME_MILLIS);
+
+                return registerValidatorsUsingJson(signedValidatorRegistrations);
+              } else {
+                return SafeFuture.completedFuture(response);
+              }
```
tbenr: you could avoid `else` here ;-)

tbenr (commented):
I think what is missing is a test checking that the fallback doesn't happen on non-http errors, like timeout or connection errors.

tbenr (approved):
LGTM. Seems like adding a missing `final` makes it good to go

PR #9067, opened 2025-01-31
Rocks subcommand

gfukushima:
## PR Description
This a lift and shift from a tool implemented on Besu that we could use to analyse the db when using RockDb.

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbHelper.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbHelper.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.function.TriConsumer;
+import org.apache.tuweni.bytes.Bytes;
 import org.bouncycastle.util.Arrays;
 import org.rocksdb.ColumnFamilyDescriptor;
 import org.rocksdb.ColumnFamilyHandle;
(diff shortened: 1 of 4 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbHelper.java
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  static void printLine(
+      final SubCommandLogger out,
+      final String cfName,
+      final String keys,
+      final String totalFilesSize,
+      final String sstFilesSize,
+      final String blobFilesSize) {
+    final String format = "| %-30s | %-15s | %-11s | %-15s | %-16s |\n";
```
rolfyone: we should fix these widths for our long column name, i think if we at least have `%-35s` we'll fit column names

teku/src/main/java/tech/pegasys/teku/cli/subcommand/storage/RocksDbCommand.java
```diff
@@ -0,0 +1,103 @@
...
+        });
+    RocksDbHelper.printTotals(SUB_COMMAND_LOG, columnFamilyUsages);
+    return 0;
+  }
+
+  @Command(
+      name = "stats",
+      description = "Print rocksdb stats",
+      mixinStandardHelpOptions = true,
+      versionProvider = PicoCliVersionProvider.class)
```
rolfyone, line 80: this one is a whole hard thing to read, Im wondering if something like a column filter or something would be useful...

rolfyone: Another option is print a better header rather than the in-line which doesnt really make anything at all clear...

rolfyone: hrm maybe i was reading it wrongly, this isnt so bad...
```
=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
Column Family: FINALIZED_STATES_BY_SLOT
```
so maybe a filter option for the column?

gfukushima: Can we add this filter on a separate PR? I don't think this is crucial for what I'm trying to get this PR merged in.

rolfyone (approved):

teku/src/main/java/tech/pegasys/teku/cli/subcommand/storage/RocksDbCommand.java
rolfyone: thats fine, the history is just so long its not usable and you need to redirect to file.

PR #9090, opened 2025-02-08
perf(ssz): implement path caching in CachingTreeAccessor

crStiv:
I noticed a TODO comment in CachingTreeAccessor about potential performance optimization and decided to implement it. The main idea is to cache the full path to the last accessed node instead of just storing the node itself.

Changes made:
- Added path caching to store the complete traversal history
- Implemented common ancestor detection
- Added logic to start traversal from the nearest common ancestor
- Updated CachedTreeNode to store the path information

This should improve performance when accessing nodes that are close to each other in the tree structure, as we won't need to start traversal from the root every time. The optimization is especially effective for sequential access patterns.

The changes maintain thread-safety and backward compatibility while reducing unnecessary tree traversals.


rolfyone:
thanks for raising. this isn't currently compiling, can you run spotless and get the build to work?

rolfyone:
Doesn't compile or even pass spotless, and no movement, closing. if you wish for help on this, please reach out in discord.

PR #9111, opened 2025-02-12
Notify new payload

gfukushima:
## PR Description
This PR adds the IL to the new `NewPayloadRequest` and the new block processor
Part of #8937 

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [ ] I thought about adding a changelog entry, and added one if I deemed necessary.

Issue #8937, opened 2024-12-18
Engine APIs


mehdi-aouadi:
##### Modified `is_valid_block_hash`

*Note*: The function `is_valid_block_hash` is modified to include the additional `inclusion_list_transactions`.

```python
def is_valid_block_hash(self: ExecutionEngine,
                        execution_payload: ExecutionPayload,
                        parent_beacon_block_root: Root,
                        execution_requests_list: Sequence[bytes],
                        inclusion_list_transactions: Sequence[Transaction]) -> bool:
    """
    Return ``True`` if and only if ``execution_payload.block_hash`` is computed correctly.
    """
    ...
```

##### Modified `notify_new_payload`

*Note*: The function `notify_new_payload` is modified to include the additional `inclusion_list_transactions`.

```python
def notify_new_payload(self: ExecutionEngine,
                       execution_payload: ExecutionPayload,
                       parent_beacon_block_root: Root,
                       execution_requests_list: Sequence[bytes],
                       inclusion_list_transactions: Sequence[Transaction]) -> bool:
    """
    Return ``True`` if and only if ``execution_payload`` and ``execution_requests_list``
    are valid with respect to ``self.execution_state``.
    """
    # TODO: move this outside of notify_new_payload.
    # If execution client returns block does not satisfy inclusion list transactions, cache the block
    # store.unsatisfied_inclusion_list_blocks.add(execution_payload.block_root)
    ...
```

##### Modified `verify_and_notify_new_payload`

*Note*: The function `verify_and_notify_new_payload` is modified to pass the additional parameter
`inclusion_list_transactions` when calling `notify_new_payload` in EIP-7805.

```python
def verify_and_notify_new_payload(self: ExecutionEngine,
                                  new_payload_request: NewPayloadRequest) -> bool:
    """
    Return ``True`` if and only if ``new_payload_request`` is valid with respect to ``self.execution_state``.
    """
    execution_payload = new_payload_request.execution_payload
    parent_beacon_block_root = new_payload_request.parent_beacon_block_root
    execution_requests_list = get_execution_requests_list(new_payload_request.execution_requests)
    inclusion_list_transactions = new_payload_request.inclusion_list_transactions # [New in EIP-7805]

    if b'' in execution_payload.transactions:
        return False

    if not self.is_valid_block_hash(
            execution_payload,
            parent_beacon_block_root,
            execution_requests_list):
        return False

    if not self.is_valid_versioned_hashes(new_payload_request):
        return False

[...]

PR #9112, opened 2025-02-12
Process execution payload

gfukushima:
## PR Description
This PR was built on top of #9111
Adds a new validation to the ProcessExecutionPayload to ensure that the number of tx in the IL in no larger than the number  defined in the spec.

## Fixed Issue(s)

Part of #8937

## Documentation

## Changelog


tbenr:
do you really want to merge this on master? :)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/BlockProcessor.java
```diff
@@ -60,7 +63,8 @@ BeaconState processAndValidateBlock(
       SignedBeaconBlock signedBlock,
       BeaconState blockSlotState,
       IndexedAttestationCache indexedAttestationCache,
-      Optional<? extends OptimisticExecutionPayloadExecutor> payloadExecutor)
+      Optional<? extends OptimisticExecutionPayloadExecutor> payloadExecutor,
+      Function<SlotAndBlockRoot, Optional<List<InclusionList>>> inclusionListSupplier)
```
mehdi-aouadi: Is there a reason for using a supplier here? We could directly get the ILs as `Optional<List<InclusionList>>` from the store since we have the slot and block root when running a fork choice on a block (the `ForkChoice::onBlock` has the slot and the block root

mehdi-aouadi (approved):
LGTM
Just a nit regarding the supplier choice

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/block/BlockProcessor.java
gfukushima: So my choice to use the supplier here was more in terms of making sure that we only try to fetch the IL when we're assembly the payload, but I think at this stage the block has been already formed so yeah we can probably just retrieve the IL at `ForkChoice::onBlock`. Thanks for the review

PR #9114, opened 2025-02-12
Improve attestation selection

tbenr:
Make sure we select the best aggregations for a given slot.
In Electra we could easily end up discarding big aggregations if we don't apply any sorting.

I also fixed `getAttestationsForBlock_shouldNotAddMoreAttestationsThanAllowedInBlock` which was broken from very long time ago

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9127, opened 2025-02-17
Enable large cache per column family

### LINKED ISSUE #9089: [RocksDB] Enable setting larger cache for specific columns
We're currently creating a 128MB cache for every column family unnecessarily. 
I believe we only require such larger caches in 2 columns atm: FINALIZED_BLOCKS_BY_SLOT and BLOB_SIDECAR_BY_SLOT_AND_BLOCK_ROOT_AND_BLOB_INDEX.

gfukushima:
## PR Description
This PR adds an extra property to the KVStoreColumn to allow specifying if a certain column should use a larger cache, allowing us to provide larger caches only to the columns that require that.

## Fixed Issue(s)

Fixes #9089 

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstanceFactory.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstanceFactory.java
@@ -165,24 +165,24 @@ public class RocksDbInstanceFactory {
     final ColumnFamilyOptions cfOptions;
     try {
       final LRUCache cache =
-              column
-                      .getIsLargerCacheAvalilable()
-                      .map(
-                              isLarger -> {
-                                if (isLarger) {
-                                  return new LRUCache(configuration.getLargerCacheCapacity());
-                                } else {
-                                  return new LRUCache(configuration.getCacheCapacity());
-                                }
-                              })
-                      .orElse(new LRUCache(configuration.getCacheCapacity()));
-      cfOptions =  new ColumnFamilyOptions()
+          column
+              .getIsLargerCacheAvalilable()
+              .map(
+                  isLarger -> {
+                    if (isLarger) {
+                      return new LRUCache(configuration.getLargerCacheCapacity());
+                    } else {
+                      return new LRUCache(configuration.getCacheCapacity());
+                    }
+                  })
+              .orElse(new LRUCache(configuration.getCacheCapacity()));
+      cfOptions =
+          new ColumnFamilyOptions()
               .setCompressionType(configuration.getCompressionType())
               .setBottommostCompressionType(configuration.getBottomMostCompressionType())
               .setLevelCompactionDynamicLevelBytes(true)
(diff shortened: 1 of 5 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstanceFactory.java
```diff
@@ -164,24 +165,71 @@ private static DBOptions createDBOptions(
...
+      final KvStoreConfiguration configuration, final KvStoreColumn<?, ?> column) {
+    final ColumnFamilyOptions cfOptions;
+    try {
+      final LRUCache cache =
+          column
+              .getIsLargerCacheAvalilable()
+              .map(
+                  isLarger -> {
+                    if (isLarger) {
+                      LOG.info("Using larger cache for column {}", column.getId().toHexString());
```
lucassaldanha, line 177: NIT: not sure if we need this as a info. It seems to be pretty low level and I belive we are happy that most users are gonna use default values anyway.

storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstanceFactory.java
```diff
@@ -164,24 +165,71 @@ private static DBOptions createDBOptions(
...
+                      return new LRUCache(configuration.getLargerCacheCapacity());
+                    } else {
+                      return new LRUCache(configuration.getCacheCapacity());
+                    }
+                  })
+              .orElse(new LRUCache(configuration.getCacheCapacity()));
```
lucassaldanha, line 183: If `getIsLargerCacheAvalilable()` is Optional.empty(), it means that we are on the default case, right? If we don't want to change the existing behaviour we should use the large capacity (matching the old value) when getIsLargerCacheAvalilable() is empty. Maybe I am missing something but it does look like to me that we would be changing every cache to 32MB.

storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstanceFactory.java
gfukushima: This was just for testing purposes.

storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstanceFactory.java
gfukushima: So long story short the default value used to be 8MB, which is low. We've changed to 128MB as a first step to get the db performing better and we're trying to narrow down what columns need larger caches, that being said the default should be something lower than 128MB, Rocks default is 32MB (as of release 8.x) which is what I believe we should be using by default as well, 8MB was set long time ago and I don't think is suitable anymore for the current needs.

gfukushima:
I'm closing this PR because I haven't seen the benefits of using lower caches.

PR #9133, opened 2025-02-18
New validator_local_validator_balances metric

lucassaldanha:
The core changes in this PR are:
1. Update the method `getValidatorStatuses()` to return a `Map<BLSPublicKey, StateValidatorData>` instead of `Map<BLSPublicKey, ValidatorStatus>`.
2. Add a new validator metric named `local_validator_balances` that has the total staked amount of all validators running on the node, labelled by their withdrawal credential type.

Change 1 does not add any extra overhead on the query side of things, given that previously we were already querying a list of `StateValidatorData` from `chainDataProvider.getStateValidators(..)`, but reducing it into a list of `ValidatorStatus`. We are removing this transformation now.

Before Electra, we could infer the total staked amount in one node by multiplying the number of validators by 32 ETH. But post-Electra this isn't always true. With this metric, we can add some extra information regarding the total staked amount for the validators on a node, and the distribution of staked amount per credential type (e.g. potentially showing how much can be consolidated).

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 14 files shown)


rolfyone (approved):
LGTM

PR #9137, opened 2025-02-18
Set Accept header in builder getPayload

### LINKED ISSUE #9131: Unset Accept header in builder getPayload call
I'm testing SSZ support with mev-boost. I've noticed that Teku doesn't set an Accept header value when calling getPayload. It does set Content-Type and Eth-Consensus-Version, but since we use Accept to determine the response, mev-boost will always return JSON here.

tbenr:
fixes #9131

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
+++ b/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
@@ -673,9 +673,12 @@ class RestBuilderClientTest {
               verifyBuilderPayloadResponse(builderPayload);
             });
     final Consumer<RecordedRequest> containsConsensusVersionHeader =
-        req ->
-            assertThat(req.getHeader("Eth-Consensus-Version"))
-                .isEqualTo(milestone.name().toLowerCase(Locale.ROOT));
+        req -> {
+          assertThat(req.getHeader("Eth-Consensus-Version"))
+              .isEqualTo(milestone.name().toLowerCase(Locale.ROOT));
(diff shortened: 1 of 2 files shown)


ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
```diff
@@ -673,9 +673,12 @@ private void runGetPayloadSuccessAsJson() {
               verifyBuilderPayloadResponse(builderPayload);
             });
     final Consumer<RecordedRequest> containsConsensusVersionHeader =
-        req ->
-            assertThat(req.getHeader("Eth-Consensus-Version"))
-                .isEqualTo(milestone.name().toLowerCase(Locale.ROOT));
+        req -> {
+          assertThat(req.getHeader("Eth-Consensus-Version"))
```
mehdi-aouadi, line 677: We'd better use the `HEADER_CONSENSUS_VERSION` constant btw

mehdi-aouadi (approved):
LGTM
Just a nit

ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/RestBuilderClientTest.java
tbenr: well... I think in test we should testing the strings cose if we mess up the constants we will pass the tests :)

mehdi-aouadi: oh my bad that's right!

PR #9140, opened 2025-02-18
Attestations pool benchmark

tbenr:
Files are in public google drive links, and are from Holesky network

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/jmh/java/tech/pegasys/teku/statetransition.validation.signatures/AggregatingAttestationPoolBenchmark.java
+++ b/ethereum/statetransition/src/jmh/java/tech/pegasys/teku/statetransition.validation.signatures/AggregatingAttestationPoolBenchmark.java
@@ -64,16 +64,16 @@ public class AggregatingAttestationPoolBenchmark {
   // a reference file can be obtained here
   // https://drive.google.com/file/d/139bA7r88riFODZ7S0FpvtO7hmWmdC_XC/view?usp=drive_link
   private static final String STATE_PATH =
-      "/tmp/BeaconStateDeneb_3630479_03664f196162fb81a4406c508674dd1ede09b883d37d0f3d0f076897f68741d2.ssz";
+      "BeaconStateDeneb_3630479_03664f196162fb81a4406c508674dd1ede09b883d37d0f3d0f076897f68741d2.ssz";
 
   // a reference file can be obtained here
   // https://drive.google.com/file/d/1I5vXK-x8ZH9wh40wNf1oACXeF_U3to8J/view?usp=drive_link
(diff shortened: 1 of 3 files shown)


ethereum/statetransition/src/jmh/java/tech/pegasys/teku/statetransition.validation.signatures/AggregatingAttestationPoolBenchmark.java
```diff
@@ -0,0 +1,175 @@
...
+      "/tmp/BeaconStateDeneb_3630479_03664f196162fb81a4406c508674dd1ede09b883d37d0f3d0f076897f68741d2.ssz";
+
+  // a reference file can be obtained here
+  // https://drive.google.com/file/d/1I5vXK-x8ZH9wh40wNf1oACXeF_U3to8J/view?usp=drive_link
+  private static final String POOL_DUMP_PATH = "/tmp/attestations_3630479.multi_ssz";
+
+  // a reference file can be obtained here
+  // https://drive.google.com/file/d/1PN0OToyNOV0SyjeQaS7oF3J4cKbmy1nX/view?usp=drive_link
+  private static final String ACTUAL_BLOCK_PATH =
+      "/tmp/block-3630480-e652bd51c7e4e528fea0728a3ad96f86ceb92e9daa227f315e96a9884ceb187b.ssz";
```
rolfyone: we should probably use the temp path from system in case this is a windows system...

tbenr: mmm I didn't want to add any logic like that cose I think whoever want to run the benchmark will end changing the path manually anyway. I was tempted to use just a placeholder like `/path/to/Beacon....`

rolfyone: i was just thinking something like `System.getProperty("java.io.tmpdir")` but if we assume it isn't usable, I guess thats ok. we may as well just have the filename and no path. then people can always change the working directory to where the files are...

rolfyone (approved):

ethereum/statetransition/src/jmh/java/tech/pegasys/teku/statetransition.validation.signatures/AggregatingAttestationPoolBenchmark.java
tbenr: just filename makes sense to me. Changing and merging.

PR #9159, opened 2025-02-24
FC validate_inclusion_lists

### LINKED ISSUE #9097: Fork choice: validate ILs
#### New `validate_inclusion_lists`

```python
def validate_inclusion_lists(store: Store,
                             inclusion_list_transactions: Sequence[Transaction],
                             execution_payload: ExecutionPayload) -> None:
    """
    The ``execution_payload`` satisfies ``inclusion_list_transactions`` validity conditions either
    when all transactions are present in payload or when any missing transactions are found to be
    invalid when appended to the end of the payload unless the block is full.
    """
    # pylint: disable=unused-argument

    # Verify inclusion list is a valid length
    assert len(inclusion_list_transactions) <= MAX_TRANSACTIONS_PER_INCLUSION_LIST * INCLUSION_LIST_COMMITTEE_SIZE

    # Verify inclusion list transactions are present in the execution payload
    contains_all_txs = all(tx in execution_payload.transactions for tx in inclusion_list_transactions)
    if contains_all_txs:
        return

    # TODO: check remaining validity conditions
```

gfukushima:
## PR Description
This PR add the implementation of `validate_inclusion_lists` to fork choice. Although it's not being used yet, this is rather a useful to check to have and to perform up on block import.

## Fixed Issue(s)

Fixes #9097

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -934,12 +934,11 @@ public class ForkChoice implements ForkChoiceUpdatedResultSubscriber {
 
   // Implements `validate_inclusion_lists` added in EIP-7805 - consensus/fork-choice
   public void validateInclusionLists(
-      final UpdatableStore store,
       final List<Transaction> inclusionListTransactions,
       final ExecutionPayload executionPayload) {
-    final SpecConfigEip7805 eip7805 = spec.getGenesisSpec().getConfig().toVersionEip7805().get();
-    final int maxTransactionPerInclusionList = eip7805.getMaxTransactionsPerInclusionList();
-    final int inclusionListCommitteeSize = eip7805.getInclusionListCommitteeSize();
+    final SpecConfigEip7805 specConfigEip7805 = spec.atSlot(recentChainData.getHeadSlot()).getConfig().toVersionEip7805().orElseThrow();
+    final int maxTransactionPerInclusionList = specConfigEip7805.getMaxTransactionsPerInclusionList();
+    final int inclusionListCommitteeSize = specConfigEip7805.getInclusionListCommitteeSize();
 
     if (inclusionListTransactions.size()
         > maxTransactionPerInclusionList * inclusionListCommitteeSize) {
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
@@ -40,17 +40,21 @@ import tech.pegasys.teku.infrastructure.async.eventthread.EventThread;
 import tech.pegasys.teku.infrastructure.exceptions.ExceptionUtil;
 import tech.pegasys.teku.infrastructure.exceptions.FatalServiceFailureException;
 import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
+import tech.pegasys.teku.infrastructure.ssz.SszList;
 import tech.pegasys.teku.infrastructure.subscribers.Subscribers;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.Spec;
 import tech.pegasys.teku.spec.SpecVersion;
 import tech.pegasys.teku.spec.cache.CapturingIndexedAttestationCache;
 import tech.pegasys.teku.spec.cache.IndexedAttestationCache;
+import tech.pegasys.teku.spec.config.SpecConfigEip7805;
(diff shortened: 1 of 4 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -927,4 +931,25 @@ private Optional<Boolean> getOptimisticSyncing() {
   public interface OptimisticHeadSubscriber {
     void onOptimisticHeadChanged(boolean isHeadOptimistic);
   }
+
+  // Implements `validate_inclusion_lists` added in EIP-7805 - consensus/fork-choice
+  public void validateInclusionLists(
+      final UpdatableStore store,
```
mehdi-aouadi: We can get the `Store` from the `recentChainData` so no need to have it in the method signature

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -927,4 +931,25 @@ private Optional<Boolean> getOptimisticSyncing() {
...
+      final List<Transaction> inclusionListTransactions,
+      final ExecutionPayload executionPayload) {
+    final SpecConfigEip7805 eip7805 = spec.getGenesisSpec().getConfig().toVersionEip7805().get();
+    final int maxTransactionPerInclusionList = eip7805.getMaxTransactionsPerInclusionList();
+    final int inclusionListCommitteeSize = eip7805.getInclusionListCommitteeSize();
+
+    if (inclusionListTransactions.size()
+        > maxTransactionPerInclusionList * inclusionListCommitteeSize) {
+      throw new IllegalStateException("Inclusion list has too many transactions");
```
mehdi-aouadi: I guess we should return a `BlockImportResult` instead of throwing here

mehdi-aouadi (changes requested):

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -927,4 +931,25 @@ private Optional<Boolean> getOptimisticSyncing() {
   public interface OptimisticHeadSubscriber {
     void onOptimisticHeadChanged(boolean isHeadOptimistic);
   }
+
+  // Implements `validate_inclusion_lists` added in EIP-7805 - consensus/fork-choice
+  public void validateInclusionLists(
+      final UpdatableStore store,
+      final List<Transaction> inclusionListTransactions,
+      final ExecutionPayload executionPayload) {
+    final SpecConfigEip7805 eip7805 = spec.getGenesisSpec().getConfig().toVersionEip7805().get();
```
mehdi-aouadi: I'd rather call it specConfigEip7805...
Not sure if the genesis spec config will work here, maybe better get the current slot from the `recentChainData` and get the spec config based on that. I'd suggest using `orElseThrow()` or `SpecConfigEipr7805::required` instead of `get`

mehdi-aouadi (changes requested):

mehdi-aouadi:
Yes this check will be performed when we import a block on slot N+1. We check that the ILs we've collected from the p2p network are included in the new block's execution payload.
From [EIP-7805](https://eips.ethereum.org/EIPS/eip-7805) attester section:

> **Attesters**
> **_Slot N+1, t=4s:_** Attesters monitor the P2P network for the proposer’s block. Upon detecting it, they verify whether all transactions from their stored ILs are included in the proposer’s execution payload, except for ILs whose sender has equivocated. Based on their frozen view of the ILs from t=9s in the previous slot, attesters check if the execution payload satisfies IL conditions. This is done either by confirming that all transactions are present or by determining if any missing transactions are invalid when appended to the end of the payload. In such cases, attesters use the EL to perform nonce and balance checks to validate the missing transactions and check whether there is enough space in the block to include the transaction(s).

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
mehdi-aouadi: Or better call `spec.getCurrentSlot(recentChainData.getStore())` to get the current slot...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
gfukushima: done

mehdi-aouadi (approved):
LGTM
I guess we can merge it for now ans keep an eye on the spec since we'd need to check if any missing transaction is invalid by calling the EL but it's not clear yet how we will do that.

Issue #9022, opened 2025-01-22
[BEACON-API] support more blobs in api


rolfyone:
https://github.com/ethereum/beacon-APIs/pull/488

mehdi-aouadi:
After checking, we don't need to make any change in order for this API to work properly in Electra since the change is related to the returned list size and we don't differentiate the schemas based on that.
In order to be more aligned with the spec though, we'd need some prior changes to implement this one:
- Add support for the anyOf property: #8433 
- Probably review the already existing `minItems` which is set set to 1 when we're dealing with lists and should rather be set to 0 imo since we could return an empty list
- Add support for the `maxItems` property

PS: We already added the new meta data: `version`, `execution_optimistic` and `finalized`

TLDR: This is non blocking for Electra but should be implemented at some point to be aligned with the spec and that requires some additional features first (`anyOf`, `minItems` and `maxItems`)

rolfyone:
- minItems doesn't have a default, we're just passing 1 in some places. you can pass Optional.empty() and have no minItems, you could equally pass 0 or another value, it's just that we're passing minItems 1 currently.
 - I did check all instances of minItems:1 and i think they're all wrong, but equally its basically documentation...
     - `/eth/v1/beacon/states/{state_id}/validators` query params `id` and `status` both have minimum 1 set, but they're query parameters so inherently optional
     - `/eth/v1/beacon/blob/sidecars/{block_id}` has `indices` - also query parameter
     - `validator_balances`- `id` is a query parameter
     - `/teku/v1/beacon/blob/sidecars/{slot}` also uses minItems in a query parameter...
     
Basically it looks like the settings we had weren't really done properly, so I'll adjust that as a first PR (the type definition isn't right)

rolfyone:
this shoudl be ok for now. closing.

PR #9161, opened 2025-02-24
fix SerializableArrayTypeDefinition minItems and add maxItems

rolfyone:
When I was looking at 9022, realised that minItems was hardcoded at a point and with a very specific use-case, so i've removed that in favor of a correct implementation.

Also implemented maxItems, and validations around min and max being at least 0 and in the right order of they'll generate a configuration exception.

I think it's fair to say this is basically only documentation currently given to nothing breaking other than our types tests.

I let this run and did some queries on a testnet and it was ok. will see if any AT's trip.

Partially addresses #9022

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM
Just a nit though, I noticed that our rendering still isn't "very clean".
From the [Beacon-APIs swagger](https://ethereum.github.io/beacon-APIs/#/Validator/getAttesterDuties), the attester duties doc looks like:
Screenshot 2025-02-24 at 08 50 39
But in our doc it looks like:
Screenshot 2025-02-24 at 08 50 47

IMU that's because we're using the global `INTEGER_TYPE` instead of a customised version of it per API

PR #9201, opened 2025-03-03
[DON'T MERGE] Incident intervals prototype

zilm13:
Implementation of incident intervals in Teku

- Dirty hack is made to use incident intervals only after client restart, because I didn't find a better way to use different flags/config in Kurtosis only after restart in the middle.
- 2 changes are not yet added to the spec and should be clarified
  - exception when trying to produce `Attestation` in incident interval in `getGenericAttestationData`
  - failed broadcast validation in `BlockManager` for blocks inside incident interval, so local block produced during incident interval is not published (otherwise it will be published even with failed block import)

`recover.yaml` for Kurtosis
```bash
participants:
  - el_type: geth
    cl_type: teku
    cl_extra_params: [
        --validators-keystore-locking-enabled=false,
	--Xincident-intervals=1-2
    ]
    cl_image: consensys/teku:develop
    cl_min_cpu: 2000
    cl_min_mem: 2000
    count: 5
network_params:
  electra_fork_epoch: 1
global_log_level: info
additional_services:
  - dora
  - spamoor_blob
  - prometheus_grafana
```

Steps to reproduce expected scenario:
```bash
# build Teku and save in local Docker repository
% ./gradlew distDocker 
# start Kurtosis with the provided config
% ./kurtosis run --enclave my-testnet github.com/ethpandaops/ethereum-package --args-file recover.yaml
```
On the start of epoch 1 stop 2-5 nodes and start them again (so incident intervals will be applied after restart):
```bash
% kurtosis service stop my-testnet cl-2-teku-geth                                                  
% kurtosis service stop my-testnet cl-3-teku-geth
% kurtosis service stop my-testnet cl-4-teku-geth
% kurtosis service stop my-testnet cl-5-teku-geth


% kurtosis service start my-testnet cl-2-teku-geth
% kurtosis service start my-testnet cl-3-teku-geth
% kurtosis service start my-testnet cl-4-teku-geth
% kurtosis service start my-testnet cl-5-teku-geth
```

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
@@ -269,6 +269,19 @@ public class Eth2NetworkOptions {
       arity = "0..1")
   private String epochsStoreBlobs;
 
+  @CommandLine.Option(
+      names = {"--Xincident-intervals"},
+      hidden = true,
+      paramLabel = "<STRING>",
+      description =
+          "Sets the interval where network is considered to be in incident. "
+              + "All the slots in these epochs should be empty. Any chain with blocks "
+              + "inside interval is considered invalid.",
+      fallbackValue = "",
+      showDefaultValue = Visibility.ALWAYS,
+      arity = "0..1")
+  private String incidentIntervals;
+
   public Eth2NetworkConfiguration getNetworkConfiguration() {
     return createEth2NetworkConfig(builder -> {});
   }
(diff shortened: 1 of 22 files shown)


zilm13:
Expected behavior:

chain created by node-1 during epochs 1-2 is orphaned
imagen

chain derived on the side from epoch 3 is finalized
imagen

There are some issues with prototype so we don't see 100% at this moment, but I will not focus on this issue at the current stage (need to abandon already imported blocks during incident intervals probably)

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
```diff
@@ -269,6 +269,19 @@ public class Eth2NetworkOptions {
...
   private String epochsStoreBlobs;
 
+  @CommandLine.Option(
+      names = {"--Xincident-intervals"},
+      hidden = true,
+      paramLabel = "<STRING>",
+      description =
+          "Sets the interval where network is considered to be in incident. "
+              + "All the slots in these epochs should be empty. Any chain with blocks "
+              + "inside interval is considered invalid.",
```
rolfyone, line 279: hrm so thats interesting, im not sure any epochs were completely empty...

zilm13: it's expected that it must fork so all the slots inside incident intervals are completely empty

zilm13:
closing, leaving it in my own repo https://github.com/zilm13/teku/pull/2
we definitely not have to keep it here

PR #9205, opened 2025-03-05
dump slots and roots from hot blocks

rolfyone:
This is a debug-tool to allow us to see what is in the hot block storage, just by slot and root.

It prints as CSV but the table is ordered by hash.  Simple enough to re-order from other tools so a raw dump is fastest as the first step.

## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
@@ -447,6 +447,37 @@ public class DebugDbCommand implements Runnable {
   }
 
   @Command(
+      name = "get-hot-block-slot-to-root",
+      description = "Writes all the slots and roots of stored hot blocks, will be in hash order.",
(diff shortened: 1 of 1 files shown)


teku/src/main/java/tech/pegasys/teku/cli/subcommand/debug/DebugDbCommand.java
```diff
@@ -446,6 +446,37 @@ public int dumpHotBlocks(
...
+  @Command(
+      name = "get-hot-block-slot-to-root",
+      description = "Writes all the slots and roots of stored hot blocks, will be in hash order.",
+      mixinStandardHelpOptions = true,
+      showDefaultValues = true,
+      abbreviateSynopsis = true,
+      versionProvider = PicoCliVersionProvider.class,
+      synopsisHeading = "%n",
+      descriptionHeading = "%nDescription:%n%n",
+      optionListHeading = "%nOptions:%n",
```
rolfyone, line 458: i should check if i need this heading

lucassaldanha (approved):
LGTM

PR #9206, opened 2025-03-06
fixed message for stateTooOldException

rolfyone:
I think the slot too old message was incorrect which is why it sometimes didnt make a lot of sense. added a test case too.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #9208, opened 2025-03-06
Do not mark block as invalid on RejectedExecutionException

### LINKED ISSUE #9207: Do not mark block as invalid on `INTERNAL_ERROR` with cause `RejectedExecutionException`
If the BN goes under pressure and the beacon execution queue becomes full, we may fail importing the block and mark it as invalid.
While I understand the principle of marking a block as invalid when it causes unexpected errors (it could be considered a liveness issue, but we can argue), in this case it is clearly a resource issue on the node itself that it is not related to any particular internal exception.

```text
2025-03-05 01:48:49.343 ERROR - Internal error while importing block: 63d06982223948c8d1a55ac8bed7084cfdb7f75cc26f27d40446e6fadd9577b3 (44039). Block content: 0x6400000092a0ebc4133c69cfed08e5e805be59b.......
...
java.util.concurrent.CompletionException: java.util.concurrent.RejectedExecutionException: Rejected execution on task queue - beaconchain
```

Image

tbenr:
fixes #9207

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/block/BlockManager.java
@@ -16,11 +16,13 @@ package tech.pegasys.teku.statetransition.block;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


StefanBratanov (approved):
LGTM

PR #9211, opened 2025-03-06
Sync common ancestor improvements

tbenr:
- Avoids falling back to last finalized slot when an exception occurs during peer rpc. To avoid starting from very far too soon
- added a `--Xp2p-sync-max-distance-from-head` to limit how far back from our local head we will ever try to sync, useful during long non-finality with nodes on the network stuck on old "wrong" chains.

## Documentation

## Changelog


lucassaldanha (approved):
LGMT

PR #9223, opened 2025-03-11
Fix HTTP headers case sensitivity

### LINKED ISSUE #9170: HTTP headers case sensitivity
HTTP headers check should be case insensitive: [RFC 7230, Section 3.2](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2)
Example: `PostAttestationV2` API fails if the `Eth-Consensus-Version` headers isn't case sensitive

StefanBratanov:
## PR Description
Use `new TreeMap<>(String.CASE_INSENSITIVE_ORDER);` to store the headers coming from `io.javalin.http.Context`

## Fixed Issue(s)
fixes #9170 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9226, opened 2025-03-12
Apply `--logging=<LEVEL>` only to Teku related classes

### LINKED ISSUE #9221: --logging=<LEVEL> should only apply to Teku and related classes?
It was raised that we should consider only changing the logging level of Teku classes and other related dependencies (e.g. discovery and p2p) when changing the log level using `--logging`.

This should not change the behaviour of changing the log level via log4j config or via the change log API endpoint.

StefanBratanov:
## PR Description
Just cater for Teku, discovery and libp2p logging

## Fixed Issue(s)
fixes #9221 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 ## Unreleased Changes
 
 ### Breaking Changes
+- `--logging` CLI option has been modified to specify only the logging level of Teku classes and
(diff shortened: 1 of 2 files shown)


zilm13 (approved):
LGTM

PR #9228, opened 2025-03-12
Fix PendingDeposit (api) type (Bytes20 -> Bytes32)

lucassaldanha:
## PR Description
- Fix PendingDeposit type (Bytes20 -> Bytes32) - this does not directly causes any issue, but can bite us in the future so it is better to get it correct now.
- Upgraded reference tests to v1.5.0-beta.2.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -324,7 +324,7 @@ allprojects {
 }
 
 def nightly = System.getenv("NIGHTLY") != null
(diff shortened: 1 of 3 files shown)


gfukushima (approved):
lgtm

Issue #9169, opened 2025-02-24
[HOLESKY PECTRA] exception trace to look at from external validator


rolfyone:
```
│ teku at tech.pegasys.teku.validator.coordinator.ValidatorApiHandler.sendSignedAttestations(ValidatorApiHandler.java:584) ~[teku-beacon-validator-25.2.0.jar:25.2.0]                                        │
│ teku at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source) ~[?:?]                                                                                                            │
│ teku at java.base/java.lang.reflect.Method.invoke(Unknown Source) ~[?:?]                                                                                                                                   │
│ geth {"t":"2025-02-24T22:47:21.532597942Z","lvl":"info","msg":"Chain head was updated","number":"3419870","hash":"0xade84a8cf42ee766ce4639bf9da07a5d0e7ecb2e80edc298c4d4a70577110fdf","root":"0x6e9b0b3779 │
│ teku at tech.pegasys.teku.infrastructure.events.DirectEventDeliverer.executeMethod(DirectEventDeliverer.java:74) ~[teku-infrastructure-events-25.2.0.jar:25.2.0]                                           │
│ geth {"t":"2025-02-24T22:47:21.609553151Z","lvl":"info","msg":"Imported new potential chain segment","number":"3419871","hash":"0x08a1be6b7532def6c02d2408380963e54d68f551d2c925d1bf5a424403931d03","block │
│ geth {"t":"2025-02-24T22:47:21.903806503Z","lvl":"info","msg":"Chain head was updated","number":"3419871","hash":"0x08a1be6b7532def6c02d2408380963e54d68f551d2c925d1bf5a424403931d03","root":"0xa0a3ed45ae │
│ geth {"t":"2025-02-24T22:47:22.009650927Z","lvl":"info","msg":"Imported new potential chain segment","number":"3419872","hash":"0xa8b0a41118a3fb8ea33e1284c135f12556c10c89676b77a55cba8370a0d8532e","block │
│ teku at tech.pegasys.teku.infrastructure.events.DirectEventDeliverer.deliverToWithResponse(DirectEventDeliverer.java:67) ~[teku-infrastructure-events-25.2.0.jar:25.2.0]                                   │
[...]

PR #9232, opened 2025-03-13
Ignore single attestations in performance tracker

mehdi-aouadi:
## PR Description
Ignore any `SingleAttestation` in the `DefaultPerformanceTracker`.
This could happen when an attestation is ignored during its processing and hence not converted to an aggregated one. In that case, we end up passing a `SingleAttestation` to the `DefaultPerformanceTracker`

## Fixed Issue(s)

#9169 

## Documentation

## Changelog

### DIFF
--- a/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
+++ b/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
@@ -14,7 +14,6 @@
 package tech.pegasys.teku.validator.coordinator.performance;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
 import it.unimi.dsi.fastutil.ints.Int2IntMap;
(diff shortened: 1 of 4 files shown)


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
```diff
@@ -433,11 +432,14 @@ private SafeFuture<Map<UInt64, List<Attestation>>> getAttestationsIncludedInEpoc
...
   @Override
   public void saveProducedAttestation(final Attestation attestation) {
-    checkState(!attestation.isSingleAttestation(), "Single attestation is not supported");
-    final UInt64 epoch = spec.computeEpochAtSlot(attestation.getData().getSlot());
-    final Set<Attestation> attestationsInEpoch =
-        producedAttestationsByEpoch.computeIfAbsent(epoch, __ -> concurrentSet());
-    attestationsInEpoch.add(attestation);
+    if (attestation.isSingleAttestation()) {
+      LOG.warn("Single attestation is not supported");
+    } else {
```
tbenr: early return here:

```java
    if (attestation.isSingleAttestation()) {
      LOG.warn("Single attestation is not supported");
      return;
    }
    final UInt64 ...
```

tbenr (approved):
LGTM. just a nit

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/performance/DefaultPerformanceTracker.java
mehdi-aouadi: I've been traumatised in the past by a code checker that forces you to always have an else when you have an if condition... 😅

Issue #9219, opened 2025-03-11
Attestation Aggregation logs/metrics


lucassaldanha:
We need to revisit our metrics associated with attestation aggregation. The main goal is to have something that we can observe in real time, and use as part of our post-deployment checklist, etc.

- [x] Explicitly print the aggregation time in our Slow Block Warning msg.
- [ ] Add all values from Slow Block Warning as a metric (similar to Block Important) - can't add any extra time to block production.

PR #9233, opened 2025-03-14
Report time to collect attestations in slow block production

gfukushima:
## PR Description
This PR adds another bit to the block performance report to inform the time taken to get the attestations during the block production. 

## Fixed Issue(s)

Part of #9219

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockOperationSelectorFactory.java
```diff
@@ -126,6 +126,7 @@ public Function<BeaconBlockBodyBuilder, SafeFuture<Void>> createSelector(
       final SszList<Attestation> attestations =
           attestationPool.getAttestationsForBlock(
               blockSlotState, new AttestationForkChecker(spec, blockSlotState));
+      blockProductionPerformance.getAttestationsForBlock();
```
tbenr, line 130: maybe we can add a `beaconBlockPreparationStarted` at the beginning of the bodyBuilder lambda, otherwise we will measure the time from the previous step which is currently `retrieve_state`

gfukushima: makes total sense to me, added the preparationstarted event

tbenr (approved):
LGTM

PR #9243, opened 2025-03-17
Avoid optionals in uint64

tbenr:
An approach to remove safe versions with optionals. And makes up to caller to handle overflow\underflow (if needed)

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #9247, opened 2025-03-18
Improve late block reorg

tbenr:
This PR makes sure we:
- correctly use `shouldOverrideForkChoiceUpdate` in forkchoice notifier
- we timely `processHead` so that we make the reorging block the head soon enough to attest and produce sync committee for it. (This should improve our sync committee accuracy in reorging blocks)

node producing a reorging block:
image

node importing a reorging block:
image

related to #6595 

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -744,16 +746,34 @@ private void onExecutionPayloadResult(
...
+      //
+      // - node producing a reorging block will attest and produce sync committee for the reorging
+      // block
+      // - node just importing (not producing it) a reorging block will produce sync committee for
+      // the new head
+      //
+      // It wouldn't be enough to trigger processHead at attestationDue because it will only
+      // cover attestation generation and not sync committee message, since it is generated
+      // VC side and requires head event to be sent before attestationDue.
+      processHead().finish(error -> LOG.error("Fork choice updating head failed", error));
```
rolfyone, line 775: what does the user do when they see that error several times in a minute? its a very generic message, i could only see them raising a bug...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceNotifierImpl.java
```diff
@@ -195,12 +198,22 @@ private void internalForkChoiceUpdated(
...
     LOG.debug("internalForkChoiceUpdated forkChoiceState {}", forkChoiceState);
 
+    final Optional<UInt64> localProposingSlot =
+        calculatePayloadAttributesSlot(forkChoiceState, proposingSlot);
+
+    if (forkChoiceLateBlockReorgEnabled
+        && localProposingSlot.isPresent()
+        && recentChainData.shouldOverrideForkChoiceUpdate(forkChoiceState.getHeadBlockRoot())) {
+      LOG.debug(
+          "internalForkChoiceUpdated forkChoiceState skipped due to late block reorg override");
```
rolfyone: proposing slot may be interesting for context

rolfyone (approved):
LGTM just consider logs

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
tbenr: humm... this replicates what happens here:
https://github.com/Consensys/teku/blob/a98149a53607d7ec1ceaa2f6d0bfbf5454ce395c/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java#L726

I think the client is in a bad shape if we cannot process head. I'd not hide the problem under the rug.. do you have any better message or idea?

tbenr: In the meantime going to merge this.
We can apply a change later if we can find a better approach.

rolfyone: well the exception trace will possibly cause more problems than it solves, at least we only want this infrequently so rate limiting it if we're including the whole exception trace. otherewise potentially busting it to a lower level and only including the message by default rather than the stacktrace would at least make it less noise

PR #9250, opened 2025-03-18
Block production metrics based on event logs

gfukushima:
## PR Description
This PR adds a set of metrics that basically export the block production events log.

## Fixed Issue(s)

Part of #9219 

## Documentation

## Changelog


ethereum/performance-trackers/src/main/java/tech/pegasys/teku/ethereum/performance/trackers/BlockProductionPerformanceImpl.java
```diff
@@ -21,22 +21,39 @@
...
+  public static final String RETRIEVE_STATE = "retrieve_state";
+  public static final String LOCAL_GET_PAYLOAD = "local_get_payload";
+  public static final String BUILDER_GET_HEADER = "builder_get_header";
+  public static final String BUILDER_BID_VALIDATED = "builder_bid_validated";
+  public static final String BEACON_BLOCK_CREATED = "beacon_block_created";
+  public static final String STATE_TRANSITION = "state_transition";
+  public static final String STATE_HASHING = "state_hashing";
+  public static final String GET_ATTESTATIONS_FOR_BLOCK = "get_attestations_for_block";
+  public static final String BEACON_BLOCK_PREPARATION_STARTED = "beacon_block_preparation_started";
+  public static final String TOTAL_PRODUCTION_TIME_LABEL = "total_production_time";
```
rolfyone, line 38: should this just be an enum?

rolfyone (approved):
LGTM overall, just curious if we use an enum for the strings if it'd be a little cleaner, but i guess functionally the same thing.

ethereum/performance-trackers/src/main/java/tech/pegasys/teku/ethereum/performance/trackers/BlockProductionPerformanceImpl.java
gfukushima: It could be, just following the same notation we used in `BlockImportPerformance` changing them will imply changes on some other classes that consume those though.

PR #9252, opened 2025-03-19
make LibP2PNetwork and LibP2PGossipNetwork constructors public

pinges:
To make these two classes reusable in Maru the constructors are made public.


tbenr:
do we still need to do this @lucassaldanha @pinges ?

lucassaldanha (approved):
I'll merge this. But remember that we can't guarantee stability in these classes APIs!

PR #9268, opened 2025-03-24
Allow reorg during sync

### LINKED ISSUE #9195: During syncing we should be able to reorg blocks
When we start syncing and our local head is on a orphaned block, we should be able to trigger the reorg. Otherwise our canonical head remains stuck on that orphaned block and we also keep printing `Syncing.. ` event with head slot stuck on that. The other bad side-effect is that if synch restarts, we will restart again from that orphaned block

tbenr:
- adds a new `SyncReorgManager` which is triggered at each successful batch import.
- `SyncReorgManager` will call a protoarray's `reorgWhileSyncing` (via ForkChoiceTrigger->ForkChoice) which will "switch" votes from the old chain to the new chain. This logic only applies to protoarray in "syncing state" (no votes applied)

It is designed to reorg only if we import a block at a more recent slot than our current head (+ 10 slot).
There could be use-cases in which we want to reorg sooner: if our DB contains a long non-canonical chain we will keep the tip of that as our canonical head until the contending chain we are syncing overcome that tip.
But for now this implementation solves the problem in which our chain tip is just on a reorged block and we just keep syncing from its parent.

fixes #9195

## Documentation

## Changelog

---

> [!NOTE]
> **High Risk**
> Changes fork-choice/protoarray behavior and adds a new reorg path triggered during sync, which is consensus-critical and could affect head selection if the reorg criteria or weight adjustments are wrong.
> 
> **Overview**
> Enables *reorging during multipeer forward sync* by emitting a new “blocks imported” event from `BatchSync` after each successful batch import and wiring it through `SyncController`.
> 
> Adds `SyncReorgManager`, invoked only for non-speculative syncs, which compares the current head to the most recently imported block and (once the imported block is at least 10 slots ahead) triggers a fork-choice reorg via `ForkChoiceTrigger`.
> 
> Implements `reorgWhileSyncing` through `ForkChoice` → `ForkChoiceStrategy` → `ProtoArray`, reassigning protoarray weights from the old head branch to the new one *only when in syncing-mode weight conditions*, with new unit tests covering event delivery and reorg/guard behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 128d46aa4b38568fc8b71632fdd44d71d1d13206. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
@@ -33,6 +33,12 @@ public interface Sync {
 
   SafeFuture<Optional<SyncProgress>> getSyncProgress();
 
+  long subscribeToBlocksImportedEvent(BlocksImportedSubscriber subscriber);
+
+  interface BlocksImportedSubscriber {
+    void onBlocksImported(SignedBeaconBlock lastImportedBlock);
+  }
+
   record SyncProgress(
       UInt64 fromSlot,
       UInt64 toSlot,
@@ -44,10 +50,4 @@ public interface Sync {
       boolean importing,
       SlotAndBlockRoot targetChainHead,
       int targetChainPeers) {}
-
-  long subscribeToBlocksImportedEvent(BlocksImportedSubscriber subscriber);
-
-  interface BlocksImportedSubscriber {
-    void onBlocksImported(SignedBeaconBlock lastImportedBlock);
-  }
 }
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
@@ -18,6 +18,7 @@ import tech.pegasys.teku.beacon.sync.forward.multipeer.chains.TargetChain;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
+import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 
 public interface Sync {
 
@@ -43,4 +44,10 @@ public interface Sync {
       boolean importing,
       SlotAndBlockRoot targetChainHead,
       int targetChainPeers) {}
+
+  long subscribeToBlocksImportedEvent(BlocksImportedSubscriber subscriber);
+
+  interface BlocksImportedSubscriber {
+    void onBlocksImported(SignedBeaconBlock lastImportedBlock);
+  }
 }
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
@@ -17,8 +17,8 @@ import java.util.Optional;
 import tech.pegasys.teku.beacon.sync.forward.multipeer.chains.TargetChain;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
-import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
+import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
 
 public interface Sync {
 
--- a/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncController.java
+++ b/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncController.java
@@ -14,7 +14,6 @@
 package tech.pegasys.teku.beacon.sync.forward.multipeer;
 
 import com.google.common.base.MoreObjects;
-import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.Executor;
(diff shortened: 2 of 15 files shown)


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
```diff
@@ -32,6 +33,12 @@ public interface Sync {
 
   SafeFuture<Optional<SyncProgress>> getSyncProgress();
 
+  long subscribeToBlocksImportedEvent(BlocksImportedSubscriber subscriber);
+
+  interface BlocksImportedSubscriber {
+    void onBlocksImported(SignedBeaconBlock lastImportedBlock);
```
zilm13, line 39: why 's, plural?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncController.java
```diff
@@ -55,12 +58,26 @@ public SyncController(
...
+    this.syncReorgManager = syncReorgManager;
     this.sync = sync;
+    sync.subscribeToBlocksImportedEvent(this);
+  }
+
+  @Override
+  public void onBlocksImported(final SignedBeaconBlock lastImportedBlock) {
+    eventThread.execute(
+        () -> {
+          if (isSyncSpeculative()) {
```
zilm13, line 87: why do we skip this case?
(sorry, I just don't understand what is speculative sync)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncReorgManager.java
```diff
@@ -0,0 +1,58 @@
...
+
+    if (lastImportedBlock.getRoot().equals(currentHead.get().getRoot())) {
+      return;
+    }
+
+    if (currentHead
+        .get()
+        .getSlot()
+        .plus(REORG_SLOT_THRESHOLD)
+        .isGreaterThan(lastImportedBlock.getSlot())) {
```
zilm13, line 52: Not sure about this rule, we determinate best chain by votes, two chains could have similar length having a big difference in votes. I guess we don't want ot switch back and forth and it could help. Is this a purpose?

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -189,6 +189,60 @@ public void setInitialCanonicalBlockRoot(final Bytes32 initialCanonicalBlockRoot
...
+   * state where all nodes have been just initialized (see {@link #setInitialCanonicalBlockRoot})
+   * and have a weight of 1 on the canonical chain and 0 on all other chains.
+   *
+   * <p>It may also do the reorg if we enter syncing mode and, while syncing, we decide to sync to a
+   * new chain so that the common ancestor has not received any votes yet.
+   *
+   * <p>The function set all weights to 0 to the reorged chain and set all weights to 1 to the new
+   * chain.
+   */
+  public void reorgWhileSyncing(
```
zilm13, line 233: so we just removed weight from old chain and put it to the new chain undoubtely? We need some reasons for this. Could we get real weights? There is no clue new chain is better than old

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
tbenr: well cose I don't want this to look like you can subscribe to it and pretend to get all blocks.
Maybe I should rename everything to be `onBatchImported(SignedBeaconBlock lastImportedBlock)`

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncController.java
tbenr: speculative syncing is when we are on sync but we "hear" about a chain that we speculatively sync. So we don't want the sync reorg logic to be applied in that case (we are actually in sync in that scenario)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncReorgManager.java
tbenr: as discussed, the sync reorg logic only applies during initial syncing when all protoarray votes are 0

`ProtoArray::reorgWhileSyncing`
```
    // check we are in syncing mode where all nodes have a max weight of 1
    if (commonAncestorNode.getWeight().isGreaterThan(1)) {
      return;
    }
```

and it is actually ok to switch back and forth, as long as we keep as canonical the longest chain we are syncing to

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
tbenr: the idea is to make it the chain we want to reorg to as canonical only by 1 vote. So once we start counting actual votes that +1 becomes irrelevant and goes away as we follow the chain (latest imported block won't have that +1 anymore)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/Sync.java
tbenr: oh, problem is that I put that on `Sync` interface, and not `BatchSync` interface. So the concept of "batch" is not part of that...
Suggestions? :)

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArray.java
```diff
@@ -189,6 +189,60 @@ public void setInitialCanonicalBlockRoot(final Bytes32 initialCanonicalBlockRoot
...
+    // let's set weight to 1 from the best descendant of the new head up to the first non-zero
+    // weight node
+    node = newHeadNode.getBestDescendantIndex().map(this::getNodeByIndex).orElse(newHeadNode);
+    while (node.getWeight().isZero()) {
+      node.adjustWeight(1);
+      if (node.getParentIndex().isEmpty()) {
+        break;
+      }
+      node = getNodeByIndex(node.getParentIndex().get());
+    }
```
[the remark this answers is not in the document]
tbenr: this is actually intended. The idea is to select the tip of that chain where new head is (if in the meantime we added a new block on top of it) and set all chain's nodes to weight 1

PR #9276, opened 2025-03-26
Small fix on StartupLog

lucassaldanha:
- We were logging details about validator API even when it was disabled.
- Also rewrote the unit tests so we have one test case per report entry.


gfukushima (approved):
lgtm

PR #9282, opened 2025-03-26
Fix timers for validator_duty metrics

tbenr:
we should not use AutoClose pattern in an async context.

The side effect was that we were actually stopping the timer twice and double-couting due to "observe" function called twice on the underlining `io.prometheus.metrics.core.datapoints.Timer`

image

before:

image

after:

image

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #9288, opened 2025-03-27
Trim doppelganger log at epoch

### LINKED ISSUE #9270: doppelganger check - log on startup
if we've got a large number of keys in our doppelganger test, we spit them all out several times, and its a huge thing for logs (and people reading logs) to see.

we should probably print out the total count, and the first handful of short keys after the first instance of printing the whole set, or only print out the count over a certain number .... I think we do this for attestations ,limit list size...

gfukushima:
## PR Description
This PR reduces the noise of the the doppelganger check performed at every epoch when the functionality is enabled.
The result would look like the snippet below:
```
2025-03-27 12:15:52.070 INFO  - Performing doppelganger check for at Epoch 12 for public keys 8c98019, b43fcda, a43484d, b5a83ac, a386212, ac210cf, b242e56, 8f14a40, 93568c9, 8eb8faf, b881d9e, b960a78, 89dc848, b9015e7, a24dd1c, 8682073, a3ee793, ad6a9d2, ad69af5, 8b4ff71… (1024 total)
```

## Fixed Issue(s)

Fixes #9270 

## Documentation

## Changelog

### DIFF
--- a/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
+++ b/infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
@@ -36,6 +36,7 @@ public class StatusLogger {
 
   public static final StatusLogger STATUS_LOG =
       new StatusLogger(LoggingConfigurator.STATUS_LOGGER_NAME);
+  public static final int KEY_LIMIT = 20;
 
   @SuppressWarnings("PrivateStaticFinalLoggers")
   final Logger log;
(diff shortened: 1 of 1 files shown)


StefanBratanov (approved):
LGTM

infrastructure/logging/src/main/java/tech/pegasys/teku/infrastructure/logging/StatusLogger.java
```diff
@@ -36,6 +36,7 @@ public class StatusLogger {
 
   public static final StatusLogger STATUS_LOG =
       new StatusLogger(LoggingConfigurator.STATUS_LOGGER_NAME);
+  public static final int KEY_LIMIT = 20;
```
StefanBratanov, line 39: Just a nit but this can be private, also maybe can rename to be more descriptive to `PUBKEY_LOG_LIMIT`

PR #9290, opened 2025-03-27
Shutdown AsyncRunner when Eth1 polling is not needed

StefanBratanov:
## PR Description
If there is a retrying deposit fetch pending and we disable Eth1 polling, the deposit fetching would still be retried. It does make sense to shutdown the AsyncRunner to fully stop all pending powchain tasks.

image

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #9294, opened 2025-03-27
refactor: expose P2P mesh gauge metrics

### LINKED ISSUE #8769: Expose p2p mesh gauge metric
It could be interesting to have a metric(s) to track how many peers are in the p2p mesh of each subscribed topic.
An addition to that could be to track how many peers "left" the mesh and how many peers "joined" the mesh between two sampling periods.

The problem is that currently the mesh is exposed `gossipRouter.mesh()` and it gives back the underlining mutable Map. Since that is a `LinkedHashMap` created via:
```
val mesh: MutableMap<Topic, MutableSet<PeerHandler>> = linkedMapOf()
```
and is not thread safe we can't safely access it to collect the data we want.

We need a more secure way of accessing the information (it could require a libp2p change)

crStiv:
## PR Description
Added metrics to track P2P mesh peers:
- `mesh_peers_count` gauge for current number of peers per topic
- `mesh_peers_joined_total` counter for peers joining mesh
- `mesh_peers_left_total` counter for peers leaving mesh

## Fixed Issue(s)
Fixes #8769


mehdi-aouadi:
Closing since the CI steps are failing and have not been fixed

PR #9298, opened 2025-03-28
Add version of Teku to VC metrics

### LINKED ISSUE #9281: add teku version to the teku validator client metrics
it looks like VC metrics isn't reporting teku version, but it'd be helpful for maintenance if it can be updated.

StefanBratanov:
## PR Description
Adds `remote_validator_teku_version_total` to standalone VC metrics which is equivalent to `beacon_teku_version_total` when running a Beacon Node

image

## Fixed Issue(s)
fix #9281 

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
@@ -29,6 +29,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
+import org.hyperledger.besu.plugin.services.metrics.Counter;
+import org.hyperledger.besu.plugin.services.metrics.LabelledMetric;
 import tech.pegasys.teku.bls.BLSPublicKey;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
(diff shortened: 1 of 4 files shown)


rolfyone:
oh i was hoping it could be the same metric... should be ok this way...

StefanBratanov:
> oh i was hoping it could be the same metric... should be ok this way...

Yeah, I wanted to keep it the same but semantically doesn't make sense to call it "beacon_..."

validator/client/src/main/java/tech/pegasys/teku/validator/client/ValidatorClientService.java
```diff
@@ -566,6 +574,18 @@ private static void addValidatorCountMetric(
         validators::getValidatorCount);
   }
 
+  /** Only available when running a standalone VC client */
+  private static void addVersionMetric(final MetricsSystem metricsSystem) {
+    final String version = VersionProvider.IMPLEMENTATION_VERSION.replaceAll("^v", "");
+    final LabelledMetric<Counter> versionCounter =
+        metricsSystem.createLabelledCounter(
+            TekuMetricCategory.REMOTE_VALIDATOR,
```
rolfyone: i cant find any other metric we have that starts with `remote_`, so we might be better off using `validator_teku_version_total`
```
# HELP beacon_teku_version_total Teku version in use
# TYPE beacon_teku_version_total counter
beacon_teku_version_total{version="25.3.0+33-g1af17ee2be"} 1.0
```

StefanBratanov: Makes sense actually, I will change it and remote REMOTE_VALIDATOR from the enun

tbenr (approved):
LGTM

PR #9302, opened 2025-03-30
Add Lodestar bootnode to Hoodi network configuration

VolodymyrBg:
The Lodestar bootnode is added to the `applyHoodiNetworkDefaults` method in the Eth2NetworkConfiguration class, complementing the existing EF and Teku bootnodes.


lucassaldanha:
Thanks, but we won't add those right now. We will wait until all teams have their release and stable infrastructure (it can change too often). I suggest next time reaching out on teku-contributors Discord before starting working on something in Teku. Or looking at one of the tickets with "good first issue" label.

On a side note, the code is not even compiling so it is a good idea to run a local build before submitting a PR to catch some of these mistakes.

PR #9318, opened 2025-04-03
Add read functionality to blobs archiving

### LINKED ISSUE #8513: Allow the option to store blobs to filesystem as they're being removed from database

StefanBratanov:
## PR Description

- Store slot -> block root indexes in 100_000 rows file chunks
- Retrieval methods by `SlotAndBlockRoot` and by only slot.
- Changes to CombinedChainDataClient to retrieve from archive

## Fixed Issue(s)
fixes #8513 

## Documentation

## Changelog


infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/TekuMetricCategory.java
```diff
@@ -27,7 +27,6 @@ public enum TekuMetricCategory implements MetricCategory {
   STORAGE("storage"),
   STORAGE_HOT_DB("storage_hot"),
   STORAGE_FINALIZED_DB("storage_finalized"),
-  REMOTE_VALIDATOR("remote_validator"),
```
tbenr, line 30: is this intentional?

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableListTypeDefinition.java
```diff
@@ -26,4 +26,8 @@ public DeserializableListTypeDefinition(
       final Optional<Integer> maxItems) {
     super(itemType, Function.identity(), minItems, maxItems);
   }
+
+  public DeserializableListTypeDefinition(final DeserializableTypeDefinition<T> itemType) {
```
tbenr: is this actually used?

tbenr (approved):
LGTM.

WDYT about adding a test in GetBlobSidecarsIntegrationTest ?

infrastructure/metrics/src/main/java/tech/pegasys/teku/infrastructure/metrics/TekuMetricCategory.java
StefanBratanov: Yes, it is not used anywhere. Even though not really relevant to this PR.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableListTypeDefinition.java
StefanBratanov: Yeah, in the Archiver

StefanBratanov:
> LGTM.
> 
> WDYT about adding a test in GetBlobSidecarsIntegrationTest ?

Makes sense to me. Will add it quickly tomorrow.

infrastructure/json/src/main/java/tech/pegasys/teku/infrastructure/json/types/DeserializableListTypeDefinition.java
tbenr: weird, intellij wasn't finding any reference

StefanBratanov: you are right, it's not, removed

PR #9334, opened 2025-04-09
Get blobs v2

zilm13:
## PR Description

API and interfaces for engine_getBlobsV2 and engine_getPayloadV5
## Documentation

## Changelog


zilm13:
updated according to this change in EL behavior:
```
Teams agreed to implement empty-or-full response behavior for getBlobsV2, meaning:

    Return empty response if any requested blobs are missing
    Only return data when all requested blobs are available
    No partial responses will be supported
```

lucassaldanha (approved):

PR #9344, opened 2025-04-13
fix: issue with checking empty variables

mdqst:
## PR Description

noticed that the original method `${!1+set}` wasn't handling empty variables properly, which could cause some issues.
so, I replaced it with `if [ -z "${!1}" ]` - this ensures that we correctly check if the variable is empty.

p.s. this should make the script more reliable when dealing with unset or empty variables.


rolfyone:
closing due to contribution guidelines.

PR #9351, opened 2025-04-14
partial 3rd party updates and errorprone updates

rolfyone:
I ended up disabling the instanceof errorprone, as it's very very common for us to be doing it apparently. I started changing things but in the interest of time will raise a ticket

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #9353, opened 2025-04-14
Make builder timeouts only for the HTTP call

### LINKED ISSUE #9327: More strict timeout application on builder API
We are actually applying the timeouts on engine\builder api using `orTimeout` on top of our rest client.
WRT `builderGetHeader` in particular,  the problem is that timeout will also include some of the work we do on our side: actual rest call preparation and, more importantly, ssz\json deserialization.

Due to timing games, the builder tends to respond very close to the timeout, so little spikes on our side could reduce the actual 1s timeout that is part of the spec.

StefanBratanov:
## PR Description
As per https://square.github.io/okio/3.x/okio/okio/okio/-timeout/timeout.html

## Fixed Issue(s)
fixes #9327 

## Documentation

## Changelog

### DIFF
--- a/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClientTest.java
+++ b/ethereum/executionclient/src/integration-test/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClientTest.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.core.JsonGenerationException;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import java.io.IOException;
+import java.io.InterruptedIOException;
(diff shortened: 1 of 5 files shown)


tbenr:
I'm worried by "as long as forward progress". So if the builder sends one byte every 500ms it can last a lot more than 1s

is `deadline` better for our use case? a good description is missing...

we really need to be sure we respect the timeout in all cases, this is the only thing that saves us from missing the block due to external builder latency

tbenr:
actually reading the javadoc it seems good (from `okhttp3.Call`)

image

the actual link to the doc is https://square.github.io/okhttp/5.x/okhttp/okhttp3/-call/timeout.html , even if I can't find the 4.x version of it. But I think we can trust the bundled lib javadoc

tbenr (approved):
LGTM

lucassaldanha (approved):

PR #9357, opened 2025-04-16
Avoid combining validators by aggregating committee when using DVT

### LINKED ISSUE #9347: Issue with DVT aggregation
From Discord:

Currently at Charon, if there are multiple validators for a slot that have aggregator duty, we receive only 1 signature from Teku VC, instead for all of them. While this makes complete sense in a traditional setup, where we wouldn't want to pollute the beacon node with the same aggregations, it breaks a bit the flow of a distributed validator. On our side we need at least threshold of signatures for a particular data in order to aggregate them. Hence, receiving 1 arbitrary signature doesn't do us a favour and it is up to the chance to receive the correct ones in threshold Charons in order to succeed in aggregating threm. (a bit of a mouthful with aggregating the aggregation duty's partial signatures 😅)

We have a kurtosis repository, which you can use to spin up a CL-Charon-VC combination - https://github.com/ObolNetwork/kurtosis-charon. I would suggest branch kalo/v1.3.0-testing, as it's with the most up to date versions compared to other branches. You need Kurtosis CLI installed - https://docs.kurtosis.com/install. After that, a simple make geth-teku-charon-teku should do the job. In any of the Charon containers' logs you will see aggregations failing after a bit:
node0  | 16:30:58.013 WARN tracker    Duty failed                              {"step": "parsig_db_external", "reason": "insufficient partial signatures received, minimum required threshold not reached", "reason_code": "insufficient_peer_signatures", "duty": "298/aggregator"}.

And if you want to trace that slot 298 aggregator duty, you will see the logs in the Charon instance like the ones in the attached file. There you can see we have 7 validators that need to aggregate, but on charon we have only received signature for validator a39_bda from peer 1, signature for validator 976_eab from peer 2 and signature for b7c_87d from our own Teku VC.

lucassaldanha:
## PR Description
By default, when we have multiple validators aggregating on the same committee, Teku VC optimizes this by only creating a single `AggregateAndProof` that is sent to the Beacon Node. However, when using DVT, it is important that each validator submits their own `AggregateAndProof`, regardless if they are to the exact same slot/committee.

In this change, `AggregationDuty` has been updated with an abstraction for the aggregator set called `AggregationDutyAggregators`. This interface hides the underlying logic of combining validators in the same committee or not from the `AggregationDuty` logic.

## Fixed Issue(s)
fixes #9347 

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/AggregationDutyTest.java
+++ b/validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/AggregationDutyTest.java
@@ -61,7 +61,9 @@ import tech.pegasys.teku.validator.api.ValidatorApiChannel;
 import tech.pegasys.teku.validator.client.ForkProvider;
 import tech.pegasys.teku.validator.client.Validator;
 import tech.pegasys.teku.validator.client.duties.attestations.AggregationDuty;
+import tech.pegasys.teku.validator.client.duties.attestations.AggregatorsGroupedByCommittee;
 import tech.pegasys.teku.validator.client.duties.attestations.BatchAttestationSendingStrategy;
+import tech.pegasys.teku.validator.client.duties.attestations.UngroupedAggregators;
 
 @TestSpecContext(milestone = {PHASE0, ELECTRA})
 class AggregationDutyTest {
@@ -108,6 +110,7 @@ class AggregationDutyTest {
             spec,
             SLOT,
             validatorApiChannel,
+            new AggregatorsGroupedByCommittee(),
             forkProvider,
             validatorLogger,
             new BatchAttestationSendingStrategy<>(validatorApiChannel::sendAggregateAndProofs),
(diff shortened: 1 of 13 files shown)


lucassaldanha:
Validated the fix using charon via kurtosis setup. Everything seems to be stable now.

validator/client/src/test/java/tech/pegasys/teku/validator/client/duties/AggregationDutyTest.java
```diff
@@ -290,6 +293,83 @@ public void shouldProduceSingleAggregateAndProofWhenMultipleValidatorsAggregateS
         .record(any(), any(AggregationDuty.class), eq(ValidatorDutyMetricsSteps.SIGN));
   }
 
+  @TestTemplate
+  public void
+      shouldProduceAllAggregateAndProofsWhenValidatorsAggregateSameCommitteeAndUsingUngroupedAggregators() {
```
rolfyone: this is probably at the point where a comment would be better than a ridiculous name...

rolfyone (approved):
mostly fine but that name is out of control.

PR #9360, opened 2025-04-16
builder API, make json media type more compatible

### LINKED ISSUE #9316: upgrade flashbots to 1.9-rc3 breaks registration call
Currently getting `Invalid remote response from the builder: {\"code\":502,\"message\":\"no successful relay response` when uppgrading to rc3. using rc2 seems to resolve the issue. This might actually be an issue on the flashbots side. or just a lack of relays in the network (?)

tbenr:
It seems that mevboost 1.9.rc3 forwards the registration call as it comes from the CL side.
So it seems that the sepolia relay we were using was rejecting the call due to the additional `charset` in the content type, and returning 415. mevboost was translating 415 to a 502 on CL side.

This is a relay issue but to avoid incompatibilities it is better to remove the useless `charset`

fixes #9316
related to https://github.com/flashbots/mev-boost/issues/766
## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM.
Looks redundant anyway since the default (if not the only) encoding for JSON is UTF-8

StefanBratanov (commented):
LGTM

PR #9362, opened 2025-04-17
New infra stream module

lucassaldanha:
## PR Description

Another module used on the das branch that is independent and can be moved over.

I have run coverage tests and these are mostly covered by unit tests (see report below). The exception seems to be `VisitorHandler` that seems to be unused. I'll keep it in for now in case it is used in the future. I'll raise a note to check its testing if we do end up using it.

Screenshot 2025-04-17 at 13 21 36

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/stream/AsyncStreamPublisherTest.java
+++ b/infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/stream/AsyncStreamPublisherTest.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright Consensys Software Inc., 2024
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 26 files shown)


zilm13 (approved):
LGTM

infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/stream/AsyncStreamPublisherTest.java
```diff
@@ -0,0 +1,169 @@
...
+    publisher.onError(new RuntimeException("test"));
+
+    assertThat(stream.toList()).isCompletedExceptionally();
+  }
+
+  @Test
+  void shouldIgnoreAnyItemsAfterOnComplete() {
+    publisher.onNext(0);
+    publisher.onComplete();
+    publisher.onNext(1);
```
zilm13, line 122: Canonically I think this should throw but let's left it as is

PR #9373, opened 2025-04-25
Add FULU to the Spec Factory

mehdi-aouadi:
## PR Description
Add FULU to the Spec Factory

## Documentation

## Changelog


StefanBratanov (approved):
Good catch. LGTM.

mehdi-aouadi:
Had to add other changes in the end to make the acceptance tests pass (the get block API call was failing). Maybe @lucassaldanha can have a look to make sure it doesn't mess the ongoing fulu plan

lucassaldanha:
Some of these changes are already in the pipeline for migration, so I won't merge this PR for now. Thanks for keeping an eye out though :)

PR #9374, opened 2025-04-25
Batch attestation duty scheduling for an epoch

### LINKED ISSUE #9292: Be smarter in scheduling aggregation duty
When running several thousands of keys, the validator client is currently  signing the slot for every validator at the beginning of each epoch.

https://github.com/Consensys/teku/blob/839a13b1a9bb98b3ed1928d1b4f4a36c2d6eeee3/validator/client/src/main/java/tech/pegasys/teku/validator/client/AttestationDutyLoader.java#L168

Due to this, if we end up being the proposer of the first slot, we could have the singing worker filled up and we could experience a delay (we are seeing even 3s) signing the block.

We could gradually pre-compute those signatures throughout the epoch and have them ready (at least the majority of them) when its time to schedule.

StefanBratanov:
## PR Description
Introduce `AttestationDutyDefaultSchedulingStrategy` and `AttestationDutyBatchSchedulingStrategy`

Batching by slots with added delays is used when there are lot of duties to be scheduled (excessive signing at the beginning of an epoch)

## Fixed Issue(s)
fixes #9292 

## Documentation

## Changelog

### DIFF
--- a/validator/client/src/main/java/tech/pegasys/teku/validator/client/AttestationDutyLoader.java
+++ b/validator/client/src/main/java/tech/pegasys/teku/validator/client/AttestationDutyLoader.java
@@ -13,119 +13,29 @@
 
 package tech.pegasys.teku.validator.client;
 
-import com.google.common.annotations.VisibleForTesting;
 import it.unimi.dsi.fastutil.ints.IntCollection;
-import java.time.Duration;
-import java.util.List;
-import java.util.Map;
 import java.util.Optional;
-import java.util.TreeMap;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.apache.tuweni.bytes.Bytes32;
-import tech.pegasys.teku.api.response.ValidatorStatus;
-import tech.pegasys.teku.bls.BLSPublicKey;
 import tech.pegasys.teku.ethereum.json.types.validator.AttesterDuties;
-import tech.pegasys.teku.ethereum.json.types.validator.AttesterDuty;
-import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
-import tech.pegasys.teku.spec.Spec;
-import tech.pegasys.teku.spec.SpecVersion;
-import tech.pegasys.teku.spec.datastructures.operations.AttestationData;
-import tech.pegasys.teku.spec.datastructures.operations.AttesterSlashing;
-import tech.pegasys.teku.spec.datastructures.operations.ProposerSlashing;
-import tech.pegasys.teku.validator.api.CommitteeSubscriptionRequest;
 import tech.pegasys.teku.validator.api.ValidatorApiChannel;
-import tech.pegasys.teku.validator.api.ValidatorTimingChannel;
-import tech.pegasys.teku.validator.client.duties.BeaconCommitteeSubscriptions;
 import tech.pegasys.teku.validator.client.duties.SlotBasedScheduledDuties;
-import tech.pegasys.teku.validator.client.duties.attestations.AggregationDuty;
-import tech.pegasys.teku.validator.client.duties.attestations.AttestationProductionDuty;
 import tech.pegasys.teku.validator.client.loader.OwnedValidators;
 
 public class AttestationDutyLoader
-    extends AbstractDutyLoader<AttesterDuties, SlotBasedScheduledDuties<?, ?>>
-    implements ValidatorTimingChannel {
-
-  private static final Logger LOG = LogManager.getLogger();
-
-  // Attestation duty scheduling is batched by slots and delays are added in order to avoid
-  // expensive aggregation slot signing in beginning of the epoch when a node is running a large
-  // number of validators
-  @VisibleForTesting
-  record SlotBatchingOptions(
-      int minSizeToBatchBySlot,
-      int currentEpochSlotsToScheduleBeforeDelay,
-      Duration currentEpochSchedulingDelay,
-      int futureEpochSlotsToScheduleBeforeDelay,
-      Duration futureEpochSchedulingDelay) {}
-
-  private static final SlotBatchingOptions DEFAULT_BATCHING_OPTIONS =
-      new SlotBatchingOptions(1000, 4, Duration.ofMillis(500), 1, Duration.ofSeconds(1));
-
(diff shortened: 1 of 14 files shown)


validator/client/src/main/java/tech/pegasys/teku/validator/client/AttestationDutyLoader.java
```diff
@@ -38,6 +43,12 @@ public class AttestationDutyLoader
     extends AbstractDutyLoader<AttesterDuties, SlotBasedScheduledDuties<?, ?>> {
 
   private static final Logger LOG = LogManager.getLogger();
+
+  // Attestation duty scheduling is batched in order to avoid expensive aggregation slot signing in
+  // beginning of the epoch when a node is running a large number of validators
+  private static final int BATCH_SIZE = 2500;
+  private static final Duration BATCH_DELAY = Duration.ofSeconds(3);
```
rolfyone: i was thinking more slots...

2500 is a lot (but may be ok if its configurable)

Theres some use cases

 - absolute worst case is bootstrapping and a large validator set. (would include a range sync here)
 - best case is we're already running, because we're just padding out our current set of duties

im not sure if we need to handle these separately, but basically, i think something slot based should solve both scenarios

 - any slots in the 'past' (by time computation) can be excluded
 - current slot needs to be known ASAP (if we're late into that we may as well not sub)
 - assuming current slot is not in our list of duties (this would be if we already have an epoch of duties computed and we're filling in epoch+1), then we can just do 1 slot of signatures and then sleep 1, then do another slot.
 - if we're computing current epoch duties, then we probably need to be doing 3-4 slots in a batch...

so at start, we have 2 epochs of duties to figure out if we're aggregators
 - exclude all 'historic' slots
 - order by slot
 - if we're current epoch, iterate 1 slot at a time for signatures. this gives gaps while we wait for that slot to complete if we're trying to get current duties in. sleep 500ms every period (3-4 slots)
   - so if we have slots 10-32 of our current to consider, then we iterate those in batches, but 1 slot at a time (sleeping 1s every ... say 3rd-4th... slot)
 - from 33-64, we would just do a single slot then sleep 1s. this would spread the sigs over a few slots.

If we have less than about 1k validators we likely don't care for this overhead, but once we get to large numbers like some of our test infra, this would help a lot.

On 32k validators, that looks like on average 1k / slot, so by spacing them out we get space to do things like sign actual duties.

Concretely,
eg1.
slot 32, i  have already computed my current epoch (from the previous time) so i get only my future epoch. I have 32k sigs to do so i do 1k (average) sleep 1s, repeat and it takes me 32 seconds to get slots 33-64 aggregation duties.

eg2.
Slot 1, if i haven't computed any duties before, then i have 64 duties (current and next epoch)
i do slot 1 computes straight away, in the hope i'm not too late. then without a rest i do 2, then 3. that gives me all the data i need for the first few slots.
I sleep 500ms, do slots 4,5, then 6. rinse, repeat until i get to 33. (would take... 6 seconds ish of sleeps)
at 33, i do that, sleep 1s. 34, sleep 1s. repeat. (would take over half a minute not including signing time)

eg3.
slot 29, i haven't computed duties before, i get duties from current/next epoch.
I exclude 1-28, as they're historic and i've already missed those aggregating tasks. 
I compute for 29. compute 30, 31, sleep 500ms.
then compute 32 which is last of current epoch, so sleep 500ms.
then 33, sleep 1s etc.

I guess theres still the edge case where you have more than 1k sigs per slot but i think at that point we have bigger issues...

This was a big comment, but kind of encapsulates some considerations i guess.

validator/client/src/main/java/tech/pegasys/teku/validator/client/AttestationDutyLoader.java
```diff
@@ -81,26 +108,68 @@ protected SafeFuture<Optional<AttesterDuties>> requestDuties(
...
-                        scheduleDuties(scheduledDuties, duty, dvtAttestationAggregationsForEpoch))
-                .toArray(SafeFuture[]::new))
+    if (duties.getDuties().size() < MIN_SIZE_TO_BATCH_BY_SLOT) {
+      return scheduleDuties(scheduledDuties, duties.getDuties(), dvtAttestationAggregations)
+          .<SlotBasedScheduledDuties<?, ?>>thenApply(__ -> scheduledDuties)
+          .alwaysRun(beaconCommitteeSubscriptions::sendRequests);
+    }
+
+    final boolean isCurrentEpoch =
+        spec.computeEpochAtSlot(currentSlot.get()).isLessThanOrEqualTo(epoch);
```
rolfyone: ahm... so current epoch should be just `equals`?

validator/client/src/main/java/tech/pegasys/teku/validator/client/AttestationDutyLoader.java
```diff
@@ -81,26 +108,68 @@ protected SafeFuture<Optional<AttesterDuties>> requestDuties(
...
+
+    final Map<UInt64, List<AttesterDuty>> dutiesBySlot =
+        duties.getDuties().stream()
+            .collect(
+                Collectors.groupingBy(AttesterDuty::getSlot, TreeMap::new, Collectors.toList()));
+
+    SafeFuture<Void> dutiesScheduling = SafeFuture.COMPLETE;
+
+    int i = 0;
```
rolfyone: i'd say we'll want to break this down a little if we're going to have a solid chance of unit testing...

validator/client/src/main/java/tech/pegasys/teku/validator/client/AttestationDutyLoader.java
```diff
@@ -13,31 +13,62 @@
...
+    extends AbstractDutyLoader<AttesterDuties, SlotBasedScheduledDuties<?, ?>>
+    implements ValidatorTimingChannel {
 
   private static final Logger LOG = LogManager.getLogger();
+
+  // Attestation duty scheduling is batched by slots and delays are added in order to avoid
+  // expensive aggregation slot signing in beginning of the epoch when a node is running a large
+  // number of validators
+  @VisibleForTesting
+  record SlotBatchingOptions(
```
rolfyone: oh i like that idea

rolfyone (approved):
LGTM, we should see how it goes. I wonder if we also want to take timings of each batch between sleeps but that could definitely be separate to this PR.

StefanBratanov:
@lucassaldanha I am not 100% sure how `DvtAttestationAggregations` works but it looks like `getCombinedSelectionProofFuture` does not complete until all other duties reach that stage. In the case of batching, I am not sure how to make this work. Any idea or I am misunderstanding possibly how it works.

StefanBratanov:
**After:**
Slow Block Publishing *** Slot: 418048 start 728ms
**Before:**
Slow Block Publishing *** Slot: 400512 start 5263ms

PR #9375, opened 2025-04-28
Added KZG computeCells method

lucassaldanha:
## PR Description

Added `computeCells` method to our KZG implementation to support Distributed cell proofs.

Created a "real" NZG no-operation implementation `NoOpKZG`, it only exist on `infra.kzg.testFixtures` so it won't make its way into production code.

Renamed KZG.NOOP to KZG.DISABLED to differentiate from the no-operation one (used by tests). This is a bit more consistent with the idea of both: no-operation implementation being safe for use on testing stubs etc; disabled means it will break if used.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9377, opened 2025-04-28
cleanup the rest of api/schema

rolfyone:
- Removes deprecated types
 - A few types got relocated as we still use them elsewhere

 This should be the bulk of the cleanup task. Certainly a huge reduction from this module.

## Documentation

## Changelog


data/serializer/src/main/java/tech/pegasys/teku/api/response/ValidatorStatusUtil.java
```diff
@@ -0,0 +1,64 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.api.response;
+
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.datastructures.state.Validator;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+
+public class ValidatorStatusUtil {
```
tbenr, line 20: should we add unit test for this as there was for ValidatorResponse ?

tbenr (approved):
LGTM. just a question

data/serializer/src/main/java/tech/pegasys/teku/api/response/ValidatorStatusUtil.java
rolfyone: it was there, i definitely didnt want to lose it. i renamed it though so its going to be able to be found now.
Noticed EventTypeTest was in the wrong home now too, also fixed.

PR #9388, opened 2025-05-02
may 3rd party updates

rolfyone:
## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ plugins {
   id 'com.github.ben-manes.versions' version '0.52.0'
   id 'com.github.jk1.dependency-license-report' version '2.9'
   id 'io.spring.dependency-management' version '1.1.7'
-  id 'net.ltgt.errorprone' version '4.1.0' apply false
+  id 'net.ltgt.errorprone' version '4.2.0' apply false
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -25,7 +25,7 @@ plugins {
   id 'com.github.ben-manes.versions' version '0.52.0'
   id 'com.github.jk1.dependency-license-report' version '2.9'
   id 'io.spring.dependency-management' version '1.1.7'
-  id 'net.ltgt.errorprone' version '4.1.0' apply false
```
gfukushima, line 28: do we still want to use the apply false?

gfukushima (approved):
lgtm, left a question regarding the `apply` we're using on errorprone

build.gradle
rolfyone: i have no idea what it even does

lucassaldanha: We should keep using it. It is just a workaround to prevent potential classpath issues and loading this plugin multiple times. I don't remember the details, but because error-prone ties into the compilation of the project, bad things can happen if you end up loading it multiple times.

TLDR This is a simple way to prevent it, and we probably don't want to remove it unless we really know what we are doing :D

Issue #9365, opened 2025-04-17
[EIP-7892] BPO for peerDAS


tbenr:
we should take a look into what does it mean for us to support BPO

https://eips.ethereum.org/EIPS/eip-7892

rolfyone:
i think for this we'll just need to bite the bullet and change from our generic reading to string.

I got everything reading, but then it was converting things like 0xaaaa -> numeric, and if we instead define those specific types as string we'd be able to have a better level of control.

This would remove all of the hackery around our SpecConfigReader, but it does mean rewriting that area and how we get to the point of being able to set config parameters.

The advantage of SpecConfigReader in this case is going to be its weakness, because we now just don't get to push everything to string, as we're introducing lists and objects to the maps.

Basically the fun starts at `SpecConfigReader::readValues`
where we 
```
      return (Map<String, String>)
          mapper
              .readerFor(
                  mapper.getTypeFactory().constructMapType(Map.class, String.class, String.class))
              .readValues(source)
              .next();
```

This forces the reading of every single field into a string map, and that's what is no longer appropriate.

From there there's field renaming and config setting/defaulting, and this would need to be updated if we ended up going and defining all the object relationships.

We want to keep our tolerance of defaulting etc, its tempting to be able to declare things like we do for `DeserializableTypeDefinition`, and have a more direct mapping into our config.

rolfyone:
ok @lucassaldanha is more creative than me today and found a simple solution, so we should be good.

PR #9390, opened 2025-05-05
Added snakeyaml to allow more complicated config reading

rolfyone:
- added tests to show reading lists of objects (BPO related)
 - added a testcase to show that hex is read correctly
 - added testcase to show specConfigReader isnt filling defaults when we read local config files

 Partially addresses #9365

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/YamlConfigReader.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/YamlConfigReader.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright Consensys Software Inc., 2025
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.spec.config;
+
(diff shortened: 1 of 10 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/YamlConfigReader.java
```diff
@@ -0,0 +1,55 @@
...
+    if (values == null) {
+      throw new NoSuchElementException();
+    }
+    return values;
+  }
+
+  public static class CustomResolver extends Resolver {
+    @Override
+    protected void addImplicitResolvers() {
+      addImplicitResolver(Tag.STR, INT, "+-0123456789.");
```
lucassaldanha, line 57: NIT: I think it is worth adding a comment re how if we don't do this, hex strings will be read as numbers. Just a reminder of what we had to go through.

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigLoaderTest.java
```diff
@@ -46,16 +44,16 @@ public void shouldLoadAllKnownNetworks(final Eth2Network network) throws Excepti
     assertAllFieldsSet(config, SpecConfigElectra.class);
   }
 
-  /**
+  /***
```
lucassaldanha: What's up with the 3 `*`? :)

lucassaldanha (approved):
LGTM. Just a few nits.

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigLoaderTest.java
rolfyone: because spotless complained loudly, i dont know.

rolfyone: will see if its less unhappy now, but it was adamant.

PR #9392, opened 2025-05-06
feat: Allow multiple entries in graffiti file via YAML configuration

### LINKED ISSUE #3471: Allow multiple entries in graffiti file
Now that graffiti can be loaded from file, we should consider allowing multiple entries in the graffiti file.

Prysm use cases: https://docs.prylabs.network/docs/prysm-usage/graffiti-file/

There's some interesting use-cases listed where randomising, or having specific per-validator graffiti might help obfuscate validators being run, or indeed for the graffiti wall.

It's worth having a discussion about use-cases and a proposed solution.

https://discord.com/channels/697535391594446898/697539289042649190/801199770956726282

CreeptoGengar:
## PR Description
Adds support for multiple graffiti entries in a YAML configuration file, similar to Prysm's functionality. This implementation allows validators to use different graffiti messages based on various criteria:
- Specific graffiti for individual validators by public key
- Ordered list of graffiti messages used in sequence
- Random selection from a list of graffiti messages
- Default fallback graffiti

Maintains backward compatibility with the existing single-value text format.

## Fixed Issue(s)
fixes #3471

## Documentation

## Changelog


validator/api/src/test/java/tech/pegasys/teku/validator/api/GraffitiParserTest.java
```diff
@@ -57,4 +74,65 @@ void testGraffitiParserDoesTrim() throws Exception {
...
+        GraffitiParser.loadGraffitiProvider(
+            Optional.of(defaultGraffiti), Optional.of(publicKey), Optional.empty());
+
+    assertThat(provider).isInstanceOf(FileBackedGraffitiProvider.class);
+  }
+
+  @Test
+  void loadGraffitiProvider_shouldCreateMultimodeProviderForYamlFile(final @TempDir Path tempDir)
+      throws IOException {
+    // Создаем временный YAML файл
```
rolfyone: i would need english comments sorry

rolfyone:
Thanks for raising this @CreeptoGengar , I'm just starting to have a look at the details here.

To fix `spotless` task, you can run `gradle spotlessJavaApply` in the teku folder and it'll update some files to conform to our coding standards and you'll be able to commit those changes...

validator/api/src/test/java/tech/pegasys/teku/validator/api/GraffitiParserTest.java
```diff
@@ -57,4 +74,65 @@ void testGraffitiParserDoesTrim() throws Exception {
...
+    final Path yamlFile = tempDir.resolve("graffiti.yaml");
+
+    // Настраиваем конфигурацию
+    MultimodeGraffitiProvider.GraffitiConfiguration config =
```
rolfyone: Im fairly sure the rules will want 'final' variables everywhere possible

validator/api/src/test/java/tech/pegasys/teku/validator/api/MultimodeGraffitiProviderTest.java
```diff
@@ -0,0 +1,239 @@
...
+    // Setup config with specific validator entry
+    MultimodeGraffitiProvider.GraffitiConfiguration config =
+        new MultimodeGraffitiProvider.GraffitiConfiguration();
+    Map<String, String> specific = new HashMap<>();
+    specific.put(pubKeyHex, "specific validator graffiti");
+    config.specific = specific;
+    config.defaultGraffiti = "file default graffiti";
+
+    // Write config to file
+    new ObjectMapper(new YAMLFactory()).writeValue(graffitiFile.toFile(), config);
```
rolfyone: probably cleanest to just have a single ObjectMapper used by the test suite

docs/multimode-graffiti.md
lucassaldanha, line 1: Our documentation for Teku is done in a separate repo. Take a look at https://github.com/Consensys/doc.teku

CreeptoGengar:
@rolfyone do i need to make ci/circleci: acceptanceTests green? also corrected what you asked me for

@lucassaldanha tnx for pointing this out, deleted the docs directory

rolfyone:
> @rolfyone do i need to make ci/circleci: acceptanceTests green? also corrected what you asked me for
ta - will have a look. The AT were green when they ran last, if they break i can see if they're related to your change etc...

validator/api/src/main/java/tech/pegasys/teku/validator/api/MultimodeGraffitiProvider.java
```diff
@@ -0,0 +1,169 @@
...
+      final Optional<Bytes32> defaultGraffiti,
+      final Optional<BLSPublicKey> validatorPublicKey,
+      final Optional<Path> graffitiFile) {
+    this.defaultGraffiti = defaultGraffiti;
+    this.validatorPublicKey = validatorPublicKey;
+    this.graffitiFile = graffitiFile;
+  }
+
+  @Override
+  public Optional<Bytes32> get() {
```
lucassaldanha, line 60: We need to make sure this method can never interfere with block production. While I was testing it, I noticed that any errors were bubbling up and not being handled here. The result was a failure in block production!

We need to ensure that any errors that happen inside `get()` are caught and the method returns Optional.empty().

One example on how I managed to trigger this was to set `defaultGraffiti:` with a value larger then 32 bytes.

lucassaldanha:
Reviewing this code, I have identified something we could do to improve our handling of errors when getting the graffiti from any `GraffitiProvider` implementation.

In `Validator.java` we do the following:
```
public Optional<Bytes32> getGraffiti() {
    return graffitiProvider.get();
}
```

This means that each `GraffitiProvider` implementation must properly handle any errors, risking it bubbling up all the way and eventually impacting block production. To be on the safer side, I recon we should change it so we capture ANY errors from `GraffitiProvider#get` and, upon an exception, we return empty. Something like this:

```
public Optional<Bytes32> getGraffiti() {
    try {
        return graffitiProvider.get();
    } catch(final Exception e) {
        LOG.warn("Error getting graffiti from graffiti provider: {}", e);
        return Optional.empty();
    }
}
```

This way we know that regardless of what happens on `GraffitiProvider`, we won't impact our block production. Of course we would like to have unit tests to verify this behaviour.

CreeptoGengar:

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -138,27 +138,6 @@ public class ValidatorOptions {
...
-      names = {"--Xvalidator-client-beacon-api-readiness-executor-threads"},
-      paramLabel = "<INTEGER>",
-      showDefaultValue = Visibility.ALWAYS,
-      description =
-          "Set the number of threads for the validator beacon node API readiness executor",
-      hidden = true,
-      converter = OptionalIntConverter.class,
-      arity = "1")
-  private OptionalInt beaconApiReadinessExecutorThreads = OptionalInt.empty();
-
```
lucassaldanha, line 161: This is likely unintentional and part of a bad merge.

teku/src/main/java/tech/pegasys/teku/cli/options/ValidatorOptions.java
```diff
@@ -191,26 +170,27 @@ public class ValidatorOptions {
...
+              .validatorKeystoreLockingEnabled(validatorKeystoreLockingEnabled)
+              .validatorPerformanceTrackingMode(validatorPerformanceTrackingMode)
+              .validatorExternalSignerSlashingProtectionEnabled(
+                  validatorExternalSignerSlashingProtectionEnabled)
+              .isLocalSlashingProtectionSynchronizedModeEnabled(
+                  isLocalSlashingProtectionSynchronizedEnabled)
+              .graffitiProvider(
+                  GraffitiParser.loadGraffitiProvider(
+                      Optional.ofNullable(graffiti),
+                      Optional.empty(),
```
lucassaldanha, line 184: I tested the graffiti logic, and it seems to be working for 'ordered', 'random' and 'defaultGraffiti'. However the `specific` by validator index or validator pub key isn't working. Setting this to empty here likely means we never check for a validator pub key?

On that note, I believe having a single optional parameter for pub key is wrong because we would want to support mapping a different graffiti to more than one validator.

validator/api/src/main/java/tech/pegasys/teku/validator/api/MultimodeGraffitiProvider.java
```diff
@@ -0,0 +1,174 @@
...
+    try {
+      // Check if we need to reload the file
+      checkAndReloadFileIfNeeded();
+
+      // If we have a configuration from file, try to get graffiti from it
+      if (graffitiConfig.isPresent()) {
+        final GraffitiConfiguration config = graffitiConfig.get();
+
+        // Priority 1: Check for specific validator entry if we have a public key
+        if (validatorPublicKey.isPresent()) {
```
lucassaldanha, line 70: This logic only works for a single validator. It does not conform with the idea of having many validator public key in the config. Probably needs a redesign of how we are instantiating this provider for a validator.

For some inspiration, check how we are dealing with an API-managed graffity (where we can set the graffiti per validator): https://github.com/Consensys/teku/blob/master/validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/MultithreadedValidatorLoader.java#L68-L75

lucassaldanha (changes requested):

lucassaldanha:
Closing PR due to inactivity.

PR #9402, opened 2025-05-08
Avoids potential mutability of the result of getCommitteeBits

tbenr:
It can't actually occur but being more defensive is better (and committeeBits is relatively small)

It also includes:
- a tiny simplification in parsing
- a better handling of copy()
- tests on aggregationBits sizing checks (including a fix in attestation generation in test)

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #9405, opened 2025-05-08
Added an info message for the highest milestone

### LINKED ISSUE #9400: report future forks that are defined on startup
Report all future fork epochs that are in the loaded spec on startup, and their activation epochs.

This should be present on both VC and BN, and will help detect when a VC loads without a fork epoch defined.

Context: currently its not possible to query this data from a VC like you could if it was a BN. Having this data in the logs will help during fork upgrades to see how a VC is configured.

rolfyone:
Also added a sanity check that will fail if the specFactory hasn't defined the highest milestone correctly, which has flow on effects.

fixes #9400

## Documentation

## Changelog


teku/src/main/java/tech/pegasys/teku/AbstractNode.java
```diff
@@ -175,8 +175,7 @@ private void reportBellatrixMergeOverrides(final TekuConfiguration tekuConfig) {
       reportTerminalBlockHashOverride = true;
       reportTerminalBlockHashEpochOverride = true;
     } else {
-      final Eth2NetworkConfiguration defaultConfiguration =
-          Eth2NetworkConfiguration.builder().applyNetworkDefaults(maybeEth2Network.get()).build();
+      final Eth2NetworkConfiguration defaultConfiguration = tekuConfig.eth2NetworkConfiguration();
```
rolfyone, line 178: this was interesting - we recreated spec here which was made evident when the log.info was added

lucassaldanha (approved):

PR #9406, opened 2025-05-09
added bpo parsing to configuration

rolfyone:
added tests, BPO information to each yaml.

Partially addresses #9365

## Documentation

## Changelog


rolfyone:
We'll now populate the fulu blob schedule from electra / deneb pre-schedule configurations that only have `max_blobs_per_block` etc defined.
I'm not sure how we go the other way yet but this is a logical point to review, as its arguably BPO compliant now i think.

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigReaderTest.java
```diff
@@ -192,12 +195,33 @@ public void read_invalidUInt64_tooLarge() {
...
   }
 
+  @Test
+  public void read_listOfBlobSchedules() {
+    final Map<String, Object> data = new HashMap<>();
+    data.put(
+        "BLOB_SCHEDULE",
+        List.of(
+            Map.of("EPOCH", "1", "MAX_BLOBS_PER_BLOCK", "2"),
+            Map.of("EPOCH", "3", "MAX_BLOBS_PER_BLOCK", "4")));
```
zilm13, line 205: Is the order important? I'd add it to the test

data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/config/GetSpecTest.java
```diff
@@ -64,12 +64,19 @@ void shouldGetCorrectMainnetConfig() throws Exception {
...
+    final JsonNode resultNode = mapper.readTree(json);
+    final JsonNode referenceNode =
+        mapper.readTree(
             Resources.toString(
                 Resources.getResource(GetSpecTest.class, "mainnetConfig.json"), UTF_8));
 
-    assertThat(result).containsExactlyInAnyOrderEntriesOf(expected);
+    // Ignoring the "data" field from the response, the new config should match the test
+    // configuration.
+    assertThat(resultNode.get("data")).isEqualTo(referenceNode);
```
zilm13: Nit: Maybe move `.get("data")` to definition of `resultNode`

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/builder/FuluBuilder.java
```diff
@@ -180,6 +191,22 @@ public void validate() {
     validateConstants();
   }
 
+  public void validateBlobSchedule(
```
zilm13, line 194: Maybe `validateBlobSchedule(
      final Optional<BlobSchedule> schedule)` and call it the same number of times how many schedules we have?

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigReaderTest.java
rolfyone: it wont be important because im fixing the spec when they back port :)

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/builder/FuluBuilder.java
rolfyone: we do call it once... but also we may not be able to have the schedule for a given fork because the variables will disappear. this is why its been done in this way, for compatibility.

zilm13: Ok, now I got it, looks good

zilm13 (approved):
LGTM

PR #9410, opened 2025-05-10
Improve jdk 24 support and add docker-jdk24

tbenr:
Added new param to avoid the following warning at startup in jdk 24

```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by org.fusesource.jansi.internal.JansiLoader in an unnamed module (file:/Users/khm/tools/maven/lib/jansi-2.4.1.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

Added a change in docker file on `jlink` command (`--add-modules ALL-MODULE-PATH` is not supported anymore)

default remains jdk21

## Documentation

## Changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -646,7 +646,7 @@ tasks.register("dockerDistUntar") {
 }
 
 def dockerImage = "consensys/teku"
-def dockerJdkVariants = [ "jdk21" ]
+def dockerJdkVariants = [ "jdk21", "jdk24" ]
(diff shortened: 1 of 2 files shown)


rolfyone (approved):
LGTM

PR #9412, opened 2025-05-11
Rename `AttestationBitsAggregator` to `AttestationBits`

tbenr:
Better reflect the usage (it is not only used to compute aggregations but also to just represent the bits)

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9427, opened 2025-05-16
BPO - remove max blobs from fulu spec

rolfyone:
## Documentation

## Changelog


rolfyone:
this should now be MVP for BPO.
 - no longer has MAX_BLOBS_PER_BLOCK_FULU
 - reads from the helper function to find blob limits from fulu
 I ran a test network and it looked to be working correctly, just needed to update the yaml to have a custom BPO. It's definitely not going to read BPO pre-fulu as it stands in this PR, but we can do that separately.
We'll want to sync DAS branch once this is done in case we need to do devnet0 from there.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/builder/FuluBuilder.java
```diff
@@ -110,7 +108,10 @@ public FuluBuilder kzgCommitmentsInclusionProofDepth(
   }
 
   public FuluBuilder blobSchedule(final List<BlobSchedule> blobSchedule) {
-    this.blobSchedule = blobSchedule;
+    checkNotNull(this.blobSchedule);
+    this.blobSchedule.clear();
+    // copy list rather than use the one passed in case we need to add to the list during validation
+    this.blobSchedule.addAll(blobSchedule);
```
zilm13, line 114: Like

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/versions/deneb/BlockContentsSchema.java
```diff
@@ -47,11 +48,18 @@ public BlockContentsSchema(
         namedSchema("block", schemaRegistry.get(BEACON_BLOCK_SCHEMA)),
         namedSchema(
             FIELD_KZG_PROOFS,
-            SszListSchema.create(SszKZGProofSchema.INSTANCE, specConfig.getMaxBlobsPerBlock())),
+            SszListSchema.create(SszKZGProofSchema.INSTANCE, maxBlobsOrCommitments(specConfig))),
```
zilm13: Looks I didn't explain this clearly, sorry. I think we should always use max blob commitments here, see, for example this structure (and any else in spec) https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/validator.md#blobsbundle

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
```diff
@@ -125,9 +128,21 @@ public int getMaxBlobsPerBlock(final UInt64 epoch) {
...
+        maybeSchedule
+            .map(BlobSchedule::maxBlobsPerBlock)
+            .orElseGet(() -> blobSchedule.getFirst().maxBlobsPerBlock());
+    LOG.debug("Max blobs at epoch {} found to be {}", epoch, maxBlobs);
+    return maxBlobs;
+  }
+
+  public int getHighestMaxBlobsPerBlockFromSchedule() {
+    checkArgument(!blobSchedule.isEmpty(), "Blob schedules not correctly defined.");
+    return blobSchedule.stream()
```
zilm13, line 142: why to sort it each time? I think we'd better expect it sorted

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
```diff
@@ -125,9 +128,21 @@ public int getMaxBlobsPerBlock(final UInt64 epoch) {
...
             .max(Comparator.comparing(BlobSchedule::epoch));
-    return maybeSchedule.isPresent()
-        ? maybeSchedule.get().maxBlobsPerBlock()
-        : blobSchedule.getFirst().maxBlobsPerBlock();
+
+    final int maxBlobs =
+        maybeSchedule
+            .map(BlobSchedule::maxBlobsPerBlock)
+            .orElseGet(() -> blobSchedule.getFirst().maxBlobsPerBlock());
+    LOG.debug("Max blobs at epoch {} found to be {}", epoch, maxBlobs);
```
zilm13, line 136: We will log it several times per each slot, maybe debug is too spammy for this?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -1095,4 +1108,22 @@ public boolean equals(final Object o) {
...
+
+  public Optional<Integer> getMaxBlobsPerBlockAtSlot(final UInt64 slot) {
+    final SpecVersion specVersion = atSlot(slot);
+    switch (specVersion.getMilestone()) {
+      case DENEB, ELECTRA -> {
+        return Optional.of(
+            specVersion.getConfig().toVersionDeneb().orElseThrow().getMaxBlobsPerBlock());
+      }
+      default -> {
+        if (specVersion.getMilestone().isGreaterThan(CAPELLA)) {
```
zilm13: why not greaterThanOrEqualTo(FULU)? It's a bit confusing with CAPELLA here and takes some cycles to realize DENEB and ELECTRA were already covered

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
```diff
@@ -80,12 +79,41 @@ public void shouldOverrideBlobRelatedValuesValues() {
...
-    final SpecConfigFulu fuluConfig =
-        specConfigAndParent.forMilestone(SpecMilestone.FULU).toVersionFulu().orElseThrow();
-
     assertThat(denebConfig.getMaxBlobsPerBlock()).isEqualTo(4);
     assertThat(electraConfig.getMaxBlobsPerBlock()).isEqualTo(8);
-    assertThat(fuluConfig.getMaxBlobsPerBlock()).isEqualTo(12);
+  }
+
+  @Test
+  public void shouldMatchElectraForkEpochBlobCount() {}
```
zilm13: Forgot?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -1095,4 +1108,22 @@ public boolean equals(final Object o) {
   public int hashCode() {
     return Objects.hash(forkSchedule);
   }
+
+  public Optional<Integer> getMaxBlobsPerBlockAtSlot(final UInt64 slot) {
```
zilm13, line 1112: It's used `.orElseThrow()` in all calls of this method. Does it make sense? Maybe return Integer and default -> throw?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/versions/deneb/BlockContentsSchema.java
rolfyone: we currently do for fulu but not for other branches, they're consistent deneb/deneb etc atm

this particular change makes a mess so i was thinking of doing it when we backport all the BPO to deneb/electra, or maybe before, but it breaks tests, removes OpenApiIntegrationTest references etc

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
rolfyone: orElseThrow because its a bug if you get there, have a deneb/electra/fulu milestone and dont have these things

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
rolfyone: because in the set of not deneb/electra is phase0, bellatrix, altair

once the implementation is fully cleaned up with deneb/electra then this becomes a single call to the BPO, but as default returning empty is what i wanted (pre-deneb fall to the final return)

this is all temporary until deneb / electra use bpo

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
rolfyone: well, i'd say if its getting super chatty we should probably be caching it somewhere... for minimal mode i was seeing it once per block

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
rolfyone: we're not sorting, we're range limiting

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
rolfyone: oh yes i think thats covered i should just delete that

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
rolfyone: for this PR that's kinda good if you've got debug enabled and you see it a lot and know something's not perfect i think.... with a tiny list the overhead is small but over time having tons of calls will be less cheap

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
rolfyone: as an aside we do a single sort during type initialization setting blobSchedule

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/blocks/versions/deneb/BlockContentsSchema.java
rolfyone: i changed that in this PR ultimately, will see if it passes AT

rolfyone: looks fine - marking resolved

lucassaldanha (approved):

PR #9438, opened 2025-05-19
Introduce `AttestationMatchingGroupV2`

tbenr:
The new matching group has:
- concurrency friendly by removing synchronized methods
- internal separation among aggregates and single attestations
- clear separation between flows: 
  - `streamForBlockProduction` - consider only aggregates
  - `streamForAggregationProduction` - consider only single attestations (for electra)
  - `streamForApiRequest` - consider all attestations (aggregates and single)
- an additional `fillUp` step which will be used by the pool v2 to improve aggregates produced by the streamForBlockProduction flow.
- timelimit on aggregation, which is used `streamForAggregationProduction` and `fillUp` steps. This allows to timebox the work and have a good guarantee we complete aggregation in a reasonable time no matter what is in the pool
- a configuration param that enables "early single attestations drop" to discard SA as aggregates come in the pool

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/utils/AttestationBitsElectra.java
```diff
@@ -329,6 +330,21 @@ public boolean isExclusivelyFromCommittee(final int committeeIndex) {
         && committeeAggregationBitsMap.containsKey(committeeIndex);
   }
 
+  @Override
+  public boolean isFromCommittee(final int committeeIndex) {
```
tbenr, line 334: I forgot to add tests for these new methods

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
```diff
@@ -0,0 +1,520 @@
...
+   *
+   * @return an iterator including attestations for every validator included in this group.
+   */
+  private Iterator<PooledAttestation> createAggregatingIterator(
+      final long timeLimitNanos,
+      final Supplier<Stream<PooledAttestation>> candidatesStreamSupplier) {
+    readLock.lock();
+    try {
+      // Capture a copy of includedValidators under lock for the iterator's isolated use
+      AttestationBits includedValidatorsCopy = this.includedValidators.copy();
```
rolfyone: final

rolfyone: also, we only need the readLock for this copy, dont need for the reduce that unlock right down.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
```diff
@@ -0,0 +1,520 @@
...
+                timeLimitNanos, aggregationProductionCandidatesStreamSupplier(committeeIndex)),
+            false)
+        .map(
+            pooledAttestation -> new PooledAttestationWithData(attestationData, pooledAttestation));
+  }
+
+  public Stream<PooledAttestationWithData> streamForApiRequest(
+      final Optional<UInt64> committeeIndex, final boolean requiresCommitteeBits) {
+    boolean noMatch;
+    readLock.lock();
```
rolfyone: we should be keeping the lock to where we use the data - so in this case this lock should be inside `noMatchingAttestations`
this improves readabilty but also centralises the lock so we know any time we call that fn its correct, otherwise if we ever use it elsewhere we'd also need locks there too

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
```diff
@@ -0,0 +1,520 @@
...
+      // Check if already seen before modifying the main includedValidators
+      if (includedValidators.isSuperSetOf(attestation)) {
+        // We've already seen and filtered out all of these bits, nothing to do
+        return 0;
+      }
+      // Mutate main includedValidators under write lock
+      includedValidators.or(attestation);
+
+      // Calculate size *before* removal for accurate delta.
+      int sizeBefore = size();
```
rolfyone: final

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
```diff
@@ -0,0 +1,520 @@
...
+      singleAttestationsByCommitteeIndex.forEach(
+          (committeeIndex, singleAttestations) -> {
+            singleAttestations.removeIf(
+                singleAttestation -> includedValidators.isSuperSetOf(singleAttestation.bits()));
+            if (singleAttestations.isEmpty()) {
+              singleAttestationsByCommitteeIndex.remove(committeeIndex);
+            }
+          });
+
+      int sizeAfter = size();
```
rolfyone: could either call final here or just use size in computation of `numRemoved`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
```diff
@@ -0,0 +1,520 @@
...
+      // Remove fully covered attestations using removeIf on the concurrent sets
+      attestationsByValidatorCount
+          .values()
+          .forEach(
+              candidates ->
+                  candidates.removeIf(
+                      candidate -> includedValidators.isSuperSetOf(candidate.bits())));
+
+      // Also remove empty sets from the outer map for cleanup
+      attestationsByValidatorCount.entrySet().removeIf(entry -> entry.getValue().isEmpty());
```
rolfyone: potentially we could do this after the cleanup from `singleAttestationsByCommitteeIndex`

This may be helpful because we'd be able to finish with included validators and release lock earlier potentially (before we do the math or this entryset we could release writelock)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
tbenr: ahh yes good catch

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
tbenr: yes, can be simplified!

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
tbenr: looking at that code section I think I can make `attestationsByValidatorCount` handling more like `singleAttestationsByCommitteeIndex` where the removal is all done in one pass

tbenr: ah no, that's not a good idea

tbenr: yes i found a good solution

rolfyone (approved):
LGTM

PR #9439, opened 2025-05-20
Added config defaulting to the Config loader

rolfyone:
Now when a configuration file is passed in, if it specifies the config_name field, that can be loaded over the top to ensure that we aren't missing parameters.

 This will make teku configuration loading more tolerant from external sources.

## Documentation

## Changelog


CHANGELOG.md
```diff
@@ -16,4 +16,6 @@
 - Added an info message on startup for the highest supported milestone and associated epoch.
 - Added jdk 24 docker image build.
 - Improved performance when scheduling attestations in the beginning of the epoch for a large number of validators.
+- Improved config loading to use builtin configurations to default any fields we need that were missing from a passed in configuration.
```
gfukushima:
```suggestion
- Improved configuration loading to use builtin configurations to default any fields we need that were missing from a passed in configuration.
```

gfukushima (approved):
Technically speaking PR lgtm. I understand that this is an improvement from the development point of view, but would this be a breaking change from the user POV? Supposing we use to error out for lack of required configs and now we default to values that users might not be aware?

rolfyone:
we'll print what we're overriding so i think its ok. mostly it insulates ppl like pandas from us needing config etc and them not providing it when we do big fork changes etc...

PR #9445, opened 2025-05-20
Introduce `AggregatingAttestationPoolV2`

tbenr:
Introduces V2 which has:
- better concurrency, removes synchronized methods
- ELECTRA oriented: when producing aggregates for block production, it sorts them by rewards
- a new aggregation for block production flow:
  1. it tries to produce all possible aggregates (timeboxed with `maxBlockAggregationTimeMillis`)
  2. it will sort them by rewards and select the top (8 for ELECTRA)
  3. it will try to improve the aggregations using single attestations that has not being part of the aggregates (so called `fillUp`, timeboxed with `maxTotalBlockAggregationTimeMillis`) 
- support `parallel` aggregation and fillUp stages

`maxBlockAggregationTimeMillis` might be in the order of `200ms`
`maxTotalBlockAggregationTimeMillis` might be in the order of `500ms`
leaving up to ~300ms to sorting and fillup.

This PR also introduces a `AggregatingAttestationPoolProfiler` interface. Implementation will come in later PRs.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
+++ b/ethereum/spec/src/testFixtures/java/tech/pegasys/teku/spec/util/DataStructureUtil.java
@@ -811,6 +811,11 @@ public final class DataStructureUtil {
         slot, randomUInt64(), randomBytes32(), randomCheckpoint(), randomCheckpoint());
   }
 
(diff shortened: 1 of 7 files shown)


rolfyone (approved):
LGTM

PR #9455, opened 2025-05-21
Early request getPayload, so that we can build the block body in parallel

tbenr:
since attestation aggregation from the aggregating pool is an heavy step, we can early initiate the payload retrieval and keep building the body in parallel.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

tbenr:
a block production made with this change, on holesky, with pool-v2 enabled
image

PR #9482, opened 2025-05-23
Added error for genesis state not matching latest block header

rolfyone:
This will give earlier warning when an invalid genesis state is in use.

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #9487, opened 2025-05-26
Include SingleAtt in block flow aggregation, remove parallel flow

tbenr:
Due to not being a game-changer (and complex to implement) I decided to remove the parallel flow.

Added the single attestation only aggregation.
On healthy networks (mainnet), only considering aggregates we do not reach 8 attestations https://light-mainnet.beaconcha.in/slots/filtered?f&f.missing=1&f.orphaned=1&f.graffiti=TK
So I added a second pass of single attestation only aggregation, still limited by the aggregation time limit.

This also removes the `TimeLimitingIteretor` wrapper because introduces a weird edge case I fund during unit tests adjustments

## Documentation

## Changelog


mehdi-aouadi (approved):
LGTM

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/attestation/MatchingDataAttestationGroupV2.java
```diff
@@ -414,6 +414,12 @@ private boolean noMatchingPreElectraAttestations(final Optional<UInt64> committe
   }
 
   private Supplier<Stream<PooledAttestation>> blockProductionCandidatesStreamSupplier() {
+    if (attestationsByValidatorCount.isEmpty()) {
+      // There are no aggregates left, which means they have all been included on-chain,
+      // so we can consider the long tail of single attestations that have not reached an aggregator
+      // in time
+      return () -> singleAttestationsByCommitteeIndex.values().stream().flatMap(Set::stream);
```
rolfyone, line 434: these would be value based surely... do we sort them somewhere?

tbenr: we do.
but i have another version locally, which is better than this simple one. I tested this and has the problem that we could start looking into SAs before completing the round on the aggregates.

PR #9491, opened 2025-05-28
Align config with specs

### LINKED ISSUE #9483: investigate differences in the minimal preset
comparing against latest consensus-spec,

in [phase0](https://github.com/ethereum/consensus-specs/blob/dev/presets/minimal/phase0.yaml), we have `SAFE_SLOTS_TO_UPDATE_JUSTIFIED` and spec does not

in [deneb](https://github.com/ethereum/consensus-specs/blob/dev/presets/minimal/deneb.yaml), we dont have `FIELD_ELEMENTS_PER_BLOB` but spec does

in [altair](https://github.com/ethereum/consensus-specs/blob/dev/presets/minimal/altair.yaml), we don't have `SYNC_COMMITTEE_BRANCH_LENGTH` but spec does

We should probably update our preset and reconcile the differences.

StefanBratanov:
## PR Description
Some cleanup of configs/presets

## Fixed Issue(s)
fixes #9483 

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9501, opened 2025-05-30
[EIP-7917] Add deterministic proposer lookahead changes to Fulu

### LINKED ISSUE #9378: [SPIKE][EIP-7917] deterministic proposer lookahead
https://github.com/ethereum/consensus-specs/pull/4190

https://eips.ethereum.org/EIPS/eip-7917

we should spike this and evaluate complexity and epoch-transition overhead

gfukushima:
## PR Description
Implementation of https://github.com/ethereum/consensus-specs/pull/4190

## Fixed Issue(s)
fixes #9378 

## Documentation

## Changelog


tbenr:
worth noting that on FULU, priming the proposerIndex doesn't make sense anymore (we don't use that transition cache anymore)

https://github.com/Consensys/teku/blob/839a13b1a9bb98b3ed1928d1b4f4a36c2d6eeee3/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/EpochCachePrimer.java#L84

StefanBratanov:
> worth noting that on FULU, priming the proposerIndex doesn't make sense anymore (we don't use that transition cache anymore)
> 
> https://github.com/Consensys/teku/blob/839a13b1a9bb98b3ed1928d1b4f4a36c2d6eeee3/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/EpochCachePrimer.java#L84

good point, to be fair, but just calling that method is super quick now anyways that probably not even worth to put an if statement in my opinion

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/BeaconStateAccessorsFulu.java
```diff
@@ -0,0 +1,66 @@
...
+        .get(lookaheadIndex)
+        .get()
+        .intValue();
+  }
+
+  public List<Integer> getBeaconProposerIndices(final BeaconState state, final UInt64 epoch) {
+    final IntList indices = getActiveValidatorIndices(state, epoch);
+    final Bytes32 seed = getSeed(state, epoch, Domain.BEACON_PROPOSER);
+    return miscHelpers.computeProposerIndices(state, epoch, seed, indices);
+  }
```
tbenr, line 56: this is maybe a thing for spec, but I don't like the method names here. It should be clearer which method lookups the state and which computes the indices.
So i guess `getBeaconProposerIndices` should be `computeBeaconProposerIndices` too

tbenr (approved):
I think we can YOLO this.
Just a comment

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/BeaconStateAccessorsFulu.java
tbenr: this has been automerged, lol
just want to make sure @StefanBratanov and @Gabriel-Trintinalia  see this comment

linoscope: Hi @tbenr ! EIP and spec coauthor here.

> but I don't like the method names here. It should be clearer which method lookups the state and which computes the indices.
> So i guess getBeaconProposerIndices should be computeBeaconProposerIndices too

Actually if you look at current specs some of these `getX` functions are already just wrappers around the corresponding `computeX` functions. For example, `get_beacon_committee` calls `compute_committee` ([code](https://github.com/ethereum/consensus-specs/blob/46306c4f8addef64defc2a52d85f46992ba2421b/specs/phase0/beacon-chain.md#L1078-L1079)) and pre-Fulu `get_beacon_proposer_index` ([code](https://github.com/ethereum/consensus-specs/blob/46306c4f8addef64defc2a52d85f46992ba2421b/specs/phase0/beacon-chain.md#L1097-L1098)) calls `compute_proposer_index`. So while I do understand your point that "getters" doing computation can be misleading, I would say it is already established pattern that exists in the specs.

tbenr: @linoscope I see what you mean. I was focused on specific changes. Widening the look that's consistent with the existing patterns. all good.

Issue #9446, opened 2025-05-20
`Single attestation is not supported` warnings may appear


tbenr:
Warning like `WARN  - Single attestation is not supported` are sometime logged from the `DefaultPerformanceTracker`. When this happens there are no other particular side effects.
It may be related to node hosting a lot of validators and SingleAttestations may reach the Tracker before being converted.
Another possibility is that we may have some codepaths that bring SA not being fully processed (early dropped for some reasons) but still reach the Tracker unconverted.
Could it also be related to validators publishing attestations to multiple BNs?

tbenr:
related to #9169

PR #9504, opened 2025-06-02
Convert the single attestations into aggregated ones when reporting the attestations performance

mehdi-aouadi:
## PR Description
The `WARN - Single attestation is not supported` log is caused by the `PerformanceTracker` when it gets a `SingleAttestation` which could happen when:
- An attestation is saved for the future: https://github.com/Consensys/teku/blob/839a13b1a9bb98b3ed1928d1b4f4a36c2d6eeee3/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/AttestationValidator.java#L123-L125
- An attestation is far from the future and hence ignored: https://github.com/Consensys/teku/blob/1a45cebda7a82a41bd8c3f40b0143acb2f8f1100/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/AttestationValidator.java#L115
- There's no state available to validate the `SingleAttestation` so it's ignored: https://github.com/Consensys/teku/blob/1a45cebda7a82a41bd8c3f40b0143acb2f8f1100/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/AttestationValidator.java#L151
- The state used to validate the `SingleAttestation` is too old so the attestation is ignored: https://github.com/Consensys/teku/blob/1a45cebda7a82a41bd8c3f40b0143acb2f8f1100/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/AttestationValidator.java#L189

This PR makes the performance tracker accept the single attestations and convert them into aggregated ones when calculating the attestations performance.
It also simplifies the single attestations conversions by merging the aggregation bits and the aggregated attestation creation.

## Fixed Issue(s)

#9446 

## Documentation

## Changelog


tbenr:
So I have a feeling that the most "healthy" condition we can hit this is also the `SAVE_FOR_FUTURE` case:

https://github.com/Consensys/teku/blob/839a13b1a9bb98b3ed1928d1b4f4a36c2d6eeee3/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/AttestationValidator.java#L123-L125

which can happen in a multi-BN setup:
BN1 (primary):
  - receives the block N
  - sends head event
  - VC react to head event and asks for attestationData and sends signed att it to BN1 and BN2
  
BN2 (secondary):
 - received the signed att but has not yet received\fully_imported block N
 - published attestation goes in SAVE_FOR_FUTURE (which is neither ignore nor reject)


@mehdi-aouadi  WDYT?

If you agree with this flow, maybe we could just "embrace" the unconverted SA condition in the default tracker and demote the log to debug

mehdi-aouadi:
As discussed I lowered the `PerformanceTracker` log level to debug in order to handle the `SAVE_FOR_THE_FUTURE` use case where we could still end up having a non converted `SingleAttestation`

tbenr (approved):
LGTM

PR #9505, opened 2025-06-02
remove unused --validators-proposer-blinded-blocks-enabled option

mehdi-aouadi:
## PR Description
The `--validators-proposer-blinded-blocks-enabled` option is not used anymore since the block v3 end point has been introduced.

## Fixed Issue(s)

#9513 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,13 @@
 # Changelog
 
 ## Upcoming Breaking Changes
-- The `--validators-proposer-blinded-blocks-enabled` is deprecated and will be removed. It's not used anymore and should be removed from any configs.
 
 ## Current Releases
 
 ## Unreleased Changes
 
 ### Breaking Changes
+- Removed `--validators-proposer-blinded-blocks-enabled` unused option. This option is not used anymore and should be removed from any config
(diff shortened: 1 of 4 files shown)


StefanBratanov (approved):
LGTM, have to remove it from the docs as well.

tbenr:
mm since this is a breaking change and we print an `info` when not enabled (assuming majority is using mev-boost) so we will break a significant number of user startups (I guess)
Wander if it would be better to do thin in two steps (deprecation warning first, then actual removal). Juts an idea.

mehdi-aouadi:
> mm since this is a breaking change and we print an `info` when not enabled (assuming majority is using mev-boost) so we will break a significant number of user startups (I guess) Wander if it would be better to do thin in two steps (deprecation warning first, then actual removal). Juts an idea.

That's right, MEV users might be still using it and that could be a problem. I raised #9512 to deprecate it first as suggested

lucassaldanha:
Can we close this PR and have a ticket for the next piece?

mehdi-aouadi:
This PR could be reopened and merged once #9512 is released

PR #9506, opened 2025-06-02
Fix block contents kzg proofs max length for fulu

jtraglia:
## PR Description

This should use the "extended" field elements per blob count.

https://github.com/ethereum/beacon-APIs/blob/7d3a0475759fdbd2ad950c2652a8dce83b5ad984/types/fulu/block_contents.yaml#L7

## Documentation

## Changelog


zilm13 (approved):
LGTM

zilm13:
Great catch.
Thank you, Justin!

PR #9512, opened 2025-06-03
deprecate unused validators-proposer-blinded-blocks-enabled option

mehdi-aouadi:
## PR Description
The `--validators-proposer-blinded-blocks-enabled` option is not used anymore since the block v3 end point has been introduced.
This is a first step deprecation. The final removal is done in #9505 

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #9516, opened 2025-06-05
Remove Deneb/Electra from blob schedule and default to Electra limit

StefanBratanov:
## PR Description
As per https://github.com/ethereum/consensus-specs/pull/4341 and https://github.com/ethereum/consensus-specs/pull/4342

## Fixed Issue(s)
N/A

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
```diff
@@ -122,27 +118,20 @@ public Optional<MiscHelpersFulu> toVersionFulu() {
...
+        .orElseGet(specConfigFulu::getMaxBlobsPerBlock);
   }
 
   public int getHighestMaxBlobsPerBlockFromSchedule() {
-    checkArgument(!blobSchedule.isEmpty(), "Blob schedule not correctly defined.");
     return blobSchedule.stream()
         .max(Comparator.comparing(BlobScheduleEntry::maxBlobsPerBlock))
         .map(BlobScheduleEntry::maxBlobsPerBlock)
-        .orElseGet(() -> blobSchedule.getLast().maxBlobsPerBlock());
+        .orElseGet(specConfigFulu::getMaxBlobsPerBlock);
```
rolfyone, line 134: not sure this is quite right, but then the previous isnt really either given they can go down... we can come back to this if its not causing issues, but i think we'd need to get the list and sort and find max blobs per block from schedule if we do need that value...

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
```diff
@@ -134,22 +150,36 @@ public void equals_electraConfigDiffer() {
...
+    assertThat(miscHelpersFulu.getMaxBlobsPerBlock(UInt64.valueOf(269569)))
+        .isEqualTo(maxBlobsPerBlockElectra);
     // last epoch of deneb
-    assertThat(miscHelpersFulu.getMaxBlobsPerBlock(UInt64.valueOf(364031))).isEqualTo(6);
+    assertThat(miscHelpersFulu.getMaxBlobsPerBlock(UInt64.valueOf(364031)))
+        .isEqualTo(maxBlobsPerBlockElectra);
     // electra boundary
-    assertThat(miscHelpersFulu.getMaxBlobsPerBlock(UInt64.valueOf(364032))).isEqualTo(9);
+    assertThat(miscHelpersFulu.getMaxBlobsPerBlock(UInt64.valueOf(364032)))
+        .isEqualTo(maxBlobsPerBlockElectra);
```
rolfyone: most of this test is meaningless now, we can just simplify

rolfyone (approved):
some minor comments...

due to the old release having the list not being empty as a requirement, this does need to be listed as a breaking change again for changelog.

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
StefanBratanov: just removed it, we can add it again when we have an actual bpo schedule

StefanBratanov:
> some minor comments...
> 
> due to the old release having the list not being empty as a requirement, this does need to be listed as a breaking change again for changelog.

Added to changelog

PR #9518, opened 2025-06-05
Use byRange RPC for sync DataColumnSidecars in FULU

zilm13:
## PR Description
Why?
byRoot RPC should return only non-finalized data, see spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/p2p-interface.md#datacolumnsidecarsbyroot-v1
`Clients MUST support requesting sidecars since minimum_request_epoch, where minimum_request_epoch = max(finalized_epoch, current_epoch - MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS, FULU_FORK_EPOCH). If any root in the request content references a block earlier than minimum_request_epoch, peers MAY respond with error code 3: ResourceUnavailable or not include the data column sidecar in the response.
`

I did it in a bit hacky way.
The right way requires removing of DataColumnReqResp, DataColumnReqRespBatchingImpl, SimpleSidecarRetriever, DataColumnSidecarRetriever, RecoveringSidecarRetriever, SyncPreImportBlockChannel,  DasPreSampler and building all new logic starting from BatchSync.

Other changes include:
- Noisy logging reduced, important logging upgraded from trace to debug
- Fixes in `byRange` handler, `earliestDataColumnSidecarSlot` and `firstIncompleteSlot` are written on finalization and we cannot rely on them

## Documentation

## Changelog


gfukushima (approved):
lgtm

PR #9525, opened 2025-06-09
Update cgc in ENR using the data from CustodyGroupCountManager

zilm13:
## PR Description
It will be actually set very early just after node updated and first finalized (if FULU is supported == not FAR_FUTURE_EPOCH scheduled). But should we care? All not known fields in enr are ignored. The size is a bit bigger, it's the only thing we may care.
Also I'm not sure, what tests could be added.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #9557, opened 2025-06-13
ConcurrentModificationException in AsyncStream

tbenr:
## Documentation

## Changelog


gfukushima (approved):

zilm13:
Added following changes:
- Test clarifying that concurrent modification is caught and logged in a right way
- Remove old useless logging as this case showed that it's not working 
- Defaults AsyncStream to a safe unmodifiable collection
- Revert impl fix as it's now fixed on infrastructure level

tbenr:
I think the failing test shows that this isn't solving the issue, you still require to have a concurrent-friendly collection to create a safe copy of the iterator

zilm13:
@tbenr 
Ahhaha sorry
I needed to change test after infrastructure method changes. Updating
Test is failing because there were no exceptions during tests and we require to have exception with the text you've added.

But I added a correct test and it's now failing on `create()`. You are right. Looks like the only thing we could do is adding a javadoc + we need concurrent safe collection. Because non-concurrent safe could fail on collection. We cannot have safe creator after all

PR #9562, opened 2025-06-18
cleanup some of the slot time methods

rolfyone:
Ultimately we're better off with these functions in mischelpers, but also, if we can avoid recomputing in many places then we're in a better position when slot times change.

This was a first pass of some obvious issues.

## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #9571, opened 2025-06-18
Disable AsyncStream concurrent windows test

tbenr:
The disabled test isn't testing production logic, rather it just checks that we do proper logging hitting the edge-case.
So not really needed to run on all platforms.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9572, opened 2025-06-19
more slot timing cleanup

rolfyone:
a couple more timing issues to reduce tech debt

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
@@ -196,7 +196,7 @@ public abstract class RecentChainData implements StoreUpdateHandler {
   }
 
   public UInt64 computeTimeAtSlot(final UInt64 slot) {
-    return spec.atSlot(slot).miscHelpers().computeTimeAtSlot(genesisTime, slot);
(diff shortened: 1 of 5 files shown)


storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
```diff
@@ -196,7 +196,7 @@ public UInt64 getGenesisTimeMillis() {
   }
 
   public UInt64 computeTimeAtSlot(final UInt64 slot) {
-    return genesisTime.plus(slot.times(spec.getSecondsPerSlot(slot)));
+    return spec.atSlot(slot).miscHelpers().computeTimeAtSlot(genesisTime, slot);
```
gfukushima: can't we just 
```suggestion
    return spec.computeTimeAtSlot(genesisTime, slot);
```

gfukushima (approved):
lgtm, i think there might be room a tiny simplification one one of the adjustments made.

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
rolfyone: yeh i guess we can now, mainly i dont want the math, so either works

PR #9580, opened 2025-06-20
fix possible recursive update

pinges:
## PR Description
This PR removes a possible recursive update. This can happen if the map entry in `pendingConnections` is removed while calling the `doConnect()` method.


rolfyone (approved):
LGTM

PR #9592, opened 2025-06-24
created local docker targets

rolfyone:
can now do
`localDocker` to build local/teku:develop
`localDockerClean` to clean the tags built from localDocker.

I basically copied the consensys/teku builds but without the extra jdk. It could probably be cleaned up a little at some point.

## Documentation

## Changelog


lucassaldanha (approved):
LGTM

Issue #9539, opened 2025-06-11
Add rpc Status v2 with earliest_available_slot


lucassaldanha:
https://github.com/ethereum/consensus-specs/pull/4374

Tasks:
- [x] Update RpcHandler to support choosing request body based on the version negotiated by multistream-select. (PR: https://github.com/Consensys/teku/pull/9594)
- [x] Update the logic sending status message to correctly use V1 or V2 based on the version negotiated (and adding `earliest_available_slot`) when using V2.
- [x] ~~Update our logic to consume the information from StatusV2 - earliest_available_slot in our syncing logic.~~ (separate ticket https://github.com/Consensys/teku/issues/9556)

lucassaldanha:
Moving this out of devnet-2 epic. We have a temporary version implemented for devnet-2 but now I am working on the final "production" version.

PR #9594, opened 2025-06-24
Updated RpcHandler with sendRequestWithBodySelector(..) method

lucassaldanha:
## PR Description

This PR is a preparation for a follow-up change where we choose the correct Status message (V1 or V2) depending on the version negotiated by libp2p's multistream-select.

The tricky part is that when we are defining the request or sending the body, we don't know what version will be used. The way to handle this is to define a function that takes a string representing the `protocolId` (version negotiated) and chooses the correct body depending on the version chosen.

Summary of changes:
- Created an interface `RpcRequestBodySelector` and two implementations, one to represent the current behavior where we always use the same body regardless of the version, and a new implementation that uses a map to choose the correct body based on the `protocolId` negotiated by multistream-select.
- Updated `RpcHandler` class with a new method `sendRequestWithBodySelector` that supports this body selector.
- Updated `RpcHandler` existing method `sendRequest` to call the new method `sendRequestWithBodySelector` using a `SingleRpcRequestBodySelector`, basically maintaining the existing behavior.

Next steps: update the logic of sending Status message to choose the correct V1 or V2

## Fixed Issue(s)
partially addresses #9539 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/bodyselector/RpcRequestBodySelector.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/networking/libp2p/rpc/bodyselector/RpcRequestBodySelector.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright Consensys Software Inc., 2025
+ *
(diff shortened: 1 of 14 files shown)


rolfyone (approved):
LGTM

Issue #9604, opened 2025-06-25
Rewrite RemoteValidatorHandler\typeDefClient using a async API calls


tbenr:
The current `OkHttpValidatorTypeDefClient` implementation is synchronous. We became async here:
https://github.com/Consensys/teku/blob/839a13b1a9bb98b3ed1928d1b4f4a36c2d6eeee3/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/RemoteValidatorApiHandler.java#L328-L340

essentially at executor level.

This means we use an executor (thread) for each inflight calls, and we are capped by the max threads (default 5, max 12).

In a multiBN scenario, where we run readiness check every slot to all configured BNs (each readiness check consist in 2 different calls) we can easily consume all available threads, and calls can be easily queued up.

I think we should move the "async" down to the actual API call, as we do in the builder rest client:

https://github.com/Consensys/teku/blob/e3259c50527694635c75b9d1920968b810a815ff/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/rest/OkHttpRestClient.java#L171-L178

EDIT:

OkHttp to handle its internal thread pool anyway (default dispatcher allows max 64 concurrent calls (threads) and max 5 calls per host, but we can have our own custom config), and it isn't using a single-threaded event base non blocking IO.
So migrating to that model is more for simplicity and not have to handle directly the asyncRunner\threadPool.

Another option would be to have a dedicated asyncRunners.
some ideas:
- one for each configured beacon node?
- one for critical apis, another for readiness checks?

anyway we should improve the client to allow having a timeout option per method, and not a common one applied to all of them

tbenr:
Some info around the actual issue:

this setup was running just 2 beacon nodes (primary and a secondarty)

`validator` executor has 5 threads (the default) and they almost always had all threads active.
<img width="1828" alt="Image" src="https://github.com/user-attachments/assets/eef0bf90-8dac-4758-908e-1dd0d31de992" />

we can see executor tasks queueing up:
<img width="1799" alt="Image" src="https://github.com/user-attachments/assets/9e595486-a7dc-4508-b917-70ee49c9ab22" />

PR #9605, opened 2025-06-25
Dedicated api executors for RemoteBeaconNodeApi

tbenr:
quick win related to #9604

Uses a dedicated asyncRunner when VC is remote and has to call a remote BN.
So we leave our main asyncRunner for internal core tasks (running scheduled tasks and even the local signer!)

In case of multiBN configuration, I introduced a dedicated asyncRunner for the readiness-related apis (sync and peer count)

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Several improvements on how validator client handles multiple beacon nodes:
   - reduced timeout for beacon node API calls down to 10 seconds
   - improved handling of unresponsive\unreachable beacon nodes.
+  - dedicated thread pools for beacon node API calls
 
 ### Bug Fixes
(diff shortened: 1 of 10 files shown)


tbenr:
before and after a VC running with 1 primary and 2 secondary (worse case, quantile=1.0):

image


comparison with a VC running old code and only with a primary node (worse case, quantile=1.0):

image

executors queue do not go up like before
image

mehdi-aouadi (commented):
Would be nice to add some test for the max threads threshold.
And wondering if it's worth it to make some parameters configurable (API max threads, number of threads for the readiness methods...)

rolfyone (approved):
LGTM we can do separate PRs for config stuff

PR #9607, opened 2025-06-25
Refactor passing of fork digest in gossip

StefanBratanov:
## PR Description

- Remove `getForkDigest(spec)` from `ForkInfo`
- Pass `forkDigest` directly to all gossip managers
- Also this avoids recomputing fork digest on every gossiped message, which is the current behaviour. Look at `AttestationSubnetSubscriptions` to see current behaviour.
- A lot of the changed classes is just refactor.
- This will greatly simplify BPO.
- Rename `BlobSchedule` to `BpoForkSchedule`

## Fixed Issue(s)
related to #9365 

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/SpecLogicFulu.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/SpecLogicFulu.java
@@ -47,7 +47,6 @@ import tech.pegasys.teku.spec.logic.versions.electra.util.AttestationUtilElectra
 import tech.pegasys.teku.spec.logic.versions.fulu.block.BlockProcessorFulu;
 import tech.pegasys.teku.spec.logic.versions.fulu.forktransition.FuluStateUpgrade;
 import tech.pegasys.teku.spec.logic.versions.fulu.helpers.BeaconStateAccessorsFulu;
-import tech.pegasys.teku.spec.logic.versions.fulu.helpers.BpoForkSchedule;
 import tech.pegasys.teku.spec.logic.versions.fulu.helpers.MiscHelpersFulu;
 import tech.pegasys.teku.spec.logic.versions.fulu.statetransition.epoch.EpochProcessorFulu;
 import tech.pegasys.teku.spec.schemas.SchemaDefinitionsFulu;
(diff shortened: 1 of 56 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/SpecLogicFulu.java
```diff
@@ -102,9 +102,9 @@ public static SpecLogicFulu create(
       final TimeProvider timeProvider) {
     // Helpers
     final PredicatesElectra predicates = new PredicatesElectra(config);
-    final BlobSchedule blobSchedule = new BlobSchedule(config);
+    final BpoForkSchedule bpoForkSchedule = new BpoForkSchedule(config);
```
rolfyone: we should just create this in MiscHelpersFulu, or at the very least not store it in here, we never use it.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
```diff
@@ -92,21 +92,21 @@ public static MiscHelpersFulu required(final MiscHelpers miscHelpers) {
...
   private final SchemaDefinitionsFulu schemaDefinitionsFulu;
-  private final BlobSchedule blobSchedule;
+  private final BpoForkSchedule bpoForkSchedule;
 
   public MiscHelpersFulu(
       final SpecConfigFulu specConfig,
       final PredicatesElectra predicates,
       final SchemaDefinitionsFulu schemaDefinitions,
-      final BlobSchedule blobSchedule) {
+      final BpoForkSchedule bpoForkSchedule) {
```
rolfyone: as said elsewhere, we can just create this in here.

rolfyone: Potentially theres 1 test case where we want to send different BPO, so maybe this becomes a package private and by default the one we use elsewhere doesnt have this 4th param...

rolfyone (approved):
overall lgtm but we probably want ot make sure we're working on devnet2 before going too far on this...

i did try to use this on devnet2 spec, but it didnt seem to behave the same, so we probably need to do some tests etc too to make sure things are working.

PR #9620, opened 2025-06-30
call create attestationData once for all committees

### LINKED ISSUE #9618: PostElectra - call attestationData once
https://ethereum.github.io/beacon-APIs/#/Validator/produceAttestationData

there is no point in calling the api for each validators in the same committee anymore.

We can call it once for all validators now.

this only affects VC running A LOT of validators (multiple validators attesting on multiple committees)

https://github.com/Consensys/teku/blob/99df3fd34b0bee7dc86d1825c90c6b57452acdff/validator/client/src/main/java/tech/pegasys/teku/validator/client/duties/attestations/AttestationProductionDuty.java#L119-L140

currently on VC running 20k validators we have bursts of 50-60 `createAttestationData` calls at the same moment.

we should also rise a beacon-API pr to clarify that post-Electra the API can be used this way (calling it once with committe_index = 0) so BNs must not expect the api to be called with the actual committee index.

tbenr:
fixes #9618

number of API calls goes down significantly, as expected:

image

but, surprisingly, the worse case in attestation duty total time goes up:

image

## Documentation

## Changelog


rolfyone:
i do really like this idea...

tbenr:
@rolfyone I updated the tests and raised a beacon api clarification PR

rolfyone (approved):
LGTM

PR #9626, opened 2025-07-02
Fix readiness for sentry nodes

tbenr:
When we configure readiness manager when sentry nodes are configured, we could end up not passing all possible nodes roles to the readiness manager. A recent PR considering non-primary nodes as non-ready. So we ended up with stuck NOT_READY nodes.

This is a quick fix but i think we should revisit how we use the readiness manager in this context.

Added an acceptance test replicating the issue

## Documentation

## Changelog

### DIFF
--- a/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
+++ b/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
@@ -94,7 +94,7 @@ public class BeaconNodeReadinessManager extends Service implements ValidatorTimi
   }
 
   public SafeFuture<Void> performPrimaryReadinessCheck() {
-    return performReadinessCheckWithInflightCheck(primaryBeaconNodeApi, true);
+    return performReadinessCheckWithInflightCheck(primaryBeaconNodeApi);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
```diff
@@ -204,12 +206,16 @@ private boolean isTimeToPerformReadinessCheck(
         lastErrorTimestamp.get().plus(ERRORED_SECONDARY_READINESS_CHECK_INTERVAL_DELAY_MS));
   }
 
-  private SafeFuture<Void> performReadinessCheck(
-      final RemoteValidatorApiChannel beaconNodeApi, final boolean isPrimaryNode) {
+  @SuppressWarnings("ReferenceComparison")
+  private boolean isPrimary(final RemoteValidatorApiChannel beaconNodeApi) {
+    return beaconNodeApi == primaryBeaconNodeApi;
```
mehdi-aouadi, line 211: Wouldn't it be better and sufficient to compare the endpoints instead of the reference?

tbenr: well, at the moment there is nothing to prevent adding different RemoteValidatorApiChannel referring to the same endpoint.
I think we should try to revisit stuff around this to enforce endpoint uniqueness OR move the internals to be more `endpoint` string oriented (allowing multiple channels against the same endpoint, if there is an actual need for it)

mehdi-aouadi (approved):
LGTM

validator/remote/src/main/java/tech/pegasys/teku/validator/remote/BeaconNodeReadinessManager.java
tbenr: I'll open an issue to track these things

PR #9643, opened 2025-07-09
fix return code for optimistic aggregates

rolfyone:
I noticed that we're returning an aggregate in the case where we're in optimistic mode, and we shouldn't be returning an aggregate in this case according to the api.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,4 +13,5 @@
 - Increase the default gas limit (`--validators-builder-registration-default-gas-limit`) to 45 million
 
 ### Bug Fixes
-- fix a regression introduced in the previous release causing a validator client configured with sentry nodes to not work properly
+- Fixed a regression introduced in the previous release causing a validator client configured with sentry nodes to not work properly
+- Fixed an issue where we would return an aggregation that was optimistic.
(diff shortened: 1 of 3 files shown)


tbenr (commented):
since we are checking the syncingState at `ValidatorApiHandler` level, I wander if we should do the check in the same place.

I guess this applies to `getAttestationData` too

rolfyone:
i think theres a race condition regardless unless we check the output and return 503 on the return once we know if the result as optimistic... theres already a syncing check at the start of the function (`validatorApiHandler.createAggregate`) but it doesnt guarantee that once we query metadata we haven't gone into syncing mode...
so all we can really do is once its all together, check if its optimistic and adjust the output, its the only point in time we can ensure our output is correct imo.
Ideally we could guarantee we're not optimistic during creation, but we're putting this data together in stages, so we need to reconcile that.
we reconcile it in `ValidatorDataProvider.createAggregateAndMetaData`, so until that, we haven't got all the data to make an accurate decision. i could potentially do trickery in the provider to get the correct data than the http handler...

tbenr:
I see what you mean but digging into it more, seems to me we hardcode the values

https://github.com/Consensys/teku/blob/fe4a9394507bd8f6f7d9835bdfdd50c3519fab98/data/provider/src/main/java/tech/pegasys/teku/api/ValidatorDataProvider.java#L263-L270

I think we use Metadata there just to get the milestone.


And finally i think that we already return an error if we are in optimistic sync cose the validatorApiHandler checks

```
  boolean isSyncActive() {
    return !syncStateProvider.getCurrentSyncState().isInSync();
  }
 ```
 
 which in the end returns `true` if we are optimistic.

rolfyone:
hrm ok, then theres no point for this PR... closing

PR #9647, opened 2025-07-09
refactor blob sidecars by range

mehdi-aouadi:
## PR Description
Refactoring the blob sidecars by range RPC method

## Documentation

## Changelog


tbenr (approved):
LGTM

tbenr:
oh good catch on single slot

Issue #9640, opened 2025-07-08
`ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0`


tbenr:
A supernode following the chain suddenly printed a big burst:
```
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.979 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.980 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.980 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
2025-07-08 13:22:04.980 ERROR - Unexpected rejected execution due to full task queue in operationPoolUpdater-async-scheduler-0
[...]

Issue #9465, opened 2025-05-22
Better heuristic to submit requests for recovery


lucassaldanha:
Check TODO-fulu `// TODO-fulu we probably need a better heuristics to submit requests for recovery` on `RecoveringSidecarRetriever`

zilm13:
- we could try just decreasing it from 5 mnutes, to, say, 30 seconds and check in Kurtosis how it will work
- the trigger could be something like we are not getting new sidecars for this slot in the last 15 seconds

zilm13:
we have reworked retriever which is currently in the work #10139

PR #9648, opened 2025-07-09
Refactor RecoveringSidecarRetriever to use less memory

### LINKED ISSUE #9638: Memory consumption while syncing
Seems like we are consuming a lot of memory while syncing a FULU network

Image

once sync is completed the mem usage goes down.

We should check where the consumption is happening and if we can (if we should) reduce it.

tbenr:
Avoids maintaining a reference to each retrieved `DataColumnSidecar` for 5 minutes.
The change now keeps only reference to `DataColumnSidecar`s that are actually pending and require a recover.

It also removes all these queued tasks on the asyncExecutor, thus it might affect #9640

fixes: #9638 

related to #9465

before:
image

after:
image

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetrieverTest.java
+++ b/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetrieverTest.java
@@ -228,7 +228,7 @@ public class RecoveringSidecarRetrieverTest {
     dbColumnIndexes.forEach(idx -> assertThat(db.addSidecar(sidecars.get(idx))).isDone());
 
     DataColumnSlotAndIdentifier id0 = createId(block, 0);
-    SafeFuture<DataColumnSidecar> res0 = delegateRetriever.retrieve(id0);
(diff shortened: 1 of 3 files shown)


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetrieverTest.java
```diff
@@ -98,81 +115,131 @@ void sanityTest() throws Exception {
...
-    assertThat(res1.get(1, TimeUnit.SECONDS)).isEqualTo(sidecars.get(1));
+    assertThat(res0).isCompletedWithValue(sidecars.get(0));
+    assertThat(res1).isCompletedWithValue(sidecars.get(1));
     assertThat(delegateRetriever.requests).allMatch(r -> r.promise().isDone());
   }
 
   @Test
-  void testMoreThanOneBlockWithBlobsOnSameSlot() throws Exception {
+  void succeededAndFailedRetrievalShouldBeImmediatelyRemovedFromPendingPromises() {
+    int blobCount = 3;
```
rolfyone, line 175: missing lots of finals

rolfyone (approved):

PR #9670, opened 2025-07-16
Find/replace indexes with indices

jtraglia:
## PR Description

I was reviewing data column sampling code and noticed several instances of "indexes" which is non-standard.

I did a find/replace to fix all instances of this.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9672, opened 2025-07-16
Make small improvements to blob reconstruction class

jtraglia:
## PR Description

This PR fixes little issues I noticed while review.

* `missingColumn` should be plural.
* I was confused about what `indices` was for (blobs or columns) so I renamed the variable.
* Improved some comments, because I think it's helpful.

## Documentation

## Changelog


rolfyone (approved):
LGTM

PR #9684, opened 2025-07-18
fix: genesis validator set cgc value

barnabasbusa:
## PR Description
This PR ensures that the cgc value will be updated based on the genesis validator set. 
You can assume that the genesis validator set is finalized, and there is no need to wait till epoch 0 gets finalized.

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -148,6 +156,42 @@ public void setCustodyGroupSyncedCount(final int custodyGroupSyncedCount) {
...
 
+  private void checkAndHandleGenesisTransition(final Set<UInt64> nodeValidatorIndices) {
+    combinedChainDataClient
+        .getLatestFinalized()
+        .ifPresent(
+            finalizedAnchor -> {
+              // Check if we've moved beyond genesis epoch (epoch 0)
+              if (finalizedAnchor.getEpoch().isGreaterThan(UInt64.ZERO)) {
+                // We've finalized beyond genesis, now get the genesis state to count validators
+                final UInt64 genesisSlot = spec.computeStartSlotAtEpoch(UInt64.ZERO);
```
gfukushima: This to me look a little redundant but I could be wrong, you can probably assume that genesis always start at slot 0 can't you? There might be a world where we have a network with genesis not starting at slot 0.
Also I need to check if it's safe to do this because in case of a restart you might be way ahead of genesis and still setting cgc based on genesis.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -148,6 +156,42 @@ public void setCustodyGroupSyncedCount(final int custodyGroupSyncedCount) {
     custodyGroupSyncedCountGauge.set(custodyGroupSyncedCount);
   }
 
+  private void checkAndHandleGenesisTransition(final Set<UInt64> nodeValidatorIndices) {
+    combinedChainDataClient
+        .getLatestFinalized()
+        .ifPresent(
+            finalizedAnchor -> {
+              // Check if we've moved beyond genesis epoch (epoch 0)
+              if (finalizedAnchor.getEpoch().isGreaterThan(UInt64.ZERO)) {
```
gfukushima: In the PR description you said you don't have to wait until epoch 0 is finalized but the whole update cgc count only runs once we have an epoch greater than 0 in this case which means we've already finalized genesis

gfukushima (commented):
I understand the intent of this PR but it probably need a bit of rework, I'll see if I can work with you on this on Monday.

tbenr:
decided to open an actual pr against master #9717

PR #9688, opened 2025-07-21
refactor: Replace magic number with named constant in EraFile

Bugarim:
## PR Description

Replaces the magic number 8 with a named constant HEADER_SIZE in EraFile.java. Using named constants instead of magic numbers improves code readability and maintainability.It makes the purpose of the value clear and allows for easier updates if the header size would ever change in the future


rolfyone:
While I agree with magic numbers idea, this is also non functional proof of concept code, will not be changing it until we actually support ERA files.

PR #9689, opened 2025-07-21
Limit the number of concurrent requests to historical data (RPC)

### LINKED ISSUE #9677: Check rate limit on serving RPC data columns
We can get dossed by remote peers quering data columns

Image

this cloggs our `storage_query` channel and makes the node unresponsive

note `irate(network_rpc_data_column_sidecars_by_root_requested_data_column_sidecars_total{instance=~"$system"}[1m])` give us requested columns per second

Image

apparently the sidecars were asked via 15 by root requests per second. So something like 80\90 columns per request

StefanBratanov:
## PR Description
Based on https://github.com/Consensys/teku/pull/6937
Introduces new hidden CLI option: `--Xp2p-historical-data-max-concurrent-queries` defaulting to 5.

## Fixed Issue(s)
fixes #9677 

## Documentation

## Changelog

### DIFF
--- a/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
+++ b/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
@@ -1581,7 +1581,7 @@ public class BeaconChainController extends Service implements BeaconChainControl
 
     // Using a throttled historical query retrieval when handling RPC requests to avoid
     // overwhelming the node in case of various DDOS attacks
-    final CombinedChainDataClient throttlingCombinedChainDataClient;
+    Optional<CombinedChainDataClient> throttlingCombinedChainDataClient = Optional.empty();
     if (beaconConfig.p2pConfig().getHistoricalDataMaxConcurrentQueries() > 0) {
       final ThrottlingStorageQueryChannel throttlingStorageQueryChannel =
           new ThrottlingStorageQueryChannel(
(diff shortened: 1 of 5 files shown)


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -1557,11 +1558,33 @@ protected void initP2PNetwork() {
...
+    } else {
+      throttlingCombinedChainDataClient = combinedChainDataClient;
+    }
+
     this.p2pNetwork =
         createEth2P2PNetworkBuilder()
             .config(beaconConfig.p2pConfig())
             .eventChannels(eventChannels)
-            .combinedChainDataClient(combinedChainDataClient)
+            .combinedChainDataClient(throttlingCombinedChainDataClient)
```
rolfyone: nit: i think if we wanted to be clearer here, we'd probably be making Throttling an optional, and then doing orElse here...
`.combinedChainDataClient(maybeThrottlingCombinedChainDataClient.orElse(combinedChainDataClient)`

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -1557,11 +1558,33 @@ protected void initP2PNetwork() {
...
+          new CombinedChainDataClient(
+              recentChainData,
+              throttlingStorageQueryChannel,
+              spec,
+              new EarliestAvailableBlockSlot(
+                  throttlingStorageQueryChannel,
+                  timeProvider,
+                  beaconConfig.storeConfig().getEarliestAvailableBlockSlotFrequency()));
+    } else {
+      throttlingCombinedChainDataClient = combinedChainDataClient;
```
rolfyone: if we did the optional bit this would disappear

rolfyone (approved):
LGTM just a nit that may make it a little clearer

rolfyone:
as an aside added #9694 to demonstrate the task queue working while i was investigating this.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
StefanBratanov: I like that, it's cleaner, done!

StefanBratanov:
I will wait for @tbenr to come back to review as well since he has done quite a bit of analysis

tbenr:
@StefanBratanov started testing it locally. Seeing some unexpected things in the surroundings. I'll give my green light tomorrow.

tbenr (approved):
LGTM.
@StefanBratanov.

I'm thinking if we need to actually drop requests or this is good enough.

so the upperbound of concurrent request is something like
peersCount * maxCouncurrentRPCcalls * maxObjectPerRPCCall

maybe we should doublecheck if that the final number is reasonable and we can actually avoid putting a maxQueue param

StefanBratanov:
> LGTM. @StefanBratanov.
> 
> I'm thinking if we need to actually drop requests or this is good enough.
> 
> so the upperbound of concurrent request is something like peersCount_maxCouncurrentRPCcalls_maxObjectPerRPCCall
> 
> maybe we should doublecheck if that the final number is reasonable and we can actually avoid putting a maxQueue param

@tbenr I am not sure about dropping requests, it can lead to peers disconnecting from us.

tbenr:
> > LGTM. @StefanBratanov.
> > I'm thinking if we need to actually drop requests or this is good enough.
> > so the upperbound of concurrent request is something like peersCount_maxCouncurrentRPCcalls_maxObjectPerRPCCall
> > maybe we should doublecheck if that the final number is reasonable and we can actually avoid putting a maxQueue param
> 
> @tbenr I am not sure about dropping requests, it can lead to peers disconnecting from us.

@StefanBratanov well, it could be a feature, not a bug. if we are connected to 100 peers coordinating to DOS us, it's not wrong to limit the damages. If they are not malicious but by chance they behave like a DOS attack, let them disconnect a bit might not be wrong in principle. The problem is finding the right limits

PR #9691, opened 2025-07-21
Default to a higher kzg-precompute value for supernodes

jtraglia:
## PR Description

I think it would be beneficial for supernodes to default to a higher KZG precompute value which makes recovery significantly faster at the cost of extra memory usage. See the following link for more information:

* https://github.com/ethereum/c-kzg-4844#precompute

## Documentation

## Changelog


rolfyone (approved):
LGTM thanks!

PR #9705, opened 2025-07-23
Introduced a default exception handler into our async runners

rolfyone:
These tasks can seemingly cause issues with onTick under some circumstances...

Increased the executors to a minimum of 6 (from 5).

## Documentation

## Changelog


tbenr:
I did some additional research on the `UncaughtExceptionHandler`. It seems to me that when there is an uncaught exception the thread will be terminated anyway, the only difference is which handler will run (a default one or a custom one).
The Executor class will detect that a thread died and will start a new one to cover the loss.

So, per my understanding, having a custom handler should just let us having a nicer log but it won't change the app logic.

It is a little bit unclear to me how a thread termination can cause this stall on the onTick. I was trying to find the logs inspiring this PR but was not able to.

rolfyone:
this isn't actually doing anything but a slight log cleanup, closing.

PR #9707, opened 2025-07-24
fix: Bytes32Converter to catch IllegalArgumentException for hex parsing

Galoretka:
Previously, Bytes32Converter only caught NumberFormatException when parsing a hex string to Bytes32. However, Bytes32.fromHexStringStrict can throw IllegalArgumentException for various invalid inputs (e.g., incorrect length, missing 0x prefix).  
This commit updates the converter to catch IllegalArgumentException, ensuring all invalid hex formats are properly handled and reported as a TypeConversionException.

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/converter/Bytes32Converter.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/converter/Bytes32Converter.java
@@ -22,7 +22,7 @@ public class Bytes32Converter implements ITypeConverter<Bytes32> {
   public Bytes32 convert(final String value) {
     try {
       return Bytes32.fromHexStringStrict(value);
-    } catch (final NumberFormatException e) {
+    } catch (final IllegalArgumentException e) {
       throw new TypeConversionException(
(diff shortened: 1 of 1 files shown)


rolfyone:
Thanks for raising.

I don't feel super strongly on this one because it's only used in CLI arguments, and our contrib guidelines are generally against this kind of non functional change.

Below is the unchanged code
```
$TEKU --Xnetwork-terminal-block-hash-override 0x00aa11
Invalid value for option '--Xnetwork-terminal-block-hash-override': cannot convert '0x00aa11' to Bytes32 (java.lang.IllegalArgumentException: Expected 32 bytes but got 3)

To display full help:
teku [COMMAND] --help
➜  teku git:(fix/bytes32converter-catch-illegalargumentexception) ✗ $TEKU --Xnetwork-terminal-block-hash-override 0xzzzz
Invalid value for option '--Xnetwork-terminal-block-hash-override': cannot convert '0xzzzz' to Bytes32 (java.lang.IllegalArgumentException: Illegal character 'z' found at index 0 in hex binary representation)

To display full help:
teku [COMMAND] --help

```

Closing.

PR #9713, opened 2025-07-25
Fix slow sync outside DA

### LINKED ISSUE #9708: Syncing outside data availability window is slow
https://github.com/Consensys/teku/blob/6faba5e5406548739c3edf158987cc7598b49f70/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceBlobSidecarsAvailabilityChecker.java#L75-L89

when we do forward sync outside the DA, we end up waiting the `.orTimeout(waitForTrackerCompletionTimeout)`  and then we realize we are outside the DA via `isBlockOutsideDataAvailabilityWindow`.

So i think the syncing logic won't try to download any blob, the BlockBlobSidecarsTracker is DA window agnostic, so the tracker never completes and we only do the DA window check when the wait timeouts.

IIRC that logic was there to handle the 

It can be replicated by running a local interop setting a very small window via:
`MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS: 3`
wait like 4 or more epochs and try to sync from genesis.

another way could be to take a old state (> 18days) from a public testnet\mainnet and try to sync from that

tbenr:
fixes #9708

I also did a class rename to be more consistent with the DataColumn availability checker

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/BlobSidecarsAvailabilityCheckerTest.java
+++ b/ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/BlobSidecarsAvailabilityCheckerTest.java
@@ -48,7 +48,7 @@ import tech.pegasys.teku.statetransition.blobs.BlockBlobSidecarsTracker;
 import tech.pegasys.teku.storage.client.RecentChainData;
 import tech.pegasys.teku.storage.store.UpdatableStore;
 
-public class ForkChoiceBlobSidecarsAvailabilityCheckerTest {
+public class BlobSidecarsAvailabilityCheckerTest {
 
   private final DataStructureUtil dataStructureUtil =
       new DataStructureUtil(TestSpecFactory.createMinimalDeneb());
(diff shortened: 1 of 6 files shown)


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/BlobSidecarsAvailabilityCheckerTest.java
```diff
@@ -192,9 +192,44 @@ void shouldReturnNotAvailableOnTimeout() throws Exception {
...
+  @Test
+  void shouldReturnNotRequiredWhenBlockIsOutsideAvailabilityWindowNoWait() throws Exception {
+    prepareForImmediateTimeoutWithBlockAndBlobSidecarsOutsideAvailabilityWindow();
+
+    final SafeFuture<DataAndValidationResult<BlobSidecar>> availabilityCheckResult =
+        blobSidecarsAvailabilityChecker.getAvailabilityCheckResult();
+
+    assertThat(blobSidecarsAvailabilityChecker.initiateDataAvailabilityCheck()).isTrue();
+
+    Waiter.waitFor(availabilityCheckResult);
```
zilm13: Maybe `Waiter.waitFor(availabilityCheckResult, Duration.ofMillis(1));` or it doesn't match  NoWait from the tests name

tbenr: ah yeah let me make it look better

zilm13 (approved):
LGTM
but from my opinion check should be "circuit broke" immediately after it's realized that it's outside DA, we shouldn't wait for timeout, download smth etc

PR #9716, opened 2025-07-28
Added exception reporting to recover sidecars

rolfyone:
Without any error catching the recovery would always print 'PLEASE FIX OR REPORT', and I'd like to avoid that.

It still looks like `onNewValidatedDataColumnSidecar` can potentially do that, but the ones i saw were in `reconstructAllDataColumnSidecars`.

Also added the slot number as context for the error.

## Documentation

## Changelog

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
@@ -292,7 +292,6 @@ public class MiscHelpersFulu extends MiscHelpersElectra {
                         new KZGCell(dataColumnSidecar.getDataColumn().get(rowIndex).getBytes()),
                         dataColumnSidecar.getIndex().intValue()))
             .collect(Collectors.toList());
-
     return kzg.verifyCellProofBatch(
         dataColumnSidecar.getSszKZGCommitments().stream()
             .map(SszKZGCommitment::getKZGCommitment)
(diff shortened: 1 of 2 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
```diff
@@ -485,8 +484,11 @@ private List<DataColumnSidecar> constructDataColumnSidecars(
   public List<DataColumnSidecar> reconstructAllDataColumnSidecars(
       final Collection<DataColumnSidecar> existingSidecars, final KZG kzg) {
     if (existingSidecars.size() < (specConfigFulu.getNumberOfColumns() / 2)) {
+      final Optional<DataColumnSidecar> maybeSidecar = existingSidecars.stream().findAny();
       throw new IllegalArgumentException(
-          "Number of sidecars must be greater than or equal to the half of column count");
+          String.format(
+              "Number of sidecars must be greater than or equal to the half of column count, slot: %s",
```
zilm13, line 490: could we also log `existingSidecars.size()` please. It's a bug call, just to understand either it's 0 or some positive number.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -279,6 +279,11 @@ private void initiateRecovery(
                   block.getSlotAndBlockRoot());
             })
         .alwaysRun(timer.closeUnchecked())
+        .exceptionally(
```
zilm13: nit:
we can do
```java
        .finish(
            error -> {
              LOG.error("Failed to recover columns, encountered error: {}", error.getMessage());
            });
```
and remove ifExceptionGetsHereRaiseABug

zilm13 (approved):
LGTM

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
rolfyone: does kind of seem weird we say not enough columns and part of the process is fetching columns is the only thing i guess...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/MiscHelpersFulu.java
rolfyone: was thinking that... slot was a good improvement, but can add that too.

PR #9717, opened 2025-07-28
Genesis cgc initialization

tbenr:
supersedes #9684

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
@@ -15,12 +15,14 @@ package tech.pegasys.teku.statetransition.datacolumns;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.units.bigints.UInt256;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import tech.pegasys.teku.ethereum.events.SlotEventsChannel;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.metrics.SettableGauge;
 import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
@@ -113,7 +115,7 @@ public class CustodyGroupCountManagerImpl implements SlotEventsChannel, CustodyG
       return;
     }
 
-    computeAndUpdateCustodyGroupCount(preparedValidators);
+    computeAndUpdateCustodyGroupCount(preparedValidators).ifExceptionGetsHereRaiseABug();
   }
 
   private boolean detectGenesisInitialization(
@@ -133,18 +135,22 @@ public class CustodyGroupCountManagerImpl implements SlotEventsChannel, CustodyG
     genesisInitialized = true;
 
     LOG.info("Validators at genesis epoch detected, initializing custody group count.");
-    computeAndUpdateCustodyGroupCount(preparedValidators);
+    computeAndUpdateCustodyGroupCount(preparedValidators)
+        .thenAccept(
+            maybeCustodyGroupCountUpdated ->
+                maybeCustodyGroupCountUpdated.ifPresent(this::setCustodyGroupSyncedCount))
+        .ifExceptionGetsHereRaiseABug();
     return true;
   }
 
-  private void computeAndUpdateCustodyGroupCount(
+  private SafeFuture<Optional<Integer>> computeAndUpdateCustodyGroupCount(
       final Map<UInt64, PreparedProposerInfo> preparedProposerInfo) {
-    combinedChainDataClient
+    return combinedChainDataClient
         .getBestFinalizedState()
-        .thenAccept(
+        .thenApply(
             maybeState -> {
               if (maybeState.isEmpty()) {
-                return;
+                return Optional.empty();
               }
 
               final int custodyGroupCountUpdated =
@@ -154,12 +160,9 @@ public class CustodyGroupCountManagerImpl implements SlotEventsChannel, CustodyG
                       .max(initCustodyGroupCount)
                       .intValue();
               updateCustodyGroupCount(custodyGroupCountUpdated);
-              if (maybeState.get().getSlot().isZero()) {
-                // genesis state
-                setCustodyGroupSyncedCount(custodyGroupCountUpdated);
-              }
-            })
-        .ifExceptionGetsHereRaiseABug();
+
+              return Optional.of(custodyGroupCountUpdated);
+            });
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -93,29 +94,69 @@ public void onSlot(final UInt64 slot) {
...
+    if (genesisInitialized) {
+      return false;
+    }
+
+    genesisInitialized = true;
+
+    LOG.info("Validators at genesis epoch detected, initializing custody group count.");
+    computeAndUpdateCustodyGroupCount(preparedValidators);
+    return true;
+  }
```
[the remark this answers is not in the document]
tbenr: onSlot is never called concurrently on multiple threads

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -158,15 +199,15 @@ public void setCustodyGroupSyncedCount(final int custodyGroupSyncedCount) {
     custodyGroupSyncedCountGauge.set(custodyGroupSyncedCount);
   }
 
-  private synchronized boolean updateEpoch(final UInt64 epoch) {
+  private boolean updateEpoch(final UInt64 epoch) {
     if (!lastEpoch.equals(epoch)) {
```
zilm13, line 236: maybe make it AtomicReference with removal of synchronized? Or probably better - check that we are increasing it

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -93,29 +94,69 @@ public void onSlot(final UInt64 slot) {
...
+              }
+
+              final int custodyGroupCountUpdated =
+                  miscHelpersFulu
+                      .getValidatorsCustodyRequirement(
+                          maybeState.get(), preparedProposerInfo.keySet())
+                      .max(initCustodyGroupCount)
+                      .intValue();
+              updateCustodyGroupCount(custodyGroupCountUpdated);
+              if (maybeState.get().getSlot().isZero()) {
```
zilm13: I don't like it being here. This method and part of code designed to handle custody group count, not synced. How about return `SafeFuture<Void>` from this method and act where we called it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -93,29 +94,69 @@ public void onSlot(final UInt64 slot) {
...
       updateCustodyGroupCount(initCustodyGroupCount);
       return;
     }
 
+    computeAndUpdateCustodyGroupCount(preparedValidators);
+  }
+
+  private boolean detectGenesisInitialization(
+      final UInt64 currentEpoch, final Map<UInt64, PreparedProposerInfo> preparedValidators) {
+    if (!currentEpoch.isZero()) {
```
zilm13: how about such order and little adjustment to logic, looks a bit safer for me
```java
    if (genesisInitialized) {
        return false;
        }
        
    if (!currentEpoch.isZero()) {
      genesisInitialized = true;  
      return false;
    }

    if (preparedValidators.isEmpty()) {
      return false;
    }

```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
tbenr: the point is that if that is supposed to be called onSlot, multiple onSlot cannot run concurrently by design and they run in in order so it shouldn't be a problem. I can revisit that anyway.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
tbenr: No I actually want to run that as soon as we have a some prepared validators.
Validator preparation depends on VC implementation (we call it at slot 3 IIRC), so the idea is to constantly check that along epoch 0 and do the initialization as soon as we have some validators.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -93,32 +96,73 @@ public void onSlot(final UInt64 slot) {
...
-            })
-        .ifExceptionGetsHereRaiseABug();
+
+              final int custodyGroupCountUpdated =
+                  miscHelpersFulu
+                      .getValidatorsCustodyRequirement(
+                          maybeState.get(), preparedProposerInfo.keySet())
+                      .max(initCustodyGroupCount)
+                      .intValue();
+              updateCustodyGroupCount(custodyGroupCountUpdated);
```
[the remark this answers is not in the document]
tbenr: I want that flow to run only once and I don't want to complicate it to handle potential multiple concurrent inflight updates. After all It is not important if we miss that initialization

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -93,32 +96,73 @@ public void onSlot(final UInt64 slot) {
...
+    }
+
+    genesisInitialized = true;
+
+    LOG.info("Validators at genesis epoch detected, initializing custody group count.");
+    computeAndUpdateCustodyGroupCount(preparedValidators)
         .thenAccept(
+            maybeCustodyGroupCountUpdated ->
+                maybeCustodyGroupCountUpdated.ifPresent(this::setCustodyGroupSyncedCount))
+        .ifExceptionGetsHereRaiseABug();
```
rolfyone: can we just finish with error rather than stack dumping and asking for a bug?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
```diff
@@ -93,32 +96,73 @@ public void onSlot(final UInt64 slot) {
...
+    }
+
+    if (preparedValidators.isEmpty()) {
       updateCustodyGroupCount(initCustodyGroupCount);
       return;
     }
 
-    combinedChainDataClient
-        .getBestFinalizedState()
+    computeAndUpdateCustodyGroupCount(preparedValidators).ifExceptionGetsHereRaiseABug();
```
rolfyone: imo also finish with error here

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
tbenr: i misread it, sorry! makes sense

zilm13 (approved):
LGTM

PR #9724, opened 2025-07-29
Implement Builder's blinded_blocks V2

zilm13:
## PR Description
Spec: https://github.com/ethereum/builder-specs/pull/123

New flow: we submit signed blinded block to the builder, builder response with 200 OK (if it's ok) and publishes block on its own without revealing it to our validator. So we will get the block with datacolumns  from gossip as usual even when we are block producer.
## Documentation

## Changelog


StefanBratanov (approved):
LGTM

PR #9726, opened 2025-07-30
proof of concept for alternative to ifExceptionGetsHereRaiseABug

rolfyone:
If this is generally accepted, the next PR i'd rename `ifExceptionGetsHereRaiseABug` to something like `finishStackTrace` or something, but that will be better to be separate from this general concept, as it'll besignificant noise.

## Documentation

## Changelog


zilm13:
I thought one of the (main) reasons we don't like `ifExceptionGetsHereRaiseABug()` because it means we (developers) didn't put enough effort to make comprehensive logging in a new code and it misses proper (user readable) message and usually useful details which could help in debugging in reproduction. This PR will "legalize" the shortcut.

Looking at code it's all ok for me here.

rolfyone:
> This PR will "legalize" the shortcut.

its just really a shortcut for what people are doing. it's not being looked at in code reviews clearly, so either we need to change processes or we need to accept that its a reality and at least not dump 'FIX OR REPORT' for every exception

mehdi-aouadi:
+1 for 

> This PR will "legalize" the shortcut

I'd prefer completely removing any "lazy" way to handle errors since we're at it. IMO you're responsible of handling your code's error use cases and should perform the appropriate actions accordingly. The new way is cleaner but still allows lazy error delegation... It's still easy to call finishxxx and delegate the error handling
We could at least deprecate those new methods immediately and remove them progressively

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
```diff
@@ -30,12 +32,29 @@
...
 public class SafeFuture<T> extends CompletableFuture<T> {
-
+  private static final Logger LOG = LogManager.getLogger();
   public static final SafeFuture<Void> COMPLETE = SafeFuture.completedFuture(null);
 
+  static void finishUsingExceptionMessage(final CompletionStage<?> future, final LogLevel level) {
+    future.exceptionally(
+        error -> {
+          final String message = Throwables.getRootCause(error).getMessage();
+          switch (level) {
```
tbenr: you can use `LOG.log()` method

rolfyone: its more ugly, it doesnt apply coloring, i did that first...

rolfyone:
> I'd prefer completely removing any "lazy" way to handle errors since we're at it.

sure, but there was a strong preference on standup to 'have an easy way' which is what i was doing, so people need to decide what they are happy to do ultimately.
the 'easy' way currently is asking for bugs, and its far too commonly used.

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
tbenr: humm why `LOG.warn(` is not equivalent to `LOG.log(Level.INFO`
We already use LOG.log in our code base....

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
```diff
@@ -315,6 +337,26 @@ public void ifExceptionGetsHereRaiseABug() {
...
+  }
+
+  public void finishDebug() {
+    finishUsingExceptionMessage(this, LogLevel.DEBUG);
+  }
+
+  public void finishTrace() {
+    finishUsingExceptionMessage(this, LogLevel.TRACE);
+  }
+
```
tbenr: I wander if, to preserve the log class, we could pass the LOG as a parameter.

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
tbenr: they literally run the same code

```java
    @Override
    public void log(final Level level, final String message) {
        logIfEnabled(FQCN, level, null, message, (Throwable) null);
    }
    
    @Override
    public void info(final String message) {
        logIfEnabled(FQCN, Level.INFO, null, message, (Throwable) null);
    }
```

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
rolfyone: well, that violates your 'lets not have parameters', but possibly

infrastructure/async/src/test/java/tech/pegasys/teku/infrastructure/async/SafeFutureTest.java
```diff
@@ -87,6 +93,66 @@ public void tearDown() {
...
+  public void shouldLogMessageAtDebugLevel() {
+    try (LogCaptor logCaptor = LogCaptor.forClass(SafeFuture.class)) {
+      final SafeFuture<Void> future = new SafeFuture<>();
+      future.finishDebug(LOG);
+
+      future.completeExceptionally(new IllegalArgumentException(COMPUTER_SAYS_NO));
+
+      assertThat(logCaptor.getDebugLogs().getFirst()).contains(COMPUTER_SAYS_NO);
+    }
+  }
```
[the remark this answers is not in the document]
lucassaldanha: NIT: Log captor seems to be capturing the expected logs anyway. We could change it to `SafeFutureTest.class` for the sake of consistency but it isn't affecting the test.

lucassaldanha:
> its just really a shortcut for what people are doing. it's not being looked at in code reviews clearly, so either we need to change processes or we need to accept that its a reality and at least not dump 'FIX OR REPORT' for every exception

IMHO there are two things being discussed in here:

1. We don't like the way we are dumping `FIX OR REPORT` in user logs. This PR will address that, and adds a "shortcut" to handle logging errors in our futures.
2. We already have `ifExceptionGetsHereRaiseABug`, but we don't like it. We should talk about it and potentially change our approach to error handling.

This PR is only taking care of point 1, so it is good to go.

lucassaldanha (approved):

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/SafeFuture.java
rolfyone: this is actioned now.

PR #9734, opened 2025-07-31
Complete late block reorg

tbenr:
## Documentation

## Changelog

### DIFF
--- a/teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
+++ b/teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
@@ -509,7 +509,6 @@ public class Eth2NetworkOptions {
             aggregatingAttestationPoolV2BlockAggregationTimeLimit)
         .aggregatingAttestationPoolV2TotalBlockAggregationTimeLimit(
             aggregatingAttestationPoolV2TotalBlockAggregationTimeLimit)
-        .lateBlockProductionEnabled(lateBlockProduction)
         .epochsStoreBlobs(epochsStoreBlobs)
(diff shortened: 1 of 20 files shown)


tbenr:

teku/src/main/java/tech/pegasys/teku/cli/options/Eth2NetworkOptions.java
```diff
@@ -142,6 +142,16 @@ public class Eth2NetworkOptions {
   private boolean forkChoiceLateBlockReorgEnabled =
       Eth2NetworkConfiguration.DEFAULT_FORK_CHOICE_LATE_BLOCK_REORG_ENABLED;
 
+  @Option(
+      names = {"--Xlate-block-production"},
```
rolfyone: i think our standards would require that we put `enabled` at the end

rolfyone: im not  100% sure what this flag does that the reorg flag above doesnt do, but i assume its used.. didnt check closely

tbenr: ahh no this won't be part of the final PR. This is only for my testing to produce late block.
You can ignore all stuff related to that param

tbenr: I removed it from the draft

tbenr:
I'm a bit concerned by the fact that the "operationsReorgCompleted" event can sometimes take too much to arrive

image


this is taken from holesky01. Let's get more data points and we can reevaluate

tbenr:
closing in favor of #9817

PR #9739, opened 2025-08-01
retry blobs fetching from local EL

mehdi-aouadi:
## PR Description
Retry blobs fetching from the local EL when the first call fails.
In this version, we retry 3 times every 500ms. These params could be configurable

## Documentation

## Changelog


ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/DataColumnSidecarELRecoveryManagerImplTest.java
```diff
@@ -229,8 +231,141 @@ public void shouldPublish_whenAllBlobsRetrieved() {
...
+      asyncRunner.executeQueuedActions();
+      verify(executionLayer, times(i + 1)).engineGetBlobAndCellProofsList(any(), any());
+    }
+    assertThat(asyncRunner.hasDelayedActions()).isFalse();
+    verifyNoInteractions(dataColumnSidecarPublisher);
+  }
+
+  @Test
+  @SuppressWarnings("unchecked")
+  public void shouldStopRetry_whenElBlobsFetchingIsDone() {
```
zilm13: Please add test for this case:
```java
              if (blobAndCellProofsList.isEmpty()) {
                LOG.debug(
                    "Blobs for {} are not found on local EL, reconstruction is not possible",
                    slotAndBlockRoot);
                return;
              }
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/util/DataColumnSidecarELRecoveryManagerImpl.java
```diff
@@ -332,12 +339,11 @@ private void onFirstSeen(
     }
 
     asyncRunner
-        .runAsync(
-            () ->
-                // fetch blobs from EL with no delay
-                fetchMissingBlobsFromLocalEL(slotAndBlockRoot))
-        .handleException(this::logLocalElBlobsLookupFailure)
-        .ifExceptionGetsHereRaiseABug();
+        .runWithRetry(
```
zilm13, line 350: Before change we were not expecting recovery started after receiving all dataColumnSidecars through gossip because we start on first sidecar (or even block) and require at least 4. We were not able to get other sidecars immediately. But now as we wait 500ms for retry, next retry could happen after receiving all required sidecars by gossip so we should stop retrying. 
TLDR: check that we have not already received sidecars in custody and sampler before next retry. The simplest way is to catch it all here as far as we already subscried onNewValidSidecar

mehdi-aouadi: As discussed, added an index tracking to check if we haven't already received the missing columns before retrying fetching them from the EL

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/util/DataColumnSidecarELRecoveryManagerImpl.java
```diff
@@ -346,6 +361,13 @@ private SafeFuture<Void> fetchMissingBlobsFromLocalEL(final SlotAndBlockRoot slo
     if (recoveryTask == null) {
       return SafeFuture.COMPLETE;
     }
+
+    if (recoveryTask
+        .recoveredColumnIndices()
+        .containsAll(custodyGroupCountManager.getCustodyColumnIndices())) {
```
zilm13: We need to switch here and in all other usages (3 total) to sampling column indices. It would be more correct for full node role. Otherwise full node sampling task will not be finished from successful recovery.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/DataColumnSidecarELRecoveryManagerImplTest.java
```diff
@@ -229,11 +239,274 @@ public void shouldPublish_whenAllBlobsRetrieved() {
...
+                    new BlobAndCellProofs(
+                        blobSidecar.getBlob(),
+                        IntStream.range(0, 128)
+                            .mapToObj(__ -> dataStructureUtil.randomKZGProof())
+                            .toList()))
+            .toList();
+    // 2 first call fails and then the 3rd succeeds
+    when(executionLayer.engineGetBlobAndCellProofsList(any(), any()))
+        .thenReturn(SafeFuture.failedFuture(new IllegalArgumentException("error")))
+        .thenReturn(SafeFuture.failedFuture(new IllegalArgumentException("error")))
```
zilm13: You have 3 retries and configurations with first 2 error responses, so you are not testing that it stopped retrying because  blobs were fetched from EL. I'd just remove second error answer and adjust verifications accordingly

zilm13 (approved):
LGTM, just few nits, see comments

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/DataColumnSidecarELRecoveryManagerImplTest.java
mehdi-aouadi: I did that and added a test to check we stop after we perform max retries...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/util/DataColumnSidecarELRecoveryManagerImpl.java
mehdi-aouadi: Makes sense, I updated this one. What other usages are you referring to?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/util/DataColumnSidecarELRecoveryManagerImpl.java
```diff
@@ -191,12 +201,14 @@ private boolean createRecoveryTaskFromDataColumnSidecar(
             new RecoveryTask(
                 dataColumnSidecar.getSignedBeaconBlockHeader(),
                 dataColumnSidecar.getSszKZGCommitments(),
-                dataColumnSidecar.getKzgCommitmentsInclusionProof().asListUnboxed()))
+                dataColumnSidecar.getKzgCommitmentsInclusionProof().asListUnboxed(),
+                Collections.newSetFromMap(new ConcurrentHashMap<>())))
```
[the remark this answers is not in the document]
mehdi-aouadi: we don't count the received data column sidecar since we abstract that away and we fetch them based on the slot and the block root

PR #9742, opened 2025-08-04
Allow BPO forks at fork epoch

zilm13:
## PR Description
Allows bpo fork to start at common fork epoch. It was working in devnets 0-2 but doesnt work anymore. This PR should fix this.

## Documentation

## Changelog


ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
```diff
@@ -113,6 +114,27 @@ public void blobParametersFuluEpoch() {
...
+                            fb ->
+                                fb.fuluForkEpoch(fuluEpoch)
+                                    .blobSchedule(
+                                        List.of(
+                                            new BlobScheduleEntry(fuluEpoch, 6),
+                                            new BlobScheduleEntry(fuluEpoch.increment(), 9),
+                                            new BlobScheduleEntry(fuluEpoch, maxBlobsPerBlock))))))
+        .hasMessageContaining("Blob schedule must be ordered and doesn't contain duplicates, while")
+        .isInstanceOf(IllegalArgumentException.class);
+  }
```
lucassaldanha: This test is only for the ordering. Are we missing a test for duplicates?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/builder/FuluBuilder.java
```diff
@@ -107,13 +107,30 @@ public FuluBuilder kzgCommitmentsInclusionProofDepth(
...
+  public void verifyBlobSchedule(final List<BlobScheduleEntry> blobSchedule) {
+    if (blobSchedule.isEmpty()) {
+      return;
+    }
+    BlobScheduleEntry previousEntry = blobSchedule.getFirst();
+    for (final BlobScheduleEntry entry : blobSchedule.subList(1, blobSchedule.size())) {
+      if (!entry.epoch().isGreaterThan(previousEntry.epoch())) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Blob schedule must be ordered and doesn't contain duplicates, while %s and %s entries were provided.",
```
rolfyone: we were previously just sorting on epoch, im not sure where that went but imo sorting is probably better. we just want to error if theres the same epoch re-specified.

we want to also be careful here with user facing wording, if we want to do this particular check, we need more like`Blob schedule must be in ascending order, and not contain duplicates.  %s and %s in the provided blob schedule do not meet this requirement.`

rolfyone:
> Allows bpo fork to start at common fork epoch. It was working in devnets 0-2 but doesnt work anymore. This PR should fix this.

bpo were only introduced devnet-2, and aren't normal forks... its just an attribute of a fork, not a fork (only digest changes)

StefanBratanov:
We need to be careful here with `.getForkDigest(epoch)` and `.getNextForkDigest(epoch)` methods in `RecentChainData`. It is a bit confusing to me that we never subscribe to Fulu fork digest in this case, we just need consistency between all those methods. In this case, if BPO and hard fork on same epoch, we prioritize BPO, just need tests for those and code change if needed. We also need to make sure `ActiveEthP2PNetwork.updateForkInfo` and `DiscoveryNetwork.setForkInfo` is not affected. `Eth2GossipTopicFilter` is another one that's important to check. All of those have tests, so it's a case of adding a test for this use case.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/builder/FuluBuilder.java
StefanBratanov: Technically, all blob schedule logic should be using `BpoForkSchedule` class which sorts it, but it is good to have it sorted regardless just in case we call `specConfig.getBlobSchedule()` in other place.

zilm13:
@StefanBratanov @rolfyone all feedback addressed

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
zilm13: You are right. I've modified the test, especially we don't need order checks

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
```diff
@@ -113,6 +114,59 @@ public void blobParametersFuluEpoch() {
         .isEqualTo(new BlobParameters(fuluEpoch, maxBlobsPerBlock));
   }
 
+  @Test
+  public void blobParameterSinglesEntryFuluEpoch() {
```
rolfyone: singles? im now sure what we mean here - `Single`?

StefanBratanov: I guess can rename to `blobParameterSameEpochAsFuluForkEpoch()`

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/ActiveEth2P2PNetworkTest.java
```diff
@@ -205,6 +210,66 @@ public void onEpoch_shouldUpdateDiscoveryNetworkForkInfo() {
...
+    network.onEpoch(UInt64.ONE);
+    reset(discoveryNetwork);
+    // At the fulu upgrade epoch, we should update fork info
+    network.onEpoch(fuluForkEpoch);
+    expectedFork = new ForkInfo(fuluFork, genesisValidatorsRoot);
+    verify(discoveryNetwork)
+        .setForkInfo(
+            expectedFork, fuluForkDigest, Optional.empty(), Optional.empty(), Optional.empty());
+
+    // At the BPO fork upgrade epoch, we should update fork info
```
StefanBratanov: this comment should be `// Processing BPO fork shouldn't cause any updates since it's the same epoch as fulu fork`

StefanBratanov (approved):
LGTM with couple of comments

ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/builder/FuluBuilder.java
StefanBratanov: I can suggest bit more simple version with ordering also:

```java
  public FuluBuilder blobSchedule(final List<BlobScheduleEntry> blobSchedule) {
    checkNotNull(blobSchedule);
    verifyBlobSchedule(blobSchedule);
    this.blobSchedule.clear();
    blobSchedule.stream()
        .sorted(Comparator.comparing(BlobScheduleEntry::epoch))
        .forEach(this.blobSchedule::add);
    return this;
  }

  private void verifyBlobSchedule(final List<BlobScheduleEntry> blobSchedule) {
    final Set<UInt64> seenEpochs = new HashSet<>();
    for (final BlobScheduleEntry entry : blobSchedule) {
      if (!seenEpochs.add(entry.epoch())) {
            String.format(
                "There are duplicate entries for epoch %s in blob schedule.", entry.epoch()));
      }
    }
  }
```

ethereum/spec/src/test/java/tech/pegasys/teku/spec/config/SpecConfigFuluTest.java
zilm13: Single. Sorry for wording. I meant single bpo defined to test that my duplicates logic doesn't fail when there is only single `BlobScheduleEntry`

PR #9748, opened 2025-08-06
Added performance test for rate tracker

rolfyone:
I was trying to replace it and simplify, but we really do need the synchronized and the duplication of the count I think ultimately.

When I was using a list i got some faster results under some conditions, but not really enough to consider changing the old one out, as I wasn't able to match speed until i included the synchronized sections and count cache i was trying to avoid having.

Renamed some classes etc once I figured out what they were to hopefully make it clearer next time.

The performance test we can use in future.

After Initial commit i'll cleanup `RateTrackerRequests`, `NewRateTrackerImpl`, `RateTrackerRequests`

`RateTrackerTest` had some complicated scenarios that were mostly covered by other test cases, will double check that bit.
## Documentation

## Changelog


rolfyone:
```
Benchmark                                  (isEmpty)  (useNewRateTracker)   Mode  Cnt         Score         Error  Units
RateTrackerBenchmark.addToList                  true                 true  thrpt   10   6902539.375 ±  222801.682  ops/s
RateTrackerBenchmark.addToList                  true                false  thrpt   10  10945036.053 ±  199994.631  ops/s
RateTrackerBenchmark.addToList                 false                 true  thrpt   10  14623544.844 ±  870838.543  ops/s
RateTrackerBenchmark.addToList                 false                false  thrpt   10  11265999.713 ± 1660072.294  ops/s
RateTrackerBenchmark.adjustRequest              true                 true  thrpt   10   5312319.274 ± 1196566.523  ops/s
RateTrackerBenchmark.adjustRequest              true                false  thrpt   10   5935261.070 ± 1468568.425  ops/s
RateTrackerBenchmark.adjustRequest             false                 true  thrpt   10   4672820.152 ±  349555.082  ops/s
RateTrackerBenchmark.adjustRequest             false                false  thrpt   10  14178382.948 ± 1892171.590  ops/s
RateTrackerBenchmark.availableObjectCount       true                 true  thrpt   10  20643865.095 ±  164705.308  ops/s
RateTrackerBenchmark.availableObjectCount       true                false  thrpt   10  20704294.045 ±  227145.734  ops/s
RateTrackerBenchmark.availableObjectCount      false                 true  thrpt   10  22579228.250 ± 3427865.904  ops/s
RateTrackerBenchmark.availableObjectCount      false                false  thrpt   10  15873835.156 ± 2988911.116  ops/s
```

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/ApprovedRequest.java
```diff
@@ -16,22 +16,26 @@
...
-public class RequestApproval {
+public class ApprovedRequest {
 
-  private final RequestsKey requestKey;
-  private final long objectsCount;
+  private final RequestKey requestKey;
+  private long requestSize;
 
-  private RequestApproval(final RequestsKey requestKey, final long objectsCount) {
+  private ApprovedRequest(final RequestKey requestKey, final long requestSize) {
```
lucassaldanha, line 24: NIT: maybe it is me but `requestSize` makes me think about size in bytes instead of the limit of objects being requested.

The previous name also wasn't clear. Maybe something like `maxObjects` or `maxObjectsPerRequest` is clearer?

lucassaldanha (approved):
LGMT. Just a suggestion re naming.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/ApprovedRequest.java
rolfyone: to me `objectsCount` was wrong, i could change it to `objectCount` or `objects`, but imo `requestSize` is more accurate than `objectsCount`... maybe just `objects`?

rolfyone: will address this in a more complete way in a followup

Issue #9556, opened 2025-06-13
Update syncing logic to consume and update earliest_slot_available from StatusV2 message


lucassaldanha:
Reference: https://github.com/ethereum/consensus-specs/pull/4374
Related to: #9539

Tasks:
- [x] Consume earliest_slot_available when selecting peers; 
- [x] Update earliest_slot_available based on local chain data (https://github.com/Consensys/teku/pull/9762)
- [x] Updated earliest_slot_available when backfilling (?)

zilm13:
- Consume earliest_slot_available when selecting peers is done in #10269 
- Updated earliest_slot_available when backfilling is done in #10226

PR #9762, opened 2025-08-12
Update earliest_available_slot for StatusV2

lucassaldanha:
## PR Description
Update `earliest_available_slot` field on StatusV2 message based on our local chain data. The update only happens once per epoch.

Spec reference: https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/p2p-interface.md?plain=1#L294-L295

## Fixed Issue(s)
part of #9556

## Documentation

## Changelog

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
@@ -100,7 +100,9 @@ public class StatusMessageFactory implements SlotEventsChannel {
 
   @Override
   public void onSlot(final UInt64 slot) {
-    updateEarliestAvailableSlot();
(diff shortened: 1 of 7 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
```diff
@@ -84,6 +90,28 @@ public Optional<StatusMessage> createStatusMessage(final StatusMessageSchema<?>
...
+                .get()
+                .or(
+                    () ->
+                        Optional.of(
+                            spec.computeStartSlotAtEpoch(recentChainData.getFinalizedEpoch())))));
+  }
+
+  @Override
+  public void onSlot(final UInt64 slot) {
+    updateEarliestAvailableSlot();
```
lucassaldanha: Maybe we don't need this every slot...

zilm13: yeah, once per epoch should be ok

zilm13 (approved):
LGTM!

PR #9767, opened 2025-08-12
put some limits onto forkChoiceTrigger

rolfyone:
we're getting into a situation where we get stuck waiting for process head and processhead isn't running, which seems odd. this should at least raise errors and traces when its happening, and maybe not get stuck for long periods

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
```diff
@@ -13,26 +13,64 @@
...
+        LOG.warn(
+            "Took {} ms waiting for fork choice to complete at slot {}, when attestations were due.",
+            duration,
+            nodeSlot);
+      } else if (duration.isGreaterThanOrEqualTo(DEBUG_TIME_MILLIS)) {
+        LOG.debug(
+            "Took {} ms waiting for fork choice to complete at slot {}, when attestations were due.",
+            duration,
+            nodeSlot);
+        ;
```
lucassaldanha: Is this a rogue `;`? :)

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTriggerTest.java
```diff
@@ -42,6 +55,77 @@ void shouldProcessHeadOnSlotStartedWhileSyncing() {
...
+    processHeadFuture.complete(true);
+    int i = 0;
+    // shouldn't take 30 x 100ms to complete, but allowing in case
+    while (i++ < 30) {
+      Thread.sleep(100);
+      if (processHeadFuture.isCompletedNormally()) {
+        break;
+      }
+    }
+    attestationsDueFuture.cancel(true);
```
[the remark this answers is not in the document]
rolfyone: it's a test case, this is fine.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
```diff
@@ -13,26 +13,63 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import tech.pegasys.teku.ethereum.performance.trackers.BlockProductionPerformance;
 import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 
 public class ForkChoiceTrigger {
-
+  public static final int WARNING_TIME_MILLIS = 250;
```
lucassaldanha, line 27: I am not sure if a warning level is the best idea here. Since it doesn't signal something the operator can take action on, it might become just noise.

Also, for the debug, if the idea is to have it log pretty much constantly, maybe a metric would suit better?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
```diff
@@ -13,26 +13,63 @@
...
+            duration,
+            nodeSlot);
+      } else if (duration.isGreaterThanOrEqualTo(DEBUG_TIME_MILLIS)) {
+        LOG.debug(
+            "Took {} ms waiting for fork choice to complete at slot {}, when attestations were due.",
+            duration,
+            nodeSlot);
+      }
+    } catch (InterruptedException | TimeoutException e) {
+      LOG.error("Failed to wait for fork choice to complete for slot " + nodeSlot, e);
```
lucassaldanha, line 68: NIT: could add the value of `attestationWaitLimitMillis` to the message so we know how long we waited for w/o having to check the config on that node.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
lucassaldanha: Or we could just leave it all as debug and enable the logs only when investigating the slowness.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
rolfyone: it'd be in every message... i can add it if you think it wont just be noise... mostly i figured slot would be most informative

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
rolfyone: potentially i could add a metric, but we'd need to add to the dash etc etc and im not sure if we want that?

I basically applied the same theory as other slow things where we print a warning...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
lucassaldanha: I guess it doesn't matter too much if we aren't planning to play too much with the time limit config. We can assume it will be 1.5s most of the time.

I was thinking something similar to what we have for slow db operations:
```
DB operation KvStoreDatabase::doUpdate took too long: 1762 ms. The alert threshold is set to: 1000 ms.
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoiceTrigger.java
lucassaldanha: That's true. We do have a warning level for slow DB operations, but the slow DB operation logs aren't enabled by default. So perhaps to keep a similar approach, we would need to have these logs disabled by default.

lucassaldanha: The metric depends on how we plan to consume this information, but it isn't something we need right now. Moving forward, it would be interesting to have the metric on the dashboard and be able to correlate that to other metrics like system resources, execution queues, etc. Potentially, it could help us identify under what scenarios FC is taking longer.

rolfyone: at the moment we just need to know so that we can see it, and we cant have it just on archive nodes, because this only executes on attesting nodes which is why i did what i did... im tempted to just do that as part one and see how it goes on some non holesky nodes

lucassaldanha (approved):

Issue #9768, opened 2025-08-13
Feature: Make BUILDER_PROPOSAL_DELAY_TOLERANCE Configurable


meta-duck:
**Description:** 

Currently, the timeout value for `BUILDER_PROPOSAL_DELAY_TOLERANCE` is hardcoded to `Duration.ofSeconds(1)` in [ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java](https://github.com/Consensys/teku/blob/master/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java#L67). Allowing this value to be configurable would enable operators to adjust the timeout as needed for different environments or use cases.

**Motivation:** 

In some scenarios, increasing the timeout would allow the builder extra time to gather more transactions or MEV value, potentially improving block value and network efficiency. Making this value configurable allows for experimentation and optimization based on network conditions or operator needs.

**Proposed Solution:**

- Expose BUILDER_PROPOSAL_DELAY_TOLERANCE as a configurable parameter, e.g., via a config file, environment variable, or command-line flag.
- Retain the current `Duration.ofSeconds(1)` as the default if not explicitly set.
- Document the new configuration option for users.

**Benefits:**

- Increase validator's yield
- Greater flexibility for node operators.
- Ability to tune performance for different conditions.
- Supports experimentation and research on the effects of timeout adjustments.

rolfyone:
We could potentially accept the idea of reducing it, but I'm not comfortable going beyond the spec'd limit. My assumption is that you're interested in exceeding 1s, not going below 1s, is that a fair assumption @meta-duck?
Should I just close this if the only answer would be to allow reducing the timeout configurably?

lucassaldanha:
Closed based on comments on https://github.com/Consensys/teku/pull/9769

PR #9769, opened 2025-08-14
feat: Make BUILDER_PROPOSAL_DELAY_TOLERANCE Configurable

meta-duck:
**Description:** 

Currently, the timeout value for `BUILDER_PROPOSAL_DELAY_TOLERANCE` is hardcoded to `Duration.ofSeconds(1)` in [ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java](https://github.com/Consensys/teku/blob/master/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java#L67). Allowing this value to be configurable would enable operators to adjust the timeout as needed for different environments or use cases.

**Motivation:** 

In some scenarios, increasing the timeout would allow the builder extra time to gather more transactions or MEV value, potentially improving block value and network efficiency. Making this value configurable allows for experimentation and optimization based on network conditions or operator needs.

**Proposed Solution:**

- Expose BUILDER_PROPOSAL_DELAY_TOLERANCE as a configurable parameter, e.g., via a config file, environment variable, or command-line flag.
- Retain the current `Duration.ofSeconds(1)` as the default if not explicitly set.
- Document the new configuration option for users.

**Benefits:**

- Increase validator's yield
- Greater flexibility for node operators.
- Ability to tune performance for different conditions.
- Supports experimentation and research on the effects of timeout adjustments.

ref: https://github.com/Consensys/teku/issues/9768


meta-duck:

meta-duck:

rolfyone:
> Currently, the timeout value for `BUILDER_PROPOSAL_DELAY_TOLERANCE` is hardcoded to `Duration.ofSeconds(1)` in [ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java](https://github.com/Consensys/teku/blob/master/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java#L67). Allowing this value to be configurable would enable operators to adjust the timeout as needed for different environments or use cases.

Hey, the bigger thing here is that it's also part of the specification for builder flow 
https://github.com/ethereum/builder-specs/blob/main/specs/bellatrix/validator.md#constants

Is there a reason that we believe the specification of the builder is incorrect in allowing a 1 second constant here?

meta-duck:
> > Currently, the timeout value for `BUILDER_PROPOSAL_DELAY_TOLERANCE` is hardcoded to `Duration.ofSeconds(1)` in [ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java](https://github.com/Consensys/teku/blob/master/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java#L67). Allowing this value to be configurable would enable operators to adjust the timeout as needed for different environments or use cases.
> 
> Hey, the bigger thing here is that it's also part of the specification for builder flow https://github.com/ethereum/builder-specs/blob/main/specs/bellatrix/validator.md#constants
> 
> Is there a reason that we believe the specification of the builder is incorrect in allowing a 1 second constant here?

We can't say it's incorrect; 1 second is a safe buffer to avoid missing a slot or signing the block too late. 
By increasing the timeout value, we could allow the builder to have an extra 1 or 1.5 seconds to collect more value and submit a larger value block to the relay. The primary goal is to boost yield.

StefanBratanov:
I remember we had a similar feature request and the main consensus was that we don't want to make it easy to have a behavior which doesn't align with the specs and potentially have effects on the network which are hard to reason about. Anyone can feel free to fork the client and change the constant if they want more hacky/custom version.

rolfyone:
> We can't say it's incorrect; 1 second is a safe buffer to avoid missing a slot or signing the block too late.
> By increasing the timeout value, we could allow the builder to have an extra 1 or 1.5 seconds to collect more value and submit a larger value block to the relay. The primary goal is to boost yield.

Ok that's what I was assuming from the description.

In reality i think i'd be not against allowing a reduction from 1 second, but we can't really openly go off spec on this, especially not purely for improving margins (no offence). Because we're implementing an honest validator, we do have certain things we should be following, and this would be one of those sensitive areas that would potentially make the game 'unfair', which I'm fairly strongly against if Im being honest.

Especially going forward towards an ePBS future, this time will all be changing anyway, so i think for now we should close this PR on that basis. Adding to that shorter slots likely in a not too distant future, this 1 second is likely to be on borrowed time.

Apologies but I can't see us being able to accept this type of feature that's openly against the spec into our supported code base.

PR #9773, opened 2025-08-14
Drop peers with false ENR subnet advertising

zilm13:
## PR Description
Include attnets, sync committees, das subnets

## Documentation

## Changelog

---

> [!NOTE]
> Adds ENR-vs-subscription validation (attnets, syncnets, DAS) to penalize and drop falsely advertising peers, updating selection/disconnect flows and related APIs.
> 
> - **Peer validation & disconnection**:
>   - Introduces `PeerAdvertisementChecker` to verify ENR-advertised subnets against actual topic subscriptions (attestation, sync committee, data column sidecars) and apply reputation penalties.
>   - Enhances `Eth2PeerSelectionStrategy` to find "false advertising" peers (throttled checks), prioritize dropping them, and return `PeerSelectionStrategy.PeerToDisconnect` with reasons (`BAD_SCORE`, `TOO_MANY_PEERS`).
>   - Updates `ConnectionManager` to use the new disconnect API and reasons.
> - **Plumbing & builders**:
>   - `Eth2P2PNetworkBuilder`: wires an `AtomicReference<ActiveEth2P2PNetwork>` into `buildNetwork` so peer checks can access active network; passes `timeProvider`/references into `Eth2PeerSelectionStrategy`.
> - **Gossip/subnet utilities**:
>   - `PeerSubnetSubscriptions`: adds optional getters (`get*IfAvailable`) and supports DAS by node id with `custodyGroupCount`.
>   - `SubnetScorer` and `NodeIdToDataColumnSidecarSubnetsCalculator`: align with DAS group count terminology.
> - **Discovery peer API**:
>   - Renames `DiscoveryPeer.getDasCustodySubnetCount()` to `getDasCustodyGroupCount()` and includes `nodeId` in equality/hash/toString.
> - **Tests & fixtures**:
>   - Adds `PeerAdvertisementCheckerTest`; expands `Eth2PeerSelectionStrategyTest` (false advertising, throttling, reasons); adapts `ConnectionManagerTest` and test fixtures (`SimplePeerSelectionStrategy`, `Eth2P2PNetworkFactory`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0bdc061a971257193378e71ad475b1d7d652d5a4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
```diff
@@ -151,7 +167,9 @@ private Optional<PeerAddress> checkCandidate(
 
   @Override
   public List<Peer> selectPeersToDisconnect(
```
StefanBratanov: So I have been looking at the code and this is called here:

```
    peerSelectionStrategy
        .selectPeersToDisconnect(network, peerPools)
        .forEach(
            peerToDrop ->
                peerToDrop.disconnectCleanly(DisconnectReason.TOO_MANY_PEERS).finishTrace(LOG));
```

So it disconnects for the reason `TOO_MANY_PEERS`. I wonder if it is better to have a different method which checks the false advertising peers and then fails with a new DisconnectReason. Probably something like `FALSE_ADVERTISEMENT` or something like that.  Otherwise `TOO_MANY_PEERS` sounds wrong.

lucassaldanha: Maybe we can change this method to return an object with the list of peers to disconnect and the reason for that group. This way, we can always report on the exact reason for disconnecting.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
```diff
@@ -180,12 +231,86 @@ public List<Peer> selectPeersToDisconnect(
...
+    }
+
+    final DiscoveryPeer discoveryPeer = discoveryPeerMap.get(discoveryNodeId.get());
+
+    // Check for sync committee subnets
+    final SszBitvector advertisedSyncCommitteeSubnets = discoveryPeer.getSyncCommitteeSubnets();
+    final SszBitvector syncCommitteeSubnets =
+        peerSubnetSubscriptions.getAttestationSubnetSubscriptions(peer.getId());
+    if (!syncCommitteeSubnets.isSuperSetOf(advertisedSyncCommitteeSubnets)) {
+      reputationManager.adjustReputation(peer.getAddress(), ReputationAdjustment.LARGE_PENALTY);
```
lucassaldanha: Maybe we could add a debug entry here saying we are downscoring peer X (not sure if we log this somewhere else). This will make it a bit easier to see this mechanism in action. The same applies to the other checks below.

networking/eth2/src/testFixtures/java/tech/pegasys/teku/networking/eth2/Eth2P2PNetworkFactory.java
```diff
@@ -351,6 +352,8 @@ protected Eth2P2PNetwork buildNetwork(final P2PConfig config) {
                                 config.getTargetSubnetSubscriberCount(),
                                 subnetPeerCountGauge),
                         reputationManager,
+                        new AtomicReference<>(),
+                        NodeIdToDataColumnSidecarSubnetsCalculator.NOOP,
```
lucassaldanha, line 358: Is this NOOP correct here?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
StefanBratanov: > Maybe we can change this method to return an object with the list of peers to disconnect and the reason for that group. This way, we can always report on the exact reason for disconnecting.

I like that actually, I guess more like a map `DisconnectReason,List<Peer>`

networking/eth2/src/testFixtures/java/tech/pegasys/teku/networking/eth2/Eth2P2PNetworkFactory.java
zilm13: It's test fixtures, don't need it at this point there, I think

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
zilm13: How about this? https://github.com/Consensys/teku/pull/9773/commits/d92d991fc52779864db2f38adecd80d99e69d3fc
I cannot introduce a reason, they are specced (actually goodbye messages are specced, not the reasons, but this doesn't change much, we could use existing I think)

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
zilm13: Added https://github.com/Consensys/teku/pull/9773/commits/daf1b10d429d2c0d85698e7566a4667acaafe975

zilm13:
Version before adding cache:

23160 calls for 3 hours (~128 calls/minute)

50th percentile: 9 ms
90th percentile: 810 ms
95th percentile: 1487 ms
99th percentile: 3078 ms

Calls are made on different nio event threads
highest 10-20% looks very bad, looking for ways to improve

zilm13:
Version with cache:

23408 calls during 3 hours (130 calls per minute)

50th percentile: 9 ms
90th percentile: 921 ms 
95th percentile: 1696 ms
99th percentile: 3410 ms

Cache is not helping at all, trying another approach

zilm13:
It's a final version.
No cache. Run every 5 seconds. Runtime is even worse, but is it important? It runs on separate thread and rarely (but it's nio thread)

50th percentile: 97 ms
90th percentile: 2365 ms
95th percentile: 3068 ms
99th percentile: 4386 ms

~6000 calls during 10 hours

zilm13:
if we decide no merge, don't forget to make a pr for renaming subnet to group where it's applicable, it should be merged anyway

tbenr:
I tested this on fusaka-devnet-5 and seems like we consider peer lying on column subnets

zilm13:
TLDR;
Bug is found and fixed and we review only peers with available connected subnets data. But still there are a lot of false positives for peers of all possible clients: Lodestar, Prysm, Lighthouse. Maybe libp2p connected subnets data is not reliable, or it takes time to be filled, anyway it cannot be used to find out peers with false advertising. 

There was and idea that this could work, but it doesn't, so closing this PR

PR #9774, opened 2025-08-15
Update README.md

dedyshkaPexto:
link is 400 error, better to delete it. thanks


rolfyone:
The link is fine. not sure why you cant see that site, but it's all valid to keep on readme. closing.

PR #9780, opened 2025-08-19
Avoid builder validator registration calls potentially delaying calls on critical path

tbenr:
Concurrent batches of validator registration may fill up our throttler and can cause getHeader\getPayload to be queued up.

I think it is better to not throttle critical path calls at all.

Our Teku VC logic sends batches in sequence so it is not causing this issue, but if there are multiple VCs connected to the same teku BN (or a VC sending batches in parallel) we may end up with multiple concurrent batches.

Since we we are not throttling critical calls anymore, we can reduce the number of concurrent non-critical calls to 3 (in practice, we won't send more than 3 registrations concurrently)

It won't affect status calls. We anyway relay on actual status API call timeout\response to decide if builder is alive or not, so if we delay it in the throttler it won't cause any bad effect.

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,5 +11,6 @@
 ### Additions and Improvements
 - Enabling, by default, a new attestation pool implementation that improves the attestation packing during block and aggregation production. It can still be disabled by setting `--Xaggregating-attestation-pool-v2-enabled=false` if needed
 - Added `--p2p-discovery-bootnodes-url` CLI option.
+- Avoid builder validator registration calls potentially delaying block production builder calls.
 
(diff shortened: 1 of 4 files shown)


StefanBratanov (approved):
LGTM, makes total sense, on top of that we only do one block building per slot in the BN, so these methods don't need throttling. I wonder how useful that whole throttling class is, but probably somewhat useful when you have many VCs connected to that BN

tbenr:
> I wonder how useful that whole throttling class is

I think it is when you have several VCs sending validator registration batches at the same time.

Issue #9463, opened 2025-05-22
Fix BeaconChainController initialization loop Network <-> DAS Custody


lucassaldanha:
Check TODO-fulu on `BeaconChainController` // TODO-fulu fix this hack to resolve the initialization loop Network <--> DAS Custody

zilm13:
The target is to eliminate `LateInitDataColumnSidecarCustody` and `CustodyGroupCountManagerLateInit`, right?

PR #9781, opened 2025-08-19
remove custody late initialisation

mehdi-aouadi:
## PR Description
Remove the `CustodyGroupCountManager` and `DataColumnSidecarCustody` late initialisation classes and fix the circular dependencies by using suppliers that resolve at last minute when used 

## Fixed Issue(s)

#9463 

## Documentation

## Changelog


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -1606,8 +1600,8 @@ protected void initP2PNetwork() {
             .eventChannels(eventChannels)
             .combinedChainDataClient(
                 throttlingCombinedChainDataClient.orElse(combinedChainDataClient))
-            .dataColumnSidecarCustody(dataColumnSidecarCustody)
-            .custodyGroupCountManager(custodyGroupCountManagerLateInit)
+            .dataColumnSidecarCustody(dataColumnSidecarCustodySupplier)
```
zilm13: let it call `dataColumnSidecarCustodySupplier` too

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -165,6 +161,18 @@ public void onNewBlock(final SignedBeaconBlock block, final Optional<RemoteOrigi
     createOrUpdateRecoveryTaskForBlock(block.getMessage());
   }
 
+  private boolean shouldSkipProcessing() {
```
zilm13: I'd add slot parameter here and put whole rule including `!isActiveSuperNode(`  inside

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -165,6 +161,18 @@ public void onNewBlock(final SignedBeaconBlock block, final Optional<RemoteOrigi
...
+    if (custodyGroupCountManagerSupplier.get().getCustodyGroupSyncedCount() == groupCount) {
```
zilm13: Looking at this code now, I think it should be `getCustodyGroupCount`, why wait for it to be synced?

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -853,12 +853,8 @@ protected void initDasCustody() {
             },
             metricsSystem,
             timeProvider);
+    this.dataColumnSidecarCustodySupplier = () -> dataColumnSidecarRecoveringCustody;
```
zilm13: So you change the object this var was linked to. Reference you already used will be still linked with old Supplier.
If you want to go this way, you need to have single wrapper object and change the object wrapped. For example AtomicReference will work for this or mutable collections like List. Of course, using list for this would be weird, but let's try with AtomicReference to see how it will look like

zilm13 (commented):
I've described why it doesn't work but let's try with AtomicReference.
Also I suggest to do in 1 commit or in other PR a version where we pass reference from `BeaconChainController` inside other classes only in those constructors that are initialized earlier and unwrap it and pass real object for all constructors initialized later. It will require thinking in advance on the order, but it something done wrong, `DasSyncAcceptanceTest` will fail. As an advantage, this reference will not spread over 5 classes. I think it would be good to see both versions to decide

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
mehdi-aouadi: As suggested by @tbenr using a `Supplier` and a method reference works too but imo using `AtomicReference` is more suitable for this use case since the goal is simply to replace the NOOP by a real implementation

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
mehdi-aouadi: Done but I had to update the test. We'd need to make sure that's ok

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -351,10 +351,11 @@ public class BeaconChainController extends Service implements BeaconChainControl
   protected volatile KZG kzg;
   protected volatile BlobSidecarManager blobSidecarManager;
   protected volatile BlobSidecarGossipValidator blobSidecarValidator;
-  protected volatile CustodyGroupCountManagerLateInit custodyGroupCountManagerLateInit;
+  protected volatile AtomicReference<CustodyGroupCountManager> custodyGroupCountManagerRef =
```
zilm13: volatile not needed for AtomicReference

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -351,10 +351,11 @@ public class BeaconChainController extends Service implements BeaconChainControl
...
+      new AtomicReference<>(CustodyGroupCountManager.NOOP);
   protected volatile DataColumnSidecarManager dataColumnSidecarManager;
-  protected volatile LateInitDataColumnSidecarCustody dataColumnSidecarCustody =
-      new LateInitDataColumnSidecarCustody();
+  protected volatile AtomicReference<DataColumnSidecarRecoveringCustody>
```
zilm13: volatile not needed for AtomicReference

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -93,11 +94,12 @@ public DataColumnSidecarRecoveringCustodyImpl(
...
     this.spec = spec;
     this.dataColumnSidecarPublisher = dataColumnSidecarPublisher;
-    this.custodyGroupCountManager = custodyGroupCountManager;
+    this.custodyGroupCountManagerSupplier = custodyGroupCountManagerSupplier;
     this.recoveryTasks =
         LimitedMap.createSynchronizedNatural(spec.getGenesisSpec().getSlotsPerEpoch());
     this.isSuperNode =
-        new AtomicBoolean(custodyGroupCountManager.getCustodyGroupCount() == groupCount);
+        new AtomicBoolean(
+            custodyGroupCountManagerSupplier.get().getCustodyGroupCount() == groupCount);
```
zilm13: So we use it right in the constructor, so it's a bit confusing as we passed supplier to delay init.
I think we'd better set it to false on var declaring string, we will override it on first usage anyway in `shouldSkipProcessing`

zilm13 (approved):
LGTM, just few nits 
Looks better than late init I think

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
mehdi-aouadi: Make sense. It will be updated at the first `onSlot` or `onBlock` and I guess that's fine.

PR #9785, opened 2025-08-20
Updated the voluntary exit validation to allow a slow retry

rolfyone:
Updated the voluntary exit validator to timeout any exit's it's seen so that they might be retried without a restart.

The result would still be re-submitting gives a 400, but the node will retry periodically if the exit is not seen in a block.

## Documentation

## Changelog


data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
```diff
@@ -158,7 +162,37 @@ public List<SignedVoluntaryExit> getVoluntaryExits() {
...
+                          }
+                          return InternalValidationResult.ACCEPT;
+                        })
+                    .thenApply(
+                        result -> {
+                          if (result.isAccept()) {
+                            return voluntaryExitPool.addLocal(exit).join();
+                          }
+                          return result;
+                        }))
```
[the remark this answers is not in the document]
lucassaldanha: I don't think this applies to this case. Can safely ignore.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/VoluntaryExitValidator.java
```diff
@@ -33,21 +34,26 @@
 
 public class VoluntaryExitValidator implements OperationValidator<SignedVoluntaryExit> {
 
+  // 117 minutes as millis
```
lucassaldanha: NIT: we could explain why we are using 177 (just below 2h that is the time the MappedOperationPool retries, etc.)

lucassaldanha (approved):
LGTM. Just need to fix the off by one reported by cursor :)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/VoluntaryExitValidator.java
rolfyone: yep sure, and 117 :)

data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
```diff
@@ -158,7 +165,46 @@ public List<SignedVoluntaryExit> getVoluntaryExits() {
...
+                                exit.getValidatorId());
+                          }
+                          return InternalValidationResult.ACCEPT;
+                        })
+                    .thenApply(
+                        result -> {
+                          if (result.isAccept()) {
+                            try {
+                              // if we can't add this in a reasonable time we should fail.
+                              return voluntaryExitPool.addLocal(exit).get(5, TimeUnit.SECONDS);
```
rolfyone, line 196: @lucassaldanha rather than join, timebox?

lucassaldanha: Sure, that makes sense. There is no sane reason for it taking longer than that.

data/provider/src/main/java/tech/pegasys/teku/api/NodeDataProvider.java
```diff
@@ -158,7 +165,46 @@ public List<SignedVoluntaryExit> getVoluntaryExits() {
...
+                            } catch (InterruptedException
+                                | ExecutionException
+                                | TimeoutException e) {
+                              return InternalValidationResult.reject(
+                                  "Failed to add voluntary exit for validator index %s to pool: %s",
+                                  exit.getValidatorId(), e.getMessage());
+                            }
+                          }
```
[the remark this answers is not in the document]
rolfyone: this is at least a limited wait, and its reasonable, so dismissing this one.

PR #9787, opened 2025-08-20
Early prepare block production

tbenr:
This is a quick approach to anticipate by 500ms the block production.

Which means we will:
- run `prepareForBlockProduction`
   - advance the time to the beginning of the slot
   - apply deferred attestations
   - process head
     - run late block reorg logic
     - eventually run reorg logic to prepare pools accordingly (upcoming PR)
- retrieve the corresponding state we will use to build the block
   
When VC calls for block production we already have the state to continue the block production (or at least we have the process started and the flow simply attaches to it)

This is faking a bit but we were already tolerating 500ms if beacon API were already called that early.

## Documentation

## Changelog


beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/BlockProductionPreparationTrigger.java
```diff
@@ -0,0 +1,65 @@
...
+    this.asyncRunner = asyncRunner;
+  }
+
+  public void onBlockProductionPreparationDue(final UInt64 slot) {
+    asyncRunner
+        .runAsync(
+            () -> {
+              final UInt64 productionSlot = slot.increment();
+              recentChainData
+                  .getBestState()
```
tbenr: this is not a good way of picking the state at epoch boundary.
I need to make sure I pick the state produced by the EpochCachePrimer

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -189,6 +197,12 @@ public ValidatorApiHandler(
     this.attesterDutiesGenerator = new AttesterDutiesGenerator(spec);
   }
 
+  @Override
+  public void onSlot(final UInt64 slot) {
+    blockProductionStateCache.keySet().removeIf(cachedSlot -> cachedSlot.isLessThan(slot));
+    localBlockProductionBySlotCache.keySet().removeIf(cachedSlot -> cachedSlot.isLessThan(slot));
+  }
```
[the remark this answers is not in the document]
tbenr: I don't think there is a resource leak here.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandler.java
```diff
@@ -330,11 +341,38 @@ public SafeFuture<Optional<BlockContainerAndMetaData>> createUnsignedBlock(
...
+              forkChoiceTrigger
+                  .prepareForBlockProduction(keySlot, productionPerformance)
+                  .thenCompose(
+                      ignored ->
+                          combinedChainDataClient.getStateForBlockProduction(
+                              keySlot, forkChoiceTrigger.isForkChoiceOverrideLateBlockEnabled()))
+                  .thenPeek(__ -> productionPerformance.getStateAtSlot());
+
+          return new BlockProductionPreparationContext(state, productionPerformance);
+        });
```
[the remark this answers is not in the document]
tbenr: not a problem to me

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/SlotProcessor.java
```diff
@@ -238,14 +255,27 @@ boolean isEpochPrecalculationDue(
...
+  }
+
+  boolean isBlockProductionPreparationDue(
+      final UInt64 calculatedSlot, final UInt64 currentTimeMillis, final UInt64 genesisTimeMillis) {
+    if (!isProcessingDueForSlot(calculatedSlot, onTickBlockProductionPreparation)) {
+      return false;
+    }
+
+    final UInt64 earliestTimeInMillis =
+        spec.computeTimeMillisAtSlot(calculatedSlot.increment(), genesisTimeMillis)
```
mehdi-aouadi, line 276: I was a bit concerned about the "hacky" way of advancing the time but I think it's ok. Maybe we could rename `isBlockProductionPreparationDue` to make it clear it's an action for the future, something like `isFutureBlockProductionPreparationDue`

mehdi-aouadi (approved):
LGTM
Just a nit

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/SlotProcessor.java
tbenr: I was worried too, at first. But then I did some checks around and I lended on being more confident.

PR #9811, opened 2025-08-26
Remove constructor's cgc set for Metadata message

### LINKED ISSUE #9461: Check Eth2PeerManager logic setting custodyGroupCount
When we update our custodyGroupCount on MetadataMessageFactory we do not have guarantee that we are sync'd. We need to review that assumption.

Search for Eth2PeerManager // TODO-fulu: we have no guarantee here that it's synced already

zilm13:
## PR Description
So we set cgc field in Metadata message:
- on initialization from genesis via channel event
- on any finalization via channel event

Which is enough to keep it uptodate and have correct value right after the client startup. 
Comment is valid and we couldn't know all required info here in constructor, so as far as it will arrive via channel a moment later, we could remove this initial value assigning

## Fixed Issue(s)

Closes #9461 

## Documentation

## Changelog


zilm13:
I'm not happy with it.
It takes sometimes 4 epochs to jump from 0 to actual value, which is bad.
Reworking

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/ActiveEth2P2PNetwork.java
```diff
@@ -174,8 +171,8 @@ private synchronized void startGossip() {
             discoveryNetwork::setSyncCommitteeSubnetSubscriptions);
     final UInt64 currentEpoch = recentChainData.getCurrentEpoch().orElseThrow();
     if (spec.isMilestoneSupported(SpecMilestone.FULU)) {
-      LOG.info("Using custody sidecar subnets count: {}", dasTotalCustodySubnetCount);
-      discoveryNetwork.setDASTotalCustodySubnetCount(dasTotalCustodySubnetCount);
+      // acttual value is updated on channel event
```
rolfyone: `actual`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerManager.java
```diff
@@ -134,12 +135,11 @@ public static Eth2PeerManager create(
...
       final KZG kzg,
       final DiscoveryNodeIdExtractor discoveryNodeIdExtractor,
-      final Optional<UInt64> custodyGroupCount,
       final DasReqRespLogger dasLogger) {
 
-    // TODO-fulu: we have no guarantee here that it's synced already
-    // (https://github.com/Consensys/teku/issues/9461)
-    custodyGroupCount.ifPresent(metadataMessagesFactory::updateCustodyGroupCount);
+    if (spec.isMilestoneSupported(SpecMilestone.FULU)) {
+      metadataMessagesFactory.updateCustodyGroupCount(UInt64.ZERO);
```
rolfyone: update it to 0? 0 is never valid to my understanding...

networking/eth2/src/testFixtures/java/tech/pegasys/teku/networking/eth2/Eth2P2PNetworkFactory.java
```diff
@@ -265,9 +261,13 @@ protected Eth2P2PNetwork buildNetwork(final P2PConfig config) {
                 spec,
                 NoOpKZG.INSTANCE,
                 __ -> Optional.of(discoveryNodeId),
-                dasTotalCustodySubnetCount,
                 DasReqRespLogger.NOOP);
 
+        if (spec.isMilestoneSupported(FULU)) {
+          metadataMessagesFactory.onCustodyGroupCountSynced(
+              SpecConfigFulu.required(spec.forMilestone(FULU).getConfig()).getCustodyRequirement());
```
rolfyone: CGC count should probably come from manager now that it's persisted?

zilm13: it's test fixture

zilm13:
As far as we've added `earliest_block_slot` in `Metadata` I think we could safely keep everything as is and remove this TODO

lucassaldanha (approved):

PR #9823, opened 2025-08-28
Bring back `getBeaconProposerIndex` validation for Fulu

StefanBratanov:
## PR Description
`validateStateCanCalculateProposerIndexAtSlot` was used before to check if requested slot is from the same epoch as the state and have a nice error message. In the current implementation of `getBeaconProposerIndex` as per spec it only works for the same epoch as the state, so it makes sense to bring back that validation.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Upcoming Breaking Changes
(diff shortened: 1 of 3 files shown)


tbenr (approved):
LGTM

PR #9825, opened 2025-08-29
fix: guard against null Accept header in ResourceLoader.urlOrFile

sashass1315:
## PR Description

Replace Optional.of(acceptHeader) with Optional.ofNullable(acceptHeader) to avoid NPE when callers pass a null Accept header. Aligns with URLResourceLoader which treats absent Accept header as optional; current call sites pass literals, so this change is backward-compatible.

## Documentation

## Changelog


sashass1315:

rolfyone:
Thanks for raising. 

did we observe this behaviour somewhere?

sashass1315:
> Thanks for raising.
> 
> did we observe this behaviour somewhere?

no we didn't

rolfyone:
So my issue here is that if we do pass in a null string, then we're going to get empty, and the problem will be 'less visible' than if they just got the nullable here... really if you call that fn with a null string then it's a bug...

We could put in a CheckNotNull in that function, or accept the NPE, but i think `ofNullable` is only going to hide a bug in this instance...

sashass1315:
> So my issue here is that if we do pass in a null string, then we're going to get empty, and the problem will be 'less visible' than if they just got the nullable here... really if you call that fn with a null string then it's a bug...
> 
> We could put in a CheckNotNull in that function, or accept the NPE, but i think `ofNullable` is only going to hide a bug in this instance...

updated

rolfyone (approved):
LGTM

Issue #9826, opened 2025-08-29
[fusaka devnet-3] Node falls into syncing mode with errors


tbenr:
(redacted logs)
[fusaka-devnet3-teku-reth.log](https://github.com/user-attachments/files/22042942/fusaka-devnet3-teku-reth.log)

To my initial analysis

- the node starts when its head is close to the chain tip.
- the EL is syncing behind, the node is able to keep following the chain optimistically but errors like this printed:
`2025-08-28 09:39:31.476 ERROR - Number of sidecars must be greater than or equal to the half of column count, slot: 257747; columns: found 52, needed at least 64`
- We hit some `rejected execution` on query storage
- It seems we handled the rejection any way
- but at a certain point we stop importing blocks probably due to `DATA_NOT_AVAILABLE`
`2025-08-28 12:02:02.533 WARN  - Unable to import blocks: DATA_NOT_AVAILABLE`
- from that moment we can't follow the chain any more.
- we transition from optimistic to real sync but our head remains at `Head slot: 258451`

tbenr:
<img width="470" height="381" alt="Image" src="https://github.com/user-attachments/assets/ca8897b4-b606-474a-95a7-1ca7543547e7" />

<img width="461" height="372" alt="Image" src="https://github.com/user-attachments/assets/aa28eb03-c03d-42b6-90e0-e6cd51c49a1f" />

<img width="473" height="757" alt="Image" src="https://github.com/user-attachments/assets/de65c173-2436-47d8-8c14-923bdc0d5c4e" />

```
2025-09-02 13:31:47.232 DEBUG - ReqResp inbound data_column_sidecars_by_root, columns: 0/1 in 70267 ms, peer 16Uiu2HAmVr2hvbo7go5FKv4buupVQKXqmPMq43KixtA9K2bbVTmc (nodeId = 0x0a7c702d...5f6cd9bd): request: 9 columns: 0x18026ce7eb8ae4cc77d2bb18e11b931cb68e52438bb5a796cbc07648a67ca9a1 colIdxs: (len: 9) [4,14,21,65,85,107,111,114,127], response: <empty>
[...]

PR #9828, opened 2025-08-29
Limit rpc storage query queue size

tbenr:
We recently introduced throttling in storage query channel used by RPC.

The problem is that we can receive a very high number of requests and we could be overloaded anyway:

image

We get this number of columns requested PER SECOND:
image

This pr introduces a limit in the queue size.

related to #9826

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandler.java
```diff
@@ -254,6 +255,14 @@ private SafeFuture<Optional<DataColumnSidecar>> retrieveDataColumnSidecar(
     return dataColumnSidecarCustodySupplier
         .get()
         .getCustodyDataColumnSidecarByRoot(identifier)
+        .exceptionally(
+            error -> {
+              LOG.warn(
+                  "Failed to retrieve data column sidecar for identifier {}: {}",
+                  identifier,
+                  ExceptionUtil.getMessageOrSimpleName(error));
+              return Optional.empty();
```
tbenr: the warning here is debatable, but I think returning partial result seems reasonable.
I if we fail everything we get downscored anyway?

zilm13: From the spec:
Clients MUST respond with at least one sidecar, if they have it. Clients MAY limit the number of blocks and sidecars in the response.

zilm13: also we don't have similar logging in byRange

zilm13: this warning will spam without throttling very badly, I'd remove it

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandler.java
```diff
@@ -254,6 +255,14 @@ private SafeFuture<Optional<DataColumnSidecar>> retrieveDataColumnSidecar(
...
+            })
         .thenCompose(
```
zilm13: We will get here and try to request non-canonical sidecar after failed future, which I think is bad, because we try to abuse db again. You could test it  in `DataColumnSidecarsByRootMessageHandlerTest`. Also I recommend to add test there for this case. Also I guess we will try this with all subsequent identifiers in this request, instead it should be good to cut response after first throttling failure

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandler.java
rolfyone: to me it does sound like a debug maybe

rolfyone: or actually maybe even a trace...

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandler.java
tbenr: that's a very good point. I missed that effect.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRangeMessageHandler.java
```diff
@@ -82,7 +82,7 @@ public DataColumnSidecarsByRangeMessageHandler(
     totalDataColumnSidecarsRequestedCounter =
         metricsSystem.createCounter(
             TekuMetricCategory.NETWORK,
-            "rpc_data_column_sidecars_by_range_requested_sidecars_total",
+            "rpc_data_column_sidecars_by_range_requested_data_column_sidecars_total",
```
rolfyone: we've got data columns in twice and the name is crazy long, can we drop the second one so its just `rpc_data_column_sidecars_by_range_requested_total`

rolfyone: or even drop requested too, `rpc_data_column_sidecars_by_range_total`

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/LimitedThrottlingTaskQueue.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package tech.pegasys.teku.infrastructure.async;
+
+import java.util.concurrent.RejectedExecutionException;
+import java.util.function.Supplier;
+import org.hyperledger.besu.plugin.services.MetricsSystem;
+import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
+
+public class LimitedThrottlingTaskQueue implements TaskQueue {
```
rolfyone: ok so we had throttling, now we have throttling and limited.... they're both semantically very similar in english so i think we'd want to understand why use limitedThrottling over throttling etc - the names really dont help us understand when to use one over the other.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRangeMessageHandler.java
tbenr: I agree that these names (all other RPCs too) have very strange (and long) names.
Here i just made it consistent with the other by root.
we can discuss how make it better (considering impacts on the dashboard, which by the way requires an update in any case)

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/LimitedThrottlingTaskQueue.java
tbenr: any suggestion? :)

rolfyone (approved):
LGTM After discussion

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRangeMessageHandler.java
tbenr: reverting this

PR #9834, opened 2025-09-01
Remove compileReferenceTestJava dependency on generateReferenceTestClasses

lucassaldanha:
## PR Description

This task dependency has the side effect of completely deleting all generated reference test classes and regenerating them every time referenceTest modules compile. This is completely unnecessary, as we only need to regenerate classes when there is a new version of reference tests.

Given that whenever we are running a reference test via gradle we compile the project, it is always taking several minutes. This should bring it back to a few seconds.

In our CI environment, we have a explicit call to `generateReferenceTestClasses` task during the `ReferenceTests` step, there is no need real need for this dependency on compile.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/build.gradle
+++ b/eth-reference-tests/build.gradle
@@ -46,7 +46,6 @@ task generateReferenceTestClasses(type: JavaExec) {
 }
 
 compileReferenceTestJava {
-  dependsOn generateReferenceTestClasses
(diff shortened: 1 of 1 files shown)


lucassaldanha:
The cursor bot comment isn't relevant in this case. Our logic in CI already handles the manual step of generating test classes. We don't want to do that on every build.

Issue #9829, opened 2025-08-31
ZKEVM L1 prototype


gfukushima:
Prototype to allow executing Teku simulating a scenario where ZKEVM are being used and execution block don't require a EL to be validated (validated via proofs).
More details: https://hackmd.io/@kevaundray/Bypupr9Yge

kevaundray:
For future reference:

- https://github.com/Consensys/teku/pull/9836
- https://github.com/Consensys/teku/pull/9853
- https://github.com/Consensys/teku/pull/9867
- https://github.com/Consensys/teku/pull/9869
- https://github.com/Consensys/teku/pull/9909
- https://github.com/Consensys/teku/pull/10002
- https://github.com/Consensys/teku/pull/10004
- https://github.com/Consensys/teku/pull/10097

PR #9836, opened 2025-09-01
[Zkevm Proto] cli options

gfukushima:
## PR Description
This PR adds some CLI options for the ZKEVM prototype. Worth saying that for now I'm using an approach similar to what we have for the pow chain, so I'm adding the features if this implementation of something I'm naming temporarily zkchain.

## Fixed Issue(s)

Part of #9829 

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProof.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public static class ExecutionProofSchema
+      extends ContainerSchema4<
+          ExecutionProof, SszBytes32, SszUInt64, SszUInt64, SszVector<SszByte>> {
+    public ExecutionProofSchema() {
+      super(
+          "ExecutionProof",
+          namedSchema("block_hash", SszPrimitiveSchemas.BYTES32_SCHEMA),
+          namedSchema("subnet_id", SszPrimitiveSchemas.UINT64_SCHEMA),
+          namedSchema("version", SszPrimitiveSchemas.UINT64_SCHEMA),
```
lucassaldanha: On LH, the version type is an unsigned 32-bit integer. We need to confirm what type will be used and ensure we use the same.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProof.java
```diff
@@ -0,0 +1,81 @@
...
+          namedSchema(
+              "proof_data",
+              SszVectorSchema.create(
+                  SszPrimitiveSchemas.BYTE_SCHEMA,
+                  65536)) // Assuming max size of proof_data is 65536 bytes
```
lucassaldanha: Let's add this to a constant, something like `PROOF_DATA_MAX_SIZE` or similar.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProof.java
```diff
@@ -0,0 +1,81 @@
...
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszBytes32;
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszUInt64;
+import tech.pegasys.teku.infrastructure.ssz.schema.SszPrimitiveSchemas;
+import tech.pegasys.teku.infrastructure.ssz.schema.SszVectorSchema;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNode;
+
+public class ExecutionProof
+    extends Container4<ExecutionProof, SszBytes32, SszUInt64, SszUInt64, SszVector<SszByte>> {
+
+  public static class ExecutionProofSchema
```
lucassaldanha: NIT: Recently, we have been moving away from using the static schema internal class. It is not a big deal, but I'd rather have it in its own class file.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProof.java
```diff
@@ -0,0 +1,81 @@
...
+          );
+    }
+
+    @Override
+    public ExecutionProof createFromBackingNode(final TreeNode node) {
+      return new ExecutionProof(this, node);
+    }
+  }
+
+  public static final ExecutionProofSchema SSZ_SCHEMA = new ExecutionProofSchema();
```
lucassaldanha: I don't think we need this public static reference to the schema. It is only used internally on this class, and any other references to the schema should be retrived from the Schema Registry.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProof.java
```diff
@@ -0,0 +1,81 @@
...
+import tech.pegasys.teku.infrastructure.ssz.containers.Container4;
+import tech.pegasys.teku.infrastructure.ssz.containers.ContainerSchema4;
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszByte;
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszBytes32;
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszUInt64;
+import tech.pegasys.teku.infrastructure.ssz.schema.SszPrimitiveSchemas;
+import tech.pegasys.teku.infrastructure.ssz.schema.SszVectorSchema;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNode;
+
+public class ExecutionProof
```
lucassaldanha: Probably worth adding a property test for this container as we do for all containers.

teku/src/main/java/tech/pegasys/teku/cli/options/ZkChainOptions.java
```diff
@@ -0,0 +1,59 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.cli.options;
+
+import picocli.CommandLine;
+import tech.pegasys.teku.config.TekuConfiguration;
+import tech.pegasys.teku.services.zkchain.ZkChainConfiguration;
+
+public class ZkChainOptions {
```
lucassaldanha, line 20: We missed the unit tests for this options object. For example, testing that both boolean parameters default to false, that we can't have negative values on the integer parameter, etc. Otherwise, anyone could come here and change it by mistake, and we won't catch it.

gfukushima: added them

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProof.java
gfukushima: going to add the container and it's tests on a separate PR once I clarify the u32 field with Kev, have removed this from this PR for now

teku/src/test/java/tech/pegasys/teku/cli/options/ZkChainOptionsTest.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Test
+  public void statelessMinProofsRequired_receivesDefaultValue() {
+    final TekuConfiguration config = getTekuConfigurationFromArguments();
+    assertThat(config.zkChainConfiguration().getStatelessMinProofsRequired()).isEqualTo(1);
+  }
+
+  @Test
+  public void statelessValidationEnabled_receivesDefaultValue() {
```
lucassaldanha: I'd rename this method to `statelessValidationEnabled_isDisabledByDefault`

teku/src/test/java/tech/pegasys/teku/cli/options/ZkChainOptionsTest.java
```diff
@@ -0,0 +1,62 @@
...
+    final TekuConfiguration config = getTekuConfigurationFromArguments();
+    assertThat(config.zkChainConfiguration().isStatelessValidationEnabled()).isFalse();
+  }
+
+  @Test
+  public void generateExecutionProofsEnabled_receivesDefaultValue() {
```
lucassaldanha: I'd rename this method to `generateExecutionProofsEnabled_isDisabledByDefault`

lucassaldanha (approved):
LGTM. Just added a suggestion for unit test names.

PR #9840, opened 2025-09-03
Enforce 32-byte JWT secret length when loading EE JWT secret

VolodymyrBg:
- Validate that the loaded JWT secret is exactly 32 bytes (64 hex chars) per Engine API expectations.
- Throw InvalidConfigurationException with a clear error message if the length is incorrect.


tbenr:
To me this looks like an AI

PR #9841, opened 2025-09-03
Be more conservative in RPC peer rate limiting

tbenr:
- removed the queue-full exception handling, a complexity introduced to improve a situation that was most likely caused by a bug
- Avoid to adjust objects in case of errors.
- reduces default storage query limits

related to #9826

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandler.java
```diff
@@ -265,7 +270,12 @@ private SafeFuture<Optional<DataColumnSidecar>> retrieveDataColumnSidecar(
               // Fallback to non-canonical sidecar if the canonical one is not found
               return getNonCanonicalDataColumnSidecar(identifier);
             })
-        .exceptionally(ThrottlingStorageQueryChannel::ignoreQueueIsFullException);
+        .exceptionally(
+            error -> {
+              ThrottlingStorageQueryChannel.mapQueueIsFullExceptionToEmptyOrRethrow(error);
```
zilm13: I'm not sure it's a good pattern, maybe make it more traditional, if (smth) else?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2Peer.java
```diff
@@ -149,6 +151,8 @@ Optional<RequestKey> approveDataColumnSidecarsRequest(
 
   boolean approveRequest();
 
+  void recordStorageLimitHit();
```
zilm13: Not sure we should disclose reason on this layer. I'd use something more general like `pauseAllRequests()` or smth like that

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerFactory.java
```diff
@@ -85,6 +85,7 @@ public Eth2Peer create(final Peer peer, final BeaconChainMethods rpcMethods) {
             timeProvider,
             "dataColumns"),
         RateTracker.create(peerRequestLimit, TIME_OUT, timeProvider, "requestTracker"),
+        RateTracker.create(1, TIME_OUT, timeProvider, "storageLimitHitTracker"),
```
zilm13: We could consider longer timeouts here in the future, but 1 minute is probably ok for the start and tests

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java
```diff
@@ -55,9 +55,9 @@ public class P2PConfig {
   // DEFAULT_BACKGROUND_THREAD_COUNT)
   // The storage query channel allows up to 10 parallel queries (STORAGE_QUERY_CHANNEL_PARALLELISM)
   // To avoid resource saturation and ensure capacity for other tasks, we limit historical data
-  // queries to 5
-  public static final int DEFAULT_HISTORICAL_DATA_MAX_CONCURRENT_QUERIES = 5;
-  public static final int DEFAULT_HISTORICAL_MAX_QUERY_QUEUE_SIZE = 100_000;
+  // queries to 3
+  public static final int DEFAULT_HISTORICAL_DATA_MAX_CONCURRENT_QUERIES = 3;
+  public static final int DEFAULT_HISTORICAL_MAX_QUERY_QUEUE_SIZE = 5_000;
```
zilm13: I'd test it somewhere. I see sepolia-archive is the most RPC active one currently

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandlerTest.java
```diff
@@ -492,6 +499,49 @@ public void shouldSendToPeerRequestedBlobSidecars() {
         .containsExactlyElementsOf(expectedDataColumnIdentifiersBlockRoots);
   }
 
+  @TestTemplate
+  public void shouldNotAdjustIfAnErrorOccurs() {
```
zilm13, line 437: I'd add a comment why we made this change, that it's protective

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandlerTest.java
```diff
@@ -492,6 +499,49 @@ public void shouldSendToPeerRequestedBlobSidecars() {
...
+    handler.onIncomingMessage(
+        protocolId, peer, messageSchema.of(dataColumnsByRootIdentifiers), callback);
+
+    verify(callback)
+        .completeWithUnexpectedError(argThat(exception -> exception.getCause().equals(error)));
+    verify(peer, never()).adjustDataColumnSidecarsRequest(any(), anyLong());
+  }
+
+  @TestTemplate
+  public void shouldRateLimitOnRequest() {
```
zilm13: We have similar `shouldNotSendDataColumnSidecarsIfPeerIsRateLimited`, I'd name them to clarify that one fails on approveSidecars, and another on approveRequest and put one after another (extremely nit)

zilm13 (approved):
LGTM just few nits
Overall I think it's a good defensive improvement and we should include it. Some numbers are magic and we may start with less risky values and adjust it after some testing.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2Peer.java
tbenr: I see. Worth noting that current implementation actually disconnects the peer, no just drop the request.
And currently I have a doubt: if we call `disconnectCleanly` when we hit some limit, then the peer reconnects here:
```
  public void onConnect(final Peer peer) {
    Eth2Peer eth2Peer = eth2PeerFactory.create(peer, rpcMethods);
```
and then we create a new `DefaultEth2Peer` which will create new brand trackers ready to be filled up again.

If I am right, that's not good to me.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java
tbenr: everything is flat to 0 there.
image

All mainnets are flat at 0 too but Holesky boxes are interesting. Going to deploy this there.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2Peer.java
tbenr: actually the ReputationManager kicks in.
With that in mind i think this PR can be simplified, there is no need to add a new tracker

tbenr:
@zilm13 I decided to reduce complexity. We can reintroduce it if we see it will be worthy after the throttling bug has been fixed.

- I applied the concept of being defensive on RPC handling errors.
- I reduced the default throttling params to values that seem reasonable based on what I'm seeing on devnets\testnets.

the problem of low-reputation peers being able to reconnect and start again is being addressed here #9887

zilm13 (approved):
LGTM

PR #9846, opened 2025-09-04
Fix deadlock and another bug in TaskQueue

tbenr:
- Fixes a deadlock that can happen between threads acquiring different locks. The issue is solved by having only one lock, on the delegated class (`ThrottlingTaskQueue`)

- Fixes a bug where if the future supplier throws, we remain with an inflight request that never completes
## Documentation

## Changelog

### DIFF
--- a/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/LimitedTaskQueue.java
+++ b/infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/LimitedTaskQueue.java
@@ -13,6 +13,7 @@
 
 package tech.pegasys.teku.infrastructure.async;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.function.Supplier;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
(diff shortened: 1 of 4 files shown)


infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/LimitedTaskQueue.java
```diff
@@ -61,21 +62,24 @@ private LimitedTaskQueue(final TaskQueue delegate, final int maximumQueueSize) {
   }
 
   @Override
-  public synchronized <T> SafeFuture<T> queueTask(final Supplier<SafeFuture<T>> request) {
-    if (delegate.getQueuedTasksCount() >= maximumQueueSize) {
-      rejectedTaskCount++;
-      return SafeFuture.failedFuture(new QueueIsFullException());
+  public <T> SafeFuture<T> queueTask(final Supplier<SafeFuture<T>> request) {
+    synchronized (delegate) {
```
zilm13, line 66: I think we could drop even this synchronization. It's high pressure cutoff for RPC, we don't care about eventual consistency, two clients will not get their data now instead of one, not a big deal

zilm13 (approved):
LGTM

infrastructure/async/src/main/java/tech/pegasys/teku/infrastructure/async/LimitedTaskQueue.java
tbenr: going to merge this as is. I'd like to revisit the whole thing

PR #9849, opened 2025-09-04
Restart sync every 60 seconds

zilm13:
## PR Description
Currently we restart sync only on peer status changes which could be rare especially in testnets when we have fixed peer pool. Looking at logs in such situation it's usually takes 5 minutes for a peer until it tries to restart syncing after "Syncing completed" with many slots remaining. This PR addresses this issue.

## Documentation

## Changelog


beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncControllerTest.java
```diff
@@ -242,6 +247,33 @@ void shouldNotNotifySubscribersWhenRunningSpeculativeTarget() {
...
   }
 
+  @Test
+  void shouldRestartSyncEverySyncAwakePeriod() {
+    final SafeFuture<SyncResult> syncResult = startFinalizedSync();
+    assertThat(syncController.isSyncActive()).isTrue();
+    syncResult.complete(SyncResult.COMPLETE);
+    assertNotSyncing();
+
+    // shouldn't restart if there are no sync targets available
```
rolfyone: so basically we've called out test cases in comments, and not made them separate tests... i'd suggest both of these comments could give rise to their own test case.

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncController.java
```diff
@@ -61,6 +65,10 @@ public SyncController(
     this.recentChainData = recentChainData;
     this.syncTargetSelector = syncTargetSelector;
     this.sync = sync;
+    asyncRunner.runWithFixedDelay(
+        () -> eventThread.execute(this::onTargetChainsUpdated),
+        SYNC_AWAKE_INTERVAL,
+        ex -> LOG.error("Unexpected error when trying to awake peer sync", ex));
```
tbenr, line 73: I'd add a comment on why this is good to have

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/SyncControllerTest.java
tbenr: I think this is part of the preconditions of this test

tbenr (approved):
LGTM. I'd address comments if they make sense, but looks ok to me

PR #9851, opened 2025-09-04
Fix DataColumnsSidecars byRoot validation

zilm13:
## PR Description

Only up to 128 identifiers is allowed in request, see [spec](https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/p2p-interface.md?plain=1#L467-L473)
Old rule is dropped now because ` <=128 * <=128 is always <=16384`, we don't need to check chunks for this

## Documentation

## Changelog


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandler.java
```diff
@@ -116,15 +116,15 @@ private SafeFuture<Boolean> validateAndMaybeRespond(
...
+    final int maxRequestIdentifiers = specConfigFulu.getMaxRequestBlocksDeneb();
+    if (request.size() > maxRequestIdentifiers) {
       requestCounter.labels("count_too_big").inc();
       return Optional.of(
           new RpcException(
               INVALID_REQUEST_CODE,
               String.format(
-                  "Only a maximum of %s data column sidecars can be requested per request",
-                  maxRequestDataColumnSidecars)));
+                  "Only a maximum of %s by root identifiers can be requested per request",
```
rolfyone: i'd strike out `per request`, as 'requested per request' is not ideal.

rolfyone: also we've done %d in the tests correctly below but not here, should probably just use %d here too.

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRootMessageHandlerTest.java
```diff
@@ -167,16 +167,15 @@ public void setup(final TestSpecInvocationContextProvider.SpecContext specContex
   }
 
   @TestTemplate
-  public void validateRequest_shouldNotAllowRequestLargerThanMaximumAllowed() {
-    final int maxRequestDataColumnSidecars =
+  public void validateRequest_shouldNotAllowNumberOfIdentifiersLargerThanMaximumAllowed() {
```
rolfyone: `validateRequest_shouldLimitMaximumRequestSize`

rolfyone (approved):
nits. lgtm overall

PR #9853, opened 2025-09-05
[Stateless Proto] Execution proof container

gfukushima:
## PR Description
This PR adds the execution proof container and its schema along with its property test. Worth highlighting that I'm adding the new schema under electra although it's not part of the electra fork officially (due to the fact that our schema registry atm requires a fork to register a schema), this is suboptimal and the alternative would be adding a fork far in the future (let's say T* fork) to register the schema under. The idea of the prototype is to be run in parallel with real networks so using a future fork would probably defeat this purpose. We should move this to the right fork when the time comes

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/schemas/registry/SchemaTypes.java
```diff
@@ -193,6 +194,9 @@ public class SchemaTypes {
       create("CONSOLIDATION_REQUEST_SCHEMA");
   public static final SchemaId<SingleAttestationSchema> SINGLE_ATTESTATION_SCHEMA =
       create("SINGLE_ATTESTATION_SCHEMA");
+  // Move
```
rolfyone: move?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProofSchema.java
```diff
@@ -0,0 +1,61 @@
...
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszBytes32;
+import tech.pegasys.teku.infrastructure.ssz.primitive.SszUInt64;
+import tech.pegasys.teku.infrastructure.ssz.schema.SszPrimitiveSchemas;
+import tech.pegasys.teku.infrastructure.ssz.schema.SszVectorSchema;
+import tech.pegasys.teku.infrastructure.ssz.tree.TreeNode;
+
+public class ExecutionProofSchema
+    extends ContainerSchema4<ExecutionProof, SszBytes32, SszUInt64, SszUInt64, SszVector<SszByte>> {
+
+  // Assuming max size of proof_data is 65536 bytes, in reality this should be much less
```
rolfyone: iirc they were talking about 300k proofs, but not sure if this is the same thing...

rolfyone (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionProofSchema.java
kevaundray: We can set this value to 1MB to accommodate where most of the provers are at now -- yep, this will eventually be 300KB in the future (possibly lower)

kevaundray (approved):

PR #9854, opened 2025-09-05
Persisted custody group count

rolfyone:
Also fixed the `debug-tools db get-variables` command. It was missing a bunch of variables.

fixes #9845

## Documentation

- [X] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [X] I thought about adding a changelog entry, and added one if I deemed necessary.

PR #9862, opened 2025-09-08
Persisted custody group count to finalized store

### LINKED ISSUE #9845: Persist CGC change
from fulu spec:
```
If a node's custody requirements decrease, it SHOULD NOT update 
the custody_group_count to reflect this reduction. The node SHOULD 
continue to custody and advertise the previous (highest) custody_group_count. 
The node SHOULD continue to respond to any DataColumnSidecar 
request corresponding to the previous (highest) custody_group_count. 
The previous (highest) custody_group_count SHOULD persist across node restarts.
```

#9838 ensures CGC no longer goes down over time, but it remains ephemeral, not surviving restarts, and should be stored in a variable and loaded at startup.

The change that ensures it doesn't reduce is in `DataColumnSidecarCustodyImpl` 
`if (!totalCustodyGroupCount.compareAndSet(oldCustodyGroupCount, newCustodyGroupCount)) {`

it might be elsewhere. but we should also store it in DB if its not already stored.

Potentially it could be managed via `CustodyGroupCountManager` probably makes the most sense. the check in the current location would remain fine, the CGCManager could just control the data it ever sees. 

There appears to be a channel etc, we just need to make sure we're managing the data properly, and loading from storage at startup.

rolfyone:
one of the review comments in #9854 was that it'd be better as finalized store, so this PR is a subset of #9854.

 - Doesn't cover the main changes to CustodyGroupCountManagerImpl (was moved to #9860)
 - Doesn't cover the nit on Database (was moved to #9861)
 - Doesn't cover the fix for variables (was moved to #9859)

fixes #9845

## Documentation

## Changelog

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
@@ -769,4 +769,12 @@ public abstract class RecentChainData implements StoreUpdateHandler, ValidatorIs
   public void setBlockTimelinessIfEmpty(final SignedBeaconBlock block) {
     lateBlockReorgLogic.setBlockTimelinessFromArrivalTime(block, store.getTimeInMillis());
   }
+
+  public Optional<UInt64> getCustodyGroupCount() {
+    return store.getCustodyGroupCount();
+  }
+
+  public SafeFuture<Void> setCustodyGroupCount(final UInt64 custodyGroupCount) {
+    return storageUpdateChannel.onCustodyGroupCountUpdated(custodyGroupCount);
+  }
 }
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDao.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDao.java
@@ -84,6 +84,8 @@ public interface KvStoreCombinedDao extends AutoCloseable {
 
   Optional<Bytes32> getLatestCanonicalBlockRoot();
 
+  Optional<UInt64> getCustodyGroupCount();
+
   Optional<? extends SignedBeaconBlock> getNonCanonicalBlock(Bytes32 root);
 
   void ingest(KvStoreCombinedDao dao, int batchSize, Consumer<String> logger);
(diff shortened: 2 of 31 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDao.java
```diff
@@ -244,38 +246,40 @@ default void addHotStates(final Map<Bytes32, BeaconState> states) {
...
 
-    void addDepositsFromBlockEvent(final DepositsFromBlockEvent event);
+    void addDepositsFromBlockEvent(DepositsFromBlockEvent event);
 
     void removeDepositsFromBlockEvent(UInt64 blockNumber);
   }
 
   interface FinalizedUpdater extends AutoCloseable {
 
-    void addFinalizedBlock(final SignedBeaconBlock block);
```
rolfyone, line 256: cleaned up finals in here that were in interface

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
```diff
@@ -769,4 +769,12 @@ public void setBlockTimelinessFromArrivalTime(
   public void setBlockTimelinessIfEmpty(final SignedBeaconBlock block) {
     lateBlockReorgLogic.setBlockTimelinessFromArrivalTime(block, store.getTimeInMillis());
   }
+
+  public Optional<UInt64> getCustodyGroupCount() {
+    return store.getCustodyGroupCount();
+  }
+
+  public SafeFuture<Void> setCustodyGroupCount(final UInt64 custodyGroupCount) {
+    return storageUpdateChannel.onCustodyGroupCountUpdated(custodyGroupCount);
```
zilm13, line 778: I see that we were not using RecentChainData for setters before. Just an observation

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDao.java
```diff
@@ -244,38 +246,40 @@ default void addHotStates(final Map<Bytes32, BeaconState> states) {
...
+    void addFinalizedBlock(SignedBeaconBlock block);
 
     void addFinalizedBlockRaw(UInt64 slot, Bytes32 blockRoot, Bytes blockBytes);
 
-    void addNonCanonicalBlock(final SignedBeaconBlock block);
+    void addNonCanonicalBlock(SignedBeaconBlock block);
+
+    void deleteFinalizedBlock(UInt64 slot, final Bytes32 blockRoot);
```
zilm13, line 264: some more left

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/KvStoreCombinedDao.java
```diff
@@ -244,38 +246,40 @@ default void addHotStates(final Map<Bytes32, BeaconState> states) {
...
 
-    void deleteFinalizedBlock(final UInt64 slot, final Bytes32 blockRoot);
+    void deleteNonCanonicalBlockOnly(Bytes32 blockRoot);
 
-    void deleteNonCanonicalBlockOnly(final Bytes32 blockRoot);
+    void addFinalizedState(Bytes32 blockRoot, final BeaconState state);
```
zilm13, line 268: and here

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
```diff
@@ -487,6 +494,16 @@ public Checkpoint getFinalizedCheckpoint() {
     }
   }
 
+  @Override
+  public Optional<UInt64> getCustodyGroupCount() {
+    readLock.lock();
+    try {
+      return custodyGroupCount;
```
zilm13, line 501: So it's set only once on Store creation and never updated. I'm not sure it's correct

zilm13 (commented):
I expect something like this should pass as recentChainData test or combinedChainData test, but I'm not sure it will with the current code. It looks like major issue for me. Sorry if I'm wrong in understanding this code:
```java

  @Test
  public void getAndUpdateCgc() {
    final Spec fuluSpec = TestSpecFactory.createMinimalWithFuluForkEpoch(ONE);

    storageSystem =
            InMemoryStorageSystemBuilder.create()
                    .specProvider(fuluSpec)
                    .storageMode(StateStorageMode.PRUNE)
                    .storeConfig(StoreConfig.builder().build())
                    .build();
    chainBuilder = storageSystem.chainBuilder();
    recentChainData = storageSystem.recentChainData();
    genesis = chainBuilder.generateGenesis();

    genesisState = genesis.getState();
    genesisBlock = genesis.getBlock().getMessage();
    recentChainData.initializeFromGenesis(genesisState, UInt64.ZERO);

    recentChainData.setCustodyGroupCount(UInt64.valueOf(4)).join();
    assertThat(recentChainData.getCustodyGroupCount()).contains(UInt64.valueOf(4));
  }
  ```

storage/src/main/java/tech/pegasys/teku/storage/client/RecentChainData.java
rolfyone: becase you wanted it in finalized and it all changes...

storage/src/main/java/tech/pegasys/teku/storage/store/Store.java
rolfyone: this is kinda why i was trying to follow the normal pattern, but i'll take a look. i'd have really preferred leaving it where the rest of them were, as now we have to forge our own path.

rolfyone: im closing this in favour of following the pattern. we should have just stuck with what i did first time, its clear that was better.

PR #9863, opened 2025-09-08
fix: mock RecentChainData before constructing AggregatingAttestationPoolV2 in benchmark

MozirDmitriy:
Reordered initialization in AggregatingAttestationPoolBenchmark.init() to create the RecentChainData mock before constructing AggregatingAttestationPoolV2. Previously, the pool captured a null reference, which could cause NullPointerExceptions if code paths accessed recentChainData (e.g., retrieving state or current slot). While current data priming often avoids these paths, the setup was fragile and could fail under minor changes. This change ensures the pool always holds a valid mock, aligning with production wiring and tests.


rolfyone:
thanks for raising - sorry our new bot wants CLA signed but then should be good to go.

rolfyone (approved):
LGTM

MozirDmitriy:

MozirDmitriy:
> thanks for raising - sorry our new bot wants CLA signed but then should be good to go.

Signed, thanks for review

Issue #9833, opened 2025-09-01
Remove tests flag to ignore Gloas


lucassaldanha:
Some unit tests are not able to run until we have all Gloas data structures and state transitions implemented. This is a ticket to remind us to go back to them.

Unit Tests:
- [x] BlobSidecarSelectorFactoryTest.java
- [x] BlobSelectorFactoryTest.java
- [x] SignedBeaconBlockTest.java
- [x] VoluntaryExitValidatorTest.java
- [x] DataColumnSidecarsByRootMessageHandlerTest

Property Tests:
- [x] SignedBlockContentsSupplier (used in a few tests)
- [x] SignedBeaconBlockSupplier (used in a few tests)

Integration Tests:
- [x] GetNewBlockV3IntegrationTest
- [x] AbstractRpcMethodIntegrationTest (see methods `generateSpecTransitionWithCombinationParams()` and `generateSpecTransition()`
- [x] ProduceBlockRequestTest

Acceptance Tests:
- [ ] MergedGenesisInteropModeAcceptanceTest (required to assert on the execution payload, so until the whole flow is implemented, this can wait)

PR #9864, opened 2025-09-08
Fix property tests for Gloas

StefanBratanov:
## PR Description
For `deneb.SignedBeaconBlockSupplier` it makes sense to limit it to Fulu, since this class is used specifically in `MiscHelpersDenebPropertyTest` which tests Deneb stuff and requires `blob_kzg_commitments` to be present. Also moved the class to be an inner class of `MiscHelpersDenebPropertyTest`. `SignedBlockContentsSupplier` shouldn't cause any breakage so removed the limitation for `Gloas`. `BeaconBlockSupplier` wasn't used anywhere, so removed it.

## Fixed Issue(s)
related to #9833 

## Documentation

## Changelog


zilm13 (approved):
LGTM

PR #9874, opened 2025-09-09
Rework recent DataColumnSidecars recovery

zilm13:
## Fixed Issue(s)

- moves sidecars to memory to remove synchronization issues between notification and getting them from DB
- clears complete tasks to clean memory
- remove synchrnization
- new tests

## Documentation

## Changelog


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -235,59 +242,53 @@ private boolean isActiveSuperNode(final UInt64 slot) {
...
+      final MetricsHistogram.Timer timer) {
+    final BeaconBlock block = recoveryTask.block.get();
+
+    LOG.debug(
+        "Recovery for block: {}. DataColumnSidecars found: {}",
+        block.getSlotAndBlockRoot(),
+        sidecars.size());
+    final List<DataColumnSidecar> recoveredSidecars =
+        miscHelpers.reconstructAllDataColumnSidecars(sidecars, kzg);
+    timer.closeUnchecked().run();
```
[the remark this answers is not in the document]
zilm13: it will close only once, check internals

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -299,28 +300,30 @@ public SafeFuture<Optional<DataColumnSidecar>> getCustodyDataColumnSidecarByRoot
...
               new AtomicBoolean(false),
               new AtomicBoolean(false));
-      recoveryTasks.put(identifier.getSlotAndBlockRoot(), recoveryTask);
+      final RecoveryTask oldValue =
+          recoveryTasks.putIfAbsent(sidecar.getSlotAndBlockRoot(), recoveryTask);
+      if (oldValue == null) {
+        return;
+      }
     }
+    final RecoveryTask existing = recoveryTasks.get(sidecar.getSlotAndBlockRoot());
```
rolfyone: i'd probably call it `task` or `existingTask` rather than `existing`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -299,28 +300,30 @@ public SafeFuture<Optional<DataColumnSidecar>> getCustodyDataColumnSidecarByRoot
...
+      sidecars.put(DataColumnSlotAndIdentifier.fromDataColumn(sidecar), sidecar);
       final RecoveryTask recoveryTask =
           new RecoveryTask(
               new AtomicReference<>(null),
-              identifiers,
+              sidecars,
               new AtomicBoolean(false),
               new AtomicBoolean(false));
-      recoveryTasks.put(identifier.getSlotAndBlockRoot(), recoveryTask);
+      final RecoveryTask oldValue =
```
rolfyone: and probably `oldTask` or `staleTask`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -299,28 +300,30 @@ public SafeFuture<Optional<DataColumnSidecar>> getCustodyDataColumnSidecarByRoot
...
-      final ConcurrentHashSet<DataColumnSlotAndIdentifier> identifiers = new ConcurrentHashSet<>();
-      identifiers.add(identifier);
+  private void createOrUpdateRecoveryTaskForDataColumnSidecar(final DataColumnSidecar sidecar) {
+    if (!recoveryTasks.containsKey(sidecar.getSlotAndBlockRoot())) {
+      final Map<DataColumnSlotAndIdentifier, DataColumnSidecar> sidecars =
+          new ConcurrentHashMap<>();
+      sidecars.put(DataColumnSlotAndIdentifier.fromDataColumn(sidecar), sidecar);
       final RecoveryTask recoveryTask =
           new RecoveryTask(
               new AtomicReference<>(null),
```
rolfyone: can we give it a better value than a null reference?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -235,59 +242,53 @@ private boolean isActiveSuperNode(final UInt64 slot) {
...
-        .finishError(LOG);
+      final RecoveryTask recoveryTask,
+      final Collection<DataColumnSidecar> sidecars,
+      final MetricsHistogram.Timer timer) {
+    final BeaconBlock block = recoveryTask.block.get();
+
+    LOG.debug(
+        "Recovery for block: {}. DataColumnSidecars found: {}",
+        block.getSlotAndBlockRoot(),
+        sidecars.size());
```
rolfyone: its a little curious that we choose to log during a timer, but it does look like we do it elsewhere too. its not really an 'ideal' time to be logging.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -173,28 +174,34 @@ private boolean shouldSkipProcessing(final UInt64 slot) {
...
+  private void createOrUpdateRecoveryTaskForBlock(final BeaconBlock block) {
+    if (!recoveryTasks.containsKey(block.getSlotAndBlockRoot())) {
+      final RecoveryTask oldValue =
+          recoveryTasks.putIfAbsent(
+              block.getSlotAndBlockRoot(),
+              new RecoveryTask(
+                  new AtomicReference<>(block),
+                  new ConcurrentHashMap<>(),
+                  new AtomicBoolean(false),
+                  new AtomicBoolean(false)));
```
rolfyone: we should just be using `putIfAbsent` (which we call second) and not checking for key...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -173,28 +174,34 @@ private boolean shouldSkipProcessing(final UInt64 slot) {
...
   }
 
-  private synchronized void createOrUpdateRecoveryTaskForBlock(final BeaconBlock block) {
-    if (recoveryTasks.containsKey(block.getSlotAndBlockRoot())) {
-      final RecoveryTask existing = recoveryTasks.get(block.getSlotAndBlockRoot());
-      if (existing.block().get() == null) {
-        existing.block().set(block);
-        maybeStartRecovery(existing);
+  private void createOrUpdateRecoveryTaskForBlock(final BeaconBlock block) {
+    if (!recoveryTasks.containsKey(block.getSlotAndBlockRoot())) {
```
rolfyone, line 178: we shouldnt really check then call putIfAbsent, just call putIfAbsent.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -235,59 +242,53 @@ private boolean isActiveSuperNode(final UInt64 slot) {
...
-            .mapAsync(delegate::getCustodyDataColumnSidecar)
-            .filter(Optional::isPresent)
-            .map(Optional::get)
-            .toList();
-    initiateRecovery(task.block().get(), list);
+    final MetricsHistogram.Timer timer = dataAvailabilityReconstructionTimeSeconds.startTimer();
+    try {
+      initiateRecovery(task, task.existingSidecars.values(), timer);
+    } catch (final Exception e) {
+      task.recoveryStarted.compareAndSet(true, false);
```
rolfyone: its far easier to reason about variables like this if they're set in a similar block, i'd be moving the set false to where it was set true

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -173,28 +174,34 @@ private boolean shouldSkipProcessing(final UInt64 slot) {
...
+  private void maybeStartRecovery(final RecoveryTask task) {
     if (readyToBeRecovered(task)) {
-      task.recoveryStarted().set(true);
-      if (task.existingColumnIds().size() != columnCount) {
+      if (task.recoveryStarted.compareAndSet(false, true)) {
+        if (task.existingSidecars.size() == columnCount) {
+          task.existingSidecars.clear();
+          return;
+        }
         asyncRunner.runAsync(() -> prepareAndInitiateRecovery(task)).finishError(LOG);
```
rolfyone: so an `alwaysRun` at the end setting the recoveryStarted -> false for example (from the move comment about recoveryStarted) would ensure we can see the place we effectively unset the `recoveryStarted`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -235,59 +242,53 @@ private boolean isActiveSuperNode(final UInt64 slot) {
...
-        .alwaysRun(timer.closeUnchecked())
-        .finishError(LOG);
+      final RecoveryTask recoveryTask,
+      final Collection<DataColumnSidecar> sidecars,
+      final MetricsHistogram.Timer timer) {
+    final BeaconBlock block = recoveryTask.block.get();
+
+    LOG.debug(
+        "Recovery for block: {}. DataColumnSidecars found: {}",
+        block.getSlotAndBlockRoot(),
```
rolfyone: so theres no guarantee block will be set... i think we should change the `RecoveryTask` object, but i'll expand on that closer to the object.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
rolfyone: It at least appears we store the whole block ,and we only care about the `slotAndBlockRoot`.
given we can get SlotAndBlockRoot from the sidecar, i'd probably lean towards just having a `SlotAndBlockRoot` attribute of the recoveryTask, and not store the block at all. then we'd no longer need the logic to set it either, and it can never be null.

rolfyone:
as a general note, in one we use `Recovering` and in one we use `Recovery` (class names)- can we pick one if its a test class for that code?

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveryCustodyTest.java
```diff
@@ -289,35 +269,86 @@ public void shouldWaitUntilTimeoutedAndHalfOfSidecars() {
...
+    when(miscHelpersFulu.reconstructAllDataColumnSidecars(anyCollection(), any()))
+        .thenReturn(sidecars.values().stream().toList());
+    custody.onNewValidatedDataColumnSidecar(sidecars.get(UInt64.ZERO));
+    stubAsyncRunner.executeDueActionsRepeatedly();
+
+    // post reconstructed
+    verify(delegate, times(config.getNumberOfColumns())).onNewValidatedDataColumnSidecar(any());
+    verify(listener, times(57)).onNewValidSidecar(any(), eq(RemoteOrigin.RECOVERED));
+    verify(dataColumnSidecarPublisher, times(57)).accept(any());
+  }
```
rolfyone: do we really need 1 test that does all that?

this looks like we're testing multiple things and putting it all in 1 test. can it be split out to the separate things its testing?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: Changed to `existingTask` in both cases

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: It's not stale, it could happen only on concurrent race, let's name it `oldTask`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: Nice catch. As far as we cannot start recovery without 64 sidecars, I've removed onBlock handler at all and changed this field to `slotAndBlockRoot` without concurrency protection as we set it only once and that's all

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: Moved logging right before starting the timer

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: I've removed all synchronization, I should consider race between the threads

tbenr: I agree with Paul, I don't get why you do this sequence, `recoveryTasks.putIfAbsent(` should be operate atomically by design

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -163,7 +163,19 @@ private void maybeStartRecovery(final RecoveryTask task) {
...
+            .runAsync(() -> prepareAndInitiateRecovery(task))
+            .whenException(
+                ex -> {
+                  LOG.debug(
+                      "Error during recovery of {} task with {} sidecars",
+                      task.slotAndBlockRoot == null ? "<unknown>" : task.slotAndBlockRoot,
+                      task.existingSidecars.size(),
+                      ex);
+                  // release task for future retries only if error happened during recovery
+                  task.recoveryStarted.compareAndSet(true, false);
```
tbenr: `task.recoveryStarted.set(false)` looks better to me here

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -298,29 +261,37 @@ public SafeFuture<Optional<DataColumnSidecar>> getCustodyDataColumnSidecarByRoot
...
+          recoveryTasks.putIfAbsent(sidecar.getSlotAndBlockRoot(), recoveryTask);
+      if (oldTask == null) {
+        return;
+      }
     }
+    final RecoveryTask existingTask = recoveryTasks.get(sidecar.getSlotAndBlockRoot());
+    existingTask
+        .existingSidecars()
+        .put(DataColumnSlotAndIdentifier.fromDataColumn(sidecar), sidecar);
+    maybeStartRecovery(existingTask);
```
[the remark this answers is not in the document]
zilm13: Negative all the code to handle race in a correct way

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: Right, I'm too paranoid

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveryCustodyTest.java
zilm13: Right, I made 2 tests which verifies expected behavior and looks more like unit-tests

zilm13:
@rolfyone Nice review, with your help I was able to remove a page of production code, which means less bugs. All feedback is addressed and ready for final round of review!

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -298,29 +275,37 @@ public SafeFuture<Optional<DataColumnSidecar>> getCustodyDataColumnSidecarByRoot
...
+      sidecars.put(DataColumnSlotAndIdentifier.fromDataColumn(sidecar), sidecar);
       final RecoveryTask recoveryTask =
           new RecoveryTask(
-              new AtomicReference<>(null),
-              identifiers,
+              sidecar.getSlotAndBlockRoot(),
+              sidecars,
               new AtomicBoolean(false),
               new AtomicBoolean(false));
-      recoveryTasks.put(identifier.getSlotAndBlockRoot(), recoveryTask);
```
rolfyone: i still think this is much better expressed as putIfAbsent, im not sure why we're so set on containsKey.
if its only because of the sidecars concurrentHashMap just abstract that away anddo the `new RecoveryTask` inline in the putIfAbsent

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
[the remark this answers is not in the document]
rolfyone: im not really sure why we're re-fetching, we'd have the reference so theres no value in this, we could just call `maybeStartRecovery(<the result of putIfAbsent>)` if its not null, so i think cursorbot is actually helpful here

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -173,31 +158,33 @@ private boolean shouldSkipProcessing(final UInt64 slot) {
...
 
-  private synchronized void createOrUpdateRecoveryTaskForBlock(final BeaconBlock block) {
-    if (recoveryTasks.containsKey(block.getSlotAndBlockRoot())) {
-      final RecoveryTask existing = recoveryTasks.get(block.getSlotAndBlockRoot());
-      if (existing.block().get() == null) {
-        existing.block().set(block);
-        maybeStartRecovery(existing);
+  protected void maybeStartRecovery(final RecoveryTask task) {
+    if (readyToBeRecovered(task)) {
+      if (task.recoveryStarted.compareAndSet(false, true)) {
```
rolfyone, line 163: so the reason i think this will be buggy, is we set true here,
then we set false but only if exception, and we never call set again...
so i guess the implication is if its started, and doesnt cause an exception that theres a single async task running? how do we rate limit that?

This seems like a source of issues in future, so we need to at least document what we mean when we're setting, and if its still true, and we've had no exception and no background task, and it didnt get all the columns, then it's just stuck in memory permanently until restart...

So i guess that drills into testing - making sure that we cant get to the task complete with no exception if all the columns aren't found etc...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: I was slow on brains, it's simple, done in https://github.com/Consensys/teku/pull/9874/commits/38c5abf3fe3cb9dbcb732e5fe69211911876ed9c

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: > how do we rate limit that?

do we need?
it's only for the head slot and started only when we get 64+ columns
I assume due to bug we can restart something and continue retries without success, but 
1. it will go away and GCed in 32 slots (max tasks quantity)
2. it will be restarted only on new sidecar (maximum 63 events with total sidecars >64 and != 128)

I've added clear comment on logic:
```
              // release task for future retries only if error happened during recovery
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -210,12 +197,12 @@ private boolean readyToBeRecovered(final RecoveryTask task) {
...
     }
 
-    if (task.existingColumnIds().size() < recoverColumnCount) {
+    if (task.existingSidecars().size() < recoverColumnCount) {
       // not enough columns collected
       return false;
     }
 
-    if (task.block().get() == null) {
+    if (task.slotAndBlockRoot == null) {
```
rolfyone: this case i don't think we can ever get to?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
rolfyone: that looks great now.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
rolfyone: should be fine, the supporting code is much better now so i think we can mark this resolved.

rolfyone (approved):
LGTM, just need to remove the if that will now never be true (we've always got slotAndBlockRoot present in that task)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
zilm13: Yeah, nice catch. Removed

PR #9877, opened 2025-09-09
Optimize `process_proposer_lookahead`

StefanBratanov:
## PR Description
We can take advantage of the mutability to optimize this code. Also, I feel this code follows spec much more closely https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/beacon-chain.md#new-process_proposer_lookahead

## Fixed Issue(s)
N/A

## Documentation

## Changelog


zilm13 (approved):
LGTM
right by the spec
Do we have non-reference tests for this?

StefanBratanov:
I will close this. Doesn't seem to add value at the moment after some benchmarking. I am also not 100% sure how calling `.set` these many times on a mutable vector affects performance.

PR #9880, opened 2025-09-10
Move execution requests processing to `ExecutionRequestsProcessor`

StefanBratanov:
## PR Description
Processing execution requests move to `process_execution_payload` in Gloas instead of the block, so to prepare for this can move all the logic from `BlockProcessor` to `ExecutionRequestsProcessor`

## Fixed Issue(s)
N/A

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/common/operations/OperationProcessor.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/common/operations/OperationProcessor.java
@@ -67,13 +67,11 @@ public interface OperationProcessor {
   void processWithdrawals(MutableBeaconState state, ExecutionPayloadSummary payloadSummary)
       throws BlockProcessingException;
 
-  void processDepositRequest(MutableBeaconState state, List<DepositRequest> depositRequest)
-      throws BlockProcessingException;
+  void processDepositRequest(MutableBeaconState state, List<DepositRequest> depositRequest);
 
-  void processWithdrawalRequest(MutableBeaconState state, List<WithdrawalRequest> withdrawalRequest)
-      throws BlockProcessingException;
+  void processWithdrawalRequest(
+      MutableBeaconState state, List<WithdrawalRequest> withdrawalRequest);
 
   void processConsolidationRequests(
-      MutableBeaconState state, List<ConsolidationRequest> consolidationRequest)
-      throws BlockProcessingException;
+      MutableBeaconState state, List<ConsolidationRequest> consolidationRequest);
 }
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
@@ -15,24 +15,18 @@ package tech.pegasys.teku.spec.logic.versions.electra.block;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static tech.pegasys.teku.infrastructure.unsigned.UInt64.ZERO;
-import static tech.pegasys.teku.spec.config.SpecConfig.FAR_FUTURE_EPOCH;
-import static tech.pegasys.teku.spec.config.SpecConfigElectra.FULL_EXIT_REQUEST_AMOUNT;
 
 import it.unimi.dsi.fastutil.ints.IntList;
 import it.unimi.dsi.fastutil.objects.Object2IntMap;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.stream.IntStream;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.bls.BLSPublicKey;
 import tech.pegasys.teku.bls.BLSSignature;
-import tech.pegasys.teku.ethereum.execution.types.Eth1Address;
-import tech.pegasys.teku.infrastructure.bytes.Bytes20;
 import tech.pegasys.teku.infrastructure.ssz.SszList;
-import tech.pegasys.teku.infrastructure.ssz.SszMutableList;
 import tech.pegasys.teku.infrastructure.ssz.collections.SszBitlist;
 import tech.pegasys.teku.infrastructure.ssz.primitive.SszBytes32;
 import tech.pegasys.teku.infrastructure.ssz.primitive.SszUInt64;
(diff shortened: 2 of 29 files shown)


eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/common/operations/OperationProcessor.java
```diff
@@ -67,13 +67,11 @@ void processBlsToExecutionChange(
   void processWithdrawals(MutableBeaconState state, ExecutionPayloadSummary payloadSummary)
       throws BlockProcessingException;
 
-  void processDepositRequest(MutableBeaconState state, List<DepositRequest> depositRequest)
-      throws BlockProcessingException;
+  void processDepositRequest(MutableBeaconState state, List<DepositRequest> depositRequest);
```
tbenr, line 70: don't we need to have a ExecutionProcessingException for consistency?

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
```diff
@@ -160,17 +146,24 @@ protected void processOperationsNoValidation(
...
-          processConsolidationRequests(state, executionRequests.getConsolidations());
-        });
+  public void processExecutionRequests(
+      final MutableBeaconState state,
+      final BeaconBlockBody body,
+      final Supplier<ValidatorExitContext> validatorExitContextSupplier) {
+    final ExecutionRequests executionRequests =
+        body.getOptionalExecutionRequests()
+            .orElseThrow(
+                () -> new NoSuchElementException("Execution requests expected as part of body"));
```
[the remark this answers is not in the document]
tbenr: I think the bot is on something here..

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/execution/ExecutionRequestsProcessorElectra.java
```diff
@@ -0,0 +1,481 @@
...
+import tech.pegasys.teku.spec.logic.common.helpers.MiscHelpers;
+import tech.pegasys.teku.spec.logic.common.helpers.Predicates;
+import tech.pegasys.teku.spec.logic.common.util.ValidatorsUtil;
+import tech.pegasys.teku.spec.logic.versions.electra.helpers.BeaconStateAccessorsElectra;
+import tech.pegasys.teku.spec.logic.versions.electra.helpers.BeaconStateMutatorsElectra;
+import tech.pegasys.teku.spec.logic.versions.electra.helpers.MiscHelpersElectra;
+import tech.pegasys.teku.spec.logic.versions.electra.helpers.PredicatesElectra;
+import tech.pegasys.teku.spec.schemas.SchemaDefinitionsElectra;
+
+public class ExecutionRequestsProcessorElectra implements ExecutionRequestsProcessor {
```
tbenr, line 54: so, despite dedicated execution processing is a Gloas thing, we moved them already on an Electra processor so we can do inheritance on Gloas and avoid code duplication?

tbenr (commented):
first pass:
So my main concern is exception handling.
We used to have a BlockProcessingException which is cought and wrapped to a StateTranstionException.

IIRC processing the payload is also a state transition so I wander if we should follow the same pattern

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/block/BlockProcessorElectra.java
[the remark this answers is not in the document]
tbenr: cc @StefanBratanov

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/electra/execution/ExecutionRequestsProcessorElectra.java
StefanBratanov: Yes, basically the execution requests processing will be part of processing the execution envelope in Gloas, so we can have processor for Electra, Fulu and then reuse that in Gloas rather than duplicating code or using the block processor as a dependency for the execution processor which is bad design imo

StefanBratanov:
> first pass: So my main concern is exception handling. We used to have a BlockProcessingException which is cought and wrapped to a StateTranstionException.
> 
> IIRC processing the payload is also a state transition so I wander if we should follow the same pattern

But we don't throw exceptions there. Dependents should use `safelyProcess` to wrap those like we do now. I think we were just adding `throw BlockProcessingException` to all methods without considering if needed or not.

image

StefanBratanov:
Oh, you are right about `NoSuchElementException`, it should be `BlockProcessingException`, since `safelyProcess` doesn't handle those

tbenr:
> > first pass: So my main concern is exception handling. We used to have a BlockProcessingException which is cought and wrapped to a StateTranstionException.
> > IIRC processing the payload is also a state transition so I wander if we should follow the same pattern
> 
> But we don't throw exceptions there. Dependents should use `safelyProcess` to wrap those like we do now. I think we were just adding `throw BlockProcessingException` to all methods without considering if needed or not.
> 
> image

But if we have a `safelyProcess` in the `ExecutionRequestsProcessor` and catch exceptions there and throw a `ExecutionRequestsProcessingException`, we will have a clean exception that will signal something wrong in that processing.


in Electra\Fulu that exception will be caught by the `safelyProcess` of the block processor, while in Gloas we will catch it directly and mapped to `StateTransitionException` in Spec (if we will have payload dedicated spec methods (`processPayload`?)  like `processBlock`)

StefanBratanov:
> > > first pass: So my main concern is exception handling. We used to have a BlockProcessingException which is cought and wrapped to a StateTranstionException.
> > > IIRC processing the payload is also a state transition so I wander if we should follow the same pattern
> > 
> > 
> > But we don't throw exceptions there. Dependents should use `safelyProcess` to wrap those like we do now. I think we were just adding `throw BlockProcessingException` to all methods without considering if needed or not.
> > image
> 
> But if we have a `safelyProcess` in the `ExecutionRequestsProcessor` and catch exceptions there and throw a `ExecutionRequestsProcessingException`, we will have a clean exception that will signal something wrong in that processing.
> 
> in Electra\Fulu that exception will be caught by the `safelyProcess` of the block processor, while in Gloas we will catch it directly and mapped to `StateTransitionException` in Spec (if we will have payload dedicated spec methods (`processPayload`?) like `processBlock`)

It won't be captured though at least in current code, `safelyProcess` captures generic errors `ArithmeticException | IllegalArgumentException | IndexOutOfBoundsException e`, anything outside will go all the way to `BlockImporter` and be considered `internalError` rather than `failedStateTransition`.

tbenr:
> It won't be captured though at least in current code, safelyProcess captures generic errors ArithmeticException | IllegalArgumentException | IndexOutOfBoundsException e, anything outside will go all the way to BlockImporter and be considered internalError rather than failedStateTransition.

this is what I'm actually proposing. I see that exception to be a "sub exception" of the block processing, in pre-gloas forks, because on those forks, payload processing is part of the block processing. It won't be anymore in gloas and beyond

StefanBratanov:
> > It won't be captured though at least in current code, safelyProcess captures generic errors ArithmeticException | IllegalArgumentException | IndexOutOfBoundsException e, anything outside will go all the way to BlockImporter and be considered internalError rather than failedStateTransition.
> 
> this is what I'm actually proposing. I see that exception to be a "sub exception" of the block processing, in pre-gloas forks, because on those forks, payload processing is part of the block processing. It won't be anymore in gloas and beyond

Yeah I think it can add value, this PR behaves exactly the same as current `master` though, the execution requests method in the interface just say throws `BlockProcessingException` but in the overriden methods none of them throw anything, it even has deleted the throws, so any generic exception would be still be caught by `safelyProcess` and wrapped.

tbenr:
> Yeah I think it can add value, this PR behaves exactly the same as current `master` though, the execution requests method in the interface just say throws `BlockProcessingException` but in the overriden methods none of them throw anything, it even has deleted the throws, so any generic exception would be still be caught by `safelyProcess` and wrapped.

I see what you mean. The code has not being changed, so what was caught previously is still caught on the `safelyProcess` on the BlockProcessor side

tbenr (approved):
LGTM. We can discuss the exception thing for a separate PR

PR #9887, opened 2025-09-11
Improve Static peer connection persistence logic

tbenr:
~~I did some tests and appears to me we are only applying reputation on outgoing connection attempt, and we do not check it on incoming connection.~~

~~So without this PR, a peer that get disconnected due to reputation can reconnect immediately without any problems, and all rate limiting trackers will be recreated brand new. We only consider reputation in `Eth2PeerSelectionStrategy` which is outbound.~~

~~There is an interaction with Static peers.~~
~~What I did:~~
~~- if 2 nodes are configured as mutually static, they both consider them trusted and the Reputation never goes down. Thus inbound\outbound connection are always allowed.~~

~~- if static is not mutually configured, the reconnection happens with increasing delay, up to matching the cooldown period. This way the connection is eventually reestablished.~~

~~TLDR: with this PR mutually configured static peers reconnect faster, non-mutual static peers reconnect slower (4 minutes: `(3+9+27+81+120) / 60`)~~

**We decided to make this PR just a static peer handling improvement, as applying reputation to inbound connection requires some additional changes**

## Documentation

## Changelog


tbenr:
The problem here is the interaction with static peers that are hitting rate limit, so they are temporary banned.

the behavior will be:
1. remote static peer hits rate limit.
2. we disconnect from it and notify reputation manager
3. we try to reconnect via `ConnectionManager::maintainPersistentConnection` after 20 second
4. we immediately disconnect it again in the `handleConnection` callback triggered by our outgoing connection attempt.

we shouldn't even try to reconnect.

~~a solution is to make the reputationManager "static peer" aware, so we simply never ban them (so they will be trusted)~~
actually, reputationManager is already "static peer" aware. 
https://github.com/Consensys/teku/blob/839a13b1a9bb98b3ed1928d1b4f4a36c2d6eeee3/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/DefaultReputationManager.java#L100-L104

Why are we doing that filtering only on `adjustReputation` and not on `reportDisconnection` and the others? cc @rolfyone 

Another option could be to make `Peer` expose an `isStatic` method, so that in `Eth2PeerFactory` we can pass a NOOP for all the trackers. So we will never disconnect it. To be discussed.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/PeerManager.java
```diff
@@ -98,6 +100,11 @@ public PeerManager(
   @Override
   public void handleConnection(@NotNull final Connection connection) {
     Peer peer = new LibP2PPeer(connection, rpcHandlers, reputationManager, peerScoreFunction);
```
rolfyone: `final` since we're here anyway?

tbenr:
if we don't want to do many changes as per my comment https://github.com/Consensys/teku/pull/9887#issuecomment-3282010509 the only disturbing thing now is:

when static peer declaration is not mutual:
the cooldown period is 2 minutes but the reconnect attempt happens every 20 seconds. So in that situation the connection will never be reestablished. If we implement an exponential time retry (not 20s constant) maybe could be a solution.

OR: if we don't reset the cooldown period when we disconnect the peer, than it would work as is, without changing the retry

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
```diff
@@ -156,8 +176,48 @@ public void shouldReconnectWhenPersistentPeerDisconnects() {
...
+    verify(network, times(1)).connect(PEER1);
+
+    // advance time enough to trigger a retry
+    timeProvider.advanceTimeBySeconds(9);
+    asyncRunner.executeDueActions();
+
     verify(network, times(2)).connect(PEER1);
+
+    // fast disconnection, last disconnection delay should not reset
+    assertThat(manager.getLastStaticPeerReconnectionDelay(PEER1)).isEqualTo(9);
```
zilm13, line 191: what is the difference between this test and `shouldRetryConnectionToStaticPeerAfterRetryAndDisconnect`?

zilm13: what is the difference between this test and `shouldRetryConnectionToStaticPeerAfterRetryAndDisconnect`?

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
```diff
@@ -440,6 +500,43 @@ public void shouldSwitchToLongDelayAfterFirstPeerIsFound() {
...
+
+    final StubPeer peer1 = new StubPeer(new MockNodeId(1));
+
+    assertThat(manager.computeStaticPeerRetryDelay(peer1.getAddress()))
+        .isEqualTo(Duration.ofSeconds(3));
+
+    final StubPeer peer2 = new StubPeer(new MockNodeId(2));
+
+    assertThat(manager.computeStaticPeerRetryDelay(peer2.getAddress()))
+        .isEqualTo(Duration.ofSeconds(3));
```
zilm13: not clear that it's per peer, it's the same

networking/p2p/src/testFixtures/java/tech/pegasys/teku/network/p2p/DiscoveryNetworkFactory.java
```diff
@@ -70,6 +71,8 @@ public class DiscoveryTestNetworkBuilder {
     private final List<String> bootnodes = new ArrayList<>();
     private final Spec spec = TestSpecFactory.createMinimalPhase0();
 
+    private volatile TimeProvider timeProvider = StubTimeProvider.withTimeInSeconds(0);
```
zilm13: it's builder, do we really need `volatile`?

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManager.java
```diff
@@ -226,32 +238,71 @@ private SafeFuture<Peer> maintainPersistentConnection(final PeerAddress peerAddr
...
+                  retryDelay.toSeconds());
               failedConnectionCounter.inc();
               return asyncRunner.runAfterDelay(
-                  () -> maintainPersistentConnection(peerAddress), RECONNECT_TIMEOUT);
+                  () -> maintainPersistentConnection(peerAddress), retryDelay);
             });
   }
 
+  private boolean connectionLastedLongEnough(final UInt64 connectionTimeMillis) {
+    final UInt64 immediateDisconnectionTime =
```
zilm13, line 277: A bit difficult to understand what it means without checking usage. I'd call it something like `meaningful connection duration`, but I don't like such name because it's long. Maybe you have more ideas

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/reputation/DefaultReputationManager.java
```diff
@@ -145,12 +159,12 @@ public void reportDisconnection(
           || reason.map(DisconnectReason::isPermanent).orElse(false)) {
         suitableAfter = Optional.of(disconnectTime.plus(BAN_PERIOD));
         score.set(DEFAULT_SCORE);
-      } else if (suitableAfter.isEmpty()) {
+      } else if (reason.isPresent() && suitableAfter.isEmpty()) {
```
zilm13: if we have remote faulty peer, we could retry reconnection immediately forever, so I think we should be in this if case when
- reason + local + suitable.empty
- !local + suitable.empty

Actually it would be good to have here the same coolDown logic as for static peers, but too much of complexity for this case.

zilm13 (commented):
LGTM
As discussed, it would be good to have metrics and test it on testnet

networking/p2p/src/test/java/tech/pegasys/teku/networking/p2p/connection/ConnectionManagerTest.java
tbenr: `shouldRetryConnectionToStaticPeerAfterRetryAndDisconnect` covers the connection failure path via future completing exceptionally

zilm13 (approved):
LGTM

PR #9888, opened 2025-09-12
Re-enable recover_cells_and_kzg_proofs_case_invalid_shuffled

### LINKED ISSUE #9832: re-enable `recover_cells_and_kzg_proofs_case_invalid_shuffled` ref tests
once CKZG is updated, we should be able to re-enable the reference tests disabled as part of alpha.6 update. (#9830)

I've excluded in `KzgRecoverCellsAndKzgProofsTestExecutor` with:

```
    if (testDefinition.getTestName().startsWith("kzg-mainnet/recover_cells_and_kzg_proofs_case_invalid_shuffled")) {
      LOG.error("SKIPPING TEST {}", testDefinition.getTestName());
      return;
    }
```

The failure will come from the CKZG library rather than our validations.

lucassaldanha:
fixes #9832


StefanBratanov (approved):
LGTM

PR #9889, opened 2025-09-13
Add per subnet subscribers target for dynamic peer requirements

zilm13:
## PR Description
We have flag and configurable variable:
`--Xp2p-target-subnet-subscriber-count` which currently defaults to 2. It's required by the spec. It's very strange, but it's required by the spec. It checks the number of subnet subscription and if there is any, it will return it's value. So with default 2 it requires 2 total peers for any number of subscriptions. No subscriptions - 0 peers, 1,2,3,100 subnets - 2 peers total.

I've introduced flag `--Xp2p-target-per-subnet-subscriber-count` which is a configuration how many peers are required for each subnet. It's goal is to dynamically require additional peers depending on the node subnet subcriptions. So if say, you are a supernode with 128 subnets, you require 128 peers, if you need 8 custody subscriptiions, you require 8 peers. It's an experimental flag and defaults to unset so it will not affect any current behavior.
## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes peer requirement calculation used by peer selection to optionally scale with the number of subscribed subnets, which can affect target peer counts and network connectivity behavior. Default remains unset so existing behavior is preserved unless the new hidden flag is enabled.
> 
> **Overview**
> Introduces a new hidden config/CLI option `--Xp2p-target-per-subnet-subscriber-count` (stored as `OptionalInt` in `P2PConfig`) to optionally specify a *per-subnet* subscriber target.
> 
> Updates `PeerSubnetSubscriptions.getSubscribersRequired()` to account for this per-subnet target by computing total subscriber shortfall across relevant attestation, sync committee, and data column sidecar subnets, and taking the max of that and the existing global/subnet-minimum based requirement.
> 
> Wires the new config through `Eth2P2PNetworkBuilder` into `PeerSubnetSubscriptions.create`, updates test fixtures, and adds/updates tests and YAML config parsing to cover the new setting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 026a65ee4dfebd60f04aad55ad6c80b6195222fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


zilm13:
I'd scrap this
We've improved connectivity with #10846 and it looks good alone. For supernodes we could suggest just higher lower and upper limits and it will be enough without complicated addons.

PR #9890, opened 2025-09-15
Cleaned up recovery map as tasks complete

### LINKED ISSUE #9875: Rework DataColumnSidecars recovery during sync
It happens in class `RecoveringSidecarRetriever`.
There are at least 2 issues with it:
- missing sidecars similar to recent sidecars recovery
- huge memory usage, could take several Gbs in certain situations

rolfyone:
Added a timed task in case we miss items going through completed, but generally speaking the completed cleanup should finish things off nicely.

Also addressed the issue of storing the block in the recoveryEntry when we only need the slot and root.

I do think there's more to do here but focused on the stated problem.

fixes #9875

## Documentation

## Changelog

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
@@ -106,7 +106,7 @@ public class RecoveringSidecarRetriever implements DataColumnSidecarRetriever {
   private void pruneRecoveryBySlot() {
     final List<UInt64> cleanEntries =
         recoveryBySlot.entrySet().stream()
-            .filter(entry -> entry.getValue().isCancelled() || entry.getValue().isDone())
+            .filter(entry -> entry.getValue().isCleanupCandidate())
             .map(Map.Entry::getKey)
             .toList();
     LOG.debug(
@@ -254,10 +254,7 @@ public class RecoveringSidecarRetriever implements DataColumnSidecarRetriever {
         (slot) -> {
           LOG.trace("Cleanup recovery for slot {}", slot);
           final RecoveryEntry entry = recoveryBySlot.remove(slot);
-          if (entry != null
-              && !entry.isCancelled()
-              && !entry.isDone()
-              && !entry.responsesByColIdx.isEmpty()) {
+          if (entry != null) {
             LOG.trace(
                 "Cleaning up after completed task but recovery entry was not completed cleanly {}:{} ",
                 entry.slot,
@@ -418,12 +415,7 @@ public class RecoveringSidecarRetriever implements DataColumnSidecarRetriever {
           .flatMap(Collection::stream)
           .forEach(response -> response.cancel(true));
       if (recoveryPeerRequests != null) {
-        recoveryPeerRequests.forEach(
-            request -> {
-              if (!request.isDone()) {
-                request.cancel(true);
-              }
-            });
+        recoveryPeerRequests.forEach(request -> request.cancel(true));
       }
       if (reconstructionInProgress != null) {
         reconstructionInProgress.cancel(true);
@@ -504,6 +496,10 @@ public class RecoveringSidecarRetriever implements DataColumnSidecarRetriever {
 
     boolean isDone() {
       return reconstructionInProgress != null && reconstructionInProgress.isDone();
+    }
+
+    boolean isCleanupCandidate() {
+      return isCancelled() || isDone();
     }
   }
 }
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
@@ -152,7 +152,7 @@ public class RecoveringSidecarRetriever implements DataColumnSidecarRetriever {
   public synchronized void stop() {
     if (recoveryBySlotCleaner != null) {
       recoveryBySlotCleaner.cancel();
-      pendingRequestsChecker = null;
+      recoveryBySlotCleaner = null;
     }
     if (pendingRequestsChecker != null) {
       pendingRequestsChecker.cancel();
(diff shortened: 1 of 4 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
```diff
@@ -202,28 +237,48 @@ private void maybeInitiateRecovery(
...
+          columnId.slot(),
+          (slot, entry) -> {
+            if (entry.blockRoot.equals(recoveryEntry.blockRoot)) {
+              entry.cancel();
+              return createNewRecovery(block);
+            }
+            return entry;
+          });
+    }
+    return recoveryEntry;
```
tbenr: I think the bot is on something.
I don't see why we should nest this.
we could have something like

```java
      recoveryBySlot.compute(columnId.slot(), (slot, existingEntry) -> {
        if (existingEntry != null && existingEntry.blockRoot.equals(block.getRoot())) {
          return existingEntry;
        } else if (existingEntry != null) {
          existingEntry.cancel();
        }
        return createNewRecovery(block);
      });
```
which preserve atomicity

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
```diff
@@ -101,11 +103,35 @@ public RecoveringSidecarRetriever(
...
+            .toList();
+    LOG.debug(
+        "Pruning recovery by slot, found {} of a total {} entries...",
+        cleanEntries.size(),
+        recoveryBySlot.size());
+
+    final int limit = Math.min(cleanEntries.size(), 500);
+    int count = 0;
+    for (; count < limit; count++) {
+      final RecoveryEntry entry = recoveryBySlot.remove(cleanEntries.get(count));
```
tbenr, line 120: I wonder if it would be cleaner to call the taskCleaner we already have prepared, so this cleaning is done in one place only

tbenr (commented):
first pass, a couple of things, waiting for a check from @zilm13 too

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
rolfyone: yep had a quick try, but will go with that ta

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
rolfyone: already prepared?

zilm13:
Looks better than it was, no complains

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
```diff
@@ -202,28 +237,48 @@ private void maybeInitiateRecovery(
...
+    return recoveryEntry;
   }
 
   private RecoveryEntry createNewRecovery(final BeaconBlock block) {
-    final RecoveryEntry recoveryEntry = new RecoveryEntry(block, kzg, specHelpers);
+    final Consumer<UInt64> cleanup =
+        (slot) -> {
+          LOG.trace("Cleanup recovery for slot {}", slot);
+          final RecoveryEntry entry = recoveryBySlot.remove(slot);
+          if (entry != null
```
zilm13: This if is too complex. Is there is a way to simplify it or refactor it?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
rolfyone: oh as in taskCleaner that the i pass to `RecoveryEntry`? and make that class wide? thats less clean that way imo. i made that consumer so that we don't have RecoveryEntry do dodgy things, because next PR i'll probably break out that into its own file and start testing it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
rolfyone: what? im just checking that the entry is not active.

rolfyone: when we store a bunch of attributes that may be active, i can either check them, or ignore them. it seems bad to ignore them. i can fix the design... but thats not a small pr.

or i can make cancel handle it and just potentially have the trace more often.
it should be clean but in a concurrent world i cant guarantee that it hasn't changed since i determined it was dead.

rolfyone:
well i'd need an approval to actually merge this, so no comlpaints is one thing but i need to merge it so i can move on...

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/RecoveringSidecarRetriever.java
tbenr: ye my idea was to stream the entries and not the keys, and call the `taskCleaner` consumer

rolfyone: for now i think this might be simpler but i can re-assess when im in that area...
the whole point of the cleaner was to not have the RecoveryEntry class go playing about in the parents attributes, but its kind of already doing that, im just preparing to do it a bunch more and dont want another thing to clean

lucassaldanha (approved):

PR #9896, opened 2025-09-16
Upgrade to Gradle 9

StefanBratanov:
## PR Description
https://docs.gradle.org/current/userguide/upgrading_major_version_9.html

## Fixed Issue(s)
N/A

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Medium risk because this upgrades the build toolchain to Gradle 9.4 and refactors many build tasks/plugins, which may change task wiring/parallelism and affect CI/release artifacts.
> 
> **Overview**
> Upgrades the Gradle wrapper to `9.4.0` and updates build configuration to be Gradle 9–compatible (e.g., converting many `task ...` definitions to `tasks.register`, updating IDEA module source configuration, and modernizing assignment syntax in publishing/repositories).
> 
> Adjusts build behavior to avoid Gradle 9 project-lock/parallelism issues: license report checks are now run via a new `licenseCheck` task that spawns a separate `gradlew ... --no-parallel` process, and `eth-reference-tests` introduces a local `refTestGenerator` configuration to avoid cross-project classpath locking. Also updates the dependency license report plugin to `3.1.1` and tweaks `docker`/`cloudsmithUpload` tasks to use injected `ExecOperations`/`FileSystemOperations`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 821cb43188c433c706dc8cd282d1d7834e678091. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


zilm13:
so there are issues left like
```
Execution failed for task ':generateLicenseReport'.
> Resolution of the configuration ':acceptance-tests:runtimeClasspath' was attempted without an exclusive lock. This is unsafe and not allowed.
```

it's not a bug it's how Gradle was changed, see https://github.com/gradle/gradle/issues/33938#issuecomment-2985249363

it could be resolved with disabling `parallel`, which is not a way we should go. So the issue (as I understand) is that subprojects have now stricter restrictions from access from root project and we try to access it too early. The first thing I'd try is move such projects in the end of execution queue by adding dependsOn. I've already tried general solutions like `mustRunAfter allprojects.collectMany { it.tasks }` or `shouldRunAfter tasks.matching { it.name != 'generateLicenseReport' }`. Probably need to add proper dependsOn.

zilm13:
There were several changes done:
- exec and copy commands are not allowed direclty in Gradle 9, so they were injected to support tasks that use it (all using command line)
- checkLicense is not working for submodules since Gradle 9 and it doesn't look like it will be fixed. The workaround was done disabling it for root project and running for all submodules instead. You could easily test it by adding some GPL license dependency and see that `assemble` will fail and generated report will include bad dependency with details
- similar lock issue with reference tests, but it's single submodule not like checkLicense, so the fix was a way easier and totally isolated in reference tests module
- small warning fixes like `=` instead of `"  "`

zilm13 (approved):
LGTM

tbenr (approved):
we have gradle 9.4 now :D

can we upgrade to it and let's try to merge

PR #9900, opened 2025-09-16
Fix SafeFuture's async methods

tbenr:
Major implications on Channels which fail to complete target features when the underlining async runner fails (ie queue is full)

## Documentation

## Changelog


lucassaldanha (approved):
🚀

PR #9901, opened 2025-09-17
Adding source error as suppressed exception when propagating async

lucassaldanha:
## PR Description

When propagating from future A to future B in a separate thread (using `SafeFuture.propagateToAsync`), failure may occur due to `AsyncRunner` rejecting the execution or for some other unknown reason.

When this happens, future B is completed exceptionally with `cause` related to the error thrown by the `AsyncRunner`. However, if future A also completes exceptionally, we don't propagate that to B.

This change keeps that link by adding A's thrown exception as a suppressed exception of B's cause exception.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


tbenr (approved):
LGTM

PR #9907, opened 2025-09-17
Replace `ExpectedWithdrawals` with `WithdrawalsHelpers`

StefanBratanov:
## PR Description
Withdrawals change a lot in Gloas https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#withdrawals
I found `ExpectedWithdrawals` class to be hard to reason about and modify. Even though not perfect, I created `WithdrawalsHelpers` which would make it easier to work on Gloas.

## Fixed Issue(s)
N/A

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/withdrawals/WithdrawalsHelpers.java
```diff
@@ -0,0 +1,50 @@
...
+import tech.pegasys.teku.infrastructure.bytes.Bytes20;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.datastructures.execution.ExecutionPayloadSummary;
+import tech.pegasys.teku.spec.datastructures.execution.versions.capella.Withdrawal;
+import tech.pegasys.teku.spec.datastructures.state.Validator;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.BeaconState;
+import tech.pegasys.teku.spec.datastructures.state.beaconstate.MutableBeaconState;
+import tech.pegasys.teku.spec.logic.common.statetransition.exceptions.BlockProcessingException;
+
+/** Helpers for post-Capella withdrawals */
```
rolfyone: not sure its 'post-capella', that would suggest for deneb onwards... this is really just helpers for withdrawals, and thats probably redundant given the class name

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/withdrawals/WithdrawalsHelpers.java
```diff
@@ -0,0 +1,50 @@
...
+public interface WithdrawalsHelpers {
+
+  ExpectedWithdrawals getExpectedWithdrawals(BeaconState state);
```
rolfyone, line 29: should probably put the `// get_expected_withdrawals` comment above this line so we can find it from spec

rolfyone (approved):
LGTM overall, haven't checked test coverage but assuming that was looked at. we do seem to have removed a bunch in tests.

StefanBratanov:
> LGTM overall, haven't checked test coverage but assuming that was looked at. we do seem to have removed a bunch in tests.

`ExpectedWithdrawalsTest` was moved into 3 test classes. For other comments, makes sense, I will cleanup tomorrow.

Issue #9835, opened 2025-09-01
[API] add getBlobs endpoint


rolfyone:
https://github.com/ethereum/beacon-APIs/pull/546

PR #9908, opened 2025-09-17
VersionedHashDBSource tests

zilm13:
## PR Description
Part of #9835 

Start from `VersionedHashDBSource`
- it's used from `KvStoreDatabase` for every actions with `VersionedHash`'es
- it's initialized with `VersionedHashDBSourceFactory` starting from `VersionedDatabaseFactory`
- method `getSidecarIdentifier` will be used later to get corresponding `Blob`
- not adding `VersionedHash`-es to `Store`, because it would be difficult to keep it update along with blobSidecars. Also DataColumnSidecars are not a part of the `Store` and niche of VersionedHash usage is only `getBlobs` RPC.

## Documentation

## Changelog

---

> [!NOTE]
> Add VersionedHash storage/lookup tests and integrate versioned-hash assertions for blob/data-column sidecars, plus minor KZG fixture tweak and small spec/util helpers.
> 
> - **Storage/tests**:
>   - Add `VersionedHashDBSourceTest` and `VersionedHashDBSourceFactoryTest` covering activation, add/remove, and serialization of sidecar identifier metadata; wire via `CustodyGroupCountChannel`.
>   - Extend `DatabaseTest` to assert `VersionedHash` presence/absence for canonical and non-canonical `BlobSidecar` and `DataColumnSidecar`, including pruning flows; add helpers to activate and verify versioned-hash mappings and combined availability checks.
>   - Update `ChainStorageTest` to verify `SidecarIdentifier` retrieval by `VersionedHash` for finalized historical blob sidecars.
> - **Spec/util**:
>   - `DataStructureUtil`: add `randomVersionedHash()`, refactor `randomVersionedHashes`, and overload `randomDataColumnSidecar(...)` to accept existing commitments.
>   - `BlobSidecar`: add `getSlotAndBlockRootAndBlobIndex()` helper.
> - **KZG test fixture**:
>   - `NoOpKZG.blobToKzgCommitment` now derives commitment from the blob’s first 48 bytes instead of returning zeros.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0e97154dae00f1352b8a45275f4b0d86cc7dc792. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
```diff
@@ -0,0 +1,89 @@
...
+
+  public VersionedHashDBSource createVersionedHashDBSource(final KvStoreCombinedDao dao) {
+    final VersionedHashDBSource versionedHashDBSource =
+        new VersionedHashDBSource(
+            dao, blobSidecarToVersionedHash, dataColumnSidecarToVersionedHash, spec);
+    eventChannels.subscribe(
+        CustodyGroupCountChannel.class,
+        new CustodyGroupCountChannel() {
+          @Override
+          public void onGroupCountUpdate(int custodyGroupCount, int samplingGroupCount) {
```
tbenr: a question here: do we receive this event here quick enough even when the node is a supernode due to number of ETH attached and not just due to explicit custody param?

I'm worried if we are able to set the flag quick enough and avoid loosing versioned hashed

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
```diff
@@ -0,0 +1,89 @@
...
+          pair ->
+              MiscHelpersDeneb.required(spec.forMilestone(SpecMilestone.DENEB).miscHelpers())
+                  .kzgCommitmentToVersionedHash(
+                      pair.getKey()
+                          .getSszKZGCommitments()
+                          .get(pair.getValue().intValue())
+                          .getKZGCommitment());
+    } else {
+      this.dataColumnSidecarToVersionedHash = pair -> VersionedHash.EMPTY;
+    }
```
tbenr: I wander if we need to provide a noop version. Is there a situation where we call these ToVersionedHash functions and it is legit to have a NOOP one? would it be a bug?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSource.java
```diff
@@ -0,0 +1,319 @@
...
+
+  static final int SLOT_OFFSET = TYPE_SIZE;
+  static final int BLOCK_ROOT_OFFSET = SLOT_OFFSET + SLOT_SIZE;
+  static final int BLOB_INDEX_OFFSET = BLOCK_ROOT_OFFSET + BLOCK_ROOT_SIZE;
+  static final int DATA_SIZE = BLOB_INDEX_OFFSET + BLOB_INDEX_SIZE;
+
+  private final Function<BlobSidecar, VersionedHash> blobSidecarToVersionedHash;
+  private final Function<Pair<DataColumnSidecar, UInt64>, VersionedHash>
+      dataColumnSidecarToVersionedHash;
+  private final AtomicBoolean storeSidecarHashes = new AtomicBoolean(false);
```
tbenr, line 66: shouldn't we control the activation of the blobSidecar's versioned hashes storage too? not sure, we want to activate this for blobs by default...

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSource.java
```diff
@@ -0,0 +1,319 @@
...
+        Bytes.wrap(Longs.toByteArray(blobSidecar.getIndex().longValue())));
+  }
+
+  private Bytes computeDataColumnSidecarMetadata(
+      final DataColumnSidecar sidecar, final UInt64 blobIndex) {
+    return Bytes.concatenate(
+        Bytes.wrap(Longs.toByteArray(sidecar.getSlot().longValue())),
+        sidecar.getBlockRoot(),
+        Bytes.wrap(Longs.toByteArray(blobIndex.longValue())));
+  }
```
tbenr, line 309: thinking about is it is wise to use DataColumnSlotAndIdentifier here too so we make sure we save data that is by design matches the key we use to lookup blobs

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
zilm13: Yeah, I was thinking about this. We could have race here. We could try to pass supernode flag earlier at least.

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
zilm13: It's basically is fulu milestone supported. I don't understand where could it fail

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSource.java
zilm13: Yeah, blobSidecar is available by default.
But for dataColumnSidecars we need to have all sidecars (to be precise at least 64) to restore blob, so it makes sense there.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSource.java
zilm13: Semantically it will be more correct to use `SlotAndBlockRootAndBlobIndex` for both but it's not SSZ and doesn't have handy serialization/deserealization

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
tbenr: Don't we save how much we custody in a db var, so that it can go up only as per spec? I just remember this, maybe i'm wrong

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
tbenr: I mean we can provide the actual function without checking if supported. Cose if we end up calling those functions we are obviously in that milestone. If not is a bug (but in this version we will return empty instead of failing)

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSource.java
tbenr: But if you are not running an L2 you won't need that api and you wont need to store those indices. I thought that api call by versionedhashes was essentially optional

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
zilm13: Yeah, we save, but we cannot put db read in constructor (or can?). I mean it's slow

storage/src/main/java/tech/pegasys/teku/storage/server/VersionedHashDBSourceFactory.java
zilm13: Ahh yeah, you are right, until it's called it cannot fail, I will update the code

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSource.java
zilm13: It replaces `getBlobSidecars`, it's not in debug space https://github.com/ethereum/beacon-APIs/pull/546
So by spec we should support it by default from my understanding

zilm13:
Finished
Splitting because of the size.
First part is #9931

storage/src/test/java/tech/pegasys/teku/storage/server/ChainStorageTest.java
```diff
@@ -257,6 +261,52 @@ public void testOnFinalized(
...
+                          .stream()
+                          .map(
+                              blobSidecar ->
+                                  new SlotAndBlockRootAndBlobIndex(
+                                      blobSidecar.getSlot(),
+                                      blobSidecar.getBlockRoot(),
+                                      blobSidecar.getIndex()))
+                          .toList();
+                  if (!expectedBlockRootAndBlobIndices.isEmpty()) {
+                    System.out.println("got something");
```
rolfyone: shouldnt really be using system.out..

rolfyone: i think this was probably just debug code while writing that can be deleted?

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
```diff
@@ -0,0 +1,351 @@
...
+      new VersionedHashDBSource(dao, commitmentToVersionedHashFunction, spec);
+
+  @BeforeEach
+  public void setup() {
+    when(commitmentToVersionedHashFunction.apply(any())).thenReturn(versionedHash);
+    versionedHashDBSource = new VersionedHashDBSource(dao, commitmentToVersionedHashFunction, spec);
+  }
+
+  @Test
+  public void storeSidecarHashes_isFunctional() {
```
rolfyone: isFunctional is kinda redundant on all of these. i'd be happy with the first part i think

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
```diff
@@ -0,0 +1,351 @@
...
+                    .apply(dataColumnSidecar.getKzgCommitments().get(index).getKZGCommitment()));
+    verify(updater, times(dataColumnSidecar.getKzgCommitments().size()))
+        .addVersionedHash(eq(versionedHash.get()), any());
+
+    // stored only once, ignores next sidecars from the same block
+    versionedHashDBSource.addSidecarVersionedHashes(
+        dataStructureUtil.randomDataColumnSidecar(
+            DataColumnSidecarFulu.required(dataColumnSidecar).getSignedBlockHeader(),
+            dataColumnSidecar.getIndex().increment()),
+        updater);
```
rolfyone, line 188: presumably we should check this did what we thought...

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
```diff
@@ -0,0 +1,351 @@
...
+    IntStream.range(0, dataColumnSidecar.getKzgCommitments().size())
+        .forEach(
+            index ->
+                verify(commitmentToVersionedHashFunction)
+                    .apply(dataColumnSidecar.getKzgCommitments().get(index).getKZGCommitment()));
+    verify(updater, times(dataColumnSidecar.getKzgCommitments().size()))
+        .addVersionedHash(eq(versionedHash.get()), any());
+
+    // stored only once, ignores next sidecars from the same block
+    versionedHashDBSource.addSidecarVersionedHashes(
```
rolfyone: as above, we probably want to check this did its thing

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
```diff
@@ -0,0 +1,351 @@
...
+        VersionedHashDBSource.computeCanonicalDataColumnSidecarMetadata(
+            dataColumnSidecar.getSlotAndBlockRoot(), expectedBlobIndex);
+    when(dao.getSidecarIdentifierData(any())).thenReturn(Optional.of(data));
+
+    final Optional<SidecarIdentifier> maybeSidecarIdentifier =
+        versionedHashDBSource.getSidecarIdentifier(versionedHash);
+    verify(dao).getSidecarIdentifierData(any());
+    assertThat(maybeSidecarIdentifier.isPresent()).isTrue();
+    final SidecarIdentifier sidecarIdentifier = maybeSidecarIdentifier.get();
+    assertThat(sidecarIdentifier.canonical()).isTrue();
```
rolfyone: should be able to check the sidecarIdentifier is the same without checking each field?

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
```diff
@@ -0,0 +1,351 @@
...
+    assertThat(sidecarIdentifier.blobSidecarIdentifier()).isEmpty();
+    assertThat(sidecarIdentifier.dataColumnSidecarsIdentifierAndBlobIndex()).isPresent();
+    final Pair<SlotAndBlockRoot, UInt64> dataColumnSlotAndIdentifierAndBlobIndex =
+        sidecarIdentifier.dataColumnSidecarsIdentifierAndBlobIndex().get();
+    assertThat(dataColumnSlotAndIdentifierAndBlobIndex.getKey())
```
rolfyone: imo potentially check the whole `dataColumnAndIdentifierAndBlobIndex`  without checking each field....

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  @Test
+  public void addBlobSidecarVersionedHash_isFunctional() {
+    final KvStoreCombinedDao.FinalizedUpdater updater =
+        mock(KvStoreCombinedDao.FinalizedUpdater.class);
+    final BlobSidecar blobSidecar = dataStructureUtil.randomBlobSidecar();
+    versionedHashDBSource.addBlobSidecarVersionedHash(blobSidecar, updater);
+    verify(commitmentToVersionedHashFunction).apply(blobSidecar.getKZGCommitment());
+    verify(updater).addVersionedHash(eq(versionedHash.get()), any());
```
rolfyone, line 80: potentially with a lot of these could `verifyNoMoreInteractions`

storage/src/test/java/tech/pegasys/teku/storage/server/ChainStorageTest.java
zilm13: sure, removing

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
zilm13: The line checks no db calls, no hash calls   `verifyNoMoreInteractions(commitmentToVersionedHashFunction, updater);` Do you want to split it into 2 for clarity?

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
zilm13: same `verifyNoMoreInteractions` check as above

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
zilm13: updated, yeah, it looks way better

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
zilm13: updated, yeah, it looks way better

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
rolfyone: we've called `add` but we dont check something was added...

storage/src/test/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/VersionedHashDBSourceTest.java
```diff
@@ -0,0 +1,295 @@
...
+    verify(updater, times(dataColumnSidecar.getKzgCommitments().size()))
+        .addVersionedHash(eq(versionedHash.get()), any());
+    verify(dao).getNonCanonicalDataColumnIdentifiers(eq(dataColumnSidecar.getSlotAndBlockRoot()));
+
+    // stored only once, ignores next sidecars from the same block
+    versionedHashDBSource.addSidecarVersionedHashes(
+        dataStructureUtil.randomDataColumnSidecar(
+            DataColumnSidecarFulu.required(dataColumnSidecar).getSignedBlockHeader(),
+            dataColumnSidecar.getIndex().increment()),
+        updater);
```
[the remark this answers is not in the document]
zilm13: we have one table for versioned hashes, canonical or non-canonical is encoded in value

infrastructure/kzg/src/testFixtures/java/tech/pegasys/teku/kzg/NoOpKZG.java
```diff
@@ -66,7 +66,7 @@ public KZGProof computeBlobKzgProof(final Bytes blob, final KZGCommitment kzgCom
 
   @Override
   public KZGCommitment blobToKzgCommitment(final Bytes blob) throws KZGException {
-    return KZGCommitment.fromBytesCompressed(Bytes48.ZERO);
+    return KZGCommitment.fromBytesCompressed(Bytes48.wrap(blob.slice(0, Bytes48.SIZE)));
```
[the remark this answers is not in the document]
zilm13: Blob is 100kb always

rolfyone (approved):
LGTM

PR #9910, opened 2025-09-18
started a new implementation of the recovering retriever.

rolfyone:
enabled via `Xp2p-reworked-sidecar-recovery-enabled: true`

Timeouts drastically changed
 - the checking task is about 3x more common but shouldn't fall on slot boundary very regularly due to the 1/3 second offset.
 - the timeout is 2 minutes rather than 5, and it could potentially be significantly shorter still

primarily more separated and the lifecycle of the recovery process is arguably better defined.

Tried to keep the timeout to the entire process rather than have a timeout that is only part of the process also.

Currently most of the interesting `PendingRecoveryRequest` logging is at debug level, trace gets very noisy.

needs work still, but pushing draft to share the idea.

need to think about how we measure its behaviour through metrics, and what is useful to report on, as well as see the recovery process actually working.

## Documentation

## Changelog

---

> [!NOTE]
> Introduce a new sidecar recovery retriever with timeouts/metrics gated by `--Xp2p-reworked-sidecar-recovery-enabled`, and add start/stop lifecycle to retrievers with updated wiring.
> 
> - **Data column sidecar recovery (reworked)**:
>   - Add `recovering` impl: `recovering/SidecarRetriever`, `PendingRecoveryRequest`, `RebuildColumnsTask` (rebuild via KZG from DB, request timeouts, periodic checks, labelled metrics).
>   - Existing `RecoveringSidecarRetriever` now implements `start/stop`.
> - **API/Lifecycle**:
>   - Extend `DataColumnSidecarRetriever` with `start()`/`stop()`; implement in `SimpleSidecarRetriever` and test stubs (`DataColumnSidecarRetrieverStub`, `DelayedDataColumnSidecarRetriever`).
> - **Wiring**:
>   - `BeaconChainController`: choose between new `recovering/SidecarRetriever` and legacy `RecoveringSidecarRetriever` based on config; manage lifecycle via `DataColumnSidecarRetriever`.
> - **Config/CLI**:
>   - `P2PConfig`: add `reworkedSidecarRecoveryEnabled` (builder, getter).
>   - `P2POptions`: add `--Xp2p-reworked-sidecar-recovery-enabled` and plumb into `P2PConfig`.
> - **Tests**:
>   - Add unit tests for `PendingRecoveryRequest`, `RebuildColumnsTask`, `recovering/SidecarRetriever`; update fixtures to support new lifecycle and behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0f82f7cf0e1e03d35697744f8298776747686106. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/SidecarRetriever.java
```diff
@@ -0,0 +1,230 @@
...
+                      columnId,
+                      downloader.retrieve(columnId),
+                      timeProvider.getTimeInMillis(),
+                      recoveryTimeout,
+                      recoveryTimeout.dividedBy(2));
+              request.getFuture().always(() -> requests.remove(columnId));
+              request
+                  .getDownloadFuture()
+                  .finish((err) -> sidecarRecoveryMetric.labels(DOWNLOAD_TIMEOUT).inc());
+              request.getFuture().finish((err) -> sidecarRecoveryMetric.labels(CANCELLED).inc());
```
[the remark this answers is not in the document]
rolfyone: for each of their respective, but i think thats ok - if download fails, we get download_timeout, if the process fails completely we get cancelled, and there's success metrics for each too...

teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
```diff
@@ -602,8 +612,9 @@ public void configure(final TekuConfiguration.Builder builder) {
                   .gossipBlobsAfterBlockEnabled(gossipBlobsAfterBlockEnabled)
                   .dasExtraCustodyGroupCount(dasExtraCustodyGroupCount)
                   .historicalDataMaxConcurrentQueries(historicalDataMaxConcurrentQueries)
-                  .historicalDataMaxQueryQueueSize(historicalDataMaxQueryQueueSize)
-                  .executionProofTopicEnabled(executionProofTopicEnabled);
+                  .executionProofTopicEnabled(executionProofTopicEnabled)
+                  .reworkedSidecarRecoveryEnabled(reworkedSidecarRecoveryEnabled)
+                  .historicalDataMaxQueryQueueSize(historicalDataMaxQueryQueueSize);
```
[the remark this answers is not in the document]
rolfyone: builders dont (shouldnt) care about order but i did update it.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/SidecarRetriever.java
```diff
@@ -0,0 +1,222 @@
...
+import tech.pegasys.teku.infrastructure.time.TimeProvider;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.kzg.KZG;
+import tech.pegasys.teku.spec.datastructures.blobs.DataColumnSidecar;
+import tech.pegasys.teku.spec.datastructures.util.DataColumnSlotAndIdentifier;
+import tech.pegasys.teku.spec.logic.versions.fulu.helpers.MiscHelpersFulu;
+import tech.pegasys.teku.statetransition.datacolumns.db.DataColumnSidecarDbAccessor;
+import tech.pegasys.teku.statetransition.datacolumns.retriever.DataColumnSidecarRetriever;
+
+public class SidecarRetriever implements DataColumnSidecarRetriever {
```
lucassaldanha, line 39: NIT: We could call this `RecoveringSidecarRetrieverV2` to clarify that we expect it to supersede the previous one.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -355,8 +356,9 @@ public class BeaconChainController extends Service implements BeaconChainControl
   protected volatile BlobSidecarGossipValidator blobSidecarValidator;
   protected volatile DataColumnSidecarManager dataColumnSidecarManager;
   protected volatile Optional<DasCustodySync> dasCustodySync = Optional.empty();
-  protected volatile Optional<RecoveringSidecarRetriever> recoveringSidecarRetriever =
+  protected volatile Optional<DataColumnSidecarRetriever> recoveringSidecarRetriever =
       Optional.empty();
+  protected volatile Optional<SidecarRetriever> sidecarRetriever = Optional.empty();
```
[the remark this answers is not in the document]
lucassaldanha: We don't expect to populate it; we can actually just remove this field, given that we are updating `recoveringSidecarRetriever` with the instance that we will use in the system.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/SidecarRetriever.java
```diff
@@ -0,0 +1,222 @@
...
+
+    requests.values().forEach(PendingRecoveryRequest::cancel);
+    requests.clear();
+  }
+
+  @Override
+  @SuppressWarnings("FutureReturnValueIgnored")
+  public SafeFuture<DataColumnSidecar> retrieve(final DataColumnSlotAndIdentifier columnId) {
+    final PendingRecoveryRequest pendingRecoveryRequest =
+        requests.computeIfAbsent(
```
lucassaldanha, line 134: The `computeIfAbsent` method isn't guaranteed to be applied atomically. Documentation states that we shouldn't use it with a `mappingFunction` that has side effects; however, we are calling `downloader.retrieve(columnId)`, which creates a side effect.

One alternative option is to use `ConcurrentHashMap`, which has guarantees that `computeIfAbsent` is atomic and won't be executed more than once.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/RebuildColumnsTask.java
```diff
@@ -0,0 +1,236 @@
...
+          && sidecar.getSlotAndBlockRoot().equals(pendingRequest.getSlotAndBlockRoot())) {
+        LOG.debug(
+            "Pending request (slotAndBlock: {}) for column {} matched completed rebuild",
+            pendingRequest.getSlotAndBlockRoot(),
+            pendingRequest.getIndex());
+
+        pendingRequest.getFuture().complete(sidecar);
+        return true;
+      } else {
+        // TODO all that i could do at this point would be fail the task, or upstream it could be
```
lucassaldanha: NIT: This comment does not look like a TODO.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/RebuildColumnsTask.java
```diff
@@ -0,0 +1,236 @@
...
+      } else {
+        // we've already got the required data for this specific request, so can complete it
+        LOG.debug(
+            "Pending request (slotAndBlock: {}) for column {} was found while preparing to rebuild",
+            pendingRequest.getSlotAndBlockRoot(),
+            pendingRequest.getIndex());
+        pendingRequest.getFuture().complete(sidecar);
+      }
+    } else {
+      // TODO so we didnt match, we'd need to either fail the future or let it be handled upstream
```
lucassaldanha: NIT: Same as before, this comment does not look like a TODO.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/PendingRecoveryRequest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.tuweni.bytes.Bytes32;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
+import tech.pegasys.teku.spec.datastructures.blobs.DataColumnSidecar;
+import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
+import tech.pegasys.teku.spec.datastructures.util.DataColumnSlotAndIdentifier;
+
+class PendingRecoveryRequest {
+  private static final Logger LOG = LogManager.getLogger();
+  private final SafeFuture<DataColumnSidecar> future = new SafeFuture<>();
```
lucassaldanha, line 29: Should we "link" this future with the downloadFuture? 

For example, in `RebuildColumnTask`, we call `pendingRequest.getFuture().cancel(true)`.  Shouldn't that also cancel the download future?

I think that maybe, instead of exposing the future as getFuture(), we should expose methods to change it, but handle any internal consistency ourselves. For example: instead of using:
```
pendingRequest.getFuture().cancel(true);
```
One could call:
```
pendingRequest.cancel();
```

And internally in the `cancel()` method, we would handle everything we need for consistency. Similar thing for other calls like: 
- `pendingRequest.getFuture().complete(sidecar)` -> `pendingRequest.complete(sidecar)`
- `pendingRequest.getFuture().isDone()` -> `pendingRequest.isDone()`

The only other cases we would need to think about are when we are using things like `request.getFuture().always(..)` and `request.getFuture().finish(..)`.

Thoughts?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/RebuildColumnsTask.java
```diff
@@ -0,0 +1,236 @@
...
+      pendingRequest.getFuture().cancel(true);
+      return false;
+    }
```
lucassaldanha, line 107: We should never call `addTask` with a pendingRequest that isn't a match for the `SlotAndBlockRoot` in the `RebuildColumnsTask`.

We probably want to have this check as the first thing in the method, so we don't even go on the `done.get()` flow in this scenario.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/RebuildColumnsTask.java
```diff
@@ -0,0 +1,236 @@
...
+    } else {
+      // TODO so we didn't match, we'd need to either fail the future or let it be handled upstream
+      pendingRequest.getFuture().cancel(true);
+      return false;
+    }
+    checkQueryResult();
+    return true;
+  }
+
+  synchronized void checkQueryResult() {
```
lucassaldanha, line 112: Looks like we are only calling this on two circumstances:
1. When we have a new `PendingRecoveryTask`
2. In a schedule defined in `SidecarRetriever.checkPendingRequests()`.

I believe there is potential to check if reconstruction is possible every time a new data column is received. So we could call `checkQueryResult()` from within `SidecarRetriever.onNewValidatedSidecar(..)`.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/SidecarRetriever.java
rolfyone: im trying to keep it separate from the rest because i want to delete it

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/SidecarRetriever.java
rolfyone: ah i may have been keeping skiplist for something else but sure, can look at that.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/RebuildColumnsTask.java
rolfyone: hrm i think i do want to revisit that area when im in that class but i can remove it if you like

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/RebuildColumnsTask.java
rolfyone: yeh same answer, this is a logic gap pretty much i think

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/PendingRecoveryRequest.java
rolfyone: yeh i think better design is call the pendingRequest.cancel if we're cancelling.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/RebuildColumnsTask.java
rolfyone: i do think most of rebuildColumnsTask is 'not done' thats what i want for next PR.... i cant easily remove from here but at the same time its not ready.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/recovering/PendingRecoveryRequest.java
rolfyone: equally completing it i think, should be via fn so everything is done right

rolfyone: so i think maybe this is the first follow up PR...

rolfyone: at least one step down the raod here i think

lucassaldanha (approved):
Approved the PR as we are gonna keep working on the discussed points in follow-ups.

PR #9914, opened 2025-09-19
Remove earliest available block slot cache

tbenr:
The field become a DB variable, and the cache default config was just making sure there was only 1 inflight request at a time.

If we want to keep this class and have a meaningful impact, we should modify `DEFAULT_EARLIEST_AVAILABLE_BLOCK_SLOT_QUERY_FREQUENCY` to be 1

related to https://github.com/Consensys/teku/pull/9906#issuecomment-3310949410

[eab_queries.log](https://github.com/user-attachments/files/22425785/eab_queries.log)

running with:
```java
  public SafeFuture<Optional<UInt64>> getEarliestAvailableBlockSlot() {
    var start = System.nanoTime();
    return historicalChainData.getEarliestAvailableBlockSlot().thenPeek(__ ->
      LOG.info("Earliest available block {}ms", (System.nanoTime() - start) / 1_000_000));
  }
```

I think, looking at the stats, it is safe to remove.

The timing spikes are because the Splitter queue up all the reads until the DB write (happening on epoch transition) completes

## Documentation

## Changelog


zilm13 (approved):
LGTM
Cache looks useless for `StatusMessageFactory`, though `BeaconBlocksByRangeMessageHandler` could query it any number of  times per slot. Looks like not a big issue

PR #9922, opened 2025-09-22
fix: assert state root mismatch on root cause

Galoretka:
Update SanityBlocksTestExecutor to assert error messages on the root cause instead of the top-level RuntimeException. Spec.processBlock throws StateTransitionException which gets wrapped in RuntimeException by applyBlocks; asserting the wrapper’s message is unreliable (often null). Root cause assertions align with repository conventions and correctly detect the state root mismatch message emitted by AbstractBlockProcessor.


Galoretka:

lucassaldanha:
Thanks for taking the time to work on Teku @Galoretka. Unfortunately, your contribution at this time does not meet our relevancy requirements outlined in our contribution guidelines (see https://github.com/Consensys/teku/blob/master/CONTRIBUTING.md).

If you want to help build Teku, I suggest you start looking at our good first issues :)

PR #9995, opened 2025-10-10
Defer block import

### LINKED ISSUE #9488: Defer block import while block production is in flight
We could defer a late block import while a block production is in-flight.
We could "close the gate" at block production start, so we can avoid resource contentions while we are in the block production critical path

we already have the concept of in-flight block production in `ValidatorApiHandler::localBlockProductionBySlotCache`, which is used keep only one block production at a time, so multiple block production api calls will effectively trigger only one block production.

We could "export" this information on the block import side and use it to gate the block import process.

tbenr:
fixes #9488 

## Documentation

## Changelog


tbenr:
closing this for now

PR #10010, opened 2025-10-13
Refactor batch signature verifier

lucassaldanha:
## PR Description

TODO

## Fixed Issue(s)
N/A

## Documentation

## Changelog


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/AsyncBatchBLSSignatureVerifier.java
```diff
@@ -77,4 +79,10 @@ public SafeFuture<Boolean> verify(
       }
     };
   }
+
+  /** Creates non-async batch verifier to comply with the {@link BLSSignatureVerifier} interface. */
+  @Override
+  public BatchBLSSignatureVerifier createBatchVerifier() {
+    return new BatchSignatureVerifier();
+  }
```
lucassaldanha, line 87: Why is this needed?

ericsson49: I had to add it since `AsyncBatchBLSSignatureVerifier` implements `BLSSignatureVerifier` interface.
It's not used anywhere, so throwing `NotImplementedException` might be an option.

PR #10035, opened 2025-10-20
chore: use self hosted for larger runners to save $$$

waskow-consensys:


waskow-consensys:

joshuafernandes (approved):
Hi @waskow-consensys could we up the concurrency on these runners please? I had popped these jobs on latest-128 to speed up the entire build process by putting them in parallel

waskow-consensys:
> Hi @waskow-consensys could we up the concurrency on these runners please? I had popped these jobs on latest-128 to speed up the entire build process by putting them in parallel
Updated max runners to 40
Thanks!

PR #10037, opened 2025-10-21
das startup order fix

rolfyone:
## Documentation

## Changelog

---

> [!NOTE]
> Replaces Supplier/refs with a direct CustodyGroupCountManager, initializes it earlier, and updates DAS, networking, and tests to propagate custody counts and start components in the correct order.
> 
> - **DAS/Core (statetransition)**:
>   - Replace `Supplier<CustodyGroupCountManager>`/`AtomicReference` with direct `CustodyGroupCountManager` in `DasSamplerBasic`, `DasPreSampler`, `DataColumnSidecarCustodyImpl`, `DataColumnSidecarRecoveringCustodyImpl`, `SidecarRetriever`, `DataColumnSidecarELRecoveryManagerImpl`, and factories.
>   - Initialize `CustodyGroupCountManager` earlier in `BeaconChainController`; pass directly to DAS components (`DasSamplerBasic`, pre-sampler, custody, EL recovery) and `DataColumnSidecarSubnetBackboneSubscriber`.
>   - Improve `CustodyGroupCountManagerImpl` init/update: set `custodyGroupSyncedCount` early, atomically update count, mark synced on first set.
> - **Networking**:
>   - Subscribe `MetadataMessagesFactory` to `CustodyGroupCountChannel`; update metadata on synced custody count; remove explicit DAS custody count wiring in `ActiveEth2P2PNetwork`/builders.
>   - `DataColumnSidecarSubnetBackboneSubscriber` now queries manager for sampling group count (no channel callbacks).
> - **Tests/Fixtures**:
>   - Update all tests and test fixtures to use direct manager (no suppliers/refs); adjust expectations (e.g., metadata seq number, genesis synced count).
>   - Reference tests: construct `DasSamplerBasic` with `DasCustodyStand.createCustodyGroupCountManager(...)` directly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9ec507588841b30c2ed79c7d79b7db73b4bb1ae7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -1024,6 +1020,33 @@ protected void initDasSyncPreSampler() {
...
+
+      final MiscHelpersFulu miscHelpersFulu =
+          MiscHelpersFulu.required(spec.forMilestone(SpecMilestone.FULU).miscHelpers());
+
+      eventChannels
+          .getPublisher(CustodyGroupCountChannel.class)
+          .onGroupCountUpdate(
+              totalMyCustodyGroups, miscHelpersFulu.getSamplingGroupCount(totalMyCustodyGroups));
+    }
+  }
```
rolfyone: something went weird here for one

rolfyone:
confirmed we've got columns after startup with this solution (according to counts in `curl http://localhost:8008/metrics |grep data_column_sidecar`)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarCustodyImpl.java
```diff
@@ -111,7 +110,8 @@ public DataColumnSidecarCustodyImpl(
     this.blockResolver = blockResolver;
     this.minCustodyPeriodSlotCalculator = minCustodyPeriodSlotCalculator;
     this.custodyGroupCountManagerSupplier = custodyGroupCountManagerSupplier;
-    this.totalCustodyGroupCount = new AtomicInteger(totalCustodyGroupCount);
+    this.totalCustodyGroupCount =
+        new AtomicInteger(custodyGroupCountManagerSupplier.get().getCustodyGroupCount());
```
zilm13: if we use it in constructor it means we 100% sure the supplier is initialized as we call it immediately, so I'd pass the `CustodyGroupCountManager` instead of Supplier in constructor

tbenr:
So we did a debug session, together with @zilm13 
This is what we found:

1. the problem is not solved: `initDataColumnSidecarSubnetBackboneSubscriber` subscribes after `CustodyGroupCountManagerImpl` initialization and the first value dispatching. Thus, the subscriber never gets the correct initial value, unless the value changes. You can simulate the problem by pretending to get something different from DB (more than 8) and you'll see that the node will remain stuck on 8 subscription.
image

If you do the same trick with current master, you'll see that we will run with the correct value (12 in the example)


2. a minor thing here is that metadataMessagesFactory gets subscribed twice: in `Eth2P2PNetworkBuilder` and `BeaconChainController.initP2PNetwork()`

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarCustodyImpl.java
rolfyone: its more complicated, because we've got a reference that may be noop at start, but we can take the atomic reference (not significantly different to supplier so limited value)

rolfyone:
> So we did a debug session, together with @zilm13 This is what we found:
> 
> 1. the problem is not solved: `initDataColumnSidecarSubnetBackboneSubscriber` subscribes after `CustodyGroupCountManagerImpl` initialization and the first value dispatching. Thus, the subscriber never gets the correct initial value, unless the value changes. You can simulate the problem by pretending to get something different from DB (more than 8) and you'll see that the node will remain stuck on 8 subscription.
> 

discussed irl briefly but basically a bug in the setup of the object in that init function getting started with the wrong value.

tbenr (approved):
this works to me now

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
tbenr, line 1687: As a subsequent change I think we can remove this, because the subscription happens in the network builder

rolfyone: im not sure why for this bit so havent raised yet...

Issue #9976, opened 2025-10-07
Publishing column sidecars while syncing far from the head?


tbenr:
```{"@timestamp":"2025-10-07T13:51:48,001","level":"INFO","thread":"sync-async-0","class":"teku-event-log","message":"Syncing     *** Slot: 5328559, Head slot: 5328061, Remaining slots: 498, Target chain: 71a1cf..fff1 (5328480) with 25 peers, Connected peers: 37","throwable":""}
{"@timestamp":"2025-10-07T13:51:48,001","level":"INFO","thread":"sync-async-0","class":"teku-event-log","message":"Sync Info   *** Range: 5328054 - 5328178 (5 batches), Downloading: none, Ready: 125 slots (5 batches), Batch import: in progress","throwable":""}
{"@timestamp":"2025-10-07T13:52:00,002","level":"INFO","thread":"sync-async-0","class":"teku-event-log","message":"Syncing     *** Slot: 5328560, Head slot: 5328061, Remaining slots: 499, Target chain: 71a1cf..fff1 (5328480) with 27 peers, Connected peers: 39","throwable":""}
{"@timestamp":"2025-10-07T13:52:00,002","level":"INFO","thread":"sync-async-0","class":"teku-event-log","message":"Sync Info   *** Range: 5328054 - 5328178 (5 batches), Downloading: none, Ready: 125 slots (5 batches), Batch import: in progress","throwable":""}
{"@timestamp":"2025-10-07T13:52:12,002","level":"INFO","thread":"sync-async-0","class":"teku-event-log","message":"Syncing     *** Slot: 5328561, Head slot: 5328061, Remaining slots: 500, Target chain: 71a1cf..fff1 (5328480) with 29 peers, Connected peers: 39","throwable":""}
{"@timestamp":"2025-10-07T13:52:12,002","level":"INFO","thread":"sync-async-0","class":"teku-event-log","message":"Sync Info   *** Range: 5328054 - 5328178 (5 batches), Downloading: none, Ready: 125 slots (5 batches), Batch import: in progress","throwable":""}
{"@timestamp":"2025-10-07T13:52:24,001","level":"INFO","thread":"sync-async-0","class":"teku-event-log","message":"Syncing     *** Slot: 5328562, Head slot: 5328061, Remaining slots: 501, Target chain: 71a1cf..fff1 (5328480) with 29 peers, Connected peers: 39","throwable":""}
[...]

PR #10086, opened 2025-10-31
Make RPC fetch delay logic reusable

tbenr:
Move the RPC fetch delay calculation outside `BlockBlobSidecarsTrackersPoolImpl` to be later reused in DAS Sampler.

relater to #9976

## Documentation

## Changelog

---

> [!NOTE]
> Externalizes RPC fetch delay logic into `RPCFetchDelayProvider` and integrates it with `BlockBlobSidecarsTrackersPoolImpl` via `PoolFactory`, with corresponding test updates.
> 
> - **Util**:
>   - Introduce `RPCFetchDelayProvider` with `create(...)`, `NO_DELAY`, and default timing constants; exposes `Duration calculate(UInt64 slot)`.
> - **Blob Sidecars Trackers Pool** (`BlockBlobSidecarsTrackersPoolImpl`):
>   - Remove internal delay constants and `TimeProvider` dependency; add `RPCFetchDelayProvider` field/ctor param.
>   - Use `rpcFetchDelayProvider.calculate(slot)` in `onFirstSeen` to delay RPC fetching.
> - **Factory Wiring** (`PoolFactory`):
>   - Construct `RPCFetchDelayProvider` using `spec`, `timeProvider`, `recentChainData`, and `CurrentSlotProvider` with default timings.
>   - Pass the provider into new pool constructors (including test-only factory).
> - **Tests**:
>   - Add `RPCFetchDelayProviderTest` covering delay calculation scenarios.
>   - Update `BlockBlobSidecarsTrackersPoolImplTest` to mock/use `RPCFetchDelayProvider` and adjusted async/time stubs; remove in-class delay calc tests; adapt delayed action assertions.
> - **Misc**:
>   - Minor cleanup in `BeaconChainController` (no functional change).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 35c2a22f453133da3680cc897ec8709dd6edb6d3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/RPCFetchDelayProvider.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/RPCFetchDelayProvider.java
@@ -41,8 +41,11 @@ public interface RPCFetchDelayProvider {
       final UInt64 targetWaitMillis) {
 
     return (slot) -> {
+      // this logic is supposed to run on "firstSeen" block\dataSidecar.
+      // The slot param is the block\dataSidecar's slot.
+
       if (slot.isLessThan(currentSlotProvider.getCurrentSlot())) {
-        // old slot
+        // old slot, we are in syncing\backfilling, we don't need to wait
         return Duration.ZERO;
       }
 
@@ -52,10 +55,16 @@ public interface RPCFetchDelayProvider {
           slotStartTimeMillis.plus(spec.getAttestationDueMillis(slot));
 
       if (nowMillis.isGreaterThanOrEqualTo(attestationDueMillis)) {
-        // late block, we already produced attestations on previous head,
+        // late block\sidecar, we already produced attestations voting for previous head,
         // so let's wait our target delay before trying to fetch
         return Duration.ofMillis(targetWaitMillis.intValue());
       }
+
+      // The idea is to wait the target delay (targetWaitMillis) if we are far
+      // enough from our deadline (maxWaitRelativeToAttDueMillis).
+      // If there is not enough time left, let's reduce the target delay so that
+      // distance from our deadline remains unchanged.
+      // The target delay reduction is capped so that we guarantee to wait at least "minWaitMillis".
 
       final UInt64 upperLimitRelativeToAttDue =
           attestationDueMillis.minus(maxWaitRelativeToAttDueMillis);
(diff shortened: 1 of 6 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/RPCFetchDelayProvider.java
```diff
@@ -0,0 +1,73 @@
...
+
+      final UInt64 nowMillis = timeProvider.getTimeInMillis();
+      final UInt64 slotStartTimeMillis = secondsToMillis(recentChainData.computeTimeAtSlot(slot));
+      final UInt64 attestationDueMillis =
+          slotStartTimeMillis.plus(spec.getAttestationDueMillis(slot));
+
+      if (nowMillis.isGreaterThanOrEqualTo(attestationDueMillis)) {
+        // late block, we already produced attestations on previous head,
+        // so let's wait our target delay before trying to fetch
+        return Duration.ofMillis(targetWaitMillis.intValue());
```
zilm13, line 60: I understand that it's old logic, but it's a bit confusing that we could have smaller delay being in time (`minWaitMillis`), than being left. It would be good to have some descriptive explanation to this logic. From the first view if we are late I expect 0 delay.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/util/RPCFetchDelayProviderTest.java
```diff
@@ -0,0 +1,140 @@
...
+    final UInt64 startSlotInMillis = startSlotInSeconds.times(1_000);
+
+    when(recentChainData.computeTimeAtSlot(currentSlot)).thenReturn(startSlotInSeconds);
+
+    // blocks arrives 200ms before attestation due
+    timeProvider.advanceTimeByMillis(startSlotInMillis.plus(3_800).longValue());
+
+    final Duration fetchDelay = rpcFetchDelayProvider.calculate(currentSlot);
+
+    // we can wait the full target
```
zilm13: *we can't wait full target, probably?

zilm13 (approved):
LGTM

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/util/RPCFetchDelayProvider.java
tbenr: The logic on why here we wait at least 500ms (when importing for the current slot) is because this delay is applied "on first seen" so the idea is to give anyway some time to receive the rest of the data which might be inflight via gossip.
Thinking twice I am not sure the way i reused this in dasSampler is correct, because it is not applied onFirst seen. I'll check in monday.

PR #10094, opened 2025-11-03
Apply RPC fetch delay to fulu

tbenr:
It applies the RPC fetch delay logic to Das Sampler

related to https://github.com/Consensys/teku/issues/9976

## Documentation

## Changelog

---

> [!NOTE]
> Adds async, configurable RPC fetch delay to DAS data column sampling and triggers availability checks immediately when required, with full wiring and tests.
> 
> - **DAS Sampling (`DasSamplerBasic`)**:
>   - Introduce `AsyncRunner` and `RPCFetchDelayProvider` to schedule RPC fetches for missing columns with a computed delay; immediate fetch when delay is zero.
>   - Defer scheduling until first-seen, and guard via `rpcFetchScheduled` flag in `DataColumnSamplingTracker`.
>   - Fetch missing columns via RPC and forward newly retrieved columns to custody only if still missing.
> - **Tracker (`DataColumnSamplingTracker`)**:
>   - Add `rpcFetchScheduled` state and expose it alongside existing tracking/completion.
> - **Availability Checker**:
>   - Start DAS `checkDataAvailability` during `initiateDataAvailabilityCheck` for REQUIRED cases and flush immediately; simplify `getAvailabilityCheckResult`.
> - **Wiring**:
>   - In `BeaconChainController`, create and pass `RPCFetchDelayProvider` and `beaconAsyncRunner` to `DasSamplerBasic`.
>   - Reference tests (`ForkChoiceTestExecutor`) instantiate an async runner and use `RPCFetchDelayProvider.NO_DELAY` for deterministic behavior.
> - **Tests**:
>   - Expand unit tests to cover delayed vs immediate RPC fetch scheduling, tracker state, and availability checker flow; adjust assertions accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 05a8363776429322d7b6f279566d1f9edcb909ae. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasic.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasic.java
@@ -18,7 +18,6 @@ import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
@@ -160,20 +159,14 @@ public class DasSamplerBasic implements DataAvailabilitySampler, SlotEventsChann
   }
 
   private DataColumnSamplingTracker getOrCreateTracker(final UInt64 slot, final Bytes32 blockRoot) {
-    final AtomicBoolean firstSeen = new AtomicBoolean(false);
-    final DataColumnSamplingTracker tracker =
-        recentlySampledColumnsByRoot.computeIfAbsent(
-            blockRoot,
-            k -> {
-              firstSeen.set(true);
-              return DataColumnSamplingTracker.create(slot, blockRoot, custodyGroupCountManager);
-            });
-
-    if (firstSeen.get()) {
-      onFirstSeen(slot, blockRoot, tracker);
-    }
-
-    return tracker;
+    return recentlySampledColumnsByRoot.computeIfAbsent(
+        blockRoot,
+        k -> {
+          final DataColumnSamplingTracker tracker =
+              DataColumnSamplingTracker.create(slot, blockRoot, custodyGroupCountManager);
+          onFirstSeen(slot, blockRoot, tracker);
+          return tracker;
+        });
   }
 
   private SafeFuture<DataColumnSidecar> retrieveColumnWithSamplingAndCustody(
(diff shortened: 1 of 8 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasic.java
```diff
@@ -91,6 +101,31 @@ public SafeFuture<List<UInt64>> checkDataAvailability(
...
+      final UInt64 slot, final Bytes32 blockRoot, final DataColumnSamplingTracker tracker) {
+    final Duration delay = rpcFetchDelayProvider.calculate(slot);
+    if (delay.isZero()) {
+      // in case of immediate RPC fetch, let's postpone the actual fetch when checkDataAvailability
+      // is called.
+      // this is needed because 0 delay means we are syncing\backfilling this slot, so we want to
+      // wait eventual known columns to be added via onAlreadyKnownDataColumn before fetching.
+      return;
+    }
+    tracker.rpcFetchScheduled().set(true);
```
zilm13, line 121: It sill can be raced. Assume `calling checkDataAvailability` at the same time, also computeIfAbsent function could be called twice+ on concurrency. I understand your goal, why you skipped if `compareAndSet` here, but I think it's not a good way to achieve this

tbenr: Ye you're right, I had a wrong interpretation of the computeIfAbsent javadoc. I'll fix it, good catch!!

tbenr: no, I got hallucinated. with java 9+ the `computeIfAbsent` function is guaranteed to be called only once

zilm13: after discussion we decided that I'm not right:
`computeIfAbsent` is guranteed to be blocked, see JavaDoc:
```
If the specified key is not already associated with a value, attempts to compute its value using the given mapping function and enters it into this map unless null. The entire method invocation is performed atomically. The supplied function is invoked exactly once per invocation of this method if the key is absent, else not at all. Some attempted update operations on this map by other threads may be blocked while computation is in progress
```
`checkDataAvailability` could not go next until get created key which includes set true, so it's safe too.

So we could close this comment

zilm13 (approved):

Issue #10124, opened 2025-11-11
fulu proposer scheduling


rolfyone:
because proposers are known for 2 epochs in fulu, we can now schedule the duties for the future epoch proposals, without waiting for epoch transition.

This will help reduce the impacts of epoch transition etc on the first proposer of the epoch. We should prioritise this one imo.

PR #10162, opened 2025-11-24
Updated BeaconStateAccessorsFulu for proposer lookahead

rolfyone:
Allows epoch duties to be fetched from the next epoch, which is contained in the proposer_lookahead for the state, we just need to return it.

also extended the number of validators in the test suite so that it's a better example.

partially addresses #10124

## Documentation

## Changelog

---

> [!NOTE]
> Allow fetching beacon proposer indices for the current or next epoch via proposer_lookahead with tests updated and a builder overload added.
> 
> - **Fulu spec logic**:
>   - `getBeaconProposerIndex`: compute `lookaheadIndex` with epoch offset to support current/next epoch; add validation to restrict to current or next epoch; add debug logging.
> - **Tests**:
>   - Update expectations for current-epoch proposer index and add next-epoch coverage by iterating `proposer_lookahead`.
>   - Adjust failure case to only allow current/next epoch; minor imports updated.
>   - Use `InMemoryStorageSystemBuilder.buildDefault(16, spec)` to increase validator count.
> - **Test fixtures**:
>   - Add `InMemoryStorageSystemBuilder.buildDefault(int numberOfValidators, Spec)` overload to configure validator count.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b0bb82f53cc5265729e905127e133d2919f4acbc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/BeaconStateAccessorsFulu.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/fulu/helpers/BeaconStateAccessorsFulu.java
@@ -17,6 +17,8 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import it.unimi.dsi.fastutil.ints.IntList;
 import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
 import tech.pegasys.teku.spec.config.SpecConfig;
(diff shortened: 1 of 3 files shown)


StefanBratanov (approved):
Makes sense to me. I remember we had this discussion before, specs are slightly different than what we use in the code since it uses the `state.slot` and we use a slot as a parameter https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/beacon-chain.md#modified-get_beacon_proposer_index

lucassaldanha (approved):

StefanBratanov:
@rolfyone I was also going to suggest to override this in `BeaconStateAccessorsFulu` since it calls `getBeaconProposerIndex(state, state.getSlot)` and does some useless checks and additional not needed computation.

```java
@Override
public int getBeaconProposerIndex(final BeaconState state) {
   return BeaconStateFulu.required(state)
       .getProposerLookahead()
       .get(state.getSlot().mod(config.getSlotsPerEpoch()).intValue())
       .get()
       .intValue();
  }
```

Issue #10185, opened 2025-11-28
Sync may stall if SimpleSidecarRetriever's RPC calls fail


tbenr:
```
2025-11-28 14:47:30.988 DEBUG - ReqResp outbound data_column_sidecars_by_root, columns: 0/1 in 1900 ms, peer 0xf7628e30...c1ef83e9: request: 8 columns: 0x0025cc167bfbf651b62c605d9d1bcafab09502bdcd8b6c404d0ae8c1b01c47c6 colIdxs: (len: 8) [24,28,52,61,71,73,90,106], response: <empty>
2025-11-28 14:47:30.989 DEBUG - SimpleSidecarRetriever.Request failed for DataColumnSlotAndIdentifier[slot=1843631, blockRoot=0x0025cc167bfbf651b62c605d9d1bcafab09502bdcd8b6c404d0ae8c1b01c47c6, columnIndex=28] due to: DasColumnNotAvailableException
2025-11-28 14:47:30.989 DEBUG - SimpleSidecarRetriever.Request failed for DataColumnSlotAndIdentifier[slot=1843631, blockRoot=0x0025cc167bfbf651b62c605d9d1bcafab09502bdcd8b6c404d0ae8c1b01c47c6, columnIndex=24] due to: DasColumnNotAvailableException
2025-11-28 14:47:30.989 DEBUG - SimpleSidecarRetriever.Request failed for DataColumnSlotAndIdentifier[slot=1843631, blockRoot=0x0025cc167bfbf651b62c605d9d1bcafab09502bdcd8b6c404d0ae8c1b01c47c6, columnIndex=61] due to: DasColumnNotAvailableException
2025-11-28 14:47:30.989 DEBUG - SimpleSidecarRetriever.Request failed for DataColumnSlotAndIdentifier[slot=1843631, blockRoot=0x0025cc167bfbf651b62c605d9d1bcafab09502bdcd8b6c404d0ae8c1b01c47c6, columnIndex=52] due to: DasColumnNotAvailableException
2025-11-28 14:47:30.989 DEBUG - SimpleSidecarRetriever.Request failed for DataColumnSlotAndIdentifier[slot=1843631, blockRoot=0x0025cc167bfbf651b62c605d9d1bcafab09502bdcd8b6c404d0ae8c1b01c47c6, columnIndex=73] due to: DasColumnNotAvailableException
[...]

PR #10193, opened 2025-12-01
Allow sampler to reissue RPC requests

tbenr:
If RPC requests issued by the sampler fail (due to timeout or other problems) the sampler stalls and never retries downloads.

This can stall the sync itself, because even in the case `SyncStallDetector` restarts the sync, the sampler will always have the same trackers with RPCFetched marked as `true`. Thus no more attempts on restart.

With this change, multiple calls to `checkDataAvailability` can cause redownloads attempt on incomplete trackers.

related to: #10185

NOTE: A limitation of the current sync design is that the sync has no feedback from the PreSampler, so the only way it has to attempt a recovery is via `SyncStallDetector`, which requires a relatively long timeout.

## Documentation

## Changelog

---

> [!NOTE]
> Allows `DasSamplerBasic` to retry RPC data fetches by resetting the scheduled flag and short-circuiting completed trackers; updates tests to cover retry and scheduling behavior.
> 
> - **Statetransition (DAS Sampler)**
>   - In `checkDataAvailability(...)`:
>     - Return early if `completionFuture` is already done.
>     - Atomically schedule RPC fetch with `rpcFetchScheduled().compareAndSet(false, true)`.
>   - In `fetchMissingColumnsViaRPC(...)`:
>     - After completion (success or failure), reset `rpcFetchScheduled` via `.alwaysRun(() -> ...set(false))` to allow retries.
> - **Tests**
>   - Add coverage for retry on failure and scheduling reset/behavior.
>   - Introduce helper to assert `rpcFetchScheduled` state; adjust expectations for zero/non-zero delays.
>   - Minor refactor to map construction for sidecars.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a9950140251d2a64a1e0a5747ccfe8d0200c4ddb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasic.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasic.java
@@ -96,6 +96,10 @@ public class DasSamplerBasic implements DataAvailabilitySampler, SlotEventsChann
       final UInt64 slot, final Bytes32 blockRoot) {
     final DataColumnSamplingTracker tracker = getOrCreateTracker(slot, blockRoot);
 
+    if (tracker.completionFuture().isDone()) {
+      return tracker.completionFuture();
+    }
+
     if (tracker.rpcFetchScheduled().compareAndSet(false, true)) {
       fetchMissingColumnsViaRPC(slot, blockRoot, tracker);
     }
@@ -150,6 +154,9 @@ public class DasSamplerBasic implements DataAvailabilitySampler, SlotEventsChann
                         tracker.samplingRequirement().size()));
               }
             })
+        // let's reset the fetched flag so that this tracker can reissue RPC requests on DA check
+        // retry
+        .alwaysRun(() -> tracker.rpcFetchScheduled().set(false))
         .finish(
             throwable -> {
               if (ExceptionUtil.hasCause(throwable, CancellationException.class)) {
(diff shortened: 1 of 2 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasic.java
```diff
@@ -150,6 +154,9 @@ private void fetchMissingColumnsViaRPC(
                         tracker.samplingRequirement().size()));
               }
             })
+        // let's reset the fetched flag so that this tracker can reissue RPC requests on DA check
+        // retry
+        .alwaysRun(() -> tracker.rpcFetchScheduled().set(false))
```
rolfyone, line 159: not sure we want this as `alwaysRun`? why don't we just do this inside the exception?

i guess this is where life might get messy if its conditional with retry limiters etc and this is maybe adequate...

tbenr: I see what you mean, but i thought is logically correct to flag that RPCs are no more scheduled when they are done, and I simply skipped the whole check on here:
https://github.com/Consensys/teku/pull/10193/files#diff-9c559eda47436a03f2b08d25a3c0fc065004df537812620d4609d978df7fcd65R99-R102
if the tracker completion future has already done (success\failure)

zilm13 (approved):
LGTM

rolfyone (approved):
LGTM

PR #10197, opened 2025-12-02
Run `kzg.loadTrustedSetup` in thread with 8MB stack size

jtraglia:
## PR Description

This PR runs `kzg.loadTrustedSetup()` in a separate, temporary, short-lived thread with an 8MB stack.

This was segfaulting with higher precompute values (greater than 10) because the default stack size for java threads is 1MB. It worked just fine in c-kzg-4844 testing because the tests ran on the main thread which is 8MB.

To test this out, you can run the following command before & after the change:

```
./build/install/teku/bin/teku --Xkzg-precompute=15 --ee-endpoint=http://blah --checkpoint-sync-url=https://mainnet.checkpoint.sigp.io
```

You can also test this by setting all threads to be 8MB 😅 

```
JAVA_OPTS="-Xss8m" ./build/install/teku/bin/teku --Xkzg-precompute=15 ...
```

## Fixed Issue(s)

* https://github.com/ethereum/c-kzg-4844/issues/611

## Documentation

## Changelog

---

> [!NOTE]
> Loads `kzg.loadTrustedSetup` on a dedicated 8MB-stack thread when `kzgPrecompute` exceeds `DEFAULT_KZG_PRECOMPUTE_SUPERNODE`, with timeout and error handling; otherwise uses current thread.
> 
> - **KZG trusted setup loading**:
>   - Conditionally executes `kzg.loadTrustedSetup(trustedSetupFile, kzgPrecompute)` on a dedicated thread with 8MB stack when `kzgPrecompute` > `DEFAULT_KZG_PRECOMPUTE_SUPERNODE`.
>   - Adds 5-minute join timeout, interruption handling, and exception propagation from the loader thread.
>   - Retains current-thread execution (1MB stack) for lower/equal precompute values.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cc197e1960cf3d3082915d61fcb4e245b3926444. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
+++ b/services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
@@ -769,30 +769,37 @@ public class BeaconChainController extends Service implements BeaconChainControl
                     return DEFAULT_KZG_PRECOMPUTE;
                   });
 
-      // Run loadTrustedSetup on a dedicated thread with a larger stack size.
-      final long stackSize = 8 * 1024 * 1024; // 8MB
-      final AtomicReference<Throwable> loadException = new AtomicReference<>();
-      final Thread loaderThread =
-          new Thread(
-              null,
-              () -> {
-                try {
-                  kzg.loadTrustedSetup(trustedSetupFile, kzgPrecompute);
-                } catch (final Throwable t) {
-                  loadException.set(t);
-                }
-              },
-              "kzg-loader",
-              stackSize);
-      loaderThread.start();
-      try {
-        loaderThread.join();
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        throw new RuntimeException("KZG trusted setup loading was interrupted", e);
-      }
-      if (loadException.get() != null) {
-        throw new RuntimeException("Failed to load KZG trusted setup", loadException.get());
+      if (kzgPrecompute <= DEFAULT_KZG_PRECOMPUTE_SUPERNODE) {
+        // By default, run loadTrustedSetup on the current thread with a 1MB stack
+        kzg.loadTrustedSetup(trustedSetupFile, kzgPrecompute);
+      } else {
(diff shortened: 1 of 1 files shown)


services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -769,7 +769,31 @@ protected void initKzg() {
...
+                  kzg.loadTrustedSetup(trustedSetupFile, kzgPrecompute);
+                } catch (final Throwable t) {
+                  loadException.set(t);
+                }
+              },
+              "kzg-loader",
+              stackSize);
+      loaderThread.start();
+      try {
+        loaderThread.join();
```
rolfyone: thoughts on maybe some kind of sane timeout here? even if its a longish one, just not unbound...

jtraglia: Would 5 or 10 minutes be too long? It takes 75 seconds to load `precompute=15` on my system.

StefanBratanov:
I think we should make this logic run only if precompute is > 10. Otherwise, the default now is 9 and the CLI option is hidden, so it's a scenario which wouldn't happen in practice, unless some knowledgeable user runs it.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
rolfyone: yeh 5 minutes or something like that should be fine - its just so that its not unbound...

jtraglia:
Hey @rolfyone & @StefanBratanov, I've updated the PR to use a five minute deadline for initializing the library and will only use the custom thread if the `kzgPrecompute` value is greater than the default supernode value (9).

rolfyone (approved):
LGTM - thanks @jtraglia !

tbenr:
so apparently we are good with this. yoloing it

PR #10209, opened 2025-12-08
Fix `BlobSidecarSelectorFactory` for Gloas

StefanBratanov:
## PR Description
In Gloas, there are no blob kzg commitments, for simplicity, we can still reuse the same logic as now and query the chain data for blob sidecars. The only thing we lose after Gloas, is checking for emptiness. But the chain data retrieval would take care of that anyways. The alternative is querying the execution payload or checking against the root of empty blob kzg commitments list based on the slot. Both seem more inefficient.

Also added some more nits + correct data structure util for gloas when stating empty commitments.

## Fixed Issue(s)
related to #9833 

## Documentation

## Changelog

---

> [!NOTE]
> Updates blob sidecar selection to work without in-block commitments (Gloas/ePBS) and replaces gossip slot check with a new isSlotCurrent; also renames payload attestation cache size constant and adds execution payload lookup by block root.
> 
> - **Blob Sidecars (provider)**:
>   - Update `BlobSidecarSelectorFactory` to handle Deneb/Gloas blocks without in-body blob commitments:
>     - Return empty list when commitments are explicitly empty, otherwise fetch sidecars; if commitments are only referenced by root (ePBS), query chain data directly.
> - **Gossip validation (statetransition)**:
>   - Replace `isCurrentSlotWithGossipDisparityAllowance` with `isSlotCurrent` in `GossipValidationHelper` and use it in payload attestation and sync-committee validators.
>   - Adjust visibility/fields in `PayloadAttestationMessageGossipValidator` and use new cache size constant.
> - **Constants**:
>   - Rename `RECENT_SEEN_PAYLOAD_ATTESTATIONS_CACHE_SIZE` to `VALID_PAYLOAD_ATTESTATION_SET_SIZE`.
> - **Data generation utilities (tests)**:
>   - Extend `DataStructureUtil` to support (ePBS) commitments via `signedExecutionPayloadBid` helpers and guard KZG fields; remove an unused vector overload.
> - **Storage**:
>   - Add `CombinedChainDataClient#getExecutionPayloadByBlockRoot` (recent-only, historical TODO).
> - **Tests**:
>   - Update/enable tests across modules to reflect new slot check, constant rename, and Gloas handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0226c2728606c3d9c9619de2fe95d46caba96a64. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
+++ b/data/provider/src/main/java/tech/pegasys/teku/api/blobselector/BlobSidecarSelectorFactory.java
@@ -27,7 +27,6 @@ import tech.pegasys.teku.spec.SpecMilestone;
 import tech.pegasys.teku.spec.datastructures.blobs.versions.deneb.BlobSidecar;
 import tech.pegasys.teku.spec.datastructures.blocks.SignedBeaconBlock;
 import tech.pegasys.teku.spec.datastructures.blocks.SlotAndBlockRoot;
-import tech.pegasys.teku.spec.datastructures.blocks.blockbody.versions.deneb.BeaconBlockBodyDeneb;
 import tech.pegasys.teku.spec.datastructures.metadata.BlobSidecarsAndMetaData;
 import tech.pegasys.teku.storage.client.BlobSidecarReconstructionProvider;
 import tech.pegasys.teku.storage.client.ChainHead;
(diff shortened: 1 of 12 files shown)


mehdi-aouadi (approved):
LGTM

PR #10210, opened 2025-12-09
Fix TOCTOU race condition for duplicate attestations

lucassaldanha:
## PR Description

When validating attestations, the node should not propagate an attestation that was previously seen. In our current implementation, the check if the attestation was seen and the resulting validation was not done atomically, which could lead to a race condition where we check two similar attestations but don't consider them seen.

This fix relies on our synchronised `seenPayloadAttestations` set `add()` function to perform a second check if the attestation has been seen.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

---

> [!NOTE]
> Makes duplicate payload attestation detection atomic in the gossip validator and adds tests to cover the race scenario.
> 
> - **Validation (PayloadAttestationMessageGossipValidator)**:
>   - Atomically handle duplicates by checking `seenPayloadAttestations.add(key)` post-validation; return `ignore(...)` if already present.
>   - Replace pre-check return with `ignoreAttestationAlreadySeenValidationResult(...)` helper and reuse it after `add(...)`.
>   - Add private helper `ignoreAttestationAlreadySeenValidationResult(...)` to centralize logging and ignore result.
> - **Tests**:
>   - Add `shouldIgnore_whenAlreadySeen_AfterInitialCheck` to simulate concurrent validations and verify ignore-on-second.
>   - Tighten existing tests to verify no further validations on duplicates and behavior when validation fails then succeeds.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9c7b9460b3e90df26a6d436331addb6b0e750ba6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,4 +17,3 @@
 ### Bug Fixes
 - Fixed a storage issue which sometimes caused Teku to crash during shut down.
 - Fixed `peer_count` metric when using `--metrics-publish-endpoint` feature. 
-- Fixed attestation validation checks to avoid broadcasting attestations already seen.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,3 +17,4 @@
 ### Bug Fixes
 - Fixed a storage issue which sometimes caused Teku to crash during shut down.
 - Fixed `peer_count` metric when using `--metrics-publish-endpoint` feature. 
+- Fixed attestation validation checks to avoid broadcasting attestations already seen.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,3 +17,4 @@
 ### Bug Fixes
 - Fixed a storage issue which sometimes caused Teku to crash during shut down.
 - Fixed `peer_count` metric when using `--metrics-publish-endpoint` feature. 
+- Fixed attestation validation checks to avoid broadcasting attestations already seen.
```
rolfyone: we can remove this, the fix is gloas only and for payload attestations (PTC) not attestations.

rolfyone (approved):
LGTM just remove changelog imo.

PR #10220, opened 2025-12-10
Fix earliest_available_slot calculation

lucassaldanha:
## PR Description

Take into consideration both the earliest block and the custody-fulfilled datacolumns when calculation `earliest_available_slot`.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

---

> [!NOTE]
> Corrects earliest_available_slot by combining earliest block and data-column sidecar slots, gated to Fulu, with tests and changelog updates.
> 
> - **Networking/eth2**:
>   - `StatusMessageFactory`: compute `earliest_available_slot` by combining `getEarliestAvailableBlockSlot` and `getEarliestDataColumnSidecarSlot` (take max) via async `thenCombine`.
>   - Update value only at epoch start and only when Fulu is supported; store in `maybeEarliestAvailableSlot` and expose via `beacon.earliest_available_slot` gauge.
> - **Tests**:
>   - Add parameterized scenarios covering combinations of earliest block/data-column slots and pre-Fulu behavior; update existing tests to include data-column sidecar.
> - **Changelog**:
>   - Add bug fix entry for `earliest_available_slot` and tidy `peer_count` entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 554be8e3f56dcceeab0b6f78605eb66ad72ec497. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
@@ -15,12 +15,14 @@ package tech.pegasys.teku.networking.eth2.rpc.beaconchain.methods;
 
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tuweni.bytes.Bytes32;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import tech.pegasys.teku.ethereum.events.SlotEventsChannel;
+import tech.pegasys.teku.infrastructure.async.SafeFuture;
 import tech.pegasys.teku.infrastructure.bytes.Bytes4;
 import tech.pegasys.teku.infrastructure.metrics.TekuMetricCategory;
 import tech.pegasys.teku.infrastructure.unsigned.UInt64;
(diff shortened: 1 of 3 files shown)


tbenr:
that seems correct for but we need to apply the combined logic only for FULU

lucassaldanha:
> that seems correct for but we need to apply the combined logic only for FULU

After discussing this with Paul, we decided to make it conditional on Fulu being scheduled, which implies support for StatusV2.

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/StatusMessageFactory.java
```diff
@@ -111,14 +116,33 @@ public Optional<StatusMessage> createStatusMessage(final StatusMessageSchema<?>
...
+        combinedChainDataClient.getEarliestAvailableBlockSlot();
+    final SafeFuture<Optional<UInt64>> earliestDataColumnSidecarSlotFuture =
+        combinedChainDataClient.getEarliestDataColumnSidecarSlot();
+
+    final BiFunction<Optional<UInt64>, Optional<UInt64>, Optional<UInt64>>
+        combineEarliestSlotFunction =
+            (blockEarliestAvailableSlot, dataColumnEarliestAvailableSlot) -> {
+              if (blockEarliestAvailableSlot.isPresent()
+                  && dataColumnEarliestAvailableSlot.isPresent()) {
+                return Optional.ofNullable(
```
tbenr, line 137: I don't feel this could be null (if so, we NPE anyway during calculation), so we can use `Optional.of`

tbenr (approved):
LGTM

PR #10219, opened 2025-12-10
Fix early tracker removal

zilm13:
## PR Description
Also fixes duplicated sidecars downloading and storing.
It's a quick fix, some refactoring will go next.

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

---

> [!NOTE]
> Adjusts on-slot pruning to only remove trackers when expired or block is imported, completing pending futures exceptionally; updates tests accordingly.
> 
> - **DAS sampling (`DasSamplerBasic`)**:
>   - Refines `onSlot` pruning: only remove trackers when `slot < firstNonFinalizedSlot` or `recentChainData.containsBlock(root)`.
>     - If tracker is already completed, remove immediately.
>     - If not completed, complete its `completionFuture` exceptionally before removal.
> - **Tests (`DasSamplerBasicTest`)**:
>   - Update `onSlot_shouldPruneTrackers` to stub `blockRoot`/`slot` and assert correct pruning and exceptional completion for expired/imported trackers, while retaining valid tracker.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a23bbb6fd58675b52fcbf9b79c30a31282331a9c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #10233, opened 2025-12-12
Add inflight db sidecars

tbenr:
This is just an idea, to be validated in real world, and it could explain why, before #10219, we were actually duplicating download requests instead of just rebuilding the tracker which then completes immediately.

if that's the case, the underlining design problem\bug is that we don't write Columns via `CombinedStorageChannelSplitter`.

`CombinedStorageChannelSplitter` guarantees that reads are happening only after writes.

this approach is beneficial because it does not wait the writes, and can be very useful for quickly serving byRoot RPCs on the block head.

We should probably fix `CombinedStorageChannelSplitter` anyway too.

## Documentation

## Changelog

---

> [!NOTE]
> Introduces in-memory tracking of inflight `DataColumnSidecar`s to surface them in reads before persistence.
> 
> - Maintain concurrent `inflightColumns` and include them in `getSidecar` (returned immediately) and `getColumnIdentifiers` (merged with DB results, deduped via `LinkedHashSet`)
> - On `addSidecar`, invalidate slot cache, add to inflight, delegate write, and remove from inflight on completion
> - Read-slot cache logic remains; added tests validating inflight reads, identifier merging, cache updates, and pruning
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d1a7bcdd0d92b25c82e76561cc6a5fe93b5d8b1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/db/ColumnIdCachingDasDb.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/db/ColumnIdCachingDasDb.java
@@ -15,10 +15,12 @@ package tech.pegasys.teku.statetransition.datacolumns.db;
 
 import java.util.BitSet;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


tbenr:
superseded

PR #10279, opened 2026-01-16
Partition sidecar byRoot batches depending on max blobs

zilm13:
## PR Description
Sidecar size depends on the number of blobs. We cannot say exactly the number without the block and commitments number but we could expect bigger sidecars when max blobs is bigger, so comes this change. Please note, we will still send all queued requests but they will be split accordingly.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes DAS data-column RPC batching behavior based on spec blob limits, which can affect request sizing and peer interactions under load. Risk is moderate due to new batching logic and dependency on correct spec-derived limits, but scope is limited to sidecar retrieval.
> 
> **Overview**
> Adjusts `DataColumnReqRespBatchingImpl` to **dynamically partition** `DataColumnsByRootIdentifier` batches based on the expected maximum blobs per block, rather than using a fixed batch size.
> 
> The batching implementation now takes a `Spec`, derives a target “cells per request” from Fulu config, computes `maxSidecarsInRequest` using `getMaxBlobsPerBlockAtSlot` (with an Electra fallback), and splits requests accordingly; wiring in `BeaconChainController` is updated and tests add coverage for the new `partitionRequests` behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ef2fe441600904538c5b9de36d2cbe59d09860f3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/DataColumnReqRespBatchingImpl.java
```diff
@@ -110,6 +126,48 @@ private void flushForNode(final UInt256 nodeId, final List<RequestEntry> nodeReq
...
   }
 
+  int getMaxBlobsForRequests(final List<RequestEntry> nodeRequests) {
+    final UInt64 lastSlot = nodeRequests.getLast().columnIdentifier.slot();
+    final Optional<Integer> maybeMaxBlobs = spec.getMaxBlobsPerBlockAtSlot(lastSlot);
+    return maybeMaxBlobs.orElseGet(
+        () ->
+            SpecConfigElectra.required(spec.forMilestone(SpecMilestone.ELECTRA).getConfig())
+                .getMaxBlobsPerBlock());
+  }
```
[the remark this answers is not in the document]
zilm13: Yeah, we could get a really highest slot, but it's an extra calculation plus one request is very rarely split over 2 BPOs. It doesn't worth extra code/calculation I think.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/DataColumnReqRespBatchingImpl.java
```diff
@@ -100,8 +107,17 @@ private void flushForNode(final UInt256 nodeId, final List<RequestEntry> nodeReq
                     byRootSchema.create(
                         byRootRequest.root(), byRootRequest.columns().stream().toList()))
             .toList();
+
+    // for 21 max blobs it will be 1560 sidecars per request, 12+ full blocks
+    // for 72 max blobs it will be 455 sidecars per request, ~3.5 full blocks
+    final int maxBlobs = getMaxBlobsForRequests(nodeRequests);
+    final int maxSidecarsInRequest =
+        SpecConfigFulu.required(spec.forMilestone(SpecMilestone.FULU).getConfig())
+                .getMaxRequestDataColumnSidecars()
```
rolfyone: should we just store this? its not going to change...

rolfyone (approved):
lgtm minor q

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/retriever/DataColumnReqRespBatchingImpl.java
zilm13: Yeah, it makes sense, introduced variable `cellsTargetPerRequest` initialized in class constructor

PR #10282, opened 2026-01-20
Add sidecars through atomic DB updates

zilm13:
## Documentation

## Changelog


zilm13:
It doesn't help with the issue. The sync is still very bad, see graph and log cut:
image

[teku-beacon.log.zip](https://github.com/user-attachments/files/24773716/teku-beacon.log.zip)

PR #10287, opened 2026-01-21
updated DataColumnSidecars to store to file

rolfyone:
This feature would change the storage mechanism of DataColumnSidecars so that they're stored to file directly.

A desirable cleanup would still be the process that streams slot by slot to find data columns, as this results in many queries to the filesystem directly now.

## Documentation

## Changelog

---

> [!NOTE]
> **Storage migration to filesystem**
> 
> - Introduces file-based storage for `DataColumnSidecars` (`FileBasedDataColumnStorage`) under `beacon/columns` with read/write, range streaming, epoch-based pruning, and metadata
> - Adds migration path: `StorageUpdateChannel.migrateDataColumnSidecarsToFilesystem` exposed via `CombinedChainDataClient.migrateDataColumnSidecarsToFilesystem()`; `KvStoreDatabase` migrates sidecars (latest→earliest), updates earliest slot metadata, and marks completion
> - Refactors `KvStoreDatabase` to use filesystem for sidecar reads/writes/counts and pruning; clears DB vars post-migration
> - Gates `DasCustodyBackfiller` and `DasCustodySync` to trigger and await migration before processing
> - Updates DB factories (`RocksDbDatabaseFactory`, `LevelDbDatabaseFactory`, `VersionedDatabaseFactory`, test builders) to pass beacon data directory
> - Adds/updates tests for migration gating and new filesystem storage; updates `CHANGELOG.md` noting migration
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 68f69c2c3c8683307fcb6a746798486cb1214005. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java
```diff
@@ -58,7 +58,8 @@ public class P2PConfig {
   public static final int DEFAULT_DOWNLOAD_TIMEOUT_MS = 240_000;
 
   public static final int DEFAULT_COLUMN_CUSTODY_BACKFILLER_POLL_PERIOD_SECONDS = 30;
-  public static final int DEFAULT_COLUMN_CUSTODY_BACKFILLER_BATCH_SIZE = 10;
+  // Align with epoch size (32 slots) for efficient filesystem scanning
+  public static final int DEFAULT_COLUMN_CUSTODY_BACKFILLER_BATCH_SIZE = 32;
```
tbenr: i'm not sure about this change, from backfiller perspective.
I choose 10 to keep the "inflight" columns batch download to a reasonable level for 2 reasosns:
1- I don't want to have too many concurrent RPC columns download, to avoid interferences with RPC requests eventually needed to follow the chain (we backfill while following the chain)
2- The earliestavailable[...] db variable is updated when each batch is completed, the bigger, the slower that var will follow the backfill

tbenr: still need to check FS scanning to see what you're talking about :)

rolfyone: it started at 100, but i suggested aligning to the mainnet epoch size because it limits to 1 folder at a time, which is cleaner... i meant 32 slots but this is 32 columns, so arguably maybe it could be 128... open to changing that for sure (batched write, its just really db tx size that matters there)

rolfyone: i'll remove the cli arg anyway and set something, hadn't realised it'd added that :)

rolfyone: actualy this is just a total red herring, i'll remove the thing it changed and not the cli arg

storage/src/test/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorageTest.java
```diff
@@ -0,0 +1,389 @@
...
+            .forEach(
+                path -> {
+                  try {
+                    Files.deleteIfExists(path);
+                  } catch (IOException e) {
+                    // Ignore
+                  }
+                });
+      }
+    }
```
rolfyone, line 66: generated, but not sure this matters on a Tempdir, they get cleaned up anyway?

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
```diff
@@ -0,0 +1,565 @@
...
+  }
+
+  private void initializeStorage() {
+    LOG.trace("Initializing file-based data column storage at: {}", baseDirectory);
+    try {
+      Files.createDirectories(baseDirectory);
+      LOG.trace("Data column storage directory created or already exists: {}", baseDirectory);
+    } catch (IOException e) {
+      LOG.error("Failed to create base directory: {}", baseDirectory, e);
+      throw new RuntimeException("Failed to initialize file-based data column storage", e);
```
rolfyone, line 99: errors all look like RuntimeException, which will basically be out of space, could potentialy make cleaner but this is probably ok...

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
```diff
@@ -0,0 +1,565 @@
...
+  private List<DataColumnSlotAndIdentifier> getIdentifiersInEpoch(final UInt64 slot) {
+    final Path epochDir = resolveEpochDirectory(slot);
+    final List<DataColumnSlotAndIdentifier> identifiers = new ArrayList<>();
+
+    if (!Files.exists(epochDir)) {
+      return identifiers;
+    }
+
+    try (DirectoryStream<Path> stream =
+        Files.newDirectoryStream(epochDir, "*" + SSZ_FILE_EXTENSION)) {
```
rolfyone: we can just filter by `slot*`

storage/src/testFixtures/java/tech/pegasys/teku/storage/server/kvstore/InMemoryKvStoreDatabaseFactory.java
```diff
@@ -63,6 +73,21 @@ public static Database createTree(
...
+        storageMode,
+        storeNonCanonicalBlocks,
+        1000,
+        spec,
+        createTempDirectory());
+  }
+
+  private static Path createTempDirectory() {
+    try {
+      return Files.createTempDirectory("teku-test-data-columns");
```
zilm13, line 88: not sure, is it cleaned up afterwards?

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
```diff
@@ -0,0 +1,565 @@
...
+
+    return identifiers.stream().sorted();
+  }
+
+  // ==================== Metadata Operations ====================
+
+  /** Get the earliest available slot (after pruning). */
+  public Optional<UInt64> getEarliestAvailableSlot() {
+    final Map<String, Object> metadata = loadMetadata();
+    final Object value = metadata.get("earliestAvailableSlot");
```
zilm13: please, put "earliestAvailableSlot" to constant

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
```diff
@@ -0,0 +1,565 @@
...
+    } catch (IOException e) {
+      LOG.warn("Failed to scan for earliest slot", e);
+      return Optional.empty();
+    }
+  }
+
+  /** Get first custody incomplete slot from metadata. */
+  public Optional<UInt64> getFirstCustodyIncompleteSlot() {
+    final Map<String, Object> metadata = loadMetadata();
+    final Object value = metadata.get("firstCustodyIncompleteSlot");
```
zilm13: please, put "firstCustodyIncompleteSlot" and other metadata fields to constant

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
```diff
@@ -0,0 +1,565 @@
...
+    final List<UInt64> epochsToPrune = new ArrayList<>();
+
+    try (Stream<Path> epochDirs = Files.list(baseDirectory)) {
+      epochDirs
+          .filter(Files::isDirectory)
+          .map(Path::getFileName)
+          .map(Path::toString)
+          .filter(this::isNumeric)
+          .map(UInt64::valueOf)
+          .filter(epoch -> epoch.isLessThanOrEqualTo(cutoffEpoch))
```
zilm13, line 426: from the description:
"Deletes entire epoch directories for epochs before the epoch containing tillSlotInclusive
We'd better keep this behavior, otherwise we will delete more slots than asked with this kind of algo, so it should be `isLessThan` without equal here

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
```diff
@@ -0,0 +1,565 @@
...
+  // ==================== Stream Operations ====================
+
+  /** Stream identifiers in a slot range (inclusive). */
+  public Stream<DataColumnSlotAndIdentifier> streamIdentifiers(
+      final UInt64 startSlot, final UInt64 endSlot) {
+    return streamIdentifiersInRange(startSlot, endSlot);
+  }
+
+  /** Stream non-canonical identifiers in a slot range (inclusive). */
+  public Stream<DataColumnSlotAndIdentifier> streamNonCanonicalIdentifiers(
```
zilm13, line 223: This interface is confusing. If we cannot distinguish it here, we'd better leave the only correct one `streamIdentifiersInRange` with Javadoc that it will give both canonical and non-canonical

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -1123,52 +1147,47 @@ public void setFinalizedDepositSnapshot(final DepositTreeSnapshot finalizedDepos
...
-    return maybePayload.map(payload -> spec.deserializeSidecar(payload, identifier.slot()));
+    return dataColumnStorage.getNonCanonicalSidecar(identifier);
   }
 
   @Override
   @MustBeClosed
   public Stream<DataColumnSlotAndIdentifier> streamDataColumnIdentifiers(
       final UInt64 firstSlot, final UInt64 lastSlot) {
-    return dao.streamDataColumnIdentifiers(firstSlot, lastSlot);
+    return dataColumnStorage.streamIdentifiers(firstSlot, lastSlot);
```
zilm13, line 1168: so we will stream both canonical and non-canonical here, we need to resolve it somehow

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -1183,91 +1202,150 @@ public Optional<List<List<KZGProof>>> getDataColumnSidecarsProofs(final UInt64 s
...
 
-    final List<UInt64> slots = prunableMap.keySet().stream().sorted().toList();
+    // Migrate in batches (32 slots = 1 epoch, aligns with filesystem structure)
+    // START FROM LATEST and work backward to EARLIEST
+    final int batchSizeSlots = 32;
+    UInt64 currentSlot = latestSlot.get();
+    final UInt64 endSlot = earliestSlot.get();
+    int totalMigrated = 0;
+
+    while (currentSlot.isGreaterThanOrEqualTo(endSlot)) {
```
zilm13, line 1280: it looks very heavy to do it one by one without delays. If it's a supernode and it has already 4096 epochs, it's moving 500Gb from DB to filesystem without any delays. We should probably make a delay after each batch or node is going to be stalled until migration is finished

storage/src/test/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorageTest.java
zilm13: we have tests with Junit's `@TempDir` it's the way when it's guaranteed to be cleaned up after test is finished

zilm13 (commented):
I've performed a test with this PR today and existing node using leveldb2 storage.
Attempt to switch to PR's binary have failed unable to sync and recover. I've reset the DB, wait until it stabilize, pause the node for 1 hour and tried to start to see how long it will take to reach the current:
- it was definitely better than LevelDB2 with sidecars DB storage
- it was still still slower than RocksDB with sidecars DB storage
- the node was less stable during the sync and later 
I think if we want this, we need to put it under the feature flag. It could improve behavior for LevelDB nodes, when someone wants to keep it, but for RocksDB it looks jsut a risk adding it when it peforms well storing sidecars in database.

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
rolfyone: fair but we dont even know its working so im probably not going to do too much cleanup unless we ultimately decide to merge...

rolfyone: its kind of removed the non canonical concept atm so that'd need to come from another source whereever is using it higher up

storage/src/main/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorage.java
rolfyone: the idea would be we're purging only epochs beyond what we want to keep, its more efficient to keep a few extra slots and just delete the epochs we dont want basically.

storage/src/test/java/tech/pegasys/teku/storage/server/filesystem/FileBasedDataColumnStorageTest.java
rolfyone: yeh i dont know why it decided to do that heh

rolfyone:
not going in this direction at the moment.

PR #10296, opened 2026-01-24
Add completed slots cache in recovering custody

zilm13:
## PR Description
We had documented cases when sidecars are submitted to recovering custody for 2 times in a row. For example with failed sync batch submission, which retries the same range after a while. If there are misses in the second retry, recovering custody will try to recover sidecars which are already in DB. I've added small light cache to `DataColumnSidecarRecoveringCustodyImpl` to avoid such behavior and skip creation of recovery task for the second time. 

Regarding the disabled test: it takes locally om m4 like 10 seconds, so it will take on CI 30+ seconds, which I think is too much for single unit test. If some could suggest a lighter way to check the same, I'd appreciate and will change the test.

## Documentation

## Changelog

---

> [!NOTE]
> Adds a lightweight in-memory cache of completed `SlotAndBlockRoot`s to skip duplicate recovery and sidecar handling, populated when all columns are present or after reconstruction.
> 
> - New `completedSlots` set with pruning (`completedSlotsSizeTarget`); pruned each `onSlot`
> - Early-return in `onNewValidatedDataColumnSidecar` when slot is completed
> - Recovery flow updates: mark slot completed when all columns already present or after recovery; clear task sidecars
> - New constructor enabling configurable `recoveryTasksSizeTarget` and `completedSlotsSizeTarget` (public ctor retains sensible defaults); minor visibility/test helpers added
> - Tests added/updated for completion, pruning, and task behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6ea69aa37aff960a055ae0fa752f99a14e92a329. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
@@ -83,7 +83,8 @@ public class DataColumnSidecarRecoveringCustodyImpl implements DataColumnSidecar
 
   private volatile boolean inSync;
 
-  public DataColumnSidecarRecoveringCustodyImpl(
+  @VisibleForTesting
+  protected DataColumnSidecarRecoveringCustodyImpl(
       final DataColumnSidecarByRootCustody delegate,
       final AsyncRunner asyncRunner,
       final Spec spec,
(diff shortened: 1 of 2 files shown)


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -160,6 +169,12 @@ private void pruneRecoveryTasks() {
     }
   }
 
+  private void pruneCompletedSlots() {
+    while (completedSlots.size() >= completedSlotsSizeTarget) {
+      completedSlots.removeFirst();
+    }
```
tbenr: `size()` in `ConcurrentSkipListSet` is `O(n)`. Since the pruning happens only here and it is guaranteed to run non-concurrently (onSlot never runs concurrently) I think you can get `size` and calculate the overshoot once and call removeFirst accordingly.

zilm13: You are right. I've pushed the changes

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -160,6 +169,15 @@ private void pruneRecoveryTasks() {
...
   }
 
+  private void pruneCompletedSlots() {
+    final int slotsToPrune = completedSlots.size() - completedSlotsSizeTarget;
+    if (slotsToPrune >= 0) {
+      for (int i = 0; i <= slotsToPrune; i++) {
+        completedSlots.removeFirst();
+      }
+    }
+  }
```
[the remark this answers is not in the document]
zilm13: Yeah we want space for at least one element, which could be added before the next prune call

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
```diff
@@ -170,8 +170,11 @@ private void pruneRecoveryTasks() {
   }
 
   private void pruneCompletedSlots() {
-    while (completedSlots.size() >= completedSlotsSizeTarget) {
-      completedSlots.removeFirst();
+    final int slotsToPrune = completedSlots.size() - completedSlotsSizeTarget;
+    if (slotsToPrune >= 0) {
```
tbenr, line 205: you don't need this if :)

zilm13: yeah, but it's more readable when explicitly checked

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarRecoveringCustodyImpl.java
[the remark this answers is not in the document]
tbenr: > Yeah we want space for at least one element, which could be added before the next prune call

I don't get this argument

tbenr (approved):
LGTM

PR #10304, opened 2026-01-29
Enable gossip during optimistic sync for data columns

### LINKED ISSUE #10275: We should subscribe to data columns subnets when following the chain optimistically
When we are in optimistic sync mode:
i.e.:
```
Syncing     *** Slot: 13471644, Head slot: 13471643, Waiting for execution layer sync, Connected peers: 100
```

We don't subscribe to data column subnets. Which means we follow the chain by downloading columns via RPC.
This is particularly relevant when running node in full custody via `p2p-subscribe-all-custody-subnets-enable`.

gfukushima:
## PR Description
This PR uses the flag `subscribedToAllCustodySubnetsEnabled` to reflect whether gossip should be enabled during optimistic sync.

## Fixed Issue(s)

Fixes #10275

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches P2P gossip subscription behavior during optimistic sync and changes when sidecar topics are enabled, which could affect network load and sync behavior if the super-node predicate or wiring is incorrect.
> 
> **Overview**
> Enables **data column sidecar gossip during optimistic sync** when the node qualifies as a *super node* (custody group count equals the chain’s max custody groups).
> 
> This introduces `SuperNodeSupplier` and threads an `isSuperNodeSupplier` through `BeaconChainController` → `Eth2P2PNetworkBuilder` → `GossipForkSubscriptions*` → `DataColumnSidecarGossipManager`, which now overrides `isEnabledDuringOptimisticSync()` to consult the supplier (and adds extra trace logging on sidecar receipt). Custody “super node” detection is centralized via new `MiscHelpersFulu.isSuperNode`, removing direct `SpecConfigFulu` usage in `CustodyGroupCountManagerImpl`, with updated/new tests covering the gating behavior and supplier dynamism.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ce4f0617578ff111549cb4c3526cd94128f24479. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarManagerImpl.java
```diff
@@ -57,6 +57,7 @@ public DataColumnSidecarManagerImpl(
   @Override
   public SafeFuture<InternalValidationResult> onDataColumnSidecarGossip(
       final DataColumnSidecar dataColumnSidecar, final Optional<UInt64> arrivalTimestamp) {
+    LOG.info("Received data column sidecar from GOSSIP: {}", dataColumnSidecar);
```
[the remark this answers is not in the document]
gfukushima: fixed

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/DataColumnSidecarGossipManager.java
```diff
@@ -21,12 +21,15 @@
...
+  private final boolean subscribedToAllCustodySubnetsEnabled;
 
   public DataColumnSidecarGossipManager(
       final DataColumnSidecarSubnetSubscriptions dataColumnSidecarSubnetSubscriptions,
-      final DasGossipLogger dasGossipLogger) {
+      final DasGossipLogger dasGossipLogger,
+      final boolean subscribeAllCustodySubnetsEnabled) {
     subnetSubscriptions = dataColumnSidecarSubnetSubscriptions;
     this.dasGossipLogger = dasGossipLogger;
+    this.subscribedToAllCustodySubnetsEnabled = subscribeAllCustodySubnetsEnabled;
```
zilm13: It's not a good way to get this information.
Node could be subscribed to all custody subnets because:
- flag subscribe-custody..
- flag custody-override
- node is supernode because of the number of the validators served

Because number of the validators served could change dynamically on any epoch, this case should be handled too. For example check `DataColumnSidecarRecoveringCustodyImpl` how the flag `isSuperNode` is set. You need something similar but the check should be recurrent (because number of the validators could change), we do it usually `onSlot`, but there is no `onSlot` here, I'm not sure what will be the best option. We should do the check at least once per epoch.

gfukushima: I've changed the PR to contemplate the scenarios where you mentioned but the check isn't recurrent, I think this is better than the current state we can work on make the check every epoch on another PR, this might be useful in case the node stays in optimistic sync for too long, correct?

zilm13: It's a bad idea trying to duplicate logic of `CustodyGroupCountManager`. Moreover, you don't cover one of the most common customer's case: validator node with 128+ validators without any command line flags. It should be subscribed to all custody subnets. And it's a reason we need updates of the custody count status: (not all) validators are loaded on initialization stage, when constructor is called, keys decryption takes time, and real numbers breaks in a bit later, that's why we always try to follow status in other components.

gfukushima: since this gossip subscription is checked (for possible new subscriptions start) every epoch I believe we don't need a mechanism to keep checking constantly, I'm relying on the DB variable to tell if the node has changed status and is supernode when the change happens, if so before the epoch transition we should start the subscription for the data column gossip. If this isn't good enough than maybe we can think about a event based logic.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarManagerImpl.java
zilm13: it's debug or trace

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/forks/versions/GossipForkSubscriptionsFulu.java
```diff
@@ -147,4 +154,22 @@ public void subscribeToDataColumnSidecarSubnet(final int subnetId) {
   public void unsubscribeFromDataColumnSidecarSubnet(final int subnetId) {
     dataColumnSidecarGossipManager.unsubscribeFromSubnetId(subnetId);
   }
+
+  private Supplier<Boolean> isSuperNode() {
```
zilm13: we shouldn't do this
how about something like this?
https://github.com/zilm13/teku/commit/24c45957cbe496ab6d7e204ae5d4e709a97ce192

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/util/SuperNodeSupplier.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class SuperNodeSupplier implements Supplier<Boolean> {
+
+  private final Spec spec;
+  private final boolean isSubscribedToAllCustodySubnetsEnabled;
+  private final Supplier<CustodyGroupCountManager> custodyGroupCountManager;
+
+  public SuperNodeSupplier(
+      final Spec spec,
+      final boolean isSubscribedToAllCustodySubnetsEnabled,
```
zilm13: You don't need to duplicate ths logic, custodyGroupCountManager will return 128 if subscribedToAllCustodySubnets flag is toggled on

You could drop gradle dependency after removing this and tests `shouldReturnTrue_whenExplicitSuperNodeConfigEnabled` and `shouldPrioritizeExplicitConfig_overCustodyCount`

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/DataColumnSidecarGossipManager.java
```diff
@@ -56,4 +60,9 @@ public void subscribe() {
   public void unsubscribe() {
     subnetSubscriptions.unsubscribe();
   }
+
+  @Override
+  public boolean isEnabledDuringOptimisticSync() {
+    return isSuperNodeSupplier.get();
```
zilm13, line 71: we could add small comment how it's linked to this

networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/gossip/DataColumnSidecarGossipManagerTest.java
```diff
@@ -262,4 +263,71 @@ public void testPublishNotLinkedToSubscription() {
...
+        new DataColumnSidecarGossipManager(subnetSubscriptions, dasGossipLogger, () -> true);
+
+    assertThat(managerWithSuperNode.isEnabledDuringOptimisticSync())
+        .as(
+            "isEnabledDuringOptimisticSync should return true when isSuperNodeSupplier returns true")
+        .isTrue();
+  }
+
+  @TestTemplate
+  public void testIsEnabledDuringOptimisticSync_whenSuperNodeSupplierReturnsFalse() {
```
zilm13: isDisabled

gfukushima: done

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/util/SuperNodeSupplier.java
gfukushima: done

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/gossip/DataColumnSidecarGossipManager.java
gfukushima: added a reference to the class that implements the supplier but not sure if there's much to be added in the comment this line of code will be tied to this PR which is tied to the original issue.

zilm13 (approved):
LGTM

PR #10312, opened 2026-02-02
Enable new column backfiller by default and fix DB variable update

tbenr:
This PR also make the backfiller the only one responsible of writing `earliestAvailableCustodySlot` db variable.
see #10318 

worth mentioning that we still need to implement https://github.com/Consensys/teku/issues/10245 which will update `StatusMessageFactory::updateEarliestAvailableSlot` logic so that the actual slot we will advertise to peers in the StatusMessage will be able to go back to the earliest available block when custody is considered fulfilled.

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [ ] I thought about adding a changelog entry, and added one if I deemed necessary.

---

> [!NOTE]
> **Medium Risk**
> Changes default behavior to enable the reworked column custody backfiller and adjusts how the custody cursor DB value is advanced, which can impact sync/backfill behavior and resource usage on real networks.
> 
> **Overview**
> Enables the *reworked data-column custody backfiller* by default by switching the default for `--Xp2p-reworked-sidecar-custody-sync-enabled` (and the underlying `P2PConfig` default) from disabled to enabled.
> 
> Fixes `DasCustodyBackfiller` so when the stored `earliestAvailableCustodySlot` is at/behind the moving `minCustodyPeriodSlot`, it **writes the cursor forward to `minCustodyPeriodSlot`** instead of just exiting, and adds a unit test covering this cursor-alignment case. Acceptance tests are updated to stop explicitly enabling the reworked custody sync, and the changelog notes the backfiller being enabled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8ba9e1fe3423a859a4e752973cde70442e34082c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #10265, opened 2026-01-11
Update earliest available data column slot after pruning

gfukushima:
## PR Description
This PR fixes the issue where we would prune the data columns and not update the earliest available slot.

## Fixed Issue(s)

Fixes #10246

## Documentation

- [x] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [x] I thought about adding a changelog entry, and added one if I deemed necessary.

---

> [!NOTE]
> **Fix: earliest available data column slot tracking after prune**
> 
> - Update `KvStoreDatabase.pruneDataColumnSidecars(...)` to set `earliestAvailableDataColumnSlot` to `slot.plus(1)` after pruning canonical `DataColumnSidecar`s (no update for non-canonical). Persist the value via `FinalizedUpdater#setEarliestAvailableDataColumnSlot` after each prune batch.
> - Add integration test `DatabaseTest.verifyDataColumnSidecarsPruningSetsDbVariable` that prunes via `database.pruneAllSidecars(...)` and asserts `database.getEarliestAvailableDataColumnSlot()` advances correctly (to `1`, then `3`).
> 
> **Misc**
> - Rename typo `MAX_REGENERATE_LOTS` → `MAX_REGENERATE_SLOTS` in `FinalizedStateCache` (constructor wiring updated).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 21b367f13952633246a9e12f7045dcb96ff70c1c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #10318, opened 2026-02-03
Revert "Update earliest available data column slot after pruning"

tbenr:
Reverts Consensys/teku#10265

As we discussed, there is an unwanted interaction with the new custody backfiller. We need to avoid both classes writing the same variable. It is better to leave the pruner as is and let the backfiller be the only one responsible for updating the variable.

done here: #10312

---

> [!NOTE]
> **Medium Risk**
> Changes data-column sidecar pruning to no longer update the `earliestAvailableDataColumnSlot` DB variable, which could affect components relying on that metadata for backfill/custody progress.
> 
> **Overview**
> Reverts the earlier behavior where data-column sidecar pruning updated the `earliestAvailableDataColumnSlot` variable: `KvStoreDatabase.pruneDataColumnSidecars` now only deletes sidecars and commits, leaving that variable to be managed elsewhere.
> 
> Removes the integration test that asserted `earliestAvailableDataColumnSlot` changes after pruning, and includes a small internal rename in `FinalizedStateCache` (`MAX_REGENERATE_SLOTS` -> `MAX_REGENERATE_LOTS`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 14ea9c888e9e8aa4f75d5eedfd6a6fa406e42f64. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/state/FinalizedStateCache.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/state/FinalizedStateCache.java
@@ -29,7 +29,7 @@ import tech.pegasys.teku.storage.server.Database;
 
 public class FinalizedStateCache {
 
-  private static final long MAX_REGENERATE_SLOTS = 10_000L;
+  private static final long MAX_REGENERATE_LOTS = 10_000L;
 
   /**
    * Note this is a best effort basis to track what states are cached. Slots are added here slightly
@@ -52,7 +52,7 @@ public class FinalizedStateCache {
         maximumCacheSize,
         useSoftReferences,
         stateRebuildTimeoutSeconds,
-        MAX_REGENERATE_SLOTS);
+        MAX_REGENERATE_LOTS);
   }
 
   FinalizedStateCache(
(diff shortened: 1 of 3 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/state/FinalizedStateCache.java
```diff
@@ -29,7 +29,7 @@
 
 public class FinalizedStateCache {
 
-  private static final long MAX_REGENERATE_SLOTS = 10_000L;
+  private static final long MAX_REGENERATE_LOTS = 10_000L;
```
zilm13, line 32: we keep this, so only reverting?

zilm13 (approved):
LGTM

storage/src/main/java/tech/pegasys/teku/storage/server/state/FinalizedStateCache.java
tbenr: let change this on another PR, i'd like to leave this as a clean revert

PR #10333, opened 2026-02-05
Add DataColumnSidecarSelectorFactory test

zilm13:
## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes runtime selector behavior to return no data and avoid storage calls pre-`FULU`, which could affect API responses around fork boundaries. Added tests reduce regression risk but milestone/commitment edge cases should be validated.
> 
> **Overview**
> **Data column sidecar selection is now explicitly `FULU`-gated.** `DataColumnSidecarSelectorFactory` short-circuits `getDataColumnSidecars*` to return `Optional.empty()` before `SpecMilestone.FULU`, and treats `FULU` Deneb blocks with no blob commitments as a valid request returning an empty list (without calling `getDataColumnSidecars`).
> 
> Adds a new `DataColumnSidecarSelectorFactoryTest` suite covering selector behaviors (head/finalized/genesis/slot/blockRoot), milestone gating (verifying storage calls are skipped pre-`FULU`), error/unsupported selector paths, and key metadata expectations (e.g., genesis optimistic=false, finalized flag propagation).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fc5fb49397fb7a74d801e56bbe06c92f668b87cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tbenr (commented):
I'm not a fan of these `if(fulu)`s

end returning `empty` the API will return a 404 even if the request is invalid, so we should trigger a 400 at API level, imo.

zilm13:
@tbenr It depends on selector if we know this info at the certain phase:
`genesis`, `slot` - we know at the beginning
`head`, `blockRoot`, `finalized` - we need to get block info to be aware if slot is in fulu
So we defintely cannot do it on earlier stage in all cases and it depends on selector and should be in selector logic.

data/provider/src/test/java/tech/pegasys/teku/api/datacolumnselector/DataColumnSidecarSelectorFactoryTest.java
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  @TestTemplate
+  public void shouldLookForDataColumnSidecarsByHeadOnlyAfterFulu(
+      final TestSpecInvocationContextProvider.SpecContext ctx)
+      throws ExecutionException, InterruptedException {
+    final DataColumnSidecarSelectorFactory dataColumnSidecarSelectorFactory =
+        new DataColumnSidecarSelectorFactory(ctx.getSpec(), client);
+
+    final SignedBlockAndState blockAndState = data.randomSignedBlockAndState(100);
```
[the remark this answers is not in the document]
rolfyone: this is correct, if we're using testTemplate we cant just use the dataStructureUtil from above, we should fix this. especially foir a block and state.

data/provider/src/test/java/tech/pegasys/teku/api/datacolumnselector/DataColumnSidecarSelectorFactoryTest.java
```diff
@@ -0,0 +1,331 @@
...
+    assertThat(maybeDataColumnSidecars.get().getData()).isEmpty();
+  }
+
+  @TestTemplate
+  public void shouldLookForDataColumnSidecarsBySlotOnlyAfterFulu(
+      final TestSpecInvocationContextProvider.SpecContext ctx)
+      throws ExecutionException, InterruptedException {
+    final DataColumnSidecarSelectorFactory dataColumnSidecarSelectorFactory =
+        new DataColumnSidecarSelectorFactory(ctx.getSpec(), client);
+    final SignedBeaconBlock block = new DataStructureUtil(ctx.getSpec()).randomSignedBeaconBlock();
```
rolfyone: we can just use dataStructureUtil from the SpecContext. this same idea fixes the cursor issue on 262.

data/provider/src/test/java/tech/pegasys/teku/api/datacolumnselector/DataColumnSidecarSelectorFactoryTest.java
```diff
@@ -0,0 +1,331 @@
...
+import tech.pegasys.teku.spec.util.DataStructureUtil;
+import tech.pegasys.teku.storage.client.ChainHead;
+import tech.pegasys.teku.storage.client.CombinedChainDataClient;
+
+@TestSpecContext(allMilestones = true)
+public class DataColumnSidecarSelectorFactoryTest {
+
+  private final CombinedChainDataClient client = mock(CombinedChainDataClient.class);
+  private final Spec spec = TestSpecFactory.createMinimalFulu();
+  private final DataStructureUtil data = new DataStructureUtil(spec);
```
rolfyone: usually we'd c all this dataStructureUtil.

data/provider/src/test/java/tech/pegasys/teku/api/datacolumnselector/DataColumnSidecarSelectorFactoryTest.java
```diff
@@ -0,0 +1,331 @@
...
+  private final List<UInt64> indices = List.of(UInt64.ZERO, UInt64.ONE);
+  private final SignedBeaconBlock block = data.randomSignedBeaconBlock();
+  private final List<DataColumnSidecar> dataColumnSidecars =
+      IntStream.range(0, 5)
+          .mapToObj(index -> data.randomDataColumnSidecar(block.asHeader(), UInt64.valueOf(index)))
+          .toList();
+  private final DataColumnSidecarSelectorFactory dataColumnSidecarSelectorFactory =
+      new DataColumnSidecarSelectorFactory(spec, client);
+
+  @Test
```
rolfyone: we've got mixed concerns here between non TestTemplate and Test, do we normally do this? ordinarily i'd split them out, not sure if all these tests will run the way we expect

data/provider/src/test/java/tech/pegasys/teku/api/datacolumnselector/DataColumnSidecarSelectorFactoryTest.java
```diff
@@ -0,0 +1,362 @@
...
+  @BeforeEach
+  public void setup(final SpecContext specContext) {
+    this.specContext = specContext;
+    spec = specContext.getSpec();
+    dataStructureUtil = specContext.getDataStructureUtil();
+    block = dataStructureUtil.randomSignedBeaconBlock();
+    dataColumnSidecarSelectorFactory = new DataColumnSidecarSelectorFactory(spec, client);
+
+    // Post-FULU initialization part
+    if (!specContext.getSpecMilestone().isGreaterThanOrEqualTo(SpecMilestone.FULU)) {
```
rolfyone: rather than if not greater than return, we should just go if greater than do this bit, which removes the negative logic and skips the early return

rolfyone (approved):
just a nit on initialization, lgtm after that.

data/provider/src/test/java/tech/pegasys/teku/api/datacolumnselector/DataColumnSidecarSelectorFactoryTest.java
```diff
@@ -0,0 +1,362 @@
...
+        .thenReturn(SafeFuture.completedFuture(dataColumnSidecars));
+    when(client.getChainHead()).thenReturn(Optional.of(ChainHead.create(blockAndState)));
+
+    final Optional<DataColumnSidecarsAndMetaData> result =
+        dataColumnSidecarSelectorFactory
+            .slotSelector(block.getSlot())
+            .getDataColumnSidecars(indices)
+            .get();
+    assertThat(result.get().getData()).isEqualTo(dataColumnSidecars);
+  }
```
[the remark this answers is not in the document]
zilm13: right, addressed

data/provider/src/test/java/tech/pegasys/teku/api/datacolumnselector/DataColumnSidecarSelectorFactoryTest.java
zilm13: updated

PR #10395, opened 2026-02-20
Progressive ssz

### LINKED ISSUE #9772: Implement ProgressiveList, ProgressiveBitList and ProgressiveContainer SSZ types
New SSZ types were introduced in the spec with [EIP-7916](https://eips.ethereum.org/EIPS/eip-7916) and [EIP-7495](https://eips.ethereum.org/EIPS/eip-7495). 7916 also defines new merkleization for progressive lists.
The goal is to do less hashes when merkleizing small lists and don't break GIndex when we need to change list size for another fork. The task could be probably broken down to the following PRs:
- [ ] Implement progressive list merkleization
- [ ] Implement `ProgressiveList` SSZ type https://github.com/ethereum/consensus-specs/pull/4445
- [ ] Implement `ProgressiveBitList` SSZ type https://github.com/ethereum/consensus-specs/pull/4480
- [ ] Implement `ProgressiveContainer` SSZ type https://github.com/ethereum/consensus-specs/pull/4529
- [ ] Enable reference tests for `ProgressiveList`, `ProgressiveBitList` and `ProgressiveContainer`

tbenr:
fixes: #9772

TODO:
- Implement `TreeNodeStore` related methods
- SuperNode hint
- We also probably need primitive versions using `CachingTreeAccessor`?
- do some benchmarking to evaluate performance impacts

---- 
The following branch
https://github.com/tbenr/teku/tree/progressive_ssz_electra_testing
implements:
- a unified Container\Progressive container, so that we can reuse the same classes but enable progressive mode just by passing the new `active_fields` param.
- introduce "progressive" constructor in `ContainerN` and `ContainerSchemaN`
- fixes progressivebitlist having size in the underlining data structure.
- implements SuperNode **hints** and `CachingTreeAccessor`.
- implements store\load functions
- introduces specialized ProgressiveUInt64List for compatibility.
- Experiment: Modifies Electra with
   1. progressiveBitList in electra attestation (aggregation bits)
   2. beaconState with validators and balances as progressive list

question: is the Unified approach the best or we want to keep Progressive classes well separated, leave common code in Util classes, and have specific `ProgressiveContainerN` and `ProgressiveBeaconState` classes too?

----

general perplexities:

- since PL are unbounded by design, our usage of schema SszLengthBounds will be partially defeated. IE: when introducing in attestation suddenly the attestation schema upperbound will be `Long.MAX_VALUE`. Which means at gossip level we wont be so restrictive anymore and the overall upperbound value we apply (10mb?) becomes applicable to more structures. I see 2 solutions: introduce the ability to "suggest" a reasonable upperbound per on PL, or haveing different limits on gossip (like 10mb for blocks only, 1mb for all the others)

- in beacon state, when upgrading the state to a PL enabled state schema, datastructures needs to be rebuilt.

- we need to be careful in types. we need to make sure we have specialized list schemas that implements the different interfaces we have (ie: `SszProgressiveUInt64ListImpl` implementing `SszUInt64List`) to be able to swap them

## Documentation

## Changelog

### DIFF
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
@@ -63,7 +63,7 @@ public class ManualReferenceTestRunner extends Eth2ReferenceTestCase {
    *
    * <p>May be overridden by the ENV_DISPLAY_NAME environment variable.
    */
-  private static final String DISPLAY_NAME = "";
+  private static final String DISPLAY_NAME = "progressive";
 
   @ParameterizedTest(name = "{0}")
   @MethodSource("loadReferenceTests")
--- a/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
+++ b/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/ManualReferenceTestRunner.java
@@ -63,7 +63,7 @@ public class ManualReferenceTestRunner extends Eth2ReferenceTestCase {
    *
    * <p>May be overridden by the ENV_DISPLAY_NAME environment variable.
    */
-  private static final String DISPLAY_NAME = "progressive";
+  private static final String DISPLAY_NAME = "";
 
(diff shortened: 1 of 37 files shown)


tbenr:
parking for now

PR #10396, opened 2026-02-23
Fix backfiller behaviour when `p2p-subscribe-all-custody-subnets-enabled` is enabled

### LINKED ISSUE #10232: Enabling `p2p-subscribe-all-custody-subnets-enabled` does not trigger backfill
We do not trigger backfill because we fail to detect the change.

In  `DasCustodyBackfiller` constructor we initialize with  the current synced custody with:

```java
this.currentSyncCustodyGroupCount = custodyGroupCountManager.getCustodyGroupSyncedCount();
```

but `custodyGroupCountManager.getCustodyGroupSyncedCount()` will return 128, which is a lie.

gfukushima:
## PR Description
This PR changes the fix the behaviour of the the backfiller when we enable the flag to override custody group count.

## Fixed Issue(s)

Fixes #10232

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes backfill/resync triggering logic and cursor handling, which can affect data column availability after restarts; coverage is improved with new targeted tests.
> 
> **Overview**
> Fixes a restart edge case where enabling `--p2p-subscribe-all-custody-subnets-enabled` (increasing custody group count) could leave the DAS custody backfiller thinking it was already complete.
> 
> `CustodyGroupCountManagerImpl` now seeds the *synced/backfilled* baseline from the previously stored DB value on first init, and `DasCustodyBackfiller` additionally detects on startup when `custodyGroupCount > syncedCount` to force a cursor reset/resync. Tests are updated and expanded to cover restart scenarios and to ensure cursor resets only when the custody group count actually increases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 351b6eae12f7b753453c05c2a69a4877aec35a26. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/CustodyGroupCountManagerImpl.java
@@ -247,7 +247,11 @@ public class CustodyGroupCountManagerImpl implements SlotEventsChannel, CustodyG
     }
     final int oldValue = custodyGroupCount.getAndSet(newCustodyGroupCount);
     if (oldValue == INITIAL_VALUE) {
-      setCustodyGroupSyncedCount(newCustodyGroupCount);
+      // Use the previously stored count as synced baseline on restart, so the backfiller detects
+      // and re-backfills if custody group count increased between restarts (e.g. when
+      // --p2p-subscribe-all-custody-subnets-enabled is added). On fresh start (no stored count),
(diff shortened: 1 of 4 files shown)


zilm13:
I think it should be done in very different manner and it's not the only case when it will return incorrect value. I will discuss the way how to fix this in a correct way with @tbenr tomorrow. 
So from my view the only case when manager could set `countSynced` is only when starting from genesis. Otherwise initialize it with 0. For all other cases we are waiting Backfiller's call to `setCustodyGroupSyncedCount`.

zilm13:
Closing in favor of #10401

PR #10426, opened 2026-02-26
Migrate `EventChannel` to virtual threads (feature flag)

tbenr:
## Summary

  Adds `--Xevent-channel-virtual-threads-enabled` hidden CLI flag to switch `EventChannel.createAsync()` from platform threads (`CachedThreadPool`) to virtual threads (`newThreadPerTaskExecutor`).

  The flag defaults to `false` (original platform thread behavior). When enabled, each async event channel uses virtual threads instead of a cached platform thread pool.

  **Why no lock changes?** [JEP 491](https://openjdk.org/jeps/491) (JDK 24) eliminates `synchronized` pinning of virtual threads, so the `ReentrantLock` conversions from the original PR are
  no longer necessary. JDK 24+ users get full virtual thread benefits without any lock migration.
  
  
## Benefits

  1. Memory savings: ~70 fewer platform threads x ~1MB stack each = ~70MB less committed memory.
  2. Reduced OS scheduling overhead: Fewer platform threads competing for CPU time slices.

(in practice, we can continue subscribing classes to `SlotEventsChannel` and others without the side effect of having a new OS thread for each subscription)

---

> [!NOTE]
> **Low Risk**
> Low risk because the new virtual-thread execution path is guarded by a hidden flag that defaults to `false`, preserving existing cached-thread-pool behavior. Main risk is limited to the opt-in virtual-thread executor choice for event delivery and its operational characteristics.
> 
> **Overview**
> Adds an opt-in (hidden) CLI flag `--Xevent-channel-virtual-threads-enabled` to run async `EventChannel` deliveries on virtual threads instead of a cached platform-thread pool.
> 
> Wires the flag through a new `BeaconNodeConfig` into `TekuConfiguration` and `AbstractNode` so `EventChannels`/`EventChannel.createAsync()` select between `Executors.newThreadPerTaskExecutor(Thread.ofVirtual...)` and the existing `newCachedThreadPool`. Includes CLI parsing tests covering default/explicit true/false behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4dfb911883f711afb91e515886ec60c2fd4a2605. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/events/CoalescingChainHeadChannel.java
```diff
@@ -45,60 +48,70 @@ public synchronized void chainHeadUpdated(
...
 
   @Override
-  public synchronized void onSyncingChange(final boolean isSyncing) {
-    syncing = isSyncing;
-    if (!syncing) {
-      pendingEvent.ifPresent(PendingEvent::send);
-      pendingEvent = Optional.empty();
+  public void onSyncingChange(final boolean isSyncing) {
+    lock.lock();
+    try {
```
rolfyone: maybe we should wrap a reentrant lock and implement `AutoCloseable` if we go down this route, so that we can just use try with resources, would make this a cleaner pattern maybe...

tbenr: I'm for these lock changes to be all reverted and assume JDK24 (and start comunicating to upgrade) But I suspect the downside in JDK21 is very limited anyway.

rolfyone: im not against the lock, i just think if we do it we should use a ReentrantLock wrapped so its autocloseable...

That would then mean we could do the try with resources pattern so its something like
```
try (AutoLock lock = lock.lock()) {
// work stuff
}
// no finish because autocloseable handles close
```

rolfyone (approved):
happy to add as feature...

PR #10454, opened 2026-03-05
CombinedChainDataClient "slot by block root" method

zilm13:
## PR Description
Implementation + tests. Usage will come in separate PR. We need it for `DataColumnSidecarsByRootMessageHandler` for example and probably in some other places. Currently we get full block in all cases just to validate its slot against minimal custody epoch, switching to slot only where possible will improve performance and decrease memory and DB usage.
## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches core storage query interfaces and DB access paths, so any implementation mismatch could surface at runtime. Logic is straightforward and covered by unit/integration tests, keeping overall risk moderate.
> 
> **Overview**
> Adds a new `CombinedChainDataClient.getSlotByBlockRoot` API to resolve a block’s slot without loading the full block, first checking hot fork-choice data, then finalized canonical slot index, and *optionally* falling back to fetching a finalized non-canonical block and extracting its slot.
> 
> Extends the storage query surface with `StorageQueryChannel.getNonCanonicalBlockByRoot` and wires it through throttling/splitting layers and database implementations (including KV-store and no-op), plus updates DAO signatures to return `SignedBeaconBlock` directly. Adds unit + integration tests covering hot/finalized/unknown roots and the non-canonical opt-in behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d1d686d48b4d14594651530a23275720477fa79. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java
```diff
@@ -579,6 +579,27 @@ public SafeFuture<Optional<SignedBeaconBlock>> getBlockByBlockRoot(final Bytes32
...
+        .getFinalizedSlotByBlockRoot(blockRoot)
+        .thenCompose(
+            maybeSlot -> {
+              if (maybeSlot.isPresent()) {
+                return SafeFuture.completedFuture(maybeSlot);
+              }
+              // 3. historical nonCanonical only: get block and extract slot
+              return historicalChainData
+                  .getNonCanonicalBlockByRoot(blockRoot)
+                  .thenApply(maybeBlock -> maybeBlock.map(SignedBeaconBlock::getSlot));
```
tbenr: I understand this is required to have a complete solution, but is there an actual usecase?
in other words, assuming the usage in RPC by root requests, are we supposed to return non canonical data in a finalized slot? I'm not sure.. If there are usages like in restAPI i can see a possible usage.
We could add a boolean in the api like "resolveNonCanonicalFinalizedSlots" but it will be `false` in the byroot RPC requests (if my doubts are solid)

tbenr (approved):
LGTM

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -313,6 +313,11 @@ protected void storeAnchorStateAndBlock(
     updater.addFinalizedBlock(block);
   }
 
+  @Override
+  public Optional<SignedBeaconBlock> getNonCanonicalBlockByRoot(final Bytes32 blockRoot) {
+    return dao.getNonCanonicalBlock(blockRoot).map(b -> b);
```
tbenr: i think we can move from`<? extends SignedBeaconBlock>` to `<SignedBeaconBlock>` in getNonCanonicalBlock and remove this strange map

tbenr (approved):
LGTM

PR #10470, opened 2026-03-10
Add option to enable blobdb on RocksDB

gfukushima:
## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches RocksDB column-family configuration and database factory wiring, which can affect storage performance and on-disk behavior when enabled. Default remains off and the option is hidden, reducing blast radius but still warrants careful validation on real DBs.
> 
> **Overview**
> Adds a hidden CLI/storage config flag (`--Xdata-storage-rocksdb-blob-db-enabled`) to optionally enable RocksDB BlobDB for column families that store *static/large* finalized data.
> 
> When enabled, RocksDB column-family descriptors are created with BlobDB settings (blob files + LZ4 compression + GC) only for marked columns (e.g., finalized blocks/states, blob sidecars, data column sidecars/proofs), with the flag plumbed through `StorageConfiguration` → `VersionedDatabaseFactory` → `KvStoreConfiguration` for V4/V5/V6 DBs.
> 
> Also adds storage latency timers around finalized reads (blocks/states/sidecars/proofs) and extends `RocksDbStats` to expose BlobDB-specific tickers and size gauges when BlobDB is enabled, with tests updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bd8f717eba82064b8ae228b16aa01442b01cac18. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbStats.java
```diff
@@ -178,6 +194,23 @@ public void registerMetrics(final RocksDB database) {
...
+        category,
+        "live_blob_file_size",
+        "Size of live blob files in bytes (excluding garbage)",
+        () -> getLongProperty(database, "rocksdb.live-blob-file-size"));
+    metricsSystem.createLongGauge(
+        category,
+        "num_blob_files",
+        "Number of blob files",
+        () -> getLongProperty(database, "rocksdb.num-blob-files"));
+
```
tbenr, line 221: maybe we can create those only if blob is enabled?

tbenr (commented):
Looks interesting.
What should be gaining?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/schema/V6SchemaCombinedSnapshot.java
```diff
@@ -60,17 +60,21 @@ public class V6SchemaCombinedSnapshot extends V6SchemaCombined
...
     slotsByFinalizedRoot =
-        KvStoreColumn.create(finalizedOffset + 1, BYTES32_SERIALIZER, UINT64_SERIALIZER);
+        KvStoreColumn.create(finalizedOffset + 1, BYTES32_SERIALIZER, UINT64_SERIALIZER, true);
     finalizedBlocksBySlot =
         KvStoreColumn.create(
             finalizedOffset + 2,
             UINT64_SERIALIZER,
-            KvStoreSerializer.createSignedBlockSerializer(spec));
+            KvStoreSerializer.createSignedBlockSerializer(spec),
+            true);
```
tbenr, line 69: the "blob-enabled" column it a benefits even when the value size is small (like 8bytes as in this case)?

gfukushima: not really, should only be used to larger values, might have included this accidentally

gfukushima: looking a second time into the gain here is the storage of the blocks into blobdb so I'd expect the block to be larger than 8 bytes, right? I think the slotsByFinalizedRoot was included in the list by accident, but the rest makes sense to me

gfukushima:
In short we should get less compaction happening on those columns where data is written intensively during sync/backfilling processes but we might, might even see some improvements on the doUpdate since that is intesive write oriented as well. 

> ...reduces write amplification, which has several potential benefits like improved SSD lifetime, and better write and read performance...

from https://github.com/facebook/rocksdb/wiki/BlobDB

It needs a bit of soaking in more realistic setups which is why it is turned off by default.

rolfyone:
more of a 'usage' question, if i enable this, then restart without, do i still have all my data? just wondering about UX impact of toggling between using and not using, and whether i have to resync to change or if we need to persist the setting or something...

gfukushima:
Ideally we shouldn't allow for toggle on/off, toggling doesn't break the database or anything from the user perspective but I reckon it might impact in the perfomance eventually. I think if we decide this feature is worth, it should be enabled for all users without the option of toggling off.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/schema/V6SchemaCombinedSnapshot.java
tbenr: yes yes, here I was referring more to the stateRoot->Slot column :)

rolfyone (approved):
LGTM

Issue #9796, opened 2025-08-22
[SPIKE] Remove Quartz Scheduler in favor of our `RepeatingTaskScheduler`


tbenr:
We could rewrite `SlotProcessor` to use `RepeatingTaskScheduler`, like we do in `TimeBasedEventAdapter` for the VC side, so we could remove a dependency and simplify the implementation.

Maybe we don't need to advance the store time (`onTick`) every 500ms, and we could do that at each event (slotStart, attDue, aggDue, etc), but to do so we have to check spec and how we use the store time.

PR #10480, opened 2026-03-12
Added a new scheduler to potentially replace quartz

rolfyone:
Added a feature toggle, will remove in future release, to clean up a little used library.

The reason it wasn't completely removed initially is in case something unexpected happens. Scheduling is critical, so as long as this works we'll remove quartz.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Slot/tick scheduling is consensus-critical; this PR changes the timer implementation and wiring (including service lifecycle) and adds a runtime toggle, which could impact timing behavior under load or during shutdown.
> 
> **Overview**
> Introduces a new executor-based `TimerService` (replacing its prior Quartz-backed implementation) and adds a separate `QuartzTimerService` that retains the Quartz scheduler behavior.
> 
> Wires timer selection through a new hidden config/CLI flag `--Xquartz-scheduler-enabled` (stored in `Eth2NetworkConfiguration` and used by `BeaconChainController`) to choose between Quartz and the new scheduler, and adds tests to validate tick cadence, exception resilience, and no catch-up bursts across both implementations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 147f4d910a0aed33f03f6a7781ba8212427af8e8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tbenr:
fixes #9796

zilm13:
Did you miss to include `QuartzTimerService` class ?

zilm13:
Except the issue bot already highlighted in this PR I like 2 other suggestions which AI recommends to use for this case:
1. Schedule not onTick event, schedule dedicatedExecutor.onTick
```java
private final ScheduledExecutorService scheduler;  // Only schedules
private final ExecutorService tickExecutor;        // Executes ticks

private void scheduleTimerTask() {
  timerTask = scheduler.scheduleAtFixedRate(
      this::dispatchTick,  // Fast: just dispatches
      0,
      millisPerSlot,
      TimeUnit.MILLISECONDS);
}

private void dispatchTick() {
  final UInt64 tickTime = UInt64.valueOf(timeProvider.getTimeInMillis().longValue());
  
  tickExecutor.execute(() -> {
    try {
      timerEventChannel.onSlotEvent(new SlotEvent(tickTime));
      consecutiveFailures.set(0);
    } catch (Throwable t) {
      LOG.error("Tick execution failed for time {}", tickTime, t);
      consecutiveFailures.incrementAndGet();
    }
  });
}
```
2. Add monitoring (maybe on separate thread) to check when onTick was actually executed, how did it lag compared to expected schedule.

rolfyone:
> Did you miss to include `QuartzTimerService` class ?

it looks like i hadn't pushed everything - will look at this now

CHANGELOG.md
```diff
@@ -14,5 +14,6 @@
 - Added `/eth/v2/node/version` endpoint to retrieve structured version information for both beacon node and execution client.
 - Added deprecation warning on startup for any leveldb database types.
 - Increased default timeout of Engine API Get Payload requests to 2 seconds.
+- Removed quartz scheduler by default, will be removed completely in a future release. Can be re-enabled with `Xquartz-scheduler-enabled=true` (defaults false).
```
zilm13: Changelog should be updated or removed (as you defaulted to quartz in this PR)

services/timer/src/test/java/tech/pegasys/teku/services/timer/TimerServiceTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    assertThat(timedOut).as("timer should continue firing after exception").isFalse();
+    assertThat(successCount.get())
+        .as("at least 2 ticks should succeed after the exception")
+        .isGreaterThanOrEqualTo(2);
+  }
+
+  @ParameterizedTest
+  @MethodSource("timeToNextTaskMs")
+  @DisabledOnOs(OS.WINDOWS)
```
zilm13: why no windows here?

services/timer/src/test/java/tech/pegasys/teku/services/timer/TimerServiceTest.java
```diff
@@ -0,0 +1,198 @@
...
+  private Service serviceUnderTest;
+
+  public static Stream<Arguments> timeToNextTaskMs() {
+    // final long millis, final int ticksPerSecond, final long expectedValue
+    return Stream.of(
+        Arguments.of(0, 2, 500),
+        Arguments.of(1, 2, 499),
+        Arguments.of(499, 2, 1),
+        Arguments.of(500, 2, 500),
+        Arguments.of(999, 2, 1));
```
zilm13: If you add ` Arguments.of(999, 3, 0)` which is passing for the current calculation, doesn't it mean that if execution goes fast enough, we will keep running tick after tick at .999 for this case? Sounds like a bug in `nextTickDue` for me if I'm right on assumption.

services/timer/src/test/java/tech/pegasys/teku/services/timer/TimerServiceTest.java
rolfyone: windows breaks on this kind of stuff, and it was early. i can retest

services/timer/src/test/java/tech/pegasys/teku/services/timer/TimerServiceTest.java
rolfyone: we cant execute things fast enough for that to be a problem. its a computation that cleans itself up because the real world wont execute things like that basically

rolfyone: its more likely that we're a second later at 999, and in that case 0 is valid and also desired.

services/timer/src/test/java/tech/pegasys/teku/services/timer/TimerServiceTest.java
zilm13: it's compute only method

zilm13 (approved):
LGTM

Issue #10095, opened 2025-11-03
Meta: Store only 1/2 of sidecars on supernodes


zilm13:
We need to decrease storage requirements for validator nodes with high count of validators and supernodes. This could be achieved by deduplicating sidecars data which is 2x extension over blobs. I suggest following changes which could be split into 2-3 PRs:
- [x] Add flag `--data-column-sidecar-extension-retention-epochs`, which first defaults to Integer.MAX, so no sidecars are pruned
- [x] Improve DataColumnSidecar RPC handlers to have special logic when queried on supernode:
  a. Use dedicated thread pool with 2 threads for this kind of tasks
  b. If node is supernode, request asks for sidecars over retention period but in DA window (Defaults: 512 < epoch < 4096) and index is more than 63, we get 0-63 sidecars from DB and reconstruct everything else using thread pool. Data is not cached and used only for one request, but we should run only one extension for several indices in one request.
- [ ] Sidecar reconstructor, which recovers 64-127 index sidecar using 0-63 sidecars + kzgProofs from DB
- [ ] DataColumnSidecarExtensionPruner, which prunes 64-127 indices over `data-column-sidecar-extension-retention-epochs` if the node is supernode and writes kzgProofs to DB
- [ ] Test with some `data-column-sidecar-extension-retention-epochs`,  I suggest 512 epochs (a bit more than 2 days)
- [ ] Add documentation for `data-column-sidecar-extension-retention-epochs`
- [ ] If test goes well and there is support for the feature make default other than Integer.MAX

NOTE: 72 blobs `DataColumnSidecar`'s reconstruction for 1 slot is 1.2 ops/second on M4

tbenr:
I like the idea. The approach makes sense to me.
What do you think about calculating the retention boundary as

``` java
oldestExtentionRetentionEpoch = Math.min(latestFinalizedEpoch, currentEpoch - sidecarExtensionEpoch)
````

So that we always keep all extensions in case of non finality. This way we better serve the RPCs in case network instability where probably several nodes are doing a checkpoint sync.

zilm13:
I wanted to deduct it from last finalized, do we deduct 4096 from head epoch?? I thought we count only finalized, not? 
I think your formula makes sense

tbenr:
> I wanted to deduct it from last finalized, do we deduct 4096 from head epoch?? I thought we count only finalized, not?

IIRC finalization is not part of the DA boundary calculation.

zilm13:
[...]

PR #10508, opened 2026-03-18
Reconstruct supernode sidecars from 50% - Reconstructor

zilm13:
## PR Description

Reconstructor part + wiring + benchmark
Part of #10095 

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches P2P data-column serving, archival pruning boundaries, and KZG-backed reconstruction; incorrect pruning or reconstruction could serve wrong/missing columns or prune too aggressively, though validation and tests mitigate this.
> 
> **Overview**
> Replaces the **NOOP** `DataColumnSidecarArchiveReconstructor` with **`DataColumnSidecarArchiveReconstructorImpl`** on Fulu-capable nodes so supernodes can **rebuild second-half data column sidecars** from stored first-half columns plus archived KZG proofs, instead of keeping full column sets forever.
> 
> Reconstruction loads first-half sidecars and slot proofs via new **`CombinedChainDataClient.getDataColumnSidecarProofs`**, merges blobs/proofs into **`BlobAndCellProofs`**, and uses **`DataColumnSidecarUtil.constructDataColumnSidecars`**. Work is **deduplicated per RPC request** (`requestId` / `onRequestCompleted`) and runs on a dedicated async runner with success/failure metrics.
> 
> **`isSidecarPruned`** marks second-half columns as pruned for supernodes past the retention window; **`onNewFinalizedCheckpoint`** publishes **`SidecarArchivePrunableSlot`** through new **`SidecarArchivePrunableChannel`**. **`DataColumnSidecarsByRangeMessageHandler`** documents that range serving of reconstructed columns stays **best-effort** under the normal sidecar cap.
> 
> Also adds a JMH benchmark mirroring the archive path, unit tests for the reconstructor, removes a stale retention TODO, and documents avoiding **`.join()`/`.get()`** on futures in **AGENTS.md**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09d7ffdf1bfd1b5415d8b3e4fff144813f548e62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
```diff
@@ -0,0 +1,293 @@
...
+    sidecarArchivePrunableChannel.onSidecarArchivePrunableSlot(lastPrunableSlot);
+  }
+
+  private int getNextTaskId() {
+    final int nextTaskId = this.nextTaskId.getAndIncrement();
+    if (this.nextTaskId.get() > (Integer.MAX_VALUE - 1000)) {
+      this.nextTaskId.set(0);
+    }
+    return nextTaskId;
+  }
```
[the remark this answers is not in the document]
zilm13: We have 1000 gap to MAX_VALUE. It could slip a bit like 2-3 in real life, but definitely not for 1000, it's a huge

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
```diff
@@ -0,0 +1,293 @@
...
+        Stream.iterate(UInt64.ZERO, UInt64::increment).limit(halfColumns).toList();
+    final SafeFuture<List<DataColumnSidecar>> dataColumnSidecars =
+        chainDataClient.getDataColumnSidecars(block.getSlot(), firstHalfOfIndices);
+    final SafeFuture<List<List<KZGProof>>> kzgProofs =
+        chainDataClient.getDataColumnSidecarProofs(block.getSlot());
+
+    return reconstructionAsyncRunner
+        .runAsync(
+            () ->
+                dataColumnSidecars.thenCombineAsync(
```
rolfyone: (claude) Bug — thenCombineAsync runs on common pool, not the reconstruction pool 
                                                                                                                                                                       
  runAsync(() -> dataColumnSidecars.thenCombineAsync(kzgProofs, combiner)) submits the lambda to reconstructionAsyncRunner, but that lambda just configures the future — the actual heavy KZG work inside the combiner runs on ForkJoinPool.commonPool() 
  (the default for thenCombineAsync without an executor arg). Should be:                                                                                                                                                                                 
  dataColumnSidecars.thenCombineAsync(kzgProofs, combiner, reconstructionAsyncRunner.getExecutor())                                                                                                                                                      
  or the wrapping runAsync serves no purpose.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
```diff
@@ -0,0 +1,293 @@
...
+                    kzgProofs,
+                    (sidecars, proofs) -> {
+                      if (sidecars.isEmpty() || proofs.isEmpty()) {
+                        return emptyResult();
+                      }
+
+                      final List<BlobAndCellProofs> blobAndCellProofsList =
+                          constructBlobAndCellProofsList(sidecars, proofs);
```
rolfyone: I think this is ok, given that columns should be ordered (iirc), but noting here as it was in the claude review...
  
(claude) Bug — constructBlobAndCellProofsList assumes sidecars are ordered by column index                                                                                                                                                                      
  sidecars.get(j) at line ~196 assumes the j-th element of the list has column index j. Nothing in the storage API guarantees getDataColumnSidecars returns results in column-index order. If the order isn't guaranteed, the blob reconstruction is     
  silently wrong (cells get assembled in the wrong order).

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
```diff
@@ -0,0 +1,293 @@
...
+    this.chainDataClient = chainDataClient;
+    this.reconstructionAsyncRunner = reconstructionAsyncRunner;
+    this.isSuperNodeSupplier = isSuperNodeSupplier;
+    this.spec = spec;
+    this.dataColumnSidecarExtensionRetentionEpochs =
+        UInt64.valueOf(dataColumnSidecarExtensionRetentionEpochs);
+    this.specConfigFulu =
+        SpecConfigFulu.required(spec.forMilestone(SpecMilestone.FULU).getConfig());
+    this.halfColumns = specConfigFulu.getNumberOfColumns() / 2;
+    this.recoveryTasks = new ConcurrentHashMap<>();
```
rolfyone, line 95: (claude) Concern — recoveryTasks is unbounded                                                                                                                                                                                                                   
  recoveryTasks grows forever if any caller fails to call onRequestCompleted. There's no TTL, no max-size, no eviction. In a long-running node this is a slow memory leak. At minimum there should be a log warning if the map grows large.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
```diff
@@ -0,0 +1,293 @@
...
+    final UInt64 lastPrunableSlot =
+        spec.computeStartSlotAtEpoch(pruningBoundaryEpoch).minusMinZero(1);
+    sidecarArchivePrunableChannel.onSidecarArchivePrunableSlot(lastPrunableSlot);
+  }
+
+  private int getNextTaskId() {
+    final int nextTaskId = this.nextTaskId.getAndIncrement();
+    if (this.nextTaskId.get() > (Integer.MAX_VALUE - 1000)) {
+      this.nextTaskId.set(0);
+    }
```
rolfyone: (claude) Nit — task ID wrap-around is non-atomic
```
  if (this.nextTaskId.get() > (Integer.MAX_VALUE - 1000)) {                                                                                                                                                                                              
      this.nextTaskId.set(0);                              
  }                                                                                                                                                                                                                                                      
```
  The get() + set(0) is not atomic. Two threads could both observe the condition true and both reset. Could use compareAndSet or just let AtomicInteger overflow naturally (wrapping to negative is fine for a map key, or use Math.abs).

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
```diff
@@ -0,0 +1,293 @@
...
+      for (int j = 0; j < halfColumns; j++) {
+        blobProofs.add(sidecars.get(j).getKzgProofs().get(blobIndex).getKZGProof());
+      }
+      for (int j = 0; j < halfColumns; j++) {
+        blobProofs.add(proofs.get(j).get(blobIndex));
+      }
+
+      final BlobSchema blobSchema =
+          SchemaDefinitionsFulu.required(spec.atSlot(firstSidecar.getSlot()).getSchemaDefinitions())
+              .getBlobSchema();
```
rolfyone: (claude) Nit — blobSchema resolved inside loop unnecessarily                                                                                                                                                                                                    
  In constructBlobAndCellProofsList, blobSchema is derived from spec.atSlot(firstSidecar.getSlot()) on every loop iteration but it's constant for the whole method. Hoist it above the loop.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
```diff
@@ -0,0 +1,293 @@
...
+        metricsSystem.createCounter(
+            TekuMetricCategory.BEACON,
+            "data_column_sidecar_archive_reconstruction_failures_total",
+            "Total number of failed data column sidecar archive reconstructions");
+  }
+
+  @Override
+  public SafeFuture<Optional<DataColumnSidecar>> reconstructDataColumnSidecar(
+      final SignedBeaconBlock block, final UInt64 index, final int requestId) {
+    final Map<SlotAndBlockRoot, SafeFuture<ReconstructionResult>> slotAndBlockRootSafeFutureMap =
```
rolfyone, line 125: probably worth a comment for future us...

(claude)   Nit — first-half index requests return Optional.empty() silently                                                                                                                                                                                       
  reconstructDataColumnSidecar with an index < halfColumns silently returns empty (the ReconstructionResult map only covers the second half). This is documented in the test but could surprise callers. An early guard/assertion at the entry point     
  would make the contract explicit.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImplTest.java
```diff
@@ -0,0 +1,272 @@
...
+        reconstructor
+            .reconstructDataColumnSidecar(block, UInt64.valueOf(halfColumns), requestId)
+            .join();
+
+    assertThat(result).isEmpty();
+  }
+
+  @TestTemplate
+  public void shouldReconstructSecondHalfSidecars() {
+    final SignedBeaconBlock block = createBlockWithBlobs(UInt64.ONE);
```
rolfyone, line 154: im less worried about this one... mentioned as was part of claude generated review.

(claude)   Concern — shouldReconstructSecondHalfSidecars may not test actual reconstruction
  createFirstHalfSidecars returns random-data sidecars and createSecondHalfProofs returns random KZG proofs. If constructDataColumnSidecars calls into the real KZG library, reconstruction of invalid data will either throw (caught by exceptionally → 
  emptyResult) making the test fail, or silently produce garbage. The test should either use a real KZG round-trip with valid blob data, or mock DataColumnSidecarUtil to verify the wiring without invoking KZG.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
zilm13: We have a guaranteed cleanup in final block, clarified with comment

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java
zilm13: added the check to the list

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImplTest.java
zilm13: used mock for DataColumnSidecarUtil

zilm13:
@rolfyone I've addressed all feedback

rolfyone:
ran another review round, I dont know how i missed this sorry ...

Findings
[P1] Proofs are read but never produced in the target branch. The reconstructor now depends on getDataColumnSidecarProofs(slot) here, but repo-wide search only finds addDataColumnSidecarsProofs in DAO/updater plumbing and tests, not in any production archival/pruning path. Once a second-half column is actually pruned, this returns List.of(), the size check fails, and reconstruction returns empty. See [DataColumnSidecarArchiveReconstructorImpl.java:145](https://github.com/Consensys/teku/blob/8ed7d2670a120b92217ca280293f6337eda6a954/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java#L145) and [CombinedChainDataClient.java:264](https://github.com/Consensys/teku/blob/8ed7d2670a120b92217ca280293f6337eda6a954/storage/src/main/java/tech/pegasys/teku/storage/client/CombinedChainDataClient.java#L264).

[P1] The new prunable event has no production subscriber. onNewFinalizedCheckpoint publishes SidecarArchivePrunableChannel, but git grep on the PR head shows only the publisher, the interface, and tests. So the retention-boundary signal never causes archive pruning, and the new reconstructor is effectively wired live without the pruning side of the workflow. See [DataColumnSidecarArchiveReconstructorImpl.java:295](https://github.com/Consensys/teku/blob/8ed7d2670a120b92217ca280293f6337eda6a954/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSidecarArchiveReconstructorImpl.java#L295) and [SidecarArchivePrunableChannel.java:20](https://github.com/Consensys/teku/blob/8ed7d2670a120b92217ca280293f6337eda6a954/storage/api/src/main/java/tech/pegasys/teku/storage/api/SidecarArchivePrunableChannel.java#L20).

[P2] By-range reconstruction can under-serve valid pruned requests. The range handler asks storage for at most maxRequestDataColumnSidecars raw identifiers, then expands only the slots present in that limited result into requested reconstructed columns. With 64 retained first-half columns per slot, a request for one pruned column across 128 slots can load identifiers for only about two slots, then finish early. See [ChainStorage.java:444](https://github.com/Consensys/teku/blob/8ed7d2670a120b92217ca280293f6337eda6a954/storage/src/main/java/tech/pegasys/teku/storage/server/ChainStorage.java#L444) and [DataColumnSidecarsByRangeMessageHandler.java:303](https://github.com/Consensys/teku/blob/8ed7d2670a120b92217ca280293f6337eda6a954/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/DataColumnSidecarsByRangeMessageHandler.java#L303).

Assumption
I’m assuming PR #10508 is intended to be functional when merged, not dependent on a follow-up PR for the proof writer and pruning subscriber.

zilm13:
@rolfyone 
I've upstreamed latest master.
Regarding your findings:
1,2  will be added in the follow-up, see meta https://github.com/Consensys/teku/issues/10095
`DataColumnSidecarExtensionPruner, which prunes 64-127 indices over data-column-sidecar-extension-retention-epochs if the node is supernode and writes kzgProofs to DB`
3, I'd keep it as is, I've added a comment for clarity. It's side-effect but it's for good:
```
 A request like count=16384, columns=[72] is valid, and if those are pruned supernode columns, fully serving it could mean reconstructing 16,384 distinct slots. Because
  reconstruction is per slot, that is much worse than serving 16,384 already-stored sidecars or reconstructing many columns from a small number of slots.

  Leaving the current behavior as a deliberate best-effort throttle can make sense
```

rolfyone:
claude has probably one issue thats worth pondering (number 1)...

  1. DataColumnSidecarArchiveReconstructorImpl.java:245-247 — archived data is fetched by slot only, ignoring the block root, so a competing block at the same slot breaks (or corrupts) reconstruction.
  createTask calls chainDataClient.getDataColumnSidecars(block.getSlot(), firstHalfOfIndices) and getDataColumnSidecarProofs(block.getSlot()), then combines the returned cells/proofs with this specific block's commitments and header. But the archive
  stores gossip-validated sidecars for any block at a slot, not just the canonical one (ChainStorage.onNewSidecar → database.addSidecar, no canonicality check), and getDataColumnIdentifiers(slot) scans all block roots at that slot.
  - Failure scenario A (availability): an orphaned block at the same slot also has its first-half sidecars stored → getDataColumnSidecars returns >halfColumns sidecars → the sidecars.size() != halfColumns guard (line 272) returns emptyResult() →
  reconstruction permanently fails for that slot, even for the legitimate canonical block, and the supernode cannot serve those columns.
  - Failure scenario B (correctness): proofs are keyed by slot as a single value (getColumnDataColumnSidecarsProofsBySlot), so if a non-canonical block wrote proofs last, getDataColumnSidecarProofs(slot) returns proofs inconsistent with the
  canonical block's cells → reconstructed sidecars whose proofs won't match the block's commitments (peers reject them). Consider filtering the fetched sidecars/proofs by block.getRoot().

  2. (minor, efficiency) DataColumnSidecarArchiveReconstructorImpl.java:223-227 — dedup is per-requestId only. Two concurrent RPC requests (different request IDs) for the same popular block each run a full, expensive KZG reconstruction. The class
  comment scopes dedup to per-request intentionally to bound memory, so this is a documented trade-off — worth a short-lived shared result cache if reconstruction of the same block by many peers proves hot.

  3. (minor, cleanup) DataColumnSidecarArchiveReconstructorImpl.java:358-363 — redundant store lookups in isSidecarPruned. chainDataClient.getFinalizedBlockSlot() is called twice (.isEmpty() then .get()) and getStore() is queried repeatedly; hoist
  into locals. This runs per requested column on every by-range/by-root request, so it's on a mildly hot path.

  Findings 2 and 3 are low-severity; finding 1 is the one I'd ask the author to confirm — the mechanism is real (verified against the write path and the by-slot query), though it only triggers when two blocks share a slot in the archive, which the
  size check partially masks by failing closed rather than serving wrong data.

Gasila5:
Bc1qz2yt84k3vlhc8nrr57m05rr4dwfyawj2le9atd

zilm13:
@rolfyone 
1. It calls `CombinedChainDataClient.getDataColumnSidecars()` which returns *only canonical sidecars*, even somehow we get an exception, the task will return an `emptyResult()`. But anyway, I've added a check and a test.
2. This is designed for old finalized blocks, there is a very small chance same block will be requested closely, so keeping some cache which consumes memory doesn't look good for me.
3. Good catch, fixed!

rolfyone (approved):
LGTM

PR #10509, opened 2026-03-19
Updated CreateAttestationData to pass optional index down

### LINKED ISSUE #10433: [beacon-api] GetAttestationData - gloas optional committee index from api
#10404 fixed the interface but at some point we need to address the handling when the optional parameter is not set.

rolfyone:
- if the committee index is used in gloas specifically, it'll currently allow and use what was specified.
 - if we're in gloas and dont specify index, it'll be determined from the inclusion of payload (which is the gloas spec)

 I was on the fence about whether to ignore the optional passed in, could be convinced to ignore it in gloas and just always compute...

 fixes #10433

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches a widely used API (`createAttestationData`) and changes default/index-selection behavior for Gloas, which could alter produced `AttestationData.index` and downstream attestation signing/publishing if callers omit or mis-handle the optional parameter.
> 
> **Overview**
> Updates the `createAttestationData` flow end-to-end (REST handler → `ValidatorDataProvider` → `ValidatorApiChannel`/handlers → remote typedef client) to accept an *optional* `committeeIndex` instead of an `int`, allowing Gloas-era requests to omit the query parameter.
> 
> Adjusts committee index selection: base `ValidatorApiHandler` defaults to `0` when not provided, while `ValidatorApiHandlerGloas` now either uses the supplied index or computes `0/1` based on payload/fullness (with added debug logging), and updates tests accordingly, including remote request building to only send `committee_index` when present.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b4e0afe97699f290b7aa1a46637400e84c8560b4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/beacon/validator/src/integration-test/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandlerIntegrationTest.java
+++ b/beacon/validator/src/integration-test/java/tech/pegasys/teku/validator/coordinator/ValidatorApiHandlerIntegrationTest.java
@@ -248,9 +248,8 @@ public class ValidatorApiHandlerIntegrationTest {
     assertThat(epochBoundaryBlock).isNotNull();
     final Checkpoint expectedTarget = new Checkpoint(targetEpoch, epochBoundaryBlock.getRoot());
 
-    final int committeeIndex = 0;
     final SafeFuture<Optional<AttestationData>> result =
-        handler.createAttestationData(targetSlot, committeeIndex);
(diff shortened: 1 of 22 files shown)


rolfyone:
theres a different scenario to cater for that makes this simpler, closing and will redo

PR #10532, opened 2026-03-25
Fix OkHttp readTimeout not being overridden for per-request Engine API timeouts

### LINKED ISSUE #10531: Engine API requests fail with read timeout when only callTimeout is overridden
## Description

When the Engine API client overrides the timeout for specific calls (e.g. `engine_newPayloadV*`, `engine_getPayloadV*`), only `callTimeout` is set on the OkHttp client — the `readTimeout` is left at its default value. This means that even though the overall call timeout is extended, the underlying socket read can still time out earlier than expected.

## Steps to Reproduce

1. Configure Teku with an Execution Layer client
2. The EL client takes longer than OkHttp's default read timeout (10s) to respond to an Engine API call, but within the configured call timeout
3. The request fails with a `SocketTimeoutException` (read timeout) even though the call timeout has not been reached

## Expected Behavior

When a custom timeout is provided for an Engine API call, both the overall call timeout **and** the read timeout should be extended, so the request is not prematurely terminated by a socket read timeout.

## Actual Behavior

Only `callTimeout` is overridden via `httpClient.newBuilder().callTimeout(timeout)`. The `readTimeout` remains at OkHttp's default (10 seconds), causing requests to fail with a read timeout before the call timeout is reached.

## Impact

This can cause intermittent failures on Engine API calls (e.g. `engine_newPayloadV*`) when the Execution Layer client is under load or processing large payloads, leading to missed block proposals or delayed chain head updates.

## Affected Code

`OkHttpHttpExecutionEngineClient.makeRequest()` in `ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/OkHttpHttpExecutionEngineClient.java`

## Fix

Set `readTimeout(timeout)` alongside `callTimeout(timeout)` when building the per-request OkHttp client override.

lucassaldanha:
## PR Description

Our per-request override of timeout did not include the readTimeout(), leaving readTimeout at the default 1s. This caused issues for the methods that required longer timeouts.

## Fixed Issue(s)
fixes #10531

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Adjusts HTTP timeout behavior for Engine API requests, which can change how long the client waits before failing calls. Impact is limited to the OkHttp execution engine client but affects networking reliability and latency handling.
> 
> **Overview**
> Ensures per-request Engine API timeouts in `OkHttpHttpExecutionEngineClient` override both OkHttp `callTimeout` *and* `readTimeout`, fixing cases where the default read timeout could prematurely abort long-running requests.
> 
> Refactors the custom-timeout path into `callWithCustomTimeout(...)` and updates the related unit test name to reflect the new helper.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9cef2ffca2909f42525fb444ccdffb7eb35a110. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/OkHttpHttpExecutionEngineClient.java
+++ b/ethereum/executionclient/src/main/java/tech/pegasys/teku/ethereum/executionclient/OkHttpHttpExecutionEngineClient.java
@@ -79,7 +79,8 @@ public class OkHttpHttpExecutionEngineClient extends OkHttpExecutionEngineClient
     final SafeFuture<Response<T>> future = new SafeFuture<>();
     final Call call;
     if (timeout.toMillis() != httpClient.callTimeoutMillis()) {
-      call = httpClient.newBuilder().callTimeout(timeout).build().newCall(httpRequest);
+      call =
+          httpClient.newBuilder().callTimeout(timeout).readTimeout(timeout).build().newCall(httpRequest);
     } else {
       call = httpClient.newCall(httpRequest);
     }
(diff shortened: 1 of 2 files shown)


rolfyone (approved):
may be better to wrap in a fn as discussed but the idea is fine

Issue #10208, opened 2025-12-08
SingleBlockProvider for Fulu


gfukushima:
just created this sub-issue to associate to PR

PR #10533, opened 2026-03-26
SingleBlockProviderFulu with soft and hard limit for sampled blocks

gfukushima:
## Fixed Issue(s)

Part of #10208 
## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes Fulu data-availability sampling memory bounds and block lookup during import; hard eviction can fail in-flight DA checks, but defaults are sized for forward sync batching.
> 
> **Overview**
> Refactors the Fulu **DAS basic sampler** into a `DasSamplerBasic` interface plus `DasSamplerBasicImpl`, and adds a **bounded in-memory cache** of per-block sampling trackers (configurable via `SyncConfig` / hidden `--Xp2p-max-concurrently-sampled-blocks`, default **128**). When the cache is full, it **soft-evicts** oldest completed trackers and, if still over **4×** the limit, **hard-evicts** oldest incomplete ones (failing their completion futures).
> 
> Trackers now optionally hold the **`SignedBeaconBlock`** so peers can be served blocks that are still undergoing column sampling. **`SingleBlockProviderResolver`** chains blob-sidecar pool lookup with the sampler; **`RecentBlocksFetchService`** skips redundant block fetches when the sampler already has the root.
> 
> Adds a **`das_recently_sampled_blocks_size`** beacon metric and wires the sampler through sync/beacon chain factories and tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67c5f3f159e92fa78e18a8ee54ac948b79cde425. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,357 @@
...
+              }
+            });
+  }
+
+  private DataColumnSamplingTracker getOrCreateTracker(final UInt64 slot, final Bytes32 blockRoot) {
+    final DataColumnSamplingTracker tracker;
+    final boolean created;
+    synchronized (this) {
+      makeRoomForNewTracker();
+      created = !recentlySampledColumnsByRoot.containsKey(blockRoot);
```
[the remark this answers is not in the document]
gfukushima: fixed in [9d7c3c6](https://github.com/Consensys/teku/pull/10533/commits/9d7c3c6f2e07401c5a2354cc40dd2f303dc09e87)

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/SyncConfig.java
```diff
@@ -123,6 +132,7 @@ public static class Builder {
     private Integer forwardSyncMaxBlocksPerMinute = DEFAULT_FORWARD_SYNC_MAX_BLOCKS_PER_MINUTE;
     private Integer forwardSyncMaxBlobSidecarsPerMinute =
         DEFAULT_FORWARD_SYNC_MAX_BLOB_SIDECARS_PER_MINUTE;
+    private final Integer maxRecentlySampledBlocks = DEFAULT_MAX_RECENTLY_SAMPLED_BLOCKS;
```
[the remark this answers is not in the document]
gfukushima: fixed in [d8f21bf](https://github.com/Consensys/teku/pull/10533/commits/d8f21bfa51cc77e35f8a744e47e69a8e8923b400)

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -596,6 +599,15 @@ protected SafeFuture<?> initialize() {
...
     final ValidatorIsConnectedProvider validatorIsConnectedProvider =
         new ValidatorIsConnectedProviderReference(() -> proposersDataManager);
+
+    final SingleBlockProvider singleBlockProviderResolver =
+        new SingleBlockProviderResolver(
+            (blockRoot) -> blockBlobSidecarsTrackersPool.getBlock(blockRoot),
+            (blockRoot) ->
+                dasSamplerBasic.containsBlock(blockRoot)
+                    ? blockBlobSidecarsTrackersPool.getBlock(blockRoot)
+                    : Optional.empty());
```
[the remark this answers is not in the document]
zilm13: @gfukushima is it something to be fixed in the next PRs?

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/gossip/blocks/RecentBlocksFetchService.java
```diff
@@ -106,6 +112,10 @@ public void requestRecentBlock(final Bytes32 blockRoot) {
       // We already have this block, waiting for blobs
       return;
     }
+    if (dasBasicSampler.containsBlock(blockRoot)) {
```
zilm13, line 115: Shouldn't we put `containsBlock` in BlockEventsListener interface and check it both for dataColumnSidecars and blobSidecars in router? Maybe add an extra interface for clarity or rename this one but it looks like we are doing the same job here, isn't it?

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,359 @@
...
+    this.rpcFetchDelayProvider = rpcFetchDelayProvider;
+    this.spec = spec;
+    this.asyncRunner = asyncRunner;
+    this.custody = custody;
+    this.retriever = retriever;
+    this.custodyGroupCountManager = custodyGroupCountManager;
+    this.recentChainData = recentChainData;
+    this.halfColumnsSamplingCompletionEnabled = halfColumnsSamplingCompletionEnabled;
+    this.maxRecentlySampledBlocks = maxRecentlySampledBlocks;
+    this.recentlySampledColumnsByRoot = new LinkedHashMap<>(maxRecentlySampledBlocks);
```
zilm13, line 88: it's not concurrent safe

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,359 @@
...
+    metricsSystem.createGauge(
+        BEACON,
+        "das_recently_sampled_blocks_size",
+        "DAS recently sampled blocks size",
+        () -> {
+          synchronized (this) {
```
zilm13: You are adding synchronization to this method. Is it really necessary? We are trying to avoid adding synchronization whenever it's possible in new classes and update old code because synchronization is slow and could be a reason for locks/stalling. Its a step backwards.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,359 @@
...
+  @Override
+  public void onSlot(final UInt64 slot) {
+    final UInt64 firstNonFinalizedSlot =
+        spec.computeStartSlotAtEpoch(recentChainData.getFinalizedEpoch()).increment();
+    synchronized (this) {
+      recentlySampledColumnsByRoot
+          .values()
+          .removeIf(
+              tracker -> {
+                if (tracker.slot().isLessThan(firstNonFinalizedSlot)
```
zilm13, line 318: there were several useful comments here, i'd restore it

zilm13 (commented):
I'm not sure all these changes required especially with loosing concurrency safety traded to synchronization. Maybe just adjust `makeNewRoomForTrackers` in the old code according to new requirements? I guess we don't care much about blocks removed in `removeAllForBlock` because those are imported/rejected.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
gfukushima: chosen so can make use of the fifo during eviction, that's why I added some synchronism when this Map is modified

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,359 @@
...
+        recentlySampledColumnsByRoot.entrySet().iterator();
+    while (it.hasNext() && recentlySampledColumnsByRoot.size() >= maxRecentlySampledBlocks) {
+      final DataColumnSamplingTracker tracker = it.next().getValue();
+      if (tracker.completionFuture().isDone()) {
+        it.remove();
+      }
+    }
+    // Hard cap: if we're at 4x the limit even after evicting completed trackers,
+    // force-evict the oldest incomplete ones to prevent unbounded growth.
+    final int hardLimit = maxRecentlySampledBlocks * 4;
```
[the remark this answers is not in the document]
gfukushima: users would be blowing his node doing this

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
gfukushima: restored

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
gfukushima: well, in this case we want to make sure that the LinkedHashMap is atomically handled, for the metrics i don't think this would be extremelly necessary but also we want metrics that provide accurate data don't we?

zilm13: I mean in general. It's just a first synchronization block of several. If we are going with concurrent safe setup without locks we could (rarely) get not contains and we will just redownload it again. It's not a big trade-off compared to locking everything here I think. 
Maybe also check it in recentChainData in the end of checks in`RecentBlocksFetchService` maybe it's already there? Does it make sense? It should be fast to check there.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
```diff
@@ -608,6 +611,11 @@ protected SafeFuture<?> initialize() {
     final ExecutionPayloadProvider executionPayloadProvider =
         createExecutionPayloadProvider(storageQueryChannel);
 
+    final SingleBlockProvider singleBlockProviderResolver =
```
zilm13, line 629: Could you please add some comment clarifying why do you need these sources for the block, why it's not recentChainData or smth like this

teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
```diff
@@ -321,6 +321,15 @@ The network interface(s) on which the node listens for P2P communication.
       arity = "1")
   private Integer forwardSyncMaxDistanceFromHead;
 
+  @Option(
+      names = {"--Xp2p-max-recently-sampled-blocks"},
+      paramLabel = "<NUMBER>",
+      showDefaultValue = Visibility.ALWAYS,
+      description = "Maximum number recent blocks we should sample when syncing.",
```
zilm13: The description is a bit confusing, same with variable name. Maybe `max-concurrently-sampled-blocks` or smth like this and update description + var name.

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicTest.java
```diff
@@ -73,7 +74,7 @@ public class DasSamplerBasicTest {
   private CurrentSlotProvider currentSlotProvider;
   private final DataStructureUtil dataStructureUtil = new DataStructureUtil(0, SPEC);
 
-  private DasSamplerBasic sampler;
+  private DasSamplerBasicImpl sampler;
```
zilm13, line 89: New pruning logic is very complex and we don't have tests for it. At the same time the size a class input, so we could easily pass something like 2 and have some good tests. We definitely need to test new behavior.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSamplingTracker.java
```diff
@@ -34,7 +39,9 @@ record DataColumnSamplingTracker(
     AtomicBoolean rpcFetchInProgress,
     SafeFuture<List<UInt64>> completionFuture,
     AtomicBoolean fullySampled,
-    Optional<Integer> earlyCompletionRequirementCount) {
+    Optional<Integer> earlyCompletionRequirementCount,
+    AtomicReference<Optional<SignedBeaconBlock>> block,
```
zilm13, line 44: As we don't need block itself, we could skip storing a reference it and just make `AtomicBoolean` flag `blockSeen`.  And leave function `blockSeen(final Bytes32 blockRoot)`. Smth like this

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  private DataColumnSamplingTracker getOrCreateTracker(final UInt64 slot, final Bytes32 blockRoot) {
+    final DataColumnSamplingTracker existing = recentlySampledColumnsByRoot.get(blockRoot);
+    if (existing != null) {
+      return existing;
+    }
+    final boolean[] created = {false};
+    final DataColumnSamplingTracker tracker =
+        recentlySampledColumnsByRoot.computeIfAbsent(
```
zilm13: 2 steps still doesn't give you any guarantees against race. To have guarantees you could go this way after existing check. `==` will check that you get the same object skipping equals:

```java
    final DataColumnSamplingTracker dataColumnSamplingTracker = DataColumnSamplingTracker.create(
            slot,
            blockRoot,
            custodyGroupCountManager,
            halfColumnsSamplingCompletionEnabled
                    ? Optional.of(
                    SpecConfigFulu.required(spec.atSlot(slot).getConfig())
                            .getNumberOfColumns()
                            / 2)
                    : Optional.empty());
    final DataColumnSamplingTracker trackerInserted =
        recentlySampledColumnsByRoot.computeIfAbsent(
            blockRoot,
            __ -> dataColumnSamplingTracker);
    if (dataColumnSamplingTracker == trackerInserted) {
      makeRoomForNewTracker();
      onFirstSeen(slot, blockRoot, dataColumnSamplingTracker);
    }
    return trackerInserted;
```

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+  }
+
+  @Override
+  public void onSlot(final UInt64 slot) {
+    final UInt64 firstNonFinalizedSlot =
+        spec.computeStartSlotAtEpoch(recentChainData.getFinalizedEpoch()).increment();
+    recentlySampledColumnsByRoot
+        .values()
+        .removeIf(
```
zilm13, line 361: if you want to make this cache useful you'd need probably to remove this cleanup, isn't it?

zilm13 (commented):
Much better keeping it without `synchronized`!
I've left some comments we need to address before merging this.

services/beaconchain/src/main/java/tech/pegasys/teku/services/beaconchain/BeaconChainController.java
gfukushima: recentChainData does rely on this interface to obtain recently validated blocks to serve them if we get an RPC request, see: https://github.com/Consensys/teku/commit/9438a285ed4bc9ab6e3b519ed351abaac090dda3, I'm adding a comment explaining this.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DataColumnSamplingTracker.java
gfukushima: as we would use this to return the block for the BlockByRootRequest requests we do need the block I believe

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
gfukushima: I think this is fine as it is, this removes from the tracker blocks that have been fully imported which makes sense, once it's fully imported we cna retrieve these blocks from recentChainData

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
gfukushima: changed to follow this suggestion

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicTest.java
gfukushima: Not sure I understood your what you mean by "new pruning logic" I haven't added anything to this class in terms of pruning. I can add some tests to this class though

zilm13 (dismissed):
Great work! Thank you

ethereum/statetransition/src/test/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicTest.java
zilm13: You've covered it with tests `getOrCreateTracker_shouldEvictOldestCompletedTrackerAtSoftLimit`, `getOrCreateTracker_shouldPreferOldestCompletedWhenSoftEvicting`, `getOrCreateTracker_shouldForceEvictOldestIncompleteAtHardLimit`, it's what I wanted.

teku/src/main/java/tech/pegasys/teku/cli/options/P2POptions.java
gfukushima: fixed

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
```diff
@@ -0,0 +1,443 @@
...
+            slot,
+            blockRoot,
+            custodyGroupCountManager,
+            halfColumnsSamplingCompletionEnabled
+                ? Optional.of(
+                    SpecConfigFulu.required(spec.atSlot(slot).getConfig()).getNumberOfColumns() / 2)
+                : Optional.empty());
+    final DataColumnSamplingTracker trackerInserted =
+        recentlySampledColumnsByRoot.computeIfAbsent(blockRoot, __ -> dataColumnSamplingTracker);
+    if (dataColumnSamplingTracker == trackerInserted) {
```
zilm13:
```suggestion
    if (isStaledTracker(blockRoot, trackerInserted)) {
      return recentlySampledColumnsByRoot.get(blockRoot);
    }

    makeRoomForNewTracker();
    onFirstSeen(slot, blockRoot, dataColumnSamplingTracker);
    return trackerInserted;
```

and remove warning supressions from the top of this method

zilm13 (dismissed):
LGTM just one nit to avoid double suppresions on what we've already solved there.
Also the merge could clash with #10825, whoever merges first, other will have a conflict.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasicImpl.java
gfukushima: makes sense, adjusted

lucassaldanha (approved):

PR #10537, opened 2026-03-26
Implement `eth/v1/beacon/execution_payload_envelopes`

### LINKED ISSUE #10416: [beacon-api] `POST /eth/v1/beacon/execution_payload_envelope` added
https://github.com/ethereum/beacon-APIs/pull/580

akronim26:
## PR Description

This PR implements `eth/v1/beacon/execution_payload_envelope` endpoint. It uses overloading of `publishSignedExecutionPayload` function to support the query param `block_validation`.

## Fixed Issue(s)

Fixes #10416 

## Documentation

## Changelog

---

> [!NOTE]
> **High Risk**
> Touches consensus-critical execution payload validation, import, and gossip paths for Gloas ePBS, including equivocation checks and blinded reconstruction from EL cache.
> 
> **Overview**
> Adds **`POST /eth/v1/beacon/execution_payload_envelopes`** so validators can broadcast Gloas signed execution payload envelopes, with optional **`broadcast_validation`** (`gossip`, `consensus`, `consensus_and_equivocation`) and **`Eth-Execution-Payload-Blinded`** selecting **`SignedExecutionPayloadEnvelopeContents`** vs **`SignedBlindedExecutionPayloadEnvelope`**.
> 
> Publishing now runs **`validateAndImportExecutionPayloadForBroadcast`**, returns **200** on successful import vs **202** when broadcast succeeds but import fails, and supports **unblinding** blinded envelopes from the cached EL `getPayload` result before gossip. Remote validators call the new route (signed envelopes are sent **blinded** when possible); failover routing pins publish to the beacon node that created the payload envelope.
> 
> Introduces **`SignedExecutionPayloadEnvelopeContents`** SSZ/schema and refactors data-column sidecar construction to accept blobs/proofs from either the cache or envelope contents.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbe7182f42f0f6a32881e0aa06b2e21c1b1fd075. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


akronim26:
@rolfyone, please review whenever you are free

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostExecutionPayloadEnvelope.java
```diff
@@ -0,0 +1,116 @@
...
+  private AsyncApiResponse processPublishSignedExecutionPayloadResult(
+      final PublishSignedExecutionPayloadResult result) {
+    return result
+        .getRejectionReason()
+        .map(
+            rejectionReason -> {
+              if (result.isPublished()) {
+                return AsyncApiResponse.respondWithCode(SC_ACCEPTED);
+              }
+              return AsyncApiResponse.respondWithError(SC_INTERNAL_SERVER_ERROR, rejectionReason);
```
rolfyone: generally we'd just 400 here, 500 is typically a coding bug...

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java
```diff
@@ -50,14 +52,20 @@ public ExecutionPayloadPublisherGloas(
   @Override
   public SafeFuture<PublishSignedExecutionPayloadResult> publishSignedExecutionPayload(
       final SignedExecutionPayloadEnvelope signedExecutionPayload) {
+    return publishSignedExecutionPayload(signedExecutionPayload, Optional.empty());
+  }
```
rolfyone, line 73: seems redundant given this is what the default of the base interface does - can remove this one.

data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostExecutionPayloadEnvelope.java
```diff
@@ -0,0 +1,116 @@
...
+        .build();
+  }
+
+  @Override
+  public void handleRequest(final RestApiRequest request) throws JsonProcessingException {
+    final SignedExecutionPayloadEnvelope signedExecutionPayloadEnvelope = request.getRequestBody();
+    final Optional<BroadcastValidationLevel> broadcastValidationLevel =
+        request
+            .getOptionalQueryParameter(PARAMETER_BROADCAST_VALIDATION)
+            .map(BroadcastValidationParameter::toInternal);
```
rolfyone: it might be worth putting an error message here for when broadcast validation is set, but we can't map what was passed in - this would help us debug if its ever an invalid value...

Basically if its a set parameter and we get an empty optional, we could 400 with a message to say that the broadcast parameter could not be parsed (bad_request iirc)

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/ExecutionPayloadGossipValidator.java
```diff
@@ -74,13 +85,24 @@ public SafeFuture<InternalValidationResult> validate(
...
-                    .map(SafeFuture::completedFuture)
-                    .orElseGet(
-                        () ->
-                            performWithStateValidation(signedExecutionPayloadEnvelope)
-                                .thenApply(result -> markAsSeen(result, envelope))));
+            maybeResult -> {
+              if (maybeResult.isPresent()) {
+                return SafeFuture.completedFuture(maybeResult.get());
+              }
+
```
rolfyone, line 96: possibly add a comment here that we're not doing extra gossip validation currently for consensus or consensus_and_equivocation - probably a `// TODO GLOAS` - we're likely to need to do something here with it at some point.

beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java
```diff
@@ -50,14 +52,20 @@ public ExecutionPayloadPublisherGloas(
...
     return executionPayloadManager
-        .validateAndImportExecutionPayload(signedExecutionPayload)
+        .validateAndImportExecutionPayload(
+            signedExecutionPayload, Optional.empty(), broadcastValidationLevel)
         .thenApply(
             result -> {
               final Bytes32 beaconBlockRoot = signedExecutionPayload.getBeaconBlockRoot();
               if (result.isAccept()) {
-                // we publish the execution payload (and data column sidecars) after passing gossip
-                // validation
```
rolfyone, line 60: maybe better to leave this comment in i think

rolfyone (commented):
looks pretty good! Some mostly minor things. Hopefully soon the other PR is merged (beacon-api) with no other changes required...

akronim26:
@rolfyone, please take a look

rolfyone:
i've updated the PR with current master etc, sorry there's been a bunch on.

 - looks like this PR has `broadcast_validation` which i can't see in the api? is that related to a PR on beacon-apis or something?
 - we should add validator required tag to endpoint
 - spec has no 202, so ideally we shouldn't be using 202, or we should be updating beacon-api if its needed.

There's some cursorbot things I haven't fully digested but potentially it might have some valid feedback here too...

akronim26:
Okay @rolfyone, ig there's some confusion: I am following [this file](https://github.com/ethereum/beacon-APIs/pull/580/changes#diff-81ee70f2d3ef064583bf30a5dd912e21e004476334fb825d16e72bc2bd411e12) as this is related to `POST` `execution_payload_envelope` and maybe you are following [this](https://github.com/ethereum/beacon-APIs/pull/580/changes#diff-70f7de196e41377e66d64da0bb5025d42f118937ee89bf86f8d2e78efa6610d0). Correct me if I am wrong thou

rolfyone:
ah ok - so yeh its in an open PR, and I was looking at the committed version thats published on beacon-apis. super confusing. will try to get the existing PR progressed so its clearer, sorry for the confusion...
may take a few days, travelling sorry :/

for reference, was looking at https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/ValidatorRequiredApi/submitPayloadAttestationMessages when i reviewed @akronim26

rolfyone:
now that 580 is merged, re-visiting review. I've updated to hopefully resolve the conflicts and bring it up to date.

Review comments:

Findings
[P1] [PostExecutionPayloadEnvelope.java (line 42)](/Users/paulharris/IdeaProjects/teku/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostExecutionPayloadEnvelope.java:42) and [ValidatorApiMethod.java (line 55)](/Users/paulharris/IdeaProjects/teku/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/apiclient/ValidatorApiMethod.java:55) still use singular /eth/v1/beacon/execution_payload_envelope. The merged [beacon-APIs#580](https://github.com/ethereum/beacon-APIs/pull/580) endpoint is plural: POST /eth/v1/beacon/execution_payload_envelopes, so spec-compliant clients will miss Teku and Teku remote clients will call the wrong path.

[P1] [PostExecutionPayloadEnvelope.java (line 62)](/Users/paulharris/IdeaProjects/teku/data/beaconrestapi/src/main/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostExecutionPayloadEnvelope.java:62) only accepts SignedExecutionPayloadEnvelope, and [PublishSignedExecutionPayloadRequest.java (line 49)](/Users/paulharris/IdeaProjects/teku/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/PublishSignedExecutionPayloadRequest.java:49) only sends Eth-Consensus-Version. Final #580 requires Eth-Execution-Payload-Blinded and either SignedExecutionPayloadEnvelopeContents or SignedBlindedExecutionPayloadEnvelope, so both stateless and stateful spec flows are currently unsupported.

[P1] [ExecutionPayloadGossipValidator.java (line 86)](/Users/paulharris/IdeaProjects/teku/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/ExecutionPayloadGossipValidator.java:86) accepts consensus / consensus_and_equivocation but only runs the existing gossip/state checks before [ExecutionPayloadPublisherGloas.java (line 62)](/Users/paulharris/IdeaProjects/teku/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java:62) broadcasts. The spec says those stronger validation levels must fail with 400 and not broadcast if not satisfied, so this silently downgrades the caller’s requested validation level.

[P2] [PublishSignedExecutionPayloadRequest.java (line 65)](/Users/paulharris/IdeaProjects/teku/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/PublishSignedExecutionPayloadRequest.java:65) maps a no-body 200 OK through an empty ResponseHandler, then returns notImported("UNKNOWN") at line 77. The test only checks isPublished(), so remote 200 success is misclassified as the 202-style result.

lmk @akronim26 if you're not able to action these, sorry its taken so long for the api merge to complete.

akronim26:
Hey @rolfyone, thanks for the detailed review. I'll address all of these in a day.

rolfyone:
> Hey @rolfyone, thanks for the detailed review. I'll address all of these in a day.

if any seem spurious just flag them and i'll check manually, they were generated messages and can be hit and miss

rolfyone:
I ran another pass on this. 

To me the first point sounds potentially like 'expected' but i've shared in case im missing something...

The second sounds potentially worth considering?

[P1] [ExecutionPayloadPublisherGloas.java (line 82)](teku/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java:82) reconstructs blinded publishes from CombinedChainDataClient.getExecutionPayloadByBlockRoot, which reads the hot store, not the block-production cache. In the stateful flow from beacon-APIs#580, the validator signs and posts the blinded envelope before the payload is imported; [Store.java (line 789)](teku/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java:789) returns empty unless the block is already in store. Remote validators hit this directly because [PublishSignedExecutionPayloadRequest.java (line 56)](teku/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/PublishSignedExecutionPayloadRequest.java:56) blinds the signed full envelope before POSTing. Result: valid stateful/remote publishes can 400 with “No cached execution payload envelope found”.

[P1] [ExecutionPayloadGossipValidator.java (line 71)](teku/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/ExecutionPayloadGossipValidator.java:71) treats gossip, consensus, and consensus_and_equivocation the same; there is no final equivocation check before [ExecutionPayloadPublisherGloas.java (line 121)](teku/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java:121) broadcasts on ACCEPT. The upstream endpoint spec says consensus_and_equivocation must reject if the envelope’s beacon block is an equivocation, and Teku’s block path already has that pattern in [BlockBroadcastValidatorImpl.java (line 152)](teku/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java:152). Builders requesting unbundling protection would not get it.

akronim26:
> [P1] [ExecutionPayloadPublisherGloas.java (line 82)](https://github.com/Consensys/teku/pull/teku/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java:82) reconstructs blinded publishes from CombinedChainDataClient.getExecutionPayloadByBlockRoot, which reads the hot store, not the block-production cache. In the stateful flow from beacon-APIs#580, the validator signs and posts the blinded envelope before the payload is imported; [Store.java (line 789)](https://github.com/Consensys/teku/pull/teku/storage/src/main/java/tech/pegasys/teku/storage/store/Store.java:789) returns empty unless the block is already in store. Remote validators hit this directly because [PublishSignedExecutionPayloadRequest.java (line 56)](https://github.com/Consensys/teku/pull/teku/validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/PublishSignedExecutionPayloadRequest.java:56) blinds the signed full envelope before POSTing. Result: valid stateful/remote publishes can 400 with “No cached execution payload envelope found”.

- Fixed it and now the payload is fetched via the cache.

> [P1] [ExecutionPayloadGossipValidator.java (line 71)](https://github.com/Consensys/teku/pull/teku/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/ExecutionPayloadGossipValidator.java:71) treats gossip, consensus, and consensus_and_equivocation the same; there is no final equivocation check before [ExecutionPayloadPublisherGloas.java (line 121)](https://github.com/Consensys/teku/pull/teku/beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java:121) broadcasts on ACCEPT. The upstream endpoint spec says consensus_and_equivocation must reject if the envelope’s beacon block is an equivocation, and Teku’s block path already has that pattern in [BlockBroadcastValidatorImpl.java (line 152)](https://github.com/Consensys/teku/pull/teku/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/BlockBroadcastValidatorImpl.java:152). Builders requesting unbundling protection would not get it.

- Fixed it and now there are equivocation checks.

There was another missing component, the payload publish workflow didn't differentiate between status code `202` (returned when the broadcast is successful but import to db failed) and `200` (returned when both broadcast and import was successful). It returned `200` as soon as the gossip was successfully published. I have fixed it and now the workflow checks for the import process and hence returns the accurate status code.

rolfyone:
This looks pretty good!

Claude had some notes, so will paste below...

  Overall the PR is well-structured. The header-based dispatch, equivocation check, failover pinning, and WebSocket stale-disconnect fix are all done correctly. Four issues below.

  ---
  1. Dead-code fallback in publishSignedExecutionPayload silently claims success (Medium)
  
  File: beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java

  In the private publishSignedExecutionPayload(SignedExecutionPayloadEnvelope, dataColumnSidecarsFuture, broadcastValidationLevel) method:

  return validateAndImportResult
      .importResult()
      .orElseGet(
          () ->
              SafeFuture.completedFuture(
                  ExecutionPayloadImportResult.successful(signedExecutionPayload)))
      .thenApply(importResult -> ...);

  The orElseGet is unreachable: the only way past the !validationResult.isAccept() guard is if validateAndImportExecutionPayloadForBroadcast returned ACCEPT, and the ACCEPT branch in that method always sets importResult to
  Optional.of(importExecutionPayload(...)) — never empty. If a future change breaks that invariant (returning ACCEPT with no import future), this fallback silently synthesises ExecutionPayloadImportResult.successful(...), returning HTTP 200 to the
  VC for a payload that was never actually imported. The fallback should be:

  .orElseThrow(() -> new IllegalStateException("ACCEPT without import future"))

  ---
  2. PostExecutionPayloadEnvelopeTest only exercises the blinded path (Medium)
  
  File: data/beaconrestapi/src/test/java/tech/pegasys/teku/beaconrestapi/handlers/v1/beacon/PostExecutionPayloadEnvelopeTest.java

  All four test cases set request.setRequestBody(envelope.blind(spec)) — every test goes through the instanceof SignedBlindedExecutionPayloadEnvelope branch. The instanceof SignedExecutionPayloadEnvelopeContents branch is completely untested.
  Additionally:

  - No test for the missing Eth-Execution-Payload-Blinded header → 400 error (exercised by isExecutionPayloadBlinded())
  - No test for the 202 response via the unblinded path

  Needed:
  // shouldReturnOkIfSuccessWithUnblindedContents
  // shouldReturnBadRequestIfBlindedHeaderMissing
  // shouldReturnAcceptedIfPublishedButRejectedWithUnblindedContents

  ---
  3. Remote VC silently blinds SignedExecutionPayloadEnvelope before transmission (Low)

  File: validator/remote/src/main/java/tech/pegasys/teku/validator/remote/typedef/handlers/PublishSignedExecutionPayloadRequest.java

  public PublishSignedExecutionPayloadResult submit(
      final SignedExecutionPayloadEnvelope signedExecutionPayload,
      final Optional<BroadcastValidationLevel> broadcastValidationLevel) {
    return submit(signedExecutionPayload.blind(spec), broadcastValidationLevel);
  }

  A full unblinded envelope is blinded before being sent to the BN. This implicitly depends on the BN's execution payload cache to unblind it — if the cache is cleared or the payload was not created by the target BN, the BN will return 400. The
  FailoverValidatorApiHandler tries to route to the originating BN via executionPayloadEnvelopeCreatorCache, but that cache has size 2 and uses remove(), so a double-publish would route the second call to any BN. At minimum this design choice needs
  a comment; ideally the VC should send the envelope as SignedExecutionPayloadEnvelopeContents (with the blobs it already holds) rather than blindifying it.

  ---
  4. Broad .exceptionally in blinded unblinding path masks programming errors (Low)
  
  File: beacon/validator/src/main/java/tech/pegasys/teku/validator/coordinator/publisher/ExecutionPayloadPublisherGloas.java

  return SafeFuture.<SignedExecutionPayloadEnvelope>of(
          () -> executionPayloadFactory.unblindSignedExecutionPayload(signedBlindedExecutionPayload))
      .thenApply(Optional::of)
      .exceptionally(error -> Optional.empty())   // catches ALL exceptions
      .thenCompose(maybeSignedExecutionPayload ->
          maybeSignedExecutionPayload
              .map(...)
              .orElseGet(() -> SafeFuture.completedFuture(
                  PublishSignedExecutionPayloadResult.rejected(
                      ..., "No cached execution payload envelope found for blinded envelope"))));

  The broad .exceptionally swallows any exception from unblindSignedExecutionPayload, including NPEs and ClassCastExceptions. If unblinding fails for a reason other than a cache miss, the VC sees the misleading "No cached execution payload envelope
  found" message instead of the real error. Consider narrowing to the specific cache-miss exception type, or at least logging the actual exception before mapping to the rejection result.

  ---
  Incidental positive: WebSocket stale-disconnect fix is correct
  
  The handleDisconnect change properly synchronises on OkHttpWebSocketExecutionEngineClient.this (the outer class) instead of this (the inner Listener), and filters stale sockets by comparing the incoming WebSocket ws reference before acting. The
  fix is sound.

akronim26:
- Implemented the changes for 1, 2 and 4
- For 3, the current implementation supports the normal single BN case, if we want to avoid sending the blind payload, then we have to fetch the blobs from the BN and then pack them and send to other BNs. This will lead to unnecessary blob round trip.

akronim26:
I think this is an unrelated CI failure. The job fails during `BeaconBlocksByRootIntegrationTest` due to a `NullPointerException`.

rolfyone:
> * For 3, the current implementation supports the normal single BN case, if we want to avoid sending the blind payload, then we have to fetch the blobs from the BN and then pack them and send to other BNs. This will lead to unnecessary blob round trip.

Ok - lets just add a comment somewhere like 
```
  public PublishSignedExecutionPayloadResult submit(
      final SignedExecutionPayloadEnvelope signedExecutionPayload,
      final Optional<BroadcastValidationLevel> broadcastValidationLevel) {
    // Blind before sending: the BN unblinds from its getPayload cache; avoids a blob round-trip.
    return submit(signedExecutionPayload.blind(spec), broadcastValidationLevel);
  }
  ```
  so that makes it clear what's happening, then we can probably merge... It's worth the comment i think just so that its very clear.

rolfyone (approved):
LGTM, theres one bot comment we can always address separately if its valid...

PR #10566, opened 2026-04-15
Bug fix: SlashingProtectedValidatorSource now wraps external signers added via API with slashing protection.

rolfyone:
The bug: When a validator key was added via the validator API using an external signer (URL-based), the addValidator(BLSPublicKey, Optional<URL>) override simply delegated to the underlying source without wrapping the returned signer in a SlashingProtectedSigner. This meant no local slashing protection was applied until the node restarted (at which point keys are loaded through the normal path that does wrap them).

  The fix: The addValidator(BLSPublicKey, Optional<URL>) override now delegates to the shared addSlashingProtectedValidator() helper (extracted from the keystore-based addValidator method), so both add paths wrap the returned signer identically.

  Test added: addValidatorWithSignerUrl_shouldWrapWithSlashingProtectedSigner — verifies that the signer returned from the URL-based add path is a SlashingProtectedSigner by confirming the second signBlock call on the same block+fork fails with a slashing violation.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches validator signing/slashing-protection behavior; while the change is small and well-tested, mistakes here could impact validator safety.
> 
> **Overview**
> Fixes a bug where `SlashingProtectedValidatorSource.addValidator(publicKey, signerUrl)` previously returned the delegate signer unwrapped, leaving validator-api added *external signer* keys without local slashing protection until restart.
> 
> Refactors signer wrapping into a shared helper so both keystore-based and URL-based add paths consistently return a `SlashingProtectedSigner`, adds a regression test covering the URL-based path, and updates the changelog with the user-facing impact/workaround.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42b574a1e46959d80e83515f5b60a4e52df188df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/SlashingProtectedValidatorSource.java
```diff
@@ -53,24 +53,29 @@ public DeleteKeyResult deleteValidator(final BLSPublicKey publicKey) {
...
+  }
+
+  @Override
+  public AddValidatorResult addValidator(
+      final BLSPublicKey publicKey, final Optional<URL> signerUrl) {
+    final AddValidatorResult delegateResult = delegate.addValidator(publicKey, signerUrl);
+    return addSlashingProtectedValidator(publicKey, delegateResult);
+  }
 
+  AddValidatorResult addSlashingProtectedValidator(
```
tbenr: maybe this can be `private`

tbenr (dismissed):
LGTM

validator/client/src/main/java/tech/pegasys/teku/validator/client/loader/SlashingProtectedValidatorSource.java
rolfyone: true, i didnt end up needing to test that directly, will change

tbenr (approved):
LGTM

PR #10573, opened 2026-04-16
Make vote tracking slot-aware and add backward-compatible vote update plumbing (3 of 7)

### LINKED ISSUE #10553: 3. Make vote tracking slot-aware and add backward-compatible vote update plumbing
Move vote-update logic into this layer by introducing slot-aware `VoteTracker`, `shouldUpdateVote(...)`, and the necessary `ForkChoiceStrategy` wiring.

Update vote serialization so legacy stored votes are still readable, while keeping pre-Gloas vote semantics unchanged.

tbenr:
fixes #10553 

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches fork-choice vote update logic and on-disk vote serialization, including legacy compatibility paths; bugs here could affect head selection or database upgrade/rollback behavior.
> 
> **Overview**
> Makes fork-choice vote tracking **slot-aware** by extending `VoteTracker` to store `nextSlot/currentSlot` plus a per-vote *FULL payload hint*, removing the old epoch-based field and updating equality/hash/toString accordingly.
> 
> Updates fork-choice plumbing to propagate this hint from attestations (default `false`, overridden in `ForkChoiceUtilGloas` based on attestation index) through deferred-attestation storage and into `ForkChoiceStrategy` vote updates, using a new `ForkChoiceUtil.shouldUpdateVote` helper.
> 
> Reworks vote persistence to be **backward-compatible**: `VoteTrackerSerializer` becomes spec-aware, reads legacy epoch-based formats (72/74 bytes) and converts epoch→slot, continues writing legacy format pre-Gloas for rollback safety, and switches to a new 84-byte slot/hint format once Gloas-era votes/hints are present. Tests, benchmarks, and compatibility fixtures are updated/added to cover these formats and behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71e8b07b78e12092009c93d49da9d2d388e1c574. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mehdi-aouadi (commented):
I prefer `payloadPresent` or `payloadFull` rather than the `hint` thing 🙈

tbenr:
> I prefer `payloadPresent` or `payloadFull` rather than the `hint` thing 🙈

this is actually intended. The point is that you can't say if that is an explicit vote for EMPTY or FULL if you don't check if it is voting for the current slot block or not. it is an "hint" because that decision will be taken by forkchoice logic later.
That boolean essentially only brings the information taken from the index.

i mean if it is "true" it will be actually voting for FULL ('cose gossip validations will drop those with index == 1 with inconsistent slot), but you can't say anything if it is "false"

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ForkChoiceStrategy.java
```diff
@@ -233,14 +249,41 @@ void processAttestation(
...
+        false,
+        spec.atSlot(UInt64.ZERO).getForkChoiceUtil());
+  }
+
+  void processAttestation(
+      final VoteUpdater voteUpdater,
+      final UInt64 validatorIndex,
+      final Bytes32 blockRoot,
+      final UInt64 targetEpoch,
+      final UInt64 slot,
```
zilm13: Slot parameter here is confusing especially as you are passing 0 knowing that it will not be used later sometimes. I think it's better to put some meaningful name here and always pass valuable data without knowledge if someone will consume it at the end or not.

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArrayScoreCalculator.java
```diff
@@ -111,9 +111,12 @@ private static void computeDelta(
           new VoteTracker(
               vote.getNextRoot(),
               vote.getNextRoot(),
-              vote.getNextEpoch(),
               vote.isNextEquivocating(),
-              vote.isNextEquivocating());
+              vote.isNextEquivocating(),
+              vote.getNextSlot(),
+              vote.isNextFullPayloadHint(),
+              vote.getNextSlot(),
```
zilm13, line 118: the input is called currentSlot, why do we pass next here?

zilm13 (commented):
LGTM just few nits

storage/src/main/java/tech/pegasys/teku/storage/protoarray/ProtoArrayScoreCalculator.java
tbenr: because all "current" values are overwritten with "next", it happens to all fields (next remain next) when the vote is applied.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/common/util/ForkChoiceUtil.java
```diff
@@ -772,6 +773,12 @@
     return parentExecutionRoot.isPresent() && !parentExecutionRoot.get().isZero();
   }
 
+  public boolean shouldUpdateVote(
+      final VoteTracker vote, final UInt64 targetEpoch, final UInt64 slot) {
```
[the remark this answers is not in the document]
rolfyone: assuming this is just needed down the line...

rolfyone (dismissed):
LGTM

rolfyone (dismissed):

StefanBratanov (approved):
LGTM

PR #10674, opened 2026-05-06
Gloas store transaction

mehdi-aouadi:
## PR Description
Fix Gloas fork choice update before hot block pruning. We need to process each block and then its matching payload immediately

## Documentation

## Changelog

---

> [!NOTE]
> **Low Risk**
> Only updates a Gloas store transaction test to reflect current execution payload lookup behavior; no production logic changes.
> 
> **Overview**
> Re-enables the previously disabled Gloas test for `retrieveSignedExecutionPayload` when the payload is sourced from an external provider.
> 
> The test now builds a real block/payload pair from `ChainBuilder`, commits the block into the store first (so `containsBlock` passes), asserts the payload isn’t already cached in hot storage, and then verifies `retrieveSignedExecutionPayload` returns the provider-supplied envelope.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc0ede4a69e12ded61ac58a5fa6b2bcf7bd8cf8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,10 @@ on:
   pull_request:
     branches:
       - master
+      - '*prototype*'
+      - '**/*prototype*'
+      - '*devnet*'
+      - '**/*devnet*'
     paths-ignore:
       - '**/*.md'
       - 'LICENSE*'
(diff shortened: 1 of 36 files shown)


tbenr:
actually we forgot we had already something related to this, waiting on devnet-0

moved to #10680 and let's see if this is required after that one

PR #10692, opened 2026-05-11
Add JaCoCo coverage reporting for unit and reference tests

rolfyone:
Applies the jacoco plugin to all subprojects so unit tests (test) and reference tests (referenceTest) generate execution data automatically. Adds jacocoRootReport and jacocoReferenceTestRootReport tasks at the root level to produce aggregate HTML and XML reports across all modules.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Build/test pipeline changes that add new JaCoCo tasks and dependencies; risk is mainly in Gradle task wiring and execution-data aggregation across modules potentially affecting CI runtime and report generation.
> 
> **Overview**
> Enables JaCoCo coverage collection across all modules by applying the `jacoco` plugin and generating per-subproject XML/HTML reports for `test`, plus a new `jacocoReferenceTestReport` wired to run after `referenceTest`.
> 
> Adds root-level aggregate report tasks: `jacocoRootReport` (unit tests), `jacocoReferenceTestRootReport` (reference tests), and `jacocoCombinedRootReport` (both), each collecting `*.exec` data across subprojects and emitting combined XML/HTML reports under `build/reports/jacoco`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4f6c743f0923eae44736e5353e52126327c9950. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1268,6 +1268,7 @@ tasks.register('jacocoReferenceTestRootReport', JacocoReport) {
 			t.name == 'jacocoReferenceTestReport'
 		}
 	}
+	mustRunAfter 'jacocoRootReport'
 
 	executionData.from fileTree(rootDir) {
 		include '**/build/jacoco/referenceTest.exec'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -74,6 +74,7 @@ def specificVersion = calculateVersion()
 def isDevelopBuild = rootProject.version.contains('develop')
 
 apply plugin: 'application'
+apply plugin: 'jacoco'
```
lucassaldanha, line 77: Why are we applying this to the root level build file?

build.gradle
```diff
@@ -1192,6 +1215,80 @@ subprojects {
...
+			t.name == 'jacocoReferenceTestReport'
+		}
+	}
+
+	executionData.from fileTree(rootDir) {
+		include '**/build/jacoco/referenceTest.exec'
+	}
+
+	subprojects.each { subproject ->
+		if (subproject.plugins.hasPlugin('java-library')) {
```
lucassaldanha, line 1278: Just checking that we sure all submodules we care about are java-library gradle modules?

lucassaldanha (dismissed):
LGTM. Just a couple of questions.

build.gradle
rolfyone: i wanted to be able to run unit test and ref test coverage and this is where claude put it... makes sense given that then the reports are then in the top level build folder?

build.gradle
rolfyone: as a starting point the reports seem to cover most of what I want i think, specifically fork choice etc. 

Will see what's broken, this might need to evolve but it looked fine locally... I think initially if we can get *something* then we can always tweak it if we're missing things we want is my theory...

lucassaldanha (approved):

PR #10698, opened 2026-05-13
Avoid signing execution-optimistic validator duties

rolfyone:
Treat execution-optimistic sync committee duties as syncing-node failures so they are retried instead of scheduled.

Sync committee duties are produced in the VC rather than via a BN production API call, so optimistic duty responses now skip scheduling and flow through the existing syncing-node retry path. Block production, attestation scheduling, and proposer-preference publishing keep the existing BN API behavior to avoid extra VC-BN roundtrips on the happy path.

Add regression tests for optimistic sync committee duty handling.

## Documentation

## Changelog


tbenr:
what I don't like here is the introduction of an extra VC-BN roundtrip on block production via `isExecutionOptimistic`.

I believe the nuance of the Sync committee duties is that they are produced in VC, not created via a BN call.
so in that case i can see an upfront call via `isExecutionOptimistic` (even if in theory we could gate on publish API).

so generally i'd more incline to call BN production apis anyway and rely on API failing with exception (and if we can't derive the nature of the problem via the http response (503?) we then could call `isExecutionOptimistic` after the fact). This would remove the constant extra roundtrip in the happy case (which is the one we should optimize for)

tbenr:
moreover, the upfront check is a delicate thing to handle. what if `isExecutionOptimistic` fails but it is just a network blip and the block production could actually run? the behaviour can vary based on the configuration (FailoverValidatorApiHandler , Sentry, and so on). So i believe the hot path should be with less calls as possible and with a "let it fail" behaviour.

tbenr:
So for those reasons i'd just touch the sync committee production flows which do not call any BN production API, and leave all the current flows untouched.

rolfyone:
Ok. I can look at trying to clean that up. I mainly want to avoid the scenario where we 'could' produce a block but haven't scheduled the duty

tbenr (approved):
LGTM

PR #10725, opened 2026-05-18
Improve data column sidecar pruning stream handling

rolfyone:
Improve data column sidecar pruning so large databases can prune efficiently.

Previously, pruning streamed data column sidecar identifiers from the start of the stored range and relied on stream filtering to stop at the prune cutoff. On large stores this could spend a long time scanning keys before reaching anything useful, and in some cases appeared to make no pruning progress.

This changes pruning to use a reverse slot lookup instead:
- find the latest data column sidecar slot at or before the prune cutoff
- stream identifiers only for the exact slot being pruned
- remove that slot’s sidecars
- repeat by seeking to the previous eligible slot

The first Fulu slot is still used as the lower bound so pruning can return quickly when the cutoff is before data column sidecars are supported.

Also:
- increases the data column pruning slot limit to 64
- reduces temporary debug logging back to summary-level pruning logs
- updates tests for the new newest-first pruning order

## Testing

```
bash
./gradlew :storage:test --tests tech.pegasys.teku.storage.server.kvstore.KvStoreDatabaseTest :storage:spotlessCheck
```

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes the data-column sidecar pruning algorithm and DAO lookups, which can impact what data is deleted and in what order if there are edge cases around slot bounds or canonical/non-canonical separation. Scoped to pruning/storage maintenance paths but touches persistent data deletion logic.
> 
> **Overview**
> Improves **data column sidecar pruning scalability** by switching from forward-stream scanning to a newest-first loop that uses DAO floor lookups to jump directly to the latest eligible slot, streams identifiers only for that slot, deletes them, then repeats for previous slots.
> 
> Adds `Spec.computeFirstSlotWithDataColumnSidecarSupport()` to bound pruning to Fulu+ networks, updates DAO APIs to support previous-slot lookups for canonical and non-canonical sidecars, and adjusts pruning defaults (data column pruning limit `12` → `64`) and tests/expectations to match the new pruning order.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63e1049cd487cb33a8cf4b5edc4d96304aad9f1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
@@ -1371,6 +1371,13 @@ public class Spec {
         .map(this::computeStartSlotAtEpoch);
   }
 
+  public Optional<UInt64> computeFirstSlotWithDataColumnSidecarSupport() {
+    return getSpecConfigFulu()
+        .map(SpecConfigFulu::getFuluForkEpoch)
+        .filter(epoch -> !epoch.equals(UInt64.MAX_VALUE))
+        .map(this::computeStartSlotAtEpoch);
+  }
+
   // Electra Utils
   public boolean isFormerDepositMechanismDisabled(final BeaconState state) {
     return atState(state).miscHelpers().isFormerDepositMechanismDisabled(state);
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
@@ -1372,9 +1372,11 @@ public class Spec {
   }
 
   public Optional<UInt64> computeFirstSlotWithDataColumnSidecarSupport() {
+    if (!supportsDataColumnSidecars()) {
+      return Optional.empty();
+    }
     return getSpecConfigFulu()
         .map(SpecConfigFulu::getFuluForkEpoch)
-        .filter(epoch -> !epoch.equals(UInt64.MAX_VALUE))
         .map(this::computeStartSlotAtEpoch);
   }
 
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
@@ -1260,17 +1260,11 @@ public class KvStoreDatabase implements Database {
     LOG.debug(
         "Pruning data column sidecars up to slot {}, limit {}", tillSlotInclusive, pruneLimit);
 
-    if (pruneDataColumnSidecars(
-        pruneLimit, tillSlotInclusive, dao::getEarliestDataSidecarColumnSlot, false, "canonical")) {
+    if (pruneDataColumnSidecars(pruneLimit, tillSlotInclusive, false, "canonical")) {
       LOG.debug("Data column sidecars pruning reached the limit of {}", pruneLimit);
     }
 
-    if (pruneDataColumnSidecars(
-        pruneLimit,
-        tillSlotInclusive,
-        dao::getEarliestNonCanonicalDataSidecarColumnSlot,
-        true,
-        "non-canonical")) {
+    if (pruneDataColumnSidecars(pruneLimit, tillSlotInclusive, true, "non-canonical")) {
       LOG.debug("Non-canonical data column sidecars pruning reached the limit of {}", pruneLimit);
     }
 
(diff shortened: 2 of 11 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/V4FinalizedKvStoreDao.java
```diff
@@ -250,8 +250,18 @@ public List<DataColumnSlotAndIdentifier> getDataColumnIdentifiers(
     }
   }
 
-  public Optional<UInt64> getEarliestDataSidecarColumnSlot() {
-    return db.getFirstEntry(schema.getColumnSidecarByColumnSlotAndIdentifier())
+  public Optional<UInt64> getLatestDataSidecarColumnSlotAtOrBefore(final UInt64 slot) {
```
gfukushima: Isn't this the earliest entry since we're using getfloor?

rolfyone: hrm it was the name codex decided on, seems not awful, but i honestly am not super concerned...

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -1371,6 +1371,13 @@ public Optional<UInt64> computeFirstSlotWithBlobSupport() {
         .map(this::computeStartSlotAtEpoch);
   }
 
+  public Optional<UInt64> computeFirstSlotWithDataColumnSidecarSupport() {
+    return getSpecConfigFulu()
+        .map(SpecConfigFulu::getFuluForkEpoch)
+        .filter(epoch -> !epoch.equals(UInt64.MAX_VALUE))
```
zilm13: nit: Not a very beautiful way to do it. We have FAR_FUTURE_EPOCH which is semantically more correct, but better to use `supportsDataColumnSidecars()`. 
Also `getSpecConfigFulu()` is already filtered on maxSupportedMilestone so the filter is not needed at all. For any doubts I'd better add a test.

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -1256,64 +1263,104 @@ public void addNonCanonicalSidecar(final DataColumnSidecar sidecar) {
...
+        "Data column sidecars pruning completed in {} ms", System.currentTimeMillis() - startTime);
   }
 
   boolean pruneDataColumnSidecars(
       final int pruneSlotLimit,
-      final Stream<DataColumnSlotAndIdentifier> dataColumnSlotAndIdentifierStream,
-      final boolean nonCanonicalBlobSidecars) {
+      final UInt64 tillSlotInclusive,
+      final boolean nonCanonicalSidecars,
+      final String sidecarType) {
```
zilm13: better make canonical/non-canonical an enum, especially as boolean was not a good way to go from the beginning

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -1256,64 +1263,104 @@ public void addNonCanonicalSidecar(final DataColumnSidecar sidecar) {
...
-        .takeWhile(
-            item -> prunableMap.size() < pruneSlotLimit || prunableMap.containsKey(item.slot()))
-        .forEach(
-            item -> prunableMap.computeIfAbsent(item.slot(), k -> new ArrayList<>()).add(item));
+    Optional<UInt64> maybeSlot =
+        findPreviousDataColumnSidecarSlot(
+            tillSlotInclusive, maybeFirstDataColumnSidecarSlot.get(), nonCanonicalSidecars);
+    while (prunedSlots < pruneSlotLimit
+        && maybeSlot.isPresent()
+        && maybeSlot.get().isLessThanOrEqualTo(tillSlotInclusive)) {
```
zilm13: The last check have happened in `findPreviousDataColumnSidecarSlot`, isn't it?

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -1256,64 +1263,104 @@ public void addNonCanonicalSidecar(final DataColumnSidecar sidecar) {
...
+      final UInt64 slot = maybeSlot.get();
+      final List<DataColumnSlotAndIdentifier> keys;
+      try (final Stream<DataColumnSlotAndIdentifier> identifiersForSlot =
+          nonCanonicalSidecars
+              ? streamNonCanonicalDataColumnIdentifiers(slot, slot)
```
zilm13: we have shortcuts for one slot already

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreDatabase.java
```diff
@@ -1256,64 +1263,104 @@ public void addNonCanonicalSidecar(final DataColumnSidecar sidecar) {
...
+
+    if (pruneDataColumnSidecars(pruneLimit, tillSlotInclusive, true, "non-canonical")) {
+      LOG.debug("Non-canonical data column sidecars pruning reached the limit of {}", pruneLimit);
     }
+
+    LOG.debug(
+        "Data column sidecars pruning completed in {} ms", System.currentTimeMillis() - startTime);
   }
 
   boolean pruneDataColumnSidecars(
```
zilm13, line 1303: I'd add Javadoc about how current strategy looks like and why it was chosen

storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/dataaccess/V4FinalizedKvStoreDao.java
rolfyone: codex's take was 
```
I updated the naming to avoid the ambiguity:

getPreviousDataColumnSidecarSlotAtOrBefore
getPreviousNonCanonicalDataColumnSidecarSlotAtOrBefore
I didn’t use getEarliest... because getFloorEntry returns the last/greatest key <= the supplied key, not the first/oldest one. git diff --check passes, but I didn’t rerun the Gradle tests after the rename because the approval was rejected.
```
which does sound reasonable... will push the name update.

zilm13 (approved):
LGTM

PR #10752, opened 2026-05-26
Fix FcU withdrawals calculation

### LINKED ISSUE #10682: [glamsterdam-devnet-3] self building could build invalid payload
seems like we autoreject what we just produced:

Image

the exception:
```
tech.pegasys.teku.spec.logic.common.execution.ExecutionPayloadVerificationException: Withdrawals of the envelope are not consistent with the expected withdrawals in the state
```

I decoded the payload envelop:

Image

So apparently the state expects some withdrawals and the envelop instead has no withdrawals

tbenr:
fixes #10682

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes withdrawal lists sent to the execution layer during payload building on GLOAS FULL-parent paths; incorrect logic would cause EL/CL payload mismatches, though behavior is covered by new spec and statetransition tests.
> 
> **Overview**
> Fixes **fork-choice updated (FcU) payload attributes** so Engine API withdrawals match the state used for GLOAS payload building when the head parent has a **FULL** execution payload.
> 
> **`ProposersDataManager`** no longer passes `spec.getExpectedWithdrawals(state)` directly. It resolves withdrawals asynchronously: for a FULL parent it loads parent **execution requests** from the blinded execution payload envelope, then calls **`Spec.getPayloadAttributeWithdrawals`**, which delegates to GLOAS fork-choice logic.
> 
> **`ForkChoiceUtilGloas.getPayloadAttributeWithdrawals`** applies parent execution payload processing on a **transient state copy** (parent execution requests, builder payment settlement, withdrawals), then returns **`payload_expected_withdrawals`** without mutating the live head state.
> 
> **`applyParentExecutionPayload`** is centralized in **`ExecutionRequestsProcessorGloas`** (moved out of **`BlockProcessorGloas`**) so block processing and FcU share the same logic. **`ForkChoiceUtilGloas`** wiring is extended with mutators, withdrawals helpers, and the execution-requests processor.
> 
> Tests cover effective withdrawals in **`ForkChoiceUtilGloasTest`** and end-to-end payload attribute calculation in **`ProposersDataManagerGloasTest`** (FULL vs EMPTY/pre-GLOAS parents, missing execution requests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 952420b65ee704fce60ce60681cdaa945e028737. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


StefanBratanov:
I agree with the logic. I don't like that we add this method to `BlockProcessor`, which should only do block processing stuff, should be part of ForkChoiceUtil in my opinion

tbenr:
> I agree with the logic. I don't like that we add this method to `BlockProcessor`, which should only do block processing stuff, should be part of ForkChoiceUtil in my opinion 

I totally agree that's the wrong place. Will fix

tbenr:
I decided to move the logic in the ExecutionRequestsProcessor and reuse it in blockProcessor and in ForkChoiceUtil.

i think it is fine to use the processor here 'cose it is actually a state transition.

rolfyone (approved):
LGTM

PR #10758, opened 2026-05-28
Limit pending block pool by retained bytes

rolfyone:
Add total weight accounting to PendingPool and configure pending beacon blocks to use SSZ size with a budget derived from MAX_PAYLOAD_SIZE. Add regression coverage for byte-budget eviction and oversized item rejection.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes block buffering during sync/reorg paths; mis-tuned limits could drop blocks that would later import, but behavior mirrors existing oldest-first eviction with an added memory bound.
> 
> **Overview**
> **Pending block pools** now enforce a **total retained byte budget** in addition to the existing max-item count. `PendingPool` tracks per-item weights, evicts oldest entries when count or cumulative weight would be exceeded, and rejects items larger than the cap or duplicates without re-indexing.
> 
> Beacon **pending block** pools (missing parent and missing parent execution payload) weight entries by **SSZ size**. `PoolFactory` sets the byte limit to **10×** networking `MAX_PAYLOAD_SIZE`. Other pending pools (e.g. attestations) keep the prior behavior via a delegating constructor with unit weight and unlimited total weight.
> 
> Tests cover weight-based eviction, oversized drops, and factory wiring for the gossip-derived byte limit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 803bcd56d9819475eb17f5b59e0bc109dd46357d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tbenr:
`block.getSchema().getSszSize(block.getBackingNode())` might be a viable alternative. It is cheap but not O(1).

rolfyone:
the size computation should be relatively ok, it accounts for individual block sizes as it turns out. I've increased the limit based on 10 blocks which is probably more reasonable, it could be many more than that if they're not too big.

lucassaldanha (approved):
LGTM

Issue #10773, opened 2026-06-01
Consensus-specs alpha.9 meta


zilm13:
- [x] [Exclude slashed validators from proposing (EIP-8045)](https://github.com/ethereum/consensus-specs/pull/5115) - consensus breaking, required - https://github.com/Consensys/teku/pull/10766
- [x] [Precompute head for update_proposer_boost_root](https://github.com/ethereum/consensus-specs/pull/5300) - opened #10783 
- [x] [Apply proposer boost if dependent roots match](https://github.com/ethereum/consensus-specs/pull/5306) - opened #10783 
- [x] [Ensure bids are for a higher slot than their parent](https://github.com/ethereum/consensus-specs/pull/5302) - merged in #10764
- [x] [Ignore PTC attestations for empty assigned slots](https://github.com/ethereum/consensus-specs/pull/5281) - merged in #10764
- [x] [Use should_build_on_full for bid parent block hash](https://github.com/ethereum/consensus-specs/pull/5310) - opened #10782 
- [x] [Limit should_build_on_full checks to the previous slot](https://github.com/ethereum/consensus-specs/pull/5309) - opened #10782 
- [x] [Remove support for old deposit mechanism in Fulu](https://github.com/ethereum/consensus-specs/pull/4704) - https://github.com/Consensys/teku/pull/10766
- [x] [Remove should_override_forkchoice_update](https://github.com/ethereum/consensus-specs/pull/5174) - covered in #10766 
- [x] [Force the proposer to reorg late payloads](https://github.com/ethereum/consensus-specs/pull/5210) - added to #10782 
- [x] [Introduce is_previous_slot_payload_decision predicate](https://github.com/ethereum/consensus-specs/pull/5317)
- [x] Reftests alpha.9 - https://github.com/Consensys/teku/pull/10766/
- [x] Specrefs alpha.9
- [ ] [Redesign fork choice abstractions](https://github.com/ethereum/consensus-specs/pull/5249) https://github.com/Consensys/teku/issues/10754

Testing
- [x] [Add `on_attestation` tests for `validate_on_attestation` updates](https://github.com/ethereum/consensus-specs/pull/5275)

gfukushima:
can this be closed?

Gasila5:
[]()417-1376

PR #10783, opened 2026-06-02
Apply proposer boost if dependent roots match

zilm13:
## PR Description
https://github.com/ethereum/consensus-specs/pull/5300
https://github.com/ethereum/consensus-specs/pull/5306

Part of #10773 

## Documentation

## Changelog

---

> [!NOTE]
> **High Risk**
> Changes fork-choice proposer boost and reorg timing on block import, which directly affects head selection and consensus behavior; incorrect logic could cause unsafe reorgs or missed boost.
> 
> **Overview**
> Aligns **proposer boost** with updated consensus-specs fork-choice rules ([5300](https://github.com/ethereum/consensus-specs/pull/5300), [5306](https://github.com/ethereum/consensus-specs/pull/5306)): boost is set only when the imported block is the first timely block of the slot **and** its **dependent root** matches the pre-import head’s dependent root (via new `getDependentRoot` / `getDependentSlot` helpers on `ForkChoice`).
> 
> The previous behavior that granted boost when head state was unavailable or when the block proposer matched the head proposer at the current slot is removed. Boost is skipped if there is no chain head at import time.
> 
> **Tests & conformance:** `ForkChoiceTest` adds matching vs divergent dependent-root scenarios; three Altair fork-choice reference tests are no longer excluded. **Specrefs** wire `get_dependent_root` (phase0/gloas) to `ForkChoice.java` and drop those entries from unimplemented exceptions. **AGENTS.md** documents `ethspecify` usage for `specrefs/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a4deca42fe55c06bc7a579c8fdfffb959f62c98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rolfyone:
Findings
[P3] specrefs/functions.yml (line 12503) is stale for the function this PR changes. It still records update_proposer_boost_root(store, root) with the old proposer-index/head-state rule, and the on_block references still call update_proposer_boost_root(store, block_root) at line 8316 (line 8316). The merged consensus-specs changes now precompute head and compare dependent roots, which the PR implements in ForkChoice.java (line 899). Since check-specrefs.yml only runs for specrefs/** or build.gradle changes (workflow trigger (line 7)), this PR leaves the reference metadata/documentation wrong unless specrefs is updated here too.

ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/forkchoice/ForkChoice.java
```diff
@@ -712,6 +710,8 @@ private BlockImportResult importBlockAndState(
         computeEarliestBlobSidecarsSlot(
             recentChainData.getStore(), dataAndValidationResult, block.getMessage());
 
+    final ForkChoiceNode preImportHead = findNewChainHead(forkChoiceStrategy).node();
```
tbenr: Do we need to run a protoarray head search here? The spec computes `head = get_head(store)` because it does not maintain a cached head (everything is "inline"). In Teku `recentChainData.getChainHead().getForkChoiceNode()` should already be the pre-import head. If there is a case where the cached head can be stale here, we should document/test that; otherwise I’d avoid the extra head walk on every block import.

zilm13: updated

tbenr (approved):
LGTM

PR #10789, opened 2026-06-04
 Avoid redundant RPC fetches for data column sidecars

zilm13:
## PR Description
Avoid redundant RPC fetches for data column sidecars once the active DA state is already satisfied.

  Skips stale delayed fetch callbacks when their tracker was removed/replaced, and ignores incoming sidecars for already imported blocks/payloads.

  Prevents Req/Resp from firing after local recovery/import has already provided the needed data.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches the Fulu data-availability sampling and Req/Resp path, including Gloas deferred-DA behavior; changes are guardrails only but timing bugs here could affect block import or sync.
> 
> **Overview**
> **`DasSamplerBasic`** stops doing redundant work once data availability is already satisfied locally, and hardens delayed RPC fetches against removed trackers.
> 
> Incoming validated data column sidecars are **ignored** when the block is already in chain and DA is complete—for non-Gloas forks that means the block is imported; for **Gloas** (deferred DA) tracking continues until the execution payload is available in the store, then late sidecars are dropped.
> 
> When a **delayed** missing-column RPC runs, it now checks that the `DataColumnSamplingTracker` is still the active one for that block root (e.g. after `removeAllForBlock` on import/reject); stale callbacks skip `retrieve` and clear the in-progress flag.
> 
> Tests cover stale delayed fetch after tracker removal, pre-Gloas imported blocks, and Gloas before/after payload import.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19095939cd65348f9317d6c11dbcdf797ce5e4d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/datacolumns/DasSamplerBasic.java
```diff
@@ -226,6 +240,20 @@ private boolean hasBlobs(final BeaconBlock block) {
...
+      return false;
+    }
+
+    if (spec.atSlot(slot)
+        .getForkChoiceUtil()
+        .isDataAvailabilityCheckDeferredToExecutionPayloadEnvelope()) {
+      return recentChainData.getStore().getExecutionPayloadIfAvailable(blockRoot).isPresent();
+    }
+
+    return true;
```
rolfyone, line 260: do we ever get here? it seems like if we do get here true might be wrong?

so 
 - if we're finalized or block hasn't been seen, then false
 - if we have the block and payload is available then true
 - default to true seems weird...

zilm13: I've added clarifying comment
so if we are there:
1) block is imported
2) DA is not deferred to ExecutionPayload


so it means we are in Fulu(preGloas) and as far as block is already imported, DA is already satisfied

tbenr (approved):
LGTM

PR #10803, opened 2026-06-08
Fix Gloas empty parent payload withdrawals - block production

### LINKED ISSUE #10682: [glamsterdam-devnet-3] self building could build invalid payload

tbenr:
fixes #10682

### Fix Gloas empty-parent payload withdrawals

For Gloas payload attributes, `PAYLOAD_STATUS_EMPTY` now uses the state's cached `payload_expected_withdrawals` instead of recomputing withdrawals via `getExpectedWithdrawals`.

This matches Gloas processing: when the parent payload is empty, withdrawals are not reprocessed, so the payload attributes must preserve the withdrawals already stored in state.

`PAYLOAD_STATUS_FULL` still applies parent execution requests first, while `PAYLOAD_STATUS_PENDING` keeps the existing pre-Gloas/default behavior.

Added a regression test covering the empty-parent case.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches withdrawal selection for Gloas payload building (consensus-adjacent fork-choice behavior); scope is narrow and covered by a targeted regression test.
> 
> **Overview**
> Fixes **Gloas block production** when the parent fork-choice node has an **empty** execution payload: Engine API payload attributes now take withdrawals from the beacon state’s cached **`payload_expected_withdrawals`** instead of **`getExpectedWithdrawals`**, matching Gloas rules where withdrawals are not reprocessed on an empty parent payload.
> 
> **`Spec.getPayloadAttributeWithdrawals`** is refactored to a **`switch` on `ForkChoicePayloadStatus`**: `PENDING` still uses expected withdrawals (pre-Gloas/default), `EMPTY` reads `BeaconStateGloas.getPayloadExpectedWithdrawals()`, and `FULL` still applies parent execution requests via **`ForkChoiceUtilGloas`** before computing withdrawals. Javadoc is updated accordingly; constructor checks use static **`checkArgument`** imports only.
> 
> A **regression test** in **`ProposersDataManagerGloasTest`** seeds distinct `payload_expected_withdrawals`, asserts they differ from expected withdrawals, and verifies empty-parent payload building uses the cached list without fetching parent execution requests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c8add035f06ee523b713502d867010f0e54b617. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
@@ -22,7 +22,6 @@ import static tech.pegasys.teku.spec.SpecMilestone.GLOAS;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import it.unimi.dsi.fastutil.ints.Int2IntMap;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
 import it.unimi.dsi.fastutil.ints.IntList;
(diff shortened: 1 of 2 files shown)


rolfyone (approved):
LGTM

PR #10805, opened 2026-06-08
Add experimental QUIC transport support

lucassaldanha:
Add QUIC transport to the libp2p network, gated behind hidden experimental flags (--Xp2p-quic-enabled, --Xp2p-quic-port, --Xp2p-advertised-quic-port and their IPv6 variants). QUIC is disabled by default and TCP remains enabled, so default behavior is unchanged.

- NetworkConfig gains QUIC listen/advertised ports and quicEnabled/ tcpEnabled flags
- LibP2PNetworkBuilder registers QuicTransport (ECDSA) and builds QUIC listen/advertised multiaddrs, making TCP conditional on tcpEnabled
- Discovery advertises a QUIC address in the ENR; DiscoveryPeer and NodeRecordConverter propagate it, and dialing prefers QUIC when a peer advertises one
- Migrate RPC stream handling from P2PChannel to Stream for libp2p 1.3.1 compatibility
- Derive the discovery node id from the libp2p identity key for QUIC peers. Over QUIC the libp2p-TLS secure session exposes the ephemeral ECDSA certificate key rather than the node's secp256k1 identity key, which previously caused discovery node id derivation to fail (a `NoSuchElementException` from `getDiscoveryNodeId().orElseThrow()` in PeerDAS components). The identity key is now recovered from the verified peer id (which inlines the key for secp256k1/Ed25519 identities), falling back to the secure-session key when not recoverable; Noise/TCP behavior is unchanged. A debug log is also added when node id calculation fails so future key-type mismatches are no longer swallowed silently.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches core P2P discovery, dialing, and peer identity used for discovery node IDs; defaults preserve TCP-only behavior but misconfiguration or QUIC-only peers could affect connectivity.
> 
> **Overview**
> Adds **experimental QUIC** libp2p transport behind hidden `--Xp2p-*` flags (QUIC off by default; TCP unchanged unless disabled). `NetworkConfig` gains QUIC/TCP listen and advertised ports plus `quicEnabled` / `tcpEnabled`; the builder registers `QuicTransport`, advertises/listens only on enabled transports, and requires at least one transport.
> 
> **Discovery and dialing:** ENRs and `DiscoveryPeer` carry an optional QUIC socket; conversion from node records prefers QUIC v4/v6 when IPv6 is supported. Outbound multiaddrs prefer QUIC when this node has QUIC enabled and the peer advertises it, otherwise TCP.
> 
> **QUIC correctness:** `LibP2PPeer` recovers the secp256k1 **identity** public key from the peer id when the secure session reports the TLS certificate key (fixes discovery node id / PeerDAS failures on QUIC). RPC handling moves from `P2PChannel` to `Stream` (`closeWrite` for half-close). Failed discovery node id calculation is logged at debug instead of swallowed silently.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a211053b3cb2594eeda5e95858f6e370b460ab4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
@@ -16,6 +16,7 @@ package tech.pegasys.teku.networking.p2p.libp2p;
 import static tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork.REMOTE_OPEN_STREAMS_RATE_LIMIT;
 import static tech.pegasys.teku.networking.p2p.libp2p.LibP2PNetwork.REMOTE_PARALLEL_OPEN_STREAMS_COUNT_LIMIT;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import identify.pb.IdentifyOuterClass;
 import io.libp2p.core.Host;
@@ -36,9 +37,8 @@ import io.libp2p.transport.tcp.TcpTransport;
 import io.netty.handler.logging.LogLevel;
 import java.net.InetSocketAddress;
 import java.time.Duration;
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Stream;
 import org.hyperledger.besu.plugin.services.MetricsSystem;
 import tech.pegasys.teku.infrastructure.async.AsyncRunner;
 import tech.pegasys.teku.infrastructure.io.IPVersionResolver;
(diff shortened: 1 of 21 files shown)


networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PNetworkBuilder.java
```diff
@@ -113,23 +115,45 @@ public P2PNetwork<Peer> build() {
...
+              new InetSocketAddress(advertisedIps.get(0), config.getAdvertisedPort()), nodeId);
+      if (config.isQuicEnabled()) {
+        final Multiaddr advertisedQuicAddr =
+            MultiaddrUtil.fromInetSocketAddressAsQuic(
+                new InetSocketAddress(advertisedIps.get(0), config.getAdvertisedQuicPort()),
+                nodeId);
+        advertisedAddresses = List.of(advertisedAddr, advertisedQuicAddr);
+      } else {
+        advertisedAddresses = Collections.singletonList(advertisedAddr);
+      }
```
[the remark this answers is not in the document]
lucassaldanha: Ah that is a good point. I think I need to adjust this.

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/network/config/NetworkConfig.java
```diff
@@ -47,9 +47,13 @@ public class NetworkConfig {
 
   public static final List<String> DEFAULT_P2P_INTERFACE = List.of("0.0.0.0");
   public static final int DEFAULT_P2P_PORT = 9000;
+  public static final int DEFAULT_P2P_QUIC_PORT = 9100;
```
rolfyone: is this a standard across clients?

lucassaldanha: There is no standard for the QUIC UDP port. We do use UDP 9000 for discv5, so we need another one when using QUIC.

rolfyone (dismissed):
LGTM

rolfyone (approved):

PR #10806, opened 2026-06-09
Added compatibility for 26.4 rocksdb

rolfyone:
## Summary

- Avoid creating RocksDB column families just by opening a database.
- Treat missing known column families as empty for reads/streams.
- Lazily create column families only when data is actually written.
- Allow databases with unknown existing column families to open, so newer Teku can avoid breaking downgrade to released versions when no new-column data has been written.

## Motivation

Opening `master` against a `26.4.0` V6 database could create newly added column families eagerly. After that, `26.4.0` could no longer reopen the database because RocksDB requires every existing column family to be explicitly opened.

This preserves compatibility without bumping the database version for open-only/no-write cases.

## Documentation

## Changelog

---

> [!NOTE]
> **High Risk**
> Touches core beacon chain persistence and column-family lifecycle; incorrect open/create behavior could corrupt data or break upgrades/downgrades between releases.
> 
> **Overview**
> Changes RocksDB V6 open behavior so **only column families already on disk** are opened, instead of eagerly creating every schema column at startup. **`createMissingColumnFamilies`** is removed; new families are created **lazily on first write** via `getColumnHandleForWrite`, while reads/streams/deletes treat a missing handle as **empty**.
> 
> `RocksDbInstanceFactory` now **lists existing column IDs** when opening and can attach **unknown** families so newer Teku can run read-only against older DBs without breaking **downgrade to 26.4** when no new-column data was written.
> 
> Adds **`RocksDbInstanceFactoryTest`**, acceptance **`DatabaseCompatibilityAcceptanceTest`** (local RocksDB V6 → last-release startup), and **`TekuNode.stopAndGetDataDirectoryFromContainer`** for capturing node data dirs in tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30e05b30faf68c1069d5bb5cb387c7c1b62d65c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstance.java
+++ b/storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstance.java
@@ -13,7 +13,6 @@
 
 package tech.pegasys.teku.storage.server.rocksdb;
 
-import com.google.common.collect.ImmutableMap;
 import com.google.errorprone.annotations.MustBeClosed;
 import java.util.HashSet;
 import java.util.List;
@@ -27,10 +26,12 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.tuweni.bytes.Bytes;
 import org.rocksdb.AbstractRocksIterator;
+import org.rocksdb.ColumnFamilyDescriptor;
 import org.rocksdb.ColumnFamilyHandle;
 import org.rocksdb.RocksDBException;
 import org.rocksdb.RocksIterator;
 import org.rocksdb.TransactionDB;
+import tech.pegasys.teku.storage.server.DatabaseStorageException;
 import tech.pegasys.teku.storage.server.ShuttingDownException;
 import tech.pegasys.teku.storage.server.kvstore.ColumnEntry;
 import tech.pegasys.teku.storage.server.kvstore.KvStoreAccessor;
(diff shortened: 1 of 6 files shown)


storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstance.java
```diff
@@ -250,7 +269,10 @@ private <K, V> Stream<ColumnEntry<byte[], byte[]>> createStreamRaw(
       final Consumer<RocksIterator> setupIterator,
       final Predicate<K> continueTest) {
     assertOpen();
-    final ColumnFamilyHandle handle = columnHandles.get(column);
+    final ColumnFamilyHandle handle = getColumnHandle(column);
+    if (handle == null) {
```
gfukushima, line 273: all those extra `if null` checks are a bit of an overhead for me, this gets called so many times and performs an extra check unnecessarily just for the case where user downgraded the db and doesn't have the column anymore. I really think this goes in the wrong direction performance wise.

gfukushima (approved):
Adding overhead to db operations that are used constantly in the client (even though they are minor) seems wrong to me. This PR also introduces a very important change in behaviour as well, the storage model becomes more permissive and less explicit meaning we might fail silently in cases where we pretend the column exists but it actually doesn't. Currently we fail fast and easily detect that there's been a change in the DB schema. For the next release this might be the fast approach we need to guarantee backwards compatibility but we should follow up with a (create an issue) proper way to handle this cases rather than add exceptional checks in hotspots for a very rare scenario.

storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstance.java
rolfyone: I don’t think we can safely remove the missing-handle check while preserving the compatibility behavior. Missing column families are expected now: reads need to treat them as empty without creating the column, otherwise a current Teku open/read can make the DB unopenable by the released version again.

I did remove the heavier parts of the original concern: reads no longer acquire a monitor and no longer allocate an Optional. The remaining cost is a lock-free map lookup plus one predictable null branch before the RocksDB JNI/read/iterator work. Creating a sentinel/wrapper would just move that branch behind virtual dispatch, and creating columns on read would break the downgrade guarantee.

PR #10807, opened 2026-06-09
Data column pruner optimizations

gfukushima:
## PR Description
This PR adds a variable in the DB so we can track where the prune left off when restarts happen.
It also restores the previous behaviour pruning forward which avoid creating a gap between old and more recent data in cases where the backlog grows.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes finalized storage pruning semantics and schema (new DB variable); incorrect watermark or pruning logic could delete or mis-report retained sidecars, though behavior is heavily tested.
> 
> **Overview**
> **Data column sidecar pruning** is switched from **newest-first to oldest-first**, removing up to `pruneLimit` distinct populated slots per run in one transaction so the retained range advances steadily when ingestion outpaces pruning.
> 
> A new persisted **`LAST_DATA_COLUMN_SLOT_PRUNED`** watermark is written atomically with canonical deletions. Canonical pruning and **`getEarliestDataColumnSidecarSlot`** resume from `watermark + 1` so restarts and later runs skip scanning deletion tombstones below already-pruned slots. Non-canonical pruning still scans from the first supported slot.
> 
> **Offline disk reclamation:** `Database.compactStorage()` / `KvStoreAccessor.compact()` trigger full blocking compaction on RocksDB and LevelDB, exposed as **`debug-db compact-db`** for use while the node is stopped.
> 
> Floor-key helpers (`getPreviousDataColumnSidecarSlotAtOrBefore`) are removed in favor of forward collection via `collectOldestSidecarsToPrune`. Integration and unit tests are updated for oldest-first behavior, watermark tracking, and compaction.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9c9a84821edfb4b406fec4566ff34567f6dbe31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/schema/V6SchemaCombined.java
```diff
@@ -281,6 +289,7 @@ public Map<String, KvStoreVariable<?>> getVariableMap() {
         .put("CUSTODY_GROUP_COUNT", getVariableCustodyGroupCount())
         .put("FIRST_CUSTODY_INCOMPLETE_SLOT", getVariableFirstCustodyIncompleteSlot())
         .put("EARLIEST_AVAILABLE_DATA_COLUMN_SLOT", getVariableEarliestAvailableDataColumnSlot())
+        .put("LAST_DATA_COLUMN_SLOT_PRUNED", getVariableLastDataColumnSidecarPrunedSlot())
```
rolfyone, line 292: worth verifying this prints out in `debug-tools db get-variables`

storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstance.java
```diff
@@ -269,6 +274,28 @@ private <K, V> Stream<byte[]> createStreamKeyRaw(
...
+    assertOpen();
+    final List<ColumnFamilyHandle> handles = new ArrayList<>(columnHandles.values());
+    handles.add(defaultHandle);
+    LOG.info("Compacting {} RocksDB column families", handles.size());
+    int compacted = 0;
+    for (final ColumnFamilyHandle handle : handles) {
+      try {
+        final long startTime = System.currentTimeMillis();
+        db.compactRange(handle);
+        LOG.info(
```
rolfyone, line 316: can we do these as debug?

CHANGELOG.md
```diff
@@ -27,4 +27,5 @@
  - Fixed a potential issue in importing blocks when data is not available.
  - Fixed potential NPE when SSE are not closed correctly.
  - Improved pruning for data column sidecars.
+ - Restored oldest-first (forward) data column sidecar pruning to prevent a retention gap from forming when sidecars are stored faster than the prune limit allows them to be removed.
```
rolfyone: probably dont need this? i think its inside this release

storage/src/main/java/tech/pegasys/teku/storage/server/rocksdb/RocksDbInstance.java
gfukushima: this is only called by the debug tool and I think it's worth keeping at info level, we should not be calling this at all in other places

rolfyone (approved):

PR #10820, opened 2026-06-11
Split new builder deposits signature verification into batches

StefanBratanov:
## PR Description
As per discussion on Discord, if there is a single invalid signature when verifying new builder deposits, the whole batch verification would fail (let's say 8192 new builder deposits and only 1 invalid one would fail the whole batch). It's smarter to split into smaller sub-batches to handle this case better.

Also added small comments to `DefaultExecutionPayloadManager` regarding the pending pool with a reasoning.

## Fixed Issue(s)
N/A

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes consensus deposit processing for Gloas builder deposits and adjusts execution-payload pending cache sizing, which could affect block processing cost and memory under edge loads.
> 
> **Overview**
> **Gloas new builder deposits** are no longer BLS batch-verified in one shot. They are split into sub-batches of **256** (`Lists.partition`); each sub-batch is verified and applied separately so a single bad signature only forces per-signature fallback within that sub-batch, not across thousands of deposits.
> 
> **Execution payload caching** renames `VALID_EXECUTION_PAYLOADS_SET_SIZE` → `VALID_EXECUTION_PAYLOAD_SET_SIZE` and updates imports. `DefaultExecutionPayloadManager` sets the pending-payload map size to a fixed **64** (with comments that it matches the non-finalized slot window under healthy finality) instead of deriving it as `VALID_EXECUTION_PAYLOAD_SET_SIZE * 2`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7c0dd598a807dad75e99fad85529e64fcc2c3cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/config/Constants.java
@@ -26,7 +26,7 @@ public class Constants {
   public static final int VALID_PAYLOAD_ATTESTATION_SET_SIZE = 512 * 2;
   // The cache is used for the `payload_present` voting, gossip and RPC fetch de-duplication, so no
   // need for a long term caching (1 epoch)
-  public static final int VALID_EXECUTION_PAYLOADS_SET_SIZE = 32;
+  public static final int VALID_EXECUTION_PAYLOAD_SET_SIZE = 32;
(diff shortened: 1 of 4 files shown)


zilm13 (approved):
LGTM

PR #10829, opened 2026-06-12
Penalize peers on hacky/bad inbound messages

### LINKED ISSUE #10339: Penalize peers causing ReqResp/Gossip exceptions
If peer made incoming ReqResp request and this request caused an exception (for example, it was malformed), the peer should be penalized.
This should include peers sending `DataColumnSidecar` or other objects not passing validation both on gossip and ReqResp.

zilm13:
## PR Description
Not sure on this because we should distinguish when it's really bad and worth large penalty and when it's not that bad. In code it looks...pretty edgy.

## Fixed Issue(s)

Fixes #10339 

## Documentation

## Changelog


rolfyone:
i do like this idea, maybe we feature toggle it and update so we've got quic etc so that we can progress it...

PR #10830, opened 2026-06-12
Disconnect bad inbound connections

### LINKED ISSUE #9924: Apply reputation to inbound connections
Initially approached here #9502 

Historically, ReputationManager started dealing with tracking disconnections and data received from peers while syncing.
It has been designed to be used only when selecting peers for outbound connections.

When we added rate limiting logic when serving RPC requests, we end up disconnecting the peer and applying a reputation penalty which leads to a cooldown period. (COOLDOWN_PERIOD)

The problem is that this "cooldown" remained applied only to outbound connection. The rate limited peer can potentially reconnect to us and our rate-limiting trackers will be freshly created and peer can continue ask us like nothing happened before.

An approach could be:

- change `ReputationManager` interface to expose a `isInboundConnectionAllowed`
- the `DefaultReputationManager` should track the last disconnection reason and direction (`locallyInitiated`)
- in `PeerManager::handleConnection` call `isInboundConnectionAllowed`
- Initially, `isInboundConnectionAllowed` implementation could evaluate `isSuitableAt` only if last disconnection was "locallyInitiated" and the disconnection reason was rate RATE_LIMITING, and allow connections on all other cases. We could evaluate case by case if we could be more restrictive.

To be evaluated what to do with `reportInitiatedConnection*` methods.

zilm13:
## PR Description
Previously, reputation bans only affected outbound peer selection: Teku checked isConnectionInitiationAllowed before dialing discovered peers.

  A banned peer could still reconnect inbound because the p2p accept path did not consult reputation.

  This adds an inbound rejection check for reputation-banned peers and sends a clean BAD_SCORE goodbye before exposing the peer as connected.

## Fixed Issue(s)

fixes #9924 

## Documentation

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,3 +11,5 @@
 ### Additions and Improvements
 
 ### Bug Fixes
+
+- Prevent RPC rate-limited peers from immediately reconnecting inbound.
--- a/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPeer.java
+++ b/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/libp2p/LibP2PPeer.java
@@ -292,7 +292,7 @@ public class LibP2PPeer implements Peer {
   public void adjustReputation(final ReputationAdjustment adjustment) {
     final boolean shouldDisconnect = reputationManager.adjustReputation(getAddress(), adjustment);
     if (shouldDisconnect) {
-      disconnectCleanly(DisconnectReason.REMOTE_FAULT).finishError(LOG);
+      disconnectCleanly(DisconnectReason.BAD_SCORE).finishError(LOG);
     }
   }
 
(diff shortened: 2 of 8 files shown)


tbenr (commented):
things we may consider:
- changelog line?
- add a test covering the staticPeer exemption
- there is a "false-negative" path:
```java
@Test
void shouldRejectInboundConnectionAfterRateLimitDisconnectDuringExistingCooldown() {
  // A prior failed outbound dial puts the peer in a benign cooldown and clears any inbound reason.
  reputationManager.reportInitiatedConnectionFailed(peerAddress);
  assertThat(reputationManager.getInboundConnectionRejectionReason(peerAddress)).isEmpty();

  // While still within that cooldown, the peer connects inbound and floods us, so we rate-limit it.
  reputationManager.reportDisconnection(
      peerAddress, Optional.of(DisconnectReason.RATE_LIMITING), true);

  // The rate-limited peer should be rejected if it reconnects inbound during the cooldown.
  assertThat(reputationManager.getInboundConnectionRejectionReason(peerAddress))
      .contains(DisconnectReason.RATE_LIMITING);
}
```

tbenr (approved):
LGTM

PR #10841, opened 2026-06-16
Add jdk.management module to Docker runtime for process_cpu_seconds_total

### LINKED ISSUE #10838: `process_cpu_seconds_total` missing from Docker image metrics
After upgrading to Teku `26.6.0`, the Prometheus metric `process_cpu_seconds_total` is no longer emitted from the Teku `/metrics` endpoint.

## Observed behavior

Teku `26.6.0` still exposes these `process_*` metrics:

```text
process_resident_memory_bytes
process_start_time_seconds
process_virtual_memory_bytes
```

But it does not expose:

```text
process_cpu_seconds_total
```

## Expected behavior

Teku should continue exposing `process_cpu_seconds_total` when process metrics are enabled.

## Investigation

Teku still enables process metrics by default via `StandardMetricCategory.PROCESS`, and still initializes Besu's Prometheus metrics system:

```java
metricsSystem = new PrometheusMetricsSystem(config.getMetricsCategories(), true);
metricsSystem.init();
```

The CPU metric is provided by Prometheus Java JVM instrumentation. That collector reads process CPU time through the JDK management API:

```java
com.sun.management.OperatingSystemMXBean.getProcessCpuTime()
```

That API is provided by the JDK module:

```text
jdk.management
```

The `jdk25` and `jdk26` Dockerfiles build a custom Java runtime using an explicit `jlink` module list:

```text
--add-modules java.base
--add-modules java.se
--add-modules java.management
--add-modules java.prefs
--add-modules jdk.httpserver
--add-modules jdk.unsupported
--add-modules jdk.jdwp.agent
```

This list does not include:

```text
jdk.management
```

As a result, Prometheus Java's `ProcessMetrics` CPU collector cannot access the JVM process CPU API, and `process_cpu_seconds_total` is omitted.

## Version notes

Observed after upgrading to `26.6.0`.

The `26.6.1` source still has the same `jdk25` and `jdk26` Dockerfile module lists, so it appears likely to be affected as well.

## Suggested fix

Add `jdk.management` to the custom runtime module list in:

```text
docker/jdk25/Dockerfile
docker/jdk26/Dockerfile
```

For example:

```text
--add-modules jdk.management
```

lucassaldanha:
## Summary

Fixes #10838.

After moving the Docker images to a custom `jlink`-built Java runtime, the `process_cpu_seconds_total` Prometheus metric stopped being emitted from the `/metrics` endpoint (other `process_*` metrics were unaffected).

**Root cause:** the `jlink` module list in `docker/jdk25/Dockerfile` and `docker/jdk26/Dockerfile` omitted the `jdk.management` module, which is the module that exports `com.sun.management`. Prometheus' `ProcessMetrics` CPU collector reads CPU time via `com.sun.management.OperatingSystemMXBean.getProcessCpuTime()`; without `jdk.management` the class cannot be resolved at runtime (`NoClassDefFoundError: com/sun/management/OperatingSystemMXBean`), so the collector silently drops the metric. Note `java.se` does not pull in any `jdk.*` modules, so it does not cover this.

## Change

Add `--add-modules jdk.management` to both Dockerfiles' `jlink` invocation.

## Testing

Reproduced and verified locally using the same JDK 25 + `jlink` setup the images use, with a small probe calling `getProcessCpuTime()`:

| Runtime | `jdk.management` present | `getProcessCpuTime()` |
|---------|--------------------------|------------------------|
| Current module list | absent | ❌ `NoClassDefFoundError: com/sun/management/OperatingSystemMXBean` |
| With `jdk.management` | present | ✅ returns CPU time |

Thanks to @yohamta0 for the detailed report and diagnosis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,3 +13,4 @@
 ### Bug Fixes
 
 - Prevent RPC rate-limited peers from immediately reconnecting inbound.
+- Fixed missing `process_cpu_seconds_total` metric in the Docker images by adding the `jdk.management` module to the custom Java runtime.
--- a/docker/jdk25/Dockerfile
+++ b/docker/jdk25/Dockerfile
@@ -6,6 +6,7 @@ RUN $JAVA_HOME/bin/jlink \
          --add-modules java.se \
          --add-modules java.management \
          --add-modules java.prefs \
+         --add-modules jdk.management \
          --add-modules jdk.httpserver \
          --add-modules jdk.unsupported \
          --add-modules jdk.jdwp.agent \
--- a/docker/jdk26/Dockerfile
+++ b/docker/jdk26/Dockerfile
@@ -6,6 +6,7 @@ RUN $JAVA_HOME/bin/jlink \
          --add-modules java.se \
          --add-modules java.management \
          --add-modules java.prefs \
+         --add-modules jdk.management \
          --add-modules jdk.httpserver \
          --add-modules jdk.unsupported \
          --add-modules jdk.jdwp.agent \


gfukushima (approved):

rolfyone (approved):
LGTM

PR #10846, opened 2026-06-17
Fix peer selection so inbound connections cannot occupy all peer slots and prevent required outbound connections

zilm13:
## PR Description
- Fix score-based peer connection calculation so reserved random outbound peers are not counted twice when determining remaining peer-count capacity (line 78)
 - Reserve capacity for required outbound peers using `minRandomlySelectedPeers` + `subnetSubscribersRequired`.
 - Count only locally initiated random peers as protected/random coverage.
 - When the remote-peer cap is exceeded, drop remotely initiated peers first before falling back to normal score-based drops.
 - Add tests for inbound saturation, remote random peers not being protected, subnet requirements, and remote-cap drop ordering.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes live P2P connect/disconnect policy for the beacon node, which can affect peer counts and gossip reach; behavior is covered by new unit tests but impacts production networking.
> 
> **Overview**
> Fixes **Eth2 peer selection** so inbound connections cannot fill the peer budget and block required **outbound** dials (minimum randomly selected peers plus attestation subnet subscribers).
> 
> **Connect:** Score-based capacity now subtracts **`randomlySelectedPeersToAdd`** (not total random pool size), and the random pool count used for targets includes only **locally initiated** random peers.
> 
> **Disconnect:** Computes a **remote-peer cap** from `upperBound - (minRandomPeers + subnet requirement)`, drops **remotely initiated** peers first (by score), then fills remaining over-cap drops via the usual path (demote locally initiated random peers to score-based, then lowest scores). Inbound random peers are **not** protected like outbound random peers.
> 
> **Config:** Default minimum random peer count rises from **20%** to **30%** of the peer lower bound; **`minRandomlySelectedPeers`** is validated against **`maxPeers`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41bb431e50331ae5428ff2aa7ab48c5d5fb3152b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
```diff
@@ -152,42 +153,124 @@ private Optional<PeerAddress> checkCandidate(
...
+            + "totalOutboundRequirement={})",
+        peersBeingDropped::size,
+        peers::size,
+        () -> peers.stream().filter(Peer::connectionInitiatedRemotely).count(),
+        remotelyInitiatedPeersBeingDropped::size,
+        locallyInitiatedRandomlySelectedPeersBeingDropped::size,
+        () ->
+            targetPeerCountRange.getMinimumRandomlySelectedPeerCount()
+                + peerSubnetSubscriptions.getSubscribersRequired());
+    return peersBeingDropped.stream().toList();
```
tbenr: `return List.copyOf(peersBeingDropped);` ?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
```diff
@@ -152,42 +153,124 @@ private Optional<PeerAddress> checkCandidate(
...
+    }
 
     final Map<PeerConnectionType, List<Peer>> peersBySource =
-        network
-            .streamPeers()
+        peers.stream()
             .collect(Collectors.groupingBy(peer -> peerPools.getPeerConnectionType(peer.getId())));
-
     final List<Peer> randomlySelectedPeers =
         peersBySource.getOrDefault(RANDOMLY_SELECTED, new ArrayList<>());
```
tbenr: `new ArrayList<>()`  can be `List.Of()`, seems like we don't need a mutable version

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
```diff
@@ -152,42 +153,124 @@ private Optional<PeerAddress> checkCandidate(
...
-    shuffler.shuffle(randomlySelectedPeers);
-    final List<Peer> randomlySelectedPeersBeingDropped =
-        randomlySelectedPeers.subList(
-            0, Math.min(randomlySelectedPeersToDrop, randomlySelectedPeers.size()));
+
+    // Fill any remaining drop slots using the normal score-based ordering.
+    final List<Peer> peersBeingDropped = new ArrayList<>(remotelyInitiatedPeersBeingDropped);
+    final int currentPeerCount = peers.size();
+    final List<Peer> locallyInitiatedRandomlySelectedPeersBeingDropped =
+        filterAndSelectLocallyInitiatedRandomPeersToDrop(randomlySelectedPeers, currentPeerCount);
```
tbenr: what about rename to:

final List<Peer> locallyInitiatedRandomPeersEligibleForDisconnection =
    selectLocallyInitiatedRandomPeersEligibleForDisconnection(randomlySelectedPeers, currentPeerCount)

?

networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/peers/Eth2PeerSelectionStrategy.java
```diff
@@ -152,42 +153,124 @@ private Optional<PeerAddress> checkCandidate(
...
+            targetPeerCountRange.getMinimumRandomlySelectedPeerCount()
+                + peerSubnetSubscriptions.getSubscribersRequired());
+    return peersBeingDropped.stream().toList();
+  }
+
+  private int calculateRemotelyInitiatedPeersToDropCount(
+      final List<Peer> peers, final PeerSubnetSubscriptions peerSubnetSubscriptions) {
+    final int totalOutboundRequirement =
+        targetPeerCountRange.getMinimumRandomlySelectedPeerCount()
+            + peerSubnetSubscriptions.getSubscribersRequired();
```
tbenr: `totalOutboundRequirement` can calculated outside and reused in the TRACE log

tbenr (commented):
some additional comments. Not sure i can approve yet :)

zilm13:
@tbenr addressed

tbenr (commented):
two more feedbacks:

1. 
`--Xp2p-minimum-randomly-selected-peer-count` should at least not allow inconsistent values compared to upperbound.
Additionally: do we want to allow the edgecase where all connections should be outbound (essentially aloways disconnecting  inbounds?) if not we may want to enforce some reasonable values.. but i don't have a strong feeling here.

2. 
The remote-cap calculation currently drops remotely initiated peers even when total peer count is well below the lower bound. For example with upper=35, minRandom=25, and subnetRequired=1, the remote limit is 9, so a node with only 18 peers and 10 remotely initiated peers still disconnects. Could we make remote-cap drops conditional on being at/near the upper bound, or otherwise account for current headroom, so inbound peers are not dropped while the node is already below target?

zilm13:
1. I have changed
2. I don't agree. I've made explicitly more aggressive behavior to protect space for outgoing peers. So what's currently in PR:
 - outboundTartget = 18 / 100, inbound = 82 / 100
 - we start dropping inbound even if outbound is not reached, so when we have 8 outbound and 83 inbound, though it's under 100 total, we will drop 1 inbound
 - if we don't do this, in the same case we come to 100 with 92 inbound and 8 outbound, `disconnect` and `connect` blocks are separated, so how disconnect block should handle this? drop 10 on 100? drop 1? If it drops 1 in `connect` block it will have space for 1 peer to connect at most.
 - As blocks are separated and there are a lot of peers, peers could easily connect between `disconnect` and `connect` blocks. 
 - For something in the middle I could suggest add protection to not drop until we are under lower bound. So if you set big number of outgoing this measure will jump in (but not on default settings, because 82 > 64)

zilm13:
I've updated check along with my suggestion https://github.com/Consensys/teku/pull/10846/commits/8e63ff0e0f0c3e2d5e7a92393932356ee60f7b03
If you think we should handle this in a different way, we could discuss it

networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/connection/TargetPeerRange.java
```diff
@@ -59,8 +59,13 @@ public int getRandomlySelectedPeersToDrop(
...
+      final int currentRemotelyInitiatedPeerCount,
+      final int totalOutboundRequirement,
+      final int currentTotalPeerCount) {
     final int remotelyInitiatedPeerLimit = Math.max(0, upperBound - totalOutboundRequirement);
-    return Math.max(0, currentRemotelyInitiatedPeerCount - remotelyInitiatedPeerLimit);
+    final int remotelyInitiatedPeersAboveLimit =
+        Math.max(0, currentRemotelyInitiatedPeerCount - remotelyInitiatedPeerLimit);
+    final int peersAboveLowerBound = Math.max(0, currentTotalPeerCount - lowerBound);
+    return Math.min(remotelyInitiatedPeersAboveLimit, peersAboveLowerBound);
   }
```
tbenr, line 73: this blows a my mind a bit.

an alternative:




```java
public int getRemotelyInitiatedPeersToDrop(
    final int currentRemotelyInitiatedPeerCount,
    final int totalOutboundRequirement,
    final int currentTotalPeerCount) {
  final int remotelyInitiatedPeerLimit = Math.max(0, upperBound - totalOutboundRequirement);
  final int maxPeersToDropWithoutGoingBelowLowerBound =
      Math.max(0, currentTotalPeerCount - lowerBound);

  return Math.clamp(
      currentRemotelyInitiatedPeerCount - remotelyInitiatedPeerLimit,
      0,
      maxPeersToDropWithoutGoingBelowLowerBound);
}
```

tbenr (dismissed):
LGTM. just a simplification option

tbenr (approved):
LGTM

PR #10858, opened 2026-06-18
Remove web3j filtering for dependency management

StefanBratanov:
## PR Description
We may not use the library soon, but in any case we can upgrade and remove the dependency filter.

https://github.com/LFDT-web3j/web3j/releases/tag/v5.0.3

## Fixed Issue(s)
N/A

## Documentation

## Changelog

---

> [!NOTE]
> **Low Risk**
> Build-only change to dependency update filtering; no application code or pinned versions are modified in this diff.
> 
> **Overview**
> Removes the **Gradle `dependencyUpdates` guard** that automatically rejected any `org.web3j` candidate at major version **5+**, so the versions plugin can now surface web3j 5.x upgrades alongside other dependencies.
> 
> The change deletes the `isWeb3jMajorUpgrade` helper and drops it from `rejectVersionIf`; **non-stable version rejection** and the existing **jqwik / snappy** update blocks are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bfa8b75dd3830a75905c90531c51853d7f807ff9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/beacon/pow/build.gradle
+++ b/beacon/pow/build.gradle
@@ -18,6 +18,7 @@ dependencies {
 
 	api 'org.web3j:core'
 
+    implementation 'com.fasterxml.jackson.core:jackson-databind'
 	implementation 'io.consensys.tuweni:tuweni-units'
 
 	testImplementation testFixtures(project(':infrastructure:async'))
--- a/beacon/pow/build.gradle
+++ b/beacon/pow/build.gradle
@@ -18,7 +18,6 @@ dependencies {
 
 	api 'org.web3j:core'
 
-	implementation 'com.fasterxml.jackson.core:jackson-databind'
 	implementation 'io.consensys.tuweni:tuweni-units'
 
 	testImplementation testFixtures(project(':infrastructure:async'))
--- a/beacon/pow/build.gradle
+++ b/beacon/pow/build.gradle
@@ -18,7 +18,7 @@ dependencies {
 
 	api 'org.web3j:core'
 
-    implementation 'com.fasterxml.jackson.core:jackson-databind'
+	implementation 'com.fasterxml.jackson.core:jackson-databind'
 	implementation 'io.consensys.tuweni:tuweni-units'
 
 	testImplementation testFixtures(project(':infrastructure:async'))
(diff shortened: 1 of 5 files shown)


gfukushima:
played with this a little bit today,  long story short I think 5.0.3 introduced what it would be a breaking change for us, I think @lucassaldanha recently changed that constructor to get a better performance out of that?

> // same call that is made in the constructor's hierarchy HttpService -> Service
    // returns the same ObjectMapper instance (singleton) so when we register the module it
    // applies to all usages within Web3jService
    final ObjectMapper objectMapper = ObjectMapperFactory.getObjectMapper(false);
    objectMapper.registerModule(new BlackbirdModule());

We could potentially upgrade to 5.0.2 which still has the constructor that we currently use

StefanBratanov:
> played with this a little bit today, long story short I think 5.0.3 introduced what it would be a breaking change for us, I think @lucassaldanha recently changed that constructor to get a better performance out of that?
> 
> > // same call that is made in the constructor's hierarchy HttpService -> Service
> > // returns the same ObjectMapper instance (singleton) so when we register the module it
> > // applies to all usages within Web3jService
> > final ObjectMapper objectMapper = ObjectMapperFactory.getObjectMapper(false);
> > objectMapper.registerModule(new BlackbirdModule());
> 
> We could potentially upgrade to 5.0.2 which still has the constructor that we currently use

Yeah, actually it's big upgrade yeah since it upgrades to Jackson 3, so bit worried about compatibility here. Maybe will leave it as draft now.

lucassaldanha:
I don't see the point in updating it now. We are close to completely removing it from our deps.

StefanBratanov:
> I don't see the point in updating it now. We are close to completely removing it from our deps.

Makes sense, I changed the scope to just remove the filtering since it's unnecessary code in my opinion and may get forgotten eventually when we remove web3j

lucassaldanha:
> Makes sense, I changed the scope to just remove the filtering since it's unnecessary code in my opinion and may get forgotten eventually when we remove web3j

This is tracked here: https://github.com/Consensys/teku/issues/10877

StefanBratanov:
> > Makes sense, I changed the scope to just remove the filtering since it's unnecessary code in my opinion and may get forgotten eventually when we remove web3j
> 
> This is tracked here: #10877

Nice, closing this one

PR #10873, opened 2026-06-22
Fix sync gloas

zilm13:
## PR Description
  **What happened** 
During range sync a batch's last block is allowed to be imported without its execution payload envelope (the last-block envelope check is deferred, never enforced).
No subsequent batch ever re-requests that envelope — the next batch starts after it, and the parent's slot is unknown (it may sit behind empty slots), so it's silently
  orphaned.

  The next block then fails import forever with UNKNOWN_PARENT_EXECUTION_PAYLOAD (parent payload missing from the store), and sync stalls 

  **What it does now**

  When a block fails import with UNKNOWN_PARENT_EXECUTION_PAYLOAD, BatchImporter fetches the parent's execution payload envelope by root (block.getParentRoot()), imports it,
  and retries the block import once. By-root needs no slot, so it doesn't matter how far back the parent is or whether the next batch is built on empty slots — the gap is
  irrelevant.

The real PR is https://github.com/Consensys/teku/pull/10873/commits/2fcace6c204fa9d81306875300eabc3577959631, everything else is merge master

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches core multipeer sync import and peer fetch paths for Gloas execution payloads; behavior is narrowly scoped to one retry on a specific failure reason with new tests.
> 
> **Overview**
> Fixes **Gloas range sync stalls** when a batch’s last block was imported without its execution payload envelope (allowed deferral) and the next block fails with `UNKNOWN_PARENT_EXECUTION_PAYLOAD` because that parent envelope was never stored.
> 
> **`BatchImporter`** now wraps block import in recovery: on that failure it requests the parent’s envelope **by `block.getParentRoot()`** from the batch sync peer, imports it via `executionPayloadManager`, and **retries the block import once**. Unrecoverable cases keep the original failure; execution or data-availability errors from the recovered envelope are surfaced as the appropriate batch outcomes (`EXECUTION_CLIENT_OFFLINE`, `DATA_NOT_AVAILABLE`, etc.).
> 
> **Networking:** `requestExecutionPayloadEnvelopeByRoot` moves from **`Eth2Peer` only** onto the **`SyncSource`** API (used by multipeer sync). **`ThrottlingSyncSource`** delegates by-root lookups **without rate limiting** so recovery does not stall the batch. Tests cover success, missing envelope, and recovered-payload import failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77a438147eb41dca9a3b52042fc9e9ffeda3c23c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
```diff
@@ -188,6 +187,55 @@ private SafeFuture<SingleImportResult> importBlock(
...
+                  .requestExecutionPayloadEnvelopeByRoot(block.getParentRoot())
+                  .thenCompose(
+                      maybeExecutionPayload -> {
+                        return maybeExecutionPayload
+                            .map(
+                                signedExecutionPayloadEnvelope ->
+                                    executionPayloadManager
+                                        .importExecutionPayload(
+                                            signedExecutionPayloadEnvelope, false)
+                                        .thenCompose(__ -> blockImporter.importBlock(block)))
```
tbenr: here you are reimporting the block blindly without checking the `importExecutionPayload` result.
I get the simplification attempt but I wander if it is better to do the import only on success and return the original failure otherwise (with some additional debug logging?)

zilm13: good catch!

beacon/sync/src/test/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporterTest.java
```diff
@@ -359,6 +361,110 @@ void shouldImportBlobSidecarsAndExecutionPayloadForSameBlock() {
...
+    firstImport.complete(BlockImportResult.FAILED_UNKNOWN_PARENT_EXECUTION_PAYLOAD);
+
+    // No envelope available: the block is not retried and the batch import fails
+    stubSyncSource.assertRequestedExecutionPayloadEnvelopeByRoot(block.getParentRoot());
+    ignoreFuture(verify(blockImporter, times(1)).importBlock(block));
+    assertThat(result).isCompletedWithValue(BatchImportResult.IMPORT_FAILED);
+  }
+
+  @Test
+  void shouldNotRetryBlockWhenRecoveredParentExecutionPayloadFailsToImport() {
```
tbenr, line 434: maybe here we should cover the new error forwarding logic

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
```diff
@@ -206,6 +207,12 @@ public SafeFuture<Void> requestExecutionPayloadEnvelopesByRange(
             });
   }
 
+  @Override
+  public SafeFuture<Optional<SignedExecutionPayloadEnvelope>> requestExecutionPayloadEnvelopeByRoot(
+      final Bytes32 beaconBlockRoot) {
+    return delegate.requestExecutionPayloadEnvelopeByRoot(beaconBlockRoot);
```
tbenr, line 215: this is not throttling.. but probably not a big deal...

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/BatchImporter.java
```diff
@@ -188,6 +187,107 @@ private SafeFuture<SingleImportResult> importBlock(
...
   }
 
+  /**
+   * Imports the block, lazily recovering from a missing parent execution payload during sync.
+   *
+   * <p>When the block fails with {@code UNKNOWN_PARENT_EXECUTION_PAYLOAD} the parent block's
+   * execution payload envelope has not been delivered (for example it was the last block of a
+   * previous batch, whose envelope is allowed to be deferred). The parent's slot is unknown - the
+   * next batch may be built on top of empty slots - so the envelope is fetched by root, imported,
+   * and the block import is retried once. If the envelope cannot be recovered the original failure
```
tbenr: this requires update too

beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/ThrottlingSyncSource.java
zilm13: it's intentional
it happens very rare, it's just 1 request and we don't want it to be slow

tbenr (approved):
LGTM

Issue #10101, opened 2025-11-05
Investigate usages of `BlockBlobSidecarsTrackersPool` 


StefanBratanov:
In Fulu, we are no longer using `BlockBlobSidecarsTrackersPool` but the pool still is doing some additional features/optimizations that we are missing in Fulu

Some things on first look, but there could be more:
- [ ] 1. `SingleBlockProvider` was used to return recently validated blocks (not yet importer)
- [ ] 2. `SingleBlobSidecarProvider` was used to return recently validated blobs (not yet imported)
- [x] 3. No longer calling `gossipValidatorSupplier.get().markForEquivocation(blobSidecar);`

regarding 1 and 2:
As per [spec](https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/p2p-interface.md#datacolumnsidecarsbyroot-v1)

> Clients SHOULD include a sidecar in the response as soon as it passes the gossip validation rules. Clients SHOULD NOT respond with sidecars related to blocks that fail gossip validation rules. Clients SHOULD NOT respond with sidecars related to blocks that fail the beacon chain state transition

We could extend the tracker in the `DasSamplerBasic` to cache the columns and blocks in memory, but we should be careful to limit that cache. Moreover, in `BlockBlobSidecarsTrackersPool` was dropping immediately all successfully imported block\blobs, the DasSampler is not currently aware of that.

gfukushima:
Had a look at the currently implementation and we used to call `markForEquivocation` after recovering using the EL to fetch the blob for the `currentSlot` on DENEB. With Fulu/DAS we wait 2 slots to start recovery of the Columns and then we would trigger the sync recovery for the missing columns. My guess is by then we don't care about marking those request as equivocation because those have already being dropped or responded. cc @zilm13 to confirm my understanding

tbenr:
> My guess is by them we don't care about marking those request as equivocation because those have already being dropped or responded. cc @zilm13 to confirm my understanding

I don't get this argument. The equivocation marking covers the case in which the node retrieves them via EL before getting anything from GOSSIP. This way we can `ignore` any sidecars coming from gossip that do not match what we successfully recovered locally via EL.

gfukushima:
Given that we disabled `BlockBlobSidecarsTrackersPool` after we're past the retention period I take that item 2 is calling the NOOP so I take that's resolved. Also we have DasSampler follwing similar mechanism and removing the block and tracker as soon as block is imported.

[...]

PR #10914, opened 2026-06-30
Add interface to allow serve gossip validated data column from blocks not imported

gfukushima:
Serve recently-validated custody data columns by-root before block import.

In Fulu, DataColumnSidecarsByRoot could not serve custody columns for a block that had passed gossip validation but was not yet imported, even though custody already persists those columns to the DB on validation.
The handler resolves the block's slot via getSlotByBlockRoot (fork-choice and finalized-DB only) before retrieving, and that returns empty for an unimported block, so nothing was served — contrary to the spec's "include
a sidecar as soon as it passes the gossip validation rules".

Resolve the missing slot from the DAS sampler, whose tracker is keyed by block root and knows the slot from the first column arrival. Surface it via a RecentlyValidatedSlotProvider held by RecentChainData (mirroring the existing validated block/blob providers) and have CombinedChainDataClient.getSlotByBlockRoot fall back to it after fork-choice and finalized-DB miss. Once the slot resolves, the existing getSidecar DB read serves the custody-persisted column. The sampler tracker is removed on import, handing slot resolution back to fork choice.

## PR Description
Fixes item 2 of #10101 

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes slot resolution on the data-column req/resp path and introduces a new fallback source; behavior is scoped to misses after fork-choice/DB but incorrect slots could affect what sidecars are served.
> 
> **Overview**
> **Fixes Fulu `DataColumnSidecarsByRoot` when custody columns exist for gossip-validated blocks that are not yet imported.** Slot lookup used to stop at fork-choice and the finalized DB, so `getSidecar` never ran even though custody had already persisted columns.
> 
> Adds **`RecentlyValidatedDataColumnSlotProvider`** (same pattern as validated block/blob providers), wired from **`DasSamplerBasic.getSlotForBlockRoot`** (DAS sampling tracker keyed by block root) into **`RecentChainData`**. **`CombinedChainDataClient.getSlotByBlockRoot`** now falls back to that index after hot/finalized (and non-canonical when enabled) miss. **`DataColumnSidecarsByRootMessageHandler`** already resolves slots via `getSlotByBlockRoot`, so peers can be served without handler changes.
> 
> Fork-choice and DB resolution still win when present; the sampler tracker is cleared on import so slot resolution moves back to the chain.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d7b514b90c17f27f59c800b2c3cbb46242c621a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rolfyone (approved):
LGTM

PR #10916, opened 2026-06-30
Fix --p2p-udp-port being ignored for advertised UDP port in ENR

### LINKED ISSUE #10915: Bug: --p2p-udp-port is ignored for ENR/discovery_addresses advertised port
## Summary

When `--p2p-udp-port` is set to a value different from `--p2p-port`, the advertised UDP port in the ENR and in the `/eth/v1/node/identity` `discovery_addresses` field still shows the TCP port instead of the configured UDP port.

## Steps to Reproduce

Start Teku with:
```
--p2p-port=9000
--p2p-udp-port=9001
--p2p-advertised-ip=<public-ip>
```

Query the identity endpoint:
```
GET /eth/v1/node/identity
```

## Expected Behavior

`discovery_addresses` should contain `/ip4/<public-ip>/udp/9001/...` and the ENR should encode UDP port 9001.

## Actual Behavior

`discovery_addresses` contains `/ip4/<public-ip>/udp/9000/...` and the ENR encodes UDP port 9000. The `--p2p-udp-port=9001` value is silently ignored for the advertised port.

## Root Cause

In `P2PConfig.build()` ([`networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java:361`](https://github.com/Consensys/teku/blob/master/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/P2PConfig.java)):

```java
discoveryConfig.advertisedUdpPortDefault(OptionalInt.of(networkConfig.getAdvertisedPort())); // TCP port = 9000
```

This runs **after** `P2POptions.configure()` has called `discoveryConfig.listenUdpPort(9001)`, and it fills in `advertisedUdpPort` with the TCP port (9000) because `advertisedUdpPort` is still empty at that point.

`DiscoveryConfig.getAdvertisedUdpPort()` is:
```java
return advertisedUdpPort.orElse(listenUdpPort);
```

The intent was: fall back to `listenUdpPort` (9001) when no explicit `--p2p-advertised-udp-port` is set. But because `advertisedUdpPortDefault` runs after and fills in the value with 9000, the fallback to `listenUdpPort` is never reached.

## Workaround

Add `--p2p-advertised-udp-port=9001` explicitly alongside `--p2p-udp-port=9001`. Since this sets `advertisedUdpPort` to a non-empty value before `P2PConfig.build()` runs, the `advertisedUdpPortDefault(9000)` call becomes a no-op.

## Fix Direction

`P2PConfig.build()` should only propagate the TCP advertised port as the UDP advertised default when `--p2p-udp-port` was **not** explicitly set. When a UDP listen port has been explicitly configured, it should take precedence over the TCP port for the UDP advertised default.

MqllR:
## Summary

Fixes #10915

When `--p2p-udp-port` is set to a different value than `--p2p-port`, the advertised UDP port in the ENR and `discovery_addresses` still reflected the TCP port rather than the configured UDP port.

**Root cause:** `P2PConfig.build()` called `advertisedUdpPortDefault(OptionalInt.of(networkConfig.getAdvertisedPort()))`, which is always non-empty. This filled in `advertisedUdpPort` before the fallback to `listenUdpPort` in `getAdvertisedUdpPort()` could take effect.

**Fix:** expose `getOptionalAdvertisedPort()` on `NetworkConfig` to return the raw `OptionalInt` (empty when `--p2p-advertised-port` was not explicitly set). Pass it directly to `advertisedUdpPortDefault` so the call is a no-op when no explicit advertised TCP port was configured, allowing `getAdvertisedUdpPort()` to fall back naturally to `listenUdpPort`.

The NAT scenario (`--p2p-advertised-port` set without `--p2p-udp-port`) continues to work correctly.

## Changes

- `NetworkConfig`: add `getOptionalAdvertisedPort()` / `getOptionalAdvertisedPortIpv6()` returning the raw `OptionalInt`
- `P2PConfig`: pass the raw optionals to `advertisedUdpPortDefault` instead of always-non-empty wrapped values
- `P2POptionsTest`: 4 new tests covering the regression case, explicit `--p2p-advertised-udp-port` taking priority, and IPv6 variants

## Test plan

- [ ] `--p2p-port=9000 --p2p-udp-port=9001` → advertised UDP port = 9001 (regression)
- [ ] `--p2p-port=9000 --p2p-udp-port=9001 --p2p-advertised-udp-port=9002` → advertised UDP port = 9002 (explicit wins)
- [ ] `--p2p-port=9000 --p2p-advertised-port=9100` → advertised UDP port = 9100 (NAT preserved)
- [ ] `--p2p-port=9000` only → advertised UDP port = 9000 (default unchanged)
- IPv6 variants of the above

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Narrow P2P discovery port wiring change with targeted CLI tests; wrong ENR ports affect peer discovery but the change restores intended CLI semantics rather than introducing new behavior paths.
> 
> **Overview**
> **Fixes incorrect advertised UDP port in the ENR** when `--p2p-udp-port` differs from `--p2p-port` (e.g. #10915). Discovery was still advertising the TCP port because `P2PConfig.build()` seeded discovery with `OptionalInt.of(networkConfig.getAdvertisedPort())`, which is never empty even when `--p2p-advertised-port` was not set.
> 
> `NetworkConfig` now exposes `getOptionalAdvertisedPort()` / `getOptionalAdvertisedPortIpv6()`, and `P2PConfig` passes those raw optionals into `advertisedUdpPortDefault` so the default is a no-op unless the user explicitly set an advertised TCP port. `DiscoveryConfig.getAdvertisedUdpPort()` can then fall back to `listenUdpPort` as intended; explicit `--p2p-advertised-udp-port` and NAT-style `--p2p-advertised-port` behavior are unchanged.
> 
> CLI integration tests were added for the regression, IPv6, and priority between UDP listen, advertised UDP, and advertised TCP ports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e5ca65d5b67f7b41761aa27a16149d82dd04b76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MqllR:

zilm13:
Looks good, big thank you!
Could you please also add these test cases (they should pass):
- IPv4: --p2p-udp-port=9001 --p2p-advertised-port=7000 advertises UDP 7000.
- IPv6: --p2p-udp-port-ipv6=9091 --p2p-advertised-port-ipv6=7070 advertises UDP IPv6 7070.

tbenr:
@zilm13 maybe we should consider taking this over, apply the changes and merge

zilm13 (dismissed):

zilm13:
I've added missing tests
Merging

zilm13 (approved):
Deliver it!

PR #10917, opened 2026-06-30
Preserve Gloas state transition caches

tbenr:
Speeds up `onboardBuildersFromPendingDeposits` especially

### Cache migration soundness

The Gloas state upgrade runs after the normal epoch transition at the fork boundary:

1. process final Fulu slot
2. run Fulu epoch transition
3. set state slot to the Gloas fork slot
4. detect milestone change
5. run `GloasStateUpgrade`

So the upgrade receives the normal post-epoch Fulu state; it is not replacing epoch processing.

This cache migration mirrors regular state mutation behavior: `BeaconState.updated(...)` creates a mutable copy using `TransitionCaches.copy()`. That copy already preserves the `ValidatorIndexCache` by reference, while copying most other cache maps.

This is sound for Fulu -> Gloas because the upgrade does not reorder validators, remove validators, or change validator public keys. Cached validator indices therefore remain valid, and `ValidatorIndexCache` already guards against stale indices beyond the current validator list size.

The intended benefit is that `onboardBuildersFromPendingDeposits` can reuse a warmed validator-index cache when checking pending-deposit pubkeys, instead of rebuilding it on the freshly-created Gloas state.

Caveat: this is intentionally Gloas-specific. Carrying all transition caches should not be generalized blindly to future forks where cache semantics or fork-specific derivation logic may change.

## Documentation

## Changelog

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 9801ccb632f5a6b01c52374d3eb77db13ce89804. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/gloas/BeaconStateGloasImpl.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/state/beaconstate/versions/gloas/BeaconStateGloasImpl.java
@@ -35,6 +35,13 @@ public class BeaconStateGloasImpl extends AbstractBeaconState<MutableBeaconState
   }
 
   BeaconStateGloasImpl(
+      final BeaconStateSchema<BeaconStateGloas, MutableBeaconStateGloas> schema,
+      final TransitionCaches transitionCaches,
+      final SlotCaches slotCaches) {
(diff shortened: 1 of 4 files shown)


mehdi-aouadi (approved):
LGTM
I find it wise to not generalise this and keep it a case by case choice

PR #10938, opened 2026-07-06
validate exeution payload block hash

### LINKED ISSUE #10884: Missing validation during historical payload sync
#10592 checks that the historical execution payload envelope is consistent with the signed Gloas bid for directly overlapping fields, including `block_hash`.

However, this only verifies that the payload header *claims* the same `block_hash` as the bid. It does not verify that the claimed `block_hash` is actually the EL block hash of the fetched payload/header.

In forward sync this is covered by `engine_newPayload`, where the EL validates the execution payload and its block hash.

In historical sync, Teku fetches full execution payload envelopes from peers, blinds them locally, and stores the blinded version without calling `engine_newPayload`. Therefore a malicious peer could provide an execution payload header with altered non-bid fields, such as `stateRoot` or `receiptsRoot`, while keeping `blockHash` equal to the signed bid’s `block_hash`.

The missing validation is to recompute the EL block hash from the fetched execution payload before blinding/storing it:

`computedBlockHash = keccak256(rlp(executionHeader(payload)))`

and require:

`computedBlockHash == bid.block_hash`

mehdi-aouadi:
## PR Description
Add Gloas execution payload block hash validation using Besu RLP/header utilities and Teku hashing. Historical sync now verifies downloaded execution payload envelopes by recomputing the payload block hash before blinding and storing them

## Fixed Issue(s)

fixes #10884 

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes historical sync acceptance of Gloas execution payloads and ties hash correctness to Besu header encoding; wrong logic would reject valid peers or accept bad data, but scope is Gloas envelope import only.
> 
> **Overview**
> Adds **Gloas execution payload block hash validation** so historical sync rejects tampered or inconsistent envelopes before they are blinded and stored.
> 
> A new `ExecutionPayloadBlockHashCalculator` recomputes the execution block hash from the envelope (payload fields, execution requests, parent beacon root, BAL, slot number, etc.) using Besu header/RLP helpers and keccak256. `HistoricalBatchFetcher.processExecutionPayload` now calls this check and throws `IllegalArgumentException` on mismatch.
> 
> The `ethereum/spec` module gains Besu dependencies for that computation. Test fixtures in `BlockProposalTestUtil` set **computed** `blockHash` on Gloas payloads so chain builders and sync tests stay consistent with the validator.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8563d278278c9f048b7c6f3e74efe74f36501107. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadBlockHashCalculator.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadBlockHashCalculator.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright Consensys Software Inc., 2026
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package tech.pegasys.teku.spec.datastructures.execution;
+
+import static tech.pegasys.teku.infrastructure.crypto.Hash.keccak256;
+import static tech.pegasys.teku.infrastructure.crypto.Hash.sha256;
+
+import java.util.List;
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import org.hyperledger.besu.datatypes.Address;
+import org.hyperledger.besu.datatypes.BlobGas;
+import org.hyperledger.besu.datatypes.GWei;
+import org.hyperledger.besu.datatypes.Hash;
+import org.hyperledger.besu.datatypes.LogsBloomFilter;
+import org.hyperledger.besu.datatypes.Wei;
+import org.hyperledger.besu.ethereum.core.BlockHeader;
+import org.hyperledger.besu.ethereum.core.BlockHeaderBuilder;
+import org.hyperledger.besu.ethereum.core.Difficulty;
+import org.hyperledger.besu.ethereum.core.Util;
+import org.hyperledger.besu.ethereum.mainnet.BodyValidation;
+import org.hyperledger.besu.ethereum.mainnet.MainnetBlockHeaderFunctions;
+import org.hyperledger.besu.ethereum.rlp.BytesValueRLPOutput;
+import tech.pegasys.teku.infrastructure.unsigned.UInt64;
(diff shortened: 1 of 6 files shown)


mehdi-aouadi:
I added some temporary logs and tested this on `devnet-6`. It works as expected:

Screenshot 2026-07-09 at 12 10 40

I didn't get any execution payload rejected because of a block hash mismatch.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/datastructures/execution/ExecutionPayloadBlockHashCalculator.java
```diff
@@ -0,0 +1,114 @@
...
+            .blobGasUsed(toLong(payload.getBlobGasUsed()))
+            .excessBlobGas(BlobGas.of(payload.getExcessBlobGas().bigIntegerValue()))
+            .parentBeaconBlockRoot(executionPayloadEnvelope.getParentBeaconBlockRoot())
+            .requestsHash(
+                computeRequestsHash(
+                    executionRequestsDataCodec.encode(
+                        executionPayloadEnvelope.getExecutionRequests())))
+            .balHash(BodyValidation.balHash(payload.getBlockAccessList().getBytes()))
+            .slotNumber(toLong(payload.getSlotNumber()))
+            .blockHeaderFunctions(new MainnetBlockHeaderFunctions())
```
tbenr, line 75: do you have an idea if this could break some non-mainnet networks?

mehdi-aouadi: Good catch! After checking, the `blockHeaderFunctions` is not used at all in the `BlockHeader` RLP encoding when calling `blockHeader.writeTo(rlpOutput)`: https://github.com/besu-eth/besu/blob/9518231a6ac2d9be26d63d500149045b760e6fc8/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeader.java#L274

The `blockHeaderFunctions(new MainnetBlockHeaderFunctions())` is basically there to satisfy the non null validation in the `BlockHeaderBuilder`: https://github.com/besu-eth/besu/blob/9518231a6ac2d9be26d63d500149045b760e6fc8/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/core/BlockHeaderBuilder.java#L356

Plus, even if it were used, the `MainnetBlockHeaderFunctions` looks to be a standard for all PoS networks and `MainnetBlockHeaderFunctions.hash()` is:
```
public static Hash createHash(final BlockHeader header) {
    final Bytes rlp = RLP.encode(header::writeTo);
    return Hash.hash(rlp);
  }
```
https://github.com/besu-eth/besu/blob/9518231a6ac2d9be26d63d500149045b760e6fc8/ethereum/core/src/main/java/org/hyperledger/besu/ethereum/mainnet/MainnetBlockHeaderFunctions.java#L33

tbenr (approved):
LGTM

PR #10939, opened 2026-07-06
Handle worker failure in AggregatingSignatureVerificationService

zilm13:
## PR Description

  A `RuntimeException` from batch BLS verification (e.g. an aggregate public key at infinity) escaped the worker loop and terminated the thread — before this fix a failed worker was never reused, so verification could stall until restart when all workers are damaged.
  The worker loop now catches such errors and falls back to verifying each task individually, keeping the thread alive and failing only the offending task in isolation.
  Already-completed tasks are skipped so the expensive verification isn't repeated; adds tests covering worker survival and the skip/isolation behaviour.

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches consensus-critical async BLS verification; incorrect fallback could mis-handle futures, but behavior is narrowly defensive with new regression tests.
> 
> **Overview**
> **Batch BLS verification** in `AggregatingSignatureVerificationService` no longer lets a `RuntimeException` (e.g. malformed inputs or aggregate key at infinity) escape the worker loop and permanently kill the thread.
> 
> The worker **`run()`** loop now catches batch failures, logs them, and falls back to **`verifyTasksIndividually`**, which verifies pending tasks with simple per-task BLS checks, skips tasks whose futures are already done, and completes only the failing task exceptionally via new **`completeExceptionallyAsync`** on **`SignatureTask`**.
> 
> Tests cover worker survival after a throwing batch and the skip/isolate behavior for mixed valid, completed, and malformed tasks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89612c05a77522c3dd81b50b01c721018b44a502. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/signatures/AggregatingSignatureVerificationService.java
+++ b/ethereum/statetransition/src/main/java/tech/pegasys/teku/statetransition/validation/signatures/AggregatingSignatureVerificationService.java
@@ -163,7 +163,34 @@ public class AggregatingSignatureVerificationService extends SignatureVerificati
     while (isRunning()) {
       final List<SignatureTask> tasks = waitForBatch();
       if (!tasks.isEmpty()) {
-        batchVerifySignatures(tasks);
+        try {
+          batchVerifySignatures(tasks);
+        } catch (final RuntimeException ex) {
+          // Batch verification can throw on malformed input (for example a BLS aggregate public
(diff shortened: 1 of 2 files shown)


tbenr (approved):
LGTM

PR #10946, opened 2026-07-07
Ssz additional optimizations

tbenr:
various optimizations for progressive and non-progressive ssz

it also contains some refactoring (moved some common logic into ListSchemaUtil)

added a benchmark to compare perf between progressive and non-progressive (called "bounded")

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

## Changelog

- [ ] I thought about adding a changelog entry, and added one if I deemed necessary.

---

> [!NOTE]
> **Medium Risk**
> Changes SSZ merkleization and deserialization for production execution payload transaction lists; behavior is heavily tested but incorrect roots or wire handling would affect consensus-critical data.
> 
> **Overview**
> Adds SSZ performance work aimed at **execution transaction lists** and **progressive** types, plus a JMH suite comparing **bounded** vs **progressive** representations.
> 
> **Packed `List[ByteList, M]` backing:** New `SszPackedByteListsHint` and `SszPackedByteListsNode` keep the list’s SSZ variable part in one buffer, compute `hashTreeRoot` by streaming merkleization, and materialize elements on demand; mutations decay to a full tree. `AbstractSszListSchema` wires hint-aware create, serialize, and deserialize. **Execution payload** `transactions` (Bellatrix through Gloas) and Heze **InclusionList** `transactions` now use `SszSchemaHints.sszPackedByteLists()`.
> 
> **Tree and deserialize refactors:** `ZeroPaddedSpineNode` compresses sparse left-aligned list trees; `TreeUtil.createTree` uses it when leaves fit a shallower depth. Packed progressive primitive lists share `ListSchemaUtil.createPackedProgressiveListTree`; boolean/bit validation runs per chunk via public `createNodeFromSszBytes`.
> 
> **Benchmarks and tests:** New `ProgressiveSszBenchmark` JMH class; unit/spec tests assert packed roots, SSZ round-trips, and oracle parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e30d136e25e541b3527def3c783f38c9b9f96c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #10951, opened 2026-07-08
implement SszPackedProgressiveByteLists hint

tbenr:
This ports the hint introduced in https://github.com/Consensys/teku/pull/10946 to progressive list

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes SSZ merkleization, serialization, and mutation paths for progressive list-of-byte-lists; incorrect roots or decay behavior would break consensus data handling, though behavior is heavily tested against an unhinted oracle.
> 
> **Overview**
> Adds **`SszPackedByteListsHint`** support for **`ProgressiveList[ProgressiveByteList]`**, mirroring the fixed-list packed optimization: the data subtree can stay as serialized SSZ bytes instead of a fully materialized progressive merkle tree until reads or mutations need it.
> 
> **`SszPackedProgressiveByteListsNode`** implements virtual progressive navigation and streaming **`hashTreeRoot`** without materializing elements; updates **decay** to a materialized tree via **`materialize()`**, which **`SszMutableProgressiveListImpl`** uses before applying chunk updates. **`AbstractSszProgressiveListSchema`** wires create/deserialize/serialize/size for the packed path and validates hint pairing (progressive byte-list elements only; mutually exclusive with super-node hint). **`AbstractSszListSchema`** now delegates pack/parse to shared **`PackedByteListsUtil`**, and fixed lists reject progressive byte-list elements for the hint so schemas cannot be mixed up.
> 
> Tests assert oracle parity (roots, SSZ, merkle proofs, mutation decay, store/load) against unhinted progressive lists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f87584b7983e76e28a5c5fff28268c58b4e5a037. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rolfyone:
claude:
  Correctness Issues
  
  1. Overflow fix in SszPackedByteListsNode.byteChunksRoot (line 387) — bundled bug fix, correct

  - final int firstByte = firstChunk * LeafNode.MAX_BYTE_SIZE;
  + final long firstByte = (long) firstChunk * LeafNode.MAX_BYTE_SIZE;

  MAX_BYTE_SIZE = 32, so firstChunk > 67M overflows int. Good catch to fix in the same PR that reuses this method via package-level access.

  2. Double-checked locking on cachedHash — safe, but subtly redundant

  // hashTreeRoot()
  if (cachedHash == null) {
      cachedHash = BranchNode.super.hashTreeRoot();  // calls hashTreeRoot(Sha256) below
      this.cachedHash = cachedHash;
  }

  // hashTreeRoot(Sha256)
  if (cachedHash == null) {
      cachedHash = spineRoot(sha256, 0);
      this.cachedHash = cachedHash;
  }

  hashTreeRoot() calls through BranchNode.super.hashTreeRoot() which dispatches to hashTreeRoot(Sha256) — that method writes this.cachedHash itself, then hashTreeRoot() writes it again. The value is always identical (immutable Bytes32, volatile
  field), so there's no correctness problem, just redundant assignment in one code path. This is exactly the pattern in SszPackedByteListsNode, so it's intentional.

  3. spineRoot recursion — bounded correctly

  The recursion terminates when levelStartSlot(level) >= getElementCount(). Since cumulativeCapacity grows as (4^(L+1)-1)/3, realistic element counts reach the termination condition well before MAX_LEVEL. Progressive lists in Ethereum spec are
  always small (e.g., blob commitments capped at 6), so no stack risk.

  4. Lazy element validation in the progressive path

  parseUnboundedPackedOffsets correctly skips per-element size bounds (progressive lists are unbounded), but element-level malformation (e.g., invalid SSZ inside an element) is deferred to access time. Structural validity (monotonicity, alignment,
  truncation) is still validated eagerly. The test deserialize_shouldRejectMalformedInputLikeOracle confirms this matches the unhinted oracle's rejection behavior for structural problems.

  ---
  Minor Observations
  
  5. SszSchemaHint visibility change appears unnecessary

  - private static class SszSchemaHint {}
  + static class SszSchemaHint {}

  SszPackedByteListsHint and SszSuperNodeHint are both public static final class. Instantiating them in tests works regardless of whether their base class is private or package-private within SszSchemaHints. This change slightly expands the API
  surface of the schema package without an obvious need. Worth removing unless there's a downstream compilation dependency I'm not seeing.

  6. AbstractSszProgressiveListSchema.sszDeserializePacked — no max-count guard

  parseUnboundedPackedOffsets passes Long.MAX_VALUE as both maxElementCount and maxElementSize, so a crafted input with a valid offset table claiming millions of elements would be accepted. Since all offsets must fit within the actual byte buffer
  (endOffset), actual memory usage is bounded by the input size, not the count. This is correct; just noting that there's no schema-level cap enforced at deserialization.

  ---
  Test Coverage
  
  The two new test classes are comprehensive:

  - SszPackedProgressiveByteListsNodeTest: Covers hash correctness at all progressive level boundaries (1, 2, 4, 5, 6, 20, 21, 22, 85, 86, 100 elements), multiple element sizes including chunk-crossing sizes (32, 33, 672 bytes), non-materialization
  of elements during hashTreeRoot, element/spine navigation via get(), Merkle proof correctness, and mutation decay.
  - SszProgressivePackedByteListsSchemaTest: Covers hint validation, rejection of incompatible schemas (bounded SszByteListSchema on the progressive path — important given SszProgressiveByteListSchema implements SszByteListSchema), roundtrip
  serialization, full decay after mutation (with tree traversal assertion), and store/load roundtrip.

  The mutation-decay test asserting no Virtual* or packed nodes survive committed trees is particularly thorough.

  ---
  Summary
  
  This is a solid PR. The one concrete suggestion is to remove the SszSchemaHint visibility change unless it's actually needed. Everything else — structure, algorithm correctness, overflow fix, and test coverage — looks good.

rolfyone (approved):

PR #10969, opened 2026-07-14
Two related bugs in DataColumnSidecarsByRoot RPC handling:

rolfyone:
Bug 1 — IllegalArgumentException for pre-Fulu epochs (Spec.java):
  isAvailabilityOfDataColumnSidecarsRequiredAtEpoch called SpecConfigFulu.required(atEpoch(epoch).getConfig()). For epochs before FULU_FORK_EPOCH, atEpoch(epoch) returns a pre-Fulu SpecVersion whose config returns Optional.empty() from
  toVersionFulu(), causing required() to throw IllegalArgumentException. This propagated through the SafeFuture chain to PeerRequiredLocalMessageHandler.handleError, which sends SERVER_ERROR (code 2) to the requesting peer and logs a full stack
  trace per request — enabling log-spam amplification by any connected peer.

  Bug 2 — Wrong error code for out-of-window block roots (DataColumnSidecarsByRootMessageHandler.java):
  The handler returned INVALID_REQUEST (code 1) when a block root references a slot outside the data column serve range. The Fulu p2p spec requires ResourceUnavailable (code 3) for this case. The blob sidecar equivalent
  (BlobSidecarsByRootMessageHandler) already returned code 3 correctly.

  Fixes:

  - Spec.isAvailabilityOfDataColumnSidecarsRequiredAtEpoch: extract the Fulu config once via getSpecConfigFulu() and add an early return false when epoch < FULU_FORK_EPOCH, implementing the spec's max(..., FULU_FORK_EPOCH) lower bound without touching atEpoch(epoch).
  - DataColumnSidecarsByRootMessageHandler.validateMinimumRequestEpoch: change the error response from INVALID_REQUEST (1) to RESOURCE_UNAVAILABLE (3), matching the spec and aligning with BlobSidecarsByRootMessageHandler.

  Tests:

  - Updated existing shouldSendResourceUnavailableIfBlockRootReferencesBlockEarlierThanTheMinimumRequestEpoch to assert code 3 (was asserting code 1).
  - Added shouldSendInvalidRequestIfBlockRootReferencesBlockEarlierThanFuluForkEpoch: sends a request with a pre-Fulu block root (slot 0, epoch 0, before FULU_FORK_EPOCH = 1) and asserts code 3 is returned rather than crashing with code 2. Skipped for GLOAS where FULU_FORK_EPOCH = 0 and no pre-Fulu epochs exist.

## Documentation

- [ ] I thought about documentation and added the `doc-change-required` label to this PR if updates are required.

[...]

PR #10970, opened 2026-07-14
Fix DataColumnSidecarsByRoot partial-response-then-abort for out-of-window block roots

rolfyone:
Summary:

  When a multi-root DataColumnSidecarsByRoot request contained a mix of in-window and out-of-window block roots (e.g. [A in-window, B out-of-window, C in-window]), Teku would serve A's sidecars, then hit B's epoch validation failure, abort the
  entire future chain, and write an error chunk onto the same libp2p stream that already had success chunks — leaving C unserved. This matches neither option the Fulu p2p spec permits: an upfront error-3 response, or silently skipping the
  unavailable sidecar and continuing to serve the rest.

  Fix:

  Removed validateMinimumRequestEpoch (which returned a failed future that propagated and aborted the chain) and inlined the epoch check directly into retrieveAndRespondForBlockRoot. When a block root is outside the data column serve range — either
  before FULU_FORK_EPOCH or beyond MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS — the method now returns SafeFuture.completedFuture(0L), silently skipping that root and continuing to process the remaining identifiers in the request. This matches the
  spec's "not include the data column sidecar in the response" option.

  Tests:

  - shouldSilentlySkipBlockRootOutsideMinimumRequestEpoch: all-out-of-window request completes successfully with no sidecars served and no error response (previously asserted an error response).
  - shouldSilentlySkipBlockRootBeforeFuluForkEpoch: pre-Fulu block root is silently skipped, completes successfully (previously asserted an error response). Skipped for GLOAS where FULU_FORK_EPOCH = 0.
  - shouldServeInWindowRootsAndSkipOutOfWindowRootInMixedRequest (new): [A in-window, B out-of-window, C in-window] → serves A and C, skips B, completes successfully with no error response.

  Depends on: #10969 (pre-Fulu epoch guard in Spec.isAvailabilityOfDataColumnSidecarsRequiredAtEpoch).

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Changes Fulu req/resp behavior on a hot P2P path; incorrect skipping or serving could break peer sync, but scope is limited to epoch-guard handling and is well covered by new tests.
> 
> **Overview**
> Fixes **DataColumnSidecarsByRoot** so mixed requests no longer abort after serving some roots: out-of-window block roots are **silently skipped** instead of returning `RESOURCE_UNAVAILABLE`, matching Fulu’s “omit unavailable sidecars and continue” behavior.
> 
> `validateMinimumRequestEpoch` is removed; the epoch check is inlined in `retrieveAndRespondForBlockRoot`, returning `0` served when the root is before Fulu or outside `MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS`. Rate limiting still adjusts for the count actually served.
> 
> Tests now expect successful completion with no error chunk for all-out-of-window and pre-Fulu roots, and add a mixed **in-window / out-of-window / in-window** case that serves only the valid roots.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a6725e6723363957c15d70bb48dce1e08069306. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tbenr (dismissed):
LGTM

gfukushima (approved):

Issue #8731, opened 2024-10-15
Feature: One password file for all validator


maximus321:
### Description
It would be nice to have a way to only use one validator password file for all my validator. 
With a high number of validators, that is easy to achieve now with lido csm, we can have an overwheliming number of .txt file that contains the same information (not optimal)

It would be nice to have an option like
`--single-password-file:pwd.txt` 

As discussed on discord here
https://discord.com/channels/697535391594446898/697539289042649190/1295781894008471604

### Steps to Reproduce (Bug)
[Please be as specific as possible]

**Expected behavior:** [What you expect to happen]

**Actual behavior:** [What actually happens]

**Frequency:** [How regularly does it occur?]

### Versions (Add all that apply)
* Software version: [`teku --version` or look for log lines starting with `Teku version:`]
* Java version: [`java -version`]
* OS Name & Version: [`cat /etc/*release`]
* Docker Version: [`docker version`]
* Cloud VM, type, size: [Amazon Web Services I3-large]

rolfyone:
Hey thanks for raising this. We'll see what we can do to improve the situation.

We do similar too for our validators so I feel like it's a use-case we have worked around and could probably improve, we just may need to specify the keys differently for this new scenario or something, but we'll schedule some time to do some design.

Just for context from others browsing, currently teku supports basically a list of key:password entries
https://docs.teku.consensys.io/reference/cli#validator-keys
which may be pairs of directories or pairs of files, 
and we'd be changing to a semantic of specifying the keys in one parameter (files or directory most likely) and taking the password from a separate entry.

The validator-keys cli arg is very commonly used currently, so a solution where current behaviour remains valid would be required. This may mean we end up using a different flag to specify just the keys, and a flag for the one password file, but I won't preclude the design process by 'solutioning' just yet.

Matilda-Clerke:
As discussed with @rolfyone and @lucassaldanha I will outline below the current functionality available

1. Specify a key file and matching password file
2. Specify a key directory and matching password directory (key files and password files must be appended with `.json` and `.txt` respectively)
3. Ability to specify mixed multiples of the above (e.g. multiple individual key files with matching password files in addition to multiple key directories with matching password directories)
 
[...]

PR #10990, opened 2026-07-21
Allow validator key directory to use single password file

Matilda-Clerke:
## PR Description
Allow `--validator-keys` to be specified as a directory of keys and a single password file, resulting in that singular password file being used for all of the keys in the directory.

## Fixed Issue(s)
https://github.com/Consensys/teku/issues/8731

## Documentation

## Changelog

---

> [!NOTE]
> **Low Risk**
> Localized validator startup configuration parsing with backward-compatible directory-to-directory behavior; no changes to signing, networking, or consensus logic.
> 
> **Overview**
> **`--validator-keys`** now supports **`&lt;KEY_DIR&gt;:&lt;PASS_FILE&gt;`**, so every `*.json` keystore under the key directory uses one shared password file instead of per-key `.txt` files in a password directory.
> 
> `KeyStoreFilesLocator` walks the key directory as before, but when the password side is a **file** it maps each keystore to that path; when it is a **directory**, behavior is unchanged (matching `.txt` per keystore). Mixing a single key file with a password **directory** is rejected with a clearer error instead of the old “both must be directories or both files” rule.
> 
> CLI help in `ValidatorKeysOptions` and the changelog document the new form; tests cover directory+shared password, mixed argument styles, and the invalid key-file/password-dir case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 788a4ec87bd9d07726fbb1a5811d31ea54c88700. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rolfyone:
i think this is likely to be much less open to regression if we leave this arg alone and implement new functionality, rather than complicate this arg...

Conceptually,  a new arg for 
--validator-keys-path - a directory to scan (must be accompanied by --validator-keys-password)
--validator-keys-password - password file that applies to all keystores found only if using the keys-path

Changing the existing param like above does cause a lot of extra complexity and space for regression, so i think we should probably avoid that...

at least one concern claude raised 
```
  Single-file branch skips the isFile() validation that the directory branch performs — KeyStoreFilesLocator.java:162
  
  The dir:dir branch at ~line 150 has an explicit if (!passwordFile.isFile()) throw InvalidConfigurationException(...). The new else branch just assigns passwordFile = passwordFileOrDirectory with no validation. A FIFO, device file, or socket passes
  passwordFile.exists() (checked earlier in parseEntry) but is not a regular file. Every keystore then attempts to read its password from the special file, blocking or producing garbage, with no user-readable error. The fix is one line: add the
  same isFile() guard in the else branch.
```

If its new cli etc we're not potentially regressing the existing functions, and potentially could even see clients using both settings. (entire directory and pass, and also a couple of keys from elsewhere) and it wouldn't cause issues.

Matilda-Clerke:
Aside from the comment from claude, which I will address shortly, I'm not sure I agree. This code is pretty comprehensively tested and the dir:dir and file:file flows remain unchanged. Introducing new CLI arguments has the potential to confuse users about which they should, and whether they can combine seemingly conflicting arguments. This solution also allows the mixing of dir:dir, dir:file, and file:file (I'll produce another unit test to show and verify this) in a more intuitive way than a separate argument.

lucassaldanha:
I am not a big fan of the already overloaded `--validator-keys` parameter. The syntax using ':' to separate the two parameters just makes it convoluted.

I think I'd take this opportunity to add the options `--validator-keys-path` and `--validator-keys-password` options, and have them support the following usages:

(current use)
--validator-keys-path=/foo          (loads all keys from /foo) 
--validator-keys-password=/bar      (loads all passwords from /bar) - what we do today

(new use case)
--validator-keys-path=/foo          (loads all keys from /foo) 
--validator-keys-password=/pwd.txt  (single password for all keys in /foo)

This way, most of the underlying logic can be reused as it is, and we can have the processing of the single password file added with low effort. All this code does is dir/file reading, so it isn't a very complicated thing to test. I'd be surprise to see any regression in the code (assuming we have the proper coverage now).

Users can keep using `--validator-keys` for now, but in the future, I'd look at deprecating `--validator-keys` (out of scope for now).

rolfyone (approved):
approving on the basis of lack of stakeholder input and discussions in team.

PR #11021, opened 2026-07-29
Fix edge case for BlocksByRange request

### LINKED ISSUE #10985: Teku BeaconBlocksByRange v2 single-block requests return empty responses
## Summary

Teku returns an empty successful response to `/eth2/beacon_chain/req/beacon_blocks_by_range/2/ssz_snappy` when the request asks for one available block in a hot range:

```text
BeaconBlocksByRange v2 request:
  start_slot = <available hot block slot>
  count      = 1
  step       = 1

Observed Teku P2P response:
  EMPTY
```

The Teku node has the block via Beacon API, and requesting a wider range that starts at the same slot returns the block. The differential testing results show that Prysm, Lighthouse, Nimbus, Lodestar, and Grandine return the block, only Teku returning `EMPTY`.

This appears to be a bug in Teku's inbound req/resp handler.

## Spec Ref
The [consensus-spec](https://github.com/ethereum/consensus-specs/blob/152790136f5f9c0e7504e14dcb5c381deb32db53/specs/phase0/p2p-interface.md?plain=1#L1436) defines `BeaconBlocksByRange` as requesting blocks in the half-open slot range `[start_slot, start_slot + count)`. For `count=1`, the requested range contains exactly `start_slot`.

The [spec](https://github.com/ethereum/consensus-specs/blob/152790136f5f9c0e7504e14dcb5c381deb32db53/specs/phase0/p2p-interface.md?plain=1#L1478-L1479) also have:
> Clients must respond with at least the first block that exists in the range if they have it. 

Teku has it, so `EMPTY` violates the spec.

## Root Cause
I think this is related to #8341. The root cause is clear:

https://github.com/Consensys/teku/blob/48f92b61acb40dfcb526987d0cb952080b0daf7d/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java#L185-L200

For `count=1`, `getAncestorRoots(start_slot, step, count)` returns a single root, so `hotRoots.lastKey() == start_slot` and `headSlot == start_slot`. `initialState.isComplete()` then returns true before any block is sent, so Teku closes the stream empty.

The unit test also assume this to pass.

https://github.com/Consensys/teku/blob/302162dad0b9f0547e89f01c9071ce8a043a265d/networking/eth2/src/test/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandlerTest.java#L340-L355

## Repro
Pick a recent, unfinalized slot that Teku has a block for (confirm via `GET /eth/v2/beacon/blocks/{slot}`). Open a stream to Teku on `/eth2/beacon_chain/req/beacon_blocks_by_range/2/ssz_snappy` and send `{ start_slot = that slot, count = 1, step = 1 }`.

- Expected: one block chunk for that slot. 
- Actual: an empty successful response.

Then send the same `start_slot` with `count = 2`. Teku returns the blocks, showing it has them and only the single-slot boundary fails.

Thanks for your attention!

gfukushima:
## PR Description
This PR fixes an edge case where we would return empty when requested a single block. Thanks @Alleysira for reporting this.

## Fixed Issue(s)

Fix #10985 

## Documentation

## Changelog

---

> [!NOTE]
> **Low Risk**
> Narrow P2P RPC handler change with targeted unit and integration tests; no auth, data migration, or broad API surface impact.
> 
> **Overview**
> Fixes **BeaconBlocksByRange** returning an empty stream when a peer asks for **one block** at the **head (hot) slot** (`start_slot == head`, `count == 1`, `step == 1`), which violated spec expectations (#10985).
> 
> **`BeaconBlocksByRangeMessageHandler`** no longer treats `RequestState.isComplete()` at setup as “send nothing.” It only skips block iteration when **`count` is zero** or **`startSlot` is strictly past `headSlot`**. When `startSlot` equals the head, it still runs **`sendNextBlock`** so the block at that slot is returned.
> 
> Coverage: unit test mocks **`getAncestorRoots`** for the real `count=1` / `startSlot == headSlot` case, plus an integration regression that requests the head slot with `count=1`. Changelog entry added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 832c56e23308cb39040075898cbfa526f95e5f28. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
@@ -196,7 +196,7 @@ public class BeaconBlocksByRangeMessageHandler
               // or a start slot strictly beyond our head. When startSlot == headSlot we must still
               // return the block at that slot, since the spec requires returning at least the first
               // block in the requested range if we have it.
-              if (count.equals(ZERO) || startSlot.isGreaterThan(headSlot)) {
+              if (count.isZero() || startSlot.isGreaterThan(headSlot)) {
                 return SafeFuture.completedFuture(initialState);
               }
               return sendNextBlock(initialState);
--- a/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
+++ b/networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
@@ -192,9 +192,11 @@ public class BeaconBlocksByRangeMessageHandler
               final UInt64 headSlot = hotRoots.isEmpty() ? headBlockSlot : hotRoots.lastKey();
               final RequestState initialState =
                   new RequestState(startSlot, step, count, headSlot, hotRoots, callback);
-              // there is an edge case when startSlot == headSlot in which case we don't return
-              // anything
-              if (initialState.isComplete()) {
+              // Only short-circuit when there is genuinely nothing to send: a zero-length request
+              // or a start slot strictly beyond our head. When startSlot == headSlot we must still
+              // return the block at that slot, since the spec requires returning at least the first
+              // block in the requested range if we have it.
+              if (count.equals(ZERO) || startSlot.isGreaterThan(headSlot)) {
                 return SafeFuture.completedFuture(initialState);
               }
               return sendNextBlock(initialState);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,3 +15,4 @@
 
 ### Bug Fixes
  - Fixed Beacon REST API socket retention when clients cancel pending asynchronous requests. Requests now time out after 30 seconds.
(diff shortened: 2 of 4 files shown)


networking/eth2/src/main/java/tech/pegasys/teku/networking/eth2/rpc/beaconchain/methods/BeaconBlocksByRangeMessageHandler.java
```diff
@@ -192,9 +192,11 @@ private SafeFuture<RequestState> sendMatchingBlocks(
...
               final RequestState initialState =
                   new RequestState(startSlot, step, count, headSlot, hotRoots, callback);
-              // there is an edge case when startSlot == headSlot in which case we don't return
-              // anything
-              if (initialState.isComplete()) {
+              // Only short-circuit when there is genuinely nothing to send: a zero-length request
+              // or a start slot strictly beyond our head. When startSlot == headSlot we must still
+              // return the block at that slot, since the spec requires returning at least the first
+              // block in the requested range if we have it.
+              if (count.equals(ZERO) || startSlot.isGreaterThan(headSlot)) {
```
tbenr: here you can use `isZero`

tbenr (dismissed):
LGTM

lucassaldanha (approved):

PR #11028, opened 2026-07-29
Fix isAttestationSameSlot in GLOAS

tbenr:
fixes https://github.com/Consensys/teku-internal/issues/305

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches GLOAS attestation head-matching and participation logic; the change only tightens validation order for same-slot attestations but could reject blocks that previously slipped through without the index check.
> 
> **Overview**
> **GLOAS `computeIsMatchingHead` now validates same-slot attestation index before target/head checks.**
> 
> For attestations classified as same-slot via `isAttestationSameSlot`, the **index must be zero** is enforced with `checkArgument` **up front**, instead of only after target and head roots already match. Invalid same-slot attestations with a non-zero index therefore fail immediately rather than being treated as a non-matching head and skipping validation.
> 
> Control flow is refactored so the early return on failed target/head match happens after that check; payload-index matching for non–same-slot attestations is unchanged.
> 
> A unit test asserts that `computeIsMatchingHead(false, false, …)` still throws when slot is zero and index is non-zero.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4aa10edc62568d93086ca235f7a4e0725d279584. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/gloas/helpers/BeaconStateAccessorsGloas.java
+++ b/ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/gloas/helpers/BeaconStateAccessorsGloas.java
@@ -318,18 +318,23 @@ public class BeaconStateAccessorsGloas extends BeaconStateAccessorsFulu {
       final boolean headRootMatches,
       final AttestationData data,
       final BeaconState state) {
+    final boolean isAttestationSameSlot = isAttestationSameSlot(state, data);
+    if (isAttestationSameSlot) {
+      checkArgument(data.getIndex().isZero(), "Index must be set to zero");
+    }
+
     if (!isMatchingTarget || !headRootMatches) {
       return false;
     }
-    if (isAttestationSameSlot(state, data)) {
-      checkArgument(data.getIndex().isZero(), "Index must be set to zero");
+
+    if (isAttestationSameSlot) {
       return true;
-    } else {
-      final int slotIndex = data.getSlot().mod(config.getSlotsPerHistoricalRoot()).intValue();
-      final boolean payloadIndex =
-          BeaconStateGloas.required(state).getExecutionPayloadAvailability().get(slotIndex).get();
-      return data.getIndex().intValue() == (payloadIndex ? 1 : 0);
     }
+
+    final int slotIndex = data.getSlot().mod(config.getSlotsPerHistoricalRoot()).intValue();
+    final boolean payloadIndex =
+        BeaconStateGloas.required(state).getExecutionPayloadAvailability().get(slotIndex).get();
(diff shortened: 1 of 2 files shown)


ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/gloas/helpers/BeaconStateAccessorsGloas.java
```diff
@@ -318,18 +318,23 @@ protected boolean computeIsMatchingHead(
       final boolean headRootMatches,
       final AttestationData data,
       final BeaconState state) {
+    final boolean isAttestationSameSlot = isAttestationSameSlot(state, data);
```
zilm13, line 321: from my point of view it could be more readable if we just copy the spec:
```java
    final boolean isAttestationSameSlot = isAttestationSameSlot(state, data);
    final boolean payloadMatches;
    if (isAttestationSameSlot) {
      checkArgument(data.getIndex().isZero(), "Index must be set to zero");
      payloadMatches = true;
    } else {
      final int slotIndex = data.getSlot().mod(config.getSlotsPerHistoricalRoot()).intValue();
      final boolean payloadIndex =
          BeaconStateGloas.required(state).getExecutionPayloadAvailability().get(slotIndex).get();
      payloadMatches = data.getIndex().intValue() == (payloadIndex ? 1 : 0);
    }

    return isMatchingTarget && headRootMatches && payloadMatches;
```

zilm13 (approved):
LGTM

ethereum/spec/src/main/java/tech/pegasys/teku/spec/logic/versions/gloas/helpers/BeaconStateAccessorsGloas.java
tbenr: Agreed. Applying.

update: but it wont short-circuit 🤔

zilm13: it will short circuit compiled, it's not an assembler
we should definitely prioritize code readability in codebase of hundred of thousand lines

tbenr: > it will short circuit compiled, it's not an assembler we should definitely prioritize code readability in codebase of hundred of thousand lines

I see but those optimizations are kind of a blackbox and you never know how far they go. A side project could be to have some kind of plugin showing how your code get optimized by compiler and JIT in different conditions :)

PR #11032, opened 2026-07-30
Implement 5473 spec changes

tbenr:
https://github.com/ethereum/consensus-specs/pull/5473

bumps to reftests to alpha.13 too

## Documentation

## Changelog

---

> [!NOTE]
> **Medium Risk**
> Touches core attestation participation, Gloas block processing order, and proposer reward math; mistakes could mis-score attestations or rewards, though coverage is expanded in unit tests.
> 
> **Overview**
> Implements **consensus-specs PR 5473** (ref tests **v1.7.0-alpha.13**) for **Gloas** attestation timing and payload availability.
> 
> **Gloas attestation / reward path:** Attestation processing and participation flags now take an explicit **`parentSlot`**. On Gloas, the attestation parent slot comes from **`latest_execution_payload_bid`**, and **timely head** uses **`execution_payload_availability`** indexed by that parent slot (not the attestation data slot). **`process_execution_payload_bid`** returns the previous bid’s slot so attestations run after the bid step with the correct parent. **`Spec.getStateForAttestationRewardCalculation`** (Gloas: optionally projects the parent availability bit onto a copy of state) is used for **block production attestation selection**, **block reward calculation**, and related paths when the parent payload is **FULL**.
> 
> **Block production:** `BlockOperationSelectorFactory` passes the projected reward state into the attestation pool when the fork-choice parent has full payload status.
> 
> Reference tests add **`fork_choice/should_apply_proposer_boost`** with one case skipped pending equivocation support.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf26c699b8fe71887f2c5fa136ee07581125a318. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mehdi-aouadi:
There could be a side effect in the `BlockRewardCalculatorUtil` since it uses the pre block state without processing the parent payload first. So after a skipped slot, the parent slot availability is still false even when the parent payload is `FULL`. I guess we should rather apply the parent payload availability to the calculation state before processing attestations. so basically we can use the state copy, apply parent payload availability and calculate the attestations rewards from it

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -1288,6 +1288,13 @@ public Optional<Integer> getBuilderIndex(final BeaconState state, final BLSPubli
   }
 
   // Attestation helpers
+  public BeaconState getStateForAttestationRewardCalculation(
```
mehdi-aouadi, line 1291: I'd suggest renaming it to `getStateForAttestationEvaluation` since it's also used for attestation packing

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
```diff
@@ -1288,6 +1288,13 @@ public Optional<Integer> getBuilderIndex(final BeaconState state, final BLSPubli
...
+      final BeaconState state, final boolean parentPayloadAvailable) {
```
mehdi-aouadi, line 1292: We could also remove the `parentPayloadAvailable` since it leaks Gloas semantics.
We can conditionally call it from `BlockRewardCalculatorUtil` and `BlockOperationSelectorFactory` only when the parent payload is FULL:

```
final BeaconState evaluationState =
    parentPayloadAvailable
        ? spec.withParentPayloadAvailableForAttestationEvaluation(state)
        : state;
```

The common `AttestationUtil` implementation can return the original state, while the Gloas takes into account the parent slot availability bit into a copy

mehdi-aouadi (approved):
LGTM
Just nits

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
tbenr: well.. attestation packing requires it because it has to sort based on rewards, so I actually prefer the current name. Feel free to push back again if you feel strongly

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
tbenr: pushing back on this too: I think this moves the semantic to every caller rather than removing it. The callers necessarily determine parent availability from different sources, but the fork-specific utility should own how both true and false map to an evaluation state. Keeping the call unconditional also avoids duplicating the branching policy and preserves the option for future fork logic to handle the unavailable case differently.

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
mehdi-aouadi: definitely not a strong opinion

ethereum/spec/src/main/java/tech/pegasys/teku/spec/Spec.java
mehdi-aouadi: good to go, sounds good

